PR #15, opened 2020-06-19
doc/async-ipc: draft

cbiffle:
Here's a manual-style discussion of the feature I'm proposing to add. I'd like feedback on the design (and its description, but mostly the design).

The "Rich Diff" feature (tiny page-with-folded-corner icon at the top of the diff below) shows the rendered version.


cbiffle:
We've managed to avoid needing async IPC for the past couple of years... let's keep trying that.

PR #98, opened 2020-11-19
doc/netstack: initial proposal

cbiffle:


cbiffle:
Obviated by later netstack work.

PR #123, opened 2021-01-04
flashing on Gemini should pull correct OpenOCD config file

bcantrill:
Currently, the Gemini `cargo xtask flash` pulls in the `openocd.cfg` for the H7 demo target.  We have our own `openocd.cfg` for Gemini, so this fix makes sure we pull in the right one.  (This becomes more than a theoretical issue when both a Gemini board and a Nucleo-H743 are attached and one wishes to use probe serial number to make sure that OpenOCD flashes the right board!)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,20 @@ name: build
 on: [push, pull_request]
 
 jobs:
+  skip_duplicate_jobs:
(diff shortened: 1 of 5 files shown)


cbiffle:
@bcantrill cleaning out the patch queue, did you want to merge this?

PR #127, opened 2021-01-10
Update AsciiDoc code blocks to use syntax highlighting

smklein:
This should be viewable under "rich diff" on GitHub.

(thanks for the great docs, btw!)


cbiffle:
This got integrated into the pre-public-release doc cleanup pass.

PR #136, opened 2021-01-27
Add RNG API for the lpc55

labbott:
This implements traits from `rand_core` so the hardware rng can
be used with APIs that need random numbers (e.g. key generation)


flihp:
closing after conversation w/ @labbott w/r/t deconflicting with #383

PR #144, opened 2021-04-01
userlib: add AtomicHolder and singleton macro

cbiffle:
Because doing static mut correctly is hard.


jamesmunns:
I'm going to go ahead and close this, I *think* the intent of this is something similar to our current `static-cell`, which already uses the armv6 atomic hack, which would be necessary to support for our current stm32g0 platforms.

Please feel free to re-open if I missed anything!

PR #149, opened 2021-04-20
gemini-bu-bare: bare metal template for Gemini SP

cbiffle:


cbiffle:
We haven't wound up wanting to Arduino our boards and let's see if we can keep it that way.

PR #159, opened 2021-05-24
Add tests for gemini-bu-rot board

labbott:
This is an lpc55 with a different memory layout


labbott:
merged as part of f8043755743e27c33e8d9f34e9265d9d4b004059

PR #180, opened 2021-09-07
Improve caching via Swatinem/rust-cache

steveklabnik:
We were trying to do this ourselves, but it has always been a bit...
suboptimal. https://matklad.github.io/2021/09/04/fast-rust-builds.html
recently points towards this Action, and it should do a better job than
we were doing, and is easier to use, to boot!

This is not ready to merge yet, I would like to verify that it actually works, which of course means running the jobs which means opening a PR.

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -77,14 +77,8 @@ jobs:
       - name: Install Rust toolchain
         run: rustup show
 
-      # cache the cargo registry & index
-      - name: Cache cargo outputs
-        uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache build output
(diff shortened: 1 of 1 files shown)


steveklabnik:
Coming back to this because apparently our caches aren't working.

image


Here's the `cleanTarget` function: https://github.com/Swatinem/rust-cache/blob/v1/src/common.ts#L193-L222

It seems to: rm:

* `path.join(targetDir, "./.rustc_info.json")`
* `path.join(targetDir, "./debug/examples")`
* `path.join(targetDir, "./debug/incremental")`
* all files from `path.join(targetDir, "./debug")`

It then fetches the list of packages, and removes everything in these directories that doesn't match:
* path.join(targetDir, "./debug/build")`
* path.join(targetDir, "./debug/.fingerprint")`

It also checks for dependencies and removes everything that's not in that list from

* `path.join(targetDir, "./debug/deps")`


That is, it does the right thing! Since we write out our stuff into `target/{board}`, this shouldn't touch any of our stuff.

kc8apf (approved):

PR #194, opened 2021-10-06
drv/stm32h7-spi-server: new config style

cbiffle:
Previously, we used conditional compilation to generate a collection of
variables in the server depending on the target board and controller. It
was not obvious what that set of variables was intended to be, or their
types. If you were adding a new board, and forgot one, you got errors
later in the file when the variable was referenced.

This commit tries to firm up the board-and-controller-specific
configuration of the server by moving the configurable bits into a
struct. We then use conditional compilation to generate a const of that
struct. This means if you forget to provide one of the fields, you get a
compile error pointing to the bad struct literal so you can go add it.

While I was formalizing the server config, I have added in two things
for near-future expansion:

1. The server now uses software-controlled CS instead of hard CS,
   allowing us (with further changes not in this commit) to address an
   arbitrary number of devices with different CS nets.

2. There are now records in the configuration for where those different
   CS nets are on each board.

### DIFF
--- a/drv/stm32h7-spi-server/src/main.rs
+++ b/drv/stm32h7-spi-server/src/main.rs
@@ -1,20 +1,8 @@
 //! Server task for the STM32H7 SPI peripheral.
 //!
-//! Currently this hardcodes the clock rate and doesn't manage chip select.
+//! Currently this hardcodes the clock rate.
 //!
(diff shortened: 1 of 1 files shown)


cbiffle:
Superceded by a different PR.

PR #250, opened 2021-11-11
Fix Sidecar app.toml to get rid of build errors.

lzrd:


cbiffle (approved):

cbiffle:
Sidecar's building on master (and in CI!) so I think this is obsolete.

PR #269, opened 2021-11-29
Update submodules to use https

steveklabnik:
We had these via git@ because they were private, now that they're
public, let's use the more usual URL.


labbott:
lpc55_support has been opened up

PR #278, opened 2021-11-30
doc/ipc: copyedits

riking:
The protocol uses leases, the Rust code uses borrows.

### DIFF
--- a/doc/ipc.adoc
+++ b/doc/ipc.adoc
@@ -272,7 +272,7 @@ A concrete mapping of these operations to IPCs might go as follows.
 
 - Message is a four-byte struct containing `fd` as a little-endian `u32`. Borrow
   0 is `buffer` and must be writable.
-- Data will be written to a prefix of borrow 0, starting at offset 0.
+- Data will be written to a prefix of lease 0, starting at offset 0.
 - On success, returns response code 0 and a four-byte response, containing the
   bytes-read count as a little-endian `u32`.
 - On failure, returns a non-zero response code that maps to an `IoError`, and a
@@ -282,7 +282,7 @@ A concrete mapping of these operations to IPCs might go as follows.
 
 - Message is a four-byte struct containing `fd` as a little-endian `u32`. Borrow
   0 is `buffer` and must be readable.
-- Data will be taken from a prefix of borrow 0, starting at offset 0.
+- Data will be taken from a prefix of lease 0, starting at offset 0.
 - On success, returns response code 0 and a four-byte response, containing the
   bytes-written count as a little-endian `u32`.
 - On failure, returns a non-zero response code that maps to an `IoError`, and a
@@ -309,7 +309,7 @@ fn read(task: TaskId, fd: u32, buffer: &mut [u8]) -> Result<usize, IoError> {
     let (rc, len) = sys_send(
         task,
         0,
-        &u32.to_le_bytes(),
+        &fd.to_le_bytes(),
         &mut response,
         &[Lease::from(buffer)],
     );
--- a/doc/timers.adoc
+++ b/doc/timers.adoc
@@ -35,7 +35,7 @@ Each task gets a timer. The timer has three properties:
 - A _notification set._
 
 At periodic intervals (ticks), if the enable bit is set, the kernel checks the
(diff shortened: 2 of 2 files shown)


doc/ipc.adoc
```diff
@@ -272,7 +272,7 @@ A concrete mapping of these operations to IPCs might go as follows.
 
 - Message is a four-byte struct containing `fd` as a little-endian `u32`. Borrow
   0 is `buffer` and must be writable.
-- Data will be written to a prefix of borrow 0, starting at offset 0.
+- Data will be written to a prefix of lease 0, starting at offset 0.
```
cbiffle, line 275: Both are technically correct here, since "borrow" is the term used at the server side, while "lease" is from the client's perspective. I'm happy to accept this change though since this section is a little informal, and the old usage clearly surprised you.

doc/ipc.adoc
```diff
@@ -309,7 +309,7 @@ fn read(task: TaskId, fd: u32, buffer: &mut [u8]) -> Result<usize, IoError> {
     let (rc, len) = sys_send(
         task,
         0,
-        &u32.to_le_bytes(),
+        &fd.to_le_bytes(),
```
cbiffle, line 312: Heh, nice.

doc/timers.adoc
```diff
@@ -35,7 +35,7 @@ Each task gets a timer. The timer has three properties:
 - A _notification set._
 
 At periodic intervals (ticks), if the enable bit is set, the kernel checks the
-deadline to see if it is in the future. If not (it is `<=` the current time),
+deadline to see if it is in the future. If not (it is &#2264; the current time),
```
cbiffle, line 38: FWIW, `<=` is a Rust operator and I'm more comfortable with that in the document than the Unicode escape for the math equivalent. Personal preference thing.

cbiffle (commented):
Hello! Thanks for fixing some stuff! I'm going to merge 2/3 of this. Comments below.

doc/timers.adoc
riking: The problem here is that asciidoc turned it into a double arrow!

fernandogmo: @riking might have noticed that the Rust operator `<=` is being rendered as <code>&#8656;</code> (`&#8656;`) which is `U+21D0`.

This is probably because GitHub renders with `asciidoctor` rather than `asciidoc`:
```
# What I think you prefer (with asciidoc):
❯ echo '`<=`' | asciidoc -s -
<div class="paragraph"><p><code>&lt;=</code></p></div>

# What you're currently getting (with asciidoctor):
❯ echo '`<=`' | asciidoctor -s -
<div class="paragraph">
<p><code>&#8656;</code></p>
</div>

# How we can fix it:
❯ echo '`\<=`' | asciidoctor -s -
<div class="paragraph">
<p><code>&lt;=</code></p>
</div>

```
We can fix the Rust operator and prevent the [character replacement substitution](https://docs.asciidoctor.org/asciidoc/latest/subs/replacements/) by using a backslash to escape the sequence, like this: `\<=`.

cbiffle: Ah! I misunderstood. Thank you @riking and @fernandogmo for walking me through that. Let's escape it as in that last suggestion.

I'm...not sure if I can edit your PR @riking? So if you'd be willing to do the `\<=` thing I'll merge.

riking: Done

doc/timers.adoc
```diff
@@ -35,7 +35,7 @@ Each task gets a timer. The timer has three properties:
 - A _notification set._
 
 At periodic intervals (ticks), if the enable bit is set, the kernel checks the
-deadline to see if it is in the future. If not (it is `<=` the current time),
+deadline to see if it is in the future. If not (it is \<=; the current time),
```
cbiffle, line 38: Sorry to round trip this again, but, `\<=;` produces `&lt;=;` in the HTML output -- this needs to be just `\<=` without the semicolon.

riking: Whoops. Done

fernandogmo: You might want to keep the backticks here.
See this example: https://gist.github.com/fernandogmo/34032124e94c6b383277decd8631c687

PR #286, opened 2021-12-03
Fix typos for flashing nucleo boards

grevian:
xtash -> xtask


cbiffle (approved):
Ha! Good catch. As you can see, our README does not have very good test coverage. ;-)

PR #287, opened 2021-12-03
Add support for A0 hardware

labbott:
The ROM API was intentionally written to only support lpc55
hardware rev 1B. It's still useful to be able to do basic verification
on 0A hardware. Add appropriate support iin the ROM API.

### DIFF
--- a/stage0/src/main.rs
+++ b/stage0/src/main.rs
@@ -43,8 +43,23 @@ pub unsafe extern "C" fn SecureFault() {
     loop {}
 }
 
+// These correspond to REV_ID in the SYSCON_DIEID field
+#[cfg(feature = "0A-hardware")]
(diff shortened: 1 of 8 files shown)


stage0/src/main.rs
```diff
@@ -45,6 +45,18 @@ pub unsafe extern "C" fn SecureFault() {
 
 #[entry]
 fn main() -> ! {
+    cfg_if::cfg_if! {
+        if #[cfg(not(feature = "a0-hardware"))] {
```
cbiffle, line 49: We probably want to do this assert in either case, changing the expected value -- because a 0A image won't work properly on 1B.

Also isn't it 0A instead of A0?

cbiffle (approved):

PR #292, opened 2021-12-04
FAQ: first draft.

cbiffle:


utzig:
... if you need a concise term to describe Hubris's design philosophy, we suggest "emokernel." :rofl:

PR #297, opened 2021-12-05
docs: Fix invalid reference to "u32.to_le_bytes()"

jdm:
The code sample doesn't make sense, since `to_le_bytes` is a method on instances of `u32`.


cbiffle:
Thanks for the close reading! @riking has fixed this in #278, we're working toward merging it.

PR #300, opened 2021-12-06
Doc: IPC - small typo fix

Guiguiprim:
Fix a small typo in a code example where a variable name was replaced by it type name.


cbiffle:
Thanks for the close reading! @riking has fixed this in #278, we're working toward merging it.

PR #303, opened 2021-12-07
README: clarify RickLink instructions on lpc55s69xpresso

kc8apf:
* Overwrite existing firmware so the new bootloader doesn't accidentally
try to boot an old main firmware
* Remove instruction to setup three terminals (pyocd gdbserver, telnet
for semihosting, gdb) as multiple folks have run into problems following
that bit.  Once RickLink and pyocd is installed, `cargo xtask flash` and
`humility itm` just work without running a gdb server.  Semihosting
isn't enabled in the lpc55s69xpresso app so no need to run a terminal
for it.  Running a pyocd gdbserver is necessary if you want to run
`cargo xtask gdb` but we don't tell people to do that.

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -257,7 +257,16 @@ $ /usr/local/lpcscrypt/bin/lpcscrypt clockslow
 $
 ```
 
-4. Run `lpcscrypt program +c <path-to-lpc4322_bl_crc.bin> BankA`:
+4. Run `lpcscrypt program +w1 0x0 BankA` to overwrite existing firmware
+
+```console
+$ /usr/local/lpcscrypt/bin/lpcscrypt program +w1 0x0 BankA
+................
+Programmed 524288 bytes to 0x1a000000 in 2.610s (196.165KB/sec)
+$
+```
+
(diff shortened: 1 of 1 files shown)


README.mkdn
```diff
@@ -260,7 +260,16 @@ $ /usr/local/lpcscrypt/bin/lpcscrypt clockslow
 $
 ```
 
-4. Run `lpcscrypt program +c <path-to-lpc4322_bl_crc.bin> BankA`:
+4. Run `lpcscrypt program +w1 0xAA BankA` to overwrite existing firmware
```
labbott, line 263: This is different than the command given in meta which is what I've been using `${LPCSCRYPT_ROOT}/bin/lpcscrypt program +w1 0x0 BANKA`
I'm guessing this is just wiping the bank so it doesn't matter what value is used?

kc8apf: It shouldn't matter but 0x0 is probably less confusing.  Ideally `lpcscrypt eraseFlash BankA` would work but it doesn't so this is the hack instead.

PR #308, opened 2021-12-08
Provide additional details for manual OpenOCD install

rmustacc:
While working through initial set up, I found myself in the place where the REAMDE suggested needing v0.11, the package manager (Ubuntu 20.04 LTS) having an 0.10.x version, and needing to build myself. Going through this, it wasn't clear what features we were relying on and so I guessed mostly that the defaults of `./configure` were correct.

I've added a bit more prose to try and help make this a little smoother for the next person. We may want something similar for how to set up pyocd at some point.


steveklabnik (approved):
Thank you!

PR #309, opened 2021-12-08
Fix OpenOCD link

### LINKED ISSUE #275: "getting OpenOCD" link is broken
On this page

https://github.com/oxidecomputer/hubris

The "getting OpenOCD" link is broken and generates a 404 error.

The link is to

http://openocd.org/getting-openocd/

But should be to

https://openocd.org/pages/getting-openocd.html

I think that the OpenOCD site/pages were reorganized recently which probably caused this mismatch.

jeffallen:
Fixes #275.


stappersg:
The new link,  https://openocd.org/pages/getting-openocd.html , is indeed valid.

PR #312, opened 2021-12-08
Added missing led for nucleo 743/753 boards.

jmingov:
Both boards have a yellow user LED connected to PE1.
PONG sample task will also make use of the yellow led :)

### DIFF
--- a/drv/user-leds/src/main.rs
+++ b/drv/user-leds/src/main.rs
@@ -41,19 +41,31 @@ enum Op {
 }
 
 cfg_if::cfg_if! {
-    if #[cfg(not(any(target_board = "gemini-bu-1", target_board = "gimletlet-2")))] {
-        #[derive(FromPrimitive)]
-        enum Led {
-            Zero = 0,
-            One = 1,
-        }
-    } else {
+    // Target boards with 4 leds
+    if #[cfg(any(target_board = "gemini-bu-1", target_board = "gimletlet-2"))] {
         #[derive(FromPrimitive)]
         enum Led {
             Zero = 0,
             One = 1,
             Two = 2,
             Three = 3,
+        }
+    }
+    // Target boards with 3 leds
+    else if #[cfg(any(target_board = "nucleo-h753zi", target_board = "nucleo-h743zi2"))] {
+        #[derive(FromPrimitive)]
+        enum Led {
+            Zero = 0,
+            One = 1,
+            Two = 2,
+        }
+    }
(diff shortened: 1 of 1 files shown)


drv/user-leds/src/main.rs
```diff
@@ -308,7 +321,7 @@ fn led_info(led: Led) -> (drv_stm32h7_gpio_api::PinSet, bool) {
     match led {
         Led::Zero => LEDS[0],
         Led::One => LEDS[1],
-        #[cfg(any(target_board = "gemini-bu-1", target_board = "gimletlet-2"))]
+        #[cfg(any(target_board = "gemini-bu-1", target_board = "gimletlet-2", target_board = "nucleo-h753zi", target_board = "nucleo-h743zi2"))]
```
kc8apf, line 324: nit: sort target_boards lexicographically.

kc8apf (approved):

jmingov:
Not sure why it failed, any hint?

kc8apf:
This had a few of us scratching our heads today.  Normal builds are done via `cargo xtask dist` which builds a full image.  Since that can be a bit slow, `cargo xtask check` will build a single package using a "standalone" config so you can iterate quickly when working on a single task.  The standalone config sets the standalone feature on the package being built but also explicitly sets the target board to nucleo-stm32h743zi2.  For user_leds, this confusingly sets an impossible situation of target_board=nucleo-stm32h743zi2 and feature=stm32f4.  This is ultimately a problem with how `cargo xtask check` is designed and it will need to drastically changed.  In the meantime, Add a commit to this PR that changes the standalone feature in drv-user-led to stm32h7 which will result in a valid feature/target_board combination during `cargo xtask check`.

jmingov:
Updated as requested @kc8apf.

----------------------------------
Hey OC Team,

Awesome work, congrats to you all, and really, thanks to the ones involved for checking the commit, I wasn't really expecting any fast response for a trivial led. :) 

Ill think about the "this is ultimately a problem with how `cargo xtask check` is designed and it will need to drastically change." issue, I may be even able to propose some good ideas - aside from adding missing leds ^^. 

Best,
Jaime

kc8apf:
Thanks for sending in a PR!  As for `cargo xtask check`, discussion internally is that we mostly wanted a way to build a subset of tasks as they would be built by `cargo xtask dist` but without building a full image.  https://github.com/oxidecomputer/hubris/pull/319 does that in an expedient way.  I have some ideas for a more elegant approach that I'll file as separate issues.

PR #313, opened 2021-12-08
One improvement to CI

steveklabnik:
* Rename build -> dist: this mirrors the command run
* Rename deploy -> documentation: this describes what it does in a less
  ambiguous way
* use the built-in submodule support of the checkout action rather than
  doing it manually

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -1,4 +1,4 @@
-name: build
+name: dist
 on: [push, pull_request]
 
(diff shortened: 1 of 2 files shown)


kc8apf (approved):

steveklabnik:
This now only does the rename, as https://github.com/oxidecomputer/hubris/pull/316 killed the submodules.

kc8apf (approved):

Issue #240, opened 2021-11-01
Cleaning everything on app.toml changes shouldn't really be necessary


cbiffle:
Currently, `xtask dist` runs around cleaning things whenever `app.toml` changes. This was originally done so that switching _between_ apps didn't accidentally link in each others' objects. It's kind of annoying, however, and shouldn't be strictly necessary ... if we've got all the various build dependencies right. It may be masking cases where we forgot to (say) emit a `cargo:rebuild-if-whatever` line from the build script.

It would be great for one of us to investigate and make this more precise, as it should dramatically reduce build times for people working on more than one app.

It's also going to become increasingly important as we move more config information into the toml, since right now, changing (say) the proposed i2c config tree will force a rebuild of _all tasks,_ which is overkill.

jamesmunns:
Marking this as still relevant, as we are still doing this. We have recently made some improvements to ensure that we are emitting the correct directives (rerun-if-changed in build-rs, storing the manifest env to a const in proc macros), it might be worth doing a pass to make sure this is complete, and consider trusting in cargo's dirty detection directly.

This could also obsolete issues like https://github.com/oxidecomputer/hubris/issues/831, as we would no longer need the `--dirty` flag.

PR #321, opened 2021-12-10
Reduce unnecessary package overbuilding

cbiffle:
Turns out we were building packages more often than we needed to, mostly because their feature sets diverged. I've gone through and fixed most of it. This has the _potential_ to knock a few minutes off a serial build of all app images by increasing build product sharing. In _practice_ we don't realize the full benefit yet, because of #240.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1205,6 +1205,10 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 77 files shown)


steveklabnik (approved):
Looks good to me, modulo that last failure of course.

PR #335, opened 2021-12-17
Bump zerocopy version to 0.6.1

cbiffle:
We were on 0.3.0, which was fine, but the new one is much better.

...and is kind of necessary for the Idol work.


steveklabnik (approved):
Will merge once tests pass.

PR #336, opened 2021-12-17
Beginnings of IDL conversion using Idol

cbiffle:
This converts most of the IPC interactions in the Gimlet image to use Idol.

Skipped for now:
- stm32h7-gpio: by accident
- i2c: on purpose to land this faster.

### DIFF
--- a/idl/user-leds.idol
+++ b/idl/user-leds.idol
@@ -1,0 +1,37 @@
+// User LED blinky API
+
+Interface(
+    name: "UserLeds",
+    ops: {
+        "led_on": (
+            args: {
+                "index": "usize",
+            },
+            reply: Result(
+                ok: "()",
+                err: CLike("LedError"),
+            ),
+            idempotent: true,
+        ),
+        "led_off": (
+            args: {
+                "index": "usize",
+            },
+            reply: Result(
+                ok: "()",
+                err: CLike("LedError"),
+            ),
+            idempotent: true,
+        ),
+        "led_toggle": (
+            args: {
+                "index": "usize",
+            },
+            reply: Result(
+                ok: "()",
+                err: CLike("LedError"),
+            ),
+            idempotent: true,
+        ),
+    },
+)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -891,6 +891,8 @@ dependencies = [
  "cfg-if 0.1.10",
  "drv-lpc55-gpio-api",
  "drv-stm32h7-gpio-api",
+ "idol",
+ "idol-runtime",
  "lpc55-pac",
  "num-traits",
  "stm32f3",
@@ -903,6 +905,8 @@ dependencies = [
 name = "drv-user-leds-api"
 version = "0.1.0"
 dependencies = [
+ "abi",
+ "idol",
  "num-traits",
  "userlib",
  "zerocopy",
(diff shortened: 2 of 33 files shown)


idl/user-leds.idol
```diff
@@ -0,0 +1,37 @@
...
+        ),
+        "led_toggle": (
+            args: {
+                "index": "usize",
+            },
+            reply: Result(
+                ok: "()",
+                err: CLike("LedError"),
+            ),
+            idempotent: true,
```
daym, line 34: Is that one idempotent? I think not. It would toggle more or fewer times.

cbiffle: Good catch!

cbiffle: Fixed in latest patchset.

cbiffle:
Amusingly the bug @daym found in the GPIO IDL is actually present in current `master` -- it's way more obvious when written in Idol.

bcantrill (approved):
I think this looks great!  In trying out `humility qspi`, I did hit a bug in idolatry -- and another opportunity for easy improvement:

https://github.com/oxidecomputer/idolatry/issues/1
https://github.com/oxidecomputer/idolatry/issues/2

It would be good to fix those -- but in debugging them, I got the opportunity to go through the idolatry-generated code, and I can't wait to build on this!

bcantrill:
Some additional comments, none of which should prevent the integration:

1. It would be very helpful to be able to correlate task to interface.
2. It would be very helpful to have the `.idol` files in the build archive

It feels like both of these can be done by adding an `interface` field (say) in the `app.toml` that points to a `.idol` file, and then pass that through to the `build.rs` for tasks that have an interface.  Clearly it would be possible to be slightly less explicit/more magical, but it seems that being explicit is most in keeping with the way we have structured the `app.toml`. (If there are no objections, I may implement a prototype of this.)

bcantrill:
> Some additional comments, none of which should prevent the integration:
> 
> 1. It would be very helpful to be able to correlate task to interface.
> 2. It would be very helpful to have the `.idol` files in the build archive
> 
> It feels like both of these can be done by adding an `interface` field (say) in the `app.toml` that points to a `.idol` file, and then pass that through to the `build.rs` for tasks that have an interface. Clearly it would be possible to be slightly less explicit/more magical, but it seems that being explicit is most in keeping with the way we have structured the `app.toml`. (If there are no objections, I may implement a prototype of this.)

Just to tie this off: I ended up implementing this in a way that didn't necessitate any changes to the `app.toml` or to the build archive.

cbiffle:
> > Some additional comments, none of which should prevent the integration:
> > 
> > 1. It would be very helpful to be able to correlate task to interface.
> > 2. It would be very helpful to have the `.idol` files in the build archive
> > 
> > It feels like both of these can be done by adding an `interface` field (say) in the `app.toml` that points to a `.idol` file, and then pass that through to the `build.rs` for tasks that have an interface. Clearly it would be possible to be slightly less explicit/more magical, but it seems that being explicit is most in keeping with the way we have structured the `app.toml`. (If there are no objections, I may implement a prototype of this.)
> 
> Just to tie this off: I ended up implementing this in a way that didn't necessitate any changes to the `app.toml` or to the build archive.

FWIW, I actually agree with both points, and would like to do them both soon.

PR #339, opened 2021-12-20
Remove standalone build

cbiffle:


steveklabnik (approved):
Love this for us 👍

labbott:
Can we get a short explanation in the commit message or the PR of why we're dropping this? bcantrill reminded me the original reason was to run `cargo check` but we should explain why the overhead of standalone isn't actually worth running that check.

cbiffle:
@labbott rewrote the commit message with details.

PR #340, opened 2021-12-21
remove Gimlet-specific functionality from eval boards

bcantrill:


labbott:
Do we have a matrix somewhere of what features are present on which boards? It's not immediately clear if things are broken by having these tasks present or it's just not necessary.

rmustacc:
> Do we have a matrix somewhere of what features are present on which boards? It's not immediately clear if things are broken by having these tasks present or it's just not necessary.

So, at least I got broken by having these tasks and others present. It made for a very confusing time to realize that I couldn't actually use various i2c buses on a gimletlet until after a bunch of confusing debugging. I've found myself needing to use gimletlet's for a lot of our ad-hoc tasks right now. And while yes, I probably should have looked over the app.toml or humility tasks output and realized that we had a bunch of things that were leftovers of development, it makes for a very confusing use of the board.

So, things aren't broken per se, but using the board correctly is not all obvious.

labbott:
> > Do we have a matrix somewhere of what features are present on which boards? It's not immediately clear if things are broken by having these tasks present or it's just not necessary.
> 
> So, at least I got broken by having these tasks and others present. It made for a very confusing time to realize that I couldn't actually use various i2c buses on a gimletlet until after a bunch of confusing debugging. I've found myself needing to use gimletlet's for a lot of our ad-hoc tasks right now. And while yes, I probably should have looked over the app.toml or humility tasks output and realized that we had a bunch of things that were leftovers of development, it makes for a very confusing use of the board.
> 
> So, things aren't broken per se, but using the board correctly is not all obvious.

Thanks for clarifying. I'm in favor of removing things if they are causing trouble like they were for you and also documenting which boards are intended for which purpose.

rmustacc:
I agree completely on having this documented. That'll be a big help going forward.

cbiffle (approved):
I've got a commit very similar to this floating on some dev branches to decouple gimletlet from hostflash, so, I'm into it.

cbiffle:
@bcantrill is this ready to go?

Issue #228, opened 2021-10-27
Task interaction integrity checks


cbiffle:
We have discussed but not implemented several kinds of integrity checks on task interactions. I'm filing this issue to keep track of them and record progress.

The cases I'm aware of are as follows.

# Uphill send

Check that SEND is always used from low to high priority. This is one of our mechanisms for avoiding deadlock. Validating this would be relatively cheap (a compare on the send path).

However, there is at least one piece of software that's currently violating the uphill send rule _deliberately,_ and that's the test suite -- specifically, the way the suite and assistant task send back and forth. We could refactor it to fix this, potentially.

However, note that this winds up being a subset of the _next_ thing:

# SEND MAC

Impose Mandatory Access Control (MAC) on the SEND operation, limiting the set of tasks that a given task can SEND to. The set would be defined by the application depending on its needs. Tasks outside the set would appear not to exist to the sender, for all intents and purposes -- that is, attempting to access one would be a fault, though it might not specifically be a `TaskIndexOutOfRange` fault.

Having this can potentially _subsume_ uphill send validation, by ensuring at build time that each task can only send to higher-priority tasks.

# Closed RECV reverse MAC validation

A task should only be able to enter closed RECV to listen to a task _that could potentially send something._ This means we should probably check the SEND MAC information on closed RECV.

# REPLY reverse MAC validation

If task A replies to task B, but static MAC rules indicate that B could _never have sent a message_ to A, A is probably malfunctioning. We can detect this case if we have MAC rules. I am inclined to try to detect it, in the interest of catching every programming error as early as possible, but because we must trust servers more than clients in general, I feel like it's slightly less important than (say) validating on SEND.

# Post filters

The POST operation does not currently exist, but I'm adding it in #224. This will let any task poke any other task's notifications, simulating interrupts and just generally being a nuisance. (This is why the guidance on interrupt handling is to check the source of truth for the interrupt events instead of simply trusting the notification bits.)

I could imagine a couple of levels of filters here.

1. We could restrict which tasks can access the POST operation in the first place. POST is usually used by specialized servers that simulate interrupts -- the set I'm aware of is the supervisor, the network stack, and the STM32 external interrupt multiplexing driver that @labbott is writing.
[...]

PR #342, opened 2021-12-21
WIP on RAM ECC driver - Open Questions

iangrunert:
The STM32H742, STM32H743/753 and STM32H750 boards have ECC modules on their RAM, however they don't automatically handle correcting detected errors nor resetting the microcontroller if the errors are unrecoverable. I started work on a driver to handle these interrupts, but there's some challenges involved though which I'm not sure how to navigate. I'm opening this WIP PR to start a discussion about how (or if) to proceed.

## Problems

When handling single errors, we need to read from that address and re-write the same value back to that address. The only place we could do this is the kernel as memory protection will prevent us from doing it otherwise.

We need access to the task table when handling double errors to determine how to handle the error based on where it is:

- in a task stack region, we can restart the task
- in a task data region, we can reload the task data from flash and restart the task
- in the kernel data region, we could reload that data from flash but might choose to reboot entirely instead
- in the kernel stack region, the only safe option may be to reboot

Unfortunately there's currently no way to get access to the task region description outside of the kernel to make this determination. There's also no way to force a task's data to be reloaded from flash.

## Solution

Perhaps handling ECC errors should be something offered by the supervisor - it already has a tighter coupling with the kernel, and it has the capability to reboot the system. The missing kernel capabilities could be added to the existing kernel IPC leveraged by the supervisor - specifically:

- Load and rewrite the same value at this memory address
- Ask what is present at a given memory address - kernel, task stack or data (including task number), or unallocated memory
- Reload the task data from flash, perhaps as an extension to restart_task

I'm not sure how best to extend the supervisor to add these board-specific capabilities. It's a small piece of code so perhaps a board-specific supervisor duplicating and extending the capabilities of jefe wouldn't be so bad in practice.

There's also an open question of the value of handling ECC faults at all - the chances of ECC catching data corruption depends on the uptime of the board. The chances of a memory fault due to cosmic rays seems small given the amount of memory we're working with, though that chance increases the longer the expected uptime of the board is.

## References

[Error correction code (ECC) management for internal memories protection on
STM32H7 Series](https://www.st.com/resource/en/application_note/dm00623136-error-correction-code-ecc-management-for-internal-memories-protection-on-stm32h7-series-stmicroelectronics.pdf)
[Reference manual for STM32H742, STM32H743/753 and STM32H750 - Section 3 RAM ECC monitoring, Page 139](https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -752,6 +752,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "drv-stm32h7-ecc"
+version = "0.1.0"
+dependencies = [
+ "num-traits",
+ "stm32h7",
+ "userlib",
+ "zerocopy",
+]
+
+[[package]]
 name = "drv-stm32h7-gpio"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ members = [
     "drv/stm32fx-rcc",
     "drv/stm32fx-usart",
 
+    "drv/stm32h7-ecc",
     "drv/stm32h7-gpio",
     "drv/stm32h7-gpio-api",
     "drv/stm32h7-rcc",
--- a/app/demo-stm32h7-nucleo/app-h753.toml
+++ b/app/demo-stm32h7-nucleo/app-h753.toml
@@ -55,6 +55,15 @@ requires = {flash = 8192, ram = 1024}
 uses = ["rcc"]
 start = true
 
+[tasks.ecc_driver]
+path = "../../drv/stm32h7-ecc"
+name = "drv-stm32h7-ecc"
+features = ["h753"]
+priority = 1
+requires = {flash = 1024, ram = 1024}
+interrupts = { 145 = 1 }
+start = true
+
 [tasks.gpio_driver]
 path = "../../drv/stm32h7-gpio"
 name = "drv-stm32h7-gpio"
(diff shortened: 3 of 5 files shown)


jeffallen:
ECC is critical in safety critical systems. Hubris is not marketed for them, but someone will sooner or later try, so it would be nice if Hubris helped them be safer.

Toyota's throttle control was criticized for not having working ECC (also called EDAC here): https://embeddedgurus.com/barr-code/2013/10/an-update-on-toyota-and-unintended-acceleration/

kc8apf:
I'm generally a fan of ECC as I've seen it be useful all too often.  I know @cbiffle has similar experiences.  As for how single-bit errors should be corrected, you're right that MPU gets in the way of a driver task accessing the affected addresses.  Any mechanism that would allow the task to do a write to an affected task's memory w/o cooperation of the affected task also becomes a primitive for violating memory protection.  Even if the kernel provided IPC mechanisms for doing so, we'll probably want #228 to limit which tasks can use those IPCs.

iangrunert:
Yeah, I agree we'd likely want to limit which tasks can use those IPCs. Part of the appeal of doing it in the supervisor was that it'd be forward compatible with proposed restrictions on kernel IPC sends:

> It is our intent to restrict kernel IPC sends to “privileged” tasks — likely just the supervisor task. We haven’t implemented this yet.

I don't see that proposed restriction referenced in #228 though it could be considered part of SEND MAC.

For single bit error correction - the kernel IPC  shape would be a request to repair a single bit error at a specified memory address, where the kernel would load and restore that memory address without revealing it's contents to the caller. That feels reasonably safe, though I'm not sure if that opens a vector for timing attacks.

PR #348, opened 2021-12-28
HIF support for Idol

bcantrill:
This is the primordial HIF support for Idol, allowing for Idol operations to be called via HIF from Humility.  This support is not complete (it does not yet include any lease support), but it is an adequate start that allows for simple services to be called without any custom Humility support whatsoever.


bcantrill:
This PR is superseded by #364.

PR #352, opened 2021-12-28
Cargo.toml: Remove 'package.metadata.build.target' leftover from 8e0b…

flihp:
…13b8.

A number of these configuration sections were removed along with the
related 'standalone' build stuff. A few appear to have been missed and
AFAIK they're no longer required.


steveklabnik:
Thanks!

PR #360, opened 2022-01-11
Shorten panic messages

mkeeter:
This uses `--remap-path-prefix` to shorten the paths embedded in panic messages.  Since we're limited to 128 bytes, this could make the difference between seeing and not seeing the full path + line number.

In particular, it remaps `git` dependencies from `/Users/mjk/.cargo/git/checkouts/...` to just `git/...` (the original will vary depending on OS and username), and crates within the same checkout from `/Users/mjk/oxide/hubris/...` to `hubris/...`.

As far as I can tell, this doesn't affect Humility's backtracing, since that's all baked into the ELF files.

However, [the docs](https://doc.rust-lang.org/rustc/command-line-arguments.html#--remap-path-prefix-remap-source-names-in-output) say this also affects "debug information, macro expansions, etc" – @steveklabnik , any idea if this is going to break something more subtly?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -926,6 +926,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dunce"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "453440c271cf5577fd2a40e4942540cb7d0d2f85e27c8d07dd0023c925a67541"
+
+[[package]]
 name = "ecdsa"
 version = "0.12.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


bcantrill:
This is great! I have never been totally comfortable with the information leakage in the DWARF files in terms of the path in which something is built -- good to get rid of it, on several fronts!

cbiffle:
I _suspect_ that this will break source-level debugging in GDB. We may be able to pass some more options to GDB from `xtask gdb` to undo the breakage. Otherwise I'm very supportive.

cbiffle:
Separately, I wonder if it's possible to strip the commit hash out of git paths from cargo. We know the commit hash because everything's pinned, it's just wasting bytes. It's not clear that we can do this with `remap-path-prefix`.

mkeeter:
Alright, GDB is now working through the magic of `substitute-path`, and I stripped the `github.com-LONGHASH` prefix, since that's apparently canonical-ish.

steveklabnik (approved):
Confirmed that this works on Windows. Thanks for putting in some extra effort to make sure this works for me.

PR #361, opened 2022-01-12
More informative size info

mkeeter:
Before:
```
flash = 8000000..8100000
ram = 20000000..20020000
Used:
flash: 0x5e100
ram: 0x13500
```

After:
```
flash = 0x08000000..0x08100000
ram   = 0x20000000..0x20020000
Used:
  flash: 0x5e100 (36%)
  ram:   0x13500 (60%)
```

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -90,7 +90,7 @@ pub fn package(
         }
     }
     for (name, range) in &memories {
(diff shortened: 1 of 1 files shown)


cbiffle (approved):
That thing with having to use `format!` to get a colon account for in the padding/justification is a thing that has bothered me for a while -- but I don't know a better way.

cbiffle (approved):

PR #364, opened 2022-01-15
Draft of Gimlet sequencing branch

bcantrill:


nathanaelhuffman (approved):
Looked through the json build.rs stuff and it looked good, I see registers and masks getting set up :)

jeffallen:
I tried this with my h753 and I got no compile errors and the lights still flash.

cbiffle:
I'll read over the diff this morning. Testing on the Nucleo, Gimletlet, and Gemini seems low priority since this shouldn't be able to affect them much (except through removal of functionality) -- testing on Gimlet is more important.

app/gimlet/app.toml
```diff
@@ -145,10 +155,10 @@ path = "../../drv/gimlet-seq-server"
 name = "drv-gimlet-seq-server"
 features = ["h753"]
 priority = 3
-requires = {flash = 32768, ram = 1024 }
+requires = {flash = 65536, ram = 8192 }
```
cbiffle, line 158: So, rodata in this change is twice as large as text. Are you aware that you appear to have pulled in unicode grapheme tables and a bunch of supporting string stuff? What sort of string manipulation is this change doing?

app/gimlet/app.toml
```diff
@@ -162,6 +172,15 @@ uses = ["quadspi"]
 interrupts = {92 = 1}
 task-slots = ["gpio_driver", "rcc_driver"]
 
+[tasks.sensor]
+path = "../../task/sensor"
+name = "task-sensor"
+features = ["itm"]
+priority = 3
+requires = {flash = 16384, ram = 2048 }
+stacksize = 2000
```
cbiffle, line 181: This seems wrong -- that leaves 48 bytes of RAM for all sensor data?

build/i2c/src/lib.rs
```diff
@@ -132,8 +169,36 @@ pub enum Disposition {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+enum Sensor {
+    Temperature,
+    Power,
+    Current,
+    Voltage,
+    Speed,
+}
+
+impl std::fmt::Display for Sensor {
```
cbiffle, line 188: Two alternatives to this as an FYI:

- I'd probably implement this as `write!(f, "{:?}", self)` - if the all-caps screamy printing is important, change the enum variant names
- If you want the output in a different case from the variants, consider calling `write_str` in each match arm. I'd put good money on it generating cheaper code. e.g. `Sensor::Temperature => f.write_str("TEMPERATURE"),`

build/i2c/src/lib.rs
```diff
@@ -261,6 +326,7 @@ impl ConfigGenerator {
         let mut s = &mut self.output;
 
         assert!(self.disposition != Disposition::Devices);
+        assert!(self.disposition != Disposition::Sensors);
```
cbiffle, line 329: So, under what circumstances should future Disposition values get added here? Should this be a match to ensure that people don't forget to extend it?

build/i2c/src/lib.rs
```diff
@@ -719,6 +780,163 @@ impl ConfigGenerator {
         Ok(())
     }
 
+    fn emit_sensor(
+        &mut self,
+        device: &str,
+        label: &str,
+        ids: &Vec<usize>,
```
cbiffle, line 787: To anticipate clippy: this should be `&[usize]` for maximum generality, callers won't need to change because Vec derefs to a slice.

build/task-link.x
```diff
@@ -94,6 +94,12 @@ SECTIONS
     KEEP(*(.task_slot_table));
   }
 
+  /* ## .idolatry */
+  .idolatry (INFO) : {
+    . = .;
```
cbiffle, line 99: ...what's this line for?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -388,6 +395,34 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         poll_for_write_complete(&self.qspi);
         Ok(())
     }
+
+    fn get_mux(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<HfMuxState, RequestError<HfError>> {
```
cbiffle, line 402: Hmmm. This should probably not return `Result`. Idol support for that might be missing though -- did you try it?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -388,6 +395,34 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
+    fn set_mux(
+        &mut self,
+        _: &RecvMessage,
+        state: HfMuxState,
+    ) -> Result<(), RequestError<HfError>> {
+        let gpio_driver = gpio_api::Gpio::from(GPIO.get_task_id());
+
+        let rv = match state {
+            HfMuxState::SP => gpio_driver.reset(self.select_pin),
+            HfMuxState::HostCPU => gpio_driver.set(self.select_pin),
```
cbiffle, line 415: I need to change the gpio operations to not return Result, because they can only return client errors -- it hasn't been ported to Idol yet. But, consider just unwrapping here. If it fails, it's because `select_pin` is invalid for the _chip,_ which isn't something the caller can do anything about.

drv/gimlet-seq-api/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#![no_std]
+
+use userlib::*;
+use zerocopy::AsBytes;
+
+#[derive(Copy, Clone, Debug, FromPrimitive, PartialEq)]
+pub enum SeqError {
+    IllegalTransition = 1,
+    MuxToHostCPUFailed = 2,
```
cbiffle, line 15: (I don't think this error code or the next one can actually happen in practice, I'd recommend removing them unless I missed something.)

drv/gimlet-seq-api/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+    MuxToSPFailed = 3,
+}
+
+impl From<SeqError> for u16 {
+    fn from(rc: SeqError) -> Self {
+        rc as u16
+    }
+}
+
+impl From<SeqError> for u32 {
```
cbiffle, line 25: Is this used anywhere?

drv/gimlet-seq-server/build.rs
```diff
@@ -12,9 +14,148 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
+    let disposition = build_i2c::Disposition::Devices;
+
+    if let Err(e) = build_i2c::codegen(disposition) {
+        println!("code generation failed: {}", e);
+        std::process::exit(1);
+    }
+
+    fs::write(out.join("gimlet_regs.rs"), regs()?)?;
+    println!("cargo:rerun-if-changed=fpga.bin");
```
cbiffle, line 26: Can you move this up to where the fpga.bin is actually being manipulated, to make it easier to find / more obvious?

drv/gimlet-seq-server/build.rs
```diff
@@ -12,9 +14,148 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            ..
+        } = child
+        {
+            writeln!(&mut output, "    {} = 0x{:x},", inst_name, addr_offset)?;
+        } else {
+            panic!("unexpected child {:?}", child);
+        }
+    }
+
+    println!("cargo:rerun-if-changed=gimlet_regs.json");
```
cbiffle, line 88: This is not necessary, `include_str!` does it as a side effect.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -180,11 +217,10 @@ fn main() -> ! {
...
     // into it should be willing to talk to us over SPI, and should be able to
     // serve up a recognizable ident code.
-    //
-    // TODO except for now we're going to skip the version check and
-    // unconditionally reprogram it because the SPI communication code ain't
-    // written, and also yolo. Replace this with a check.
-    let reprogram = true;
+    let seq = seq_spi::SequencerFpga::new(spi.device(SEQ_SPI_DEVICE));
+
+    let reprogram = if seq.valid_ident() { false } else { true };
```
cbiffle, line 222: ... should this be `!seq.valid_ident()`?

drv/stm32h7-i2c/src/lib.rs
```diff
@@ -220,24 +220,49 @@ impl<'a> I2cController<'a> {
...
                     .sdadel().bits(0)
                 });
+
+                #[cfg(feature = "amd_erratum_1394")]
+                compile_error!("no support for amd_erratum_1394 on h7b3");
             } else if #[cfg(any(feature = "h743", feature = "h753"))] {
+                cfg_if::cfg_if! {
+                    // Due to AMD Milan erratum 1394, the processor needs an
+                    // abnormally long data setup time from an I2C target
+                    // sending an ACK. (According to the erratum,
```
cbiffle, line 230: Gross, AMD.

idl/gimlet-hf.idol
```diff
@@ -57,5 +57,28 @@ Interface(
                 err: CLike("HfError"),
             ),
         ),
+	"get_mux": (
```
cbiffle, line 60: Why are these de-dented like this?

task/sensor-api/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+
+impl From<SensorId> for usize {
+    fn from(id: SensorId) -> Self {
+        id.0
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum Reading {
+    None,
```
cbiffle, line 30: Consider eliminating `None` and putting this in an option where it might be missing, so that people who know it _isn't_ missing don't have to cope with a non-removable `None` variant.

task/sensor/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn nodata(
+        &mut self,
+        _: &RecvMessage,
+        id: SensorId,
+        nodata: NoData,
+    ) -> Result<(), RequestError<SensorError>> {
+        let index = id.0;
+
+        if index < NUM_SENSORS {
+            self.data[index] = Reading::NoData(nodata);
```
cbiffle, line 75: We should talk about this at a whiteboard (possibly after this gets merged), but -- it's not clear to me that you want to throw away any existing valid data if the sensor flakes.

task/thermal/src/main.rs
```diff
@@ -12,111 +12,239 @@
...
+            if pwm <= 100 {
+                match self.fctrl.set_pwm(fan, PWMDuty(pwm)) {
+                    Ok(_) => Ok(()),
+                    Err(_) => Err(ThermalError::DeviceError.into()),
+                }
+            } else {
+                Err(ThermalError::InvalidPWM.into())
+            }
+        } else {
+            Err(ThermalError::InvalidFan.into())
```
cbiffle, line 184: FWIW, this and the error on line 181 look like client programming errors, which is a great use for `RequestError::BadMessage` (which will panic the client).

cbiffle (approved):
This is probably okay to merge as-is. My comments (which GitHub won't let me actually _see_ while writing in this box because what's UX) fall into three rough categories.

- FYIs, which you can address if they excite you.
- A couple of "considers" where I think the change is worth making but can be talked out of it.
- A request for a higher-level discussion about the sensor task data model, which doesn't need to block the merge.

app/gimlet/app.toml
cbiffle: So, the increase in rodata seems to fall into a few categories.

- There are tables for a lot of enums giving the Debug representation, which seem to be almost entirely `unwrap()` on IPC returns
- There's a large binary data table I can't really explain.
- Panic messages.

build/task-link.x
bcantrill: This is to assure that the `.idolatry` section isn't chucked.

task/sensor-api/src/lib.rs
bcantrill: Yes, good call.

idl/gimlet-hf.idol
bcantrill: Ha ha -- it's an errant hard tab because my editor didn't recognize `.idl` when I first started editing them.  Fixed; thank you!

task/thermal/src/main.rs
bcantrill: In this case, the client is often `hiffy`, which I would rather not detonate. ;)

task/sensor/src/main.rs
bcantrill: I think this first sensor task is very much a sketch -- I am assuming that we will want to improve it in many ways, including how data is retained/discarded.

drv/stm32h7-i2c/src/lib.rs
bcantrill: Indeed!

drv/gimlet-seq-server/build.rs
bcantrill: TIL!

drv/gimlet-seq-server/src/main.rs
bcantrill: Yes -- fixed

drv/gimlet-seq-server/build.rs
bcantrill: Done

drv/gimlet-seq-api/src/lib.rs
bcantrill: No --- gone

drv/gimlet-seq-api/src/lib.rs
bcantrill: The `IllegalTransition` can happen due to an illegal call from `hiffy`. `MuxToHostCPUFailed` *shouldn't* happen, but the task that is propagating this error (namely, the sequencer) doesn't know that:  it is propagating this on a failure from the host flash server.

drv/gimlet-hf-server/src/main.rs
bcantrill: My disposition has always been to return a `Result` just to avoid changing an interface when a failure mode is discovered, but perhaps that's not warranted here.

build/i2c/src/lib.rs
bcantrill: Fixed.

build/i2c/src/lib.rs
bcantrill: Yes, good call.

app/gimlet/app.toml
bcantrill: That really *does* seem wrong; investigating.

bcantrill: And of course, this is because the sensor data is currently stored on the stack. Adding a comment!

PR #371, opened 2022-01-18
Netstack

cbiffle:


app/gimletlet/src/main.rs
```diff
@@ -77,6 +77,17 @@ unsafe fn system_pre_init() {
...
+    let rcc = &*device::RCC::ptr();
+    rcc.ahb2enr.modify(|_, w| {
+        w.sram1en()
+            .set_bit()
+            .sram2en()
+            .set_bit()
+            .sram3en()
+            .set_bit()
+    });
+
```
mkeeter, line 90: Want to add these changes to Sidecar's `main.rs` as well?

drv/stm32h7-eth/src/lib.rs
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    fn is_smi_busy(&self) -> bool {
+        self.mac.macmdioar.read().mb().bit()
+    }
+}
+
+/// Standard MDIO registers laid out in IEEE 802.3 standard clause 22. Vendors
+/// often add to this set in the 16+ range.
+pub enum SmiClause22Register {
```
mkeeter, line 373: `vsc7448-pac` autogenerates a PHY register map based on Microchip's SDK, e.g. [`phy.rs` here](https://github.com/oxidecomputer/vsc7448/blob/master/vsc7448-pac/src/phy.rs) and the standard page [here](https://github.com/oxidecomputer/vsc7448/blob/master/vsc7448-pac/src/phy/standard.rs).

That's what I'm using for PHY bringup in `task-vsc7448`, since I need to use a bunch of registers from the extended page – does it make sense to also use it here?

drv/stm32h7-eth/src/ring.rs
```diff
@@ -0,0 +1,429 @@
...
+/// carefully control accesses to its contents.
+#[repr(transparent)]
+pub struct TxDesc {
+    tdes: [AtomicU32; 4],
+}
+
+impl TxDesc {
+    pub const fn new() -> Self {
+        Self {
+            tdes: [
```
mkeeter, line 51: You could save a line or two with
```
    const ZERO: AtomicU32 = AtomicU32::new(0);
    Self {
        tdes: [ZERO; 4],
    }
```
but I'm not sure that's worth the cleverness.

drv/stm32h7-eth/src/ring.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// packet into it.
+    ///
+    /// If the next buffer in the ring is not holding a pending packet, this
+    /// will borrow it and call `body` with its address. `body` returns the
+    /// number of bytes in the packet; this routine will then set up a pending
+    /// transmit descriptor for that prefix of the buffer, and return
+    /// `Some(len)`.
+    ///
+    /// If the next buffer in the ring is pending, that means we have run out of
+    /// TX ring slots. In that case, this will return `None` without invoking
```
mkeeter, line 157: The docstring seems out of date, since this returns a `Result`, not an `Option`.

drv/stm32h7-eth/src/ring.rs
```diff
@@ -0,0 +1,429 @@
...
+    ///
+    /// # Panics
+    ///
+    /// If a buffer is available, and we call `body`, and `body` returns a valid
+    /// length larger than `BUFSZ`. Because that's obviously wrong.
+    pub fn try_with_next<R, F: FnOnce(&mut [u8]) -> R>(
+        &self,
+        len: usize,
+        body: F,
+    ) -> Result<R, F> {
```
mkeeter, line 168: Speaking of which, it seems weird to return the closure in the `Err` condition.  Is there a subtle reason for it?

drv/stm32h7-eth/src/ring.rs
```diff
@@ -0,0 +1,429 @@
...
+            let buffer = unsafe { &mut *buffer };
+            let buffer = &mut buffer[..len];
+
+            let result = body(buffer);
+
+            // Program the descriptor to represent the packet. We program
+            // carefully to ensure that the memory accesses happen in the right
+            // order: the entire descriptor must be written before the OWN bit
+            // is set in TDES3 using a RELEASE store.
+            d.tdes[0].store(buffer.as_ptr() as u32, Ordering::Relaxed);
```
mkeeter, line 196: Could we make the `struct TxDesc` have four members with meaningful names, instead of this array?

cbiffle: They're named `TDES0` through `TDES3` in the reference manual. I had fields named that at one point and then switched to the array. There aren't obviously better names available imo.

drv/stm32h7-eth/src/ring.rs
cbiffle: There definitely was at one point (about resource management and timing when exactly buffers get dropped), but that may no longer be the case...the current structure of the code isn't using this for anything. I will try simplifying it and see if the reason becomes clearer to me.

drv/stm32h7-eth/src/ring.rs
cbiffle: Fixing!

drv/stm32h7-eth/src/ring.rs
cbiffle: Being able to use it in two places is kind of nice (rxdesc has similar code). I'll play with it.

drv/stm32h7-eth/src/lib.rs
cbiffle: Maaaaaybe. That's a lot of code to bring in for boards that don't have a vsc7448 on them.

mkeeter: After chatting at the Hubris sync, I'm going to make a separate `drv/vsc85xx` crate and use it in both the netstack and the `vsc7448` code.  That can wait until after this branch is merged, though.

app/gimletlet/src/main.rs
cbiffle: Really, portions of this commit should be added to every stm32h7x3 main.rs, which serves as an excellent reminder that we need to centralize this stuff.

drv/stm32h7-eth/src/ring.rs
cbiffle: Wound up adopting this, thanks!

app/gimletlet/src/main.rs
cbiffle: I'd like to take centralizing this as a TODO after this merges, if it's okay with you.

mkeeter: Sure, that makes sense – want to copypasta it into Sidecar in this PR, or should I do that in `task-vsc7448` after this is merged and I rebase?

cbiffle: Go ahead and do the latter -- I've started the centralization work on a branch.

idl/net.idol
```diff
@@ -0,0 +1,36 @@
...
+            leases: {
+                "payload": (type: "[u8]", read: true),
+            },
+            reply: Result(
+                ok: "()",
+                err: CLike("NetError"),
+            ),
+        ),
+    },
+)
```
mkeeter, line 36: I'd appreciate an SMI read/write API in here for Sidecar VSC8504 configuration, until rev B moves it off that bus.

cbiffle: Ah, yes. I will add that. If this PR gets approved before I finish, I'll do a followup.

drv/stm32h7-eth/src/ring.rs
cbiffle: @mkeeter I have switched these back to Option.

PR #372, opened 2022-01-18
kern/arm_m: properly dispatch to irq0 handler

kc8apf:
Off-by-one error in DefaultHandler() meant IRQ0 would result in an
unknown exception panic instead of looking for a task to dispatch to.

Fixes #367


labbott (approved):
LGTM

PR #377, opened 2022-01-19
Move SPI device definitions into config data

cbiffle:
Previously, on stm32h7, we had a big conditional compilation block in the SPI server source code that defined its behavior on each target board. That's lame. This moves it into app.toml.


bcantrill (approved):
Looks great! Also: this is very clean code for code that generates code; definitely some good patterns here to use elsewhere!

PR #381, opened 2022-01-19
Build sidecar in CI

steveklabnik:

PR #382, opened 2022-01-20
Fix sidecar mux options.

cbiffle:
Toml is evil.


steveklabnik:
cherry-picked into #381

PR #383, opened 2022-01-20
[RFC] RNG driver API

flihp:
This is a request for comments on some work to improve the interface to the RNGs. Currently it's limited to the lpc55 dev board but the goal is for this to be a common API crate for our RNG drivers. This was needed for compatibility with the 'getrandom' crate that's a dependency of the 'ring' crate currently required by the SPDM crate @andrewjstone has been developing. General support for ring in a hubris task is likely not possible (https://github.com/briansmith/ring/issues/813) but supporting 'getrandom' may still have its benefits. TIA for any and all feedback.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -671,6 +671,7 @@ name = "drv-lpc55-rng"
 version = "0.1.0"
 dependencies = [
  "drv-lpc55-syscon-api",
+ "drv-rng-api",
  "lpc55-pac",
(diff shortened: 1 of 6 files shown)


flihp:
closing after seeing #136

labbott:
heh that needs some work and kind of got pushed aside. I'm happy to have this take over from that one.

PR #384, opened 2022-01-20
Fix SPI device ordering and potentially break lots of things

cbiffle:


bcantrill (approved):
Looks good -- and confirmed that this works on Gimlet.

PR #386, opened 2022-01-20
Add per-device SPI clock divisors.

cbiffle:
This also changes the clock rate on stm32h7 nucleo to demonstrate that it works.

### DIFF
--- a/app/demo-stm32h7-nucleo/app-h743.toml
+++ b/app/demo-stm32h7-nucleo/app-h743.toml
@@ -292,6 +292,7 @@ input = {port = "A", pin = 6, af = 5} # miso
 [config.spi.spi1.devices.pins]
 mux = "cn7_arduino"
 cs = {port = "D", pin = 14}
+clock_divider = "DIV32"
 
 
 [config.net]
--- a/app/demo-stm32h7-nucleo/app-h743.toml
+++ b/app/demo-stm32h7-nucleo/app-h743.toml
@@ -115,15 +115,15 @@ path = "../../drv/stm32h7-spi-server"
 name = "drv-stm32h7-spi-server"
 priority = 2
 requires = {flash = 16384, ram = 2048}
-features = ["spi3", "h743"]
-uses = ["spi3"]
+features = ["spi1", "h743"]
+uses = ["spi1"]
 start = true
-interrupts = {51 = 1}
+interrupts = {35 = 1}
 stacksize = 1000
 task-slots = ["gpio_driver", "rcc_driver"]
 
 [tasks.spi_driver.config.spi]
-global_config = "spi3"
+global_config = "spi1"
 
 [tasks.net]
 path = "../../task/net"
@@ -279,19 +279,19 @@ af = 4
 # driver = "ltc4306"
 # address = 0b1001_010
 
-[config.spi.spi3]
-controller = 3
+[config.spi.spi1]
+controller = 1
 
-[config.spi.spi3.mux_options.port_b]
+[config.spi.spi1.mux_options.cn7_arduino]
 outputs = [
-    {port = "B", pins = [3], af = 6},
-    {port = "B", pins = [5], af = 7},
+    {port = "A", pins = [5], af = 5}, # sck
+    {port = "B", pins = [5], af = 5}, # mosi
 ]
-input = {port = "B", pin = 4, af = 6}
+input = {port = "A", pin = 6, af = 5} # miso
 
-[config.spi.spi3.devices.pins]
-mux = "port_b"
-cs = {port = "A", pin = 4}
+[config.spi.spi1.devices.pins]
+mux = "cn7_arduino"
+cs = {port = "D", pin = 14}
 
 
 [config.net]
--- a/drv/stm32h7-spi-server/build.rs
+++ b/drv/stm32h7-spi-server/build.rs
@@ -109,7 +109,29 @@ struct Af(usize);
 #[derive(Clone, Debug, Deserialize)]
 struct DeviceDescriptorConfig {
     mux: String,
+    #[serde(default)]
+    clock_divider: ClockDivider,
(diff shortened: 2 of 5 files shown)


drv/stm32h7-spi-server/build.rs
```diff
@@ -160,27 +182,31 @@ impl ToTokens for SpiConfig {
...
+            syn::parse_str(&format!("Spi{}", self.controller)).unwrap();
+
         // We don't derive ToTokens for DeviceDescriptorConfig because it needs
         // extra knowledge (the mux_indices map) to do the conversion. Instead,
         // convert it here:
         let device_code = self.devices.values().map(|dev| {
             let mux_index = mux_indices[&dev.mux];
             let cs = &dev.cs;
+            let div: syn::Ident =
+                syn::parse_str(&format!("{:?}", dev.clock_divider)).unwrap();
```
kc8apf, line 199: Not a fan on using Debug formatting in a non-debug usage like this.  I'd rather see Display implemented and used.

drv/stm32h7-spi-server/build.rs
```diff
@@ -160,27 +182,31 @@ impl ToTokens for SpiConfig {
...
             quote::quote! {
                 DeviceDescriptor {
                     mux_index: #mux_index,
                     cs: #cs,
+                    // `spi1` here is _not_ a typo/oversight.
```
kc8apf, line 204: why?

cbiffle: Because the PAC crate has decided to call all SPI types spi1. There are comments to this effect at several places in the driver, I can add another one if you like.

drv/stm32h7-spi-server/build.rs
cbiffle: I'd argue that Debug is a good choice here because we're trying to convert the enum back to its code representation, whereas Display is normally a human-readable output.

drv/stm32h7-spi-server/build.rs
kc8apf: I saw other mentions of spi1 being correct but not why.

drv/stm32h7-spi-server/build.rs
kc8apf: https://doc.rust-lang.org/std/fmt/trait.Debug.html says `Debug should format the output in a programmer-facing, debugging context.` while https://doc.rust-lang.org/std/fmt/trait.Display.html says `Display is similar to Debug, but Display is for user-facing output, and so cannot be derived.` so neither is really correct for this situation.  strum's `AsRefStr` is probably the closest to "correct" and avoids the string allocation.

drv/stm32h7-spi-server/build.rs
cbiffle: Would you like me to add a comment?

drv/stm32h7-spi-server/build.rs
cbiffle: Given that the type I'm formatting and its sole use are both within this same file, I don't see how this code is a problem. I'm basically trying to give a bit more feedback than just using String, but just using String would also work here.

cbiffle: What if I add a large comment on the divider type noting that the names in the enum must match the names used in the PAC crate?

kc8apf: I understand you are trying to get the name of an enum variant as a &str.  I found `&format!("{:?}", variant)` difficult to comprehend as doing that.  More generally, `Debug` is intended to be programmer-facing for debug contexts so using it in code generation feels wrong.

drv/stm32h7-spi-server/build.rs
kc8apf: Yes

app/demo-stm32h7-nucleo/app-h743.toml
```diff
@@ -279,19 +279,20 @@ af = 4
 # driver = "ltc4306"
 # address = 0b1001_010
 
-[config.spi.spi3]
-controller = 3
+[config.spi.spi1]
```
bcantrill, line 282: Worth adding a cautionary note here about SPI3?  (It may not be...)  FWIW, here is the comment that I had when carefully avoiding TRACESWO:

```rust
                            //
                            // It may seem reasonable -- especially given the
                            // groupings and silkscreen on CN7 -- to use PB3
                            // for SPI3_SCK. However, this pin is TRACESWO on
                            // AF0:  if we configure this pin for SPI, ITM
                            // will no longer work (and with a failure mode
                            // that can be difficult to debug).  We instead
                            // opt to use PC10 for SPI3_SCK, which is also
                            // easily accessible:  it's pin 6 on CN8.
                            //
```

Again, not sure if this is worth it -- it's just so painful to get wrong.

bcantrill (approved):
Looks great -- will be good to get ITM working again again on the h743. ;)

bcantrill (changes requested):
I realize that we need to make these same SPI changes to the H753 Nucleo `app.toml` as well.

app/demo-stm32h7-nucleo/app-h743.toml
cbiffle: I don't think the note is necessary, because anyone following the SPI labels on the board won't consider mapping it to PB3.

bcantrill (approved):

app/demo-stm32h7-nucleo/app-h743.toml
```diff
@@ -279,19 +279,20 @@ af = 4
...
+controller = 1
 
-[config.spi.spi3.mux_options.port_b]
+[config.spi.spi1.mux_options.cn7_arduino]
```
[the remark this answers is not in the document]
cbiffle: That portion of the Nucleo board is the Arduino-compatible subset.

PR #388, opened 2022-01-21
Gimlet clock programming + CS workaround

bcantrill:
This adds programming for the 8A34003 on Gimlet, as well as an explicit
panic for those unlucky Gimlets with broken CS pins.


drv/gimlet-seq-server/src/payload.rs
```diff
@@ -0,0 +1,3812 @@
...
+/// generator.  This code was generated by "humility rencm -g" given
+/// Aardvark output generated by running Renesas configuration software.
+///
+#[rustfmt::skip]
+pub fn idt8a3xxxx_payload<E>(
+    mut func: impl FnMut(&[u8]) -> Result<(), E>
+) -> Result<(), E> {
+
+    // PAGE_ADDR = 0xc000
+    func(&[
```
cbiffle: I suggested this before, so maybe you've already tried it and I missed the response, but:

I suspect this code would be significantly smaller if you restructured this as an array of things to send, and then called `func` in a loop. That is,

```rust
const PAYLOAD: &[&[u8]] = &[ &[0xfc ..], ..];
for chunk in PAYLOAD {
    func(chunk); // lol it rhymes
}
```

Would you be willing to try this and check the text size difference?

cbiffle (approved):
Some questions below; if the answers turn out to be "meh" then I'm cool with this merging.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -241,11 +253,22 @@ fn main() -> ! {
...
-                    // We're gonna try again.
+
+                    laps += 1;
+
+                    if laps >= 3 {
+                        panic!(
+                            "could not reprogram FPGA after {} \
+                            attempts; if CS has been reworked, \
+                            look for \
+                            \"ATTENTION REWORKED CS\" in app.toml",
```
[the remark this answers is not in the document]
bcantrill: It does, yes. It's been reworked to live within the limit. ;)

drv/gimlet-seq-server/src/payload.rs
bcantrill: Yes, will do!

bcantrill: Good intuition! This is a significant savings. Here is the code before:

```
section              size        addr
.text               27420   134348800
.rodata             33188   134376220
.addr_table             0   134409408
.data                1292   536938496
.bss                    0   536939788
```

And afterwards:

```
section              size        addr
.text               15292   134348800
.rodata             37300   134364092
.addr_table             0   134401392
.data                1292   536938496
.bss                    0   536939788
```

So more `.rodata` -- but much, much less text, and significant net savings:

```
$ size ./gimlet_seq ./gimlet_seq-after 
   text	   data	    bss	    dec	    hex	filename
  60608	   1292	      0	  61900	   f1cc	./gimlet_seq
  52592	   1292	      0	  53884	   d27c	./gimlet_seq-after
```

Clearly a very good change; thank you for the suggestion!

app/gimlet/app.toml
```diff
@@ -163,7 +163,7 @@ name = "drv-gimlet-seq-server"
 features = ["h753"]
 priority = 3
 requires = {flash = 65536, ram = 8192 }
-stacksize = 1024
+stacksize = 2048
```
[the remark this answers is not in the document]
bcantrill: This has been obviated by @mkeeter's excellent `cargo xtask sizes` work.

PR #389, opened 2022-01-21
`assembler` → `assembly`

jasikpark:


cbiffle:
@jasikpark ?

jasikpark:
In my mind:
C -> assembly
compiler -> assembler

I wasn't aware of "assembler" being the name of a language rather than being the name of a tool, similar to a compiler.

Good to know I suppose!

PR #390, opened 2022-01-21
ringbuf: add per-buffer enable flag.

cbiffle:
This changes the meaning of the `disabled` feature: it now controls the
initial value of the enable flags. These flags can be changed from
within a program if required, but they're really intended to be written
by the debugger.

### DIFF
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -142,6 +142,14 @@
 /// macros is guaranteed to be able to find them.
 pub use userlib::util::StaticCell;
 
+use core::sync::atomic::{AtomicBool, Ordering};
+
+#[cfg(not(feature = "disabled"))]
+pub const DEFAULT_ENABLE: bool = true;
+
+#[cfg(feature = "disabled")]
+pub const DEFAULT_ENABLE: bool = false;
+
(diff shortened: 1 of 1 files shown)


cbiffle:
(This is a draft to prevent accidentally merging it without corresponding Humility support.)

lib/ringbuf/src/lib.rs
```diff
@@ -235,6 +231,24 @@ pub struct RingbufEntry<T: Copy + PartialEq> {
     pub payload: T,
 }
 
+/// A container for a `Ringbuf` at static scope, which includes a `StaticCell`
+/// to access it safely, and an enable flag.
+pub struct StaticRingbuf<T: Copy + PartialEq, const N: usize> {
+    pub ringbuf_cell: StaticCell<Ringbuf<T, N>>,
+    pub enable: AtomicBool,
```
bcantrill, line 238: Getting greedy here, but how difficult would it be for us to locate this address from Humility? It would be incredible to be able to turn this on dynamically (and apologies if you were thinking in this direction all along!).

bcantrill (commented):
I think this is looking good -- just want to make sure we get the versioning sorted out as part of this so we can continue to evolve Hubris and give Humility crisp failure modes on a mismatch.

PR #391, opened 2022-01-21
Bring up management network on Sidecar

mkeeter:
This PR brings up the management network on Sidecar.  It doesn't _quite_ work: the `task-vsc7448` fails to lock the Tx PLLs on the VSC7448, so it ends up in a reboot loop.  However, it's big enough that we should start looking at it now.

As part of management network bringup, this PR also updates the `net` task to support multiple boards through a compile-time `Bsp`.  Right now, it just supports Sidecar and the Nucleo dev board, but will also support Gimlet through a very similar configuration as Sidecar.

It adds support for various PHYs, and a new `PhyRw` trait, because we talk to PHYs in three different ways:

- Directly (in the `net` task) through the MAC registers
- Indirectly by making IPC calls to the `net` task
- Indirectly by making SPI transactions to the VSC7448's MIIM peripheral

The `PhyRw` trait abstracts over these communication channels, so configuration can just care about what registers to read and write.

There's still a bunch of changes to come, but early feedback is welcome, especially about the code is structured and split into crates!


drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+ringbuf!(Trace, 16, Trace::None);
+
+const fn register_offset(address: u16) -> u16 {
+    let addr10_2 = address >> 2;
+    let mask_shift = 2 /* turn around bits */ + (2 * ((address >> 1) & 0x1));
+    (addr10_2 << 6) | ((0x3 as u16) << mask_shift)
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[repr(u16)]
```
cbiffle, line 32: Nit: do you need this? We've got a lot of enums with explicit `repr`s but it's usually to support deriving `zerocopy`, which you're not.

drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+        Self {
+            spi,
+            nrst,
+            slow_reset,
+        }
+    }
+
+    pub fn read(&self, r: Register) -> Result<u16, SpiError> {
+        let cmd = (r as u16).to_be_bytes();
+        let request = [cmd[0], cmd[1]];
```
cbiffle, line 72: This appears to be a no-op.

drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+        self.spi.exchange(&request, &mut response)?;
+        let v = u16::from_le_bytes(response[2..].try_into().unwrap());
+        ringbuf_entry!(Trace::Read(r, v));
+
+        Ok(v)
+    }
+
+    pub fn write(&self, r: Register, v: u16) -> Result<(), SpiError> {
+        let cmd = (r as u16 | 0x8000).to_be_bytes(); // Set MSB to indicate write.
+        let data = v.to_le_bytes();
```
cbiffle, line 84: (It might be a good idea to leave a comment here so that the be -> le switch doesn't look like a typo.)

drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+        Ok(())
+    }
+
+    pub fn write_masked(
+        &self,
+        r: Register,
+        v: u16,
+        mask: u16,
+    ) -> Result<(), SpiError> {
+        let _v = (self.read(r)? & !mask) | (v & mask);
```
cbiffle, line 98: It's pretty deeply ingrained in Rust that variables starting with underscore are unused, so turning around and using this seems confusing.

drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn enable(&self) -> Result<(), SpiError> {
+        self.write(Register::CIDER, 1)
+    }
+
+    pub fn disable(&self) -> Result<(), SpiError> {
+        self.write(Register::CIDER, 0)
+    }
+
+    /// Configures the KSZ8463 switch in 100BASE-FX mode.
+    pub fn configure(&self) {
```
cbiffle, line 115: I'd consider passing in a `&Gpio` here, actually, since you don't seem to use it anywhere else.

drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+
+use core::convert::TryInto;
+
+use drv_spi_api::{SpiDevice, SpiError};
+use drv_stm32h7_gpio_api as gpio_api;
+use gpio_api::{Gpio, OutputType, PinSet, Pull, Speed};
+use ringbuf::*;
+use userlib::{hl::sleep_for, task_slot};
+
+task_slot!(GPIO, gpio_driver);
```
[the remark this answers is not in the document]
cbiffle: (Further comment on `configure` below suggests an alternative.)

drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        // The datasheet recommends a particular combination of diodes and
+        // capacitors which dramatically slow down the rise of the reset
+        // line, meaning you have to wait for extra long here.
+        //
+        // Otherwise, the minimum wait time is 1 µs, so 1 ms is fine.
+        sleep_for(if self.slow_reset { 150 } else { 1 });
+
+        let id = self.read(Register::CIDER).unwrap();
+        assert_eq!(id & !1, 0x8452);
```
cbiffle, line 141: Will panicking the task that's calling this function fix the communication with the switch? I'd suggest making this an `Err` return and letting the caller decide.

drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+        ringbuf_entry!(Trace::Id(id));
+
+        // Configure for 100BASE-FX operation
+        self.enable().unwrap();
```
cbiffle, line 145: Why is `unwrap` ok here? It has the effect of panicking the task that's calling this if the SPI driver has restarted -- is that a good thing? (Same q on following lines.)

drv/stm32h7-eth/Cargo.toml
```diff
@@ -12,6 +12,7 @@ ipv4 = []
 ipv6 = []
 
 [dependencies]
+cfg-if = "0.1.10"
```
cbiffle, line 15: Nit: `cfg-if = "1"` -- I'm in the process of moving things to the 1.0 version

drv/stm32h7-eth/src/lib.rs
```diff
@@ -387,6 +391,7 @@ pub enum SmiClause22Register {
     MmdAccessControl = 13,
     MmdAccessAddressData = 14,
     ExtendedStatus = 15,
+    TxFxExtendedStatus = 16,
```
cbiffle, line 394: This register does not appear in 802.3 clause 22, at least in the version I've got. Is this a vendor-specific extension?

drv/stm32h7-spi-server/src/main.rs
```diff
@@ -528,7 +528,7 @@ fn activate_mux_option(
             pins.pin_mask,
             gpio_api::Mode::Alternate,
             gpio_api::OutputType::PushPull,
-            gpio_api::Speed::High,
+            gpio_api::Speed::Low,
```
cbiffle, line 531: Hm. This really ought to be configurable, but, this is more likely to work than what we had before as long as people stay below about 10MHz.

drv/user-leds/src/main.rs
```diff
@@ -35,7 +35,11 @@ use userlib::*;
 
 cfg_if::cfg_if! {
     // Target boards with 4 leds
-    if #[cfg(any(target_board = "gemini-bu-1", target_board = "gimletlet-2"))] {
+    if #[cfg(any(
+            target_board = "gemini-bu-1",
+            target_board = "gimletlet-1",
```
cbiffle, line 40: Out of curiosity, I see several deltas for supporting gimletlet-1; do you actually have one you're using?

drv/vsc7448/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "vsc7448"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+cfg-if = "0.1.10"
```
cbiffle, line 7: (1.0)

drv/ksz8463/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "ksz8463"
+version = "0.1.0"
+edition = "2018"
```
cbiffle, line 4: (Consider setting 2021 for new crates)

drv/vsc7448/src/dev.rs
```diff
@@ -0,0 +1,206 @@
...
+/// that use either one.
+#[derive(Copy, Clone)]
+pub enum DevGeneric {
+    Dev1g(u32),
+    Dev2g5(u32),
+}
+
+impl DevGeneric {
+    pub fn new_1g(d: u32) -> Self {
+        assert!(d < 24);
```
cbiffle, line 22: I'd be more comfortable with these asserts if the functions were documented, I'm not sure how to tell if my `d` is valid or why.

drv/vsc7448/src/dev.rs
```diff
@@ -0,0 +1,206 @@
...
+    spi::Vsc7448Spi,
+    VscError,
+};
+use vsc7448_pac::Vsc7448;
+
+/// DEV1G and DEV2G5 share the same register layout, so we can write functions
+/// that use either one.
+#[derive(Copy, Clone)]
+pub enum DevGeneric {
+    Dev1g(u32),
```
cbiffle, line 16: It looks like the fields here have a range of 0..23 and 0..28, respectively; you could use `u8` here and make them cheaper to store. I don't know if you store them or anything though.

drv/vsc7448/src/dev.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub fn new_1g(d: u32) -> Self {
+        assert!(d < 24);
+        DevGeneric::Dev1g(d)
+    }
+    pub fn new_2g5(d: u32) -> Self {
+        assert!(d < 29);
+        DevGeneric::Dev2g5(d)
+    }
+    /// Convert from a DEV to a port number, based on Table 12 in the datasheet
+    pub fn port(&self) -> u32 {
```
cbiffle, line 30: It looks like the 1g and 2g5 ports will alias in the result of this function, is that intentional?

drv/vsc7448/src/dev.rs
cbiffle: (It's worth noting that all of the arithmetic you're doing on dev numbers below implies overflow checks -- storing a u8 and converting to a u32 before arithmetic _should_ make it clear to the compiler that overflow can't happen. This may or may not result in a text size decrease, depending on whether the compiler had already inlined enough things to see this on its own.)

drv/vsc7448/src/dev.rs
```diff
@@ -0,0 +1,206 @@
...
+                vsc7448_pac::DEV1G::from_raw_unchecked_address(
+                    vsc7448_pac::DEV2G5::BASE + d * vsc7448_pac::DEV1G::SIZE,
+                )
+            }
+        }
+    }
+
+    /// Based on `jr2_port_conf_1g_set` in the SDK
+    pub fn init_sgmii(&self, v: &Vsc7448Spi) -> Result<(), VscError> {
+        // Flush the port before doing anything else
```
cbiffle, line 77: (I'm taking the contents of these blocks on faith :-) )

drv/vsc7448/src/port.rs
```diff
@@ -0,0 +1,166 @@
...
+                empty &= value.maxuse() == 0;
+                // Keep looping, because these registers are clear-on-read,
+                // so it's more efficient to read them all, even if we know
+                // that the port isn't currently empty.
+            }
+        }
+        if empty {
+            return Ok(());
+        }
+        hl::sleep_for(1);
```
cbiffle, line 163: Your other sleeps have comments justifying the duration choices. Do we have a justification for this one?

Do we know how often this has to retry? If the answer is >50%, should we sleep _before_ checking to reduce load?

drv/vsc7448/src/serdes1g.rs
```diff
@@ -0,0 +1,95 @@
...
+    mbtr_ctrl: u32,
+    des_bw_ana: u32,
+    ena_lane: bool,
+}
+/// Based on `jr2_sd1g_cfg`
+impl Config {
+    pub fn new(m: Mode) -> Self {
+        match m {
+            Mode::Sgmii => Self {
+                ob_amp_ctrl: 12,
```
cbiffle, line 30: Where would I check these against, were I to wish to?

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+    None,
+    Read { addr: u32, value: u32 },
+    Write { addr: u32, value: u32 },
+}
+
+/// This indicates how many bytes we pad between (writing) the address bytes
+/// and (reading) data back, during SPI transactions to the VSC7448.  See
+/// section 5.5.2 for details.  1 padding byte should be good up to 6.5 MHz
+/// SPI clock.
+pub const SPI_PAD_BYTES: u8 = 1;
```
cbiffle, line 21: You seem to use this const as a usize -- I suggest making it a usize. That would prevent misinterpretation that 1 _is the pad byte._

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+
+// Flags to tune ringbuf output while developing
+const DEBUG_TRACE_SPI: u8 = 1 << 0;
+const DEBUG_MASK: u8 = 0;
+
+/// Writes the given value to the ringbuf if allowed by the global `DEBUG_MASK`
```
cbiffle, line 27: You are gonna love #390 once I figure out the Humility side.

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+            ringbuf_entry!($value);
+        }
+    };
+}
+ringbuf!(Trace, 16, Trace::None);
+
+////////////////////////////////////////////////////////////////////////////////
+
+/// Helper struct to read and write from the VSC7448 over SPI
+pub struct Vsc7448Spi(pub SpiDevice);
```
cbiffle, line 40: Does other code jump past this type and do direct spi transactions? If not, the field should probably not be `pub`.

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+impl Vsc7448Spi {
+    /// Reads from a VSC7448 register
+    pub fn read<T>(&self, reg: RegisterAddress<T>) -> Result<T, VscError>
+    where
+        T: From<u32>,
+    {
+        assert!(reg.addr >= 0x71000000);
```
cbiffle, line 47: Having these preconditions mentioned in the comment on this function would be good.

(Text size nit: doing one assert using `&&` would likely be cheaper.)

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+        assert!(reg.addr <= 0x72000000);
+        let addr = (reg.addr & 0x00FFFFFF) >> 2;
+        let data: [u8; 3] = [
+            ((addr >> 16) & 0xFF) as u8,
```
cbiffle, line 51: Nit 1: you don't need the `& 0xFF` if you then `as u8`, it's a truncating conversion.

Nit 2: this is doing a big-endian 24-bit integer. I had to think for a bit to work that out. Consider whether there's a way to leverage `to_be_bytes` here.

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+        ];
+
+        // We read back 7 + padding bytes in total:
+        // - 3 bytes of address
+        // - Some number of padding bytes
+        // - 4 bytes of data
+        const SIZE: usize = 7 + SPI_PAD_BYTES as usize;
+        let mut out = [0; SIZE];
+        self.0.exchange(&data[..], &mut out[..])?;
+        let value = (out[SIZE - 1] as u32)
```
cbiffle, line 63: Consider
```rust
let value = u32::from_le_bytes(out[SIZE - 4..].try_into().unwrap());
```

cbiffle: ...or is it `be_bytes`? I think it might be big-endian? Negative indices are confusing, please consider using an operation with a name. :-)

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+                panic!(
+                    "Got invalid read sentinel value while reading IF_CFGSTAT"
+                );
+            }
+            // This read should _never_ fail for timing reasons because the
+            // DEVCPU_ORG register block can be accessed faster than all other
+            // registers (section 5.3.2 of the datasheet).
+            let v = match self.read(if_cfgstat) {
+                Err(e) => {
+                    panic!("Failed read while trying to check IF_STAT: {:?}", e)
```
cbiffle, line 108: Nit: this is equivalent to

```rust
let v = self.read(if_cfgstat)
    .expect("Failed read while trying to check IF_STAT");
```

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// if you want to modify it, then use [Self::modify] instead.
+    pub fn write<T>(
+        &self,
+        reg: RegisterAddress<T>,
+        value: T,
+    ) -> Result<(), VscError>
+    where
+        u32: From<T>,
+    {
+        assert!(reg.addr >= 0x71000000);
```
cbiffle, line 129: (same comment about documenting this precondition)

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+        self.write(reg, (((value >> 32) as u32) & 0x1FFFFF).into())
+    }
+
+    /// Performs a write operation on the given register, where the value is
+    /// calculated by calling f(0).  This is helpful as a way to reduce manual
+    /// type information.
+    pub fn write_with<T, F>(
+        &self,
+        reg: RegisterAddress<T>,
+        f: F,
```
cbiffle, line 177: (fwiw, I have taken to writing this as `f: impl Fn(&mut T)` to cut down on parameters and where-clause-lines, but the two are equivalent as long as you don't otherwise need to refer to `F` by name)

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Reads from a specific SERDES6G instance, which is done by writing its
+    /// value (as a bitmask) to a particular register with a read flag set,
+    /// then waiting for the flag to autoclear.
+    pub fn serdes6g_read(&self, instance: u32) -> Result<(), VscError> {
+        let mut reg: vsc7448_pac::hsio::mcb_serdes6g_cfg::MCB_SERDES6G_ADDR_CFG =
+            0.into();
+        reg.set_serdes6g_rd_one_shot(1);
+        reg.set_serdes6g_addr(1 << instance);
+        let addr = Vsc7448::HSIO().MCB_SERDES6G_CFG().MCB_SERDES6G_ADDR_CFG();
+        self.write(addr, reg)?;
```
cbiffle, line 214: Could this be done using `write_with`?

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+        for _ in 0..32 {
```
cbiffle, line 215: This fixed iteration count seems suspicious. Should it be expressed in terms of wall time? Either way, a comment would be good.

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+            if self.read(addr)?.serdes6g_rd_one_shot() != 1 {
+                return Ok(());
+            }
+        }
+        return Err(VscError::Serdes6gReadTimeout { instance });
```
cbiffle, line 220: (nit: equivalent to `Err(Vsc::...)`)

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Writes to a specific SERDES6G instance, which is done by writing its
+    /// value (as a bitmask) to a particular register with a read flag set,
+    /// then waiting for the flag to autoclear.
+    pub fn serdes6g_write(&self, instance: u32) -> Result<(), VscError> {
+        let mut reg: vsc7448_pac::hsio::mcb_serdes6g_cfg::MCB_SERDES6G_ADDR_CFG =
+            0.into();
+        reg.set_serdes6g_wr_one_shot(1);
+        reg.set_serdes6g_addr(1 << instance);
+        let addr = Vsc7448::HSIO().MCB_SERDES6G_CFG().MCB_SERDES6G_ADDR_CFG();
+        self.write(addr, reg)?;
```
cbiffle, line 232: (this also seems like it could be `write_with`)

drv/vsc7448/src/spi.rs
cbiffle: (same in other cases in this file)

task/net/src/bsp/nucleo_h7.rs
```diff
@@ -0,0 +1,107 @@
...
+//
+use crate::GPIO;
+use drv_stm32h7_eth as eth;
+
+/// Address used on the MDIO link by our Ethernet PHY. Different
+/// vendors have different defaults for this, it will likely need to
+/// become configurable.
+const PHYADDR: u8 = 0x01;
+
+pub struct Bsp {
```
cbiffle, line 13:
```rust
pub struct Bsp;
```

task/net/src/bsp/sidecar_1.rs
```diff
@@ -0,0 +1,244 @@
...
+use drv_stm32h7_eth as eth;
+use drv_stm32h7_gpio_api as gpio_api;
+use ksz8463::{Ksz8463, Register as KszRegister};
+use ringbuf::*;
+use userlib::{hl::sleep_for, task_slot};
+use vsc7448_pac::types::PhyRegisterAddress;
+use vsc85xx::{Phy, PhyRw, PhyVsc85xx, VscError};
+
+task_slot!(SPI, spi_driver);
+const KSZ8463_SPI_DEVICE: u8 = 0; // Based on app.toml ordering
```
cbiffle, line 16: Yeah, sorry about this, I need to finish the code to generate a named enumeration. :-\

task/net/src/bsp/sidecar_1.rs
```diff
@@ -0,0 +1,244 @@
...
+        let gpio = Gpio::from(GPIO.get_task_id());
+        let eth_af = Alternate::AF11;
+
+        // RMII
+        gpio.configure(
+            Port::A,
+            (1 << 1) | (1 << 7),
+            Mode::Alternate,
+            OutputType::PushPull,
+            Speed::VeryHigh,
```
cbiffle, line 72: (FWIW, from the datasheet, VeryHigh may actually be higher than we need for 50MHz RMII, but it'll depend on capacitance.)

task/net/src/bsp/sidecar_1.rs
```diff
@@ -0,0 +1,244 @@
...
+        // The KSZ8463 connects to the SP over RMII, then sends data to the
+        // VSC8552 over 100-BASE FX
+        self.ksz.configure();
+
+        // The VSC8552 connects the KSZ switch to the management network
+        // over SGMII
+        configure_vsc8552(eth);
+    }
+
+    pub fn wake(&self, eth: &mut eth::Ethernet) {
```
cbiffle, line 131: So...what's this for?

task/net/src/bsp/gimletlet_mgmt.rs
```diff
@@ -0,0 +1,223 @@
...
+            ringbuf_entry!(Trace::Vsc8552Status100 { port, status });
+        }
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+/// Helper struct to implement the `PhyRw` trait using direct access through
+/// `eth`'s MIIM registers.
+struct MiimBridge<'a> {
```
cbiffle, line 162: This looks identical to the one in sidecar_1?

task/vsc7448/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "task-vsc7448"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+cfg-if = "0.1.10"
```
cbiffle, line 7: ("1.0")

task/vsc7448/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+userlib = {path = "../../sys/userlib", features = ["panic-messages"]}
+ringbuf = {path = "../../lib/ringbuf" }
+drv-spi-api = {path = "../../drv/spi-api"}
+drv-stm32h7-gpio-api = {path = "../../drv/stm32h7-gpio-api", default-features = false}
```
cbiffle, line 11: FWIW this library _has no_ features, so, you don't need to turn them off.

drv/user-leds/src/main.rs
mkeeter: OH YEAH!
IMG_5982

drv/ksz8463/src/lib.rs
mkeeter: It appears to be needed, otherwise I get errors of the form
```
  --> drv/ksz8463/src/lib.rs:32:13
   |
32 |     CIDER = register_offset(0x000),
   |             ^^^^^^^^^^^^^^^^^^^^^^ expected `isize`, found `u16`
```

I want to do a second pass through this, either before or after merging, since this is left over from Arjen's original `100base-fx` branch and doesn't match in style.

drv/vsc7448/src/serdes1g.rs
mkeeter: There's a docstring at the [top of the struct](https://github.com/oxidecomputer/hubris/pull/391/files/1d809c08e4e9703dd8b811394b6d8faaed3ec4be#diff-0de703a68cb440b6446654586e5e97a06f16a35eaf2424b390d395664390f1c7R25)

drv/vsc7448/src/spi.rs
mkeeter: Good idea; I also renamed it to `SPI_NUM_PAD_BYTES` for clarity.

drv/vsc7448/src/spi.rs
mkeeter: Good catch, I've made it private (I think it was `pub` just to avoid needing a constructor, but this is better)

drv/vsc7448/src/spi.rs
mkeeter: Oddly enough, this doesn't seem equivalent!  If I make that change, I get these errors:

```
error[E0282]: type annotations needed
   --> drv/vsc7448/src/dev.rs:129:11
    |
129 |         v.write_with(dev1g.MAC_CFG_STATUS().MAC_ENA_CFG(), |r| {
    |           ^^^^^^^^^^ cannot infer type for type parameter `F` declared on the associated function `write_with`

error[E0282]: type annotations needed
  --> drv/vsc7448/src/lib.rs:37:7
   |
37 |     v.write_with(Vsc7448::DEVCPU_ORG().DEVCPU_ORG().IF_CFGSTAT(), |r| {
   |       ^^^^^^^^^^ cannot infer type for type parameter `F` declared on the associated function `write_with`
```

drv/vsc7448/src/spi.rs
mkeeter: Yup, and a few others in this file as well!

task/vsc7448/Cargo.toml
mkeeter: Okay!  This is wrong in other places, so I opened #418 to fix it elsewhere.

drv/ksz8463/src/lib.rs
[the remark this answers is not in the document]
mkeeter: Good rule of thumb, fixed!

drv/ksz8463/src/lib.rs
mkeeter: Done!

drv/ksz8463/src/lib.rs
mkeeter: Fixed!

drv/vsc7448/src/spi.rs
mkeeter: I switched to returning an error if the caller is out of range, instead of panicking.

drv/vsc7448/src/spi.rs
mkeeter: I ended up reworking this whole section to return errors instead!

drv/vsc7448/src/spi.rs
mkeeter: (same reply about now returning an error instead 😄)

drv/vsc7448/src/dev.rs
mkeeter: Yup, that's intentional; I added a more detailed explanation.

drv/stm32h7-eth/src/lib.rs
mkeeter: Yes, this is a vendor-specific extension.  I've removed in favor of using registers from the VSC7448 PAC, which is already used in the BSP that was using `TxFxExtendedStatus`.

I'm not totally happy having the definitions in two places, though.

[Here's the main definitions in the PAC](https://github.com/oxidecomputer/vsc7448/blob/master/vsc7448-pac/src/phy.rs#L44), and [here's an example of register structs](https://github.com/oxidecomputer/vsc7448/blob/master/vsc7448-pac/src/phy/standard.rs#L32).

We should chat about this when we discuss live.  We could remove `eth::SmiClause22Register` entirely; the downside would be compiling an extra relatively heavy crate (once).

drv/vsc7448/src/dev.rs
mkeeter: I updated the PAC to use the minimum-sized integers for indexes of things like `DEV1G`, and updated all of this code as well!

task/net/src/bsp/gimletlet_mgmt.rs
mkeeter: Yup, deduplicated and moved to a single place.

drv/vsc7448/src/dev.rs
mkeeter: Added a docstring and switched to returning an error on invalid `d`.

drv/vsc7448/src/spi.rs
mkeeter: Good idea! (I had to check on playground, but it turns out to be `be_bytes`)

drv/vsc7448/src/spi.rs
mkeeter: Done!

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,271 @@
...
+    where
+        u32: From<T>,
+    {
+        if reg.addr < 0x71000000 || reg.addr > 0x72000000 {
+            return Err(VscError::BadRegAddr(reg.addr));
+        }
+
+        let addr = (reg.addr & 0x00FFFFFF) >> 2;
+        let value = u32::from(value);
+        let mut data: [u8; 7] = [0; 7];
```
mkeeter, line 132: @cbiffle I also switched to using `to_be_bytes` here!

drv/vsc7448/src/port.rs
mkeeter: Bump this to 2000, which is what the SDK does, and leave a comment to that effect

drv/vsc7448/src/spi.rs
```diff
@@ -0,0 +1,271 @@
...
+        // DEVCPU_ORG:IF_CFGSTAT.IF_STAT.  If that bit is set, then the sentinel
+        // value _actually_ indicates an error (and not just an unfortunate
+        // coincidence).
+        if value == 0x88888888 {
+            // Return immediately if we got an invalid read sentinel while
+            // reading IF_CFGSTAT itself.  This check also protects us from a
+            // stack overflow.
+            let if_cfgstat = Vsc7448::DEVCPU_ORG().DEVCPU_ORG().IF_CFGSTAT();
+            if reg.addr == if_cfgstat.addr {
+                return Err(VscError::InvalidRegisterReadNested);
```
mkeeter, line 100: Consider a `panic!`

drv/ksz8463/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+task_slot!(GPIO, gpio_driver);
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum Trace {
+    None,
+    Read(Register, u16),
+    Write(Register, u16),
+    Id(u16),
+}
+ringbuf!(Trace, 16, Trace::None);
```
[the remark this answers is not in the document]
mkeeter: Decided not to do anything about, since it would be a weird choice to have two tasks that both drive KSZ chips (we have no hardware with > 1 KSZ chip, and no plans to build such a thing).

drv/ksz8463/src/lib.rs
mkeeter: We talked about this in person and decided that panicking is okay, since this is only called at initialization and rebooting the task is a reasonable response to such a failure.

drv/ksz8463/src/lib.rs
mkeeter: (same as the panic above, this is during initialization, so panicking seems okay-ish)

drv/stm32h7-eth/src/lib.rs
mkeeter: When talked live, the consensus was to preserve `eth::SmiClause22Register` as a lightweight way to talk to registers.  I added a detailed comment suggesting using the PAC register addresses if that crate is pulled in.

drv/vsc7448/src/spi.rs
mkeeter: Until then, I turned this on to run unconditionally; the masking was useful when it was part of a larger collection of trace options, but now this is in a standalone file, so it doesn't crowd out other info.

drv/vsc7448/src/spi.rs
mkeeter: ...and change my mind and switched back, after documenting the precondition.

drv/vsc7448/src/spi.rs
mkeeter: ... then switched the nested read failure to a panic, because it indicates something went _deeply_ wrong.

drv/vsc7448/src/spi.rs
mkeeter: Added some documentation about this and moved it to a constant.  There's a pending `TODO` to see whether this actually takes more than one iteration...

task/net/src/bsp/sidecar_1.rs
mkeeter: Moved out of the Idol API, added a new `Option<u64>` to ask the `net` task to run periodic logging at a BSP-specified interval.

drv/ksz8463/src/lib.rs
mkeeter: I ended up reworking where this conversion happens, so the `repr(u16)` is no longer needed

drv/ksz8463/src/lib.rs
mkeeter: Done, along with lots of docs on the terrible addressing mode!

PR #393, opened 2022-01-22
Allocate kernel task resources statically

### LINKED ISSUE #236: Kernel out of startup-heap behavior is currently difficult to debug
The kernel currently uses a bump-pointer allocator to allocate task structures at startup, from its free RAM section. When the free RAM section is not long enough for this, it will panic, and this panic is difficult to diagnose if you're not watching over ITM at the time. We should fix this. Potential approaches off the top of my head include

- Figure out how to allocate that stuff statically, converting this into a build failure.
- Make the panic behavior better
- Add kernel panic diagnosis to humilty

cbiffle:
This should fix #236 by moving kernel RAM allocation to link time, and so pulling the corresponding failure to build time. The decision to allocate "dynamically" (by bump pointer) during kernel startup predates the build system; now that we have a build system, we can use it to stop doing this.

This is a _very_ breaking change from Humility's perspective, so, merging it will depend on some other work.


build/xtask/src/dist.rs
```diff
@@ -1224,6 +1215,31 @@ fn allocate_all(
...
+fn allocate_k(
+    region: &str,
+    size: u32,
+    avail: &mut Range<u32>,
+) -> Result<Range<u32>> {
+    // Our base address will be larger than avail.start if it doesn't meet our
+    // minimum requirements. Round up.
+    let base = (avail.start + 15) & !15;
+
+    if base >= avail.end || size > avail.end - base {
```
mkeeter, line 1227: Nit – could you do
```
    let end = base + size;
    if end > avail.end {
        ...
```
instead of checking both conditions?

This would produce different results if `base == avail.end && size == 0`, but I'm not sure what answer is correct there...

sys/kern/build.rs
```diff
@@ -45,3 +54,110 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    )?;
+    writeln!(
+        file,
+        "const HUBRIS_TASK_COUNT: usize = {};",
+        kconfig.tasks.len()
+    )?;
+
+    writeln!(
+        file,
+        "static HUBRIS_TASK_DESCS: [abi::TaskDesc; {}] = [",
```
mkeeter, line 81: Nit – You could do
```rust
"static HUBRIS_TASK_DESCS: [abi::TaskDesc; HUBRIS_TASK_COUNT] = ["
```
since you defined `HUBRIS_TASK_COUNT` above

sys/kern/build.rs
```diff
@@ -45,3 +54,110 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            task.flags.bits()
+        )?;
+        writeln!(file, "    }},")?;
+    }
+    writeln!(file, "];")?;
+
+    writeln!(
+        file,
+        "static mut HUBRIS_TASK_TABLE_SPACE: \
+        core::mem::MaybeUninit<[crate::task::Task; {}]> = \
```
mkeeter, line 107: Same here!

sys/kern/build.rs
```diff
@@ -45,3 +54,110 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        writeln!(file, "        notification: 0b{:b},", irq.notification)?;
+        writeln!(file, "    }},")?;
+    }
+    writeln!(file, "];")?;
+
+    Ok(())
+}
+
+#[derive(Deserialize)]
+struct KernelConfig {
```
mkeeter, line 158: Is there an easy way to share this with the `xtask/src/dist.rs` definition?

sys/kern/src/startup.rs
```diff
@@ -161,25 +127,40 @@ fn safe_start_kernel(
     // As a small optimization, we equip each task with an array of references
     // to RegionDecs, instead of looking them up by index each time. Generate
     // these.
-    let region_tables = alloc.gimme_n(app_header.task_count as usize, |i| {
-        let mut table = [&region_descs[0]; app::REGIONS_PER_TASK];
+
+    // Safety: MaybeUninit<[T]> -> [MaybeUninit<T>] is defined as safe.
+    let region_tables: &mut [[MaybeUninit<&'static app::RegionDesc>; app::REGIONS_PER_TASK];
+             HUBRIS_TASK_COUNT] =
+        unsafe { &mut *(region_tables as *mut _ as *mut _) };
```
mkeeter, line 134: Out of curiosity, what's the intermediate state in this chain of casts?

cbiffle: `&mut MaybeUninit<[T]>`
to `*mut MaybeUninit<[T]>`
to `*mut [MaybeUninit<T>]`
then `&mut [MaybeUninit<T>]`

cbiffle: I can document this further if you want, but this is basically a mechanical dance I do to avoid needing transmute in cases like this -- the start and end types pretty much tell the story.

sys/kern/build.rs
cbiffle: Nooooot right now. :-( I would like to improve this in a followup, we have a number of config data structures that could use to be factored out into a crate.

build/xtask/src/dist.rs
cbiffle: ...I _believe_ that yours is equivalent, yes.

I also note that `Range::contains` has been stabilized in 1.35.0, so we could probably make this even less silly.

sys/kern/build.rs
cbiffle: Correct! Will fix.

bcantrill (approved):
Looks great -- and I have verified that Humility post https://github.com/oxidecomputer/humility/pull/87 works with the new `HUBRIS_IMAGE_ID`.

mkeeter (approved):
Looks good to me!

build/xtask/src/dist.rs
```diff
@@ -1530,58 +1553,18 @@ fn make_descriptors(
...
-
-    Ok(words)
+    Ok(KernelConfig {
+        irqs,
+        tasks: task_descs,
+        regions,
+        supervisor_notification: if let Some(supervisor) = supervisor {
+            supervisor.notification
+        } else {
+            // TODO: this exists for back-compat with incredibly early Hubris,
```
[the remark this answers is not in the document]
cbiffle: I can do that in a followup, it involves changes to the TOML schema so it'll be a bit of a shotgun change.

PR #396, opened 2022-01-24
Networking for h753

jeffallen:
It turns out the h743 and the h753 are so similar for
DMA and Etherent that they can use the same config.


jeffallen:
@mkeeter I wanted to see if my growing understanding of Hubris would let me get my h753 board talking on the network... and I managed. I don't know if this is useful or interesting to you, but it sure forced me to learn a lot. :)

drv/stm32h7-eth/Cargo.toml
```diff
@@ -12,9 +12,10 @@ ipv4 = []
 ipv6 = []
 
 [dependencies]
+cfg-if = "0.1.10"
 cortex-m = "0.7"
 userlib = {path = "../../sys/userlib"}
-stm32h7 = {version = "0.13", default-features = false}
+stm32h7 = {version = "0.13"}
```
cbiffle, line 18: You want to keep `default-features = false` here, the stm32 PAC crates turn on the `rt` feature by default, which is inappropriate unprivileged task code like this.

task/net/Cargo.toml
```diff
@@ -10,12 +10,12 @@ cortex-m = {version = "0.7", features = ["inline-asm"]}
...
 serde = {version = "1", default-features = false, features = ["derive"]}
 ssmarshal = {version = "1", default-features = false}
-stm32h7 = {version = "0.13", features = ["stm32h743"]}
+stm32h7 = {version = "0.13"}
 zerocopy = "0.6"
 
 userlib = {path = "../../sys/userlib", features = ["panic-messages"]}
 hubris-num-tasks = {path = "../../sys/num-tasks", features = ["task-enum"]}
-drv-stm32h7-eth = {path = "../../drv/stm32h7-eth", features = ["with-smoltcp", "ipv6", "h743"]}
+drv-stm32h7-eth = {path = "../../drv/stm32h7-eth"}
```
cbiffle, line 18: This won't work without the ipv6 and with-smoltcp features, so, please move them back up here; only the h743/53 part needs to move below.

app/demo-stm32h7-nucleo/app-h753.toml
```diff
@@ -247,6 +278,13 @@ size = 4096
 #address = 0x48021000
 #size = 4096
 
+[peripherals.eth]
+address = 0x40028000
+size = 0x1000
+
+[peripherals.eth_dma]
+address = 0x40029000
+size = 0x400
```
cbiffle, line 287: nit: blank line after this

task/net/Cargo.toml
```diff
@@ -10,12 +10,12 @@ cortex-m = {version = "0.7", features = ["inline-asm"]}
 num-traits = {version = "0.2.12", default-features = false}
 serde = {version = "1", default-features = false, features = ["derive"]}
 ssmarshal = {version = "1", default-features = false}
-stm32h7 = {version = "0.13", features = ["stm32h743"]}
+stm32h7 = {version = "0.13"}
```
cbiffle, line 13: ...and actually it looks like I accidentally omitted `default-features = false` from this line, before you arrived. :-)

task/net/src/bsp/sidecar_1.rs
```diff
@@ -22,7 +22,7 @@ enum Trace {
     KszId(u16),
     Vsc8552Status { port: u8, status: u16 },
 }
-ringbuf!(Trace, 16, Trace::None);
+ringbuf!(SIDECAR_RINGBUF, Trace, 16, Trace::None);
```
cbiffle, line 25: I'm confused about this part of the change, can you describe what happened without this? It may point to a bug in the ringbuf macros.

task/net/src/main.rs
```diff
@@ -10,15 +10,32 @@ mod buf;
 mod server;
 
 use core::sync::atomic::{AtomicU32, Ordering};
-use stm32h7::stm32h743 as device;
-
+cfg_if::cfg_if! {
+    if #[cfg(any(target_board = "nucleo-h743zi2", target_board = "sidecar-1"))] {
+    use stm32h7::stm32h743 as device;
```
cbiffle, line 15: rustfmt doesn't fix macros every time, unfortunately -- please indent the bodies of these curly braced blocks.

task/net/src/main.rs
```diff
@@ -10,15 +10,32 @@ mod buf;
 mod server;
 
 use core::sync::atomic::{AtomicU32, Ordering};
-use stm32h7::stm32h743 as device;
-
+cfg_if::cfg_if! {
```
cbiffle, line 13: This should probably be controlled by an h743/h753 feature rather than keying off board name. It's possible I just don't understand why you did it this way, however, so, let me know.

task/net/src/main.rs
```diff
@@ -85,6 +102,10 @@ fn main() -> ! {
     let mac = EthernetAddress::from_bytes(&FAKE_MAC);
 
     let ipv6_addr = link_local_iface_addr(mac);
+    ringbuf_entry!(Trace::Addr {
```
cbiffle, line 105: You definitely don't need 16 elements in a ringbuffer that's only written once. If your goal is to discover the IP address of the Nucleo board, a global variable accessed by `readvar` would be much more space-efficient. (and also the IP address is fixed to `fe80::4:6ff:fe08:0a0c`)

task/net/src/bsp/sidecar_1.rs
jeffallen: I understood that sidecar.rs and main.rs can't both have unnamed rings, but maybe that's wrong.

task/net/src/main.rs
jeffallen: Frankly, I was playing and learning, and thought I might add other trace messages. But for checking it in for real, I'll nuke all this.

Thanks for the feedback!

task/net/Cargo.toml
jeffallen: I understood that the features from the [features] section below replace these. I'll try and see if it's additive instead. I am a strong reader of Rust, but a very new writer, and Cargo is a total mystery to me...

drv/stm32h7-eth/Cargo.toml
jeffallen: OK, understood. In my $HOME/.cargo/registry/src/github.com-1ecc6299db9ec823/stm32h7-0.13.0/Cargo.toml file, I see:

```
[features]
default = []
```

So I thought that removing it like net/task/Cargo.toml was more right.

task/net/src/main.rs
jeffallen: Because mac will only be known at runtime, what you're asking for is to use a mutable static, and Rust really would prefer I didn't do that (`error[E0133]: use of mutable static is unsafe and requires unsafe function or block`). I looked into the technique used by buf.rs, and I guess it would work, but it would be ugly for just this debug info. I also did it with unsafe, and that worked, but then the variable didn't end up visible with humility or objdump. Is there such a thing as "inlined statics"? I can't figure this out, so I nuked it for the time being.

drv/stm32h7-eth/Cargo.toml
cbiffle: They changed the default in 0.14, and I'm in the process of upgrading things. Thanks for future-proofing it. :-)

app/sidecar/app.toml
```diff
@@ -124,7 +124,7 @@ path = "../../task/net"
 name = "task-net"
 stacksize = 3800
 priority = 2
-features = ["mgmt"]
+features = ["mgmt", "h743"]
```
mkeeter, line 127: Sidecar is actually an `h753`!

mkeeter:
Thanks, this looks good other than one config error!

jeffallen:
@mkeeter If the CI turns green, then this is ready for you to take into your vsc7448 branch.

Have a nice weekend!

mkeeter:
Thanks!

PR #398, opened 2022-01-25
Introduce UnwrapLite.

cbiffle:
This provides a syntactically-convenient equivalent to unwrap() on
Result/Option that is _significantly_ cheaper from a code generation
perspective, because it doesn't invoke any type's formatting routines.

This knocked like 5 kiB off one incarnation of the kernel.

### DIFF
--- a/sys/userlib/src/kipc.rs
+++ b/sys/userlib/src/kipc.rs
@@ -4,6 +4,7 @@
 
 //! Operations implemented by IPC with the kernel task.
 
+use unwrap_lite::UnwrapLite;
 use zerocopy::AsBytes;
 
 use crate::*;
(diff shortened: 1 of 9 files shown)


cbiffle:
(updated to remove an experimental use of `unsafe` that accidentally leaked into this commit from another change.)

sys/userlib/src/kipc.rs
```diff
@@ -15,19 +16,16 @@ pub fn read_task_status(task: usize) -> abi::TaskState {
     let (rc, len) =
         sys_send(TaskId::KERNEL, 1, task.as_bytes(), &mut response, &[]);
     assert_eq!(rc, 0);
-    ssmarshal::deserialize(&response[..len])
-        .map_err(|_| ())
-        .unwrap()
+    ssmarshal::deserialize(unsafe { response.get_unchecked(..len) })
```
mkeeter, line 19: Can you add a `SAFETY` comment here?

cbiffle: Branch mixup, unsafe removed.

mkeeter (approved):

PR #399, opened 2022-01-25
Add cargo xtask sizes, to print sizes and suggest improvements

mkeeter:
Sample output for `cargo xtask sizes app/sidecar/app.toml`:

```
jefe
  flash:  7148 bytes (43%)
  ram:    1620 bytes (79%)

rcc_driver
  flash:  4156 bytes (50%)
  ram:    1024 bytes (100%)

gpio_driver
  flash:  5400 bytes (65%)
  ram:    1024 bytes (100%)

spi2_driver
  flash:  9256 bytes (56%)
  ram:    2036 bytes (49%)

spi5_driver
  flash:  9276 bytes (56%)
  ram:    2036 bytes (49%)

i2c_driver
  flash:  8756 bytes (53%)
  ram:    2008 bytes (98%)

hiffy
  flash: 15456 bytes (47%)
  ram:    8752 bytes (26%)

sensor
  flash:  4372 bytes (26%)
  ram:    2000 bytes (97%)

sidecar_seq
  flash: 19296 bytes (58%)
  ram:    2316 bytes (56%)

idle
  flash:    88 bytes (34%)
  ram:     256 bytes (100%)

kernel
  flash: 31732 bytes (96%)
  ram:    1072 bytes (26%)

jefe:
  flash: 8192 (from 16384)
spi2_driver:
  ram:   2048 (from 4096)
spi5_driver:
  ram:   2048 (from 4096)
hiffy:
  flash: 16384 (from 32768)
  ram:   16384 (from 32768)
sensor:
  flash: 8192 (from 16384)
idle:
  flash: 128 (from 256)
kernel:
  ram:   2048 (from 4096)

------------ Total savings ------------
  flash: 32896
  ram:   22528
```

### DIFF
--- a/build/xtask/src/sizes.rs
+++ b/build/xtask/src/sizes.rs
@@ -1,0 +1,159 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::convert::TryInto;
+use std::path::{Path, PathBuf};
+
+use anyhow::bail;
+use goblin::Object;
+use indexmap::IndexMap;
+
+use crate::Config;
+
+fn armv7m_suggest(size: u64) -> u64 {
+    // Nearest power of two
+    let mut i = 1;
+    while i < size {
+        i *= 2;
+    }
+    i
+}
+fn armv8m_suggest(size: u64) -> u64 {
+    // Nearest chunk of 32
+    ((size + 31) / 32) * 32
+}
+
+pub fn run(cfg: &Path) -> anyhow::Result<()> {
+    let cfg_contents = std::fs::read(&cfg)?;
+    let toml: Config = toml::from_slice(&cfg_contents)?;
+
+    let mut dist_dir = PathBuf::from("target");
+    dist_dir.push(&toml.name);
+    dist_dir.push("dist");
+
+    let mut memories = IndexMap::new();
+    for (name, out) in &toml.outputs {
+        // This is called after `dist`, which logs this overflow gracefully
+        let end = out.address.checked_add(out.size).unwrap();
+        memories.insert(name.clone(), out.address..end);
+    }
+
+    let output_region = |vaddr: u64| {
+        memories
+            .iter()
+            .find(|(_, region)| region.contains(&vaddr.try_into().unwrap()))
+            .map(|(name, _)| name.as_str())
+    };
+
+    let mut first = false;
+
+    struct Task<'a> {
+        name: &'a str,
+        stacksize: u32,
+        requires: &'a IndexMap<String, u32>,
+    }
+    let mut tasks: Vec<Task> = toml
+        .tasks
+        .iter()
+        .map(|(name, task)| Task {
+            name,
+            stacksize: task
+                .stacksize
+                .unwrap_or_else(|| toml.stacksize.unwrap()),
+            requires: &task.requires,
+        })
+        .collect();
+    tasks.push(Task {
+        name: "kernel",
+        stacksize: toml.stacksize.unwrap(),
+        requires: &toml.kernel.requires,
+    });
+
+    let suggest = if toml.target.starts_with("thumbv7m")
+        || toml.target.starts_with("thumbv7em")
+    {
+        armv7m_suggest
+    } else if toml.target.starts_with("thumbv8m") {
+        armv8m_suggest
+    } else {
+        panic!("Unknown target {}", toml.target);
+    };
+
+    let mut suggestions = Vec::new();
+    for task in tasks {
+        if !first {
+            first = true;
+        } else {
+            println!();
+        }
+        let mut elf_name = dist_dir.clone();
+        elf_name.push(task.name);
+        let buffer = std::fs::read(elf_name)?;
+        let elf = match Object::parse(&buffer)? {
+            Object::Elf(elf) => elf,
+            o => bail!("Invalid Object {:?}", o),
+        };
+
+        let mut memory_sizes = memories
+            .keys()
+            .map(|name| (name.as_str(), 0))
+            .collect::<IndexMap<_, _>>();
+        for phdr in &elf.program_headers {
+            if let Some(region) = output_region(phdr.p_vaddr) {
+                memory_sizes[region] += phdr.p_memsz;
+            }
+        }
+        memory_sizes["ram"] += task.stacksize as u64;
+
+        println!("{}", task.name);
+        let mut my_suggestions = Vec::new();
+        for name in memories.keys() {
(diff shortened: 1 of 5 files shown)


build/xtask/src/sizes.rs
```diff
@@ -0,0 +1,158 @@
...
+use std::convert::TryInto;
+use std::path::{Path, PathBuf};
+
+use anyhow::bail;
+use goblin::Object;
+use indexmap::IndexMap;
+
+use crate::Config;
+
+fn armv7m_suggest(size: u64) -> u64 {
```
cbiffle, line 14: FYI - https://doc.rust-lang.org/std/primitive.u64.html#method.next_power_of_two

build/xtask/src/sizes.rs
```diff
@@ -0,0 +1,158 @@
...
+            requires: &task.requires,
+        })
+        .collect();
+    tasks.push(Task {
+        name: "kernel",
+        stacksize: toml.stacksize.unwrap(),
+        requires: &toml.kernel.requires,
+    });
+
+    let suggest = if toml.target.starts_with("thumbv7m")
```
cbiffle, line 71: `thumbv7` is probably enough, there aren't going to be targets that start with `thumbv7` and don't have this restriction.

build/xtask/src/sizes.rs
```diff
@@ -0,0 +1,158 @@
...
+    {
+        armv7m_suggest
+    } else if toml.target.starts_with("thumbv8m") {
+        armv8m_suggest
+    } else {
+        panic!("Unknown target {}", toml.target);
+    };
+
+    let mut suggestions = Vec::new();
+    let mut first = false; // used to print a newline between tasks
```
cbiffle, line 82: I believe your use of this variable is correct, but its name is inverted. It gets set on all _but_ first.

build/xtask/src/sizes.rs
```diff
@@ -0,0 +1,158 @@
...
+            o => bail!("Invalid Object {:?}", o),
+        };
+
+        let mut memory_sizes = memories
+            .keys()
+            .map(|name| (name.as_str(), 0))
+            .collect::<IndexMap<_, _>>();
+        for phdr in &elf.program_headers {
+            if let Some(region) = output_region(phdr.p_vaddr) {
+                memory_sizes[region] += phdr.p_memsz;
```
cbiffle, line 103: Unless I missed a compensating routine somewhere else, this isn't quite right. Segments loaded into Flash need to have their `filsz` attributed to flash and their `memsz` attributed to RAM. You can spot this case when the vaddr and paddr differ -- the filsz should be assigned to the region associated with the paddr, while the memsz with the vaddr.

build/xtask/src/sizes.rs
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+        memory_sizes["ram"] += task.stacksize as u64;
+
+        println!("{}", task.name);
+        let mut my_suggestions = Vec::new();
+        for name in memories.keys() {
+            let used = memory_sizes[name.as_str()];
```
cbiffle, line 111: `for (name, used) in &memory_sizes`

build/xtask/src/sizes.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+        let mut elf_name = dist_dir.clone();
+        elf_name.push(task.name);
+        let buffer = std::fs::read(elf_name)?;
+        let elf = match Object::parse(&buffer)? {
+            Object::Elf(elf) => elf,
+            o => bail!("Invalid Object {:?}", o),
+        };
+
+        let mut memory_sizes = memories
```
cbiffle, line 97: You might not prefer this, but an alternative to initializing the map in this way is to leave it initially empty:
```rust
let mut memory_sizes = IndexMap::<&str, u64>::new();
```
...and then update it using the `entry` API.
```rust
*memory_sizes.entry(region).or_default() += phdr.p_memsz;
```

build/xtask/src/sizes.rs
```diff
@@ -0,0 +1,158 @@
...
+        for (name, list) in suggestions {
+            println!("{}:", name);
+            for (mem, prev, value) in list {
+                println!(
+                    "  {:<6} {} (from {})",
+                    format!("{}:", mem),
+                    value,
+                    prev
+                );
+                savings[mem.as_str()] += prev as u64 - value;
```
cbiffle, line 148: (You might consider the entry API here too)

cbiffle (commented):
I believe the algorithm on line 103 is wrong, and I've left a comment there. Otherwise this is all just suggestions/FYIs.

build/xtask/src/sizes.rs
mkeeter: ✅

build/xtask/src/sizes.rs
mkeeter: Gotcha – I was matching the behavior [here](https://github.com/oxidecomputer/hubris/blob/master/build/util/src/lib.rs#L17), but have simplified mine to `thumbv7`.

build/xtask/src/sizes.rs
mkeeter: Fixed (by removing it as part of other refactoring)

build/xtask/src/sizes.rs
mkeeter: ✅

build/xtask/src/sizes.rs
mkeeter: This ends up being slightly shorter, so I did it.  It also looks like there's a [`defaultmap` crate](https://docs.rs/defaultmap/0.5.0/defaultmap/), but probably not worth pulling in a new dependency...

build/xtask/src/sizes.rs
mkeeter: Done!

mkeeter:
I think this is good to go!  I added a `*` to the kernel's report, since otherwise it will suggest reducing its RAM allocation and then fail at runtime (until #393 is merged).

build/xtask/src/sizes.rs
```diff
@@ -0,0 +1,208 @@
...
+                        } else {
+                            Color::Red
+                        }));
+                        out.set_color(&color)?;
+                        write!(out, " ({}%)", percent,)?;
+                        out.reset()?;
+                        writeln!(out, "{}", if asterisk { "*" } else { "" })?;
+                    }
+                    let suggestion = suggest(used);
+                    if suggestion != size as u64 && !asterisk {
```
cbiffle, line 133: Change this to `<`, or change the subtract at the end of the routine (line 195) -- otherwise if the region is _smaller_ than the script thinks it should be, it crashes unceremoniously.

cbiffle (approved):
Looks good! I have used it successfully with one change (below).

Issue #395, opened 2022-01-24
Cortex-M0+ port


cbiffle:
It would be nice to have Cortex-M0+ support for situations in the future where we need a low-cost micro.

@jperkin has a prototype port.

We should talk. :-)

cbiffle:
@jperkin thanks for doing this! It's neat.

My initial unsolicited thoughts:

The port looks very plausible. I would like to split the syscall stubs out of `userlib/src/lib.rs` now that there are two architectures -- perhaps into `src/armv6m.rs` and `src/armv6m.rs` and pull them in based on build settings. (This and the RISC-V port also cause me to want to generate the syscall stubs automatically, because there's a lot of duplication going on -- but that's probably a separate issue.)

The branch currently doesn't build. The commit it references in the stm32-nightlies repo has disappeared. The stm32 folks (for reasons I can't explain) don't guarantee that commits _on master_ that repo today will still exist tomorrow -- this is part of why we had to fork it. However, the stm32g070 demo at least _does_ build with the published 0.14 version of stm32g0, so I was able to inspect the results.

I expected that armv6m code would be larger, in general, than armv7m code because of the very limited instruction set, particularly addressing modes. I started by looking at the largest single hunk of code, which is the kernel. Comparison is proving involved, though, because of configuration differences -- since the stm32g0 parts don't implement ITM, and you're leaving them in their default HSI16 clock configuration, the code that actually gets compiled into the kernel is rather different for reasons that have nothing to do with the instruction set. (Also, so that I don't have to do this math again: the stm32h7's vector table is 476 bytes larger than the stm32g0's.)

A more apples-to-apples comparison is the `user-leds` task.

```
   text    data     bss     dec     hex filename
   6068       0       0    6068    17b4 target/demo-stm32g070-nucleo/dist/user_leds
   5664       0       0    5664    1620 target/demo-stm32h743-nucleo/dist/user_leds
```

That's a 7% difference, and the g070-nucleo's user-leds implementation is simpler than the h743-nucleo's because their LED counts differ. Based on past experience and making numbers up, I'd expect a ~10% increase in code size, which means our nearly-32-kiB kernel is going to start becoming a problem on the smaller m0s.

[...]

PR #401, opened 2022-01-26
Initial cortex-m0+ port (mostly stm32g0)

cbiffle:
This builds on @jperkin's port and my notes in #395, applying some optimizations and fixes to support more and smaller targets, and rebasing it to take our recent system changes into account.

~~Note that this is currently contingent on the `static-tcbs` PR #393. This is because the changes in that PR make it easier to conserve kernel memory, which is important on devices with down to 8 kiB of RAM.~~ #393 was merged a while back.

Outstanding issues:
- [x] figure out why Humility doesn't work reliably
- [x] change fault mapping to not say "Illegal Instruction"
- [x] don't check in anything that uses semihosting by default because it makes images hard to test
- [x] test DebugBI support like Humility-Hiffy
- [x] get test suite working


drv/stm32g0-rcc-api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+//
+// A few macros for purposes of defining the Peripheral enum in terms that our
+// driver is expecting:
+//
+// - RCC_IOPENR[31:0] and RCC_IOPRSTR[31:0] are indices 31-0.
+// - RCC_AHBENR[31:0] and RCC_AHBRSTR[31:0] are indices 63-32.
+// - RCC_APBENR1[31:0] and RCC_APBRSTR1[31:0] are indices 95-64.
+// - RCC_APBENR2[31:0] and RCC_APBRSTR2[31:0] are indices 127-96.
+//
+macro_rules! iop {
```
mkeeter, line 93: What about using a `const fn` here instead of a macro?

(it looks like this is also applicable to the `stm32h7-rcc-api`!)

drv/stm32g0-rcc/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+    // Safety: this is needlessly unsafe in the API. The RCC is essentially a
+    // static, and we access it through a & reference so aliasing is not a
+    // concern. Were it literally a static, we could just reference it.
+    let rcc = unsafe { &*device::RCC::ptr() };
+
+    // Any global setup we required would go here.
+
+    // Field messages.
+    // Ensure our buffer is aligned properly for a u32 by declaring it as one.
+    let mut buffer = [0u8; idl::INCOMING_SIZE];
```
mkeeter, line 132: ...this doesn't look like a `u32`, contrary to the comment above

(same issue exists in the H7 RCC task)

lib/armv6m-atomic-hack/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+//! including it on armv7m or later won't pull in the bogus implementations.
+//!
+//! # Why not just disable interrupts
+//!
+//! Because (1) we can't from unprivileged mode and (2) we don't out of
+//! principle anyway.
+
+#![no_std]
+
+use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
```
mkeeter, line 32: You probably want `#[cfg(armv6m)]` here too so that Github Actions doesn't get _really mad_ (or wrap the whole file in that conditional?)

lib/armv6m-atomic-hack/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[cfg(armv6m)]
+fn rmw_ordering(o: Ordering) -> (Ordering, Ordering) {
+    match o {
+        Ordering::AcqRel => (Ordering::Acquire, Ordering::Release),
+        Ordering::Relaxed => (o, o),
+        Ordering::SeqCst => (o, o),
+        Ordering::Acquire => (Ordering::Acquire, Ordering::Relaxed),
+        Ordering::Release => (Ordering::Relaxed, Ordering::Release),
+        _ => panic!(),
```
mkeeter, line 92: It looks like the `match` covers all options – why do we need a default case here?

sys/kern/src/arch/arm_m.rs
```diff
@@ -1151,6 +1457,54 @@ bitflags::bitflags! {
...
+    }
+
+    // This is clearly bogus, but until we have the ability to figure out from a single HardFault
+    // what the actual problem was we need to just pick one...
+    let fault = FaultInfo::IllegalInstruction;
+
+    // We are now going to force a fault on our current task and directly
+    // switch to a task to run.  (It may be tempting to use PendSV here,
+    // but that won't work on ARMv8-M in the presence of MPU faults on
+    // PSP:  even with PendSV pending, ARMv8-M will generate a MUNSTKERR
```
mkeeter, line 1483: This part of the comment isn't relevant in an explicitly `armv6` block

sys/userlib/src/macros.rs
```diff
@@ -33,7 +33,18 @@ macro_rules! sys_log {
...
 
-#[cfg(not(any(feature = "log-semihosting", feature = "log-itm")))]
+#[cfg(feature = "log-null")]
+#[macro_export]
+macro_rules! sys_log {
+    ($s:expr) => {};
+    ($s:expr, $($tt:tt)*) => {};
+}
+
+#[cfg(not(any(
```
mkeeter, line 43: YMMV, but whenever I see both positive and negative `#[cfg...` branches, I prefer to replace it with a `cfg_if` block (and then use `else` for the negative case).

drv/stm32g0-rcc-api/src/lib.rs
cbiffle: I believe that, at the time the original code (which was for the f4) was written, const fns were not permitted in this syntactic position (the definition of enum values)? But they definitely are now! Will fix this one. Would you be willing to file a brief issue to ensure that we go get the f4/h7/etc?

drv/stm32g0-rcc/src/main.rs
cbiffle: That commit is obsolete -- I should have removed it when I did the stm32h7 rcc Idol conversion, and it looks like it got copied into the stm32g0 version. Will fix.

lib/armv6m-atomic-hack/src/lib.rs
cbiffle: Ha, good point. Wrapping the whole file in the conditional made things unhappy, iirc.

lib/armv6m-atomic-hack/src/lib.rs
cbiffle: Because of the `#[non_exhaustive]` attribute on https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html - that's the authors reserving the right to add new variants in the future.

sys/userlib/src/macros.rs
cbiffle: Heh, amusingly, I did this for a _lot_ of cases in this changeset, but not this one. Fixing.

task/pong/Cargo.toml
```diff
@@ -3,11 +3,12 @@ name = "task-pong"
 version = "0.1.0"
 edition = "2018"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+panic-messages = ["userlib/panic-messages"]
```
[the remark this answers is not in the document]
cbiffle: It does change that, yes. We could edit all callers; I haven't gone ahead and done that because... well, okay, the most proximal reason is likely laziness, but my thought process was

- pong doesn't really panic.
- pong is a dev demo that isn't in production.

I can be talked out of this.

drv/stm32g0-rcc-api/src/lib.rs
mkeeter: Done, opened #404 [PR NOT FOUND]

lib/armv6m-atomic-hack/src/lib.rs
mkeeter: Aha, makes sense!

task/pong/Cargo.toml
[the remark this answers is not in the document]
mkeeter: Alright, I don't feel strongly about it!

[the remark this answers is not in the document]
cbiffle: Frankly we kind of need better app-level controls for this sort of thing, so that making something like this configurable doesn't require running around adding a bazillion `features = ["panic-messages"]` to every app.toml. I'm not sure what that looks like yet though. Very open to suggestions.

cbiffle:
FWIW as of last night, most of Humility (including Hiffy-driven commands like gpio) is working on m0+. I think we're just waiting on stack traces, which is in progress.

cbiffle:
- Rebased atop recent master
- Got the kernel test suite passing (which required adding semihosting support to it)
- Fixed one asm mistake that revealed, in the `reply_fault` syscall stub.

cbiffle:
- Updated STM32G031 port to actually work on the Nucleo G031 board.
- Fixed an xtask bug along the way
- Noticed that the cortex-m0 port was acting like exception frames contained floating point registers; fixed this for a size reduction.

cbiffle:
Okay, you can now replicate my hiffy demo from Friday using this branch if you're so inclined. I'm hoping to get this merged soon.

mkeeter (approved):

PR #402, opened 2022-01-26
STM32H7: more faster SPI

### LINKED ISSUE #387: STM32H7 SPI overrun handling isn't great
If you increase the STM32H7 SPI rates high enough, you get overrun errors (which is good if annoying); but there's an intermediate setting on the Nucleo board where you _don't,_ and you instead get the SPI driver hanging forever.

This is because we don't detect overruns on every code path through the loop, and in particular, we can wind up in the EOT-polling loop at the end of the transaction with the OVR bit set. This means the interrupt recurs forever, because we're not clearing it.

The easiest way for this to happen is to observe OVR sometime well before EOT, because the section that observes OVR doesn't currently _clear it._ So you'll wind up getting trapped in the polling loop.

(My plan is to rewrite this event state machine completely, as I think it's nudged over my complexity tolerance -- but I'm filing this in case anyone gets to it first.)

cbiffle:
Fixes #387.

### DIFF
--- a/drv/stm32h7-spi-server/src/main.rs
+++ b/drv/stm32h7-spi-server/src/main.rs
@@ -17,7 +17,9 @@
 #![no_main]
 
 use drv_spi_api::*;
-use idol_runtime::{Leased, LenLimit, RequestError, R, W};
+use idol_runtime::{
+    LeaseBufReader, LeaseBufWriter, Leased, LenLimit, RequestError, R, W,
+};
 use ringbuf::*;
 
 #[cfg(feature = "h7b3")]
@@ -41,8 +43,8 @@ task_slot!(GPIO, gpio_driver);
 #[derive(Copy, Clone, PartialEq)]
 enum Trace {
     Start(SpiOperation, (u16, u16)),
-    Tx(u16, u8),
-    Rx(u16, u8),
+    Tx(u8),
+    Rx(u8),
     WaitISR(u32),
     None,
 }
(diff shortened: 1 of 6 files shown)


drv/stm32h7-spi-server/src/main.rs
```diff
@@ -360,79 +364,85 @@ impl ServerImpl {
...
-                    if *tx_pos == overall_len {
-                        // We continue to transmit beyond our src_len
-                        // (transmitting dummy bytes, above), but once we
-                        // hit overall_len we are done transmitting.
-                        tx = None;
-                        break;
-                    }
+                }
+                if tx_count == overall_len {
+                    // Optimization: stop feeding the FIFO and don't repeat
```
mkeeter, line 412: It would be helpful to add a comment that disabling CAN interrupts _also_ protects us from an IRQ storm in the weird case where the hardware gets stuck (until we find the actual root cause there).

mkeeter (approved):
As of this morning, this works with the `vsc7448` branch with SPI2 running at `DIV16`, both when bringing up the VSC7448 and doing miscellaneous one-byte writes directly through Humility!

task/net/src/server.rs
```diff
@@ -56,7 +56,7 @@ impl<'a> ServerImpl<'a> {
         self.socket_handles
             .get(index)
             .cloned()
-            .ok_or(RequestError::Fail(ClientError::BadMessage))
+            .ok_or(RequestError::Fail(ClientError::BadMessageContents))
```
mkeeter, line 59: Is this change intentionally in this PR?

cbiffle: It is! It's a side effect of the idol-runtime bump. This was the only place in the code that used that enum variant.

drv/stm32h7-spi-server/src/main.rs
cbiffle: Done!

PR #403, opened 2022-01-27
Add STM32H7 HASH block driver and QSPI API to hash regions of flash.

lzrd:
Add STM32 HASH block driver.
Add QSPI function to call HASH on region and return SHA256 value.
Enable hash and qspi on Gemini and Nucleo 753
  - removed U801, testing not succesful.)
  - SI was not good on dupont-wired QSPI flash on Nucleo.
  Testing on Nucleo shows that the software is probably correct.
  Standard SHA256 test vectors pass.

TODO: Suspend/Resume HASH state if we have use cases for multiple client tasks.
TODO: Use 4x mode instead of 1x mode on QSPI to improve throughput.
TODO: HASH DMA of flash contents if performance requirements dictate its use.
TODO: Use HMAC from HASH block if there are use cases.
TODO: Resolve ambiguity about when NBLW must be set, either at the
      begining of a hash data block or before the last word of a block is fed
      to the hash DIN register.
TODO: Some performance gain might be realized by taking advantage of
      already aligned data when feeding the HASH DIN register. However, if
      that is an issue, DMA may become worth the complexity.

### DIFF
--- a/drv/gimlet-hf-server/src/main.rs
+++ b/drv/gimlet-hf-server/src/main.rs
@@ -16,17 +16,24 @@ use drv_stm32h7_qspi::Qspi;
 use drv_stm32xx_sys_api as sys_api;
 use idol_runtime::{ClientError, Leased, LenLimit, RequestError, R, W};
 
-// Note: h7b3 has QUADSPI but has not been used in this project.
-
 #[cfg(feature = "h743")]
 use stm32h7::stm32h743 as device;
 
 #[cfg(feature = "h753")]
 use stm32h7::stm32h753 as device;
 
+// hash_api is optional, but idl files don't have support for optional APIs.
+// So, always include and return a "not implemented" error if the
+// feature is absent.
+#[cfg(feature = "hash")]
+use drv_hash_api as hash_api;
+use drv_hash_api::SHA256_SZ;
+
 use drv_gimlet_hf_api::{HfError, HfMuxState};
 
 task_slot!(SYS, sys);
+#[cfg(feature = "hash")]
+task_slot!(HASH, hash_driver);
 
 const QSPI_IRQ: u32 = 1;
 
@@ -42,11 +49,11 @@ fn main() -> ! {
     // Board specific goo
     cfg_if::cfg_if! {
         if #[cfg(target_board = "gimlet-1")] {
+            let clock = 5; // 200MHz kernel / 5 = 40MHz clock
             qspi.configure(
-                5, // 200MHz kernel / 5 = 40MHz clock
+                clock,
                 25, // 2**25 = 32MiB = 256Mib
             );
-
             // Gimlet pin mapping
             // PF6 SP_QSPI1_IO3
             // PF7 SP_QSPI1_IO2
@@ -94,8 +101,9 @@ fn main() -> ! {
             let select_pin = sys_api::Port::B.pin(1);
             let reset_pin = sys_api::Port::B.pin(2);
         } else if #[cfg(target_board = "gimletlet-2")] {
+            let clock = 5; // 200MHz kernel / 5 = 40MHz clock
             qspi.configure(
-                5, // 200MHz kernel / 5 = 40MHz clock
+                clock,
                 25, // 2**25 = 32MiB = 256Mib
             );
             // Gimletlet pin mapping
@@ -145,7 +153,6 @@ fn main() -> ! {
 
             let select_pin = sys_api::Port::F.pin(5);
             let reset_pin = sys_api::Port::F.pin(4);
-
         } else if #[cfg(target_board = "gemini-bu-1")] {
             // PF4 HOST_ACCESS
             // PF5 RESET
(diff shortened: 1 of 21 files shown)


drv/gimlet-hf-server/src/main.rs
```diff
@@ -420,6 +423,54 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
             }
         }
     }
+
+    cfg_if::cfg_if! {
```
andrewjstone, line 427: I see the macro form is needed because of the IDL generation. It would be cool if instead we could configure the generation to not include the hash functionality and use `#[cfg(feature = "hash")]` to eliminate the runtime error. 

Nothing to do now, just some ideas for the future.

drv/stm32h7-hash-server/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+
+task_slot!(SYS, sys);
+
+const IRQ_MASK: u32 = 1;
+
+fn hash_hw_reset() {
+    let sys = sys_api::Sys::from(SYS.get_task_id());
+    sys.enter_reset(sys_api::Peripheral::Hash);
+    sys.disable_clock(sys_api::Peripheral::Hash);
+    hl::sleep_for(1);
```
andrewjstone, line 33: Maybe this is common in embedded code, but using sleeps feels icky to me. Will `sleep_for` always deterministically satisfy the purpose it is being used for here?

drv/stm32h7-hash-server/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+        Ok(sha256_sum)
+    }
+
+    fn digest_sha256(
+        &mut self,
+        _: &RecvMessage,
+        len: u32,
+        data: LenLimit<Leased<R, [u8]>, 512>,
+    ) -> Result<[u8; 32], RequestError<HashError>> {
+        let mut sha256_sum = [0; 32];
```
andrewjstone, line 104: Nit: You use `[0; 256 / 8]` above. I would just be consistent and choose one way to do it. Maybe use a const.

drv/stm32h7-hash/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+
+// Other SKUs in the STM32 line having the HASH block:
+// stm32{f21[57],f4{05,07,27,29,69},f7{45,65,x6,x7,x9},h7{47cm[47],53{,v},b3},l4x6}
+#[cfg(feature = "h753")]
+use stm32h7::stm32h753 as device;
+
+use userlib::*;
+use zerocopy::AsBytes;
+
+/// TODO: For the purpose of performance analysis, the hash driver could be
```
andrewjstone, line 19: The location of this comment makes it look like it's commenting `use core::mem::size_of`. I would move it up top to the module comment `//!`.

drv/stm32h7-hash/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+        // From the STM32H7 reference:
+        //
+        //  "...message string “abc” with a bit string representation of
+        //  “01100001 01100010 01100011” is represented by a 32-bit word
+        //  0x00636261, and 8-bit words 0x61626300."
+
+        // Deal with the remainder bytes from last update if any.
+        let mut offset = 0;
+        if self.nvalid > 0 {
+            while self.nvalid < 32 {
```
andrewjstone, line 220: May want to put an `assert!(self.nvalid <= 32);` here.

drv/stm32h7-hash/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+            if self.nvalid == 32 {
+                self.write_word(self.remainder, SIZEOF_U32);
+                self.nvalid = 0;
+                self.remainder = 0;
+            }
+        }
+
+        // Hash all of the whole words available.
+        // The words might not be aligned.
+        // TODO: How do the cool kids translate 4 unaligned bytes into a u32?
```
andrewjstone, line 237: You can try using the hideous strategy for slices [here](https://doc.rust-lang.org/std/primitive.u32.html#method.from_le_bytes). But this code looks fine to me.

drv/stm32h7-hash-server/src/main.rs
lzrd: There is already a constant in the hash-api SHA256_SZ, Changed to use that consistently.
I'd like to be able to refer to that constant in the idol file too.

drv/stm32h7-hash/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+                u32::from_be(self.reg.hash_hr0.read().bits()),
+                u32::from_be(self.reg.hash_hr1.read().bits()),
+                u32::from_be(self.reg.hash_hr2.read().bits()),
+                u32::from_be(self.reg.hash_hr3.read().bits()),
+                u32::from_be(self.reg.hash_hr4.read().bits()),
+                u32::from_be(self.reg.hash_hr5.read().bits()),
+                u32::from_be(self.reg.hash_hr6.read().bits()),
+                u32::from_be(self.reg.hash_hr7.read().bits()),
+            ],
+        };
```
labbott, line 336: This seems to be the only place the `Sha256` type is used, do you still need it?

lzrd: I no longer need it. I would like to iterate over the register set directly into the destination, but the Sha256 struct can go.

PR #405, opened 2022-01-27
SPI server needs more memory in lpc55 app-a0.toml

flihp:


labbott:
Oh no do you have 0A hardware

flihp:
No but your old RNG patch used this app toml. Since then it looks these bits have rotted a bit.

cbiffle (approved):

PR #406, opened 2022-01-27
Image header

labbott:
Add an image header for Hubris image for bootloader work


stage0/src/main.rs
```diff
@@ -50,6 +52,92 @@ const ROM_VER: u32 = 0;
...
+            core::ptr::write_volatile(sau_rbar, r.rbar);
+            core::ptr::write_volatile(sau_rlar, r.rlar);
+        }
+    }
+
+    core::ptr::write_volatile(sau_ctrl, 1);
+
+    let mut peripherals = match Peripherals::take() {
+        Some(p) => p,
+        None => loop {},
```
mkeeter, line 74: Is `loop {}` going to get compiled to something odd?  Cliff said at one point

> (`loop {}` for better or worse becomes a panic when LLVM finds it)

and suggested putting a `cortex_m::asm::nop()` in there

cbiffle: What about `panic!`? The `panic-halt` crate ought to work in this context.

stage0/src/image_header.rs
```diff
@@ -132,25 +42,41 @@ impl ImageHeader {
...
 
-    fn get_cert_table(&self) -> *const CertHeader {
-        self.get_table_start() as *const CertHeader
+    pub fn get_vectors(&self) -> u32 {
+        self.0.vector
     }
 
     pub fn get_pc(&self) -> u32 {
-        self.pc
+        let vec = self.get_vectors();
```
cbiffle, line 64: `get_pc` needs to be `unsafe` now, because it can only be used safely from a context where you know the ImageHeader is intended to be valid. (As opposed to declaring a nonsense one on the stack and then using this to dereference nonsense pointers from apparently safe code.)

The `unsafe` should propagate up to somewhere that we're confident the header is as valid as we can achieve.

In practice it looks like `get_pc` is only called from `unsafe` code that is already operating under this requirement, so, should be an easy change.

stage0/src/image_header.rs
```diff
@@ -132,25 +42,41 @@ impl ImageHeader {
...
+
+        unsafe { core::ptr::read_volatile((vec + 0x4) as *const u32) }
     }
 
     pub fn get_sp(&self) -> u32 {
-        self.sp
+        let vec = self.get_vectors();
```
cbiffle, line 70: (same)

stage0/src/image_header.rs
```diff
@@ -132,25 +42,41 @@ impl ImageHeader {
...
+
+        unsafe { core::ptr::read_volatile(vec as *const u32) }
+    }
+
+    pub fn get_sau_entry<'a>(&self, i: usize) -> Option<&'a abi::SAUEntry> {
+        if i > 8 {
+            return None;
+        }
+
+        Some(&self.0.sau_entries[i])
```
cbiffle, line 80: This routine is equivalent to `self.0.sau_entries.get(i)`, fwiw

sys/abi/src/lib.rs
```diff
@@ -546,3 +548,44 @@ impl core::convert::TryFrom<u32> for Sysnum {
...
+
+#[repr(C)]
+#[derive(Default, Copy, Clone, Debug)]
+#[cfg_attr(feature = "std", derive(PackedStruct))]
+#[cfg_attr(
+    feature = "std",
+    packed_struct(size_bytes = "0x8", bit_numbering = "msb0", endian = "msb")
+)]
+pub struct SAUEntry {
+    #[cfg_attr(feature = "std", packed_field(endian = "lsb"))]
```
cbiffle, line 560: Is this all trying to get us accesses using little endian from host software? Should we just use `zerocopy::U32<LittleEndian>` here and avoid the packed struct dependency?

sys/abi/src/lib.rs
```diff
@@ -546,3 +548,44 @@ impl core::convert::TryFrom<u32> for Sysnum {
...
+    pub vector: u32,
+    #[cfg_attr(feature = "std", packed_field(endian = "lsb"))]
+    pub total_image_len: u32,
+    #[cfg_attr(
+        feature = "std",
+        packed_field(endian = "lsb", element_size_bytes = "0x8")
+    )]
+    pub sau_entries: [SAUEntry; 8],
+    // pad for VTOR alignment (128)
```
cbiffle, line 589: I'd suggest letting the linker script manage the padding, particularly because different implementations have different VTOR alignment requirements.

stage0/src/image_header.rs
```diff
@@ -2,124 +2,34 @@
...
-    key_size: u32,
-    // The u32 here represents the start of the key which comes after this
-    // structure.
-    key: u32,
-}
+pub struct Image(&'static ImageHeader);
 
-pub fn get_image_a() -> Option<&'static ImageHeader> {
+pub fn get_image_a() -> Option<Image> {
     // Taking the reference to our supposed imagea
```
cbiffle, line 14: Here's a better safety comment:

```rust
// Safety: this is unsafe because `IMAGEA` is coming from
// an extern, and might violate alignment rules or even be
// modified externally and subject to data races. In our case
// we have to assume that neither of these is true, since it's
// being furnished by our linker script, which we trust.
```

sys/abi/src/lib.rs
labbott: `packed_struct` has been my preferred crate to guarantee structure layouts. I thought the last time this came up there were some questions if `zerocopy` would guarantee the appropriate structure layout vs just guarantee safety. Going `AsBytes` -> `FromBytes` should always work correctly but it wasn't clear if you used `AsBytes` if the representation was guaranteed to be in a form that could be interpreted without using `FromBytes`. Some of this memory may have been muddled from discussion about using `packed_struct` for actual bitfields. Since we (thankfully) have no need for bitfields here I would be okay with switching to zerocopy instead.

build/xtask/src/dist.rs
```diff
@@ -640,6 +662,60 @@ pub fn package(
...
+
+    // Our peripherals
+    header.sau_entries[2].rbar = 0x4000_0000;
+    header.sau_entries[2].rlar = 0x4fff_ffe0 | 1;
+
+    header
+        .write_to_prefix(&mut bytes[header_byte_offset..])
+        .unwrap();
+
+    let mut out = std::fs::OpenOptions::new()
```
cbiffle, line 708: FWIW, this looks like it's equivalent to the options implied by `File::create`?

build/kernel-link.x
```diff
@@ -0,0 +1,262 @@
...
+
+    /* Exceptions */
+    KEEP(*(.vector_table.exceptions)); /* this is the `__EXCEPTIONS` symbol */
+    __eexceptions = .;
+
+    /* Device specific interrupts */
+    KEEP(*(.vector_table.interrupts)); /* this is the `__INTERRUPTS` symbol */
+  } > FLASH
+
+  __vector_size = SIZEOF(.vector_table);
```
cbiffle, line 65: `__vector_size` is effectively a zero-length item, which means when you extern it, you should type it as `()` instead of `u32`. Otherwise it will alias whatever comes next in memory, which is bad.

stage0/src/image_header.rs
```diff
@@ -2,155 +2,106 @@
...
     }
 
-    fn get_table_start(&self) -> u32 {
-        self.get_img_start() + self.header_offset
+    fn get_header(&self) -> *const ImageHeader {
+        // SAFETY: This generated by the linker script which we trust
+        // Note that this is generated from _this_ image's linker script
+        // as opposed to the _image_ linker script but those two _must_
+        // be the same value!
+        let vector_size = unsafe { core::ptr::addr_of!(__vector_size) as u32 };
```
cbiffle, line 41: Nit: `__vector_end` might be clearer? Though if you change the type to `()` as I noted on the linker script, the chances of confusion from the name seem _very low,_ so, up to you.

stage0/src/image_header.rs
```diff
@@ -2,155 +2,106 @@
...
+        // SAFETY: We've validated the header location is programmed so this
+        // will not trigger a fault. This is generated from our build scripts
+        // which we trust.
+        let header = unsafe { &*header_ptr };
+
         // Next make sure the marked image length is programmed
         let valid = lpc55_romapi::validate_programmed(
             img_start,
-            (self.image_length + 0x1ff) & !(0x1ff),
+            (header.total_image_len + 0x1ff) & !(0x1ff),
```
cbiffle, line 73: Nit: these `0x200` and `0x1ff`s -- are these the programming page size? A named constant might be good.

stage0/src/image_header.rs
```diff
@@ -2,155 +2,106 @@
...
 
-    fn get_cert_table(&self) -> *const CertHeader {
-        self.get_table_start() as *const CertHeader
+    pub fn get_vectors(&self) -> u32 {
+        self.get_img_start()
     }
 
-    pub fn get_pc(&self) -> u32 {
-        self.pc
+    pub unsafe fn get_pc(&self) -> u32 {
```
cbiffle, line 92: So, in an uncharacteristic comment for me, I don't think this needs to be `unsafe`. Here's my argument.

- `Image`'s only field the reference to `ImageHeader`, is private to this module, and you are very careful to only build it with valid information (sourced from the build system and `validate`-d). This means anyone holding an `Image` is holding positive proof that the header can be accessed correctly, I believe.
- Because it's a reference (`&'static ImageHeader`) we're telling the Rust compiler that it's initialized and safe to dereference, so any operation on `self.0` should be ok. (We can do this because of your careful validation above.)

Whatcha think?

stage0/src/image_header.rs
```diff
@@ -2,155 +2,106 @@
...
+        self.0.entry
     }
 
-    pub fn get_sp(&self) -> u32 {
-        self.sp
+    pub unsafe fn get_sp(&self) -> u32 {
```
cbiffle, line 96: (Same question as pc)

stage0/src/main.rs
```diff
@@ -50,44 +53,108 @@ const ROM_VER: u32 = 0;
...
+        if let Some(r) = image.get_sau_entry(i) {
+            core::ptr::write_volatile(sau_rnr, i as u32);
+            core::ptr::write_volatile(sau_rbar, r.rbar);
+            core::ptr::write_volatile(sau_rlar, r.rlar);
+        }
+    }
+
+    core::ptr::write_volatile(sau_ctrl, 1);
+
+    let mut peripherals = match Peripherals::take() {
```
cbiffle, line 73: Consider `Peripherals::steal()`

stage0/src/main.rs
```diff
@@ -50,44 +53,108 @@ const ROM_VER: u32 = 0;
...
+        panic!()
     }
 
-    let imagea = image_header::get_image_a().unwrap();
-
-    let entry_pt = imagea.get_pc();
-    let stack = imagea.get_sp();
-
-    let mut peripherals = Peripherals::take().unwrap();
+    let imagea = match image_header::get_image_a() {
```
cbiffle, line 152: Consider the `unwrap-lite` crate here.

app/lpc55xpresso/app.toml
```diff
@@ -3,7 +3,7 @@ target = "thumbv8m.main-none-eabihf"
 board = "lpcxpresso55s69"
 chip = "../../chips/lpc55.toml"
 stacksize = 1024
-secure = true
+secure = false
```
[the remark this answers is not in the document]
labbott: This is confusing. The feature on this line indicates that _hubris_ isn't secure and the selected feature on 64 indicates that the _bootloader_ is secure. Let me see if I can rework this to something easier to understand and less error prone.

stage0/src/image_header.rs
labbott: This isn't the end of the vectors, `__vector_size` is holding the actual size of the vector table.

build/kernel-link.x
cbiffle: It looks like you decided to go with `allow(improper_ctypes)` which feels right to me, so I'll resolve this.

build/xtask/src/dist.rs
```diff
@@ -462,7 +477,19 @@ pub fn package(
     )?;
 
     if let Some(signing) = toml.signing.get("combined") {
-        do_sign_file(signing, &out, &src_dir, "combined")?;
+        match ksymbol_table.get("HEADER") {
+            None =>  bail!("Didn't find header symbol -- does the image need a placeholder?"),
```
cbiffle, line 481: (I think this error message is going to make someone very happy down the road, nicely done)

cbiffle (approved):

PR #412, opened 2022-01-31
userlib: Reserve static memory for panic output.

cbiffle:
Previously, we tried to use stack to generate the panic message, which
seemed clever at the time, but had an annoying result of turning panics
into stack overflows. We considered an alternative approach that would
use "however much stack is left," but that would cause panic messages to
become unexpectedly truncated, which is annoying for debugging.

This change allocates 128 bytes to receive any panic message. This
causes many tasks to require more RAM -- though in practice their stack
sizes can likely be reduced by a corresponding amount. I haven't done
that in this change because I haven't tested them all under load; stack
sizing is tricky.

Any task not wishing to pay this cost can, of course, opt out of panic
messages by removing the Cargo feature from `userlib`.

I've also taken this opportunity to rewrite the panic handler to be more
robust in the face of compiler changes. Instead of being clever about
letting the compiler prove that code can't panic, I've just written it
using unsafe so that it _literally_ will never panic.


cbiffle:
Note that this PR is currently atop the `cortex-m0` PR, as I'm trying not to snipe my own PRs.

sys/userlib/src/lib.rs
```diff
@@ -1284,71 +1284,161 @@ pub unsafe extern "C" fn _start() -> ! {
...
+            // the compiler, which otherwise can't see our invariant.
+            if self.pos >= self.buf.len() {
+                return Ok(());
             }
+
+            // We can write more. How much more? Let's find out. We're going to
+            // do this using unchecked operations to avoid a recursive panic.
+            //
+            // Safety: this is unsafe because of the risk of passing in an
+            // out-of-bounds index for the split. However, our type invariant
```
steveklabnik, line 1351: is it the "type invariant" or the check that we made above on 1343? the type of pos is usize, which could hold a value bigger than buf.len()

sys/userlib/src/lib.rs
```diff
@@ -1284,71 +1284,161 @@ pub unsafe extern "C" fn _start() -> ! {
...
+    //
+    // However, it is possible to produce an alias if the panic handler is
+    // called reentrantly. This can only happen if the code in the panic handler
+    // itself panics, which is what we're working very hard to prevent here.
+    let panic_buffer = unsafe { &mut PANIC_BUFFER };
+
+    // Whew! Time to write the darn message.
+    //
+    // Note that if we provided a different value of `pos` here we could destroy
+    // PrefixWrite's type invariant, so, don't do that.
```
steveklabnik, line 1422: ah yes so i guess given that there is literal access to the pos, that's why i don't think of it as a type invariant. however, *really* isolating this stuff would make the code more convoluted for little gain, so im not advocating that exactly

steveklabnik (approved):
other than a marginal philosophical question about "what is a type invariant" this unsafe code looks good to me. i also share the "the more anxious i am the more comments there should be" and this feels appropriate and good for what's going on here.

sys/userlib/src/lib.rs
cbiffle: Ah, sry, by "our type invariant" I mean the invariant on PrefixWrite (our type). I could clarify the comments.

sys/userlib/src/lib.rs
cbiffle: Right -- strictly I'd need to move this into a submodule and protect its internals. So far I couldn't be bothered, so, you do have to read the whole function to verify the invariant. Open to better terms!

sys/userlib/src/lib.rs
steveklabnik: nah i think you're totally good to merge, hence the approval.

PR #414, opened 2022-01-31
Ugh cortex-m0 again.

cbiffle:
This is a re-push of the reverted cortex-m0+ support, with fix. I have tested this.


steveklabnik:
Might be nice for future reference to comment on what the fix was.

cbiffle:
Yeah, I tried to reopen the original PR and continue there, but that's apparently not a thing.

In brief: the original third-party port duplicated interrupt handlers like SVCall with #[cfg(armv6m)] vs #[cfg(everyone_else)] attributes. I didn't like that so I combined them and used cfg_if internally to switch based on architecture. ...but I didn't remove the #[cfg(armv6m)] from the original patch in two cases.

As a result armv7m was routing syscalls to the default handler, which just panics. This implies that I had not run the ARMv7-M test suite on my cortex-m0 branch. I have done so now and the fix seems to work.

This manifested as Humility rejecting the archives because, as a result of the syscall handlers getting compiled out, large swaths of the kernel got compiled out, including the image ID header that Humility now uses to ensure that your archive matches what's in flash.

This sort of mistake is exactly why I wanted to take the cfg off the functions and move it to sections of the body -- if you get it wrong, it's like you deleted the code from the compiler's perspective, just like #ifdef. Just, need to do it correctly. :-)

PR #416, opened 2022-02-01
Factor common GPIO code out of stm32{g0,h7}

cbiffle:
This is a step toward reducing the overall high level of code
duplication between stm32 targets. It exploits some traits that the PACs
started generating circa 0.14, which are unfortunately not put into a
common crate, but, hey, at least they exist now.

Knocks some bytes (100+) off GPIO servers by reducing macro-driven code
duplication and giving the compiler an easier task.

This is sort of a HAL, but unlike most HALs it's very careful to allow
static monomorphization on properties you know at compile time (like
your SoC model target) and only use dynamic dispatch for things you vary
at runtime (like which gpio port you're poking).

Scary bits are mostly hidden in the server.rs file.

### DIFF
--- a/drv/stm32g0-sys/src/main.rs
+++ b/drv/stm32g0-sys/src/main.rs
@@ -16,7 +16,8 @@ use stm32g0::stm32g070 as device;
 #[cfg(feature = "g0b1")]
 use stm32g0::stm32g0b1 as device;
 
-use drv_stm32g0_sys_api::{GpioError, Port, RccError};
+use drv_stm32g0_sys_api::{GpioError, RccError};
+use drv_stm32xx_gpio_common::{server::get_gpio_regs, Port};
 use idol_runtime::RequestError;
 use userlib::*;
 
@@ -79,106 +80,6 @@ fn main() -> ! {
     loop {
         idol_runtime::dispatch(&mut buffer, &mut server);
     }
-}
-
-// These macros are required as the different GPIO ports return different types
-// for some reason (GPIOA returns gpioa, GPIO[B-F] return gpiob?!).
-macro_rules! do_configure {
-    ($reg:expr, $pins:expr, $packed_attributes:expr) => {{
-        // The GPIO config registers come in 1, 2, and 4-bit per
-        // field variants. The user-submitted mask is already
-        // correct for the 1-bit fields; we need to expand it
-        // into corresponding 2- and 4-bit masks. We use an
-        // outer perfect shuffle operation for this, which
-        // interleaves zeroes from the top 16 bits into the
-        // bottom 16.
-
-        // 1 in each targeted 1bit field.
-        let mask_1 = u32::from($pins);
-        // 0b01 in each targeted 2bit field.
-        let lsbs_2 = outer_perfect_shuffle(mask_1);
-        // 0b0001 in each targeted 4bit field for low half.
-        let lsbs_4l = outer_perfect_shuffle(lsbs_2 & 0xFFFF);
-        // Same for high half.
-        let lsbs_4h = outer_perfect_shuffle(lsbs_2 >> 16);
-
-        // Corresponding masks, with 1s in all field bits
-        // instead of just the LSB:
-        let mask_2 = lsbs_2.wrapping_mul(0b11);
-        let mask_4l = lsbs_4l.wrapping_mul(0b1111);
-        let mask_4h = lsbs_4h.wrapping_mul(0b1111);
-
-        let atts = $packed_attributes;
-
-        // MODER contains 16x 2-bit fields.
-        let moder_val = u32::from(atts & 0b11);
-        $reg.moder.write(|w| unsafe {
-            w.bits(
-                ($reg.moder.read().bits() & !mask_2)
-                    | (moder_val.wrapping_mul(lsbs_2)),
-            )
-        });
-        // OTYPER contains 16x 1-bit fields.
-        let otyper_val = u32::from((atts >> 2) & 1);
-        $reg.otyper.write(|w| unsafe {
-            w.bits(
-                ($reg.otyper.read().bits() & !mask_1)
-                    | (otyper_val.wrapping_mul(mask_1)),
-            )
-        });
-        // OSPEEDR contains 16x 2-bit fields.
-        let ospeedr_val = u32::from((atts >> 3) & 0b11);
-        $reg.ospeedr.write(|w| unsafe {
-            w.bits(
-                ($reg.ospeedr.read().bits() & !mask_2)
-                    | (ospeedr_val.wrapping_mul(lsbs_2)),
-            )
-        });
-        // PUPDR contains 16x 2-bit fields.
-        let pupdr_val = u32::from((atts >> 5) & 0b11);
-        $reg.pupdr.write(|w| unsafe {
-            w.bits(
-                ($reg.pupdr.read().bits() & !mask_2)
-                    | (pupdr_val.wrapping_mul(lsbs_2)),
-            )
-        });
-        // AFRx contains 8x 4-bit fields.
-        let af_val = u32::from((atts >> 7) & 0b1111);
-        $reg.afrl.write(|w| unsafe {
-            w.bits(
-                ($reg.afrl.read().bits() & !mask_4l)
-                    | (af_val.wrapping_mul(lsbs_4l)),
-            )
-        });
-        $reg.afrh.write(|w| unsafe {
-            w.bits(
-                ($reg.afrh.read().bits() & !mask_4h)
-                    | (af_val.wrapping_mul(lsbs_4h)),
-            )
-        });
-    }};
-}
-
-macro_rules! do_set_reset {
-    ($reg:expr, $set:expr, $reset:expr) => {
-        $reg.bsrr.write(|w| unsafe {
-            w.bits((u32::from($reset) << 16) | u32::from($set))
-        })
-    };
-}
-
-macro_rules! do_toggle {
-    ($reg:expr, $pins:expr) => {{
-        // Read current pin *output* states.
-        let state = $reg.odr.read().bits();
-        // Compute BSRR value to toggle all pins. That is, move
-        // currently set bits into reset position, and set unset
-        // bits.
-        let bsrr_all = state << 16 | state ^ 0xFFFF;
-        // Write that value, but masked as the user requested.
-        let bsrr_mask = u32::from($pins) | u32::from($pins) << 16;
-        $reg.bsrr.write(|w| unsafe { w.bits(bsrr_all & bsrr_mask) });
-    }};
 }
 
 struct ServerImpl<'a> {
@@ -257,24 +158,7 @@ impl idl::InOrderSysImpl for ServerImpl<'_> {
         pins: u16,
         packed_attributes: u16,
     ) -> Result<(), RequestError<GpioError>> {
-        match port {
-            Port::A => {
-                let gpio = unsafe { &*device::GPIOA::ptr() };
-                do_configure!(gpio, pins, packed_attributes);
-            }
-            _ => {
-                let gpio = match port {
-                    Port::B => unsafe { &*device::GPIOB::ptr() },
-                    Port::C => unsafe { &*device::GPIOC::ptr() },
-                    Port::D => unsafe { &*device::GPIOD::ptr() },
-                    #[cfg(feature = "g0b1")]
-                    Port::E => unsafe { &*device::GPIOE::ptr() },
-                    Port::F => unsafe { &*device::GPIOF::ptr() },
-                    _ => unreachable!(),
-                };
-                do_configure!(gpio, pins, packed_attributes);
-            }
-        }
+        unsafe { get_gpio_regs(port) }.configure(pins, packed_attributes);
         Ok(())
     }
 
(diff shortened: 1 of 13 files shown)


drv/stm32xx-gpio-common/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    I,
+    #[cfg(feature = "has-port-gpioj")]
+    J,
+    #[cfg(feature = "has-port-gpiok")]
+    K,
+}
+
+impl Port {
+    /// Turns a `Port` into a `PinSet` containing one pin, number `index`.
+    pub const fn pin(self, index: usize) -> PinSet {
```
mkeeter, line 43: Does `#[inline(always)]` improve code size here?

drv/stm32xx-gpio-common/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+pub struct PinSet {
+    /// Port we're talking about.
+    pub port: Port,
+    /// Mask with 1s in affected positions, 0s in others.
+    pub pin_mask: u16,
+}
+
+impl PinSet {
+    /// Derives a `PinSet` by setting mask bit `index`.
+    pub const fn and_pin(self, index: usize) -> Self {
```
mkeeter, line 67: Same question! (Does `#[inline(always)]` improve code size here?)

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+        let lsbs_2 = outer_perfect_shuffle(mask_1);
+        // 0b0001 in each targeted 4bit field for low half.
+        let lsbs_4l = outer_perfect_shuffle(lsbs_2 & 0xFFFF);
+        // Same for high half.
+        let lsbs_4h = outer_perfect_shuffle(lsbs_2 >> 16);
+
+        // Corresponding masks, with 1s in all field bits instead of just the
+        // LSB.  We use multiplication to distribute a constant over all 1
+        // positions because multiplication is cheap on ARMvx-M devices.
+        let mask_2 = lsbs_2.wrapping_mul(0b11);
```
mkeeter, line 112: Out of curiosity, are we using `wrapping_mul` here to save an overflow check (since overflow shouldn't be possible)?

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+        let mask_4l = lsbs_4l.wrapping_mul(0b1111);
+        let mask_4h = lsbs_4h.wrapping_mul(0b1111);
+
+        // MODER contains 16x 2-bit fields.
+        let moder_val = u32::from(atts & 0b11);
```
mkeeter, line 117: A comment about how `atts` are packed would be helpful.  I was confused by the name `packed_attributes` used elsewhere, and thought that it was supposed to define attributes for multiple pins.

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+        self.moder().modify(|r, w| unsafe {
+            w.bits((r.bits() & !mask_2) | (moder_val.wrapping_mul(lsbs_2)))
+        });
+
+        // OTYPER contains 16x 1-bit fields.
+        let otyper_val = u32::from((atts >> 2) & 1);
+        self.otyper().modify(|r, w| unsafe {
+            w.bits((r.bits() & !mask_1) | (otyper_val.wrapping_mul(mask_1)))
```
mkeeter, line 125: It would be easier to read if we used `lsbs_1` here (with the trivial `let mask_1 = lsbs_1;` in the block above)

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+            w.bits((r.bits() & !mask_4l) | (af_val.wrapping_mul(lsbs_4l)))
+        });
+        self.afrh().modify(|r, w| unsafe {
+            w.bits((r.bits() & !mask_4h) | (af_val.wrapping_mul(lsbs_4h)))
+        });
+    }
+
+    fn set_reset(&self, set: u16, reset: u16) {
+        self.bsrr().write(|w| unsafe {
+            w.bits((u32::from(reset) << 16) | u32::from(set))
```
mkeeter, line 149: General question: why use `u32::from` vs `.into()`?

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+        });
+    }
+
+    fn toggle(&self, pins: u16) {
+        // Read current pin *output* states.
```
mkeeter, line 154: Nit – what about implementing this using `set_reset`, e.g.
```
// Read current pin *output* states, masked by the user's request
let state = self.odr().read().bits() as u16;
self.set_reset((state ^ 0xFFFF) & pins, state & pins);
```

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+/// Adapter from `GpioPeriph`, the trait implemented for disjoint GPIO
+/// peripheral types, to the general `AnyGpioPeriph`.
+#[cfg(feature = "server-support")]
+impl<T: GpioPeriph> AnyGpioPeriph for T {
+    fn configure(&self, pins: u16, atts: u16) {
+        // The GPIO config registers come in 1, 2, and 4-bit per field variants.
+        // The user-submitted mask is already correct for the 1-bit fields; we
+        // need to expand it into corresponding 2- and 4-bit masks. We use an
+        // outer perfect shuffle operation for this, which interleaves zeroes
+        // from the top 16 bits into the bottom 16.
```
mkeeter, line 98: I love the use of perfect shuffling here!

drv/stm32g0-sys/src/main.rs
```diff
@@ -257,24 +158,7 @@ impl idl::InOrderSysImpl for ServerImpl<'_> {
...
-                    Port::D => unsafe { &*device::GPIOD::ptr() },
-                    #[cfg(feature = "g0b1")]
-                    Port::E => unsafe { &*device::GPIOE::ptr() },
-                    Port::F => unsafe { &*device::GPIOF::ptr() },
-                    _ => unreachable!(),
-                };
-                do_configure!(gpio, pins, packed_attributes);
-            }
-        }
+        unsafe { get_gpio_regs(port) }.configure(pins, packed_attributes);
```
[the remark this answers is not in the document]
cbiffle: siiiiiiiiigh

So.

The family of APIs that grew up around svd2rust decided, early on, to try to manage peripheral ownership using little "tokens" that can't be duplicated. You do `Peripherals::take()` at boot and it gives you a struct of tokens, and you can access peripherals through those tokens. This approach is useful in some contexts.

We don't use it, though, because it assumes that all code is in the same address space and is simply unsound in unprivileged isolated mode.

This means we're sort of on the gray areas of the svd2rust-generated APIs. In particular, they decided that any access to peripherals _outside_ of this token scheme would be unsafe. This is an abuse of `unsafe` to mean something other than `unsafe` -- technically there are peripherals one could poke that could violate memory safety, like anything capable of remapping memory or a DMA controller, but they are exceptions and the GPIO peripheral is certainly not among them.

So, a side effect of their decision is that there's no static `GPIOA` item defined, to prevent "ambient access" to the GPIOA peripheral bypassing the token scheme. Instead, all they'll do is give us a raw pointer from the `GPIOA::ptr()` call, which is the only call we can make without holding one of their tokens. We have to indirect it to get at the actual `RegisterBlock`, which is unsafe-in-the-Rust-sense.

Now, in this case, why did I also make `get_gpio_regs` unsafe, rather than hiding the `unsafe` within it? I could go either way on that.

drv/stm32xx-gpio-common/src/lib.rs
cbiffle: I've never seen this routine _not_ get inlined. In practice that's probably because it's almost always called in `const` context though. I can add a directive.

drv/stm32xx-gpio-common/src/lib.rs
cbiffle: (same answer)

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+//! Server support functions, types, etc.
+
+use crate::Port;
+
+// Every PAC includes the same traits and types describing register access. One
+// would expect them to be defined in a common crate that all PACs depend upon,
+// but no, they are generated whole-cloth in every PAC. As a result, we can't
+// simply depend on the common types here, and we need to know which SoC family
+// we're targeting to generate these SoC-family-independent APIs. :-(
+#[cfg(feature = "server-support")]
```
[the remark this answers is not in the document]
cbiffle: Yes, because otherwise you have to specify which precise SoC model you're targeting to build the crate (see comment just above).

drv/stm32xx-gpio-common/src/server.rs
cbiffle: Bingo. Want me to add a comment?

drv/stm32xx-gpio-common/src/server.rs
cbiffle: That's totally fair. Will document.

drv/stm32xx-gpio-common/src/server.rs
cbiffle: Oh, interesting. I used `mask_1` to make it syntactically parallel with the other computations, where using `lsbs` in place of `mask` would be a typo. I could go either way.

drv/stm32xx-gpio-common/src/server.rs
cbiffle: Because it's explicit about the destination type, which makes the code easier for me to read. (I habitually use `T::from` over `.into()` for this reason. This is possibly idiosyncratic of me.)

drv/stm32xx-gpio-common/src/server.rs
cbiffle: Hm. It seems like you'd find that clearer, can you help me understand why?

From my perspective it doesn't reduce the size of the routine, and inserts a conceptual indirection (i.e. when checking this routine you now need to go read another routine), so I didn't do it by default. Open to changing that for clarity though.

drv/stm32xx-gpio-common/src/server.rs
mkeeter: Yes please!

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -0,0 +1,278 @@
...
+                &self.odr
+            }
+            fn idr(&self) -> &pac::Reg<Self::IdSpec> {
+                &self.idr
+            }
+        }
+    };
+}
+
+#[cfg(feature = "has-gpioa-type")]
```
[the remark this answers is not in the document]
cbiffle: Ok. (And, yes, your inference is correct. Though if we ever need to support the stm32l4 series a separate `gpioc` type _will_ appear. svd2rust is a capricious master.)

drv/stm32xx-gpio-common/src/server.rs
cbiffle: Heh, thanks. The Hacker's Delight outer perfect shuffle routine is also _ridiculously_ cheap on ARMv7-M because the constants happen to match patterns that can be materialized cheaply by Thumb-2 instructions. It is _much worse_ on ARMv6-M, since those particular instructions were among the ones removed to simplify the architecture. But, ARMv6-M takes more instructions to do _everything,_ so I'm not terribly concerned.

drv/stm32xx-gpio-common/src/server.rs
mkeeter: Makes sense (and there was follow-up discussion in #oxide-rust)

drv/stm32xx-gpio-common/src/server.rs
mkeeter: I don't feel _too_ strongly about it; the only reason it would be more clear is that abstracts away the inner details of `BSSR`.

drv/stm32xx-gpio-common/src/server.rs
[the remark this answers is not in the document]
cbiffle: Ohhhh I misunderstood your question. I think the answer is no.

drv/stm32g0-sys/src/main.rs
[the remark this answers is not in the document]
mkeeter: Got it, that all makes sense.  I don't feel strongly about this, so matching `svd2rust`'s decisions (however misguided) seems reasonable.

drv/stm32xx-gpio-common/src/server.rs
cbiffle: (I have added a comment)

drv/stm32g0-sys/src/main.rs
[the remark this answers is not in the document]
cbiffle: (This is related to the reason why accessing raw `bits` is `unsafe` on _some_ registers and not others. Basically the default is to make any bit manipulation `unsafe`; someone has to go through and edit a giant YAML file to mark operations as ok to do from safe code. People have not done that for our processors. It gets really fun when they've done it for _some_ of our processors and not others, which gets us "unnecessary unsafe" warnings when compiling for some targets. This is part of why I'm concerned that our use case is diverging too far from what the PAC authors intend.)

drv/stm32xx-gpio-common/src/server.rs
cbiffle: I think I'm coming around to your perspective here, because while I tend to think of the code as "how many hoops must I jump through to ensure that it matches the hardware reference manual," using `set_reset` here would centralize the logic around bsrr bit pattern generation in one place. Which, while technically a hoop, is one less opportunity for divergence. So, I'll make the change.

mkeeter (approved):

drv/stm32xx-gpio-common/src/server.rs
cbiffle: Ohhhhh I read your comment backwards. Yes, I agree that would be clearer.

PR #417, opened 2022-02-03
fix breakage introduced by stack size change

bcantrill:


rmustacc (approved):
Thanks for doing this. Given that we have 2k stack sizes in others restoring this back up seems reasonable. We may want to come back and figure out what the right size is across these again later.

PR #421, opened 2022-02-05
Generalize stm32g0-sys to support stm32h7

cbiffle:
This replaces the stm32h7 rcc and gpio servers with a combined `sys` server. I initially used this approach on the stm32g0 port to save space, but it has turned out to have other advantages:

- The GPIO server only ever sent messages to the RCC server; merging the servers removes the need to handle potential errors / restarts of the two tasks.
- This creates a good place for support for EXTI to live, which is on my todo list. EXTI (the external interrupt controller) is tricky because it kind of cross-cuts GPIO and SYSCFG; we currently don't have a SYSCFG driver, but it's so intimately tied to device setup that it probably belongs in the same place as RCC, which is to say, in the `sys` server.
- This also gives us a centralized place to start implementing access controls without duplicating the information in gpio and rcc.

I've done this by generalizing the stm32g0 code to cover h7, and then renaming it to `stm32xx-sys`. This parallels the `stm32xx-gpio-common` stuff I did last week. There is some g0/h7 specific code in the sys task but it's largely target-neutral code, which is nice.

That in turn has the advantage of making ports to other stm32 devices _much easier;_ I'm rewriting my STM32L0 support in terms of this. Agility across SoC families is turning out to be _kind of important_ here in 2022. :-\

This PR may be easiest to review as individual commits; each commit builds and is fairly focused on a single type of change.


drv/stm32xx-sys/src/main.rs
```diff
@@ -0,0 +1,370 @@
...
+                    .gpiogen()
+                    .set_bit()
+                    .gpiohen()
+                    .set_bit()
+                    .gpioien()
+                    .set_bit()
+                    .gpiojen()
+                    .set_bit()
+                    .gpioken()
+                    .set_bit()
```
bcantrill, line 122: /me shakes fist at rustfmt

task/hiffy/src/stm32g0.rs
```diff
@@ -18,49 +18,49 @@ pub struct Buffer(u8);
 pub enum Functions {
     Sleep(u16, u32),
     #[cfg(feature = "gpio")]
-    GpioInput(drv_stm32g0_sys_api::Port, drv_stm32g0_sys_api::GpioError),
+    GpioInput(drv_stm32xx_sys_api::Port, drv_stm32xx_sys_api::GpioError),
```
bcantrill, line 21: This is a reminder that I really need to retool Humility to be using the IDL calls for `humliity gpio`!

bcantrill (approved):
Looks great -- let me know if you'd like me to test on Gimlet, though the risk here frankly seems very low.

PR #423, opened 2022-02-07
Move chip peripherals into a separate file

mkeeter:
This PR adds a new `chip` field to `app.toml` which specifies a file from which to load peripheral addresses.

It's still backwards-compatible, but I've gone through and changed all of the `app.toml`s that I could lay my hands on.

Having peripheral addresses in one place reduces copy-pasta, and will make it easier to switch to named interrupts consistently through all of our `app.toml`'s.

### DIFF
--- a/build/xtask/src/main.rs
+++ b/build/xtask/src/main.rs
@@ -198,9 +198,8 @@ impl Config {
             if !toml.peripherals.is_empty() {
                 bail!("Cannot specify both chip and peripherals");
             }
-            let mut src_dir = cfg.to_path_buf();
-            src_dir.pop();
-            let chip_contents = std::fs::read(src_dir.join(chip))?;
+            let chip_file = cfg.parent().unwrap().join(chip);
+            let chip_contents = std::fs::read(chip_file)?;
             hasher.write(&chip_contents);
             toml::from_slice(&chip_contents)?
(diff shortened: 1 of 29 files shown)


build/xtask/src/main.rs
```diff
@@ -159,6 +164,71 @@ struct Config {
...
+        hasher.write(&cfg_contents);
+
+        // If the app.toml specifies a `chip` key, then load the peripheral
+        // register map from a separate file and accumulate that file in the
+        // buildhash.
+        let peripherals = if let Some(chip) = toml.chip {
+            if !toml.peripherals.is_empty() {
+                bail!("Cannot specify both chip and peripherals");
+            }
+            let mut src_dir = cfg.to_path_buf();
```
cbiffle, line 201: FYI - `cfg.parent().join(chip)`

cbiffle (approved):

build/xtask/src/main.rs
cbiffle: ... `cfg.parent().unwrap().join(chip)` I think. The `src_dir.pop()` implies the equivalent of that `unwrap` if I'm reading correctly.

mkeeter: Cool!

PR #424, opened 2022-02-07
Switch to using named IRQs whenever possible

mkeeter:
Begone, cryptic numbers!


mkeeter:
This isn't quite ready to merge – we discovered that Humility won't know how to load named IRQs from the separate chip file, so we'll need to make changes there as well.

bcantrill:
FWIW, the Humility changes have been made and verified against this branch. Clear to land!

PR #425, opened 2022-02-07
Fix broken IRQ by using numbers instead of name

mkeeter:
#423 broke `humility manifest` because the IRQ names are no longer part of the `app.toml` file.  This fixes it temporarily, until our bright future of named IRQs arrives (in #424).


bcantrill (approved):
I'll take it!

PR #426, opened 2022-02-08
Test humility manifest

steveklabnik:
This change:

* downloads the latest nightly humility
* makes it executable
* runs humility manifest

It only does this on linux platforms because windows is already very
slow and we only need to test in one place.


bcantrill (approved):
This looks great -- so sorry that I was so slow to review it!

PR #429, opened 2022-02-11
spdm build on stoltz

lzrd:
- WIP - added toy SPDM send, recv, exchange.
- WIP - toy SPDM plumbing


task/spdm/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+
+struct ServerImpl {
+    // _responder: spdm::Responder<'a, FakeSigner>,
+    valid: usize,
+    message: [u8; MAX_SPDM_MSG_SIZE],
+}
+
+impl idl::InOrderSpdmImpl for ServerImpl {
+    /// A client sends a message for SPDM processing.
+    fn send(
```
andrewjstone, line 82: The `send` and `recv` terminology from a server PoV is already confusing me. It makes sense from a client/humility POV where we use humility as a "step based" debugger, but I don't know why a SPDM requester or responder would need to be told to `send`. A requester does need to be told when to send `GET_MEASUREMENTS` perhaps, but that could also be encoded in the SPDM requester code. There will also be some proxy behavior where the SP essentially just forwards signing requests to the RoT. 

I'm thinking we'll likely end up just wanting semantic operations in humility and we may or may not want to map them 1:1 with IDL.

In short, I know I pushed a bit for send/recv, but maybe we should remove them and just use `exchange` for now while we build out the SPDM responder code only here. We can use humility as the requester and once we have a flow worked out we can go back and change the IDL...

I think

PR #431, opened 2022-02-14
Don't run Cargo twice per task

mkeeter:
`cargo_output_dir` always maps to `target`, but takes a significant number of milliseconds per invocation.  This speeds up a no-op build by about 40%.

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -957,17 +957,7 @@ fn build(
         cmd.arg(features.join(","));
     }
 
-    let mut cargo_out = cargo_output_dir(target, path)?;
-
-    // the target dir for each project is set to the same dir, but it ends up
-    // looking something like:
-    //
-    // foo/bar/../target
-    // foo/baz/../target
-    //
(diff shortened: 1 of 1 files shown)


build/xtask/src/dist.rs
```diff
@@ -957,17 +957,7 @@ fn build(
...
-
-    // the target dir for each project is set to the same dir, but it ends up
-    // looking something like:
-    //
-    // foo/bar/../target
-    // foo/baz/../target
-    //
-    // these resolve to the same dirs but since their text changes, this causes
-    // rebuilds. by canonicalizing it, you get foo/target for every one.
-    let canonical_cargo_out_dir = fs::canonicalize(&cargo_out)?;
```
steveklabnik, line 970: it might be nice to have a comment here explaining that we are able to do this because we do not set `CARGO_TARGET` and if that changes, this will need to change.

steveklabnik (approved):
Big fan. Not only of the approach but also to not having to deal with canonicalization. 

One note, not blocking. :)

Issue #430, opened 2022-02-11
LED API should expose a `set(usize, bool)` function


mkeeter:
This would save me from having to write stuff like
```rust
if any_link {
    self.leds.led_on(1).unwrap();
} else {
    self.leds.led_off(1).unwrap();
}
```

PR #432, opened 2022-02-15
user_leds: Add set(usize, bool) function

rumpuslabs:
My attempt at #430 - first time contributing to Hubris.

I wasn't completely sure if this was the right approach but as `bool` doesn't implement `FromPrimitive` or `FromBytes` I modelled the API on spi's `CsState` and added a wrapper function to convert from `bool`.

### DIFF
--- a/drv/user-leds-api/src/lib.rs
+++ b/drv/user-leds-api/src/lib.rs
@@ -22,4 +22,25 @@ impl From<u32> for LedError {
     }
 }
 
+#[derive(
+    Copy, Clone, Debug, Eq, PartialEq, zerocopy::AsBytes, FromPrimitive,
+)]
+#[repr(u8)]
+pub enum LedState {
+    Off = 0,
+    On = 1,
+}
+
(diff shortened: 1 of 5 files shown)


drv/user-leds-api/src/lib.rs
```diff
@@ -22,4 +22,25 @@ impl From<u32> for LedError {
...
+    pub fn led_set(&self, index: usize, state: bool) -> Result<(), LedError> {
+        self.led_set_state(
+            index,
+            match state {
+                true => LedState::On,
+                false => LedState::Off,
+            },
+        )
+    }
+}
```
rumpuslabs, line 46: Of course a much simpler version of this would be simply:
```rust
impl UserLeds {
    pub fn led_set(&self, index: usize, state: bool) -> Result<(), LedError> {
        match state {
            true => self.led_on(index),
            false => self.led_off(index),
        }
    }
}
```
and dispense with all the other changes, perhaps this was what was intended? Seems too simple for the OP to have bothered raising an issue for, though I don't know Hubris well enough to judge what the benefit of the more complex approach is.

PR #433, opened 2022-02-17
Drastically simplify LPC55 SPI driver

labbott:
The LPC55 SPI is designed to be used as a target, responding to what
is sent. This does not align well with the server design that was
written originally. Combine the spi-loopback task into the spi
driver to better match what this driver will eventually do.


drv/lpc55-spi-server/src/main.rs
```diff
@@ -117,172 +79,64 @@ fn main() -> ! {
...
-                    datref.spi.drain();
-                    datref.spi.enable_tx();
-                    datref.spi.enable_rx();
-
-                    ringbuf_entry!(Trace::Start(Op::Exchange));
-                    Ok(())
-                }
-            },
-        );
+        let _m = match sys_recv_closed(&mut [], 1, TaskId::KERNEL) {
```
andrewjstone, line 101: Since you're ignoring the values here you can likely simplify with 

```rust
if sys_recv_closed(&mut [], 1, TaskId::Kernel).is_err() {
    panic!()
}
```

drv/lpc55-spi-server/src/main.rs
```diff
@@ -117,172 +79,64 @@ fn main() -> ! {
...
+            Ok(m) => m,
+            Err(_) => panic!(),
+        };
+
+        ringbuf_entry!(Trace::IRQ);
+
+        while spi.can_tx() && !tx_done {
+            let b = tx[tx_cnt];
+            ringbuf_entry!(Trace::Tx(b));
+            spi.send_u8(tx[tx_cnt]);
```
andrewjstone, line 111: nit: replace `tx[tx_cnt]` with `b`

drv/lpc55-spi-server/src/main.rs
```diff
@@ -117,172 +79,64 @@ fn main() -> ! {
...
+            let b = spi.read_u8();
+            ringbuf_entry!(Trace::Rx(b));
+            rx[rx_cnt] = b;
+            rx_cnt += 1;
+            if rx_cnt == rx.len() {
+                rx_done = true;
+            }
+        }
+
+        if tx_done && rx_done {
```
andrewjstone, line 128: What happens if the loops bail early and you never get `tx_done` or `rx_done`? Is that a safe condition?

labbott: this is designed to be used in very limited circumstances right now, mostly for hardware signal testing. Even so, I'm not sure I understand what you mean by "safe".  If `tx_done` and `rx_done` are not set this will loop to wait for the next interrupt that either the TX or RX FIFOs are ready. It's fine to wait indefinitely.

andrewjstone: Thanks!

andrewjstone: Thanks. I just meant was there any special condition that would need to be taken the next time around the loop. Looks like the answer is no, which clarifies my understanding.

PR #434, opened 2022-02-21
WIP: LPC55 RNG driver and API crate

flihp:
This is mostly complete but I've marked it as a 'work in progress' because:
1. The RNG driver task uses more flash than seems reasonable (> 16384). I've been tipped off to `cargo bloat` and will be starting there.
2. My error propagation through the rand / rand_core / getrandom traits is awkward at best. I'm expecting this is mostly me using the wrong traits / types for the job.
3. I'm working on a humility interface to this for testing. Currently I'm driving this manually from another [task](https://github.com/flihp/hubris/tree/drv-rng-api-test).
4. The API crate is intended to be used on top of more than just the LPC55 RNG hardware and I'm working up another driver to prove this out on another board.

Advice on any of the above + generic code review / feedback would be greatly appreciated.


flihp:
I couldn't get `cargo bloat` to work. My hunch that it was related to the `ChaCha20RngCore` from the `chacha20` crate panned out (not much else to blame it on really). Replacing the `ChaCha20RngCore` from `chacha20` crate with the `ChaCha20Core` from the `rand_chahca` crate has reduced the FLASH usage by > 6k in my initial testing. The tradeoff seems to be that this type uses more stack space but trading ~300 bytes of stack to get back  ~6k of flash seems lake a fair trade. An update is incoming.

PR #435, opened 2022-02-23
add support for "humility flash"

bcantrill:
Draft commit message:

```
This work adds two bits to the archive:  the final binaries that we
actually flashed, and some metadata to explain how that can be
reproduced via a flashing program (i.e., OpenOCD or pyOCD depending on
the target).
```

### DIFF
--- a/build/xtask/src/flash.rs
+++ b/build/xtask/src/flash.rs
@@ -12,23 +12,46 @@ use anyhow::Context;
 
 use crate::Config;
 
+//
+// We allow for enough information to be put in the archive for the image to
+// be flashed based only on the archive (e.g., by Humility).  Because flashing
+// is itself a bit of a mess (requiring different programs for different
+// targets), this is a bit gritty (e.g., any required external configuration
+// files must themselves put in the archive).  If these structures need to
+// change, be sure to make corresponding changes to Humility.
+//
 #[derive(Debug, Serialize)]
 pub enum FlashProgram {
     PyOcd(Vec<FlashArgument>),
     OpenOcd(FlashProgramConfig),
 }
 
+//
+// Enum describing flash programs configuration (e.g., "openocd.cfg" for
+// OpenOCD), either as a path in the file system or with the entire contents.
+//
 #[derive(Debug, Serialize)]
 pub enum FlashProgramConfig {
     Path(Vec<String>),
     Payload(String),
 }
 
+//
+// An enum describing a single command-line argument to the flash program.
+//
 #[derive(Debug, Serialize)]
 pub enum FlashArgument {
+    // A direct string
     Direct(String),
+
+    // The filesystem path of the binary flash payload itself
     Payload,
+
+    // A single argument consisting of a prefix, the path of the flash
+    // payload, and a suffix
     FormattedPayload(String, String),
+
+    // The filesystem path of the flash program configuration
     Config,
 }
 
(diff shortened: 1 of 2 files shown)


build/xtask/src/flash.rs
```diff
@@ -11,59 +12,233 @@ use anyhow::Context;
...
-                .arg("--format")
-                .arg("hex")
-                .arg(out.join("final.ihex"));
+    // The filesystem path of the binary flash payload itself
+    Payload,
 
-            let mut reset = Command::new("pyocd");
-            reset.arg("reset").arg("-t").arg("lpc55s69");
+    // A single argument consisting of a prefix, the path of the flash
+    // payload, and a suffix
```
mkeeter, line 51: There are only two things here, not three (and it looks like we hardcode `final.srec` as the payload name?)

build/xtask/src/flash.rs
```diff
@@ -11,59 +12,233 @@ use anyhow::Context;
...
+    // Slurp in any flash program configuration file and flatten it into
+    // our overall configuration
+    //
+    pub fn flatten(&mut self) -> anyhow::Result<()> {
+        if let FlashProgram::OpenOcd(ref config) = self.program {
+            if let FlashProgramConfig::Path(ref path) = config {
+                let p: PathBuf = path.iter().collect();
+                let text = std::fs::read_to_string(p)?;
+                self.program =
+                    FlashProgram::OpenOcd(FlashProgramConfig::Payload(text));
```
mkeeter, line 129: Vague architectural complaint:

`FlashProgramConfig::Path` is only ever used here in `run`; `FlashProgramConfig::Payload` is only ever written to a `flash.ron` file.  They're the same (enum) type, but having the wrong enum will break things either here or in Humility.

It would be less error-prone to _always_ flatten (and modify `run` below); then we could do flattening immediately in `FlashProgramConfig::new` and remove the `enum` entirely.

build/xtask/src/flash.rs
```diff
@@ -11,59 +12,233 @@ use anyhow::Context;
...
+                .config()
+                .arg("-c")
+                .formatted_payload("program", "verify reset")
+                .arg("-c")
+                .arg("exit");
+
+            Ok(flash)
+        }
+        _ => {
+            anyhow::bail!("unrecnogized board {}", board);
```
mkeeter, line 201: typo

build/xtask/src/flash.rs
```diff
@@ -11,59 +12,233 @@ use anyhow::Context;
...
-    ctrlc::set_handler(|| {}).expect("Error setting Ctrl-C handler");
+//
+// We allow for enough information to be put in the archive for the image to
+// be flashed based only on the archive (e.g., by Humility).  Because flashing
+// is itself a bit of a mess (requiring different programs for different
+// targets), this is a bit gritty (e.g., any required external configuration
+// files must themselves put in the archive).  If these structures need to
+// change, be sure to make corresponding changes to Humility.
+//
+#[derive(Debug, Serialize)]
```
mkeeter, line 23: One day, we need a common crate so Hubris and Humility can share stuff like this (but that shouldn't block this PR).

mkeeter (approved):
A few minor notes, but LGTM

build/xtask/src/flash.rs
bcantrill: Will clarify the comment -- it's up to the user of this thing to inject the path to the payload in the middle.

build/xtask/src/flash.rs
bcantrill: Agreed and agreed.

build/xtask/src/flash.rs
bcantrill: So, that would mean that in the case of flashing live, we are (unnecessarily) slurping in the file and then writing it back out again to a temporary file.  Happy to do that, but it feels like unnecessary complexity?  (Especially because the temporary file would be blown away by default.)

build/xtask/src/flash.rs
bcantrill: Thanks -- will fix!

build/xtask/src/flash.rs
mkeeter: Yeah, I'm now on the fence after having read the Humility side of things (and noticing that it reads + writes a file).  I think this is okay for now, since it will fail loudly if you forget to flatten it.

Presumably, the future "shared Hubris + Humility functionality" crate would handle both cases 😄

PR #438, opened 2022-02-24
bump toolchain to last night's nightly

steveklabnik:
The decision to re-purpose the `asm` feature flag [was changed](https://github.com/rust-lang/rust/issues/72016#issuecomment-1021679663), which is great. Let's get us on a version of the toolchain that uses the new flags.

~~This won't be quite ready just yet; I tested using the f4 demo; gonna use one CI run to show me what else breaks so I can fix it up.~~ Now ready!

### DIFF
--- a/app/demo-stm32f4-discovery/app.toml
+++ b/app/demo-stm32f4-discovery/app.toml
@@ -44,7 +44,7 @@ stacksize = 1536
 name = "drv-stm32fx-rcc"
 features = ["f4"]
 priority = 1
-max-sizes = {flash = 4096, ram = 1024}
+max-sizes = {flash = 8192, ram = 1024}
 uses = ["rcc"]
 start = true
 
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -551,15 +551,14 @@ impl BuildConfig<'_> {
             None => PathBuf::from("cargo"),
         });
 
-        // nightly features that we use: asm_sym,named-profiles,naked_functions
-        // cmse_nonsecure_entry,array_methods,destructuring_assignment
-        // nightly features that our dependencies use: proc_macro_span,backtrace
-        // asm,llvm_asm
+        // nightly features that we use: asm_sym,asm_const,
+        // named-profiles,naked_functions,cmse_nonsecure_entry,array_methods
+        //
+        // nightly features that our dependencies use: backtrace,proc_macro_span
 
         cmd.arg(
-            "-Zallow-features=asm,asm_sym,named-profiles,naked_functions,\
-cmse_nonsecure_entry,array_methods,destructuring_assignment,proc_macro_span,\
-backtrace,llvm_asm",
+            "-Zallow-features=asm_sym,asm_const,named-profiles,naked_functions,\
+cmse_nonsecure_entry,array_methods,backtrace,proc_macro_span",
         );
 
         cmd.arg(subcommand);
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -1287,6 +1287,7 @@ fn link(
     cmd.arg("-m").arg(m);
     cmd.arg("-z").arg("common-page-size=0x20");
     cmd.arg("-z").arg("max-page-size=0x20");
+    cmd.arg("-rustc-lld-flavor=ld");
 
     cmd.current_dir(working_dir);
 
--- a/drv/stm32xx-sys/src/main.rs
+++ b/drv/stm32xx-sys/src/main.rs
@@ -6,6 +6,7 @@
 
 #![no_std]
 #![no_main]
+#![deny(unsafe_op_in_unsafe_fn)]
 
 cfg_if::cfg_if! {
     if #[cfg(feature = "family-stm32g0")] {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -363,9 +363,9 @@ checksum = "6245d59a3e82a7fc217c5828a6692dbc6dfb63a0c8c90495621f7b9d79704a0e"
 
 [[package]]
 name = "cortex-m"
-version = "0.7.3"
+version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2ac919ef424449ec8c08d515590ce15d9262c0ca5f0da5b0c901e971a3b783b3"
+checksum = "cd20d4ac4aa86f4f75f239d59e542ef67de87cce2c282818dc6e84155d3ea126"
 dependencies = [
  "bare-metal 0.2.5",
  "bitfield",
@@ -399,6 +399,15 @@ name = "cortex-m-semihosting"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6bffa6c1454368a6aa4811ae60964c38e6996d397ff8095a8b9211b1c1f749bc"
+dependencies = [
+ "cortex-m",
+]
+
+[[package]]
+name = "cortex-m-semihosting"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c23234600452033cc77e4b761e740e02d2c4168e11dbf36ab14a0f58973592b0"
 dependencies = [
  "cortex-m",
 ]
@@ -1167,7 +1176,7 @@ name = "drv-stm32g0-usart"
 version = "0.1.0"
 dependencies = [
  "cortex-m",
- "cortex-m-semihosting",
+ "cortex-m-semihosting 0.5.0",
  "drv-stm32xx-sys-api",
  "num-traits",
  "stm32g0",
@@ -2243,7 +2252,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3d55dedd501dfd02514646e0af4d7016ce36bc12ae177ef52056989966a1eec"
 dependencies = [
  "cortex-m",
- "cortex-m-semihosting",
+ "cortex-m-semihosting 0.3.7",
 ]
 
 [[package]]
@@ -3083,7 +3092,7 @@ dependencies = [
  "armv6m-atomic-hack",
  "build-util",
  "cortex-m",
- "cortex-m-semihosting",
+ "cortex-m-semihosting 0.5.0",
  "hubris-num-tasks",
  "idol",
  "idol-runtime",
@@ -3425,7 +3434,7 @@ version = "0.1.0"
 dependencies = [
  "build-util",
  "cortex-m",
- "cortex-m-semihosting",
+ "cortex-m-semihosting 0.5.0",
  "hubris-num-tasks",
  "num-traits",
  "test-api",
@@ -3468,7 +3477,7 @@ dependencies = [
  "build-util",
  "cfg-if",
  "cortex-m",
- "cortex-m-semihosting",
+ "cortex-m-semihosting 0.5.0",
  "hubris-num-tasks",
  "num-traits",
  "test-api",
--- a/app/demo-stm32f4-discovery/app-f3.toml
+++ b/app/demo-stm32f4-discovery/app-f3.toml
@@ -44,7 +44,7 @@ stacksize = 1536
 name = "drv-stm32fx-rcc"
 features = ["f3"]
 priority = 1
-max-sizes = {flash = 4096, ram = 1024}
+max-sizes = {flash = 8192, ram = 1024}
 uses = ["rcc"]
 start = true
 
@@ -96,4 +96,3 @@ priority = 5
 max-sizes = {flash = 128, ram = 256}
 stacksize = 256
 start = true
-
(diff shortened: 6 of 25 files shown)


app/demo-stm32f4-discovery/app.toml
```diff
@@ -50,7 +50,7 @@ path = "../../drv/stm32fx-rcc"
 name = "drv-stm32fx-rcc"
 features = ["f4"]
 priority = 1
-requires = {flash = 4096, ram = 1024}
+requires = {flash = 8192, ram = 1024}
```
steveklabnik, line 53: This now needs 720 more bytes of text and 884 bytes of rodata. Curse you, powers of two!

(But I haven't determined *why* yet, I am guessing that it is either simply "new compiler generates different code" or "cortex-m@0.7.3 uses more for some reason". We'll see more once I get the whole build going, I guess.

steveklabnik: See https://github.com/oxidecomputer/hubris/pull/438/commits/fe85db89a51ba62b311a805a1c518ebbd911837f for all of the various size changes.

drv/lpc55-spi-server/src/main.rs
```diff
@@ -216,7 +216,7 @@ fn muck_with_gpios(syscon: &Syscon) {
     ];
 
     for (pin, alt, mode, slew, invert, digi, od) in
-        core::array::IntoIter::new(pin_settings)
+        IntoIterator::into_iter(pin_settings)
```
steveklabnik, line 219: this one was a deprecation that i fixed while I was here. the right move may be to just bump this to rust 2021 but that felt like a larger change, so I kept it at this.

drv/stm32xx-sys/src/main.rs
```diff
@@ -61,11 +61,11 @@ where
     S: pac::RegisterSpec<Ux = u32> + pac::Readable + pac::Writable,
 {
     unsafe fn set_bit(&self, index: u8) {
-        self.modify(|r, w| unsafe { w.bits(r.bits() | 1 << index) });
+        self.modify(|r, w| w.bits(r.bits() | 1 << index));
     }
 
     unsafe fn clear_bit(&self, index: u8) {
-        self.modify(|r, w| unsafe { w.bits(r.bits() & !(1 << index)) });
+        self.modify(|r, w| w.bits(r.bits() & !(1 << index)));
```
steveklabnik, line 68: both of these lines are warnings that I fixed, I assume they're new. Can leave this out if someone objects for some reason.

app/gimletlet/app-mgmt.toml
```diff
@@ -85,7 +85,7 @@ path = "../../drv/user-leds"
 name = "drv-user-leds"
 features = ["stm32h7"]
 priority = 2
-requires = {flash = 2048, ram = 1024}
```
mkeeter, line 88: This seems very suspicious – the `user-leds` task should be _tiny_, so this would be a good starting place to investigate the size changes.

drv/stm32xx-sys/src/main.rs
mkeeter: I also ran into this when bumping one of the PAC crates for an experiment, so I think this is fine.

app/gimletlet/app-mgmt.toml
steveklabnik: 99% of the changes in size are on the order of 100-200 bytes; only one of them was around 1000 bytes.

The issue is that these must be powers of two, and so if you need 2049 bytes of flash, you end up needing to put in 4096, even if almost all of it is wasted space.

mkeeter: I did some detective work (`arm-none-eabi-objdump -d --demangle -t user_leds | grep " F .text" | sort -k 6 | cut -w -f5-`) and found that the main change is in `core::fmt::Formatter::pad`, which went from 690 bytes to 1970 bytes!

Here's the full set of functions that changed size
```diff
4c4
< 000002b2      core::fmt::Formatter::pad
---
> 000007b2      core::fmt::Formatter::pad
13d12
< 00000014      drv_user_leds::idl::InOrderUserLedsImpl::closed_recv_fail
15c14
< 00000208      main
---
> 0000020c      main
```

I found [this Rust commit](https://github.com/rust-lang/rust/commit/6c9e708f4be04a5da9267171602be135789b12b8) which touched the function recently, but dunno if it's worth digging further...

drv/stm32xx-sys/src/main.rs
cbiffle: This is fine, the PAC crates occasionally mark `bits` accessors as safe (sometimes in ways that are not correct), so any time we bump the version expect to move some unsafes around.

drv/lpc55-spi-server/src/main.rs
cbiffle: That's probably fine, though bumping to 2021 probably doesn't have _that_ many implications for this crate

steveklabnik:
Rebased.

I heard from a fairly reputable source that Rust 2021 supposedly made panics not use the formatting machinery, with the express goal of reducing binary size. However, it's our dependencies, not us, that are the problem here. I changed all of our packages to 2021 and it didn't change anything.

steveklabnik:
@cbiffle during the all hands you had said that you had investigated a bit too, but I forget what you said, and it didn't make it to a comment here. Is this okay to merge?

steveklabnik:
Since this still has not landed and it's been a long while, and due to https://github.com/rust-lang/rust/issues/95228, I plan on updating this tomorrow to use the nightly that ^ has landed in, so we can try it out.

Gankra:
yesss... fix your pointer crimes......

mkeeter:
I prefer to think of them as pointer peccadillos

(mostly because it's funnier sounding)

steveklabnik:
I.... did some work this morning to update this, and went to push it, and got

```text
> git push origin update-toolchain
Enumerating objects: 63, done.
Counting objects: 100% (63/63), done.
Delta compression using up to 32 threads
Compressing objects: 100% (29/29), done.
Writing objects: 100% (33/33), 3.14 KiB | 1.05 MiB/s, done.
Total 33 (delta 24), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (24/24), completed with 23 local objects.
remote:
remote: Create a pull request for 'update-toolchain' on GitHub by visiting:
remote:      https://github.com/oxidecomputer/hubris/pull/new/update-toolchain
remote:
To https://github.com/oxidecomputer/hubris
 * [new branch]        update-toolchain -> update-toolchain
```

... why did I name two branches `update-toolchain` and `bump-toolchain`, sigh.

I have to go move some furniture, so this will get updated this afternoon instead of this morning, I was so excited, haha.

steveklabnik:
I have tortured my poor computer so much today. Sorry, computer.

I completely redid this branch from scratch, with today's nightly. I tested every single app, even those not on CI. 

https://github.com/oxidecomputer/hubris/pull/438/commits/92e126b3fb082f464cf6293d78e8c5f8f47e2e97 in particular will end up breaking by tomorrow thanks to upstream; we will have to fix this before landing.

I adjusted every app that had size suggestions *except* the lpc55 ones because those were not powers of 2, which felt like introducing a *lot* of churn as code changes land. I kept these in separate commits so that if we don't want them, I can easily take them out of this PR.

drv/stm32xx-gpio-common/Cargo.toml
```diff
@@ -11,12 +11,7 @@ stm32h7 = { version = "0.14", default-features = false, optional = true }
 userlib = {path = "../../sys/userlib"}
 num-traits = {version = "0.2", default-features = false}
 zerocopy = "0.6.1"
-
-[dependencies.stm32g0]
-optional = true
-git = "https://github.com/oxidecomputer/stm32-rs-nightlies"
-branch = "stm32g0b1-initial-support"
-default-features = false
+stm32g0 = { git = "https://github.com/oxidecomputer/stm32-rs-nightlies", branch = "20220401-snap", default-features = false, optional = true }
```
steveklabnik, line 14: i do prefer the old way of doing this, but the codebase overwhelmingly uses one line, so I've normalized this, since I was here.

steveklabnik:
Okay! This (assuming tests pass) should now be good to go.

steveklabnik:
Unfortunately, this seems to be blocked on https://github.com/rust-lang/rust/issues/90357 😭

steveklabnik:
https://github.com/rust-lang/rust/pull/95685 was merged and so this is now unblocked! It's deeply bitrotted of course, but I'm going to pick this back up now.

steveklabnik:
This has now been updated; it also got a bit smaller! 

@cbiffle ready for a re-review whenever :)

build/xtask/src/dist.rs
```diff
@@ -1287,6 +1287,7 @@ fn link(
     cmd.arg("-m").arg(m);
     cmd.arg("-z").arg("common-page-size=0x20");
     cmd.arg("-z").arg("max-page-size=0x20");
+    cmd.arg("-rustc-lld-flavor=ld");
```
steveklabnik, line 1290: FWIW: I don't know why I need to make this change. In theory this is supposed to be detected, but isn't now? The docs on this are effectively non-existent...

build/xtask/src/config.rs
```diff
@@ -551,13 +551,14 @@ impl BuildConfig<'_> {
...
-        // cmse_nonsecure_entry,array_methods,destructuring_assignment
+        // nightly features that we use: asm_sym,asm_const,
+        // named-profiles,naked_functions,cmse_nonsecure_entry,array_methods,
+        // destructuring_assignment
         // nightly features that our dependencies use: proc_macro_span,backtrace
         // asm,llvm_asm
 
         cmd.arg(
-            "-Zallow-features=asm,asm_sym,named-profiles,naked_functions,\
+            "-Zallow-features=asm,asm_sym,asm_const,named-profiles,naked_functions,\
```
cbiffle, line 561: You ought to be able to remove `asm` from this list now, right?

cbiffle (approved):
One note below

build/xtask/src/dist.rs
luqmana: A brief recap about what led to needing to provide `-rustc-lld-flavor`:

Calling `ld` directly to link is a bit of a pain as you need to deal with all the platform specific logic (e.g. like where to find crt1.o) which you can avoid by invoking it via `cc/gcc/clang`. Up until recently, you couldn't pass the linker via an absolute path to `gcc` but instead had to add a search path that would have the `ld` binary. Hence why `$SYSROOT/lib/rustlib/$HOST_TARGET/bin/gcc-ld` exists. This is what happens when you pass `-Zgcc-ld=lld` (soon to be stabilized as `-Clink-self-contained=linker -Clinker-flavor=gcc-lld`) to `rustc` to use the bundled `rust-lld`.

Now, `lld` decides what flavour to use based on either the name of the binary invoked or via the first argument (i.e., `arg[0]` or `arg[1]`). Unfortunately, rust couldn't just ship `ld` and `ld64` as symlinks to `rust-lld` so the `gcc-ld` folder just had multiple copies which meant some bloat.

All's fine for Hubris at this point because we just directly invoked `$SYSROOT/lib/rustlib/$HOST_TARGET/bin/gcc-ld/ld` and so `lld` operated with the `ld` flavour.

Around Oct last year to reduce the bloat (3 copies of `rust-lld`), the binaries under `gcc-ld` were replaced with a small tool `lld-wrapper` that would call `../rust-lld` with the right flavor. So now the `ld` and `ld64` under `bin/gcc-ld/` were `lld-wrapper` compiled with different `cfg` flags choosing the flavour to pass to `../rust-lld`. Up until this point, I'd expect `hubris` to still work when invoking this `ld`.

But this was still a pain, with the duplication and lacked some other support. So in May-ish, it was simplified to a single copy that would correctly place the flavour argument as the first one passed to `rust-lld` via `-rustc-lld-flavor`. Although, looks like this did come with some [downsides](https://github.com/rust-lang/rust/issues/97755) that might mean further changes coming.

Looking at hubris, looks like we're manually invoking the linker here and not through `cc/gcc/clang`. I would say we should just call `rust-lld` directly instead of using whatever is in under `bin/gcc-ld` (really all that is an implementation detail for `rustc` hence the name of the flag).

build/xtask/src/config.rs
```diff
@@ -551,13 +551,14 @@ impl BuildConfig<'_> {
...
-        // cmse_nonsecure_entry,array_methods,destructuring_assignment
+        // nightly features that we use: asm_sym,asm_const,
+        // named-profiles,naked_functions,cmse_nonsecure_entry,array_methods,
+        // destructuring_assignment
         // nightly features that our dependencies use: proc_macro_span,backtrace
         // asm,llvm_asm
 
         cmd.arg(
-            "-Zallow-features=asm,asm_sym,named-profiles,naked_functions,\
+            "-Zallow-features=asm_sym,asm_const,named-profiles,naked_functions,\
```
luqmana, line 561: Should be able to remove `destructuring_assignment` right?

steveklabnik: Yes!

I probably should just re-do the list again manually... I guess I really don't want just an allowlist, but like, bidirectional checking that they're actually required, ha!

I'll fix this soon

steveklabnik: I've re-done this now; our dependencies no longer use `asm` and `llvm_asm` too :)

PR #441, opened 2022-02-24
port xtask to env_logger

steveklabnik:
So, after doing this for Humility, we had talked about doing it for Hubris. Here's what the naïve version of it looks like:

image
image

As you can see, it is, uh. Lots noisier. We can change that, but before doing so I figured I'd at least show the defaults. I also didn't touch the memory allocation warning just yet, it looks so good as it is now, and this would uh, make it a bit worse.

So, thoughts? I'm not sure if this is an improvement strictly speaking. In general I kind of feel like the build output is a bit of a mess at the moment, but I'm also not 100% sure what I would want it to be like in an ideal world yet.


steveklabnik:
If we do decide to do this, I'd just do it over again at this point.

PR #444, opened 2022-03-08
Stop with an error if cargo xtask build has an invalid task

### LINKED ISSUE #443: cargo xtask build should error on invalid task name
If you run `cargo xtask build app/bla/app.toml not-a-real-task`, it will complete successfully.  It should instead produce an error, and ideally print the list of valid task names.

mkeeter:
Fixes #443


steveklabnik (approved):
Awesome, thank you!

bcantrill (approved):
THANK YOU. I have run into this too, and should have fixed it on the spot -- thank you for fixing it!

PR #446, opened 2022-03-11
allow for multiple CS pins

bcantrill:
Draft commit message:

```
allow for multiple CS + Gimlet sequencing improvements

The first Gimlet boards have an annoying habit of getting their CS
blown on the sequencer FPGA.  To allow for these kind of busted
snowflakes, we extend SPI to allow for multiple CS pins for a single
device: for most devices most of the time, a single CS will be
specified, but when rework is required and another pin is used, this
allows for a single application definition across the reworked and
non-reworked boards.

This change also includes a few modest improvements to the Gimlet
sequencing:  

- Always load the clock generator's config as part of going to A2
- Update the clock generator config to reflect the management network
- Always enable the SP-to-SP3 UART to allow for transmission into SP3
- Track the tasks making Sequencer.get_state calls

```

Gimlet verification to be done:

- [x] Test on grimey (A2, A0, UART)
- [x] Test on lucky (A2, A0, UART)
- [x] Test on sweaty  (A2, A0, UART)
- [x] Test on zappy (A2)
- [x] Test on zombie (A2)


drv/gimlet-seq-server/src/main.rs
```diff
@@ -536,6 +524,31 @@ cfg_if::cfg_if! {
...
+        // into the SP3's UART
+        //
+        const UART_TX_ENABLE_PIN: (sys_api::Port, u16) =
+            (sys_api::Port::A, 1 << 5);
+
+        fn uart_sp_to_sp3_enable() {
+            let sys = sys_api::Sys::from(SYS.get_task_id());
+            let (port, pin_mask) = UART_TX_ENABLE_PIN;
+
+            sys.gpio_configure(
```
mkeeter, line 538: You can use `sys.gpio_configure_output` here to simplify

drv/gimlet-seq-server/src/main.rs
```diff
@@ -307,35 +297,52 @@ fn main() -> ! {
...
+    hl::sleep_for(150);
+
+    //
+    // And now load our clock configuration
+    //
+    let mut packet = 0;
+    let clockgen = i2c_config::devices::idt8a34003(I2C.get_task_id())[0];
+
+    payload::idt8a3xxxx_payload(|buf| match clockgen.write(buf) {
+        Err(err) => Err(err),
```
mkeeter, line 313: It looks like we previously logged these results in the ringbuf – is it deliberate to not log them now?

If so, this could be slightly simplified to
```rust
|buf| {
    clockgen.write(buf)?;
    packet += 1;
    Ok(());
}
```

drv/stm32h7-spi-server/src/main.rs
```diff
@@ -81,18 +81,19 @@ fn main() -> ! {
     // Configure all devices' CS pins to be deasserted (set).
     // We leave them in GPIO output mode from this point forward.
     for device in CONFIG.devices {
-        sys.gpio_set_reset(device.cs.port, device.cs.pin_mask, 0)
+        for pin in device.cs {
+            sys.gpio_set_reset(pin.port, pin.pin_mask, 0).unwrap();
```
mkeeter, line 85: Use `sys.gpio_set(pin)` and `sys.gpio_configure_output(pin, ...)` here instead of the fully general functions

drv/stm32h7-spi-server/src/main.rs
```diff
@@ -197,16 +198,19 @@ impl InOrderSpiImpl for ServerImpl {
...
-                device.cs.port,
-                if cs_asserted { 0 } else { pin_mask },
-                if cs_asserted { pin_mask } else { 0 },
-            )
-            .unwrap();
+        for pin in device.cs {
+            // If we're asserting CS, we want to *reset* the pin. If
+            // we're not, we want to *set* it. Because CS is active low.
+            let pin_mask = pin.pin_mask;
+            self.sys
```
mkeeter, line 205: What about this?
```rust
if cs_asserted {
    self.sys.gpio_reset(pin).unwrap();
} else {
    self.sys.gpio_set(pin).unwrap();
}
```

drv/stm32h7-spi-server/src/main.rs
```diff
@@ -351,9 +356,9 @@ impl ServerImpl {
         // We're doing this! Check if we need to control CS.
         let cs_override = self.lock_holder.is_some();
         if !cs_override {
-            self.sys
-                .gpio_set_reset(device.cs.port, 0, device.cs.pin_mask)
-                .unwrap();
+            for pin in device.cs {
+                self.sys.gpio_set_reset(pin.port, 0, pin.pin_mask).unwrap();
```
mkeeter, line 360:
```suggestion
                self.sys.gpio_reset(pin).unwrap();
```

drv/stm32h7-spi-server/src/main.rs
```diff
@@ -499,9 +504,9 @@ impl ServerImpl {
 
         // Deassert (set) CS, if we asserted it in the first place.
         if !cs_override {
-            self.sys
-                .gpio_set_reset(device.cs.port, device.cs.pin_mask, 0)
-                .unwrap();
+            for pin in device.cs {
+                self.sys.gpio_set_reset(pin.port, pin.pin_mask, 0).unwrap();
```
mkeeter, line 508:
```suggestion
                self.sys.gpio_set(pin).unwrap();
```

mkeeter (approved):
Lots of suggestions to use nicer GPIO APIs, but otherwise looks good!

drv/gimlet-seq-server/src/main.rs
bcantrill: Done -- thank you!

drv/gimlet-seq-server/src/main.rs
bcantrill: I *did* purposely not log them -- and agreed on the simplification!

drv/gimlet-seq-server/src/main.rs
```diff
@@ -536,6 +524,31 @@ cfg_if::cfg_if! {
...
+                pin_mask,
+                sys_api::Mode::Output,
+                sys_api::OutputType::PushPull,
+                sys_api::Speed::High,
+                sys_api::Pull::None,
+                sys_api::Alternate::AF0, // doesn't matter
+            )
+            .unwrap();
+
+            sys.gpio_set_reset(port, 0, pin_mask).unwrap();
```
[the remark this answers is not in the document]
bcantrill: Yes, this can be even simpler -- the `gpio_reset` API is handy, it turns out!

drv/gimlet-seq-server/src/payload.rs
```diff
@@ -64,12 +64,7 @@ pub fn idt8a3xxxx_payload<E>(
         &[
```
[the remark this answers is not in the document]
bcantrill: Yes, that would be good follow-on work to do.

drv/stm32h7-spi-server/src/main.rs
bcantrill: Definitely prefer it -- fixing, and thank you!

PR #447, opened 2022-03-14
Minor README installation fixes

jgallagher:
The update to `cargo install... humility` is required to avoid an error from cargo in which it doesn't know whether you want to install `humility` or `xtask`.

The updates to the package names are for ubuntu.


steveklabnik (approved):

steveklabnik:
Thank you! Shoulda realized that with the `xtask`.

PR #448, opened 2022-03-14
RNG

flihp:
This PR creates a common API crate for using RNGs on Hubris. The interfaces is defined using the IDL. It's a single function that takes a byte slice from the caller and fills it with entropy from the RNG. The API crate wraps this new API in the RngCore trait from rand_core & exposes a [custom 'getrandom' implementation](https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html) for convenience.

The Lpc55 RNG driver to is updated to support this new interface and its output is conditioned through a ChaCha20 CPRNG that is reseeded from the RNG hardware every 1MiB output to clients. We do this because this part is not validated against any standard statistical tests by NXP (section 48.15.2 from UM11126 rev 2.4). NOTE: The chi^2 test for the Lpc55 has been removed because it keeps getting wedged.

A driver for the RNG on the stm32h743 & h753 are added as well. According to STM this hardware RNG is certified against BSI AIS-31 statistical tests and so we return entropy from the hardware directly to the caller.

Finally we add a function to hiffy to allow humility to interact with the RNG drivers.


drv/lpc55-rng/src/main.rs
```diff
@@ -2,81 +2,146 @@
...
 
-    let rng = unsafe { &*device::RNG::ptr() };
-    let pmc = unsafe { &*device::PMC::ptr() };
+struct Lpc55Rng(BlockRng<Lpc55Core>);
 
-    let mut buffer = [0u32; 1];
+impl Lpc55Rng {
+    fn new() -> Self {
+        Lpc55Rng {
+            0: BlockRng::new(Lpc55Core::new()),
```
kc8apf, line 56: Unusual to use this style of initialization for a tuple struct.

flihp: Fixed. There's a clippy lint for this but it's only a few months old and hasn't trickled down yet: https://github.com/rust-lang/rust-clippy/pull/8170

PR #455, opened 2022-03-16
Fix lpc55 clocks

labbott:


bcantrill (approved):
Looks good, even with the ugly checks! ;)

PR #456, opened 2022-03-16
Hiffy SPI operations should specify device

bcantrill:


mkeeter:
The code looks reasonable, but is there any way to make it fail loudly on a Humility version mismatch?

- Old Humility -> New Hubris will send too few arguments, so that should fail (unless there's something else in the Hiffy stack)
- New Humility -> Old Hubris will send the device argument where firmware expects `len`, so that would do something unexpected

I know we've got an explicit `HIFFY_VERSION_MAJOR`, but that's just the HIF version; it seems like we also want a `HIFFY_API_VERSION` or similar.

bcantrill:
> The code looks reasonable, but is there any way to make it fail loudly on a Humility version mismatch?
> 
> * Old Humility -> New Hubris will send too few arguments, so that should fail (unless there's something else in the Hiffy stack)
> * New Humility -> Old Hubris will send the device argument where firmware expects `len`, so that would do something unexpected
> 
> I know we've got an explicit `HIFFY_VERSION_MAJOR`, but that's just the HIF version; it seems like we also want a `HIFFY_API_VERSION` or similar.

The good news is that the failure mode is pretty explicit; new Humility on old Hubris will fail with:

```
humility spi failed: mismatched function signature on SpiRead (args.len:3, nargs:4
```

And old Humility on new Hubris will fail with:

```
humility spi failed: mismatched function signature on SpiRead (args.len:4, nargs:3
````

This message can be improved a bit, but it should at least point folks in the right direction!

**Comment by bcantrill** [comment]
> > The code looks reasonable, but is there any way to make it fail loudly on a Humility version mismatch?
> > 
> > * Old Humility -> New Hubris will send too few arguments, so that should fail (unless there's something else in the Hiffy stack)
> > * New Humility -> Old Hubris will send the device argument where firmware expects `len`, so that would do something unexpected
> > 
> > I know we've got an explicit `HIFFY_VERSION_MAJOR`, but that's just the HIF version; it seems like we also want a `HIFFY_API_VERSION` or similar.
> 
> The good news is that the failure mode is pretty explicit; new Humility on old Hubris will fail with:
> 
> ```
> humility spi failed: mismatched function signature on SpiRead (args.len:3, nargs:4
> ```
> 
> And old Humility on new Hubris will fail with:
> 
> ```
> humility spi failed: mismatched function signature on SpiRead (args.len:4, nargs:3
> ```
> 
> This message can be improved a bit, but it should at least point folks in the right direction!

In the spirit of improving that message, I am changing it to:

```
humility spi failed: mismatched function signature on SpiRead (target has 4 args, but we expected 3); missing Hubris/Humility flag day?
```

mkeeter (approved):
Great, that seems clear enough!

PR #461, opened 2022-03-18
Add derive macro for Idol errors

mkeeter:
This removes a bunch of boilerplate for error types which are used in Idol / Hiffy.

(I also removed an unused task from `gimletlet/app-vsc7448`, since it was complaining about being oversized)


steveklabnik (approved):
This looks good to me! Getting rid of this boilerplate seems quite nice.

PR #462, opened 2022-03-18
Reduce kernel-link deltas relative to cortex-m-rt

cbiffle:
While this contains a number of small formatting/layout changes to try
and reduce the diff, the critical change is the re-insertion of the
PreResetTrampoline section and its accompanying comment. Removing these
broke the build on M0, where branches have very limited range.

### DIFF
--- a/build/kernel-link.x
+++ b/build/kernel-link.x
@@ -3,9 +3,8 @@
 INCLUDE memory.x
 
 /* # Entry point = reset vector */
-EXTERN(__RESET_VECTOR);
-EXTERN(Reset);
 ENTRY(Reset);
+EXTERN(__RESET_VECTOR); /* depends on the `Reset` symbol */
 
 /* # Exception vectors */
 /* This is effectively weak aliasing at the linker level */
@@ -41,7 +40,7 @@ PROVIDE(__pre_init = DefaultPreInit);
 SECTIONS
 {
   PROVIDE(_stack_start = ORIGIN(RAM) + LENGTH(RAM));
- 
+
   /* ## Sections in FLASH */
   /* ### Vector table */
   .vector_table ORIGIN(FLASH) :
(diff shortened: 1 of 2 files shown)


mkeeter (approved):

build/kernel-link.x
```diff
@@ -70,12 +69,14 @@ SECTIONS
 	KEEP(*(.image_header));
   } > FLASH
 
-  PROVIDE(_stext = ADDR(.vector_table) + SIZEOF(.vector_table) + SIZEOF(.header));
+  PROVIDE(_stext = ADDR(.header) + SIZEOF(.header));
```
labbott, line 72: I'm not sure about changing this definition. This now sets `_stext` relative to the header as opposed to the vector table. It's the same technically but spelling it out explicitly seems clearer.

cbiffle: The text will follow the header, the original expression could become wrong if the header's alignment were changed (which would make it more than just a sizeof-offset from the previous section). This expression of it is more robust.

labbott: Do we want it to be robust or do we want it to fail so we know the alignment has changed? If we can get it to fail I'd prefer that over finding something with the header changed.

cbiffle: I believe if I rolled back to what you had, it would fail if the alignment were changed, though I don't totally understand why that's good. I'm willing to do that with a large comment explaining why it's built that way when it looks wrong/ increases the delta.

cbiffle: Change rolled back and comment added.

labbott: My real concern is something silently succeeding. If the alignment has changed without us intending it I don't want other things that may be making that assumption to break. Maybe nothing else cares about the alignment but I'm deeply distrustful of linker scripts at this point.

PR #465, opened 2022-03-21
Add `expand` command to xtask

jgallagher:
One minor (I think) implementation wart/question; I'll call it out in the diff below.

### DIFF
--- a/build/xtask/src/expand.rs
+++ b/build/xtask/src/expand.rs
@@ -1,0 +1,69 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::process::Command;
+use std::{env, path::Path};
+
+use anyhow::{bail, Result};
+
+use crate::Config;
+
+pub fn run(
+    cfg: Option<&Path>,
+    package: Option<String>,
+    target: Option<String>,
+) -> Result<()> {
+    let package = package.unwrap_or_else(|| {
+        let path = env::current_dir().unwrap();
+        let manifest_path = path.join("Cargo.toml");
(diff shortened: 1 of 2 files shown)


build/xtask/src/expand.rs
```diff
@@ -0,0 +1,69 @@
...
+    // this is only actually used for demo-stm32h7 but is harmless to include,
+    // so let's do it unconditionally
+    cmd.env("HUBRIS_BOARD", "nucleo-h743zi2");
+
+    // Expanding tasks that include build-time config from their app config
+    // (e.g., lists of I2C devices) requires specifing the app config.
+    if let Some(cfg) = cfg {
+        let toml = Config::from_file(cfg)?;
+        cmd.env("HUBRIS_APP_CONFIG", toml::to_string(&toml.config).unwrap());
+    }
```
jgallagher, line 60: This block seems a bit awkward to me, but it was the only way I could get the `sensor` task to expand, since it requires the I2C devices from the app config at build time.

This file is heavily derived from the `clippy` subcommand, which (to my surprise) does not have this block. I think it may need it to run on `sensor`; currently if I try that it fails clippy-ifying build time dependencies.

build/xtask/src/expand.rs
```diff
@@ -0,0 +1,69 @@
...
+        let path = env::current_dir().unwrap();
+        let manifest_path = path.join("Cargo.toml");
+        let contents = std::fs::read(manifest_path).unwrap();
+        let toml: toml::Value = toml::from_slice(&contents).unwrap();
+
+        // someday, try blocks will be stable...
+        toml.get("package")
+            .and_then(|v| v.get("name"))
+            .and_then(|v| v.as_str())
+            .expect("Couldn't find [package.name]; pass -p <package> to run clippy on a specific package or --all to check all packages")
```
mkeeter, line 27: s/clippy/expand

PR #466, opened 2022-03-21
task-config! macro

mkeeter:
**Docstring pasted below**

The `task_config!` macro defines a `struct TASK_CONFIG` which is pulled
from the Hubris task config.

For example, the following definition could live in a task's `main.rs`:
```rust
task_config::task_config! {
    count: usize,
    leds: &'static [(drv_stm32xx_sys_api::PinSet, bool)],
}
```

Then, it look for a `config` block in the `user_leds` task:
```toml
[tasks.user_leds.config]
count = 4
leds = [
    ["drv_stm32xx_sys_api::Port::C.pin(6)", true],
    ["drv_stm32xx_sys_api::Port::I.pin(8)", false],
    ["drv_stm32xx_sys_api::Port::I.pin(9)", false],
    ["drv_stm32xx_sys_api::Port::I.pin(10)", false],
    ["drv_stm32xx_sys_api::Port::I.pin(11)", false],
]
```

This would generate the following Rust code:
```rust
struct TaskConfig {
    count: usize,
    leds: &'static [(drv_stm32xx_sys_api::PinSet, bool)],
}
const TASK_CONFIG: TaskConfig {
    count: 4,
    leds: &[
        (drv_stm32xx_sys_api::Port::C.pin(6), true),
        (drv_stm32xx_sys_api::Port::I.pin(8), false),
        (drv_stm32xx_sys_api::Port::I.pin(9), false),
        (drv_stm32xx_sys_api::Port::I.pin(10), false),
        (drv_stm32xx_sys_api::Port::I.pin(11), false),
    ]
}
```

At the moment, this only supports tasks which are instantiated _once_ and
configured through the task configuration block (e.g. the SPI driver
cannot be configured using this macro).

**end of docstring**

Sadly, this PR doesn't actually _use_ the new macro anywhere; there are a bunch of places where it's 90% applicable, but it needs extra glue code to actually fit it in.

While looking into this, I realized I don't actually understand why we split configuration between APP and TASK.

For example, I don't know why we do this:
```
[tasks.spi4_driver.config.spi]
global_config = "spi4"

[config.spi.spi4]
controller = 4
...
```

Instead of this:
```
[tasks.spi4_driver.config]
controller = 4
...
```

Anyone have historical insights?  If we could always use the second option, it would make configuration much simpler.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1159,7 +1159,6 @@ dependencies = [
  "num-traits",
  "stm32f3",
  "stm32f4",
- "task-config",
  "userlib",
  "zerocopy",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3080,6 +3080,7 @@ dependencies = [
  "cortex-m",
  "hubris-num-tasks",
  "num-traits",
+ "task-config",
  "test-api",
  "test-idol-api",
  "userlib",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3079,6 +3079,7 @@ dependencies = [
  "cfg-if 0.1.10",
  "cortex-m",
  "hubris-num-tasks",
+ "hypocalls",
  "num-traits",
  "task-config",
  "test-api",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1159,6 +1159,7 @@ dependencies = [
  "num-traits",
  "stm32f3",
  "stm32f4",
+ "task-config",
  "userlib",
  "zerocopy",
 ]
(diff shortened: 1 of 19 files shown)


mkeeter:
Yes, right now it just builds a single `const TASK_CONFIG: TaskConfig = TaskConfig { ... }` (with the struct and type name hardcoded).

We _could_ tweak the macro syntax to enable multiple blocks with user-defined names and types, e.g. something like

```rust
task_config::task_config! {
    const THIS_TASK_CONFIG = TaskConfig {
        count: usize,
        leds: &'static [(drv_stm32xx_sys_api::PinSet, bool)],
    };
    const MORE_TASK_CONFIG = MoreTaskConfig {
        foo: bool,
    };
}
```
```toml
[tasks.user_leds.config.this_task_config]
count = 4
leds = [
    ["drv_stm32xx_sys_api::Port::C.pin(6)", true],
    ["drv_stm32xx_sys_api::Port::I.pin(8)", false],
    ["drv_stm32xx_sys_api::Port::I.pin(9)", false],
    ["drv_stm32xx_sys_api::Port::I.pin(10)", false],
    ["drv_stm32xx_sys_api::Port::I.pin(11)", false],
]
[tasks.user_leds.config.more_task_config]
foo = false
```

One downside is that it's not obvious how to name the TOML block. I guess we could also do a conversion from snake_case to either SCREAMING_SNAKE_CASE or UpperCamelCase, depending on if we want to match the struct name or type name...

(in the example above, I assume snake_case in the TOML and SCREAMING_SNAKE_CASE in the Rust file)

PR #472, opened 2022-03-23
Flash using `humility flash`

labbott:
Now that humility has a flash command as of humility commit
5459f01d0b18f4ef8f4505e049302085d6a0d3dd we can use that directly
instead of calling out to {py, Open}OCD.

Note that one of the boards doesn't work out of the box at the
moment with probe-rs.


andrewjstone (approved):
🎉  Thank you!

PR #475, opened 2022-03-24
Add `usart` lib and `uartecho` task with tx/rx uart support

jgallagher:
I only added the new `uartecho` task to the nucleo h743 board (the one I have at hand), and I did not (yet?) remove the uart driver (which is still used by the ping tasks in other apps).


lib/usart/src/stm32h7.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Debug, Clone, Copy)]
+pub enum DeviceId {
+    Usart3,
+}
+
+pub struct Device(&'static device::usart1::RegisterBlock);
+
+impl Device {
+    /// Turn on the `device` USART with the given baud rate.
+    // TODO passing in a `Sys` seems weird. Should we take the sys task ID
```
cbiffle, line 27: If you're IPCing, the `Sys` is the handle to the remote server, so, it's probably the right thing to pass in.

lib/usart/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+
+#[cfg(any(feature = "stm32h743", feature = "stm32h753"))]
+use self::stm32h7::Device;
+
+use tinyvec::ArrayVec;
+
+/// Supported baud rates
+#[derive(Debug, Clone, Copy)]
+#[repr(u32)]
+pub enum BaudRate {
```
cbiffle, line 26: What value is this enum adding? Could the user just pass in an integer? Otherwise I foresee us adding a bunch of baud rates here as applications develop.

lib/usart/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    Usart<TX_BUF_LEN, RX_BUF_LEN>
+{
+    /// Start managing `device`.
+    ///
+    /// Enables the `irq` interrupt; the caller is responsible for calling
+    /// [`Usart::handle_interrupt()`] when that interrupt fires.
+    pub fn new(usart: Device, irq_mask: u32) -> Self {
+        // Turn on our interrupt. We haven't enabled any interrupt sources at
+        // the USART side yet, but we will momentarily.
+        userlib::sys_irq_control(irq_mask, true);
```
cbiffle, line 57: (Nit: I'd probably leave this in control of the calling code, an operation like `new` turning on some interrupts might be surprising.)

lib/usart/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /// Handle (by stepping tx/rx if possible) and then reenable the USART
+    /// interrupt.
+    pub fn handle_interrupt(&mut self) {
+        // See if we have any data to transmit.
+        if let Some(&byte) = self.tx_buf.get(0) {
+            // Write it to the transmitter, if possible.
+            if self.usart.try_write_tx(byte) {
+                // TODO? `remove()` shifts all remaining tx data down; we could
```
cbiffle, line 78: awwww I was kinda hoping they also provided `ArrayVecDeque` but they do not. :-(

lib/usart/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+                self.rx_overrun = true;
+            }
+        }
+
+        // reenable interrupt
+        userlib::sys_irq_control(self.irq_mask, true);
+    }
+
+    /// Get mutable access to the tx/rx buffers held by `self`.
+    pub fn buffers(
```
cbiffle, line 101: This API is a little odd, could you have read and write functions directly on the Usart? It is rare in practice for the read and write sides to be handled from different code (you may have partially gotten this from the original stm32h7 usart demo task, which is a bit of an anomaly).

lib/usart/src/stm32h7.rs
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+}
+
+fn configure_pins(sys: &Sys, device: DeviceId) {
+    use drv_stm32xx_sys_api::*;
+
+    let tx_rx_mask = match device {
+        DeviceId::Usart3 => {
+            // TODO these are really board configs, not SoC configs!
```
cbiffle, line 135: I'd suggest leaving this to the caller so you don't have to become board-specific here.

lib/usart/src/stm32h7.rs
```diff
@@ -0,0 +1,149 @@
...
+            Port::D.pin(8).and_pin(9)
+        }
+    };
+
+    sys.gpio_configure_alternate(
+        tx_rx_mask,
+        OutputType::PushPull,
+        Speed::High,
+        Pull::None,
+        Alternate::AF7,
```
cbiffle, line 146: (this AF value is also specific to a subset of UARTs, I believe.)

lib/usart/src/stm32h7.rs
```diff
@@ -0,0 +1,149 @@
...
+            // TODO these are really board configs, not SoC configs!
+            #[cfg(any(feature = "stm32h743", feature = "stm32h753"))]
+            Port::D.pin(8).and_pin(9)
+        }
+    };
+
+    sys.gpio_configure_alternate(
+        tx_rx_mask,
+        OutputType::PushPull,
+        Speed::High,
```
cbiffle, line 144: Please default things to Low.

lib/usart/src/stm32h7.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        let usart = match device {
+            DeviceId::Usart3 => {
+                // From thin air, pluck a pointer to the USART register block.
+                //
+                // Safety: this is needlessly unsafe in the API. The USART is
+                // essentially a static, and we access it through a & reference
+                // so aliasing is not a concern. Were it literally a static, we
+                // could just reference it.
+                #[cfg(any(feature = "stm32h743", feature = "stm32h753"))]
```
cbiffle, line 46: (this cfg should probably get hoisted higher, since if it fails there'll be a weird error message about `()`. Or, removed!)

cbiffle (commented):
The UART driver logic looks good, my comments are mostly structure/API related. We talked in chat about making this crate explicitly stm32h7-specific, which will avoid eliciting comments about ways in which its API might not fit some of our other target parts.

cbiffle:
One other thought ... it would be nice to have a core Usart driver that doesn't include mandatory buffering, in case the user wants a different buffering strategy (or is doing something slow enough that moving single bytes around is fine). Once you turn the FIFO on, this'll become more important, because there's often no need for _two_ levels of buffering.

lib/usart/src/lib.rs
jgallagher: I was on the fence about this. I thought it might be nice to give the user constants to avoid typos, but I'm happy to remove it and take an integer instead.

lib/usart/src/lib.rs
jgallagher: Agreed on the "a little odd" part. I originally had read/write directly on the Usart, but ran into lifetime issues trying to read the contents of the rx buffer in a loop that also pushed things into tx. I'll roll this in with the work to use the FIFO and get rid of the buffering altogether.

lib/usart/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+
+/// Rx buffer owned by a [`Usart`].
+pub struct RxBuf<'a, const N: usize> {
+    usart: &'a Device,
+    buf: &'a mut ArrayVec<[u8; N]>,
+    overrun: &'a mut bool,
+}
+
+impl<'a, const N: usize> RxBuf<'a, N> {
+    /// Drain `self`, returning the bytes it currently contains.
```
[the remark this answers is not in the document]
jgallagher: You didn't miss it. Will roll this in with API changes/removing buffering/using FIFO too.

drv/stm32h7-usart/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Turn on the `USART` described by `usart`, `peripheral`, `tx_rx_mask`,
+    /// and `alternate`, with the baud rate defined by `clock_hz` and
+    /// `baud_rate`.
+    ///
+    /// Enables interrupts from the `USART` when bytes are available to receive;
+    /// the caller is responsible for enabling and handling the corresponding
+    /// kernel interrupt.
+    pub fn turn_on(
+        sys: &Sys,
+        usart: &'static device::usart1::RegisterBlock,
```
andrewjstone, line 41: As far as I can find (in the code in `drv/` at least) `usart1::RegisterBlock` is just a pointer. Where are all the fields and methods coming from?

drv/stm32h7-usart/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+        Self { usart }
+    }
+
+    /// Try to push `byte` into the USART's TX FIFO, returning `true` on success
+    /// or `false` if the FIFO is currently full.
+    pub fn try_tx_push(&self, byte: u8) -> bool {
+        // Check if TX fifo is not full
+        if self.usart.isr.read().txe().bit() {
+            // Stuff byte into TX fifo.
+            self.usart.tdr.write(|w| w.tdr().bits(u16::from(byte)));
```
andrewjstone, line 99: Why are we wrapping a byte in a `u16` here?

andrewjstone (commented):
To my untrained I this looks good. I actually understood the `uartecho` task code which is nice. This change also got me to look through the different registers for the USART that are used in this change in the STM32H7 reference manual. It's the first time in at least 15 years that I've looked at a ref manual like this. Yay embedded!

All of my questions are simply noob questions, but figured it can't help posting publicly for those that are also noobs and interested to follow along.

drv/stm32h7-usart/src/lib.rs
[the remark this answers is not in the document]
jgallagher: This is a quirk of the stm32h7 crate. Not only are the same structure, but `USART3::ptr()` literally returns a `usart1::RegisterBlock`.

jgallagher: Linking to the stm32h7 crate is annoyingly difficult - the docs.rs build is broken, and the repo doesn't have code in it since it's autogenerated. `usart1::RegisterBlock` is a normal Rust struct; the `USART3::ptr()` method gives us a pointer to it, but we then convert that to a `&'static ref`, at which point we can use it like any other ref-to-a-struct. It has fields for all the relevant USART registers:

```rust
#[repr(C)]
pub struct RegisterBlock {
    pub cr1: Reg<CR1_SPEC>,
    pub cr2: Reg<CR2_SPEC>,
    pub cr3: Reg<CR3_SPEC>,
    pub brr: Reg<BRR_SPEC>,
    pub gtpr: Reg<GTPR_SPEC>,
    pub rtor: Reg<RTOR_SPEC>,
    pub rqr: Reg<RQR_SPEC>,
    pub isr: Reg<ISR_SPEC>,
    pub icr: Reg<ICR_SPEC>,
    pub rdr: Reg<RDR_SPEC>,
    pub tdr: Reg<TDR_SPEC>,
    pub presc: Reg<PRESC_SPEC>,
}
```

drv/stm32h7-usart/src/lib.rs
jgallagher: Short answer - that's the type required by `w.tdr().bits(_)`!

Longer answer - it looks like the TDR register accepts 9 bits, but the 9th bit is ignored/replaced by the parity calculation if parity calculations are enabled (which they are by default). I assume the function takes a u16 to account for that.

drv/stm32h7-usart/src/lib.rs
flihp: I found the [stm32h7-nightly](https://github.com/stm32-rs/stm32-rs-nightlies) builds helpful here. I got lost trying to sort out some quirks in stm32h7 related to the RNG and having the generated code (not just the SVDs) was bigly helpful.

PR #478, opened 2022-03-29
Update infrastructure

labbott:


chips/stm32h7.toml
```diff
@@ -96,6 +96,14 @@ size = 0x20000
 address = 0x48021800
 size = 4096
 
+[flash_controller]
+address = 0x52002000
+size = 0x20000
```
mkeeter, line 101: Is this deliberately claiming a larger block than just the flash interface registers?

Screen Shot 2022-03-30 at 9 38 56 AM

drv/stm32h7-update-server/README.mkdn
```diff
@@ -0,0 +1,5 @@
+# Task Template
```
mkeeter, line 1: We can probably delete this (or fill in more relevant info)

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+// heavy work and decision making should be handled in other tasks.
+#![no_std]
+#![no_main]
+
+use drv_update_api::UpdateError;
+use ringbuf::*;
+use stm32h7::stm32h753 as device;
+use userlib::*;
+
+const FLASH_KEY1: u32 = 0x4567_0123;
```
mkeeter, line 17: I'd appreciate a comment here saying these specific values are specified / mandated by ST, not something that we picked ourselves, and citing section 4.5.1 in the reference manual.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+const BANK_ADDR: u32 = 0x08100000;
+const BANK_END: u32 = 0x08200000;
+
+// Writes are indexed by flash words, BANK_ADDR is word 0,
+// BANK_ADDR + FLASH_WORD_BYTES is word 1 etc.
+const BANK_WORD_LIMIT: usize =
+    (BANK_END - BANK_ADDR) as usize / FLASH_WORD_BYTES;
+
+// RM0433 Rev 7 section 4.3.9
+// Flash word is defined as 256 bytes
```
mkeeter, line 32:
```suggestion
// Flash word is defined as 256 bits
```

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// Total length of a word in bytes (i.e. our array size)
+const FLASH_WORD_BYTES: usize = FLASH_WORD_BITS / 8;
+
+// number of 32-bit words we will have to write
+const FLASH_CPU_WORDS: usize = FLASH_WORD_BYTES / 4;
+
+// Block is an abstract concept here. It represents the size of data the
+// driver will process at a time.
+const BLOCK_SIZE: usize = FLASH_WORD_BYTES * 32;
```
mkeeter, line 43: This is a little confusing – maybe define
```rust
const FLASH_WORDS_PER_BLOCK: usize = 32;
```
first, then define `BLOCK_SIZE_BYTES` in terms of it (instead of just `BLOCK_SIZE`, which is ambiguously unitless)?

lib/update/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+
+// Write data to the specified block number. It is expected that the driver
+// has called the start function before this.
+//
+// The data must be at least `block_size` in length.
+pub fn write_block(
+    update_task: TaskId,
+    start_block: usize,
+    data: &[u8],
+) -> Result<(), UpdateError> {
```
jgallagher, line 35: I have a few questions about this API, but for context, do updates write the contiguous blocks `0..sizeof(update)/block_size`, or is it more scattered? I'm assuming it's the former; if that's wrong these questions might make less sense.

1. Does passing a `start_block` of anything other than 0 require the caller to also call `block_size()` first so they can convert whatever offset they have (maybe the number of bytes written so far?) into a block number?
2. Does `data.len()` need to be a multiple of the block size? (If not, why does it need to be at least `block_size` in length?)
3. Should this function return the number of blocks written? Returning a `usize` might be confusing at a glance (is it "number of blocks" or "number of bytes"), but I think that potential confusion is related to all of these questions.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+            word.clone_from_slice(&bytes[byte_offset..byte_offset + 4]);
+
+            core::ptr::write_volatile(
+                (start + (i * 4) as u32) as *mut u32,
+                u32::from_ne_bytes(word),
+            );
+        }
+    }
+
+    poll_flash_done()?;
```
mkeeter, line 188: In `bank_erase`, we logged the end state in the ringbuf regardless of `poll_flash_done` status.  Here, we return early if `poll_flash_done` returns an error, without logging to the ringbuf.

I don't feel too strongly one way or the other, but they should probably be consistent.  Maybe even log `poll_flash_done`'s result in the ringbuf, e.g. 
```rust
let b = poll_flash_done();
ringbuf_entry!(Trace::WriteEnd(b));
b
```

drv/update-api/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#![no_std]
+
+//use core::cell::Cell;
+use userlib::{sys_send, FromPrimitive};
+use zerocopy::{AsBytes, FromBytes};
+
+#[derive(FromPrimitive)]
```
mkeeter, line 11: If you add `derive-idol-err` to dependencies, you can do `#[derive(FromPrimitive, IdolError)]` and skip the manual `From` implementations below!

lib/update/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+//
+// Device independent helpers to do image update
+#![no_std]
+
+pub use drv_update_api::UpdateError;
+use userlib::TaskId;
+
+// Size of processing for the driver. A driver may pass less than this
+// size but it cannot pass greater than this size.
+pub fn block_size(update_task: TaskId) -> Result<usize, UpdateError> {
```
mkeeter, line 13: I'm a little confused why this module needs to exist.  It seems like `block_size` and `start_update` forward stuff straight to the `struct Update` in `update-api`.

Looking at the Hiffy calls below, is the advantage slightly shorter code, e.g.
```rust
    func_err(libupdate::start_update(UPDATE.get_task_id()))?;
```
vs
```rust
    func_err(drv_update_api::Update::from(UPDATE.get_task_id()).start_update())?;
```
?

`write_block` does a little more work, but I wonder if it could be added as an extra `impl` block on the (autogenerated) `struct Update` in `update-api/lib.rs`, e.g.
```rust
impl Update {
    pub fn write_block(start_block: usize, data: &[u8]) -> Result<(), UpdateError> {
        let block_size = update.block_size()?;

        for (i, c) in data.chunks(block_size).enumerate() {
            update.write_one_block(start_block + i, c)?;
        }

        Ok(())
    }
}
```

drv/stm32h7-update-server/src/main.rs
labbott: Hmmm, I don't think that makes it clearer. `FLASH_WORDS_PER_BLOCK` is closer to an internal constant than something that we actually want to change (see where it gets used). `BLOCK_SIZE` is what we would actually want to adjust but I'll see about changing that to `BLOCK_SIZE_BYTES` to indicate the unit.

chips/stm32h7.toml
labbott: I think this is a typo or something else on my end

drv/stm32h7-update-server/src/main.rs
labbott: This is more a side effect of the ringbuf being used for debugging (and wondering what was up with erase at one point). I'll see about making it consistent

lib/update/src/lib.rs
labbott: I think you're right. The extra module is a bit of a historical artifact as I was writing this the wrong way.

lib/update/src/lib.rs
labbott: Correct, the expectation right now is that we'll be writing the blocks contiguously. We don't really have a use for doing scattered.

1. Correct
2. It does not need to be a multiple. It's actually the opposite for the `write_one_block` call, the data can be at most a block size (was there a check elsewhere that had the constraint?). This is a limitation of the current design with leases where we need to copy the data from the other task into a fixed size buffer before we can write it out. There might be some room for improvement.
3. What use would you see in returning the number of block written?

jgallagher: On 3 - Assuming I'm calling this function multiple times to write ultimately-contiguous data, I have to do some bookkeeping to know what `start_block` should be for 2nd-and-later calls; this function could return it directly since it already knows. That said, maybe this doesn't actually help, because if I'm calling this multiple times I already have to know that `data` is an exact multiple of `block_size` (unless it's the final block), right? I'm trying to picture what streaming an update in over UDP looks like; assuming those packets aren't delivering data that's an exact multiple of the block size, the task receiving them is already going to have to know the block size to do some buffering, etc.

I think the thing that's bugging me about this is that it's taking a start block explicitly and a number of blocks implicitly (`data.len() / block_size`, where `block_size` is determined internally by this function). The caller can independently ask what the block size this function is going to use is, but that seems slightly awkward.

labbott: The example in the README was what I had in mind how this would work
```
prep_image_update()
for (i, block) in image_byte_array {
	write_one_block(i, block)
}
finish_image_update()
```
Your example of receiving an update over UDP (which is much closer to how this will actually work) shows the weaknesses of this design. I do think the `write_one_block` which takes a block number and data is the right approach at the lowest level, the question is then what to build on top of it.

I think having a structure that you can feed data without having to worry about the block number would be easier to use. That unfortunately doesn't work with doing things through hiffy because there isn't an actual task where you can keep the structure. Let me give this some thought and see if I can come up with a design/commenting that's easier to use.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+    ringbuf_entry!(Trace::WriteEnd);
+    Ok(())
+}
+
+#[export_name = "main"]
+fn main() -> ! {
+    let mut buffer = [0; 4];
+    let mut flash_page: [u8; BLOCK_SIZE] = [0; BLOCK_SIZE];
+
+    loop {
```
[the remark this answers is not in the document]
labbott: Does idol support server generation? I knew it could generate the API but I didn't realize it had full server support. I'm pro-using idol if it can!

lib/update/src/lib.rs
jgallagher: It also might be perfectly reasonable to not provide a block-number-free abstraction at this point, and let the task receiving an update over UDP worry about it, especially if that's the only place we think it will be used. At that point I might argue for dropping this function altogether and only providing `write_one_block`?

drv/stm32h7-update-server/src/main.rs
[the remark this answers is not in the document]
mkeeter: Yes! Look at [`test-idol-server`](https://github.com/oxidecomputer/hubris/blob/master/test/test-idol-server/src/main.rs) (and associated [`build.rs`](https://github.com/oxidecomputer/hubris/blob/master/test/test-idol-server/build.rs)) for a simple example.

lib/update/src/lib.rs
labbott: based on other suggestions from mkeeter I dropped this library all together and just folded the implementation in hiffy for now. Once we have a few more details about how updates are going to come in we can evaluate what's a good abstraction for state.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+
+enum UpdateState {
+    NoUpdate,
+    InProgress,
+    Finished,
+}
+
+ringbuf!(Trace, 64, Trace::None);
+
+const FLASH_TIMEOUT: usize = 10000;
```
cbiffle, line 70: Nit: this is a retry counter, not a timeout, which I discovered when I went looking for what time unit it's expressed in.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+
+struct ServerImpl<'a> {
+    flash: &'a device::flash::RegisterBlock,
+    state: UpdateState,
+}
+
+impl<'a> ServerImpl<'a> {
+    // See RM0433 Rev 7 section 4.3.13
+    fn swap_banks(&mut self) -> Result<(), RequestError<UpdateError>> {
+        let flash = unsafe { &*device::FLASH::ptr() };
```
cbiffle, line 80: Isn't this equivalent to `self.flash`?

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+                cnt = 0;
+            }
+
+            if seen && !self.flash.bank2().sr.read().qw().bit() {
+                break;
+            } else {
+                cnt += 1;
+            }
+
+            if cnt == FLASH_TIMEOUT {
```
cbiffle, line 119: So, does the hardware specify a maximum programming time? Could we express this as a deadline (in time) that we will retry until? I worry that a canned counter like this could be fragile.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+                break;
+            }
+        }
+
+        if cnt == FLASH_TIMEOUT {
+            return Err(UpdateError::Timeout.into());
+        }
+
+        if self.flash.bank2().sr.read().dbeccerr().bit() {
```
cbiffle, line 128: Nit: Consider reading sr once. It looks like the flash SR doesn't contain any read-to-clear bits but it'd be more obviously correct this way, I think:

`let sr = self.flash.bank2().sr.read();`

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+        if self.flash.bank2().sr.read().strberr().bit() {
+            return Err(UpdateError::StrobeErr.into());
+        }
+
+        if self.flash.bank2().sr.read().pgserr().bit() {
+            return Err(UpdateError::ProgSeqErr.into());
+        }
+
+        if self.flash.bank2().sr.read().wrperr().bit() {
+            return Err(UpdateError::WriteProtErr.into());
```
cbiffle, line 161: I wonder if it's worth having an error type that can describe more than one of these cases. Like, potentially one that contains the literal SR value.)

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+    // Following Single write sequence
+    fn write_word(
+        &mut self,
+        word_number: usize,
+        bytes: &[u8],
+    ) -> Result<(), RequestError<UpdateError>> {
+        ringbuf_entry!(Trace::WriteStart);
+        self.flash.bank2().cr.write(|w| w.pg().set_bit());
+
+        if word_number > BANK_WORD_LIMIT {
```
cbiffle, line 177: Move this check before the side-effecting CR write on line 175. Also, this feels like a panic to me, not an err, since it's a function precondition.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+            return Err(UpdateError::OutOfBounds.into());
+        }
+
+        let start = BANK_ADDR + (word_number * FLASH_WORD_BYTES) as u32;
+
+        if bytes.len() != FLASH_WORD_BYTES {
+            return Err(UpdateError::BadLength.into());
+        }
+
+        for i in 0..FLASH_CPU_WORDS {
```
cbiffle, line 187: (Really the CR write on line 175 should probably move here.)

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+            unsafe {
+                let byte_offset = i * 4;
+                let mut word: [u8; 4] = [0; 4];
+                word.clone_from_slice(&bytes[byte_offset..byte_offset + 4]);
```
cbiffle, line 191: (Nit: could be `copy_from_slice` -- chances are pretty good that it'll generate the same code either way but the use of clone threw me at first)

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+
+                core::ptr::write_volatile(
+                    (start + (i * 4) as u32) as *mut u32,
+                    u32::from_ne_bytes(word),
```
cbiffle, line 195: (Nit: I think `from_le_bytes` would be clearer)

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+        }
+
+        let len = block.len();
+        let mut flash_page: [u8; BLOCK_SIZE_BYTES] = [0; BLOCK_SIZE_BYTES];
+
+        block
+            .read_range(0..len as usize, &mut flash_page)
+            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+
+        if len != BLOCK_SIZE_BYTES {
```
cbiffle, line 279: Okay, I _think_ I see what this is doing, but I think it could use a comment. I suggest

- Changing this check to `<` to make it clear that you're filling in the tail of any too-short block.
- Add a comment stating that.
- Change the while-loop-acting-as-a-for-loop to:

```rust
flash_page[len..].fill(0);
```

With that alternative implementation you can actually lose the check on line 279 if you want.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+        }
+
+        let start = BANK_ADDR + (word_number * FLASH_WORD_BYTES) as u32;
+
+        if bytes.len() != FLASH_WORD_BYTES {
+            return Err(UpdateError::BadLength.into());
+        }
+
+        for i in 0..FLASH_CPU_WORDS {
+            unsafe {
```
cbiffle, line 188: Please narrow the scope of this unsafe to include only the unsafe operation and add a comment explaining why you think it's sound.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+        ringbuf_entry!(Trace::WriteStart);
+        self.flash.bank2().cr.write(|w| w.pg().set_bit());
+
+        if word_number > BANK_WORD_LIMIT {
+            return Err(UpdateError::OutOfBounds.into());
+        }
+
+        let start = BANK_ADDR + (word_number * FLASH_WORD_BYTES) as u32;
+
+        if bytes.len() != FLASH_WORD_BYTES {
```
cbiffle, line 183: (If `bytes` had type `&[u8; FLASH_WORD_BYTES]` you could move this check to compile time.)

drv/stm32h7-update-server/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+            return Err(UpdateError::BadLength.into());
+        }
+
+        for i in 0..FLASH_CPU_WORDS {
+            unsafe {
+                let byte_offset = i * 4;
```
cbiffle, line 189: This could probably be simplified by looping over `bytes.chunks_exact(4)`.

cbiffle (commented):
I agree with the design. I have some comments on part of the implementation.

drv/stm32h7-update-server/src/main.rs
cbiffle: Infinite delays seem ok if the hardware is genuinely hosed, we can always restart the task or reboot to attempt to recover.

This may prove to be famous last words.

drv/stm32h7-update-server/src/main.rs
labbott: So I took a look at this and unfortunately I don't think we can get the full SR value. If we use `idol` we need the error value to be representable as `u16` The bits we care about in SR can only be represented as `u16` (too many for `u8`). I like the idea of doing this long term but for now I think the specific enums are better.

drv/stm32h7-update-server/src/main.rs
labbott: I removed the timeout.

PR #481, opened 2022-03-30
Implement Gimletlet NIC BSP

mkeeter:
This PR implements the BSP for the Gimletlet NIC, which breaks out a KSZ8463.

There are also a few drive-by changes:
- Adding `gpio_init_reset_pulse`, to combine a common series of GPIO operations
- Adding a dedicated error type for KSZ8463 errors (and then using it in the driver)

### DIFF
--- a/drv/ksz8463/src/lib.rs
+++ b/drv/ksz8463/src/lib.rs
@@ -10,6 +10,46 @@ use userlib::hl::sleep_for;
 mod registers;
 pub use registers::{MIBCounter, Register};
 
+////////////////////////////////////////////////////////////////////////////////
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum Error {
+    SpiError(SpiError),
+    WrongChipId(u16),
+}
+
+impl From<SpiError> for Error {
+    fn from(s: SpiError) -> Self {
+        Self::SpiError(s)
+    }
+}
+
+pub enum VLanMode {
+    /// Configure VLAN tags 0x301 and 0x302 for (upstream) ports 1 and 2
+    /// respectively.  Allow untagged frames on any port, but drop tagged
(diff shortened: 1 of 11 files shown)


drv/ksz8463/src/lib.rs
```diff
@@ -19,6 +59,8 @@ enum Trace {
 }
 ringbuf!(Trace, 16, Trace::None);
 
+////////////////////////////////////////////////////////////////////////////////
```
bcantrill, line 62: Out of curiosity (but *actually* out of curiosity!), what do you you use these breaks to denote? In particular would be curious to know if you use this in your workflow by searching forwards or backwards for them?  And obviously feel free to ignore this question entirely. ;)

mkeeter: It's purely visual – I use them to break up logically distinct-ish sections of the file, so I can scan through and see major chunks at a glance.

drv/stm32xx-sys-api/src/lib.rs
```diff
@@ -265,6 +265,33 @@ impl Sys {
...
+    /// - Set the given GPIO pin(s) as low, to avoid glitches when setting it
+    ///   as an output.
+    /// - Configures the given GPIO pin as a push-pull output.  At this point,
+    ///   the pin is actively driven low.
+    /// - Holds it low for `low_time_ms`
+    /// - Brings it high, then waits for `wait_time_ms`
+    pub fn gpio_init_reset_pulse(
+        &self,
+        pinset: PinSet,
+        low_time_ms: u32,
```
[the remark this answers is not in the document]
mkeeter: Yeah, I agree, but not in this PR 😄 

(since the fix would be adding a newtype to `userlib::sleep_for`, which is used _everywhere_!)

PR #486, opened 2022-03-31
Avoid unwrapping while waiting for clock config

mkeeter:
Otherwise, the `net` and `vsc7448` tasks will crash if the sequencer task reboots.

### DIFF
--- a/task/net/src/bsp/gimlet_1.rs
+++ b/task/net/src/bsp/gimlet_1.rs
@@ -45,7 +45,7 @@ pub struct Bsp(mgmt::Bsp);
 pub fn preinit() {
     // Wait for the sequencer to turn on the clock
     let seq = Sequencer::from(SEQ.get_task_id());
-    while seq.is_clock_config_loaded().map(|b| b == 0).unwrap_or(true) {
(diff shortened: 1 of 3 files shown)


task/net/src/bsp/gimlet_1.rs
```diff
@@ -45,7 +45,7 @@ pub struct Bsp(mgmt::Bsp);
 pub fn preinit() {
     // Wait for the sequencer to turn on the clock
     let seq = Sequencer::from(SEQ.get_task_id());
-    while seq.is_clock_config_loaded().unwrap() == 0 {
+    while seq.is_clock_config_loaded().map(|b| b == 0).unwrap_or(true) {
```
cbiffle, line 48: Could also: `.unwrap_or(0) == 0` for slightly less code.

cbiffle (approved):
One suggestion to consider

task/net/src/bsp/gimlet_1.rs
mkeeter: Thanks, I was wondering if there was a cleaner way to write this!

PR #492, opened 2022-04-04
Expand list of uarts/usarts in `chips/stm32h7.toml`

jgallagher:
Also adds the `uartecho` task to `gimletlet`.

Builds on #475; I pointed this PR to that branch to show only the new changes. Should merge the other first.

### DIFF
--- a/task/uartecho/src/main.rs
+++ b/task/uartecho/src/main.rs
@@ -173,19 +173,79 @@ fn configure_uart_device() -> Usart {
 
     const BAUD_RATE: u32 = 115_200;
 
-    // From thin air, pluck a pointer to the USART register block.
-    //
-    // Safety: this is needlessly unsafe in the API. The USART is essentially a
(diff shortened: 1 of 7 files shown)


task/uartecho/src/main.rs
```diff
@@ -173,19 +173,79 @@ fn configure_uart_device() -> Usart {
...
-    //
-    // Safety: this is needlessly unsafe in the API. The USART is essentially a
-    // static, and we access it through a & reference so aliasing is not a
-    // concern. Were it literally a static, we could just reference it.
-    let usart = unsafe { &*device::USART3::ptr() };
+    let usart;
+    let peripheral;
+    let pins;
+
+    cfg_if::cfg_if! {
```
jgallagher, line 180: Is there a cleaner way to list all of these than a big (compile time) if/else chain? Maybe not super relevant since this is a demo task and real uart consumers presumably won't need to do this.

PR #494, opened 2022-04-06
Idol size improvements

cbiffle:
This (combined with a matching change in Idol which I've already snuck onto main) switches servers and clients to using `unwrap_lite` internally, for a substantial text size improvement.


cbiffle:
@mkeeter - indeed we could! Done.

mkeeter (approved):
Looks good!

PR #495, opened 2022-04-06
Fix USART1 pin mapping for tx and rx in uartecho to match gimletlet

andrewjstone:


jgallagher (approved):

andrewjstone:
Big thanks to @jgallagher for finding the correct pin mapping in the datasheet and showing me the how to do the same next time in the datasheet and gimletlet schematic.

PR #498, opened 2022-04-11
Move pin configuration for lpc55 targets into TOML

labbott:
Hubris now supports TOML based configuration. Move the settings
for UART and SPI.

### DIFF
--- a/build/lpc55pins/src/lib.rs
+++ b/build/lpc55pins/src/lib.rs
@@ -1,0 +1,151 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use anyhow::Result;
+use proc_macro2::TokenStream;
+use quote::{format_ident, ToTokens, TokenStreamExt};
+use serde::Deserialize;
+use std::io::Write;
+
+#[derive(Clone, Debug, Deserialize)]
+struct Pin {
+    port: usize,
+    pin: usize,
+}
+
+impl Pin {
+    fn get_port_pin(&self) -> (usize, usize) {
+        assert!(self.pin < 32, "Invalid pin {}", self.pin);
+        assert!(self.port < 2, "Invalid port {}", self.port);
+
+        (self.port, self.pin)
+    }
+}
+
+impl ToTokens for Pin {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let (port, pin) = self.get_port_pin();
+        let final_pin = format_ident!("PIO{}_{}", port, pin);
+        tokens.append_all(quote::quote! {
+            // Yes we want the trailing comma
+            Pin::#final_pin,
+        });
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct PinConfig {
+    pin: Pin,
+    alt: usize,
+    mode: Option<String>,
+    slew: Option<String>,
+    invert: Option<String>,
+    digimode: Option<String>,
+    opendrain: Option<String>,
+    direction: Option<String>,
+}
+
+impl PinConfig {
+    fn get_mode(&self) -> String {
+        match &self.mode {
+            None => "NoPull".to_string(),
+            Some(s) => s.to_string(),
+        }
+    }
+
+    fn get_slew(&self) -> String {
+        match &self.slew {
+            None => "Standard".to_string(),
+            Some(s) => s.to_string(),
+        }
+    }
+
+    fn get_invert(&self) -> String {
+        match &self.invert {
+            None => "Disable".to_string(),
+            Some(s) => s.to_string(),
+        }
+    }
+
+    fn get_digimode(&self) -> String {
+        match &self.digimode {
+            None => "Digital".to_string(),
+            Some(s) => s.to_string(),
+        }
+    }
+
+    fn get_opendrain(&self) -> String {
+        match &self.opendrain {
+            None => "Normal".to_string(),
+            Some(s) => s.to_string(),
+        }
+    }
+
+    fn get_alt(&self) -> usize {
+        if self.alt > 9 {
+            panic!("Invalid alt setting {}", self.alt);
+        }
+
+        self.alt
+    }
+}
+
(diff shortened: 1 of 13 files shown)


build/lpc55pins/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+
+#[derive(Clone, Debug, Deserialize)]
+struct Pin {
+    port: usize,
+    pin: usize,
+}
+
+impl Pin {
+    fn get_port_pin(&self) -> (usize, usize) {
+        if self.pin > 32 {
```
mkeeter, line 19: This could be a one-liner: `assert!(self.pin <= 32, "Invalid pin {}", self.pin)` (same for the port check below)

I'm skeptical that `32` is a valid pin number, so maybe it should be
```
assert!(self.pin < 32, "Invalid pin {}", self.pin)
```

drv/lpc55-spi-server/src/main.rs
```diff
@@ -4,27 +4,10 @@
 
 //! A driver for the LPC55 HighSpeed SPI interface.
 //!
-//! Mostly for demonstration purposes, write is verified read is not
```
mkeeter, line 7: Not exactly related to this PR, but I find myself confused as to what this driver is doing!  Is it actually a SPI server, or still in a development state where it's just sending the same buffers over and over?  A comment to that effect would be helpful!

drv/lpc55-spi-server/src/main.rs
```diff
@@ -160,66 +143,10 @@ fn turn_on_flexcomm(syscon: &Syscon) {
     syscon.leave_reset(Peripheral::Fc3).unwrap_lite();
 }
 
-fn muck_with_gpios(syscon: &Syscon) {
-    syscon.enable_clock(Peripheral::Iocon).unwrap_lite();
```
mkeeter, line 164: These `syscon` calls don't appear to exist in the generated code – is it okay to delete them?

drv/lpc55-spi-server/src/main.rs
```diff
@@ -160,66 +143,10 @@ fn turn_on_flexcomm(syscon: &Syscon) {
...
-
-    for (pin, alt, mode, slew, invert, digi, od) in
-        core::array::IntoIter::new(pin_settings)
-    {
-        iocon
-            .iocon_configure(pin, alt, mode, slew, invert, digi, od)
-            .unwrap_lite();
-    }
+
+    setup_pins(gpio_driver).unwrap_lite();
```
mkeeter, line 149: At this point, might as well delete `muck_with_gpios()` entirely and just put these two lines into `main`?

drv/lpc55-usart/src/main.rs
```diff
@@ -193,33 +192,8 @@ fn turn_on_flexcomm() {
...
-            Pin::PIO0_30,
-            AltFn::Alt1,
-            Mode::NoPull,
-            Slew::Standard,
-            Invert::Disable,
-            Digimode::Digital,
-            Opendrain::Normal,
-        )
-        .unwrap();
+
```
mkeeter, line 195: Same comment applies here (just move these two lines into `main`)

build/lpc55pins/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+        });
+    }
+}
+
+pub fn codegen(pins: Vec<PinConfig>) -> Result<()> {
+    let out_dir = std::env::var("OUT_DIR")?;
+    let dest_path = std::path::Path::new(&out_dir).join("pin_config.rs");
+    let mut file = std::fs::File::create(&dest_path)?;
+
+    writeln!(
```
mkeeter, line 127: I _tried_ to convince `quote` to do this, but did not succeed due to the logic required in the loop!

build/lpc55pins/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+        });
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct PinConfig {
+    pin: Pin,
+    alt: usize,
+    mode: Option<String>,
```
mkeeter, line 47: Handwavey question – would it make sense to use strongly-typed enums here, instead of `String`?

I ask because we're validate the pin / port numbers in this build script, rather than letting them fail when building the application itself, but doing the opposite for these values (where they pass through codegen just fine, but would then fail during the application compilation)..

build/lpc55pins/src/lib.rs
labbott: asserts are usually expensive because of debug formatting so I tend to shy away from using them. You're correct about the pin number though.

build/lpc55pins/src/lib.rs
labbott: I hadn't thought about enums because I didn't know if they were possible. That's a good idea.

drv/lpc55-spi-server/src/main.rs
labbott: This is still in a development state. Once we get the RoT -> SP communication finalized it will certainly be more useful.

drv/lpc55-spi-server/src/main.rs
labbott: Yes, this was leftover from before we had the IOCON as part of the GPIO driver.

build/lpc55pins/src/lib.rs
labbott: After taking a quick look I'm not sure it's worth it. I found this comment in the net build crate:
```
/// TODO: this type really wants to be an enum, but the toml crate's enum
/// handling is really, really fragile, and currently it would be an enum with a
/// single variant anyway.
```
So it sounds like enum handling may be questionable. Even if we did add enum support, we're currently using these with `format_ident` as the name of a field in an existing enum. Maybe there's a better way to use this with `quote` but if not we'd be having to go back and forth to a string anyway. The current failure mode of the existing code will be at compilation time but it should be pretty easy to determine the issue since it will complain about a mismatch of an enum field. If I'm mistaken about the level of annoyance of doing this I'd appreciate pointers.

mkeeter: I haven't tried TOML enums, so probably not worth digging into right now (and failing slightly-later-but-still-at-compile-time is fine by me)

PR #499, opened 2022-04-13
FRU ID driver

mkeeter:
This PR implements a Hubris-level driver for the FRU ID EEPROM, which is an AT24CSW080.

It includes a new driver in the `i2c-devices` crate, plus a minimal `eeprom` task, which just exists to expose it to Hiffy.

There's also a relatively elaborate test suite. The new tests are only built when the `fru-id-eeprom` feature is enabled in `test-suite`; so far, the only test image to do so is the brand new `test/tests-psc/app.toml` image.

The pull request is best reviewed with the [datasheet](https://ww1.microchip.com/downloads/aemdocument/MPD/ProductDocuments/DataSheets/AT24CSW04X-AT24CSW08X-I2C-Compatible-(Two-Wire)-Serial-EEPROM-with-a-Security-Register-and-Software-Write-Protection-4-Kbit-8-Kbit-20006426B.pdf) at your side.


test/test-suite/src/main.rs
```diff
@@ -528,6 +530,272 @@ fn test_idol_ssmarshal() {
...
+    const EEPROM_SIZE: u16 = 1024;
+    const PAGE_SIZE: u16 = 16;
+
+    pub(super) fn test_at24csw080() {
+        let i2c_task = I2C.get_task_id();
+        let dev =
+            At24csw080::new(i2c_config::devices::at24csw080_local(i2c_task)[0]);
+
+        // The FRU ID EEPROM is 1KByte.  Because we want to exercise the
+        // ability to rewrite the EEPROM (and have that persist through power
```
andrewjstone, line 556: Are you testing this by manually power cycling?

andrewjstone (approved):
@mkeeter This looks great. I learned a lot from this review. I actually deleted questions as I continued the review and they became clear.

test/test-suite/src/main.rs
mkeeter: Yup!

PR #502, opened 2022-04-15
Allow Gnarle to decompress chunked input data

arjenroodselaar:
While attempting to decompress a chunked input (a compressed bitstream, sent in blocks between tasks) it was found (thanks @kc8apf!) that run sequences spanning chunks would get mangled and result in an incorrect output. This diff adds two more states to the compressor, allowing it to continue across such chunk boundaries.

This was tested by decompressing a stream using 128 byte buffers for both the input and output.


cbiffle (approved):
Nice catch.

PR #509, opened 2022-04-21
RoT -> SP control via SWD

labbott:


drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,641 @@
...
+        dest: LenLimit<Leased<W, [u8]>, 4096>,
+    ) -> Result<(), RequestError<SpCtrlError>> {
+        ringbuf_entry!(Trace::ReadCmd);
+        if !self.init {
+            return Err(SpCtrlError::NeedInit.into());
+        }
+        let cnt = dest.len();
+        let mut buf = LeaseBufWriter::<_, 32>::from(dest.into_inner());
+
+        for i in 0..cnt {
```
kc8apf, line 195: I'm confused.  Isn't cnt in bytes?  Just below this you treat i as an offset in words.  Then later, the write to buf appears to always write a word.  Won't this attempt to overread by 3x only to fail because the buffer is full?

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,641 @@
...
+    fn reset(&mut self) {
+        // Spec says hold high for 51 clock cycles, more is okay
+        // this gives us 56
+        for _ in 0..7 {
+            self.tx_byte(0xff);
+        }
+    }
+
+    fn idle_cycles(&mut self, cnt: usize) {
+        for _ in 0..(cnt >> 3) {
```
kc8apf, line 368: Took me a second to realize (cnt >> 3) is bits->bytes conversion.  I would have recognized (cnt/8) faster.  If nothing else, a comment about bits->bytes may be warranted.

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,641 @@
...
+
+        self.io_in();
+
+        // We need to read 33 bits total
+        for i in 0..7 {
+            let b = (self.read_nibble().reverse_bits() >> 4) as u32;
+            result |= b << (i * 4);
+        }
+
+        let b = ((self.read_trailing().reverse_bits() >> 3) & 0xF) as u32;
```
kc8apf, line 391: Oooh.  Now I understand why this is called read_trailing().  That wasn't obvious at all at the function definition.

drv/lpc55-swd/src/main.rs
labbott: You are correct. This was a holdover when I initially thought I could treat it as an array of `u32`. The code was just silently returning when this case happened.

drv/lpc55-swd/src/main.rs
labbott: Yes, will update with a comment and also round up just to be cautious.

drv/lpc55-swd/src/main.rs
labbott: I'll change the name. The entire function to could use a few more comments as well.

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+
+#[derive(Copy, Clone, PartialEq)]
+enum Ack {
+    //Ok,
+    Wait,
+    Fault,
+    Protocol,
+}
+
+// Several of these bits in the CSW are  implementation defined according to
```
mkeeter, line 44: Nit: extra space after `are`

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+        RawSwdReg::ApRead(v) => (1 << RDWR_BIT, v.addr_bits() << ADDR_BITS),
+        RawSwdReg::ApWrite(v) => (0 << RDWR_BIT, v.addr_bits() << ADDR_BITS),
+    }
+}
+
+// The parity is only over 4 of the bits
+fn calc_parity(val: u8) -> u8 {
+    let b = val >> 3 & 0xf;
+
+    if b.count_ones() % 2 == 0 {
```
mkeeter, line 169: Could this simply be `(b.count_ones() % 2) << PARITY BIT`?

drv/lpc55-swd/build.rs
```diff
@@ -0,0 +1,89 @@
...
+
+#[derive(Deserialize)]
+struct TaskConfig {
+    in_cfg: Vec<PinConfig>,
+    out_cfg: Vec<PinConfig>,
+    pins: Vec<PinConfig>,
+    spi_num: usize,
+}
+
+fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
```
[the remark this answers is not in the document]
cbiffle: I've done this in a couple cases, but one warning: `quote!` discards comments. So your io_out/io_in comments there would be visible in the code generator, but not the generated code. This may or may not concern you.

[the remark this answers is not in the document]
cbiffle: Oh also the display impl for `quote` that you used there prints all generated code on one line. I've got the `lib/call_rustfmt` crate if you want to fix that.

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+enum Ack {
+    //Ok,
+    Wait,
+    Fault,
+    Protocol,
+}
+
+// Several of these bits in the CSW are  implementation defined according to
+// the official specification but thankfully RM0433 60.4.2 actually gives us the
+// defintions
```
cbiffle, line 46: You reference the ARM ADI docs below -- they're the authoritative definition of the CSW contents for an AHB MEM-AP, not the ST RM, fwiw.

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+enum DpRead {
+    IDCode = 0x0,
+    Ctrl = 0x4,
+    //Resend = 0x8,
+    //Rdbuf = 0xc,
+}
+
+impl DpRead {
+    fn addr_bits(&self) -> u8 {
+        // we need this to be as bits [2:3]
```
cbiffle, line 90: This comment and the implementation of this and DpWrite below are confusing me. Can you expand on it? I think you're deliberately bit-reversing?

drv/lpc55-swd/src/main.rs
cbiffle: Reckon so.

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+        _: &RecvMessage,
+        addr: u32,
+        dest: LenLimit<Leased<W, [u8]>, 4096>,
+    ) -> Result<(), RequestError<SpCtrlError>> {
+        ringbuf_entry!(Trace::ReadCmd);
+        if !self.init {
+            return Err(SpCtrlError::NeedInit.into());
+        }
+        let cnt = dest.len();
+        if cnt % 4 != 0 {
```
cbiffle, line 193: One option to satisfy these properties statically, and make your code below simpler, is to use a lease of `u32` instead of `u8`.

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+            }
+            Err(_) => Err(SpCtrlError::Fault.into()),
+        }
+    }
+}
+
+impl ServerImpl {
+    fn io_out(&mut self) {
+        while !self.spi.mstidle() {
+            cortex_m::asm::nop();
```
cbiffle, line 268: Is there an interrupt you could wait for here? Right now you're starving all tasks at lower and equal priority. (If the answer is no that would be a good thing to put in a comment.)

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+        }
+
+        let gpio_driver = GPIO.get_task_id();
```
cbiffle, line 271: Rather than making a syscall every time to look up the current generation, I recommend creating the `Pins` driver facade _once_ and storing it as a field in your server. It will update its own stored notion of the generation on any failure. This should save text and time.

cbiffle (commented):
This looks pretty darn reasonable. Some comments below asking for more documentation or simplifications.

There's a high-level comment that I think this code needs, which is: a statement of the design intent for (ab)using Flexcomm SPI to do SWD. Ideally this would contain

- A description of what the SWD read and write transactions look like
- A discussion of how 4/8/9/whatever bit SPI transactions are combined to implement this
- How turnaround bits are handled.

drv/lpc55-swd/src/main.rs
labbott: It turns out this doesn't work out of the box:
```
  --> /hubris/target/thumbv8m.main-none-eabihf/release/build/drv-sp-ctrl-api-7da50a9d9a750d44/out/client_stub.rs:61:44
   |
61 |                 userlib::Lease::write_only(arg_sink),
   |                                            ^^^^^^^^ expected `u8`, found `u32`
   |
   = note: expected mutable reference `&mut [u8]`
              found mutable reference `&mut [u32]`
```
Leases work on `[u8]` at the moment. I'm going to leave this as is for now.

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,735 @@
...
+// together and configure exactly one of MOSI or MISO for use with SPI at a time.
+// A side effect of this choice is that reading needs to be precise with the
+// specification. Adding extra idle cycles is fairly easy when writing but that
+// is not possible with reading.
+// - The LPC55 can transmit between 4 and 16 bits at a time. This makes
+// makes transmitting the various combinations of phases and turnaround bits
+// a bit of a pain. This is broken up into the following combinations:
+// -- 8-bit packet write
+// -- 4-bit ACK read (three bits of response + turnaround)
+// -- 34-bit write (32 bits data, one bit partiy, one bit turnaround) broken up
```
cbiffle, line 42: There is no turnaround bit required at the end of a write, this should be 33 bits.

drv/lpc55-swd/src/main.rs
```diff
@@ -0,0 +1,735 @@
...
+//    into 9 bit + 8 bit + 8 bit + 9 bit writes
+// -- 33-bit read (32 bits data, one bit parity) broken up into
```
cbiffle, line 44: We _do_ need a turnaround bit at the end of a read, so, this should be 34.

PR #510, opened 2022-04-21
Implement VLAN support on the H7

mkeeter:
This PR implements support for VLANs on the STM32H7, enabling the feature for
the Gimletlet + NIC hardware (for now).

It's primarily intended to enable the networking described in RFD 250, where
the KSZ8463 + STM32H7 look (to the outside world) like a single device with
two MACs.

# Configuration and build
VLAN support is enabled through an `h7-vlan` feature in the `net` task, and
`vlan` features in all dependent crates. This is checked in the `net-api` build
system, which raises an error if it's not enabled for all relevant crates
(e.g. if you enabled it for the `net` task but forgot to enable it for the
`udpecho` task).

When the feature is enabled, the build system reads a `vlan` dictionary from
`config.net`, which specifies the start VLAN (as a VID) and the total number
of VLANs. This information is used to generate **arrays of arrays** in the
`net` build system; where we would previously make a single array of
`SOCKET_COUNT` items, we're now building a nested array with
`SOCKET_COUNT * VLAN_COUNT` total items.

# Basic architecture
Each VLAN runs an independent instance of `smoltcp` with `SOCKET_COUNT`
independent sockets. These instances are VLAN-unaware; they think that
everything is normal. Instead of owning the `Ethernet` device directly, they
each own a `VLanEthernet` facade. This facade stores a (non-mutable) reference
to the `Ethernet` peripheral and a VID.

## Receive path
VLAN tags are stripped by the STM32H7's ethernet peripheral. The tag's presence
and value (VID) are available in the RDES DMA descriptor. When a `smoltcp`
instance is polled, it calls into its `VLanEthernet` to check if packets are
available. This checks for both an available descriptor _and_ that the stripped
VID matches the `VLanEthernet`'s VID.

Naively, this lead to congestion: if a packet arrives that doesn't match any
VLAN, each instance would refuse to receive it. To avoid this issue, the
check _also_ **discards invalid packets**. Invalid packets include
- Packets that did not have a VID
- Packets with a VID outside of the entire VLAN VID range
- Descriptors with the error flags set
- Descriptors without first and last bits set (`FD` / `LD`)

The `can_recv` function continues to discard invalid packets until either the
ring buffer is empty, or the next packet is valid. Note that it doesn't have
to be valid for _the particular `smoltcp` instance that called `can_recv`_;
as long as it has VID that matches _one_ of the instances, it will be consumed
eventually.

The VLAN receive path should be infallible once `can_recv` returns true. Clever
lifetime tricks using the `RxToken` ensure that no one else can take packets
while it's held by the interface, meaning the validity of `can_recv` should
not change.

(This isn't 100% foolproof, because someone else could call into the `Ethernet`
peripheral directly, but it's relatively easy to audit: `can_recv` is only
called by `VLanEthernet::receive`, and `vlan_recv` is only called by
`VLanRxToken::consume`)

This also led me to discover a potential issue in the existing receive path:
if the first and only receive descriptor is an error descriptor, `can_recv`
will return `true`, but `RxRing::try_with_next` will skip that packet and
return `None`, which means `Ethernet::try_recv` will _also_ return `None`,
which will panic in `OurRxToken::consume`.

(I haven't fixed this yet; I'm not sure how to trigger it, so it's hard to
tell whether it's a true issue)

## Transmit path
When transmitting in VLAN mode, we use DMA _context descriptors_ to set the
VLAN tag, which is then automatically injected into packets. For simplicity,
we _always_ set the VLAN tag with a context descriptor before writing the
transmit descriptor. This doubles the size of our Tx descriptor ring, since
each item is now `[[u32; 4]; 2]` instead of `[u32; 4]`.

Other than that, the Tx path is largely the same.

## Task API
To tasks seeking to communicate on the network, the API is _nearly_ the same.
They still use `send_packet` and `recv_packet` IPC calls to the `net` API, and
use the same socket numbering as before.

However, the `struct UdpMetadata` used in these calls has an extra field when
the `vlan` feature is enabled: `vid: u16`.  This `vid` directs the packet
in `send_packet` to a particular VLAN, and is returned in `recv_packet` so
that the caller can know the source VLAN of a packet.

If the task is only replying to packets (i.e. `udpecho`), this is invisible:
the task simply includes the source `meta` in its reply, so the reply goes
out to the same VLAN.

If a task is _generating_ packets independently, it has to think a little more
about where the packets are going, which is good and intentional given our
system design.


task/net/src/vlan.rs
```diff
@@ -0,0 +1,407 @@
...
+
+        // We're iterating over a bunch of things together.  The standard
+        // library doesn't have a great multi-element zip, so we'll just
+        // manually use mutable iterators.
+        let mut neighbor_cache_iter = storage.neighbor_cache_storage.iter_mut();
+        let mut socket_storage_iter = storage.socket_storage.iter_mut();
+        let mut socket_handles_iter = socket_handles.iter_mut();
+        let mut vid_iter = generated::VLAN_RANGE;
+        let mut ifaces_iter = ifaces.iter_mut();
+        let mut ip_addr_slice = &mut storage.ipv6_net[..];
```
jgallagher, line 150: I think this could be

```rust
let mut ip_addr_iter = storage.ipv6_net.chunks_mut(1);
```

to allow it to be an iterator like the others instead of having to manually carve it up via `.split_at_mut(1)`.

task/net/src/vlan.rs
```diff
@@ -0,0 +1,407 @@
...
+            .ok_or(RequestError::Fail(ClientError::BadMessageContents))
+            .and_then(|s| {
+                s.get(index)
+                    .cloned()
+                    .ok_or(RequestError::Fail(ClientError::BadMessageContents))
+            })
+    }
+
+    /// Gets the socket `index`. If `index` is out of range, returns
+    /// `BadMessage`.
```
jgallagher, line 259: Is it worth noting that an out of bounds `vlan_index` will panic? I wouldn't normally suggest that since this is a private function, but since the docstring explicitly states what happens with an out of bounds `index` it might be worth it.

task/net/src/vlan.rs
```diff
@@ -0,0 +1,407 @@
...
+        }
+
+        // Convert from absolute VID to an index in our VLAN array
+        if !VLAN_RANGE.contains(&metadata.vid) {
+            return Err(NetError::InvalidVLan.into());
+        }
+        let vlan_index = metadata
+            .vid
+            .checked_sub(VLAN_RANGE.start)
+            .ok_or(NetError::InvalidVLan)?;
```
jgallagher, line 349: Could this be a normal subtraction instead, given the `VLAN_RANGE.contains(..)` check that precedes it?

jgallagher:
This all looks awesome. I wanted to clarify something in your description:

> This also led me to discover a potential issue in the existing receive path:
if the first and only receive descriptor is an error descriptor, can_recv
will return true, but RxRing::try_with_next will skip that packet and
return None, which means Ethernet::try_recv will also return None,
which will panic in OurRxToken::consume.
> 
> (I haven't fixed this yet; I'm not sure how to trigger it, so it's hard to
tell whether it's a true issue)

It looks like the VLAN path is not subject to this issue because `vlan_is_next_free()` itself discards error descriptors, meaning it won't return `true` for them, right? Putting in a fix without being able to trigger the problem feels icky, but would the fix be to make the normal path look more like the VLAN path (where `can_recv()` loops until it finds a valid packet or the end, and `try_recv()` is infallible once `can_recv()` returns)?

build/net/src/lib.rs
```diff
@@ -18,6 +18,11 @@ pub struct GlobalConfig {
 pub struct NetConfig {
     /// Sockets known to the system, indexed by name.
     pub sockets: BTreeMap<String, SocketConfig>,
+
+    /// VLAN configuration, or None. This is checked against enabled features
+    /// during the `net` build, so it must be present iff the `vlan` feature
+    /// is turned on.
+    pub vlan: Option<VLanConfig>,
```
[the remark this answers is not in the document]
mkeeter: This is an interesting idea!  I tested it out, and have ended up _slightly in favor_ of the current setup.

Making the `vlan` member only exist if the feature is enabled means replacing a bunch of the `Option`-based code in `net/build.rs` with `#[cfg(...)]` blocks.

I'm iffy about this, because it creates code paths that are _conditionally_ compiled. When I tested it, I could only check by building a VLAN-enabled image, then building a no-VLAN image, and often found that I had broken one or the other.  (In general, I see conditional compilation as a necessary evil, but would rather keep all the code building all the time).

It was also more code compared to the `Option`-based version!

drv/stm32h7-eth/src/lib.rs
```diff
@@ -378,6 +345,115 @@ impl Ethernet {
...
+        core::sync::atomic::fence(core::sync::atomic::Ordering::Release);
+        self.dma.dmacrx_dtpr.write(|w| unsafe {
+            w.rdt().bits(self.rx_ring.tail_ptr() as u32 >> 2)
+        });
+        result
+    }
+
+    /// Checks whether the next slot on the Rx buffer is owned by userspace
+    /// and has a matching VLAN id. Packets without a VID or with a
+    /// non-matching VID are dropped by the Rx ring during this function
```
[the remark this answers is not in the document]
mkeeter: Yes, clarified the comment

drv/stm32h7-eth/src/ring.rs
```diff
@@ -213,6 +255,94 @@ impl TxRing {
...
+            // also free.
+            let buffer = self.buffers[self.next.get()].0.get();
+            // Safety: we're dereferencing a raw *mut to get a &mut into the
+            // buffer. We must ensure that the pointer is valid (which we can
+            // tell trivially from the fact that we got it from an UnsafeCell)
+            // and that there is no aliasing. We can say there's no aliasing
+            // because the descriptor is free (above), so we're about to produce
+            // the sole reference to it, which won't outlive this block.
+            let buffer = unsafe { &mut *buffer };
+            let buffer = &mut buffer[..len];
```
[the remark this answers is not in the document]
mkeeter: I sure hope so!  This is copied from the existing code, so I'm not totally sure what upholds that invariant – @cbiffle any hints?

task/net/src/vlan.rs
mkeeter: Nice!

task/net/src/vlan.rs
mkeeter: Sure, added a comment

task/net/src/vlan.rs
mkeeter: Yes, good catch – this `checked_sub` predated `VLAN_RANGE`

mkeeter:
Yup, that's exactly right.

I spent a while this morning trying to send a packet that was messed up enough to give a DMA error descriptor, but without success – I think I _might_ be able to trigger it with a jumbo packet, but having trouble convincing my hardware to do that...

drv/stm32h7-eth/src/ring.rs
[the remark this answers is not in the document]
cbiffle: Hmmmm so in the send path, the only thing that "ensures" this is that the network stack shouldn't be generating packets larger than the MTU (buffers are the size of the MTU). However, I think it's conceivable that some interaction bug could cause this line to panic.

Panicking might be appropriate -- there's really no runtime way to handle this condition -- but is heavy-handed and would be annoying to debug. We might want to rejigger this so we can return an error / drop the packet instead.

Not necessarily in this PR.

app/gimletlet/app.toml
```diff
@@ -161,10 +161,10 @@ task-slots = ["sys"]
 [tasks.net]
 path = "../../task/net"
 name = "task-net"
-stacksize = 3800
+stacksize = 8192
```
cbiffle, line 164: Dang, that's a lot of stack -- how close are we to using all of that, or is this padding left over from development? (i.e. what does `humility stackmargin` tell you)

build/net/src/lib.rs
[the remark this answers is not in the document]
cbiffle: The down side of this approach is that the vlan code is likely getting included in the netstack text for non-vlan targets. That said, our main focus is the vlan targets right now, so we can fix that if it becomes a problem.

build/net/src/lib.rs
```diff
@@ -45,7 +58,35 @@ pub struct TaskNote {
 }
 
 pub fn load_net_config() -> Result<NetConfig, Box<dyn std::error::Error>> {
-    Ok(build_util::config::<GlobalConfig>()?.net)
+    let cfg = build_util::config::<GlobalConfig>()?.net;
+
+    match (cfg!(feature = "vlan"), cfg.vlan.is_some()) {
```
cbiffle, line 63: Are you sure this check is working? I thought you had to detect features from build scripts using the `CARGO_FEATURE_VLAN` (in this case) environment variable.

build/net/src/lib.rs
```diff
@@ -45,7 +58,35 @@ pub struct TaskNote {
...
+pub fn generate_vlan_consts(
+    config: &NetConfig,
+    mut out: impl std::io::Write,
+) -> Result<(), std::io::Error> {
+    let vlan = config.vlan.unwrap();
+    let end = vlan.start + vlan.count;
+    writeln!(
+        out,
+        "
+pub const VLAN_RANGE: core::ops::Range<u16> = {:#x}..{:#x};
```
cbiffle, line 85: Would we ever want to be able to include `0xFFFF` here? If so, you may wish to use a `RangeInclusive`.

cbiffle (approved):

build/net/src/lib.rs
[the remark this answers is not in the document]
mkeeter: This is in the build script 😄 

(and I'm pretty sure the embedded side is doing the right thing)

build/net/src/lib.rs
mkeeter: Nope, the VID is actually only 12 bits. I just added checks + masking to ignore the upper 4 bits.

PR #511, opened 2022-04-22
Add drv/{fpga-api, fpga-devices, fpga-server}

arjenroodselaar:
This diff is a first pass for a somewhat generic FPGA driver and server which
allows other tasks in the system to manage actions such as loading bitstreams.
In addition the server task exposes a register interface abstraction for a SPI
peripheral running as part of the bitstream application.

These two  combined interfaces if you will, allow a process such as a sequencer
to manage both the device and interact with application logic through a single
entry point.

The driver is generic enough that in addition to the ECP5 implementation found
in this diff, we can support the iCE40 and possibly future devices.

The intend is for the FPGA server to not use a separate SPI driver if not required
reduce system resources by absorbing the functionality needed to handle the
SPI peripheral. This work requires some refactoring to expose more of the SPI
server functionality as a library and this will be done in a follow-up.

Note that the ECP5 driver is split into a top and bottom implementation.
Currently only one bottom implementation is provided, but a second (involving
transparent use of a SPI mux and GPIO extend) is in the works.

Finally, Hiffy/Humility support for interacting with an FPGA will be provided in
a follow up diff.


drv/fpga-api/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+        u16::from(e) as u32
+    }
+}
+
+impl core::convert::TryFrom<u16> for FpgaError {
+    type Error = ();
+
+    fn try_from(v: u16) -> Result<Self, Self::Error> {
+        match v & 0xff00 {
+            0x0100 => Ok(FpgaError::ImplError((v & 0x00ff) as u8)),
```
mkeeter, line 54: Style nit: the mask is unnecessary, since it's a truncating conversion, but `v as u8` doesn't make it obvious – maybe `u8::from(v)`?

drv/fpga-api/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+            _ => Err(()),
```
mkeeter, line 55: What about the other errors in the enum?

(If there's a reason we shouldn't convert them, there should be a comment here for clarity)

drv/fpga-api/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+//! API crate for the FPGA server.
+
+#![no_std]
+
+use drv_spi_api::SpiError;
+use userlib::*;
+use zerocopy::AsBytes;
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[repr(u16)]
```
mkeeter, line 14: What does this `#[repr(u16)]` do?

[Testing on Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2024a0df8df09c9621e3c645b180ef40), I find that it increases the size of the type from 2 bytes to 4 bytes.

drv/fpga-api/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+            _ => Err(()),
+        }
+    }
+}
+
+impl core::convert::TryFrom<u32> for FpgaError {
+    type Error = ();
+
+    fn try_from(v: u32) -> Result<Self, Self::Error> {
+        Self::try_from(v as u16)
```
mkeeter, line 64: Nit to correctly handle large values, rather than masking upper bits:
```suggestion
        let v: u16 = v.try_into().map_err(|_| ())?;
        Self::try_from(v)
```

drv/fpga-api/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+#[repr(u8)]
+pub enum BitstreamType {
+    Uncompressed = 0,
+    Compressed = 1,
+}
+
+#[derive(Copy, Clone, Debug, FromPrimitive, PartialEq, AsBytes)]
+#[repr(u8)]
+pub enum WriteOp {
+    Write = 0,
```
mkeeter, line 88: Nit: Why is there a gap here?

drv/fpga-api/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+impl From<WriteOp> for u8 {
+    fn from(op: WriteOp) -> Self {
+        op as u8
+    }
+}
+
+include!(concat!(env!("OUT_DIR"), "/client_stub.rs"));
+
+impl Fpga {
+    pub fn application_read8(
```
mkeeter, line 102: Non-mandatory suggestion:

If you wanted to be more flexible, you could write something like
```rust
pub fn application_read<T>(&self, addr: impl Into<u16>) -> Result<T, FpgaError>
    where T: AsBytes + Default
{
    let mut v = T::default();
    self.application_read_raw(addr.into(), v.as_bytes_mut())?;
    Ok(v)
}
```

(renaming the IDL `application_read` function to `application_read_raw`)

The same is applicable to generalize `application_write`.

idl/fpga.idol
```diff
@@ -0,0 +1,133 @@
+// FPGA server IPC interface
+
+Interface(
+    name: "Fpga",
+    ops: {
+        "device_enabled": (
+            doc: "Return true if the FPGA is enabled, false otherwise",
+            reply: Result(
+                ok: "u8",
```
mkeeter, line 9: Idol supports `bool` variables (as of https://github.com/oxidecomputer/idolatry/pull/14); this also applies to `set_device_enable` / `application_enabled` / `set_application_enabled`

drv/fpga-server/build.rs
```diff
@@ -0,0 +1,17 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//use std::{env, fs, path::PathBuf};
```
mkeeter, line 5: Nit: delete commented-out line

drv/fpga-api/src/lib.rs
arjenroodselaar: Ah yes, I like the use of `from` here.

drv/fpga-api/src/lib.rs
arjenroodselaar: We should convert them. This went through a few iterations and I missed this as it took its final form. Thanks!

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+        if #[cfg(target_board = "sidecar-1")] {
+            let driver = Ecp5UsingSpi {
+                sys,
+                spi,
+                done: sys_api::Port::J.pin(15),
+                init_n: sys_api::Port::J.pin(12),
+                program_n: sys_api::Port::J.pin(13),
+                design_reset_n: sys_api::Port::J.pin(14),
+            };
+            driver.configure_gpio();
```
mkeeter, line 48: Nit: these two lines (and their repetition below) could be pulled out of the `cfg_if!` block, since they're common to both boards.

drv/fpga-api/src/lib.rs
arjenroodselaar: Because the RTL implementation uses a more generic command type which has `Read` with value 1. But because the API is already split in read/write I wanted an Op type which wouldn't accidentally let you do the wrong thing.

drv/fpga-api/src/lib.rs
arjenroodselaar: Yes, this is a better solution. Thanks!

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+    buffer: [u8; 128],
+    decompressor: Option<gnarle::Decompressor>,
+}
+
+type RequestError = idol_runtime::RequestError<FpgaError>;
+type ReadDataLease = LenLimit<Leased<R, [u8]>, 128>;
+type WriteDataLease = LenLimit<Leased<W, [u8]>, 128>;
+
+impl<FpgaT: Fpga> idl::InOrderFpgaImpl for ServerImpl<FpgaT> {
+    fn device_enabled(&mut self, _: &RecvMessage) -> Result<u8, RequestError> {
```
mkeeter, line 99: I commented on the Idol file as well, but we should use `bool` instead of `u8` for the `*_enabled` / `set_*_enable` functions.

(and if that doesn't work, let me know, since that would be an Idolatry bug!)

drv/fpga-api/src/lib.rs
arjenroodselaar: `From<u16> for u8` doesn't exist (and for good reason). But `v as u8` does what I intend here without the explicit mask.

drv/fpga-api/src/lib.rs
arjenroodselaar: Ah also left over from an earlier iteration, removed. Thanks!

drv/fpga-api/src/lib.rs
arjenroodselaar: Yes, this is a better way to do this. Thanks!

drv/fpga-server/src/main.rs
arjenroodselaar: No, that seems to work as intended and makes this whole interface less awkward. Thanks!

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn set_device_enable(
+        &mut self,
+        _: &RecvMessage,
+        enabled: u8,
+    ) -> Result<(), RequestError> {
+        Ok(self.device.set_device_enable(enabled != 0)?)
+    }
+
+    fn reset_device(&mut self, _: &RecvMessage) -> Result<(), RequestError> {
+        Ok(self.device.reset_device(self.device_reset_ticks)?)
```
mkeeter, line 112: The `Ok(function()?)` pattern is kinda weird, but I now understand what it's doing.

The alternative would be something like
```
self.device.reset_device(self.device_reset_ticks).map_err(Into::into)
```
which is slightly longer, but _maybe_ more obvious?

(I don't feel too strongly about this, but wanted to propose it as an option)

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+            .reset_application(self.application_reset_ticks)?)
+    }
+
+    fn start_bitstream_load(
+        &mut self,
+        _: &RecvMessage,
+        bitstream_type: BitstreamType,
+    ) -> Result<(), RequestError> {
+        if let BitstreamType::Compressed = bitstream_type {
+            self.decompressor = Some(gnarle::Decompressor::default())
```
mkeeter, line 156: Should we check if `self.decompression` is `None` here, which would indicate that we're mid-load (and something is probably wrong)?

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        let chunk = &mut &self.buffer[..data.len()];
+        let mut decompress_buffer = [0; 256];
+
+        match self.decompressor.as_mut() {
+            Some(decompressor) => {
+                while !chunk.is_empty() {
+                    let decompressed_chunk = gnarle::decompress(
+                        decompressor,
+                        chunk,
```
mkeeter, line 179: It's not immediately obvious that this modifies `chunk`. Maybe don't take the reference above, so you can pass `&mut chunk` here?

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+    ) -> Result<(), RequestError> {
+        let header = ApplicationRequestHeader {
+            cmd: 0x1,
+            addr: U16::new(addr),
+        };
+
+        self.application
+            .lock(spi_api::CsState::Asserted)
+            .map_err(FpgaError::from)?;
+        self.application
```
mkeeter, line 217: If these functions fail, do we still need to call `self.application.release()`?

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+        } else {
+            compile_error!("Board is not supported by the task/fpga");
+        }
+    }
+
+    let mut incoming = [0u8; idl::INCOMING_SIZE];
+    let mut server = ServerImpl {
+        device: ecp5,
+        device_reset_ticks: ecp5::DEVICE_RESET_DURATION,
+        application: Spi::from(SPI.get_task_id()).device(1),
```
mkeeter, line 67: `application` is a confusing name for the SPI peripheral, I'd normally pick something like `spi`.

drv/fpga-api/src/lib.rs
mkeeter: Oh yeah, that makes sense 😅 

`v as u8` sounds good

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,453 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use super::Fpga;
```
mkeeter, line 5: I normally only see `use super` for in-line modules that want to use code from their parent context; it's more common to `use crate` to include things from the crate root.

```suggestion
use crate::Fpga
```

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,453 @@
...
+use bitfield::bitfield;
+use drv_fpga_api::{DeviceState, FpgaError};
+use ringbuf::*;
+use userlib::*;
+use zerocopy::{AsBytes, FromBytes};
+
+/// ECP5 IDCODE values.
+#[derive(Copy, Clone, Debug, FromPrimitive, PartialEq, AsBytes)]
+#[repr(u8)]
```
mkeeter, line 14: Why not make this a `#[repr(u32)]` using the values (`0x21111043`, etc) below, so you get `FromPrimitive` / `ToPrimitive` for free?

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// Send a `Command` and read back two bytes of data.
+    pub fn read16(&self, c: Command) -> Result<u16, ImplT::Error> {
+        Ok(u16::from_be(self.read(c)?))
+    }
+
+    /// Send a `Command` and read back four bytes of data.
+    pub fn read32(&self, c: Command) -> Result<u32, ImplT::Error> {
+        let v = u32::from_be(self.read(c)?);
+
+        match c {
```
mkeeter, line 250: Nit: I'd rather have this `ringbuf` logging in `status` and `device_id`, since `read32`'s name implies that it's low-level and agnostic to the value of `c`

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// Wait for the DONE flag to go high.
+    pub fn await_done(&self, sleep_ticks: u64) -> Result<(), ImplT::Error> {
+        while !self.status()?.done() {
+            hl::sleep_for(sleep_ticks);
+        }
+        Ok(())
+    }
+}
+
+impl<ImplT: Ecp5Impl> From<ImplT> for Ecp5<ImplT> {
```
mkeeter, line 298: This seems weird to me – are there places where we want to use `from` / `into` instead of the plain `Ecp5(...)` constructor?

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,453 @@
...
+        self.set_device_enable(false)?;
+        hl::sleep_for(ticks);
+        self.set_device_enable(true)?;
+        Ok(())
+    }
+
+    fn device_state(&self) -> Result<DeviceState, FpgaError> {
+        if !self.0.program_n()? {
+            Ok(DeviceState::Disabled)
+        } else {
```
mkeeter, line 339: Clippy will probably complain about this, since you can rewrite it as an `else if` chain

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,453 @@
...
+        Ok(self.read32(Command::ReadId)?)
+    }
+
+    fn start_bitstream_load(&mut self) -> Result<(), FpgaError> {
+        // Put device in configuration mode if required.
+        if !self.status()?.write_enabled() {
+            self.enable_configuration_mode()?;
+        }
+
+        if !self.status()?.write_enabled() {
```
mkeeter, line 362: Should this block be inside the upper conditional, since it's checking to make sure `enable_configuration_mode` actually worked?

drv/fpga-server/src/main.rs
arjenroodselaar: The bitstream start/continue/finish sequence should probably be implemented as https://cliffle.com/blog/rust-typestate/ in order to force that sequence. The typestate object can then hold the decompressor and the finish step will automatically discard it.

drv/fpga-devices/src/ecp5.rs
arjenroodselaar: Thanks, this cleans up nicely.

drv/fpga-api/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+            FpgaError::InvalidValue => 0x0301,
+            FpgaError::PortDisabled => 0x0400,
+            FpgaError::NotLocked => 0x0500,
+        }
+    }
+}
+
+impl From<SpiError> for FpgaError {
+    fn from(e: SpiError) -> Self {
+        FpgaError::ImplError(u32::from(e) as u8)
```
mkeeter, line 42: Could this be `u8::from(e)`?

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,496 @@
...
+        self.lock = None
+    }
+}
+
+impl<Driver: Ecp5Driver> FpgaBitstream for Ecp5Bitstream<'_, Driver>
+where
+    FpgaError: From<<Driver as Ecp5Driver>::Error>,
+{
+    fn continue_load(&mut self, buf: &[u8]) -> Result<(), FpgaError> {
+        self.lock.as_ref().unwrap().0.configuration_write(buf)?;
```
mkeeter, line 406: Nit: it would be nice to return `InvalidState` instead of unwrapping here.

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,496 @@
...
+        ringbuf_entry!(Trace::WriteBitstreamChunk);
+        Ok(())
+    }
+
+    fn finish_load(&mut self) -> Result<(), FpgaError> {
+        // Release the locked resources in the driver.
+        self.lock = None;
```
mkeeter, line 413: Should we return an error if `!self.lock.is_none()`?

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,496 @@
...
+    /// The following primitives allow the upper half of the driver to issue
+    /// atomic commands.
+    ///
+    /// If no lock control of the medium is needed these can be implemented as
+    /// no-op.
+    fn configuration_lock(&self) -> Result<(), Self::Error>;
+    fn configuration_release(&self) -> Result<(), Self::Error>;
+
+    /// User design reset.
+    fn user_design_reset_n(&self) -> Result<bool, Self::Error>;
```
mkeeter, line 178: There's a lot of overlap between this trait and `FpgaUserDesign` below.

Could we instead define the trait as
```rust
pub trait Ecp5Driver : FpgaUserDesign {
    ...
```
and remove some duplication (since that would force anyone implementing `Ecp5Driver` to also implement `FpgaUserDesign`)?

drv/fpga-api/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub fn continue_load(&mut self, data: &[u8]) -> Result<(), FpgaError> {
+        (*self.0).continue_bitstream_load(data)
+    }
+
+    pub fn finish_load(&mut self) -> Result<(), FpgaError> {
+        (*self.0).finish_bitstream_load()
+    }
+}
+
+pub struct FpgaUserDesign(idl::Fpga);
```
mkeeter, line 184: Nit: It's a little confusing that this is both a `struct` in `fpga-api` and a `trait` in `fpga-driver`...

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,380 @@
...
+
+    fn closed_recv_fail(&mut self) {
+        // Welp, someone had asked us to lock and then died. Release the
+        // lock and any resources acquired from the device driver.
+        self.lock_holder = None;
+        self.bitstream_loader = None;
+    }
+
+    fn lock(&mut self, msg: &userlib::RecvMessage) -> Result<(), RequestError> {
+        if let Some(task) = self.lock_holder {
```
mkeeter, line 147: Do we actually expect tasks to lock the server multiple times? It seems more intuitive to return an error if the `lock()` function is called when the server is already locked.

drv/fpga-api/src/lib.rs
mkeeter: Never mind, that doesn't work!

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,380 @@
...
+    fn continue_bitstream_load(
+        &mut self,
+        _: &RecvMessage,
+        data: LenLimit<Leased<R, [u8]>, 128>,
+    ) -> Result<(), RequestError> {
+        data.read_range(0..data.len(), &mut self.buffer[..data.len()])
+            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+
+        let mut chunk = &self.buffer[..data.len()];
+        let mut decompress_buffer = [0; 512];
```
mkeeter, line 263: Nite: this only needs to be scoped to the block below, e.g.
```rust
Some(BitstreamLoader::Compressed(decompressor, bitstream, len)) => {
    let mut decompress_buffer = [0; 512]; // moved here
    ...
}
```

mkeeter (approved):

drv/fpga-server/src/main.rs
arjenroodselaar: This has been implemented using the type state pattern and the server will return an `InvalidState` error if a client were to call these functions out of order.

drv/fpga-server/src/main.rs
arjenroodselaar: This has been resolved using `Drop` types.

drv/fpga-server/src/main.rs
arjenroodselaar: This has been renamed to `user_design` throughout the diff.

drv/fpga-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+        data.read_range(0..data.len(), &mut self.buffer[..data.len()])
+            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+
+        let header = ApplicationRequestHeader {
+            cmd: u8::from(op),
+            addr: U16::new(addr),
+        };
+
+        self.application
+            .lock(spi_api::CsState::Asserted)
```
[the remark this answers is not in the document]
arjenroodselaar: This has been refactored using `Drop` types.

drv/fpga-devices/src/ecp5.rs
arjenroodselaar: This was removed.

drv/fpga-devices/src/ecp5.rs
```diff
@@ -0,0 +1,496 @@
...
+
+        if status.encrypt_preamble_detected() {
+            ringbuf_entry!(Trace::EncryptedBitstreamDetected);
+        }
+        if status.standard_preamble_detected() {
+            ringbuf_entry!(Trace::StandardBitstreamDetected);
+        }
+
+        let error = status.bitstream_error();
+        if error != BitstreamError::None {
```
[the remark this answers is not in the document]
arjenroodselaar: Yup, this now closer to other device APIs. Thanks!

drv/fpga-server/src/main.rs
arjenroodselaar: I added an error for this.

drv/fpga-devices/src/ecp5.rs
arjenroodselaar: Yes, I added an early return here, because that would mean either a code bug or the user tried to call this out of order (or twice, or both).

PR #512, opened 2022-04-22
Use # to prepend `0x` to hex numbers

mkeeter:


build/xtask/src/dist.rs
```diff
@@ -84,7 +84,7 @@ pub fn package(
         }
     }
     for (name, range) in &memories {
-        println!("{:<5} = 0x{:0>8x}..0x{:0>8x}", name, range.start, range.end);
+        println!("{:<5} = {:0>#8x}..{:0>#8x}", name, range.start, range.end);
```
jgallagher, line 87: I don't think this one works:

```rust
    let x = 12345;
    println!("{:0>#8x}", x);
```

prints

```
000x3039
```

jgallagher: Looks like `{:0>#010x}` works though

build/xtask/src/dist.rs
```diff
@@ -924,7 +924,7 @@ fn generate_task_linker_script(
     fn emit(linkscr: &mut File, sec: &str, o: u32, l: u32) -> Result<()> {
         writeln!(
             linkscr,
-            "{} (rwx) : ORIGIN = 0x{:08x}, LENGTH = 0x{:08x}",
+            "{} (rwx) : ORIGIN = {:#08x}, LENGTH = {:#08x}",
```
jgallagher, line 927:
```suggestion
            "{} (rwx) : ORIGIN = {:#010x}, LENGTH = {:#010x}",
```

jgallagher: I'll stop repeating this, but all the ones with explicit field widths need 2 added to them; I think the field width is including the `0x` in its calculation after the change to use `#`.

build/xtask/src/dist.rs
mkeeter: Now that I think about it, I don't think the `0>` is even necessary, since we're using `0` padding, which necessarily goes onto the left; it seems like `{:#010x}` should work fine.

mkeeter:
Thanks, I fixed up all the padding discrepancies!

PR #514, opened 2022-04-26
lpc55-pac: Bump version from 0.3.0 to 0.4.

flihp:
After sending a patch upstream to the lpc55-pac adding a missing feature yesterday I took a swing at upgrading to the latest version of this crate. This was an attempt to determine how painful an upgrade will be if lpc55-pac merges my PR and we then upgrade to benefit. Going from 0.3 to 0.4 was thankfully trivial.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2078,9 +2078,9 @@ dependencies = [
 
 [[package]]
 name = "lpc55-pac"
-version = "0.3.0"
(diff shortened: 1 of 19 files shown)


flihp:
breakage here seems to be caused by new crates added to hubris since this PR that need to be updated to use the newer pac crate as well

kc8apf (approved):

PR #518, opened 2022-04-29
Add very basic kernel profiling support.

cbiffle:
More details in the comments on sys/kern/src/profiling.rs.

PR #519, opened 2022-05-02
Use perfect hashing for IRQ tables

mkeeter:
In our existing code, we search through all of `HUBRIS_INTERRUPTS` in both the generic IRQ handler and the `IrqControl` syscall. This is `O(n)` with the number of interrupts, and there's a `TODO` indicating that we could do something more clever.

This PR does something more clever!

It generates a [perfect hash table](https://en.wikipedia.org/wiki/Perfect_hash_function) at compile time, then uses that table to do lookups in `O(1)` time.

This is a _dumb_ perfect hash table with a single stage of `index = ((a * b) & 0xFFFFFFFF) % n`, and including empty slots when necessary.

I did some performance testing by rebasing onto #518 and tracking
- time spent in the `IrqControl` syscall
- time spent in the generic IRQ handler
while handling a ping flood: `sudo ping6 -f fe80::c1d:9aff:fe64:b8c2%en0`

Data was captured with a Saleae and post-processed using something like [this script](https://gist.github.com/mkeeter/981cca838fed3b7d909336689b069888).

I captured data on an STM32H753 Nucleo board under two conditions:
- Stock firmware, which has 6 IRQs in the table
- Modified `app.toml` to include 12 IRQs in total, which matches Gimlet and Sidecar

Here are some performance numbers, recorded for a 1-second span.  Times in nanoseconds, in the form mean ± standard deviation (count)

|| 6 IRQs | | 12 IRQs | |
|-|-|-|-|-|
|| Linear | Perfect | Linear | Perfect |
|`IrqControl`|592 ± 44 (418)|557 ± 75 (906)|667 ±49 (758)|558 ± 74 (813)|
|ISR handler|426 ± 76 (620)|394 ±62 (1288)| 467 ± 70 (1081)|382 ±61 (1119)|

I'm not sure why – in the 6 IRQ case – the number of interrupt invocations is much higher for the perfect hashing implementation.  (this could also be a red herring, since `ping6 -f` is probably dependent on OS / network load?)

Other than that one weird observation, this looks like a solid win. In a system with 12 IRQs in its table, this is a ~20% speedup!

Remaining questions:
- [ ] This probably won't work on the G0, which lack the hardware support for efficient remainder calculations. How should we handle this (feature in `kern`, something different in codegen, something else?).  It's possible to iterate over the `PerfectHashMap` just like before, so that's an easy option.
- [ ] Does removing `HUBRIS_INTERRUPTS` break `humility`?
- [ ] The single-layer hash table scales poorly. Should I add a fallback to a dual-level table (e.g. from [this commit in the branch](https://github.com/oxidecomputer/hubris/blob/e7bdbe06db9c29b5814a7702a4abdda2ebcc37a0/lib/phash/src/lib.rs#L32-L45))
- [ ] Should we use an OTS crate instead, e.g. [`phf`](https://crates.io/crates/phf)?


cbiffle:
> * This probably won't work on the G0

I believe it will _work_ but may not show an equivalent speedup. I have a box of G0 hardware here and I'm happy to benchmark it later this afternoon.

cbiffle:
For posterity -- I benchmarked it and it slow. :-) This increases interrupt processing costs by about 3x on M0 targets with >1 IRQ in use, because it calls into an expensive and fully general software div/mod routine. Which is a little surprising, tbh, since it's mod-by-a-constant -- I thought compilers used to optimize such expressions using CORDIC or similar. So, @mkeeter and I talked about the possibility of conditionally selecting the IRQ dispatch implementation based on your target architecture.

mkeeter:
4a24cf46ab713099dd45f2737416e322e229e1c6 switch back to linear search for `armv6`!

lib/phash/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    /// Looks up a value in the table by key, using a linear search.
+    ///
+    /// This is slower than [Self::get] in most cases, but is faster for small
+    /// tables on a system without hardware division.
+    #[inline(always)]
+    pub fn get_linear(&self, key: K) -> Option<&V> {
+        self.values.iter().find(|v| v.0 == key).map(|v| &v.1)
```
cbiffle, line 47: TIL `find_map`!

```rust
self.values.iter().find_map(|(k, v)| if k == key { Some(v) } else {None })
```

Not 100% sure that's an improvement, but, I didn't actually realize it was there.

PR #520, opened 2022-05-03
Allow humility xtask to set the humility path via HUMILITY_PATH

aphistic:
I don't have Humility installed on my `$PATH` at the moment, and I anticipate needing to iterate on both Hubris and Humility at the same time, so I added a way to configure the path! I made it so if you provide `HUMILITY_PATH` in the environment, the `cargo xtask humility` command will use that as the path to the Humility binary instead of just `"humility"`.

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -461,7 +461,7 @@ ID ADDR     TASK               GEN STATE
  6 200003a8 idle                 0 Healthy(Runnable)         
 ```
 
-If the `humility` binary is not available on your `$PATH`, the `HUMILITY_PATH` environment
+If the `humility` binary is not available on your `$PATH`, the `HUBRIS_HUMILITY_PATH` environment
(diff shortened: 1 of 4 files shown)


bcantrill (commented):
This looks good!  My only comment would be the name of the environment variable:  we have generally used `HUMILITY_` to denote Humility environment variables and `HUBRIS_` to denote Hubris environment variables.  This is a confluence of both of them, of course, but it's ultimately a Hubris environment variable -- and I would recommend `HUBRIS_HUMILITY_PATH`.

aphistic:
Ahh, that makes sense! I'll get that updated tonight. :)

aphistic:
As I was trying to run the tests tonight I saw the `test` and `flash` xtasks also use Humility in the same way, so I added the same env var to both of those as well. I felt they still fit within the spirit of the PR.

PR #526, opened 2022-05-06
Gimlet build ergonomics improvements for Rev B

cbiffle:
- The FPGA image filename is no longer hardcoded and is now set in the configuration, which allows rev A and B to reference different bitstreams.
- Same for the register layout JSON file, so they can diverge _if they need to._
- Build output is now called `gimlet-a` or `gimlet-b` instead of just `gimlet` because the latter was hella confusing.


bcantrill (approved):
Looks great -- changing the image name is a very good idea. Our future selves will never know the pain we saved them!

PR #529, opened 2022-05-06
gimlet: implement board rev detection.

cbiffle:
This can now distinguish Rev A from Rev B, and (with any luck) both from
any future revs.

- gimlet-a firmware will now refuse to boot on rev B or C+
- gimlet-b firmware will now refuse to boot on rev A or C+
- Board mismatch is detected while all pins are still tristate, not even
  the crystal is started.

### DIFF
--- a/app/gimlet/src/main.rs
+++ b/app/gimlet/src/main.rs
@@ -36,47 +36,121 @@ fn main() -> ! {
 }
 
 fn system_init() {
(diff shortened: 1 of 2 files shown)


cbiffle:
FYI I have verified that this (1) boots on the correct rev and (2) does not boot on the incorrect rev

bcantrill (approved):
This looks good -- but I'm very concerned that I will now have "What Does the Fox Say" stuck in my head for the rest of the day.

PR #534, opened 2022-05-08
Add TMP451 driver

mkeeter:
This implements a small driver for the TMP451 and adds it to the `task-thermal` sensor list


bcantrill (approved):
Looks great -- thank you!

PR #538, opened 2022-05-09
Initial rp2040 support.

cbiffle:
This adds support for one of the two cores. Note that you can't actually use this to flash it; I'm currently feeding the eval board UF2 files. The upstream `elf2uf2` tool is buggy and won't accept our ELF files, so, I wrote a new one. Basically, don't get too excited yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -518,6 +518,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "76fbd10dce159c002b9c688ae8ab7cd531151e185e0ad360f4bfea3b0eede3a8"
 
 [[package]]
+name = "demo-pi-pico"
+version = "0.1.0"
+dependencies = [
+ "build-util",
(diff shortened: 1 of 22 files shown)


moosingin3space:
Hi, I actually attempted this a few months ago as an attempt to learn Hubris! Some useful things I discovered:

1. `probe-rs` supports the RP2040 "normally" -- via an external JTAG/SWD probe. I used a CMSIS-DAP with both a `defmt`/RTIC demo application and with my (not-really-working) RP2040 port of Hubris.
2. I was able to use RTT (via the `defmt`/`rtt_target` crates) in order to log messages to the host. I noticed Hubris has ITM support for the system log, would an RTT backend be useful? Could possibly integrate it with Humility (via the `probe-rs` libraries) to have support for RP2040 debugging.

mchodzikiewicz:
This is also what my thinking was reading the docs - why no `rtt` when there is such an excellent support from knurling.rs and one pin less to be used

PR #541, opened 2022-05-09
xtask: include kernel config in image ID

cbiffle:
Noticed that Humility was saying I already had an image flashed when I
changed the start flag on a task. This fixes that.

The main thing left out of image ID generation now is the kernel build
itself, because we have a causality problem (the kernel build includes
the image ID). I'll fix this later.


bcantrill (approved):
Looks good!

PR #542, opened 2022-05-09
Add temperature sensing to DIMMs

mkeeter:
This adds temperature sensing for all of the DIMMs in Gimlet, based on the TSE2004av standard.

(this is orthogonal to the existing `spd` task, because the temperature sensors have their own addresses and register sets, and are actively polled during operation)

### DIFF
--- a/app/gimlet/rev-a.toml
+++ b/app/gimlet/rev-a.toml
@@ -551,145 +551,161 @@ refdes = "U431"
 bus = "mid"
 address = 0x18
 device = "tse2004av"
-name = "A0"
+name = "DIMM_A0"
 description = "DIMM A0"
 sensors = { temperature = 1 }
 refdes = "M0"
+removable = true
 
 [[config.i2c.devices]]
 bus = "mid"
 address = 0x19
 device = "tse2004av"
-name = "A1"
+name = "DIMM_A1"
 description = "DIMM A1"
 sensors = { temperature = 1 }
 refdes = "M8"
+removable = true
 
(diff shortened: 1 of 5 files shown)


app/gimlet/rev-a.toml
```diff
@@ -545,6 +545,153 @@ pmbus = { rails = [ "V12_SYS_A2" ] }
 sensors = { temperature = 1, power = 1, voltage = 1, current = 1 }
 refdes = "U431"
 
+################################################################################
+# DIMM slots
+[[config.i2c.devices]]
```
bcantrill, line 550: I think on these we want to indicate that `removable = true` -- which should allow `humility validate` to do the right thing when they're gone.

bcantrill (approved):
This looks awesome! Seems like it came together really smoothly -- thanks for doing this!

drv/i2c-devices/src/tse2004av.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+pub struct Tse2004Av {
+    device: I2cDevice,
+}
+
+impl core::fmt::Display for Tse2004Av {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "tmp451: {}", &self.device)
```
rmustacc, line 44: Is the use of TMP451 intentional here?

mkeeter: nope, good catch

PR #543, opened 2022-05-10
Linker fill

labbott:
Pad out flash with a known value. This is useful for measurement testing and we're going to want to do this anyway.


build/kernel-link.x
```diff
@@ -131,15 +131,21 @@ SECTIONS
...
     __veneer_base = .;
     *(.gnu.sgstubs*)
     . = ALIGN(32);
     __veneer_limit = .;
   } > FLASH
 
+  /*
+   * Fill the remaining flash space with a known value
+   */
+  .fill : ALIGN(1) {
```
cbiffle, line 145: FWIW, the ALIGN(1) here shouldn't be necessary.

build/task-link.x
```diff
@@ -55,6 +55,14 @@ SECTIONS
     __edata = .;
   } > RAM
 
+  /*
+   * Fill the remaining flash space with a known value
+   */
+  .fill (LOADADDR(.data) + SIZEOF(.data)) : AT(LOADADDR(.data) +  SIZEOF(.data)) {
+    . = .;
+    . += ((ORIGIN(FLASH) + LENGTH(FLASH)) - ABSOLUTE(.));
```
cbiffle, line 63: I think this is equivalent to the (simpler, imo) code:
```
. = ORIGIN(FLASH) + LENGTH(FLASH);
```

cbiffle (approved):
One simplification for your consideration.

PR #548, opened 2022-05-12
Swd transaction

labbott:


drv/lpc55-swd/src/main.rs
```diff
@@ -74,6 +74,7 @@ enum Trace {
     WriteCmd,
     None,
     AckErr(Ack),
+    ParityFail(u32, u16),
```
cbiffle, line 77: Speaking as a person who may want to debug this later, could you name these fields? From the code I _think_ that these would be reasonable names:
```rust
ParityFail {
    data: u32,
    received_parity: bool,
}
```
(also does `parity` need to be a u16?)

drv/lpc55-swd/src/main.rs
```diff
@@ -543,7 +608,11 @@ impl ServerImpl {
             // for the required turnaround bit!
             self.swd_finish();
 
-            return Ok(ret);
+            if let Some(v) = ret {
```
cbiffle, line 611: FWIW this is equivalent to
```rust
ret.ok_or(Ack::Fault)
```

cbiffle (approved):
High-level thought: since this is an IPC server, it might make sense to use closed receive once a read transaction has started so that you only accept calls from the one client task. Similar to the SPI lock API.

I don't really expect this to _stay_ an IPC server (afaict it's mostly a server so we can poke it from Hiffy) so I don't feel super strongly about this.

PR #555, opened 2022-05-17
Use FPGA server to load Sidecar Mainboard bitstream, sequencer improvements

arjenroodselaar:
This diff does the following:

- Use the FPGA machinary to load the mainboard controller bitstream on boot
- Add an API for the various functional units in the mainboard controller
- Use the above to implement Tofino sequencer improvements (slightly more
  control, better visbility when faults occur)
- Small tweaks in the Idol interface of the sequencer server to pick up recent
  Hubris improvements
- Implement a rudimentary Sidecar mainboard I2C emulator, allowing the sequencer
  to run on a Gimletlet connected to a ECP5 EVN board


mkeeter:
Is it intentional to commit `drv/sidecar-mainboard-controller-api/sidecar_mainboard_emulator_ecp5_evn.bit.old`?

drv/sidecar-mainboard-controller-api/src/tofino2.rs
```diff
@@ -0,0 +1,143 @@
...
+    A2 = 1,
+    A0 = 2,
+    InPowerUp = 3,
+    InPowerDown = 4,
+}
+
+#[derive(Copy, Clone, PartialEq, FromPrimitive, AsBytes)]
+#[repr(u8)]
+pub enum TofinoSeqError {
+    None = 0,
```
mkeeter, line 23: You can remove `None` and use `Option<TofinoSeqError>` below; Rust is smart enough to still pack it into a single byte ([test](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ccef8f80f580dd619f81bb0f80261cf7))

drv/sidecar-mainboard-controller-api/src/tofino2.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    pub fn clear_error(&self) -> Result<(), FpgaError> {
+        self.write_ctrl(WriteOp::BitSet, Reg::TOFINO_SEQ_CTRL::CLEAR_ERROR)
+    }
+
+    pub fn enabled(&self) -> Result<bool, FpgaError> {
+        Ok(
+            self.read_masked(Addr::TOFINO_SEQ_CTRL, Reg::TOFINO_SEQ_CTRL::EN)?
+                != 0,
+        )
```
mkeeter, line 84: Small suggestion:
```suggestion
        self.read_masked(Addr::TOFINO_SEQ_CTRL, Reg::TOFINO_SEQ_CTRL::EN)
            .map(|v| v != 0)
```
is a little shorter, up to you whether you think it's more clear

drv/sidecar-mainboard-controller-api/src/tofino2.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    pub fn power_status(&self) -> Result<u32, FpgaError> {
+        self.fpga.read(Addr::TOFINO_POWER_ENABLE)
+    }
+
+    pub fn vid(&self) -> Result<Option<Tofino2Vid>, FpgaError> {
+        let mask =
+            Reg::TOFINO_POWER_VID::VID_VALID | Reg::TOFINO_POWER_VID::VID;
+        let v = self.read_masked(Addr::TOFINO_POWER_VID, mask)?;
```
mkeeter, line 123: Nit: constructing `mask` and calling `read_masked` isn't necessary here, since you're re-applying the `VID_VALID` and `VID` masks below.

drv/sidecar-mainboard-controller-api/src/tofino2.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        if (v & Reg::TOFINO_POWER_VID::VID_VALID) != 0 {
+            match Tofino2Vid::from_u8(v & Reg::TOFINO_POWER_VID::VID) {
+                None => Err(FpgaError::InvalidValue),
+                some_vid => Ok(some_vid),
+            }
+        } else {
+            Ok(None)
```
mkeeter, line 131: It's not intuitive that without the `VID_VALID` set, we returns `Ok(None)` here.

Maybe add a docstring for this function, or simplify the return type to just `Result<Tofino2Vid, FpgaError>` (and also return`Err(InvalidValid)` if the `VID_VALID` bit isn't set)?

drv/sidecar-mainboard-i2c-emulator/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+//! An I2C device emulator for Sidecar Mainboard, intended to convince the
+//! sequencer task it is running on a Sidecar Mainboard rather than say a
+//! Gimletlet.
+
+#![no_std]
+#![no_main]
+
+use drv_i2c_api::*;
+use ringbuf::*;
+use userlib::*; //{hl, FromPrimitive};
```
mkeeter, line 14: Not sure what the comment is doing here.

drv/sidecar-seq-server/src/main.rs
```diff
@@ -24,177 +29,335 @@ use i2c_config::devices;
...
+        match vid {
+            Tofino2Vid::V0P922 => set_vout(Volts(0.922)),
+            Tofino2Vid::V0P893 => set_vout(Volts(0.893)),
+            Tofino2Vid::V0P867 => set_vout(Volts(0.867)),
+            Tofino2Vid::V0P847 => set_vout(Volts(0.847)),
+            Tofino2Vid::V0P831 => set_vout(Volts(0.831)),
+            Tofino2Vid::V0P815 => set_vout(Volts(0.815)),
+            Tofino2Vid::V0P790 => set_vout(Volts(0.790)),
+            Tofino2Vid::V0P759 => set_vout(Volts(0.759)),
+        }
```
mkeeter, line 113: Nit: this is a little over-complicated, how about
```suggestion
        let value = Volts(match vid {
            Tofino2Vid::V0P922 => 0.922,
            Tofino2Vid::V0P893 => 0.893,
            Tofino2Vid::V0P867 => 0.867,
            Tofino2Vid::V0P847 => 0.847,
            Tofino2Vid::V0P831 => 0.831,
            Tofino2Vid::V0P815 => 0.815,
            Tofino2Vid::V0P790 => 0.790,
            Tofino2Vid::V0P759 => 0.759,
        });
        self.vddcore
            .set_vout(value)
            .map_err(|_| SeqError::SetVddCoreVoutFailed)?;

        ringbuf_entry!(Trace::SetVddCoreVout(value));
        Ok(())
```

drv/sidecar-seq-server/src/main.rs
```diff
@@ -24,177 +29,335 @@ use i2c_config::devices;
...
+
+        Err(SeqError::SequencerTimeout)
+    }
+
+    fn power_down(&mut self) -> Result<(), SeqError> {
+        ringbuf_entry!(Trace::InitiateTofinoPowerDown);
+        Ok(self
+            .sequencer
+            .set_enable(false)
+            .map_err(|_| SeqError::SequencerError)?)
```
mkeeter, line 154:
```suggestion
        self.sequencer
            .set_enable(false)
            .map_err(|_| SeqError::SequencerError)
```

drv/sidecar-seq-server/src/main.rs
```diff
@@ -24,177 +29,335 @@ use i2c_config::devices;
...
+            (TofinoSequencerPolicy::Disabled, TofinoSeqState::A0, _) => {
+                self.power_down()
+            }
+            // Power up
+            (
+                TofinoSequencerPolicy::LatchOffOnFault,
+                TofinoSeqState::A2,
+                TofinoSeqError::None,
+            ) => self.power_up(),
+            _ => Ok(()), // Do nothing by default.
```
mkeeter, line 177: Why doesn't `RestartOnFault, A2, None` cause the system to power up?

For that matter, why doesn't `RestartOnFault, A2, Err(...)` cause the system to restart?

drv/sidecar-seq-server/src/main.rs
```diff
@@ -24,177 +29,335 @@ use i2c_config::devices;
...
+    }
+
+    fn clear_tofino_seq_error(
         &mut self,
         _: &RecvMessage,
-        state: PowerState,
     ) -> Result<(), RequestError<SeqError>> {
-        ringbuf_entry!(Trace::SetState(self.state, state));
+        ringbuf_entry!(Trace::ClearingTofinoSequencerFault(
+            self.tofino.sequencer.error().map_err(SeqError::from)?
```
mkeeter, line 225: Do we want to return early if logging `error()` fails, or just if `clear_error` fails below?

drv/sidecar-seq-server/src/main.rs
```diff
@@ -24,177 +29,335 @@ use i2c_config::devices;
...
-        }
+    fn tofino_power_status(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<u32, RequestError<SeqError>> {
+        Ok(self
+            .tofino
+            .sequencer
+            .power_status()
+            .map_err(SeqError::from)?)
```
mkeeter, line 242:
```suggestion
        self.tofino.sequencer.power_status().map_err(SeqError::from)
```

drv/sidecar-seq-server/src/payload.rs
```diff
@@ -7,69 +7,48 @@
...
     // Unknown write to 0xc171
-    func(&[
-        0x71, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-    ])?;
+    func(&[0x71, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])?;
 
     // Unknown write to 0xc180
     func(&[
-        0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
```
mkeeter, line 46: There's lots of free real estate for `gnarle` here, but that doesn't need to happen right away.

task/net/src/bsp/sidecar_1.rs
```diff
@@ -45,7 +45,7 @@ pub struct Bsp(mgmt::Bsp);
 pub fn preinit() {
     // Wait for the sequencer to turn on the clock
     let seq = Sequencer::from(SEQ.get_task_id());
-    while seq.is_clock_config_loaded().unwrap_or(0) == 0 {
+    while seq.is_clock_config_loaded().unwrap_or(false) {
```
mkeeter, line 48: Thanks for cleaning this up (I think it predated Idolatry supporting `bool` 😆)

mkeeter:
This generally looks good, with a few comments.

It would also be good to run it through clippy (`cargo xtask clippy app/sidecar/app.toml ecp5_mainboard sequencer`) and see if there's anything worth fixing in there.

app/gimletlet/app-sidecar-emulator.toml
```diff
@@ -0,0 +1,258 @@
+name = "gimletlet-sidecar-emulator"
+target = "thumbv7em-none-eabihf"
+board = "gimletlet-2"
+chip = "../../chips/stm32h7.toml"
```
mkeeter, line 4: Now that #539 is merged, you'll need to edit this line to read
```suggestion
chip = "../../chips/stm32h7"
```
(i.e. no trailing `.toml`)

drv/sidecar-mainboard-controller-api/src/tofino2.rs
arjenroodselaar: The sequencer server wants to pass `TofinoSeqError` though an Idol interface. The `None` variant lets me do this without much fuss. Can `Option<TofinoSeqError>` likewise pass through Idol without much additional work?

drv/sidecar-seq-server/src/main.rs
arjenroodselaar: This is significantly cleaner. Thanks!

drv/sidecar-mainboard-controller-api/src/tofino2.rs
arjenroodselaar: The short answer is that `None` is a valid answer when Tofino is not yet powered up far enough that the value is valid. But the state of the system is valid at that point, so I would not want to make that an error. I added a docstring.

drv/sidecar-seq-server/src/payload.rs
arjenroodselaar: Yes, we should do this soon.

drv/sidecar-seq-server/src/main.rs
```diff
@@ -24,177 +29,335 @@ use i2c_config::devices;
...
+        let start = sys_get_timer().now;
+
+        if let Err(e) = self.tofino.handle_tick() {
+            ringbuf_entry!(Trace::TofinoSequencerError(e));
+        }
+
+        let finish = sys_get_timer().now;
+
+        // We now know when we were notified and when any work was completed.
+        // The assumption here is that `start` < `finish`, with the this won't
```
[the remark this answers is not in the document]
arjenroodselaar: Fixed :$

PR #564, opened 2022-05-23
First draft of thermal control loop

mkeeter:
This is a head-to-toes reworking of the existing `thermal` task, which was
previously described as a "primordial control loop".

The new control loop is in [`tasks/thermal/src/control.rs`](https://github.com/oxidecomputer/hubris/blob/thermal-control-loop/task/thermal/src/control.rs).
In general, it expects the system to have some number of sensors which are used
to control fan speed, and some which are just logged.  Temperature sensors are
associated with a particular power state (e.g. `A0`), and have per-sensor
margins above which they adjust fan speed.

I've finally found a board support package (BSP) pattern that I like:
- There's a top-level `BspT` trait, which defines functions which every BSP
  must implement.
- Every BSP `mod` is declared in the top-level `bsp.rs`, so they're all compiled
- A dummy `const _: () = { ... }` is used to check that every BSP type actually
  implements the `BspT` trait, which simultaneously silences warnings about
  dead code.  (This is the new trick!)
- Based on the `target_board`, we conditionally import one or the other module
  (this is the same as previous `bsp.rs`, e.g. the one in `net`)

After this PR, it is still not a _production-ready_ control loop: the control
system is effectively an integral-only controller, so it's not stable.

However, all of the infrastructure is in place, and the default behavior is
identical to before (just run the fans at a low duty cycle); the new control
loop must be selected with a Hiffy command before it begins running.

As such, I think it's reasonable to merge this in now.  It's a large PR that
breaks every time anything changes, and rebasing it is a pain.

There are also a few drive-by changes:
- Some of the I2C devices included a local cache of configuration values.  I've
  updated those caches to use interior mutability, so that the `read` and
  `write` functions can take `&self` instead of `&mut self`.  Requiring
  `&mut self` was cascading through the entire codebase and making it hard to
  store arrays of thermal sensors.
- In that same vein, many of those I2C devices were not _invalidating_ their
  local caches, so the values would never change, even if the user modified
  the calibration.  This is now fixed.
- There's now an `enable_watchdog` and `disable_watchdog` Hiffy function
  attached to the `thermal` task, so that you can disable the fan watchdog
  timer before taking a `humility dump` (otherwise, it will trigger and spin
  the fans up to 100%).


task/spd/src/main.rs
```diff
@@ -380,17 +380,7 @@ fn main() -> ! {
...
-                    // when performing a read at offset 0xff, the next read
-                    // should be at offset 0x00), but Rust (rightfully) isn't
-                    // so into that -- so unwrap what we're doing.
-                    //
-                    voffs[ndx] = if voffs[ndx] == u8::MAX {
-                        0
-                    } else {
-                        voffs[ndx] + 1
-                    };
+                    voffs[ndx] = voffs[ndx].wrapping_add(1);
```
bcantrill, line 383: I think it might be worth retaining the comment (or some fraction of it?) to explain why this is a `wrapping_add`?

task/thermal/src/control.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /// An extremely simple thermal control loop.
+    ///
+    /// Returns an error if the control loop failed to read critical sensors;
+    /// the caller should set us to some kind of fail-safe mode if this
+    /// occurs.
+    pub fn run_control(&mut self) -> Result<(), ThermalError> {
+        // Ignore errors and missing readings
+        // TODO: handle this better
```
bcantrill, line 273: Are these being ignored? It looks like they are being propagated, leaving it to someone else to ignore them. ;)

bcantrill (approved):
Looks great -- and I think disabling the watchdog by default per Robert's suggestion for development is a good idea.

task/spd/src/main.rs
mkeeter: Good idea, done

task/thermal/src/control.rs
mkeeter: Yup, removed comment

PR #566, opened 2022-05-23
Kernel spring cleaning

cbiffle:
This is a collection of cleanups I made last week. You can review it as a giant multipurpose diff, if you're into that, but it's probably best read as a series of individual commits. (I could create a separate PR for each but oh god the paperwork and conflicts.)

- Switches the whole kernel to `forbid(unsafe_op_in_unsafe_fn)`, meaning that all unsafe operations will now need to be marked and justified even if they're in an `unsafe fn`.
- Simplifies kernel startup by removing old app table validation code, which has been superfluous for several months.
- Simplifies syscall argument handling by removing the arg proxy abstraction that _seemed like a good idea at the time_ but really wasn't
- Remove vestigial app module
- Remove nearly-vestigial `klog!` facility
- Improve docs


sys/kern/src/arch/arm_m.rs
```diff
@@ -914,9 +889,9 @@ pub unsafe extern "C" fn SVCall() {
                 ",
                 exc_return = const EXC_RETURN_CONST as u32,
```
mkeeter, line 890: Nit: indentation is a little weird now that the top `asm!("` has shifted right.  Sadly, it won't be autofixed by `rustfmt` because it's within the body of the `cfg_if!` macro.

sys/kern/src/arch/arm_m.rs
```diff
@@ -1312,8 +1319,9 @@ unsafe extern "C" fn configurable_fault() {
         @ resume
         bx lr
         ",
```
mkeeter, line 1321: ...and here

sys/kern/src/arch/arm_m.rs
```diff
@@ -1349,8 +1366,9 @@ pub unsafe extern "C" fn UsageFault() {
 #[naked]
 #[cfg(armv6m)]
 pub unsafe extern "C" fn HardFault() {
-    asm!(
-        "
+    unsafe {
+        asm!(
+            "
```
mkeeter, line 1371: Final case of weird indentation!

sys/kern/src/syscalls.rs
```diff
@@ -60,19 +66,25 @@ pub unsafe extern "C" fn syscall_entry(nr: u32, task: *mut Task) {
     arch::with_task_table(|tasks| {
         // Work out the task index based on the pointer into the task table
         // slice. We could store the index *and* the pointer in globals,
-        // avoiding this divde, but divides are pretty cheap....
+        // avoiding this divide, but divides are pretty cheap....
```
mkeeter, line 69: Not required for this PR, but what would be the advantage of storing the index in globals, vs including the index in the `struct Task` itself (and populating it when building the task table in kernel startup)?

mkeeter (approved):
A few comments (mostly indentation) and questions, but generally looks good!

PR #568, opened 2022-05-25
Improve `cargo xtask dist`

mkeeter:
This is a large refactoring of `dist.rs`, focusing specifically on breaking `package(...)` from a single 600-line function into something more manageable.

In general, I split subsections of `package(...)` into functions (to isolate their inputs and outputs), then collected repeated input arguments into a new `struct PackageConfig`.

The resulting build should be the same!

Other drive-by changes:
- Clippy fixes to make `cargo clippy -p xtask` _mostly_ clean
- Change signing method to be an `enum` instead of a `String`
- Add warnings if multiple tasks are at supervisor priority (0), or _no_ tasks are at priority 0
- Make `cargo xtask build` and `cargo xtask clippy` work on the kernel (!)
- Remove `supervisor.notification` configuration parameter, and update the docs to say it's always bit 0
- Remove `info/map.txt` and `info/allocations.txt` from the app bundle, since they were wrong and unused

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -94,9 +94,7 @@ impl PackageConfig {
         // path for its --remap-path-prefix argument, so we use
         // `dunce::canonicalize` instead
         let cargo_home = dunce::canonicalize(std::env::var("CARGO_HOME")?)?;
-        let mut cargo_git = cargo_home.clone();
-        cargo_git.push("git");
-        cargo_git.push("checkouts");
+        let cargo_git = cargo_home.join("git").join("checkouts");
         remap_paths.insert(cargo_git, "/git");
 
         // This hash is canonical-ish: Cargo tries hard not to change it
@@ -105,10 +103,10 @@ impl PackageConfig {
         // It depends on system architecture, so this won't work on (for example)
         // a Raspberry Pi, but the only downside is that panic messages will
         // be longer.
-        let mut cargo_registry = cargo_home;
-        cargo_registry.push("registry");
-        cargo_registry.push("src");
-        cargo_registry.push("github.com-1ecc6299db9ec823");
+        let cargo_registry = cargo_home
+            .join("registry")
+            .join("src")
+            .join("github.com-1ecc6299db9ec823");
         remap_paths.insert(cargo_registry, "/crates.io");
 
         let mut hubris_dir =
@@ -1076,11 +1074,11 @@ fn build(
         bail!("command failed, see output for details");
     }
 
-    cargo_out.push(build_config.out_path);
+    let out_file = cargo_out.join(build_config.out_path);
 
     let dest = cfg.dist_file(name);
-    println!("{} -> {}", cargo_out.display(), dest.display());
-    std::fs::copy(&cargo_out, dest)?;
+    println!("{} -> {}", out_file.display(), dest.display());
+    std::fs::copy(&out_file, dest)?;
 
     Ok(())
 }


build/xtask/src/clippy.rs
```diff
@@ -16,7 +16,7 @@ pub fn run(
 ) -> Result<()> {
     let toml = Config::from_file(&cfg)?;
 
-    let mut src_dir = cfg.to_path_buf();
+    let mut src_dir = cfg;
```
steveklabnik, line 19: is there a reason to prefer this over making `cfg` `mut` above?

build/xtask/src/dist.rs
```diff
@@ -130,80 +138,66 @@ pub fn package(
...
-        edges,
-        &allocs,
-        &mut all_output_sections,
-        &entry_points,
-        &dist_dir,
-        &remap_paths,
-    )?;
-
     // Generate combined SREC, which is our source of truth for combined images.
+    let (kentry, ksymbol_table) = kern_build.unwrap();
```
steveklabnik, line 221: So, i asked myself, "why this unwrap?" Answer: this function returns `Result`, and this is an Option. Makes sense.

But if we look at things above... `build_kernel` returns a `Result`, which we `?` and then wrap in an option, only to unwrap it here. So I was like "Could we get rid of that extra layer of indirection?" And after trying some things, the answer is "no". Or at least, I couldn't figure out something better. Kudos on finding a legit situation for unwrap 😅

build/xtask/src/dist.rs
```diff
@@ -697,18 +705,33 @@ fn check_task_priorities(toml: &Config) -> Result<()> {
...
                     name, task.priority, callee, p
                 )?;
-            } else if task.priority >= idle_priority && name != "idle" {
-                bail!("task {} has priority that's >= idle priority", name);
             }
         }
+        if task.priority >= idle_priority && name != "idle" {
+            bail!("task {} has priority that's >= idle priority", name);
+        }
+        if task.priority == 0 {
```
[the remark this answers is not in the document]
mkeeter: Fixed, and decided to enforce that it's both task 0 _and_ the only task at priority 0!

build/xtask/src/dist.rs
```diff
@@ -662,10 +428,251 @@ Did you mean to run `cargo xtask dist`?"
...
+}
+
+fn check_task_names(toml: &Config, task_names: &[String]) -> Result<()> {
+    // Quick sanity-check if we're trying to build individual tasks which
+    // aren't present in the app.toml, or ran `cargo xtask build ...` without
+    // any specified tasks.
+    if task_names.is_empty() {
+        bail!(
+            "Running `cargo xtask build` without specifying tasks has no effect.
+Did you mean to run `cargo xtask dist`?"
```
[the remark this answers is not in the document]
mkeeter: Thanks, I always forget about that trick

build/xtask/src/dist.rs
```diff
@@ -53,45 +55,103 @@ struct PackageConfig {
...
+        // path for its --remap-path-prefix argument, so we use
+        // `dunce::canonicalize` instead
+        let cargo_home = dunce::canonicalize(std::env::var("CARGO_HOME")?)?;
+        let mut cargo_git = cargo_home.clone();
+        cargo_git.push("git");
+        cargo_git.push("checkouts");
+        remap_paths.insert(cargo_git, "/git");
+
+        // This hash is canonical-ish: Cargo tries hard not to change it
+        // https://github.com/rust-lang/cargo/blob/master/src/cargo/core/source/source_id.rs#L607-L630
```
steveklabnik, line 103: how will we know if this changes? purely long error messages, something else? I wonder if there's a good way to be like "this is the place you come fix this"

steveklabnik: (I know you just moved this code around so I don't think addressing this should block the PR, just wondering now...)

build/xtask/src/clippy.rs
mkeeter: I decided on
```
    let src_dir = cfg
        .parent()
        .ok_or_else(|| anyhow!("Could not get src dir"))?;
```
which has nothing mutable!

build/xtask/src/dist.rs
mkeeter: Alright, I figured out a way to do it without repeating anything or invoking `unwrap()`, but I think it's worse-looking:
```rust
    // Build the kernel!
    let kern_build = if tasks_to_build.contains("kernel") {
        Some(build_kernel(
            &cfg,
            &allocs,
            &mut all_output_sections,
            &entry_points,
        )?)
        .filter(|_| !partial_build) // <-- this is new
    } else {
        None
    };

    // If we've done a partial build (which may have included the kernel), bail
    // out here before linking stuff.
    let kern_build = match kern_build { // <-- this is new
        Some(k) => k,
        None => return Ok(()),
    };

    // Generate combined SREC, which is our source of truth for combined images.
    let (kentry, ksymbol_table) = kern_build;
```

build/xtask/src/dist.rs
mkeeter: Good question – there's no good way to check it automatically, but it should only change when we bump compiler versions, so it's not too hard to check manually.  If we end up writing a playbook for "bumping the Hubris compiler", this should be in there!

PR #569, opened 2022-05-25
Clean up taskdesc and index handling

cbiffle:
This cleans up some cruft in the abi crate, but the biggest part of the change is best summarized by the message of the top commit:

---

    kern: add task index to TaskDesc instead of computing
    
    From time immemorial, the kernel has computed the task index of the
    current task during the syscall/ISR entry sequences using a subtraction
    and a division. This is problematic on ARMv6-M, where hardware division
    is unavailable.
    
    This commit applies mkeeter's idea of just stashing a task index in the
    TaskDesc. By adjusting the size of some existing fields that had room to
    spare, this doesn't even make the TaskDesc larger.
    
    This knocks 6% or 600+ bytes off the STM32G0 kernel by removing the last
    use of division, and should take hundreds of cycles off the Cortex-M0
    syscall and interrupt latency.
    
    Interestingly it also makes the M7 kernel smaller, by removing a panic
    path: the old index computation code could panic on subtraction
    underflow if the task address was lower than the start of the task
    table, and contained code to handle this case at runtime.

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -1327,7 +1327,6 @@ pub fn make_kconfig(
         base: 0,
         size: 32, // smallest legal size on ARMv7-M
         attributes: abi::RegionAttributes::empty(), // no rights
-        reserved_zero: 0,
     });
 
     // Regions 1.. are the fixed peripheral regions, shared by tasks that
@@ -1373,7 +1372,6 @@ pub fn make_kconfig(
             base: p.address,
             size: p.size,
             attributes,
-            reserved_zero: 0,
         });
     }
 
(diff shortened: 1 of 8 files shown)


build/xtask/src/dist.rs
```diff
@@ -1478,6 +1474,7 @@ pub fn make_kconfig(
                 + task.stacksize.or(toml.stacksize).unwrap(),
             priority: task.priority,
             flags,
+            index: u16::try_from(i).expect("more than 2**16 tasks?"),
```
steveklabnik, line 1477: i love this error message, hahah

steveklabnik (approved):

PR #572, opened 2022-05-26
Consolidate `humility` subcmds and restore `xtask gdb`

mkeeter:
This consolidates the three Humility-based subcommands (`flash` / `test` / `humility`) to all run through `humility::run`, instead of having separate-but-mostly-overlapping implementations.  It also restores `cargo xtask gdb`, using the same code path.

Drive-by fixes:
- Remove an unneeded `mut`
- Make `flash.ron` pretty-printed


build/xtask/src/main.rs
```diff
@@ -121,11 +120,28 @@ enum Xtask {
...
+    #[clap(short, long)]
+    verbose: bool,
+
+    /// Extra options to pass to Humility
+    #[clap(last = true)]
+    extra_options: Vec<String>,
+}
+
 // For commands which may execute on specific packages, this enum
 // identifies the set of packages that should be operated upon.
```
steveklabnik, line 138: I know you didn't modify this line, but I now find it pretty confusing? it is supposedly describing an enum but is describing main?

build/xtask/src/main.rs
```diff
@@ -121,11 +120,28 @@ enum Xtask {
...
 fn main() -> Result<()> {
     let xtask = Xtask::parse();
+    run(xtask)
+}
 
+fn run(xtask: Xtask) -> Result<()> {
```
steveklabnik, line 144: what's the reason to extract everything but arg parsing into a separate function?

steveklabnik (approved):
This looks great overall! I have two little non-blocking questions.

build/xtask/src/main.rs
mkeeter: So I can re-call into `run` [further down](https://github.com/oxidecomputer/hubris/blob/f194910e81ae34e1ac21a07b8a48ddf78697ff78/build/xtask/src/main.rs#L187)

build/xtask/src/main.rs
mkeeter: lol yeah, it looks like someone (probably me) forgot to delete it when removing `enum RequestedPackages`
```rust
// For commands which may execute on specific packages, this enum
// identifies the set of packages that should be operated upon.
enum RequestedPackages {
    // Specifies a single specific (Package, Target) pair.
    Specific(Option<String>, Option<String>),
    // Specifies the command should operate on all packages.
    All,
}
```

I deleted it now!

PR #573, opened 2022-05-26
Expose `HIFFY_SCRATCH` buffer

### LINKED ISSUE #473: Make hiffy scratch buffer size discoverable by humility
See https://github.com/oxidecomputer/humility/issues/116

andrewjstone:
The scratch buffer used by hiffy is now a static and no longer on the stack.

Fixes #473

### DIFF
--- a/task/hiffy/src/main.rs
+++ b/task/hiffy/src/main.rs
@@ -17,7 +17,7 @@
 
 use core::sync::atomic::{AtomicU32, Ordering};
 use hif::*;
-use userlib::*;
+use userlib::{util::StaticCell, *};
 
 #[cfg(armv6m)]
 use armv6m_atomic_hack::AtomicU32Ext;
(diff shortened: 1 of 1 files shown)


andrewjstone:
This has a corresponding humility change that should probably be merged first. https://github.com/oxidecomputer/humility/pull/152

task/hiffy/src/main.rs
```diff
@@ -61,13 +61,16 @@ cfg_if::cfg_if! {
         const HIFFY_RSTACK_SIZE: usize = 2048;
     }
 }
+const HIFFY_SCRATCH_SIZE: usize = 256;
```
cbiffle, line 64: As a result of this change you've made it so that `HIFFY_SCRATCH_SIZE` could usefully be changed -- since Humility can detect its size now. I'd suggest adding a comment on this constant to that effect, something like

```rust
/// Number of "scratch" bytes available to Hiffy programs. Humility uses this to deliver data
/// used by some operations. This number can be increased at the cost of RAM.

cbiffle (approved):

task/hiffy/src/main.rs
andrewjstone: Done.

PR #574, opened 2022-05-26
Build crates by name instead of path

mkeeter:
In our build system, we specify both crate names _and_ paths, and use the latter to run `cargo rustc ...` in the crate's directory.

However, crate names are guaranteed to be unique by dint of being in the same workspace.  This means we can instead use `cargo rustc -p ...` to run on that specific crate, saving a bunch of boilerplate in `app.toml`.

Other drive-by fixes:
- Added `sysroot` to `BuildConfig`, which speeds up a no-op Sidecar rebuild by 12% by skipping `rustup` façade (!)
- When doing `cargo clean` after changing apps, clean all the apps at once (`cargo clean -p hf -p spi_driver ...`).  This is _much faster_.

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -481,25 +481,23 @@ fn check_rebuild(toml: &Config) -> Result<()> {
     // if we need to rebuild, we should clean everything before we start building
     if rebuild {
         println!("app.toml has changed; rebuilding all tasks");
-        let names = std::iter::once(toml.kernel.name.as_str())
-            .chain(
-                toml.bootloader
-                    .as_ref()
-                    .map(|b| b.name.as_str())
-                    .into_iter(),
-            )
-            .chain(toml.tasks.keys().map(|name| {
-                // this feels redundant, don't we already have the name?
-                // consider our supervisor:
-                //
-                // [tasks.jefe]
-                // name = "task-jefe"
-                //
-                // the "name" in the key is jefe, but the package name is in
-                // tasks.jefe.name, and that's what we need to give to cargo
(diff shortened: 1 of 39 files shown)


build/xtask/src/dist.rs
```diff
@@ -481,24 +495,23 @@ fn check_rebuild(toml: &Config) -> Result<()> {
...
             // [tasks.jefe]
-            // path = "../task-jefe"
             // name = "task-jefe"
             //
-            // the "name" in the key is jefe, but the package name is in
-            // tasks.jefe.name, and that's what we need to give to cargo
-            let task_toml = &toml.tasks[name];
-
-            cargo_clean(&task_toml.name, &toml.target)?;
+            // The "name" in the key is `jefei`, but the package (crate)
```
steveklabnik, line 509:
```suggestion
            // The "name" in the key is `jefe`, but the package (crate)
```

build/xtask/src/dist.rs
```diff
@@ -1026,7 +1046,7 @@ fn build(
     name: impl AsRef<Path>,
     build_config: BuildConfig,
 ) -> Result<()> {
-    println!("building path {}", build_config.crate_path.display());
+    println!("building crate {}", build_config.crate_name);
```
steveklabnik, line 1049: I love this change because it's been bugging me for a while, which is that this on master looks like this to me:

```
building path .\app\demo-stm32f4-discovery\../../task/jefe
```

both slashes, the `..`s... now this output will look good again :)

steveklabnik (approved):
Love this. Both great changes. I have found one tiny typo, but other than that, this looks great to me.

PR #576, opened 2022-05-26
Some work towards a sizing action

steveklabnik:
I talked with @mkeeter about this yesterday; rather than wait until I've figured out the actions stuff, here's a PR for just the code that calculates the sizes. The gist of it is:

This PR adds two new flags to `cargo xtask sizes`:

* --save, which writes out the sizes to a file
* --compare, which calculates sizes and then compares to a previously
  saved file

These files are named after the app, for example,
`demo-stm32f4-discovery.json`, and are written to the project root.

I am open to any and all suggestions about this code. I think it's... fine. I'm not sure what I don't like about it. Also happy to revise the overall strategy.

And one open question: should I put `*.json` in the `.gitignore`, or maybe include something a bit more recognizable in the filename so we could glob it?

### DIFF
--- a/build/xtask/src/main.rs
+++ b/build/xtask/src/main.rs
@@ -77,6 +77,14 @@ enum Xtask {
         verbose: bool,
         /// Path to the image configuration file, in TOML.
         cfg: PathBuf,
+
+        /// Compare this to a previously saved file of sizes
+        #[clap(long)]
+        compare: bool,
+
+        /// Write JSON out to a file?
+        #[clap(long)]
+        save: bool,
     },
 
     /// Runs `humility`, passing any arguments
@@ -161,7 +169,7 @@ fn run(xtask: Xtask) -> Result<()> {
             cfg,
         } => {
             dist::package(verbose, edges, &cfg, None)?;
-            sizes::run(&cfg, true)?;
+            sizes::run(&cfg, true, false, false)?;
         }
         Xtask::Build {
             verbose,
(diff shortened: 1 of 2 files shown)


mkeeter (approved):
LGTM, although it's probably time to introduce a type for
```rust
(
    IndexMap<String, IndexMap<String, u64>>,
    IndexMap<String, Vec<Vec<(String, u32, u64)>>>,
)
```
(up to you if you want to do that now or in later work)

steveklabnik:
I am happy to do it as part of this, given it shouldn't be a big thing. I am also not 100% sure what exactly the best way to type this is. Let me explain.

I imagine we'd still want to keep this as a tuple of `(sizes, suggestions)`. I'm not sure making it an explicit struct adds a ton of clarity. 

So that means that `sizes` is a `IndexMap<String, IndexMap<String, u64>>`. The only way we could really type the outside is with is via a type alias, or a tuple struct. And the problem is kinda replicated on the inside too. This process then kinda repeats itself for the suggestions, except that I do think that doing some sort of `Suggestion` type is much more clear than `(String, u32, u64)`.

So.. what's your opinion on the right thing here? Do you think the aliases would pull their weight? If so, I'm not *against* using them. Or maybe there's a good way to describe this type I'm sorta missing.

steveklabnik:
(Matt and Cliff both suggest replacing the outer tuple with a struct, so I'm going to do that now!)

PR #582, opened 2022-05-31
Eliminate or atomize most kernel statics, reducing unsafe

cbiffle:
This PR eliminates kernel static variables that are not technically necessary for kernel operation (and were preserved mostly for Humility), and converts the rest to `Atomic` types. This lets us access the statics from safe code with well-defined semantics. (In practice, as the kernel is not preemptible, this was already true -- but now the compiler is aware of it.)

These uses of unsafe were originally less obvious / more easily tolerated because we were allowing `unsafe_op_in_unsafe_fn`, as Rust always has historically.

This should be compatible with recent Humility after @bcantrill's updates. The main part of this that I, personally, am nervous about and would like review on is the timestamp handling changes.


sys/kern/src/arch/arm_m.rs
```diff
@@ -1001,51 +906,61 @@ pub(crate) fn with_task_table<R>(
...
-        // in the range of nanoseconds to milliseconds -- meaning over 500 years.
-        // However, we do not use wrapping add here because, if we _do_ overflow due
-        // to e.g. memory corruption, we'd rather panic and reboot than attempt to
-        // limp forward.
-        *ticks += 1;
-        // Now, give up mutable access to *ticks so there's no chance of a
-        // double-increment due to bugs below.
-        let now = Timestamp::from(*ticks);
-        drop(ticks);
+        // Load the time before this tick event.
```
mkeeter, line 942: Is there any advantage to building the `Timestamp` outside of the closure and passing it in, to minimize how long the task table is held?

cbiffle: Not really -- the task table in-use flag is not a mutex, and this is not a critical section (as the kernel is not preemptible). Moving the bounds of the closure would just rearrange instructions.

mkeeter (approved):
Looks good, one non-blocking question.

PR #583, opened 2022-05-31
`staticlib` build

mkeeter:
This is step 2 in refactoring the build system for task autosizing.

I converted all of the tasks from executables to `staticlib`, which lets us build and link them in separate steps.  Right now, it _always_ performs the link step (even if the `.a` file is unchanged); this is to avoid also needing to track the linker scripts.

This is a draft PR because I'm seeing changes in the resulting images that need further investigation.


mkeeter:
Closed in favor of #584

PR #584, opened 2022-06-02
Use relocatable linking to split build and link steps

mkeeter:
This is an alternative path towards task autosizing (as an alternative to #583 ).

It uses the `-r` linker flag to enable relocations, meaning the build produces a relocatable ELF file, then runs the linker a second time to make it static (and located at the correct point in memory).


build/xtask/src/dist.rs
```diff
@@ -194,6 +210,15 @@ pub fn package(
         }
     }
 
+    // Build all tasks (which are static binaries, so they are not linked yet)
```
cbiffle, line 213: Nit: "which are partially linked relocatable binaries"

build/xtask/src/dist.rs
```diff
@@ -203,7 +228,11 @@ pub fn package(
         .keys()
         .map(|name| {
             let ep = if tasks_to_build.contains(name.as_str()) {
-                build_task(&cfg, name, &allocs, &mut all_output_sections)
+                // Link tasks regardless of whether they have changed, because
```
cbiffle, line 231: Note that this technique is probably not sufficient for handling changes to the `rlink` script.

cbiffle (approved):

PR #580, opened 2022-05-29
Sidecar Power and Thermal management

arjenroodselaar:
This diff adds a first cut for power and thermal management of a Sidecar mainboard/chassis by adding support for Sidecar to the `power` and `thermal` tasks.

PR #586, opened 2022-06-06
supress a warning about an unused result

steveklabnik:
We've got a couple of warnings in the build that are leaving comments on PRs. This is the simplest one; while this TODO is being figured out, at least we can stop all of this:
image


mkeeter:
FYI there are drive-by fixes for these warnings in #580

steveklabnik:
Ah cool! I'll just close then.

PR #589, opened 2022-06-07
Fix A2->A0 leakage

nathanaelhuffman:
New FPGA with tri-state outputs for cross-domain signals.
Disable UART mux when transitioning A0->A2.


drv/gimlet-seq-server/src/main.rs
```diff
@@ -513,6 +518,19 @@ cfg_if::cfg_if! {
             sys.gpio_reset(UART_TX_ENABLE).unwrap();
         }
 
+        fn uart_sp_to_sp3_disable() {
+            let sys = sys_api::Sys::from(SYS.get_task_id());
+
+            sys.gpio_configure_output(
+                UART_TX_ENABLE,
+                sys_api::OutputType::PushPull,
+                sys_api::Speed::High,
```
mkeeter, line 527:
```suggestion
                sys_api::Speed::Low,
```
😉

nathanaelhuffman: hehe, I can make that change, probably also in the enable function which I copied to make this?

mkeeter: Sure, that makes sense!

kc8apf (approved):
Verified that A4 -> A3 -> A2 leaves V3P3_SYS_A0 at 0V.  A2 -> A0 bring V3P3_SYS_A0 up to 3.3V as expected and FTDI module connected to UART0 was able to interact with nanobl-rs so buffer was enabled correctly.  A0 -> A2 returns V3P3_SYS_A0 to 0V so UART0 buffer appears to be disabled correctly.

PR #592, opened 2022-06-07
jefe system state API

cbiffle:
This adds a system state tracking API to Jefe. Currently the only things that can be done with that API are:
- Read the current state
- Alter the state
- Get notified if someone alters the state.

In support of that third one, Jefe now has configuration: it can be asked to poke tasks when the state changes.

This PR also includes `gimlet-seq-server` support for posting sequencing state to Jefe, and corresponding client code in `net` to use Jefe's state instead of directly contacting the sequencer task.

### DIFF
--- a/build/util/src/lib.rs
+++ b/build/util/src/lib.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use serde::de::DeserializeOwned;
 use std::env;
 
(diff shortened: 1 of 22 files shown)


build/util/src/lib.rs
```diff
@@ -48,38 +48,53 @@ pub fn expose_target_board() {
 /// reflect that by having its member (or members) be an `Option` type.
 ///
 pub fn config<T: DeserializeOwned>() -> Result<T> {
-    toml_from_env("HUBRIS_APP_CONFIG", Some("[config]"))
+    toml_from_env("HUBRIS_APP_CONFIG")?.ok_or_else(|| {
```
mkeeter, line 51: Out of curiosity, why `?.ok_or_else(|| { anyhow!(...) })` instead of `.context(...)?`?

task/jefe/src/main.rs
```diff
@@ -110,72 +121,127 @@ fn main() -> ! {
...
+        FAULT_MASK | TIMER_MASK
+    }
+
+    fn handle_notification(&mut self, bits: u32) {
+        // Check to see if we have any external requests
+        let changed = external::check(self.disposition);
+
+        // If our timer went off, we need to reestablish it
+        if bits & TIMER_MASK != 0 {
+            self.deadline += TIMER_INTERVAL;
```
mkeeter, line 200: `self.deadline` is initialized to `TIMER_INTERVAL` above.  If `jefe` restarts, I think this would cause it to spam timer interrupts until it catches up to kernel time.

We should either
- Initialize it to `sys_get_timer().now + TIMER_INTERVAL` when constructing the `ServerImpl`
- Update it with `self.deadline = sys_get_timer().now + TIMER_INTERVAL` when it goes off

build/util/src/lib.rs
cbiffle: Good point. Will fix.

task/jefe/src/main.rs
cbiffle: Well, technically, if `jefe` restarts, we're resetting the processor. However, initializing it to the current time would be more obviously correct, and I'll make that first change.

I'm not sure whether the second change is right -- this timeout thing is something @bcantrill added for the Hiffy interface, and adding the increment to it every time ensures that it doesn't drift. Adding the increment to the current time will cause drift if stuff is happening that delays execution of the code. In this case it likely doesn't matter, given that the timeout is set to be "fast for a human but slow for a computer," so I'm inclined to stick with the cheaper (fewer syscalls) option.

build/util/src/lib.rs
cbiffle: Ahhhhhh -- the `?` is load bearing.

This is converting a `Result<Option<T>>` into a `Result<T>` by (1) unwrapping the outer bits and then (2) applying [`Option::ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) to the interior to convert `None` into an error.

cbiffle:
@mkeeter - I have
1. [kebab'd the config](https://github.com/oxidecomputer/hubris/pull/592/commits/81c635a2da66246659e058a7db635d599021efa9#diff-6b46bc7f39e9c9bbc7e3ef491842c4da26c1614d5b66e353dae42296f2d12a06R37)
2. [made Jefe's initial deadline more plausible](https://github.com/oxidecomputer/hubris/pull/592/commits/319261386054991b9163acd8e4c198d4d5fc3e19#diff-9d0363bbabda71825a67857c2e4a0263721b5d1d4757b6ffc514bcee99355895R124)

I have not
1. Changed the `?.ok_or_else` sequences as they appear to actually be the right thing
2. Changed the interval computation in Jefe's loop to avoid drift.

Thoughts?

mkeeter (approved):

Issue #439, opened 2022-02-24
"Insufficient flash" error message doesn't always seem to work


steveklabnik:

steveklabnik:
As an example, after fixing the `app.toml` to have the right amount for `task-net`, here's the next error I got:

```
error: linking with `rust-lld` failed: exit code: 1
  |
  = note: "rust-lld" "-flavor" "gnu" "C:\\Users\\steve\\Documents\\Oxide\\hubris\\target\\thumbv7em-none-eabihf\\release\\deps\\task_udpecho-25fcf3909630cdca.task_udpecho.740dc942-cgu.0.rcgu.o" "--as-needed" "-L" "C:\\Users\\steve\\Documents\\Oxide\\hubris\\target\\thumbv7em-none-eabihf\\release\\deps" "-L" "C:\\Users\\steve\\Documents\\Oxide\\hubris\\target\\release\\deps" "-L" "target" "-L" "C:\\Users\\steve\\.rustup\\toolchains\\nightly-2022-02-22-x86_64-pc-windows-msvc\\lib\\rustlib\\thumbv7em-none-eabihf\\lib" "--start-group" "--end-group" "-Bstatic" "C:\\Users\\steve\\.rustup\\toolchains\\nightly-2022-02-22-x86_64-pc-windows-msvc\\lib\\rustlib\\thumbv7em-none-eabihf\\lib\\libcompiler_builtins-e8b9a93ab984480c.rlib" "-Bdynamic" "--eh-frame-hdr" "-znoexecstack" "-L" "C:\\Users\\steve\\.rustup\\toolchains\\nightly-2022-02-22-x86_64-pc-windows-msvc\\lib\\rustlib\\thumbv7em-none-eabihf\\lib" "-o" "C:\\Users\\steve\\Documents\\Oxide\\hubris\\target\\thumbv7em-none-eabihf\\release\\deps\\task_udpecho-25fcf3909630cdca" "--gc-sections" "-Tlink.x" "-z" "common-page-size=0x20" "-z" "max-page-size=0x20"
  = note: rust-lld: error: section '.rodata' will not fit in region 'FLASH': overflowed by 57 bytes
          rust-lld: error: section '.rodata' will not fit in region 'FLASH': overflowed by 76 bytes
          rust-lld: error: section '.rodata' will not fit in region 'FLASH': overflowed by 92 bytes
          rust-lld: error: section '.rodata' will not fit in region 'FLASH': overflowed by 118 bytes
[...]

PR #595, opened 2022-06-08
Automatically size tasks

### LINKED ISSUE #474: `xtask sizes` is incorrect with regards to the kernel
The kernel doesn't require a power of two memory size, but `xtask sizes` doesn't know that, and so will give incorrect results.

mkeeter:
This PR implements task autosizing, at long last!

It builds on the previous work with relocatable task builds (#584). After
building the relocatable task ELF file, it runs a "dummy link" against a
linker script with "infinite" memory (in practice, the entirety of memory
available on the chip). It then parses the resulting (static) binary to extract
sizes.

After finding sizes for every task, it runs the same memory packer as before,
then relinks each task with the resulting memory.

Task sizes are based on the target microcontroller, with a new `alignment`
parameter passed to `allocate_one`.

There are extensive changes to `cargo xtask sizes` to make it more generically
useful, decoupling the suggestions from the "find the size of a static ELF".

WARNING: this changes the format of the exported JSON files!

In addition, there are a bunch of new helper functions in `Config` to help with
task and memory sizing / alignment.

This fixes #474 and maybe #439, and deprecates #476

### DIFF
--- a/build/xtask/src/clippy.rs
+++ b/build/xtask/src/clippy.rs
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use std::collections::HashMap;
 use std::path::PathBuf;
 
 use anyhow::{bail, Result};
@@ -47,7 +48,10 @@ pub fn run(
         }
 
         let build_config = if name == "kernel" {
-            let (allocs, _) = crate::dist::allocate_all(&toml)?;
+            // Build dummy allocations for each task
+            let mut task_sizes = HashMap::new();
+
+            let (allocs, _) = crate::dist::allocate_all(&toml, &task_sizes)?;
             // Pick dummy entry points for each task
             let entry_points = allocs
                 .tasks
--- a/build/xtask/src/clippy.rs
+++ b/build/xtask/src/clippy.rs
@@ -2,7 +2,6 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use std::collections::HashMap;
 use std::path::PathBuf;
 
 use anyhow::{bail, Result};
@@ -49,7 +48,14 @@ pub fn run(
 
         let build_config = if name == "kernel" {
             // Build dummy allocations for each task
-            let mut task_sizes = HashMap::new();
+            let task_sizes = toml
+                .tasks
+                .keys()
+                .map(|name| name.as_str())
+                .zip(std::iter::repeat_with(|| {
+                    [("flash", 64), ("ram", 64)].into_iter().collect()
+                }))
+                .collect();
 
             let (allocs, _) = crate::dist::allocate_all(&toml, &task_sizes)?;
             // Pick dummy entry points for each task
--- a/build/xtask/src/clippy.rs
+++ b/build/xtask/src/clippy.rs
@@ -5,6 +5,7 @@
 use std::path::PathBuf;
 
 use anyhow::{bail, Result};
+use indexmap::IndexMap;
 
 use crate::config::Config;
 
@@ -48,13 +49,12 @@ pub fn run(
 
         let build_config = if name == "kernel" {
             // Build dummy allocations for each task
+            let fake_sizes: IndexMap<_, _> =
+                [("flash", 64), ("ram", 64)].into_iter().collect();
             let task_sizes = toml
                 .tasks
                 .keys()
-                .map(|name| name.as_str())
-                .zip(std::iter::repeat_with(|| {
-                    [("flash", 64), ("ram", 64)].into_iter().collect()
-                }))
+                .map(|name| (name.as_str(), fake_sizes.clone()))
                 .collect();
 
             let (allocs, _) = crate::dist::allocate_all(&toml, &task_sizes)?;
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -7,7 +7,7 @@ use std::hash::Hasher;
 use std::ops::Range;
 use std::path::{Path, PathBuf};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
 use indexmap::IndexMap;
 use serde::Deserialize;
 
@@ -59,6 +59,9 @@ impl Config {
     pub fn from_file(cfg: &Path) -> Result<Self> {
         let cfg_contents = std::fs::read(&cfg)?;
         let toml: RawConfig = toml::from_slice(&cfg_contents)?;
+        if toml.tasks.contains_key("kernel") {
+            bail!("'kernel' is reserved and cannot be used as a task name");
+        }
 
         let mut hasher = DefaultHasher::new();
         hasher.write(&cfg_contents);
@@ -74,7 +77,6 @@ impl Config {
             toml::from_slice(&chip_contents)?
         };
 
-        assert!(!toml.tasks.contains_key("kernel"));
         let buildhash = hasher.finish();
 
         Ok(Config {
@@ -291,10 +293,7 @@ impl Config {
 
     /// Calculates the output region which contains the given address
     pub fn output_region(&self, vaddr: u64) -> Option<&str> {
-        let vaddr: u32 = match vaddr.try_into() {
-            Ok(v) => v,
-            Err(_) => return None,
-        };
+        let vaddr = u32::try_from(vaddr).ok()?;
         self.outputs
             .iter()
             .find(|(_name, out)| {
@@ -317,10 +316,13 @@ impl Config {
 
     /// Checks whether the given chip's MPU requires power-of-two sized regions
     pub fn mpu_power_of_two_required(&self) -> bool {
-        matches!(self.mpu_alignment(), MpuAlignment::PowerOfTwo)
+        self.mpu_alignment() == MpuAlignment::PowerOfTwo
     }
 
-    /// Looks up the `requires` array for the given task or the kernel
+    /// Looks up the sizing array for the given task or the kernel.  For tasks,
+    /// this is the `max_sizes` array establishing upper bounds on autosizing;
+    /// for the kernel, this is the `requires` array which defines its size
+    /// (there is no kernel autosizing).
     pub fn requires(&self, name: &str) -> &IndexMap<String, u32> {
         match name {
             "kernel" => &self.kernel.requires,
@@ -337,32 +339,44 @@ impl Config {
                 // Nearest chunk of 16
                 ((size + 15) / 16) * 16
             }
-            _ => match self.mpu_alignment() {
-                MpuAlignment::PowerOfTwo => size.next_power_of_two(),
-                MpuAlignment::Chunk(c) => ((size + c - 1) / c) * c,
-            },
+            _ => self.mpu_alignment().suggest_memory_region_size(size),
         }
     }
 
     /// Returns the desired alignment for a task memory region. This is
     /// dependent on the processor's MMU.
     pub fn task_memory_alignment(&self, size: u32) -> u32 {
-        match self.mpu_alignment() {
-            MpuAlignment::PowerOfTwo => {
-                assert!(size.is_power_of_two());
-                size
-            }
-            MpuAlignment::Chunk(c) => c.try_into().unwrap(),
-        }
+        self.mpu_alignment().memory_region_alignment(size)
     }
 }
 
 /// Represents an MPU's desired alignment strategy
+#[derive(Copy, Clone, Debug, PartialEq)]
 enum MpuAlignment {
     /// Regions should be power-of-two sized and aligned
     PowerOfTwo,
     /// Regions should be aligned to chunks with a particular granularity
     Chunk(u64),
+}
+
+impl MpuAlignment {
+    /// Suggests a minimal memory region size fitting the given number of bytes
+    fn suggest_memory_region_size(&self, size: u64) -> u64 {
+        match self {
+            MpuAlignment::PowerOfTwo => size.next_power_of_two(),
+            MpuAlignment::Chunk(c) => ((size + c - 1) / c) * c,
+        }
+    }
+    /// Returns the desired alignment for a region of a particular size
+    fn memory_region_alignment(&self, size: u32) -> u32 {
+        match self {
+            MpuAlignment::PowerOfTwo => {
+                assert!(size.is_power_of_two());
+                size
+            }
+            MpuAlignment::Chunk(c) => (*c).try_into().unwrap(),
+        }
+    }
 }
 
 #[derive(Clone, Debug, Deserialize)]
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -319,17 +319,6 @@ impl Config {
         self.mpu_alignment() == MpuAlignment::PowerOfTwo
     }
 
-    /// Looks up the sizing array for the given task or the kernel.  For tasks,
-    /// this is the `max_sizes` array establishing upper bounds on autosizing;
-    /// for the kernel, this is the `requires` array which defines its size
-    /// (there is no kernel autosizing).
-    pub fn requires(&self, name: &str) -> &IndexMap<String, u32> {
-        match name {
-            "kernel" => &self.kernel.requires,
-            name => &self.tasks[name].max_sizes,
-        }
-    }
-
     /// Suggests an appropriate size for the given task (or "kernel"), given
     /// its true size.  The size depends on MMU implementation, dispatched
     /// based on the `target` in the config file.
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -303,15 +303,21 @@ impl Config {
             .map(|(name, _out)| name.as_str())
     }
 
-    /// Checks whether the given chip's MPU requires power-of-two sized regions
-    pub fn mpu_power_of_two_required(&self) -> bool {
+    fn mpu_alignment(&self) -> MpuAlignment {
         // ARMv6-M and ARMv7-M require that memory regions be a power of two.
         // ARMv8-M does not.
         match self.target.as_str() {
-            "thumbv8m.main-none-eabihf" => false,
-            "thumbv7em-none-eabihf" | "thumbv6m-none-eabi" => true,
+            "thumbv8m.main-none-eabihf" => MpuAlignment::Chunk(32),
+            "thumbv7em-none-eabihf" | "thumbv6m-none-eabi" => {
+                MpuAlignment::PowerOfTwo
+            }
             t => panic!("Unknown mpu requirements for target '{}'", t),
         }
+    }
+
+    /// Checks whether the given chip's MPU requires power-of-two sized regions
+    pub fn mpu_power_of_two_required(&self) -> bool {
+        matches!(self.mpu_alignment(), MpuAlignment::PowerOfTwo)
     }
 
     /// Looks up the `requires` array for the given task or the kernel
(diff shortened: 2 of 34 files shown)


build/xtask/src/clippy.rs
```diff
@@ -47,7 +47,17 @@ pub fn run(
         }
 
         let build_config = if name == "kernel" {
-            let (allocs, _) = crate::dist::allocate_all(&toml)?;
+            // Build dummy allocations for each task
+            let task_sizes = toml
+                .tasks
+                .keys()
+                .map(|name| name.as_str())
+                .zip(std::iter::repeat_with(|| {
```
cbiffle, line 55: I'm an iterator nerd, but, I had to stare at this for a bit. What about:

```rust
// above
let fake_sizes: HashMap<_, _> = [("flash", 64), ("ram", 64)].into_iter().collect();

// below that
.map(|name| (name.as_str(), fake_sizes.clone())
```

instead of map + zip + repeat_with?

Alternatively, based on your use of the `allocate_all` function, I'd consider changing its `task_sizes` argument type to

`impl FnMut(&str) -> u64`

and then for _this_ use you can pass in `|_| 64`, whereas for the original use, it'd be `|name| task_sizes[name]`

build/xtask/src/config.rs
```diff
@@ -74,6 +74,7 @@ impl Config {
             toml::from_slice(&chip_contents)?
         };
 
+        assert!(!toml.tasks.contains_key("kernel"));
```
cbiffle, line 77: This seems like kind of a random place for this, should it go higher? Should it `bail!` instead with a useful error message, such as `"sorry, you can't name a task kernel in the current system"` ?

build/xtask/src/config.rs
```diff
@@ -287,6 +288,81 @@ impl Config {
             })
             .collect()
     }
+
+    /// Calculates the output region which contains the given address
+    pub fn output_region(&self, vaddr: u64) -> Option<&str> {
+        let vaddr: u32 = match vaddr.try_into() {
```
cbiffle, line 294:
```rust
let vaddr = u32::try_from(vaddr).ok()?;
```

...or use `try_into` if you prefer it, it just takes more typing -- the `.ok()?` is the important bit.

build/xtask/src/config.rs
```diff
@@ -287,6 +288,81 @@ impl Config {
...
+            Ok(v) => v,
+            Err(_) => return None,
+        };
+        self.outputs
+            .iter()
+            .find(|(_name, out)| {
+                vaddr >= out.address && vaddr < out.address + out.size
+            })
+            .map(|(name, _out)| name.as_str())
```
cbiffle, line 303: Just a note, `find` returns `Option`, so you don't have to use `map` here if you don't want to (your current use is fine if you prefer it). You could write:

```rust
let name = self.outputs
    .iter()
    .find(stuff)?;
Some(name.as_str())
```

build/xtask/src/config.rs
```diff
@@ -287,6 +288,81 @@ impl Config {
...
+            "thumbv7em-none-eabihf" | "thumbv6m-none-eabi" => {
+                MpuAlignment::PowerOfTwo
+            }
+            t => panic!("Unknown mpu requirements for target '{}'", t),
+        }
+    }
+
+    /// Checks whether the given chip's MPU requires power-of-two sized regions
+    pub fn mpu_power_of_two_required(&self) -> bool {
+        matches!(self.mpu_alignment(), MpuAlignment::PowerOfTwo)
```
cbiffle, line 320: I feel like `self.mpu_alignment() == MpuAlignment::PowerOfTwo` is clearer, since you're not really using the Special Matches Powers

build/xtask/src/config.rs
```diff
@@ -287,6 +288,81 @@ impl Config {
...
+    }
+
+    /// Looks up the `requires` array for the given task or the kernel
```
cbiffle, line 323: This description is now incorrect -- this looks up max-sizes for tasks. Consider renaming the function. Based on your use (which mostly seems to be in `xtask sizes` maybe something like `expected_sizes`?

build/xtask/src/config.rs
```diff
@@ -287,6 +288,81 @@ impl Config {
...
+            name => &self.tasks[name].max_sizes,
+        }
+    }
+
+    /// Suggests an appropriate size for the given task (or "kernel"), given
+    /// its true size.  The size depends on MMU implementation, dispatched
+    /// based on the `target` in the config file.
+    pub fn suggest_memory_region_size(&self, name: &str, size: u64) -> u64 {
+        match name {
+            "kernel" => {
```
cbiffle, line 336: So, I'm a big fan of allowing `kernel` on the command line, but I wonder if we've gotten a little stringly-typed here. Now everything taking a name needs to handle `kernel`...or not, if it can't, but then it might still get passed `kernel`. This is kind of a validate-over-parsing decision. Would it make more sense to convert the task-or-kernel-name on the command line rather immediately into something like

```rust
enum Name<'a> {
    Kernel,
    Task(&'a str),
 }
 ```
 
 ?

File: build/xtask/src/config.rs [hunk 8]
```diff
@@ -287,6 +288,81 @@ impl Config {
...
+                // Nearest chunk of 16
+                ((size + 15) / 16) * 16
+            }
+            _ => match self.mpu_alignment() {
```
cbiffle, line 340: (This operation seems like a candidate for a function on `MpuAlignment`)

build/xtask/src/dist.rs
```diff
@@ -232,6 +215,32 @@ pub fn package(
         }
     }
 
+    // Calculate the sizes of tasks, assigning dummy sizes to tasks that
+    // aren't active in this build.
+    let task_sizes: HashMap<_, _> = cfg
+        .toml
```
cbiffle, line 221: I
love
rustfmt
so
much
can
I
just
tell
you

build/xtask/src/dist.rs
```diff
@@ -1300,16 +1359,20 @@ pub fn allocate_all(
...
-            if !amt.is_power_of_two() {
-                bail!("task {}, memory region {}: requirement {} is not a power of two.",
-                    task.name, name, amt);
+    for name in tasks.keys() {
+        for (mem, amt) in task_sizes[name.as_str()].iter() {
+            let bytes = toml.suggest_memory_region_size(name, *amt);
+            if let Some(r) = tasks[name].max_sizes.get(&mem.to_string()) {
+                if bytes > *r as u64 {
+                    bail!(
+                        "task {}, memory region {}: requirement {} is too small (needs {}).",
```
cbiffle, line 1368: I'd suggest removing the word "requirement" here since it's no longer `requires` -- maybe a message like

`task {name}: needs {bytes} bytes of {mem} but max-sizes limits it to {max}`

build/xtask/src/dist.rs
```diff
@@ -1300,16 +1359,20 @@ pub fn allocate_all(
...
+                        name, name, r, bytes);
```
cbiffle, line 1369: This is wrong, for the current format string it should be `name, mem, r, bytes`

build/xtask/src/dist.rs
```diff
@@ -1423,12 +1488,10 @@ fn allocate_k(
...
     region: &str,
     size: u32,
+    align: u32,
     avail: &mut Range<u32>,
 ) -> Result<Range<u32>> {
-    // This condition is ensured by allocate_all.
-    assert!(size.is_power_of_two());
-
-    let size_mask = size - 1;
+    let size_mask = align - 1;
```
cbiffle, line 1494: If you're going to compute the mask this way, you still need to assert that `align.is_power_of_two()`, because the result is nonsense for non-power-of-two inputs.

build/xtask/src/dist.rs
```diff
@@ -1497,15 +1560,7 @@ pub fn make_kconfig(
...
-    // ARMv6-M and ARMv7-M require that memory regions be a power of two.
-    // ARMv8-M does not.
-    let power_of_two_required = match toml.target.as_str() {
-        "thumbv8m.main-none-eabihf" => false,
-        "thumbv7em-none-eabihf" => true,
-        "thumbv6m-none-eabi" => true,
-        t => panic!("Unknown mpu requirements for target '{}'", t),
-    };
-
+    let power_of_two_required = toml.mpu_power_of_two_required();
```
cbiffle, line 1563: Given that this routine is already consulting the MPU requirement data, could you move the computation of `align` into here, rather than consulting the MPU requirement data outside this function and then calling it?

build/xtask/src/dist.rs
```diff
@@ -1551,24 +1606,32 @@ pub fn make_kconfig(
     // The remaining regions are allocated to tasks on a first-come first-serve
     // basis.
     for (i, (name, task)) in toml.tasks.iter().enumerate() {
-        if power_of_two_required && !task.requires["flash"].is_power_of_two() {
-            panic!("Flash for task '{}' is required to be a power of two, but has size {}", task.name, task.requires["flash"]);
+        if power_of_two_required
+            && !task_allocations[name]["flash"].len().is_power_of_two()
```
cbiffle, line 1610: This is not a new bug, but: these conditions need to be applied to _all_ regions, not just `"flash"` and `"ram"`. For instance the `sram1` area used by the ethernet driver still needs to be pow2.

build/xtask/src/sizes.rs
```diff
@@ -79,233 +68,235 @@ pub fn run(
...
-        }
-        Ok((sizes, suggestions))
-    };
+    }
+    let mut memory_sizes: IndexMap<&str, u64> = memory_sizes
+        .into_iter()
+        .map(|(name, range)| (name, range.end - range.start))
+        .collect();
+
+    // XXX: are there alignment issues with the stack here?
```
cbiffle, line 254: Yes, the stack is 8-byte aligned on all of our current targets -- so, technically, this could underestimate RAM usage by 7 bytes in corner cases.

cbiffle (commented):
This is looking pretty good. There are a couple of cases I've pointed out here that I think are bugs, but most of my comments are more informational than anything.

cbiffle:
I'm playing around with this branch, and I'm confused by the new xtask sizes output. (Yeah, I know, I read the code, I should probably understand it.)

What's this mean?

```
thermal
  flash: 12608 bytes (38%) [autosized to 16384]
  ram:    5284 bytes (64%)
```

12608 is 77% of 16384, so this is fine. (The yellow numbers used to indicate cases where space is being wasted; this is not one.)

Is it suggesting that the `max-size` could be reduced? If so, it would be nice to print the max-size it's mad about.

cbiffle:
Okay, how about this:

The use case for `xtask sizes` is, imo, showing the sizes of things. In an autosized world, its previous use case of "see how close I am to needing to adjust size up/down" is no longer the main reason you'd want to do that. So, we may want to reorient the output, maybe something like

```
kernel
  flash: 32768 bytes / 32768 (0% fragmentation) [limited to 32768]
  ram:    4596 bytes / 8192 (44% fragmentation) [limited to 8192]

thermal
  flash: 12608 bytes / 16384 (23% fragmentation) [limited to 32768]
  ram:    5284 bytes / 8192 (35% fragmentation) [limited to 8192]
```

Changes I'm suggesting there:
- Print the size and chosen partition size, `X / Y`, as the primary piece of information
- Flip the sense of the percentage to show how much space is being _wasted_.
- Add a `[limited to X]` note only for cases that have limits specified in config, which today is essentially everything, but in the future will not be.

Thoughts?

build/xtask/src/clippy.rs
mkeeter: Yeah, I'll do the simpler option (of cloning a separate `fake_sizes` variable)

build/xtask/src/config.rs
mkeeter: Sure, moved it to right after the `RawConfig` is loaded and switched to a more useful error message (`'kernel' is reserved and cannot be used as a task name`)

build/xtask/src/config.rs
mkeeter: Thanks, I always forget that `?` works on `Option` as well!

build/xtask/src/config.rs
mkeeter: Yeah, I was trying to avoid a `#[derive(..., PartialEq)]` block, but I suppose that's silly

build/xtask/src/config.rs
```diff
@@ -287,6 +288,81 @@ impl Config {
...
+                assert!(size.is_power_of_two());
+                size
+            }
+            MpuAlignment::Chunk(c) => c.try_into().unwrap(),
+        }
+    }
+}
+
+/// Represents an MPU's desired alignment strategy
+enum MpuAlignment {
```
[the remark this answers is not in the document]
mkeeter: Yup, done

build/xtask/src/dist.rs
mkeeter: Good idea, although we can't use nice format strings due to compiler version 😢

build/xtask/src/dist.rs
mkeeter: Yup, fixed

build/xtask/src/config.rs
mkeeter: I experimented with this a little bit in [a `strongly-typed-task-names` branch](https://github.com/oxidecomputer/hubris/tree/strongly-typed-task-names), but am not totally satisfied with the result – it ripples through a lot of code, and ends up feeling un-ergonomic.  I'd vote to open a ticket for this and revisit later.

cbiffle: Fine by me.

build/xtask/src/sizes.rs
```diff
@@ -79,233 +68,235 @@ pub fn run(
...
-        || toml.target.starts_with("thumbv6m")
-    {
-        pow2_suggest
-    } else if toml.target.starts_with("thumbv8m") {
-        armv8m_suggest
-    } else {
-        panic!("Unknown target: {}", toml.target);
+/// Loads the size of the given task (or kernel)
+pub fn load_task_size<'a>(
+    toml: &'a Config,
```
steveklabnik, line 199: I am not sure i prefer `toml` as a name, given that the old name was `config` because it's a `Config`, but I don't feel strongly enough to argue that you should change it back :)

build/xtask/src/sizes.rs
```diff
@@ -79,233 +68,235 @@ pub fn run(
...
+                _ => format!("{}.tmp", name),
+            });
+
+    let buffer = std::fs::read(elf_name)?;
+    let elf = match Object::parse(&buffer)? {
+        Object::Elf(elf) => elf,
+        o => bail!("Invalid Object {:?}", o),
     };
 
-    let check_task = move |name: &str,
```
steveklabnik, line 208: i am glad this monstrosity is leaving, haha. i *almost* wanted to turn it into a free function, but it captured one or two variables and that was convenient.

build/xtask/src/sizes.rs
```diff
@@ -79,233 +68,235 @@ pub fn run(
...
-                }
-            } else {
-                assert!(used == 0);
+    };
+    for phdr in &elf.program_headers {
+        record_size(phdr.p_vaddr, phdr.p_memsz);
+
+        // If the VirtAddr disagrees with the PhysAddr, then this is a
+        // section which is relocated into RAM, so we also accumulate
+        // its FileSiz in the physical address (which is presumably
```
steveklabnik, line 241: is `FileSiz` a typo? Shouldn't it be "file size" or `FileSz` or something? an extremely minor thing.

steveklabnik (approved):
Looks good to me! A few anecdotes and one maybe-typo.

build/xtask/src/sizes.rs
mkeeter: heh, the names in this section based on the output of `readelf`, so it's intentional
```
Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  LOAD           0x000120 0x08039400 0x08039400 0x00060 0x00060 R E 0x20
  LOAD           0x000180 0x08039460 0x08039460 0x00020 0x00020 R E 0x20
  GNU_STACK      0x000000 0x00000000 0x00000000 0x00000 0x00000 RW  0x0
```

build/xtask/src/config.rs
mkeeter: I ended up removing this entirely while editing `xtask sizes`

mkeeter:
Alright, I've refactored the `cargo xtask sizes` output and think this is ready to go!

Here's an example:
```
flash = 0x08000000..0x08100000
ram   = 0x20000000..0x20020000
sram1 = 0x30000000..0x30020000
Used:
  flash: 0x66880 (40%)
  ram:   0x1d500 (91%)
  sram1: 0x4000 (12%)

flash:
      ADDRESS  |   PROGRAM    |   USED |   SIZE | LIMIT
    0x08000000 | kernel       |  32768 |  32768 | (fixed)
    0x08008000 | hiffy        |  20120 |  32768 | 32768
    0x08010000 | gimlet_seq   |  51980 |  65536 | 65536
    0x08020000 | net          |  86080 | 131072 | 131072
    0x08040000 | spi4_driver  |  11436 |  16384 | 16384
    0x08044000 | spi2_driver  |  11852 |  16384 | 16384
    0x08048000 | i2c_driver   |  11916 |  16384 | 16384
    0x0804c000 | spd          |  11608 |  16384 | 16384
    0x08050000 | thermal      |  12608 |  16384 | 32768
    0x08054000 | power        |  12012 |  16384 | 16384
    0x08058000 | hf           |   8768 |  16384 | 16384
    0x0805c000 | jefe         |   7200 |   8192 | 8192
    0x0805e000 | sensor       |   4160 |   8192 | 8192
    0x08060000 | udpecho      |   7456 |   8192 | 16384
    0x08062000 | udpbroadcast |   5024 |   8192 | 16384
    0x08064000 | validate     |   8012 |   8192 | 8192
    0x08066000 | sys          |   1664 |   2048 | 2048
    0x08066800 | idle         |     96 |    128 | 128

ram:
      ADDRESS  |   PROGRAM    |   USED |   SIZE | LIMIT
    0x20000000 | kernel       |   4596 |   8192 | (fixed)
    0x20002000 | thermal      |   5284 |   8192 | 8192
    0x20004000 | net          |  16208 |  16384 | 16384
    0x20008000 | hiffy        |  27312 |  32768 | 32768
    0x20010000 | spd          |  10264 |  16384 | 16384
    0x20014000 | udpecho      |   4228 |   8192 | 8192
    0x20016000 | power        |   2828 |   4096 | 4096
    0x20017000 | gimlet_seq   |   2508 |   4096 | 4096
    0x20018000 | udpbroadcast |   2180 |   4096 | 8192
    0x20019000 | validate     |   2164 |   4096 | 4096
    0x2001a000 | jefe         |   1620 |   2048 | 2048
    0x2001a800 | spi4_driver  |   2036 |   2048 | 2048
    0x2001b000 | spi2_driver  |   2036 |   2048 | 2048
    0x2001b800 | i2c_driver   |   2008 |   2048 | 2048
    0x2001c000 | hf           |   2048 |   2048 | 2048
    0x2001c800 | sensor       |   2048 |   2048 | 2048
    0x2001d000 | sys          |    896 |   1024 | 1024
    0x2001d400 | idle         |    256 |    256 | 256

sram1:
      ADDRESS  |   PROGRAM    |   USED |   SIZE | LIMIT
    0x30000000 | net          |  12480 |  16384 | 16384

PROGRAM       REGION  USED    SIZE    LIMIT
kernel        flash   32768   32768   (fixed)
              ram     4596    8192    (fixed)
jefe          flash   7200    8192    8192
              ram     1620    2048    2048
net           flash   86080   131072  131072
              ram     16208   16384   16384
              sram1   12480   16384   16384
sys           flash   1664    2048    2048
              ram     896     1024    1024
spi4_driver   flash   11436   16384   16384
              ram     2036    2048    2048
spi2_driver   flash   11852   16384   16384
              ram     2036    2048    2048
i2c_driver    flash   11916   16384   16384
              ram     2008    2048    2048
spd           flash   11608   16384   16384
              ram     10264   16384   16384
thermal       flash   12608   16384   32768
              ram     5284    8192    8192
power         flash   12012   16384   16384
              ram     2828    4096    4096
hiffy         flash   20120   32768   32768
              ram     27312   32768   32768
gimlet_seq    flash   51980   65536   65536
              ram     2508    4096    4096
hf            flash   8768    16384   16384
              ram     2048    2048    2048
sensor        flash   4160    8192    8192
              ram     2048    2048    2048
udpecho       flash   7456    8192    16384
              ram     4228    8192    8192
udpbroadcast  flash   5024    8192    16384
              ram     2180    4096    8192
validate      flash   8012    8192    8192
              ram     2164    4096    4096
idle          flash   96      128     128
              ram     256     256     256

kernel:
  ram:    4608 (currently 8192)
```

PR #601, opened 2022-06-10
stm32h7-i2c: rework target state machine

### LINKED ISSUE #544: STM32H7 I2C: target code sometimes fails to handle all interrupt flags, causing busyloop
When we raised `spd`'s priority on gimlet we noticed that other things were getting starved, which seems darn suspicious given that spd should be spending like 98.7% of its life waiting for interrupts.

Unless it's not waiting.

Inspection of the ringbufs showed a lot of this:

```
  11  846    30017        1 TxISR(0x6c0005)
```

Bit 2 is set, which is RXNE, which is bad. The Tx loop doesn't _handle_ RXNE. So, it is immediately popping out of wfi. The transmit still completes, but only after having left a bloody trail of preemption behind it.

I believe the problem is in the rx loop, specifically around [here](https://github.com/oxidecomputer/hubris/blob/27cb5cd4d7bf118037c8733095f639bb580979b9/drv/stm32h7-i2c/src/lib.rs#L811). If we enter that loop with RXNE set _and_ either STOPF or NACKF, we will handle the STOPF or NACKF and bail before handling the RXNE.

Generally, when responding to an interrupt, we need to make sure to clear _all_ relevant flags before moving on, which in this case means that the loop needs to potentially deal with _all three_ before leaving the loop, rather than early exiting.

(@bcantrill fyi)

cbiffle:
This rework is focused on handling interrupt flags comprehensively and
at the right times, to try and fix #544 and related issues.

Broad themes you should expect to see here:

1. Enable peripheral interrupt generation _only_ for the cases that are relevant to the target state machine and that we're prepared to handle. This should fix #544 implicitly since that's a case where we spin with a flag we weren't expecting.
2. Make sure all flags that might accumulate during a particular phase are handled in that phase. (In #544 there was a case where the rx loop would fail to dequeue a single byte, leaving RXNE set for the later phases to discover.)
3. Enable kernel-side interrupt only when you want it, which in this case means right before `wfi`.

### DIFF
--- a/drv/stm32h7-i2c/src/lib.rs
+++ b/drv/stm32h7-i2c/src/lib.rs
@@ -142,6 +142,8 @@ enum Trace {
     WaitRx,
     WaitTx,
     BusySleep,
+    Stop,
+    RepeatedStart(bool),
     None,
 }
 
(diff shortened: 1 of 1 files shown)


cbiffle:
Ok, I've exercised this some and confirmed that it behaves correctly in cases similar to what we've seen in #544.

bcantrill (approved):
This looks great -- comments are terrific. My only comment would be that it may be worth striking the two "TODO"s: one is a question that we should probably answer one way or the other (even if by acknowledging if it's unclear if flushing TXDR matters) and the other is probably worth elaborating on a bit more one way or the other.

PR #603, opened 2022-06-15
Add spi-msg communications from SP to RoT

lzrd:
EchoReq/Rsp, StatusReq/Rsp are implemented.
The SprocketsReq/Rsp messages are reserved.
See drv/spi-msg/README.md for details.


drv/lpc55-spi/src/lib.rs
```diff
@@ -225,13 +255,74 @@ impl Spi {
...
+    // Master idle status flag - true when master function is fully idle.
+    pub fn intstat(&self) -> (bool, bool, bool) {
+        let stat = self.reg.intstat.read();
+        (stat.ssa().bits(), stat.ssd().bits(), stat.mstidle().bits())
+    }
+
+    pub fn stat(
+        &mut self,
+    ) -> (bool, bool, bool, bool, bool, bool, bool, u8, u8) {
+        let stat = self.reg.fifostat.read();
```
labbott, line 272: It's hard to follow all the values of the returned tuple here and other places. Can you just return the value of `fifostat.read()` and call the appropriate function instead?

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,24 +8,329 @@
...
+// To support that semantic, the RoT will always
+// process incoming bytes even if it is transmitting a previously queued
+// message.
+// TODO: properly test that case.
+//
+// Messages for protocol 1 are composed of a four-byte header:
+//   - a protocol version (0x01)
+//   - the lease significant byte of the payload length
+//   - the most significant byte of the payload length
+//   - a message type
```
mkeeter, line 47: How is message type encoded?

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,24 +8,329 @@
...
+    Idle = 0,       // Idle, waiting for message
+    Header = 1,     // Expecting message length LSB, MSB, and MsgType
+    Payload = 2,    // receiving message payload
+    Dispatch = 3,   // Receive complete, message processing is pending.
+    Responding = 4, // Tx frame is queued. Terminal state until next SOT.
+    Invalid = 5,    // bad message or normally ignoring incoming data until SOT.
+    Error = 0xff,   // Errors, Terminal state until next SOT.
+}
+
+impl From<u8> for RxState {
```
mkeeter, line 77: Are we encoding `RxState` and `TxState` as a `u8` anywhere? If not, I'd suggest converting them to pure enums – remove the explicit integer values, `#[repr(u8)]`, and this `From` implementation, since we aren't using any of that.

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,24 +8,329 @@
...
+    HeaderComplete,
+    HeaderOnly,
+    Initialized,
+    InvalidRxState(RxState),
+    Irq(bool, bool, bool),
+    Line,
+    SSAsserted,
+    SSDeasserted,
+    Loop(u16, u16, TxState, RxState),
+    LStat(bool, bool, bool, bool, bool, bool, bool, u8, u8),
```
mkeeter, line 137: A `struct LStat { ... }` with named fields would clarify this variant.

drv/spi-msg/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+    SpiServerError = 7,
+}
+
+/// Protocol version
+pub const SPI_MSG_IGNORE: u8 = 0; // To be ignored
+pub const SPI_MSG_VERSION: u8 = 1; // Supported message format
+
+/// SPI Message types will allow for multiplexing and forward compatibility.
+#[derive(
+    Copy, Clone, PartialEq, Eq, Debug, userlib::FromPrimitive, AsBytes,
```
mkeeter, line 44: Nit: you don't need an explicit `userlib::` here

drv/spi-msg/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+    EchoReq = 2,
+    EchoRsp = 3,
+    StatusReq = 4,
+    StatusRsp = 5,
+    SprocketsReq = 6,
+    SprocketsRsp = 7,
+    Unknown = 0xff,
+}
+
+impl From<u8> for MsgType {
```
mkeeter, line 59: You're already deriving `FromPrimitive` – do you really need `From` as well?

If you do, you can implement it in terms of `FromPrimitive`, e.g. `MsgType::from_u8(msgtype).unwrap_or(MsgType::Unknown)`

drv/spi-msg/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+            _ => MsgType::Unknown,
+        }
+    }
+}
+
+#[derive(FromBytes, AsBytes, Unaligned, Debug)]
+#[repr(C)]
+pub struct MsgHeader {
+    version: u8,
+    len_lsb: u8,
```
mkeeter, line 95: You could use [`zerocopy::byteorder::U16`](https://docs.rs/zerocopy/latest/zerocopy/byteorder/struct.U16.html) to explicitly store a `u16` here, instead of having to pack it yourself.

drv/spi-msg/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn set_len(&mut self, len: usize) {
+        self.header.len_lsb = (len & 0xff) as u8;
+        self.header.len_msb = (len >> 8) as u8;
+    }
+
+    pub fn set_msgtype(&mut self, msgtype: MsgType) {
+        self.header.msgtype = msgtype as u8;
+    }
+
+    pub fn payload_buf(&'a mut self) -> &'a mut [u8] {
```
mkeeter, line 150: Nit: no need for an explicit lifetime here

drv/spi-msg/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+        &mut self.body[..]
+    }
+
+    pub fn payload_get(&'a self) -> Result<&'a [u8], MsgError> {
```
mkeeter, line 154: Same here (no need for explicit lifetime)

drv/spi-msg/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+```sh
+SERIAL=/dev/ttyUSB0
+```
+
+In a [sprockets workspace](https://github.com/oxidecomputer/sprockets):
+```
+  cargo run -- -b 115200 -p $SERIAL get-certificates
+  cargo run -- -b 115200 -p $SERIAL get-measurements
+
```
mkeeter, line 205: Missing the end of this code fence

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+                perint,
+                txempty,
+                txnotfull,
+                rxnotempty,
+                rxfull,
+                txlvl,
+                rxlvl,
+            ) = spi.stat();
+
+            if (prev_txerr != prev_txerr)
```
mkeeter, line 487: Am I missing something, or are all the comparisons here trivially `false`?

drv/spi-msg/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "drv-spi-msg"
```
mkeeter, line 2: Since this is a client API, it should probably be named `drv-spi-msg-api` and be in a `spi-msg-api` folder.

drv/spi-msg/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+edition = "2018"
```
mkeeter, line 4: `"2021"`?

drv/stm32h7-spi-rot-server/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "drv-stm32h7-spi-rot-server"
+version = "0.1.0"
+edition = "2018"
```
mkeeter, line 4:
```suggestion
edition = "2021"
```

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+use userlib::*;
+
+task_slot!(SPI, spi_driver);
+task_slot!(SYS, sys);
+
+use ringbuf::*;
+
+const SPI_TO_ROT_DEVICE: u8 = 0;
+
+// On Gemini, the STM32H753 is in a LQFP176 package with ROT_IRQ on pin2/PE3
```
mkeeter, line 22: Can you add a config block here that will fail to compile on non-Gemini boards, so no one tries to run this on Gimlet by accident?

app/gemini-bu-rot/app.toml
```diff
@@ -118,24 +118,29 @@ task-slots = ["syscon_driver"]
 [tasks.spi0_driver]
 name = "drv-lpc55-spi-server"
 priority = 4
-max-sizes = {flash = 16384, ram = 2048}
+max-sizes = {flash = 65536, ram = 32768}
```
mkeeter, line 121: Running `cargo xtask sizes` shows this task is actually using 13K of flash and 19K of RAM.  Any idea why the RAM usage is so high?

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,24 +8,329 @@
...
+    RxState(RxState),
+    Tx(TxState, u8, usize),
+    StartTx(bool, bool),
     None,
 }
 
-ringbuf!(Trace, 64, Trace::None);
+ringbuf!(Trace, 128, Trace::None);
+
+struct TxContext<'a> {
```
mkeeter, line 154: `TxContext` and `RxContext` seem like a great place to use the [typestate pattern](http://cliffle.com/blog/rust-typestate/).

For example, in `TxContext`, you could use a richer `TxState` that only includes what's used in each phase of the state machine:
```rust
enum TxState {
    /// Idle, waiting for message to be queued.
    Idle,
    /// Message queued, waiting for start of Tx frame.
    Queued { end: u32 },
    /// Message is transferring from local buffer to FIFO.
    Writing { count: u32, end: u32 },
    /// Final bytes are being transmitted from FIFO to SPI bus.
    Finish,
    /// Error during transmit.
    Error,
}
```

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,24 +8,329 @@
...
+                }
+            }
+            RxState::Header => {
+                self.rx[self.count] = b;
+                self.count += 1;
+                ringbuf_entry!(Trace::HeaderCollect);
+                if self.count == SPI_HEADER_SIZE {
+                    ringbuf_entry!(Trace::HeaderComplete);
+                    let max_len = self.rx.len();
+                    let msg = Msg::parse(&mut *self.rx).unwrap_lite();
```
mkeeter, line 295: I don't love that `Msg::parse()` is literally always followed by `unwrap_lite` – that suggests that we could be encoding things differently.

Taking a step back, why does `RxContext` store a single slice of arbitrary size, rather than _owning_ separate slices for the header and contents, e.g.

```
struct RxContext<'a> {
    state: RxState,
    header: MsgHeader,
    buf: [u8; MAX_SPI_MSG_PAYLOAD_SIZE],
    count: usize, // cursor for receive
    end: usize,   // cursor for receive
}
```

When in the `Header` state, we could write directly to `self.header.as_bytes()[self.count]`; in `Payload` we write to `self.payload[self.count]`, and never have to do runtime checks of slice sizes (so the entire `struct Msg` is eliminated).

This is a _tiny_ bit tricker on the `Tx` side, but is still doable / worthwhile: we'd split `TxState::Responding` into `TxState::Header` and `TxState::Payload`, then write out either `self.header` or `self.payload` in the same way.

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,24 +8,329 @@
...
+            ringbuf_entry!(Trace::EnqueueTooBig);
+            return false;
+        }
+
+        let mut msg = Msg::parse(&mut *self.tx).unwrap_lite();
+        msg.set_version();
+        msg.set_len(len);
+        msg.set_msgtype(msgtype);
+
+        if let EnqueueBuf::Copy(input) = buf {
```
mkeeter, line 210: This would be more clear as an explicit `match`, so that we know that the other cases are deliberately ignored / don't need to copy any data.

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,24 +8,329 @@
...
+                        self.state = RxState::Error;
+                        //let mut buf = [0u8; SPI_HEADER_SIZE];
+                        //buf[..SPI_HEADER_SIZE].clone_from_slice(&self.rx[..SPI_HEADER_SIZE]);
+
+                        if tx.enqueue(
+                            MsgType::ErrorRsp,
+                            EnqueueBuf::Copy(&self.rx[0..SPI_HEADER_SIZE]),
+                        ) {
+                            self.state = RxState::Dispatch;
+                        }
```
mkeeter, line 308: What happens if we fail to enqueue an error message? In the equivalent case above, there's `ringbuf_entry!(Trace::EnqueueFail);`

Is it _correct_ to do nothing if we fail to enqueue an error message, vs waiting until Tx is idle then sending it? Will the host be waiting on a reply?

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
 
-    spi.drain();
     spi.enable_rx();
-    spi.enable_tx();
-
-    let mut tx_done = false;
-    let mut rx_done = false;
+    spi.disable_tx(); // Ignore Tx interrupts for now.
+    spi.send_u8(0x00); // Throw away data, but this sets bit-level framing.
+    spi.drain(); // Nothing in the FIFOs is interesting yet.
```
mkeeter, line 402: This seems odd – does this introduce a race between `0x00` being sent out by the SPI peripheral and the FIFO being cleared by `spi.drain()`?

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+    // The assert interrupt happens before the first byte of Rx data is
+    // available. That could make us more reponsive if keeping up with realtime
+    // requirements becomes an issue.
+    // On de-assert, we'll process any received message and then queue up a
+    // response for the next CSn assert.
+    spi.ssa_enable(); // Interrupt on CSn changing to asserted.
+    spi.ssd_enable(); // Interrupt on CSn changing to deasserted.
+    ringbuf_entry!(Trace::Initialized); // XXX
+
+    let mut olc = 0u16;
```
mkeeter, line 416: Add comment explaining `olc`

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+    // TODO: Should the status data returned be in some extensible format?
+    // Information should include items needed before a system is fully configured
+    // or other cases where the RoT may not be in a compliant state.
+    // The information should guide the next steps for
+    //   - RoT FW version: used to assess the need for FW update.
+    //   - Boot ROM CRC: a proxy for machine readable NXP LPC55 revision.
+    pub const CRC32: Crc<u32> = Crc::<u32>::new(&CRC_32_CKSUM);
+    let bootrom_crc32 = CRC32.checksum(&bootrom().data[..]).to_le_bytes();
+
+    let (
```
mkeeter, line 428: I already said this, but this would be great as a `struct LStat`. Plus, then you could `#[derive(Eq, PartialEq)]` and not have to compare all of the values individually down below!

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
-                rx_cnt += 1;
-                if rx_cnt == rx.len() {
-                    rx_done = true;
+            // TODO: check for tx underflow.
+            // TODO: check for rx overflow.
+            // TODO: Queue a failure message on those conditions.
+            // TODO: Clear the error condition.
+
+            // If we are asserting ROT_IRQ, check for conditions to de-assert.
+            if tctx.rot_irq && tctx.state == TxState::Finish {
```
mkeeter, line 556: Is there any case where `tctx.state == TxState::Finish` but `tctx.rot_irq != true`?

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+            // several bytes left. Get them now because it is then time to
+            // process the received message.
+            loop {
+                if !spi.has_byte() {
+                    break;
+                }
+                let (b, csn, sot) = spi.read_u8_csn_sot(); // always read
+                ringbuf_entry!(Trace::Rx(rctx.state, b, rctx.count, csn, sot));
+                rctx.rx_byte(b, sot, &mut tctx);
+                if rctx.state == RxState::Header
```
mkeeter, line 609: Why does this conditional only care if `rctx.state == RxState::Header`? Isn't receiving payload bytes while transmitting _also_ bad?

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+                // Process transitory states
+                if rctx.state == RxState::Dispatch {
+                    // Receive is complete.
+                    // Processing may take a long time and detract from keeping up
+                    // with Tx and Rx FIFOs.
+                    // Processing has been delayed until CSn was de-asserted.
+                    rctx.state = RxState::Responding;
+                    ringbuf_entry!(Trace::Responding);
+
+                    if tctx.state == TxState::Idle {
```
mkeeter, line 660: Is there any way for this condition not to be true?

More generally: it seems like splitting `TxState` and `RxState` up is a mistake.  Having separate enums gives us `N * M` possible state combinations, many of which are nonsensical.  In practice, we want a single unified state machine that's something like
```
enum State {
    Idle,
    RxHeader { count: u32 },
    RxPayload  { count: u32, end: u32 },
    Dispatch,
    TxHeader  { count: u32 },
    TxPayload  { count: u32, end: u32 },
    Invalid,
    Error,
}
```
which gives is `N + M` states instead of `N * M` (and all of the states are meaningful)

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+                                tctx.enqueue(
+                                    MsgType::ErrorRsp,
+                                    EnqueueBuf::Empty,
+                                );
+                            }
+                        }
+                    }
+                }
+            }
             if !again {
```
mkeeter, line 701: It seems like the inner loop can run multiple times without returning to the kernel, and the inner loop also checks and clears interrupt flags on its own.  However, the inner loop leaves (hardware) interrupts on, which means we could have a somewhat confusing series of events:

- Inner loop runs, setting `again = true`
- Interrupt fires (e.g. SSA)
  - Kernel marks that interrupt as fired
- Inner loop runs again, noticing the interrupt and clearing its bit
- Inner loop ends
- Return to outer loop, re-enabling irqs from the kernel
- The kernel immediately notifies us of the IRQ that we already saw + handled

This suggests to me that we should keep (hardware) interrupts disabled when in the inner loop, then re-enable them before enabling (kernel) interrupts.

drv/spi-msg/README.md
```diff
@@ -0,0 +1,218 @@
...
+only one of which is available in production:
+
+  - An SWD interface that can be used for development.
+  - A serial port for personalization during manufacturing.
+  - A SPI link to the SP for production use.
+
+The use and security of the SWD and serial interfaces is out of scope for this
+document.
+
+SWD from RoT to SP is also used to securely bring-up the SP and can be used to establish shared data and retrieve SP measurements, but is not available beyond early SP boot.
```
mkeeter, line 16: Nit: overly long line

drv/spi-msg/README.md
```diff
@@ -0,0 +1,218 @@
...
+  - `DB9:3(TXD)` to `J302:5(RS232_RX)`
+  - `DB9:5(GND)` to `J302:9(GND)`
+
+DB9, looking at the pins:
+```
+ ___________
+( 1 2 3 4 5 )
+ \ 6 7 8 9 /
+  \_______/
+```
```
mkeeter, line 182: I love this diagram!

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+use drv_spi_msg::*;
+use drv_stm32xx_sys_api as sys_api;
+use userlib::*;
+
+task_slot!(SPI, spi_driver);
+task_slot!(SYS, sys);
+
+use ringbuf::*;
+
+const SPI_TO_ROT_DEVICE: u8 = 0;
```
mkeeter, line 20: Nit: add comment that this is based on the `app.toml` file

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+
+// On Gemini, the STM32H753 is in a LQFP176 package with ROT_IRQ on pin2/PE3
+const ROT_IRQ: sys_api::PinSet = sys_api::PinSet {
+    port: sys_api::Port::E,
+    pin_mask: 1 << 3,
+};
```
mkeeter, line 26:
```suggestion
const ROT_IRQ: sys_api::PinSet = sys_api::Port::E.pin(3);
```

drv/lpc55-spi-server/src/main.rs
mkeeter: Alternatively, you could make the `struct Msg` an owner of its data:
```
#[derive(FromBytes, AsBytes, Unaligned, Debug)]
#[repr(C)]
pub struct Msg {
    header: MsgHeader,
    payload: [u8; MAX_SPI_MSG_PAYLOAD_SIZE],
}
```
then use `msg.as_bytes()[...]` when reading / writing to it.  Either way, it eliminates the runtime checks in `Msg::parse`.

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#![no_std]
+#![no_main]
+
+use idol_runtime::{ClientError, Leased, RequestError, R, W};
+// use idol_runtime::{NotificationHandler, ClientError, Leased, RequestError, R, W};
```
mkeeter, line 9: Nit: delete dead code

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        sys,
+        spi,
+        message: *message,
+    };
+
+    loop {
+        idol_runtime::dispatch(&mut buffer, &mut server);
+    }
+}
+
```
mkeeter, line 68: Nit: it would be nice if `struct ServerImpl` was defined up here, to make it easier to read the code.

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        MsgType::EchoReq | MsgType::SprocketsReq | MsgType::StatusReq => (),
+        _ => {
+            ringbuf_entry!(Trace::Line);
+            return Err(MsgError::BadMessageType);
+        }
+    }
+    let xmit_len = SPI_HEADER_SIZE + msg.payload_len();
+    ringbuf_entry!(Trace::Line);
+
+    let port = server.sys.gpio_read_input(ROT_IRQ.port).unwrap_lite();
```
mkeeter, line 90: You can read in a single pass: `server.sys.gpio_read(ROT_IRQ)`, which returns a masked `u16`.  I see that you did this below then commented it out, not sure why.

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        spi,
+        message: *message,
+    };
+
+    loop {
+        idol_runtime::dispatch(&mut buffer, &mut server);
+    }
+}
+
+fn do_send_recv(server: &mut ServerImpl) -> Result<usize, MsgError> {
```
mkeeter, line 69: A docstring here would be nice, explaining what this function is supposed to do.

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        return Err(MsgError::SpiServerError);
+    }
+
+    // We sleep and poll for what should be long enough for the RoT
+    // to queue a response.
+    // TODO: For better performance and power efficiency,
+    // take an interrupt on ROT_IRQ falling edge with timeout.
+    const LIMIT_START: u64 = 1_000_000; // TODO: Something more reasonable.
+    const NAP: u64 = 1;
+    let mut limit = LIMIT_START; // 1.04ms Time limit is somewhat arbitrary.
```
mkeeter, line 126: `sleep_for` is in terms of kernel ticks, not clock ticks. Each kernel tick is 1 ms, so this will actually wait for 16 minutes!

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+        limit -= 1;
+        if limit == 0 {
+            ringbuf_entry!(Trace::RotIrqTimeout);
+            break;
+        }
+        hl::sleep_for(NAP);
+    }
+
+    /*
```
mkeeter, line 141: Remove dead code / comment here, otherwise it's confusing.

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+    {
+        ringbuf_entry!(Trace::SpiError(spi_error));
+        server.spi.release().unwrap_lite();
+        return Err(MsgError::SpiServerError);
+    }
+    server.spi.release().unwrap_lite();
+
+    let msg = Msg::parse(&mut server.message[0..rlen + SPI_HEADER_SIZE])
+        .unwrap_lite();
+    match msg.payload_get() {
```
mkeeter, line 207: Nit: this could be `msg.payload_get()?.len()`

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        server.message[0],
+        server.message[1],
+        server.message[2],
+        server.message[3]
+    ));
+
+    let msg = Msg::parse(&mut server.message[..]).unwrap_lite();
+    if !msg.is_supported_version() {
+        ringbuf_entry!(Trace::BadProtocol(server.message[0]));
+        server.spi.release().unwrap_lite();
```
mkeeter, line 185: What happens on the client side here? Does it end up confused if CS becomes deasserted midway through a transmission?

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+    sys_irq_control(self.interrupt, true);
+    // And wait for it to arrive.
+    let _rm =
+        sys_recv_closed(&mut [], self.interrupt, TaskId::KERNEL)
+            .unwrap_lite();
+    */
+
+    // Read just the header.
+    // Keep CSn asserted over the two reads.
+    server
```
mkeeter, line 162: Not related to your PR, but it sure would be nice to have an RAII handle for SPI locking...

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        msg.set_version();
+        msg.set_len(source.len());
+        msg.set_msgtype(msgtype);
+        ringbuf_entry!(Trace::SendRecv(
+            msgtype,
+            source.len(),
+            msg.payload_buf().len()
+        ));
+        if source.len() > msg.payload_buf().len() {
+            ringbuf_entry!(Trace::Line);
```
mkeeter, line 241: What's the purpose of this tracing? Shouldn't this return an error?

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+                &mut msg.payload_buf()[0..source.len()],
+            )
+            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+        ringbuf_entry!(Trace::Line);
+
+        // Send message, then receive response using the same local buffer.
+        do_send_recv(&mut *self)?;
+        ringbuf_entry!(Trace::Line);
+
+        let msg = drv_spi_msg::Msg::parse(&mut self.message[..]).unwrap_lite();
```
mkeeter, line 256: Probably want to return an error if the `sink` lease isn't big enough to contain the whole payload

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+                prev_txerr = txerr;
+                prev_rxerr = rxerr;
+                prev_perint = perint;
+                prev_txempty = txempty;
+                prev_txnotfull = txnotfull;
+                prev_rxnotempty = rxnotempty;
+                prev_rxfull = rxfull;
+                prev_txlvl = txlvl;
+                prev_rxlvl = rxlvl;
+            }
```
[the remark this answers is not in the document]
lzrd: Yes. I'm not ready to remove debugging code until the implementation has proven itself on Gimlet and over time.

drv/lpc55-spi-server/src/main.rs
lzrd: This is a bit of redundant documentation. I'll clean this up or refer to the complete version.

app/gemini-bu-rot/app.toml
lzrd: Sprockets gets linked into this task. The flash limit can be reduced.

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,24 +8,329 @@
...
 use ringbuf::*;
+
+use crc::{Crc, CRC_32_CKSUM};
+use lpc55_romapi::bootrom;
 use userlib::*;
 
 task_slot!(SYSCON, syscon_driver);
 task_slot!(GPIO, gpio_driver);
 
+// See drv/spi-msg/README.md
```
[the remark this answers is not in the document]
lzrd: I'm not at peace with the naming yet.
I'm thinking that on the SP side, there should be a single task for the SP:spi4 driver that incorporates the low level SPI and the protocol instead of a spi server and a spi-rot server.
The UART to SP task might be able to be merged as well, so one task would own a SPI and UART interface as well as an Idl interface. I'm not ready for that level of complexity yet if that were actually a good idea.

drv/lpc55-spi-server/src/main.rs
lzrd: u8 encoding is not necessary. From for Rx/TxState also not necessary.

drv/lpc55-spi-server/src/main.rs
lzrd: done

drv/lpc55-spi-server/src/main.rs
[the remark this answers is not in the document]
mkeeter: Hmmm, do we have a name for the "SP to RoT" channel?

I don't like using `2` for "to" – how does `{stm32h7,lpc55}-spi-rot-server` sound?  It's a bit shorter and still pretty unambiguous, since on both sides the server is handling RoT-related SPI communication.

drv/lpc55-spi-server/src/main.rs
lzrd: added a more detailed description.

drv/lpc55-spi-server/src/main.rs
lzrd: I took your suggestions and am happier with the code now. Thanks.

drv/spi-msg/Cargo.toml
lzrd: I did a rename of all spirot variants to "sprot" with drv/spi-msg changing to drv/sprot-api.

drv/spi-msg/Cargo.toml
lzrd: done

drv/spi-msg/README.md
```diff
@@ -0,0 +1,218 @@
...
+  - A retry scheme may be useful to improve reliability.
+  - The SP may need to make a trade-off between SPI clock speed and reliability.
+	  + It may be useful to do retries at slower speeds.
+	  + Logging of retries, clock speed changes, and other information related to reliability is needed to make any improvements and to identify problems early.
+  - Ensure that bytes transmitted beyond payload length are properly ignored/processed (log unexpected conditions, do not impede driving towards compliance).
+  - Agree on identification information needed at earliest, unsecured boot for manufacturing and production workflows.
+  - Agree on RoT firmware update workflows.
+  - If more a more detailed MsgType::Error payload is useful, work that out early.
+  - Complete testing is needed to ensure that the interface has a high degree of stability and reliability.
+  - This doc belongs in an RFD.
```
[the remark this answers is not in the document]
lzrd: I've fixed what I was able to spot on my own.

drv/spi-msg/src/lib.rs
lzrd: I was able to get rid of the issues I had by using accessors outside of sprot-api (nee spi-msg) and using u8 as the internal type.

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
-
     sys_irq_control(1, true);
 
-    let mut tx = &mut a_bytes;
-    let mut rx = &mut b_bytes;
-
-    let mut tx_cnt = 0;
-    let mut rx_cnt = 0;
+    // Configure ROT_IRQ
+    let rot_irq = Pin::PIO0_18; // XXX Should be in app.toml
```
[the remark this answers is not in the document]
lzrd: There was already a use of the config block. I extended it to include optional naming of a pin.

drv/lpc55-spi-server/src/main.rs
lzrd: Done

drv/lpc55-spi-server/src/main.rs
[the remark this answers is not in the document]
lzrd: struct LStat reduces verbosity.

drv/lpc55-spi-server/src/main.rs
lzrd: You are correct.

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+                    ringbuf_entry!(Trace::Line);
+                    // Note: SP is going to be edge-triggered, so this signal
+                    // does not need to be asserted for the entire transaction.
+                    // However, it is nice to see it on the logic analyzer for
+                    // the whole frame.
+                    gpio.set_val(rot_irq, Value::One).unwrap_lite();
+                    ringbuf_entry!(Trace::DeassertRotIrq);
+                    tctx.rot_irq = false;
+                    tctx.state = TxState::Idle;
+                    ringbuf_entry!(Trace::DeassertRotIrq);
```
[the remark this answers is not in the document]
lzrd: remvoed

drv/lpc55-spi-server/src/main.rs
```diff
@@ -63,74 +369,339 @@ fn main() -> ! {
...
+                    // If we have nothing to send in the next frame, then
+                    // the LPC55 will send a byte of unknown origin.
+                    spi.send_u8(0); // write our no-op byte.
+                    spi.drain_tx(); // Throw away byte just written.
```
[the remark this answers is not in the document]
lzrd: This seems to be a quirk of the LPC55 SPI hardware.
If the FIFO does not have some last explicit contents, it will clock out what seems to be garbage.
This is visible on a logic analyzer, but we may also find some corner case where the SP is clocking out data
and LPC55 is out of sync. Rather than waste time, (like I did when discovering this quirk), default to zeros.

drv/stm32h7-spi-rot-server/src/main.rs
lzrd: Yeah. That wasn't right. Hence the TODO.

drv/stm32h7-spi-rot-server/src/main.rs
lzrd: The CSn assertion bounds the SPI frame. If the message doesn't validate, there will be an error response.

drv/stm32h7-spi-rot-server/src/main.rs
lzrd: Errors now handled.

drv/stm32h7-spi-rot-server/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        source
+            .read_range(
+                0..source.len(),
+                &mut msg.payload_buf()[0..source.len()],
+            )
+            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+        ringbuf_entry!(Trace::Line);
+
+        // Send message, then receive response using the same local buffer.
+        do_send_recv(&mut *self)?;
```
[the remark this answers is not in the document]
lzrd: yes

drv/stm32h7-spi-rot-server/src/main.rs
lzrd: write_range does a range check and returns error.

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -0,0 +1,593 @@
...
+        let rot_irq_begin = self.is_rot_irq_asserted();
+        self.spi
+            .lock(CsState::Asserted)
+            .map_err(|_| MsgError::CannotAssertCSn)?;
+        if delay != 0 {
+            hl::sleep_for(delay);
+        }
+        self.spi.release().unwrap_lite();
+        if delay_after != 0 {
+            hl::sleep_for(delay); // TODO: make this a 2nd parameter?
```
jgallagher, line 361:
```suggestion
            hl::sleep_for(delay_after);
```

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -0,0 +1,593 @@
...
+    // Poll ROT_IRQ until asserted (true) or deasserted (false).
+    fn wait_rot_irq(&mut self, desired: bool, max_sleep: u32) -> bool {
+        let mut slept = 0;
+        while self.is_rot_irq_asserted() != desired {
+            if slept == max_sleep {
+                ringbuf_entry!(Trace::RotReadyTimeout);
+                return false;
+            }
+            hl::sleep_for(1);
+            slept += 1;
```
jgallagher, line 384: IIRC `hl::sleep_for()` only promises that it sleeps _at least as long as_ the argument we pass it, so this could potentially end up sleeping for much longer than `max_sleep`. It would be a bit more complex, but should we convert `max_sleep` to a deadline on entry to this function and break based on how much time we've actually slept?

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -0,0 +1,593 @@
...
+                                    }
+                                },
+                                MsgType::ErrorRsp => {
+                                    Err(MsgError::Unknown)
+                                },
+                                _ => {
+                                    // Other codes from RoT
+                                    // are not recoverable
+                                    // with a retry.
+                                    Err(MsgError::Unknown)
```
jgallagher, line 544: Should this be 

```suggestion
                                    Err(MsgError::BadMessageType)
```
?

idl/sprot.idol
```diff
@@ -0,0 +1,72 @@
...
+        "send_recv": (
+            doc: "Send then receive a message from the RoT",
+            args: {
+                "msgtype": (
+                    type: "MsgType",
+                    recv: FromPrimitive("u8"),
+                ),
+            },
+            leases: {
+                "message": (type: "[u8]", read: true),
```
jgallagher, line 15: Do we want `max_len` caps on all the leases in this interface?

jgallagher (commented):
A kinda vague comment that may not be very actionable - I know there's a lot of `if let Some(..) = ...` to reduce the number of possible panics, but I find that style quite a bit more difficult to read. It doesn't seem easy to make it obvious when those `if`s should always be true vs when they're legitimate runtime conditions, which makes it hard to decide whether the `else` (or lack of an `else`) is correct. I wonder if there's some middle ground here that reads more like the panicking style but doesn't have the text overhead.

app/gimlet-rot/app.toml
```diff
@@ -65,27 +65,38 @@ pins = [
...
+{ pin = { port = 0, pin = 26}, alt = 9, mode = "NoPull"},
 # HS_SPI_MISO = P1_3 = FUN6
-{ pin = { port = 1, pin = 3}, alt = 6},
+{ pin = { port = 1, pin = 3}, alt = 6, mode = "NoPull"},
 # HS_SPI_SCK = P1_2 = FUN6
-{ pin = { port = 1, pin = 2}, alt = 6},
+{ pin = { port = 1, pin = 2}, alt = 6, mode = "NoPull"},
 # HS_SSEL1 = P1_1 = FUN5
-{ pin = { port = 1, pin = 1}, alt = 5},
+{ pin = { port = 1, pin = 1}, alt = 5, mode = "NoPull"},
```
labbott, line 97: NoPull should be the default, do you actually need to specify it here?

drv/lpc55-spi/src/lib.rs
```diff
@@ -196,11 +229,7 @@ impl Spi {
         len_bits: u8,
     ) {
         // SPI hardware only supports lengths of range 4-16 bits
-        #[allow(clippy::manual_range_contains)]
-        if len_bits > 16 || len_bits < 4 {
-            panic!()
-        }
-
+        assert!((4..=16).contains(&len_bits));
```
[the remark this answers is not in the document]
lzrd: It's my understanding that the compiler would elide this test in a release build. All the calls to this fn are using constants for this field. This shows intent without affecting production code size.

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -0,0 +1,426 @@
...
+// the Tx buffer.
+//
+
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub enum IoStatus {
+    Flush,
+    IOResult { overrun: bool, underrun: bool },
+}
+
+#[repr(C, align(8))]
```
labbott, line 76: Why the 8 byte align?

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -0,0 +1,426 @@
...
+        self.rxcount = 0;
+        self.txremainder = 0;
+        self.overrun = false;
+        self.underrun = false;
+
+        if !transmit {
+            // Ensure that unused Tx buffer is zero-filled.
+            // TODO: Is there a way to do this that is more efficient while in
+            // critical real-time sections? DMA may require this too.
+            self.tx.fill(0);
```
labbott, line 238: I'd be interested what this is getting to compiled at the moment

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -0,0 +1,426 @@
...
+
+                // Note that while Tx is done at end of frame,
+                // FIFORD may still have bytes to read.
+                if intstat.ssd().bit() {
+                    self.spi.ssd_clear();
+                    // ROT_IRQ will only be asserted if there was a Tx
+                    // message from RoT to SP.
+                    // For simplicity, deassert unconditionally.
+                    if self.gpio.set_val(ROT_IRQ, Value::One).is_err() {
+                        // XXX this can never happen
```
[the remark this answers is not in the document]
labbott: I'd make this `unreachable!()` or `panic!()` if we're really asserting this can never happen

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+fn turn_on_flexcomm(syscon: &Syscon) {
+    // HSLSPI = High Speed Spi = Flexcomm 8
+    // The L stands for Let this just be named consistently for once
+    syscon.enable_clock(Peripheral::HsLspi).unwrap_lite();
+    syscon.leave_reset(Peripheral::HsLspi).unwrap_lite();
+
+    syscon.enable_clock(Peripheral::Fc3).unwrap_lite();
+    syscon.leave_reset(Peripheral::Fc3).unwrap_lite();
```
labbott, line 423: Can go ahead and remove this since we're only using the HS-SPI

drv/lpc55-spi/src/lib.rs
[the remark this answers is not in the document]
jgallagher: Asserts are not removed from release builds. `debug_assert!()` are removed from release builds, but we only make release builds. :)

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -0,0 +1,593 @@
...
+            //
+            // The RoT must be able to observe SP resets.
+            // During the normal start-up seqeunce, the RoT is controlling the
+            // SP's boot up sequence. However, the SP can reset itself and
+            // individual Hubris tasks may fail and be restarted.
+            // then we'll see ROT_IRQ asserted without having first sent a message.
+            // If SP and RoT are out of sync, e.g. this task restarts and an old
+            // response is still in the RoT's transmit FIFO, then we can also see
+            // ROT_IRQ asserted when not expected.
+            // TODO: configuration parameters for delays below
```
labbott, line 132: I spent a lot of time thinking about this case and what it means for both the boot flows of the RoT and SP. I see two plausible scenarios
1) RoT and SP always boot asynchronously and the protocol handles this
2) On power up the RoT always resets the SP to ensure it is in a known state.

I thought we were going the async route which may be a sign this discussion needs to RFD 317 for discussion.

drv/sprot-api/src/lib.rs
```diff
@@ -0,0 +1,490 @@
...
+//
+// It may mean that the two parties are out of phase, some RoT task
+// is hampering the sprot task from meeting its realtime requirements,
+// or there is some other bug.
+pub const VERSION_BUSY: u8 = 0xB2;
+
+/// SPI Message types will allow for multiplexing and forward compatibility.
+#[derive(Copy, Clone, Eq, PartialEq, FromPrimitive, Unaligned, AsBytes)]
+#[repr(u8)]
+pub enum MsgType {
```
labbott, line 102: Is this where I would add a message type for sending an RoT update from the SP? Put another way, we had talked about the RoT injecting a session key over SWD, is this where we would add that message type?

drv/sprot-api/README.md
```diff
@@ -0,0 +1,510 @@
...
+for the test.
+
+#### Test StatusReq and StatusRsp:
+
+The exact values in the status structure are expected to change and stabilize
+before first customer ship.
+
+/Discussion on contents are appreciated/.
+
+  $ cargo xtask humility $APP -- sprot --status
```
jgallagher, line 431: Similar to above - I think we can do this via hiffy (`hiffy -c SpRot.status`).

drv/sprot-api/README.md
```diff
@@ -0,0 +1,510 @@
...
+
+```sh
+SERIAL=/dev/ttyUSB0
+```
+
+In a [sprockets workspace](https://github.com/oxidecomputer/sprockets):
+
+```sh
+  cargo run -- -b 115200 -p $SERIAL get-certificates
+  cargo run -- -b 115200 -p $SERIAL get-measurements
```
jgallagher, line 488: I don't see any uart config / setup in this PR - how is this making its way to sprot?

drv/sprot-api/src/lib.rs
```diff
@@ -0,0 +1,490 @@
...
+    /// Supply RoT status.
+    // TODO: decide on appropriate content for the StatusRsp message payload.
+    StatusRsp = 5,
+    /// Payload contains a sprockets request.
+    SprocketsReq = 6,
+    /// Payload contains a sprockets response.
+    SprocketsRsp = 7,
+    /// RoT sinks this message sending back a SinkReq with no payload.
+    SinkReq = 8,
+    /// Acknowledge recepit of SinkReq
```
andrewjstone, line 122: Typo: receipt

drv/sprot-api/src/lib.rs
andrewjstone: Possibly. I envisioned all authenticated messages would go through sprockets communication originally. Although currently, messages between the SP and RoT are just proxied messages from the host. https://github.com/oxidecomputer/sprockets/blob/main/common/src/msgs.rs

I'm certainly open to using this message path for this, but I'd like to open the option of converting all this hand-written serialization code to just use hubpack instead.

drv/lpc55-sprot-server/src/handler/sprockets.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub fn init() -> RotSprocket {
+    let manufacturing_keypair = salty::Keypair::from(&random_buf());
+    let config = RotConfig::bootstrap_for_testing(
+        &manufacturing_keypair,
+        salty::Keypair::from(&random_buf()),
+        SerialNumber(random_buf()),
+    );
+    RotSprocket::new(config)
+}
```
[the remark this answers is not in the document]
andrewjstone: This code is many months old, and was written to provide a way for sprockets to work before the DICE stuff and actual certs were provisioned. The real certs provisioned to the RoT will have to be used here. Furthermore, we're changing from the placeholder cert code to use x509 certs: https://github.com/oxidecomputer/sprockets/issues/36

drv/spi-msg/src/lib.rs
lzrd: Most of the manual serialization/serialization throughout the code has been addressed through use of hubpack.

drv/stm32h7-spi-rot-server/src/main.rs
lzrd: Done

drv/lpc55-spi/src/lib.rs
```diff
@@ -157,6 +164,32 @@ impl Spi {
         self.reg.fifointenclr.write(|w| w.rxlvl().set_bit());
     }
 
+    pub fn ssa_enable(&mut self) {
+        self.reg.intenset.write(|w| w.ssaen().enabled());
+    }
+
+    pub fn ssa_disable(&mut self) {
+        self.reg.intenclr.write(|w| w.ssaen().set_bit());
```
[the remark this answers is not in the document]
lzrd: This is a register where only writing ones has an effect. Set_bit is clearer for me.

drv/lpc55-spi/src/lib.rs
```diff
@@ -157,6 +164,32 @@ impl Spi {
...
+    pub fn ssa_clear(&mut self) {
+        self.reg.stat.write(|w| w.ssa().set_bit());
+    }
+
+    pub fn ssd_enable(&mut self) {
+        self.reg.intenset.write(|w| w.ssden().enabled());
+    }
+
+    pub fn ssd_disable(&mut self) {
+        self.reg.intenclr.write(|w| w.ssden().set_bit());
```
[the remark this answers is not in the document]
lzrd: similar answer. Does it make sense or need a clarifying comment/reference to docs?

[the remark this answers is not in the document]
jgallagher: I think two things happened when I read this the first time (and similarly for the above case):

1. I thought this was acting on the same register as on line 181; I missed that this method writes to `intenclr` while the other writes to `intenset`
2. I was confused by the difference on 181/185 where one was `w.ssden().enabled()` and the other was `w.ssden().set_bit()`, compounded by the previous point

I don't have a strong preference between `.enabled()` and `.set_bit()`, it seems a little surprising to see both options chosen back-to-back. Could the `.enabled()`s be changed to `.set_bit()` if your preference is for `.set_bit()`?

drv/stm32h7-sprot-server/src/main.rs
lzrd: Sleeping longer affects performance, but is tolerable. To tune this properly, we need to know the LPC55 interrupt response times. They all should be sub-millisecond, but the variability remains unknown.
There is certainly room for improvement. Retries by the SP have worked well so far.

drv/stm32h7-sprot-server/src/main.rs
lzrd: Done

app/gimlet-rot/app.toml
lzrd: Done

drv/lpc55-sprot-server/src/main.rs
lzrd: removed

drv/lpc55-sprot-server/src/main.rs
lzrd: Done

drv/stm32h7-sprot-server/src/main.rs
lzrd: Moving this to the RFD but there remains handling of the unexpected IRQ asserted.

drv/sprot-api/src/lib.rs
lzrd: All converted to hubpack whenever practical. (1-byte payloads and the sink test messages are not hubpack).

drv/sprot-api/README.md
```diff
@@ -0,0 +1,510 @@
...
+The RoT `EchoReq` and `EchoRsp` handling can be tested through the SP using
+the existing humility spi command:
+
+#### Test EchoReq and EchoRsp:
+
+```sh
+  $ SPI=3 # if Gimletlet
+  $ SPI=4 # if Gemini
+  $ APP=<path to app.toml>
+  $ cargo xtask humility $APP -- spi -p $SPI -w 1,2,1,0,35,236,249,41,9
```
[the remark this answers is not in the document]
lzrd: The readme has been updated with a working example.

drv/sprot-api/README.md
lzrd: Updated. The hiffy calls are now used in the examples. My changes to humility aren't needed. However, humility doesn't deal with read plus write leases, so a change there may be needed.

app/gemini-bu/app.toml
```diff
@@ -137,12 +137,20 @@ task-slots = ["sys"]
...
 stacksize = 2048
 start = true
-task-slots = ["sys", "i2c_driver", "hf", "hash_driver"]
+task-slots = ["sys", "i2c_driver", "hf", "hash_driver", "sprot"]
+
+[tasks.sprot]
+name = "drv-stm32h7-sprot-server"
+priority = 5
+max-sizes = {flash = 16384, ram = 32768}
+stacksize = 16384
```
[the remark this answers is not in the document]
lzrd: The stack size on the SP was inflated. I can take it down to 4K based on running the tests and what humility stackmargin says.
On the RoT, it is pulling Sprockets into the task. That plus other un-identified things push the stack just over 8K.
I expect it will improve over time.

drv/lpc55-spi/src/lib.rs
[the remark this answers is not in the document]
lzrd: That seems more consistent for this set of registers. I can do that.

drv/stm32h7-sprot-server/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+ringbuf = {path = "../../lib/ringbuf"}
+zerocopy = "0.6.1"
+num-traits = { version = "0.2.12", default-features = false }
+drv-spi-api = {path = "../../drv/spi-api"}
+drv-sprot-api = {path = "../../drv/sprot-api"}
+drv-stm32xx-sys-api = {path = "../../drv/stm32xx-sys-api", features = ["family-stm32h7"]}
+idol-runtime = {git = "https://github.com/oxidecomputer/idolatry.git"}
+cfg-if = "1"
+if_chain = "1"
+# byteorder = {version = "1.4", default-features = false}
```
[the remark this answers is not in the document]
lzrd: yes

idl/sprot.idol
lzrd: The source of truth for the limit is in drv/sprot-api/src/lib.rs: PAYLOAD_SIZE_MAX which, for SP and RoT compiled in the same release, is the same value.
We could change this between releases and during RoT update, the SP and RoT may have a value for PAYLOAD_SIZE_MAX that is different from each other.
Normally, one would use the constants generated by idolatry to size the buffers in the implementation.
The RoT doesn't offer the SpRot idol client interface and doesn't have those idol constants handy.
I don't want to have two sources of truth to be maintained.
Should I still put limits here?

app/rot-carrier/app.toml
```diff
@@ -137,37 +139,22 @@ pins = [
...
-start = true
-task-slots = [{peer = "pong"}, "usart_driver"]
-
-[tasks.pong]
-name = "task-pong"
-priority = 5
-max-sizes = {flash = 8192, ram = 1024}
-start = true
-task-slots = ["user_leds"]
-
```
labbott, line 158: Did those get removed intentionally? It's nice to see the comforting blinking lights when testing on the carrier if we still have space.

app/gimlet-rot/app-b.toml
```diff
@@ -125,9 +139,13 @@ spi_num = 5
 [tasks.sp_measure]
 name = "task-sp-measure"
 priority = 5
-max-sizes = {flash = 32768, ram = 8192}
+max-sizes = {flash = 131072, ram = 8192}
 task-slots = ["swd"]
 stacksize = 2048
```
labbott, line 144: Given sp_measure was always supposed to be a sample task I'm not sure if it's worth it to increase the flash size vs just removing it if actually requires more flash.

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -0,0 +1,420 @@
...
+            self.gpio.set_val(ROT_IRQ, Value::Zero).unwrap_lite();
+            ringbuf_entry!(Trace::RotIrqAssert);
+        }
+
+        // TODO: SP RESET needs to be monitored, otherwise, any
+        // forced looping here could be a denial of service attack against
+        // observation of SP resetting. SP resetting without invalidating
+        // security related state means a compromised SP could operate using
+        // the trust gained in the previous session.
+        // Upper layers may mitigate that, but check on it.
```
labbott, line 268: Do we expect SP RESET to be monitored here in this driver or another driver?

app/gemini-bu-rot/app.toml
lzrd: I'll leave further memory footprint work for a future PR.

idl/sprot.idol
lzrd: This is an open question. I'll think about this for a future PR.

drv/lpc55-spi/src/lib.rs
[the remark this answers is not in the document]
lzrd: I put it back the way you had it.

drv/lpc55-update-server/src/main.rs
```diff
@@ -168,6 +168,13 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
         })
     }
+
+    fn clock_speed(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<u32, RequestError<UpdateError>> {
+        Err(UpdateError::Unknown.into()) // TODO
+    }
 }
```
labbott, line 178: What's the intention here? Eventually get the clock speed so the SP knows what to run at? If it's not being used currently I'd prefer to just see it removed.

idl/sprot.idol
```diff
@@ -0,0 +1,133 @@
...
+                "count": "u16",
+                "size": "u16",
+            },
+            reply: Result(
+                ok: "SinkStatus",
+                err: CLike("MsgError"),
+            ),
+        ),
+
+        // The RoT update API is copy and pasted from idl/update.idol.
```
labbott, line 73: It would be nice if there was a way to include that. Definitely long term work/wish.

drv/sprot-api/README.md
```diff
@@ -0,0 +1,510 @@
...
+#### Test EchoReq and pulse CSn to clear RoT Tx and `ROT_IRQ`:
+
+Same as above, but pulse CSn to clear the RoT Tx buffer.
+
+```sh
+$ cargo xtask humility $APP -- spi -p $SPI -w 1,2,1,0,35,236,249,41,9
+...
+[Ok([])]
+
+$ cargo xtask humility $APP -- sprot --pulse 10
```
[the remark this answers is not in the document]
lzrd: I found that the hiffy calls are sufficient for testing. I don't need to merge my custom humility commands. The drv/sprot-api/README.md file has been updated with commands that will work for you.

task/hiffy/src/common.rs
```diff
@@ -514,6 +518,236 @@ pub(crate) fn spi_write(
     Ok(0)
 }
 
+#[cfg(feature = "sprot")]
+task_slot!(SPROT, sprot);
+
+#[cfg(feature = "sprot")]
+pub(crate) fn sprot_send_recv(
```
labbott, line 525: There are warnings here about unused functions, can you double check config settings?

drv/stm32h7-update-server/src/main.rs
```diff
@@ -388,6 +388,13 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
             version: HUBRIS_BUILD_VERSION,
         })
     }
+
+    fn clock_speed(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<u32, RequestError<UpdateError>> {
+        Err(UpdateError::Unknown.into()) // Not implemented for the STM32
+    }
```
labbott, line 397: Since this doesn't seem to be used currently can we just remove this API and add it later?

labbott (approved):
Some very minor comments but overall LGTM

drv/stm32h7-update-server/src/main.rs
lzrd: I'll remove it.

drv/lpc55-sprot-server/src/main.rs
lzrd: This is one place it could be monitored, but only because it is monitoring other things that come from the SP.
My thinking is that sprockets is linked into this task and may have state that needs to be invalidated.
There will be work to do elsewhere, so I don't think that sprockets is forcing the decision.

drv/lpc55-update-server/src/main.rs
lzrd: RoT clock speed affects the data rate that the SP can safely use for the SPI interface to RoT.
The SP could make a runtime decision on that.
I'm removing this for now.

idl/sprot.idol
lzrd: Yes.

app/rot-carrier/app.toml
lzrd: I'll look. At one time I was concerned about the crashing one interfering with real time stuff.

lzrd: I put them back in with priority changes to keep them from preempting anything else.

app/gimlet-rot/app-b.toml
lzrd: I made this consistent with other app.toml files. I should have left it alone, but it can be a clean-up PR later.

PR #620, opened 2022-06-21
Support for Meanwell application

bcantrill:
Draft commit message:

```
This adds support for a Meanwell application definition for Gimletlet,
which allows for a Gimletlet to be easily used to manage Meanwell
power supplies on a benchtop. There is nothing particularly
Meanwell-specific (or indeed even power-specific) here; this is really
just a thin veneer on the GPIOs present on the J16 header on the
Gimletlet, allowing them to be used to manage external power supplies
without forcing the user to wade into the intricacies of GPIO pin
configuration.  Note that the Meanwell application also indicates a
different LED pattern (a quick blue LED blink, once per second) to
make it clear from across a room that a particularly Gimletlet is
serving to manage Meanwell supplies.
```

### DIFF
--- a/drv/meanwell/src/main.rs
+++ b/drv/meanwell/src/main.rs
@@ -2,12 +2,21 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+//!
+//! An exceedingly simple driver to deal with the management of Meanwell power
+//! supplies from a Gimletlet on a benchtop.  This driver has nothing
+//! Meanwell-specific (or indeed even power-specific); this is really just a
+//! wrapper around the seven GPIOs present on J16 on the Gimletlet.  We also
+//! use this opportunity to indicate a different LED pattern (a quick blue LED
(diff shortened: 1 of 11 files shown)


drv/meanwell/src/main.rs
```diff
@@ -0,0 +1,179 @@
...
+    fn current_notification_mask(&self) -> u32 {
+        TIMER_MASK
+    }
+
+    fn handle_notification(&mut self, _bits: u32) {
+        let user_leds =
+            drv_user_leds_api::UserLeds::from(USER_LEDS.get_task_id());
+
+        if !self.led_on {
+            user_leds.led_on(LED_INDEX).unwrap();
```
mkeeter, line 124: Nit: `UserLeds` exposes an `led_toggle` command

mkeeter (approved):

drv/meanwell/src/main.rs
bcantrill: If you don't mind, I think I want to leave this as an explicit on to assure that its state is what the driver thinks it should be. (The LED is on for 100 ms and off for 900 ms.)

PR #622, opened 2022-06-23
Improve VSC7448 control

mkeeter:
This PR does a bunch of things:

- Renames the `vsc7448` task to `monorail-server`
- Adds an Idolatry API and `ServerImpl` to that task, removing the old main loop
- Adds `monorail-api` to interface with that task, with functions to get/set registers on the VSC7448 and PHY, and VSC7448 / PHY status.
- Changes VSC7448 configuration to be declaratively determined by an array of port modes, rather than imperatively in the BSP.
- Makes the `Phy`'s "last page" cache use interior mutability, then change a bunch of functions from `&mut self` to `&self`


bcantrill (approved):
This looks really good -- can't wait to use `humility monorail`!

PR #625, opened 2022-06-24
Rename At24csw080 -> At24Csw080, implement Validate

arjenroodselaar:
The `at24csw080` type breaks some of the I2C auto-generated code since that expects the type to be `At24Csw080`. With this driver not really in use yet renaming seems like the lowest friction path. In addition the `Validate` trait is implemented so this can be used to test presence.

### DIFF
--- a/drv/i2c-devices/src/at24csw080.rs
+++ b/drv/i2c-devices/src/at24csw080.rs
@@ -4,6 +4,7 @@
 
 //! Driver for the AT24CSW080/4 I2C EEPROM
 
+use crate::Validate;
 use core::convert::TryInto;
 use drv_i2c_api::*;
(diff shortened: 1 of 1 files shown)


bcantrill (approved):
Looks great -- ship it!

drv/i2c-devices/src/at24csw080.rs
```diff
@@ -64,6 +65,15 @@ impl From<ResponseCode> for Error {
     }
 }
 
+impl From<Error> for ResponseCode {
+    fn from(err: Error) -> Self {
+        match err {
+            Error::I2cError(code) => code,
+            _ => panic!(),
```
arjenroodselaar, line 72: I wonder if it would make sense to add a catch-all "DeviceSpecific" variant to `ResponseCode`, for safe conversions like this.

PR #626, opened 2022-06-25
QSFP x32 FPGA driver, server multiple device from fpga-server

arjenroodselaar:
This diff implements the following:

- Allow a `fpga-server` task to server requests for multiple FPGA devices
- Add a driver for hardware found on the QSFP x32 board, where multiple ECP5
  devices are controlled using a mux'd SPI link and an I2C GPIO expander

### DIFF
--- a/drv/fpga-devices/Cargo.toml
+++ b/drv/fpga-devices/Cargo.toml
@@ -6,10 +6,12 @@ edition = "2021"
 [dependencies]
 userlib = {path = "../../sys/userlib"}
 ringbuf = {path = "../../lib/ringbuf"}
+drv-i2c-api = {path = "../../drv/i2c-api"}
+drv-i2c-devices = {path = "../../drv/i2c-devices"}
+drv-fpga-api = {path = "../../drv/fpga-api"}
 drv-spi-api = {path = "../../drv/spi-api"}
 drv-stm32xx-sys-api = {path = "../stm32xx-sys-api"}
-drv-fpga-api = {path = "../../drv/fpga-api"}
-num-traits = { version = "0.2.12", default-features = false }
+num-traits = {version = "0.2.12", default-features = false}
 cfg-if = "1"
 bitfield = "0.13"
 zerocopy = "0.6.1"
--- a/drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
+++ b/drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
@@ -1,0 +1,362 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::ecp5::{Command, Ecp5, Ecp5Driver};
+use crate::FpgaUserDesign;
+use drv_fpga_api::FpgaError;
+use drv_i2c_api::ResponseCode;
+use drv_i2c_devices::pca9538;
+use drv_spi_api::{self as spi_api, SpiDevice, SpiError};
+use drv_stm32xx_sys_api::{self as sys_api, GpioError, Sys};
+
+/// This module implements an ECP5 driver which exposes two physical devices,
+/// which share a single SPI bus using a mux and are controlled through a shared
+/// PCA9538 GPIO expander.
+
+/// Impl Error type, with conversion from `GpioError`, `SpiError` and
+/// `ResponseCode`.
+#[derive(Copy, Clone, Debug)]
+pub enum Error {
+    GpioError(GpioError),
+    SpiError(SpiError),
+    I2cError(ResponseCode),
+}
+
+impl From<GpioError> for Error {
+    fn from(e: GpioError) -> Self {
+        Self::GpioError(e)
+    }
+}
+
+impl From<SpiError> for Error {
+    fn from(e: SpiError) -> Self {
+        Self::SpiError(e)
+    }
+}
+
+impl From<ResponseCode> for Error {
+    fn from(e: ResponseCode) -> Self {
+        Self::I2cError(e)
+    }
+}
+
+impl From<Error> for u8 {
+    fn from(e: Error) -> Self {
+        match e {
+            Error::GpioError(e) => match e {
+                GpioError::BadArg => 2,
+            },
+            Error::SpiError(e) => match e {
+                SpiError::BadTransferSize => 3,
+                SpiError::ServerRestarted => 4,
+                SpiError::NothingToRelease => 5,
+                SpiError::BadDevice => 6,
+                SpiError::DataOverrun => 7,
+            },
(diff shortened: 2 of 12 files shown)


drv/fpga-server/src/main.rs
```diff
@@ -117,11 +179,45 @@ impl<Device: FpgaUserDesign> Drop for UserDesignLock<'_, Device> {
...
+                } else {
+                    Err(FpgaError::BadDevice)
+                }
+            }
+            None => {
+                if device_index < self.devices.len() {
+                    Ok(&self.devices[device_index])
+                } else {
+                    Err(FpgaError::BadDevice)
+                }
```
mkeeter, line 207: Nit:
```suggestion
                self.devices.get(device_index).ok_or(FpgaError::BadDevice),
```

drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+pub struct Pins {
+    pub done: pca9538::PinSet,
+    pub init_n: pca9538::PinSet,
+    pub program_n: pca9538::PinSet,
+    pub user_design_reset_n: pca9538::PinSet,
+}
+
+pub struct Driver {
```
mkeeter, line 75: I don't love having every member `pub` here – since there's a cache in `instance_select`, it makes it possible to invalidate state.  However, I _do_ like having the constructor specify variables by name.

How about making internal variables of the `Driver` private, then adding
```
struct DriverConfig {
    pub sys: Sys,
    pub configuration_port: SpiDevice,
    pub user_design: SpiDevice,
    pub spi_mux_select: sys_api::PinSet,
    pub gpio: pca9538::Pca9538,
    pub user_design_reset_duration: u64,
}
```
with an initialization function
```
impl Driver {
    pub fn init(config: DriverConfig) -> Self {
       ...
    }
}
```
The `init` function should also does GPIO configuration to ensure that the `instance_selected` is correct (currently done in `configure_gpio`).  That way, the only way to get a `struct Driver` ensures that it's valid.

drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub instance_selected: core::cell::Cell<usize>,
+    pub gpio: pca9538::Pca9538,
+    pub user_design_reset_duration: u64,
+}
+
+impl Driver {
+    pub fn select_instance(&self, instance: usize) -> Result<(), GpioError> {
+        if self.instance_selected.get() != instance {
+            self.instance_selected.set(instance);
+            self.sys.gpio_set_to(
```
mkeeter, line 89: I like to store `Cell<Option<usize>>` for caches like this one.

That way, if `gpio_set_to` returns an error, we'd store `None` in the cache, representing an unknown state (since we don't know if the GPIO task crashed before or after setting the pin).

drv/fpga-devices/Cargo.toml
```diff
@@ -6,14 +6,19 @@ edition = "2021"
...
 drv-stm32xx-sys-api = {path = "../stm32xx-sys-api"}
-drv-fpga-api = {path = "../../drv/fpga-api"}
-num-traits = { version = "0.2.12", default-features = false }
+num-traits = {version = "0.2.12", default-features = false}
 cfg-if = "1"
 bitfield = "0.13"
 zerocopy = "0.6.1"
 
+[features]
+front_io = ["drv-i2c-api", "drv-i2c-devices"]
```
mkeeter, line 20: This feature doesn't appear to be used; `drv-fpga-devices` builds both `ecp5_spi::Ecp5UsingSpi` and `ecp5_spi_mux_pca9538::Driver` all the time.

I think this is fine, just delete the feature.

drv/fpga-server/Cargo.toml
```diff
@@ -7,19 +7,26 @@ edition = "2021"
...
 num-traits = { version = "0.2.12", default-features = false }
 zerocopy = "0.6.1"
 
 [build-dependencies]
+build-i2c = {path = "../../build/i2c"}
 build-util = {path = "../../build/util"}
 idol = {git = "https://github.com/oxidecomputer/idolatry.git"}
 
+[features]
+mainboard = []
```
mkeeter, line 27: Since these are mutually exclusive, there should be a guard in `drv/fpga-server/src/main.rs`, e.g.
```rust
#[cfg(all(target_board = "sidecar-1", feature = "mainboard", feature = "front_io"))]
compile_error!("Cannot enable both mainboard and front_io simultaneously");
```

mkeeter (commented):
A few notes, but this is looking pretty good!  You'll probably want to run `rustfmt` and `cargo xtask clippy` before merging.

drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
arjenroodselaar: Yup, this split is a good idea. It's a common pattern I use in Bluespec code, not sure why I didn't think of it here.

drv/fpga-server/src/main.rs
arjenroodselaar: Thanks! I managed to make this function even shorter.

drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
arjenroodselaar: I've refactored this quite a bit and overall I am pretty happy with the result. The only thing that's nagging a bit is the `config` member in `Driver` causing deeper nesting elsewhere in the driver. But the immediate alternative is copying the fields from `DriverConfig` to `Driver`, which seemed equally meh. I was contemplating something with `Deref`, but that didn't look that great either. Any suggestions on something I have not thought of?

drv/fpga-devices/Cargo.toml
arjenroodselaar: Ah yes, it needs a `feature` annotation in `lib.rs` to be active. But yes, it's fine to leave this out.

drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
arjenroodselaar: I updated the function with the suggestion, but let me know if there is a more concise way to do this.

drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
mkeeter: Yeah, it seems like all the options are meh:
- Storing the `DriverConfig` in the driver adds an extra `.config` all over
- Copying fields is a lot of duplication
- Using `Deref` for anything other than a smart pointer is heavily frowned upon

I think options 1 vs 2 are about the same, so no objections to the one you picked.

PR #628, opened 2022-06-27
Minimal Front IO board initialization, PHY SMI API

arjenroodselaar:
This diff:

- Breaks up the Sidecar sequencer into separate modules in an attempt to improve
  readability
- Adds basic support to initialize a connected Front IO board
- Exposes a PHY SMI API, to allow initializing the VSC8562 found on a QSFP x32
  board


drv/monorail-api/src/lib.rs
```diff
@@ -39,6 +39,7 @@ pub struct PortCounters {
 pub enum MonorailError {
     SpiError,
     ServerDied,
+    ProxyError,
```
arjenroodselaar, line 42: Looking for input on what this should be. `ProxyError` was the best I could think of.

drv/fpga-api/src/lib.rs
```diff
@@ -267,6 +267,28 @@ impl FpgaUserDesign {
...
 
+/// Poll the device state of the FPGA to determine if it is ready to receive
+/// a bitstream, resetting the device if needed.
+pub fn await_fpga_ready(
+    fpga: &mut Fpga,
+    sleep_ticks: u64,
+) -> Result<DeviceState, FpgaError> {
+    let mut state = fpga.state()?;
+
+    while match state {
```
mkeeter, line 278: Nit: this could be
```
while match!(state, DeviceState::Unknown | DeviceState::Disabled | DeviceState::Error) {
```
(using the `match!` macro and removing one level of inversion, since I found it confusing at first glance)

mkeeter: Alternatively, remove the separate `mut state` variable entirely and do everything in the loop, e.g.
```rust
loop {
    let state = fpga.state()?;
    match state {
        DeviceState::AwaitingBitstream | DeviceState::RunningUserDesign => return Ok(state),
        _ => (),
    }
    fpga.reset()?;
    userlib::hl::sleep_for(sleep_ticks);
}
```
This hides the loop condition in the `match` statement, but makes everything else a little cleaner, so I don't feel too strongly about it...

drv/sidecar-front-io-api/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+num-traits = {version = "0.2.12", default-features = false}
+zerocopy = "0.6.1"
+
+[features]
+controller = []
+phy_smi = ["vsc85xx", "vsc7448-pac"]
+
+# a target for `cargo xtask check`
+[package.metadata.build]
+target = "thumbv7em-none-eabihf"
```
mkeeter, line 21: `cargo xtask check` doesn't exist, so I think we can delete these lines.

drv/sidecar-front-io-api/build.rs
```diff
@@ -0,0 +1,166 @@
...
+use std::{env, fs, path::PathBuf};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    build_util::expose_target_board();
+
+    let out_dir = &PathBuf::from(env::var_os("OUT_DIR").unwrap());
+
+    fs::write(out_dir.join("sidecar_qsfp_x32_controller.rs"), regs()?)?;
+
+    let ecp5_bitstream_name = match env::var("HUBRIS_BOARD")?.as_str() {
```
mkeeter, line 16: Why are we deploying the FPGA image in `sidecar-front-io-api`, rather than the `-server` task?

e.g. on Gimlet, the image is in `gimlet-seq-server`, rather than `gimlet-seq-api`.

app/sidecar/app.toml
```diff
@@ -214,10 +214,14 @@ task-slots = ["sys", "i2c_driver", {spi_driver = "spi1_driver"}]
 name = "drv-sidecar-seq-server"
 features = []
 priority = 3
-max-sizes = {flash = 131072, ram = 2048}
+max-sizes = {flash = 262144, ram = 2048}
```
mkeeter, line 217: No need to change anything for this PR, but I wonder if this is now big enough to make `miniz` worthwhile 👀...

(It wasn't worth it for Gimlet, because it saved 10K of FPGA image size and added 11K of decompressor code)

drv/sidecar-front-io-api/build.rs
```diff
@@ -0,0 +1,166 @@
...
+    // Make sure the app image is rebuilt if the bitstream file for this target
+    // changes.
+    println!("cargo:rerun-if-changed={}", ecp5_bitstream_name);
+
+    Ok(())
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(tag = "type", rename_all = "lowercase")]
+enum Node {
```
mkeeter, line 37: This code is now duplicated 3x (`gimlet-seq-server/build.rs`, `sidecar-mainboard-controller-api`, and `sidecar-front-io-api`).  It's probably time to make a separate crate (`build/fpga-regmap`?) and call into it!

drv/sidecar-front-io-api/build.rs
mkeeter: After reading more, I think the crate is just confusingly named.

It doesn't feel like a `-api` crate, which is normally a thin shim around Idolatry client bindings.

How about renaming to `drv-sidecar-front-io` and `drv-sidecar-mainboard-controller`?  Then, it doesn't feel weird that it contains the FPGA image.

drv/sidecar-front-io-api/build.rs
```diff
@@ -0,0 +1,166 @@
...
+            panic!("unexpected child {:?}", child);
+        }
+    }
+
+    writeln!(&mut output, "}}")?;
+
+    Ok(output)
+}
+
+fn compress(input: &[u8]) -> Vec<u8> {
```
mkeeter, line 156: I opened #630 to remove this duplicate code, but no need to handle it here.

drv/sidecar-front-io-api/src/controller.rs
```diff
@@ -0,0 +1,56 @@
...
+
+static COMPRESSED_BITSTREAM: &[u8] =
+    include_bytes!(concat!(env!("OUT_DIR"), "/ecp5.bin.rle"));
+
+pub struct FrontIOController {
+    fpga: Fpga,
+    user_design: FpgaUserDesign,
+}
+
+impl FrontIOController {
```
mkeeter, line 16: Is it worth combining `FrontIOController` and `MainboardController`, which share basically the same logic?

drv/sidecar-front-io-api/src/phy_smi.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+impl PhyRw for PhySmi {
+    #[inline]
+    fn read_raw<T: From<u16>>(
+        &self,
+        phy: u8,
+        reg: PhyRegisterAddress<T>,
+    ) -> Result<T, VscError> {
+        self.read_raw(phy, reg)
```
mkeeter, line 139: It's not obvious that this calls `PhySmi::read_raw`, instead of recursion forever – maybe rename the inner functions to `read/write_raw_inner`?

drv/fpga-api/src/lib.rs
arjenroodselaar: I keep forgetting there's a plain `loop {..}`. The `while` is a bit awkward and this preserves the linear nature of what I am trying to accomplish.

drv/sidecar-front-io-api/build.rs
arjenroodselaar: Sure, we can do that. In a different world I would have called these something along the lines of `libfront_io` and `libmainboard_controller`. I can see the argument why `-api` crates are closely tied to Idol client bindings.

app/sidecar/app.toml
arjenroodselaar: Maybe? I think we soon need to do the work to enable the additional flash we have on this board to store these images. That'll free up this as well as other init blobs.

drv/sidecar-front-io-api/build.rs
arjenroodselaar: Fair, I'll do that.

drv/sidecar-front-io-api/src/controller.rs
arjenroodselaar: I want to keep them separate structs, because they convey a different context. But I agree that duplicate code is meh, so I've started moving most into `fpga-api` where it can be used by anything.

drv/fpga-api/src/lib.rs
arjenroodselaar: Thanks, this suggestion improves this function significantly. And I fixed the edge case where the device may be held in reset, which would otherwise cause this function to hang indefinite.

PR #634, opened 2022-06-30
Sidecar support for VSC7448 -> QSFP board

mkeeter:
This PR adds support for the front IO (QSFP) board to the `monorail` management task.

### DIFF
--- a/drv/vsc85xx/src/vsc8562.rs
+++ b/drv/vsc85xx/src/vsc8562.rs
@@ -95,14 +95,6 @@ impl<'a, 'b, P: PhyRw> Vsc8562Phy<'a, 'b, P> {
             self.sd6g_patch(true)?;
         }
 
-        /*
-        self.phy.broadcast(|v| {
-            v.modify(phy::EXTENDED_3::MAC_SERDES_PCS_CONTROL(), |r| {
-                r.0 |= 1 << 7; // Enable Clause 37 MAC Autonegotiation
-            })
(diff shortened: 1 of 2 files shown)


drv/vsc85xx/src/vsc8562.rs
```diff
@@ -97,10 +129,33 @@ impl<'a, 'b, P: PhyRw> Vsc8562Phy<'a, 'b, P> {
...
             v.write(phy::TR::TR_16(), 0x8fa4.into())
-        })?;
+        })
+    }
 
-        // In the SDK, there's more configuration for 100BT, which we don't use
+    // Based on `vtss_phy_sd1g_patch_private` in the SDK
+    fn sd1g_patch(&mut self, is_100fx: bool) -> Result<(), VscError> {
+        for p in 0..2 {
+            // XXX The SDK just does self.port * 2 (including any offset); based on
```
bcantrill, line 138: Feels like the `XXX` isn't merited as it seems to actually be correct? ;)

bcantrill (approved):

drv/vsc85xx/src/vsc8562.rs
mkeeter: Maybe! It's hard to tell if it's actually correct, because it's not clear what this actually _does_ – I'm going to leave the comment as a breadcrumb in case we ever get suspicious of the SerDes.

PR #637, opened 2022-07-05
gimlet: need to enable/disable the T6 in response to power enable

### LINKED ISSUE #605: gimlet: need to enable/disable the T6 in response to power enable
The SP is responsible for dealing with kicking off part of the T6 power state machine. Effectively we have the signal `SP3_TO_SP_NIC_PWREN_L` (PF4) which comes out of the SP3 GPIO expander network. This signal basically indicates the intended power state of the device. When the T6 is supposed to be on this signal will be asserted and expect to remain asserted. When it goes away, it will be deasserted. When we enter A0 initially this will remain deasserted. When this is asserted and deasserted we will need to work with the Sequencer state machine to cause it to set and clear the various reset flags and/or activate the broader T6 state machine. @nathanaelhuffman can hopefully provide more on the exact sequencer SPI transactions.

bcantrill:
Fixes #605 and #638.


drv/gimlet-seq-server/src/main.rs
```diff
@@ -316,21 +323,94 @@ struct ServerImpl {
...
+            }
+            _ => {}
+        }
+
+        match self.state {
+            PowerState::A0 => {
+                let sys = sys_api::Sys::from(SYS.get_task_id());
+                let pwren_l = sys.gpio_read_input(NIC_PWREN_L_PORT).unwrap()
+                    & NIC_PWREN_L_MASK
+                    != 0;
```
mkeeter, line 367:
```suggestion
                let pwren_l = sys.gpio_read(NIC_PWREN_L_PINS).unwrap() != 0;
```

mkeeter (approved):
Looks good, two small suggestions

PR #650, opened 2022-07-11
Add FPGA checksum, remove Gimlet A

mkeeter:
- Use new FPGA checksum strategy to decide when to reflash on Gimlet
- Switch to generating register map with `fpga-regmap` crate, removing duplicated code
- Remove rev A FPGA image, since it no longer matches the register map
- Remove Gimlet rev A build entirely, since it no longer has an FPGA image
- Remove `all_ps_enabled.bin`, since it's no longer used


nathanaelhuffman (approved):
lgtm

PR #656, opened 2022-07-15
Multi image

labbott:
Add support for building the same image but at a different address. As part of this change, the stage0 bootloader on the LPC55 targets now needs to be built separately.

### DIFF
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -313,7 +313,6 @@ impl Config {
         &self,
         region: String,
     ) -> Result<IndexMap<String, Range<u32>>> {
-        // self.image_names.iter().fold(|a| self.memories(a)? ).collect()
         let mut memories: IndexMap<String, Range<u32>> = IndexMap::new();
         for a in &self.external_images {
             if let Some(r) = self.memories(&a)?.get(&region) {
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -82,14 +82,13 @@ impl Config {
             toml::from_slice(&chip_contents)?
         };
 
-        let outputs : IndexMap<String, Vec<Output>> = {
+        let outputs: IndexMap<String, Vec<Output>> = {
             let fname = if let Some(n) = toml.memory {
                 n
             } else {
                 "memory.toml".to_string()
             };
-            let chip_file =
-                cfg.parent().unwrap().join(&toml.chip).join(fname);
+            let chip_file = cfg.parent().unwrap().join(&toml.chip).join(fname);
             let chip_contents = std::fs::read(chip_file)?;
             hasher.write(&chip_contents);
             toml::from_slice::<IndexMap<String, Vec<Output>>>(&chip_contents)?
@@ -312,9 +311,12 @@ impl Config {
             .collect()
     }
 
-    pub fn image_memories(&self, region : String) -> Result<IndexMap<String, Range<u32>>> {
-       // self.image_names.iter().fold(|a| self.memories(a)? ).collect()
-        let mut memories : IndexMap<String, Range<u32>> = IndexMap::new();
+    pub fn image_memories(
+        &self,
+        region: String,
+    ) -> Result<IndexMap<String, Range<u32>>> {
+        // self.image_names.iter().fold(|a| self.memories(a)? ).collect()
+        let mut memories: IndexMap<String, Range<u32>> = IndexMap::new();
         for a in &self.external_images {
             if let Some(r) = self.memories(&a)?.get(&region) {
                 memories.insert(a.clone(), r.start..r.end);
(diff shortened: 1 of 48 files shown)


build/xtask/src/config.rs
```diff
@@ -269,35 +277,64 @@ impl Config {
...
     pub fn output_region(&self, vaddr: u64) -> Option<&str> {
         let vaddr = u32::try_from(vaddr).ok()?;
         self.outputs
             .iter()
             .find(|(_name, out)| {
-                vaddr >= out.address && vaddr < out.address + out.size
+                let mut stat = false;
+                for o in out.iter() {
+                    stat = stat
+                        || vaddr >= o.address && vaddr < o.address + o.size;
```
mkeeter, line 335: Nit: parenthesis to clarify operator precedence (which will also cause `rustfmt` to not make this look like a closure!)

mkeeter: Actually, you could replace the whole loop with something like
```
let stat = out.iter().any(|o| vaddr >= o.address && vaddr < o.address + o.size);
```

build/xtask/src/dist.rs
```diff
@@ -122,6 +120,14 @@ impl PackageConfig {
         })
     }
 
+    fn img_dir(&self, img_name: &str) -> PathBuf {
+        self.dist_dir.join(img_name)
+    }
+
+    fn img_file(&self, name: impl AsRef<Path>, img_name: &str) -> PathBuf {
+        self.img_dir(img_name).join(name)
```
mkeeter, line 128: Is this going to break Humility?

build/xtask/src/dist.rs
```diff
@@ -816,16 +750,112 @@ fn build_kernel(
...
     let (kentry, _) = load_elf(
-        &cfg.dist_file("kernel"),
+        &cfg.img_file("kernel", image_name),
         all_output_sections,
         &mut ksymbol_table,
     )?;
     Ok((kentry, ksymbol_table))
 }
 
+fn update_elf(
```
mkeeter, line 783: I'd still like a docstring explaining what this function does!

mkeeter (approved):
Small notes, but looks good!

build/xtask/src/humility.rs
```diff
@@ -24,6 +25,7 @@ pub fn run(
     let archive = Path::new("target")
         .join(&toml.name)
         .join("dist")
+        .join(image_name)
```
mkeeter, line 28: It would be good to check the `image_name` against the `app.toml` before calling Humility, since the failure mode would be confusing (Humility would complain about a missing archive, I think?)

build/xtask/src/config.rs
labbott: I was hoping someone would turn this into a one liner on review (because I couldn't :))

build/xtask/src/dist.rs
labbott: It will not! Right now we generate a separate build.zip for each image so from humilty's perspective it's just loading another archive. Cliff and I briefly chatted about having a single build.zip but that's a longer term TODO and would require a humility bump.

build/xtask/src/config.rs
jgallagher: You could also replace `.find(..).map(..)` with `.find_map(..)`; I think clippy might suggest this? Untested, maybe typos:

```rust
self.outputs.iter().find_map(|(name, out)| {
    if vaddr >= o.address && vaddr < o.address + o.size {
        Some(name.as_str())
    } else {
        None
    }
})
```

A couple golfing ideas from this that might not be good:

1. If `Output` had a method to return the range `address..address+size`, the condition could become `o.address_range().contains(&vaddr)`
2. If going with the `.find_map()` version, the body could use `bool::then_some()` instead of an explicit if/else

build/xtask/src/dist.rs
```diff
@@ -1346,7 +1216,10 @@ pub struct Allocations {
 pub fn allocate_all(
     toml: &Config,
     task_sizes: &HashMap<&str, IndexMap<&str, u64>>,
-) -> Result<(Allocations, IndexMap<String, Range<u32>>)> {
+) -> Result<(
+    BTreeMap<String, Allocations>,
```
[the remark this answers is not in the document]
labbott: Hmmm I see your point. That could work as well and might eliminate a few `get` calls. The one quirk is figuring out what makes sense to return from `package` to run `sizes` with `dist`. The current version returns Allocations from all images and runs sizes on each but that might be overkill.

build/xtask/src/humility.rs
labbott: Yes, good idea.

build/xtask/src/dist.rs
mkeeter: 👍🏻

PR #659, opened 2022-07-26
Use globs in Cargo.toml

steveklabnik:
I saw this on twitter a day ago, and figured I'd give it a try. This ends up including everything, which is nicer than listing each out
individually. However, I can also see the benefits of the "explicit allowlist" approach, so I wouldn't be sad if this isn't accepted.

I also deleted a seemingly useless driver, and added a Cargo.toml for
one of the test packages. Are those changes correct? I'm not sure!


Cargo.lock
```diff
@@ -3466,6 +3478,21 @@ dependencies = [
...
+dependencies = [
+ "build-util",
+ "cfg-if",
+ "cortex-m",
+ "cortex-m-rt",
+ "kern",
+ "panic-halt",
+ "panic-itm",
+ "panic-semihosting",
+ "stm32h7 0.13.0",
```
jgallagher, line 3493: Should this be updated to `0.14.0`? It looks like it's the only crate pulling in `0.13.0`.

steveklabnik: possibly! I tried to change as little as possible, but that does seem like a good idea.

cbiffle (approved):
Well that seems obviously better.

Cargo.lock
cbiffle: (yes, fwiw.)

PR #660, opened 2022-07-26
Configure mux before locking the SPI device

mkeeter:
Calling `SpiDevice::lock` asserts the chip select line.  Because the chip select line is passed through the SPI mux, we need to configure the mux beforehand; otherwise, the CS line will be pulsed for the _other_ FPGA when changing targets.

This fixes [hardware-qsfp-x32#36](https://github.com/oxidecomputer/hardware-qsfp-x32/issues/36) and [humility#191](https://github.com/oxidecomputer/humility/issues/191), tested on the `niles` Sidecar.


arjenroodselaar (approved):
Thanks for looking into this.

PR #670, opened 2022-07-28
Add `net` PHY read/write API

mkeeter:
This is initial infrastructure for a `humility net` subcommand, to query the SP's perspective on the management network.

These functions are distinct from the `smi_read/write` commands, since we need to
- change the page
- do a read/write
without anything else interrupting us.

In passing, I refactored the "VLAN Server" vs "Basic Server" distinction, moving them into better-named files, then added `server.rs` for a common trait.  This removes duplicate code from the two Idol implementations.  Unfortunately, moving `server.rs -> server_basic.rs`, then re-adding `server.rs` makes the diff unhappy; sorry about that!

In addition, the new API uses register types from the `vsc7448_pac` crates, so that crate is now non-optional.  This means I finally removed the `enum SmiClause22Register`, which duplicated some of that code (but was previously saved us an extra import).  The new dependency is only new for the STM32H7-Nucleo task; everyone else was already using it.

Finally, I noticed that the STM32H7-Nucleo `net` BSP was wrong: the PHY is at address 0, rather than 1.  I fixed this, then removed the loop-until-link-up block, because none of the other boards have that kind of blocking behavior in their BSP constructor.  Previously, the loop-until-link-up was accidentally returning right away, because the PHY address was wrong and MDIO returned `0xFFFF`.

### DIFF
--- a/task/udpecho/src/main.rs
+++ b/task/udpecho/src/main.rs
@@ -40,8 +40,10 @@ fn main() -> ! {
                         Err(NetError::NotYours) => panic!(),
                         Err(NetError::InvalidVLan) => panic!(),
                         Err(NetError::Other) => panic!(),
-                        // `send_packet()` can't return QueueEmpty
+                        // `send_packet()` can't return QueueEmpty or
+                        // InvalidPort
                         Err(NetError::QueueEmpty) => unreachable!(),
+                        Err(NetError::InvalidPort) => unreachable!(),
                     }
                 }
             }
@@ -52,8 +54,9 @@ fn main() -> ! {
             Err(NetError::NotYours) => panic!(),
             Err(NetError::InvalidVLan) => panic!(),
             Err(NetError::Other) => panic!(),
-            // `recv_packet()` can't return QueueFull
+            // `recv_packet()` can't return QueueFull or InvalidPort
             Err(NetError::QueueFull) => unreachable!(),
+            Err(NetError::InvalidPort) => unreachable!(),
         }
 
         // Try again.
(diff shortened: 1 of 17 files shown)


task/udpecho/src/main.rs
```diff
@@ -40,8 +40,10 @@ fn main() -> ! {
                         Err(NetError::NotYours) => panic!(),
                         Err(NetError::InvalidVLan) => panic!(),
                         Err(NetError::Other) => panic!(),
-                        // `send_packet()` can't return QueueEmpty
+                        // `send_packet()` can't return these errors
```
cbiffle, line 43: As these diverge more, it becomes more likely that we should separate these error types. There's no particular reason why every operation on a server needs to return the same error type, and in these cases it seems like we're approaching having disjoint sets of "MAC related errors" vs "PHY related errors."

cbiffle (approved):
This looks pretty good.

My favorite thing about this is, with the trait seam you've introduced, we're one step closer to making `net` itself not depend on stm32h7, which will be valuable.

I am a little :-\ about relying on the PHY types from the vsc crates even on boards that don't have VSC-series chips on them. FWIW, the original APIs had a separate Clause22 register enum because that was intended to define the _standard_ registers from clause22 -- and the PHY operations took any type that could be converted into a u8 so that you could add additional types alongside that for vendor-specific register sets. I believe this change puts us into a state where (1) everybody's building those PACs, which are not small, whether they need 'em or not, and (2) we have no compiler help determining which registers exist in the system.

Since our production boards currently use VSC-series chips, I don't feel like this is a serious problem to block your refactor, but I wanted to highlight it as a concern.

mkeeter:
Yeah, I don't _love_ the PAC PHY registers, because the source files are incomplete / different PHYs have different subsets of registers and bits.  It would be better to hand-craft them on a per-PHY basis, like we do with `pmbus`, but that would take extra time.  I'm going to open an issue to revisit this.

Still, I think it makes sense to merge this now to unblock @jgallagher 's error refactoring work, so I'll queue it up for automerge.

PR #672, opened 2022-08-01
LPC55 flash write / update API

labbott:


build/xtask/src/dist.rs
```diff
@@ -836,12 +949,80 @@ fn update_image_header(
...
+                    // This means the entry for our flash gets broken
+                    // down into three entries:
+                    // 1) Non-secure range before the secure task
+                    // 2) non-secure range after the secure task
+                    // 3) NSC region in the secure task
+                    for (_, range) in map.iter() {
+                        if range.contains(&s.secure.start) {
+                            header.sau_entries[i].rbar = range.start;
+                            header.sau_entries[i].rlar =
+                                (s.secure.start - 1) & !0x1f | 1;
```
kc8apf, line 994: comment or function to make it clearer what this line is doing.  I'm pretty confident this is setting 1 to mark the entry valid but it'd help to have it spelled out here and elsewhere.

drv/lpc55-update-server/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    fn write_one_block(
+        &mut self,
+        _: &RecvMessage,
+        block_num: usize,
+        block: LenLimit<Leased<R, [u8]>, MAX_LEASE>,
+    ) -> Result<(), RequestError<UpdateError>> {
+        match self.state {
+            UpdateState::NoUpdate | UpdateState::Finished => {
+                return Err(UpdateError::UpdateInProgress.into())
```
kc8apf, line 68: Isn't the error that an update _isn't_ in progress?

drv/lpc55-update-server/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    fn finish_image_update(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<(), RequestError<UpdateError>> {
+        match self.state {
+            UpdateState::NoUpdate | UpdateState::Finished => {
+                return Err(UpdateError::UpdateInProgress.into())
```
kc8apf, line 122: Again, isn't the error that an update isn't in progress?

lib/lpc55-flash/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn __write_block(
+    image_num: ImageType,
+    page_num: u32,
+    buffer: *mut u32,
+) -> HypoStatus {
+    // Can only update opposite image
+    if same_image(image_num) {
+        return HypoStatus::SameImage;
```
kc8apf, line 115: maybe RunningImage?  I read "Same" as equal content, not the same slot.

task/real/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "task-real"
```
kc8apf, line 2: Is this a test of some kind?

labbott: Yes, this was an old test. I'll remove it.

lib/lpc55-flash/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+    if let Err(result) = flash_erase(write_addr, FLASH_PAGE_SIZE as u32) {
+        return HypoStatus::FlashError(result);
+    }
+
+    if let Err(result) = flash_write(write_addr, buffer, FLASH_PAGE_SIZE as u32)
+    {
+        return HypoStatus::FlashError(result);
+    }
+
+    return HypoStatus::Success;
```
cbiffle, line 155: FWIW, this function contains a lot of noise because it isn't returning `Result`, so you can't use `?` and its automatic conversions. Consider factoring the body of the function into a routine using `Result`. You can then do things like chained checked arithmetic as

```rust
let x = y.checked_mul(z).and_then(|product| product.checked_add(base)).ok_or(Whatever)?;
```

You could then call that routine from this `extern "C"` routine and convert the result.

task/hiffy/src/common.rs
```diff
@@ -4,7 +4,12 @@
 
 use hif::{Failure, Fault};
 use hubris_num_tasks::NUM_TASKS;
-#[cfg(any(feature = "qspi", feature = "hash", feature = "rng"))]
+#[cfg(any(
```
cbiffle, line 7: This is getting kind of elaborate, how about

```rust
#[allow(unused_imports)]
```

build/task-link.x
```diff
@@ -15,6 +15,16 @@ SECTIONS
     __etext = .;
   } > FLASH =0xdededede
 
+   /* 32-byte alignment requirement for SAU */
```
cbiffle, line 18: What effect does this have on non-TZ binaries? It seems like it'll inject some padding.

task/secure/README.mkdn
```diff
@@ -0,0 +1,5 @@
+# Task Template
```
cbiffle, line 1: Remove or fill out please.

task/secure/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use hypocalls::*;
+#[allow(unused_imports)]
+use userlib::*;
+
+#[link_section = ".tz_table"]
+#[no_mangle]
+#[used]
+static TZ_TABLE: SecureTable = SecureTable {
+    magic: 0xaaaa_bbbb,
```
cbiffle, line 16: Is this number important?

task/secure/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+    write_to_flash: write_to_flash,
+};
+
+#[export_name = "main"]
+fn main() -> ! {
+    loop {
+        cortex_m::asm::nop();
```
cbiffle, line 23: What is going on here? This will burn all cycles at its priority level and starve anything below.

cbiffle (commented):
Design level thoughts:

- I'm concerned about punning "uses" to mean both peripherals and TZ interactions. Let's add a task attribute instead, maybe something like `uses-trustzone = true`
- I don't understand the secure task, why it appears to burn all CPU cycles, and what its role is in the system. This at least needs docs.
- I would rather not hardcode expectations about task naming into the build system, let's designate the secure task with a top-level key (or configuration key) instead.

drv/lpc55-update-server/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+            }
+            _ => (),
+        }
+
+        let len = block.len();
+
+        // The max lease length is longer than our block size, double
```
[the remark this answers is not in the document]
labbott: The problem is we're using the same IDL interface for both the SP and RoT and the two have different maximums. It seemed silly to reimplement the same API and I couldn't find a way to change the block size at build time.

task/secure/README.mkdn
labbott: This is where I should put all the documentation about our TZ interface :)

task/secure/src/main.rs
labbott: Not particularly. I added the magic number mostly to be an extra check to make sure the table was consistent in the task using the table. Clearly I didn't end up implementing that. I'll see about actually adding that check.

task/secure/src/main.rs
labbott: Added details in README (tl;dr this isn't really a "task")

task/secure/README.mkdn
```diff
@@ -0,0 +1,22 @@
...
+
+Hubris can run on ARMv8m targets, some of which may support
+[TrustZone](https://developer.arm.com/documentation/100688/0200/Secure-and-Normal-Worlds?lang=en).
+ARM nominially expects the secure world to be compiled as a
+[static library](https://developer.arm.com/documentation/ecm0359818/latest/)
+but making that work with Hubris has turned out to be troublesome (static
+libraries are big, too many build extensions required among other reasons).
+Instead of building a static library, Hubris designates a single "task" as
+the secure task where all memory will be marked as secure (or NSC as
+appropriate). "Task" is being used very generiously as this is not designed to
```
mkeeter, line 11:
```suggestion
appropriate). "Task" is being used very generously as this is not designed to
```

mkeeter (approved):

PR #673, opened 2022-08-01
Add `large_payload_behavior` arg to `net::recv_packet()`

jgallagher:
This fixes the ability of an external UDP source crashing `net` clients with `ClientError::BadLease` by sending packets in between their lease size and the max size allowed by the `net` task.

Unfortunately it continues to pour fuel on the fire of `server.rs` + `vlan.rs`.


cbiffle (approved):
Looks right to me.

One reason to do this as two separate operations (...in the future), now that I see the code, is so that they could return more precise error types, since the Discard variant _can't_ produce TooLarge, yet right now clients are forced to handle it by the compiler.

PR #675, opened 2022-08-02
Split `NetError` into `{Phy,Recv,Send}Error`

### LINKED ISSUE #674: Split `NetError` into `{Send,Recv}Error`
As of #673, `NetError` contains 1 variant that applies only to send (`QueueFull`) and 2 variants that apply to recv (`QueueEmpty`, `PayloadTooLarge`), with the remaining 3 variants being common. We should split this up to clean up callers' error handling.

Wait to do this until #670 lands, which also touches `NetError`.

jgallagher:
Fixes #674


mkeeter (approved):
Very nice!

PR #676, opened 2022-08-02
xtask: add --dirty flag to suppress cleaning

cbiffle:


mkeeter:
One potential point of confusion: if I do a `--dirty` run, then I think future non-dirty runs _won't clean_ beforehand.

In other words, this could change behavior of later builds in a confusing way: if we expect `cargo xtask dist` to clean when switching between images, this lets us switching _without_ cleaning.

```
build sidecar
build gimlet --dirty
build gimlet # no cleaning here
```

```
build sidecar
build gimlet # cleaning here
```

Is this a legit concern?

cbiffle:
I don't think the particular sequence you highlighted is a problem, because `--dirty` also suppresses update of the build stamp file. So the sequence gimlet / gimlet `--dirty` / sidecar will clean at the sidecar build.

Now, sidecar / gimlet `--dirty` / sidecar might produce weird effects.

This flag is intended, in part, to help us learn what those weird effects are.

mkeeter (approved):
Makes sense, let's bring it in

PR #677, opened 2022-08-02
Sidecar flash image size improvements

cbiffle:
These changes, all together, free up
- 133,248 bytes of Flash
- 1,280 bytes of RAM

The bulk of the savings is from the monomorphization-control pattern I applied to the vsc crates, a change which should be stable across compiler revs and reduce the rate of code growth as more logic is added.

### DIFF
--- a/task/monorail-server/src/bsp/sidecar_1.rs
+++ b/task/monorail-server/src/bsp/sidecar_1.rs
@@ -135,32 +135,12 @@ pub use map::PORT_MAP;
 /// which owns the ethernet peripheral containing the MDIO block.
 pub struct NetPhyRw<'a>(&'a mut task_net_api::Net);
 impl<'a> PhyRw for NetPhyRw<'a> {
-    #[inline(always)]
-    fn read_raw<T: From<u16>>(
-        &self,
-        port: u8,
-        reg: PhyRegisterAddress<T>,
-    ) -> Result<T, VscError> {
-        self.0
-            .smi_read(port, reg.addr)
-            .map(|r| r.into())
-            .map_err(|e| e.into())
+    fn read_raw(&self, port: u8, reg: u8) -> Result<u16, VscError> {
+        Ok(self.0.smi_read(port, reg)?)
     }
 
-    #[inline(always)]
-    fn write_raw<T>(
-        &self,
-        port: u8,
-        reg: PhyRegisterAddress<T>,
-        value: T,
-    ) -> Result<(), VscError>
-    where
-        u16: From<T>,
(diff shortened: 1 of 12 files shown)


task/monorail-server/src/bsp/sidecar_1.rs
```diff
@@ -135,32 +135,12 @@ pub use map::PORT_MAP;
...
-    fn read_raw<T: From<u16>>(
-        &self,
-        port: u8,
-        reg: PhyRegisterAddress<T>,
-    ) -> Result<T, VscError> {
-        self.0
-            .smi_read(port, reg.addr)
-            .map(|r| r.into())
-            .map_err(|e| e.into())
+    fn read_raw(&self, port: u8, reg: u8) -> Result<u16, VscError> {
```
mkeeter, line 138: Should this be annotated with `#[inline(never)]`, per the comment on the trait?

task/monorail-server/src/bsp/sidecar_1.rs
```diff
@@ -511,27 +491,14 @@ pub enum GenericPhyRw<'a> {
 
 impl<'a> PhyRw for GenericPhyRw<'a> {
     #[inline(always)]
-    fn read_raw<T: From<u16>>(
-        &self,
-        port: u8,
-        reg: PhyRegisterAddress<T>,
-    ) -> Result<T, VscError> {
+    fn read_raw(&self, port: u8, reg: u8) -> Result<u16, VscError> {
```
mkeeter, line 494: Same question here!

mkeeter:
Generally looks good!  My only question is that there's a comment on `trait PhyRw` saying

```
/// This means that an impl of this trait should probably apply
/// `#[inline(never)]` to each function in the impl. We can't do that at the
/// trait level; it has to be on the impl.
```

but then all of the implementers of the trait don't do that (and some of them are even `#[inline(always)]`!).

 Is this correct / intentional?

task/monorail-server/src/bsp/sidecar_1.rs
cbiffle: So, in this case, the answer is no. This is just passing off to `smi_read` and should ideally be inlined.

cbiffle:
The ones that aren't `inline(never)` are passing off to other functions instead of embedding significant logic, and didn't come up as significant users of flash space, so I did not annotate them. Is there a rewording of that comment that would cover this?

cbiffle:
Matt and I discussed this in person and I'm going to go ahead and merge.

PR #679, opened 2022-08-02
sidecar: hack to allow larger bitstreams for now

cbiffle:
This allows sidecar to use both banks of the STM32H7 flash for its
images, which in turn allows the use of FPGA bitstreams that are larger
than ~100kiB. I've tested this with up to 512 kiB of bitstreams and
there's space to spare.

Note that this destroys sidecar's ability to do online updates or have
redundant firmware images. So, this is explicitly intended as a
short-term hack to unblock bringup while we get the external flash
online.

### DIFF
--- a/chips/stm32h7-nonredundant/memory.toml
+++ b/chips/stm32h7-nonredundant/memory.toml
@@ -1,0 +1,24 @@
+# Flash sections are mapped into both flash banks, YOLO
+[[flash]]
+address = 0x08000000
+size = 2097152
+read = true
+execute = true
+
+# RAM sections are currently mapped into DTCM, a small but fast SRAM.
+[[ram]]
+address = 0x20000000
+size = 131072
+read = true
+write = true
+execute = false  # let's assume XN until proven otherwise
+
+# Network buffers are placed in sram1, which is directly accessible by the
+# Ethernet MAC.
+[[sram1]]
+address = 0x30000000
+size = 0x20000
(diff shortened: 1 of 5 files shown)


chips/stm32h7-nonredundant/memory.toml
```diff
@@ -0,0 +1,24 @@
+# Flash sections are mapped into flash bank 1 (of 2).
```
mkeeter, line 1: This comment doesn't seem right 😄

chips/stm32h7-nonredundant/openocd.cfg
```diff
@@ -0,0 +1,43 @@
+gdb_port 3343
+telnet_port 4453
+
+source [find interface/stlink.cfg]
+source [find target/stm32h7x_dual_bank.cfg]
```
mkeeter, line 5: Should this be the normal `stm32h7x.cfg`?

chips/stm32h7-nonredundant/memory.toml
cbiffle: Altered

chips/stm32h7-nonredundant/openocd.cfg
cbiffle: No, that config is for any part with the dual-bank flash controller.

mkeeter (approved):

PR #684, opened 2022-08-04
Add system reset kipc and associated jefe interface

labbott:


idl/jefe.idol
```diff
@@ -14,5 +14,9 @@ Interface(
             reply: Simple("()"),
             idempotent: true,
         ),
+	"request_reset": (
```
cbiffle, line 17: Nit: indent here is different from the rest of the file.

sys/abi/src/lib.rs
```diff
@@ -598,6 +598,32 @@ impl core::convert::TryFrom<u32> for Sysnum {
...
+#[repr(u16)]
+pub enum Kipcnum {
+    ReadTaskStatus = 1,
+    RestartTask = 2,
+    FaultTask = 3,
+    ReadImageId = 4,
+    Reset = 5,
+}
+
+impl core::convert::TryFrom<u16> for Kipcnum {
```
cbiffle, line 612: Recommending deriving `FromPrimitive` for this.

sys/abi/src/lib.rs
```diff
@@ -598,6 +598,32 @@ impl core::convert::TryFrom<u32> for Sysnum {
     }
 }
 
+/// Representation of kipc numbers
+/// These get passed as an operation which is passed as a u16
+#[repr(u16)]
```
cbiffle, line 603: Explicit repr here is probably not necessary since you're not zerocopying them.

sys/userlib/src/kipc.rs
```diff
@@ -24,13 +29,31 @@ pub fn restart_task(task: usize, start: bool) {
...
+        TaskId::KERNEL,
+        Kipcnum::FaultTask as u16,
+        task.as_bytes(),
+        &mut [],
+        &[],
+    );
+    assert_eq!(rc, 0);
+}
+
+pub fn system_restart() {
```
cbiffle, line 55: This should probably return `!`, shouldn't it?

sys/abi/src/lib.rs
labbott: I was following the convention of of defining it manually based on this comment for other functions

```
/// We're using an explicit `TryFrom` impl for `Sysnum` instead of
/// `FromPrimitive` because the kernel doesn't currently depend on `num-traits`
/// and this seems okay.
```

PR #697, opened 2022-08-09
Add Hubris ID to stage0

labbott:
While not a true 'hubris' image, adding a hubris ID to stage0 lets
it be used with some humility commands.


mkeeter (approved):

flihp (approved):
Thanks!

PR #698, opened 2022-08-09
minimum self-signed DICE

flihp:
This PR adds a minimal, self-signed DeviceId and Alias cert creation process to stage0. The majority of the functionality is collected in a library that provides a types for various artifacts (Cdi, CdiL0, AliasOkm etc), functions for their derivation, and an interface to build the DeviceId and Alias certs. The artifacts created by stage0 (Alias seed, Alias cert & DeviceId cert) are then passed through RAM to a demonstration / test task in hubris. This task extracts the relevant artifacts, checks signatures on the certs and dumps the FWID out to a ringbuf.

NOTE: The demonstration task should be removed from this PR before merging as it serves no useful purpose beyond demo / debug.


lib/dice/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+            // Support from the lpc55-pac would help:
+            // https://github.com/lpc55/lpc55-pac/issues/15
+            // https://github.com/lpc55/lpc55-pac/pull/16
+            let n = unsafe { ptr::read_volatile(p) };
+
+            cdi[offset..offset + step].copy_from_slice(&n.to_ne_bytes());
+
+            offset += step;
+            addr += step as u32;
+        }
```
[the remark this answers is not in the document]
jgallagher: Untested, but I think something like this would work (with the caveat that I'm not sure why `ptr::read_volatile(p)` was used instead of `ptr::read(p)`, and if that's important this needs tweaking):

```rust
let mut cdi = [0u8; SEED_LENGTH];
// TODO document why this is safe
let source = unsafe {
    core::slice::from_raw_parts(
        Self::REG_ADDR_NONSEC as *const u32,
        SEED_LENGTH / mem::size_of::<u32>(),
    )
};

for (dst, src) in cdi.chunks_mut(mem::size_of::<u32>()).zip(source) {
    dst.copy_from_slice(&src.to_ne_bytes());
}
```

task/dice-data/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+        ringbuf_entry!(Trace::AliasSigCheck(true));
+    } else {
+        ringbuf_entry!(Trace::AliasSigCheck(false));
+        panic!();
+    }
+
+    // check that device id cert was signed by device id key (self signed)
+    let data = deviceid_cert.get_signdata();
+    let sig: Signature = deviceid_cert.get_sig().into();
+    if deviceid_pub.verify(data, &sig).is_ok() {
```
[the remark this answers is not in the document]
flihp: The only data I have is relative and generic (not specific to the salty implementation): https://www.cryptopp.com/wiki/Ed25519. I don't know of any existing performance metrics for the salty ed25519 implementation. We could probably hack up a test to do verify operations in a loop toggling a GPIO to collect timing metrics.

stage0/src/dice.rs
```diff
@@ -0,0 +1,87 @@
...
+
+fn get_deviceid_keypair(cdi: &Cdi) -> Keypair {
+    let devid_okm = DeviceIdOkm::from_cdi(cdi);
+
+    Keypair::from(devid_okm.as_bytes())
+}
+
+fn get_serial_number() -> SerialNumber {
+    // get serial number from somewhere
+    SerialNumber::from_str("0123456789ab").expect("SerialNumber::from_str")
```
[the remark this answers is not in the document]
flihp: discussion of serial numbers is in two RFDs:
- https://rfd.shared.oxide.computer/rfd/0219
- https://rfd.shared.oxide.computer/rfd/0273

From RFD 273 it looks like we may need to accommodate a part number as well but I need to follow up with @kc8apf.

lib/dice/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+
+            unsafe {
+                ptr::write(p, 0);
+            }
+
+            offset += step;
+            addr += step as u32;
+        }
+    }
+}
```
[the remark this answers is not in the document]
flihp: oof. Doesn't look like there's any difference. Looks like duplicate code from a copy / pasta. Removing.

lib/dice/src/lib.rs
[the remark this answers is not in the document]
flihp: Great feedback and thanks for the code snip. I hadn't discovered `core::slice::from_raw_parts` till I had to turn the Hubris image into a slice for hashing. I didn't see it's applicability here but this makes perfect sense. I've made a similar change to the `clear_register` function as well.

stage0/src/dice.rs
```diff
@@ -0,0 +1,87 @@
...
+    cert_sn += 1;
+
+    // Collect hash(es) of TCB. The first TCB Component Identifier (TCI)
+    // calculated is the Hubris image. The DICE specs call this collection
+    // of TCIs the FWID. This hash is stored in keeys certified by the
+    // DeviceId. This hash should be 'updated' with relevant configuration
+    // and code as FWID for Hubris becomes known.
+    let mut fwid = Sha3_256::new();
+    fwid.update(image.as_bytes());
+    let fwid = fwid.finalize();
```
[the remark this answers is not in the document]
flihp: Initially I had expected that we would need to measure a number of components as part of the FWID. But after reading through your contribution to rfd 287 and the stage0 code it was clear that all of the data making up the TCB (at least the stuff I'm aware of) comes directly from the hubris image (vector table, entry point, sp etc). This makes the calculation a single hash / 'update' and a dedicated function seemed like overkill. If this becomes more complicated I agree that moving it into a function would make sense.

lib/dice/src/deviceid_cert_tmpl.rs
```diff
@@ -0,0 +1,97 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// NOTE: This DER blob, offsets & lengths is mostly generated code.
+// TODO: generate cert template DER from ASN.1 & text config
```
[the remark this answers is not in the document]
flihp: My first pass at implementing this did generate this stuff at build time. After bike shedding on that code for a while I decided that committing it would be a distraction for a few reasons: The implementation was pretty gnarly on account of shelling out to the openssl ca tools adding an additional dependency that's awkward (openssl-sys?) and slow to build. The ad-hoc parsing of the certs was ugly too and brittle. All of this would be added only to be removed and replaced with something less horrific.

Adding this back is possible but it'll add stuff to the build that I think we don't want (openssl command line tools) and create a lot of churn in the future. I've added the comment you requested here in my latest update to this PR and I'll mark this conversation as resolved. Don't hesitate to re-open the conversation if you disagree or have any other questions.

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,340 @@
...
+    #[test]
+    fn issuer_sn_from_new() {
+        let sn = SerialNumber::from_str("0123456789ab").expect("SN from_str");
+        let cert = DeviceIdCert::new().set_issuer_sn(&sn);
+
+        assert_eq!(get_issuer_sn(&cert), sn.as_bytes());
+    }
+
+    #[test]
+    fn subject_cn_from_new() {
```
kc8apf, line 281:
```suggestion
    fn subject_sn_from_new() {
```

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,73 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::{AliasCert, AliasOkm, DeviceIdCert};
+use hubpack::SerializedSize;
+use lpc55_pac::syscon::RegisterBlock;
+use serde::{Deserialize, Serialize};
+
+// This memory is the USB peripheral SRAM that's 0x4000 bytes long. Changes
+// to this address must be coordinated with the [dice_*] tables in
+// chips/lpc55/chips.toml
```
kc8apf, line 12: build.rs could generate these from the config.

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,73 @@
...
+            .presetctrl2
+            .modify(|_, w| w.usb1_ram_rst().asserted());
+        self.0.ahbclkctrl2.modify(|_, w| w.usb1_ram().disable());
+    }
+
+    pub fn alias(&self, alias_handoff: &AliasHandoff) -> usize {
+        let dst_ptr = ALIAS_START as *mut [u8; AliasHandoff::MAX_SIZE];
+        // SAFETY: Dereferencing this raw pointer is necessary to write to the
+        // memory region used to handoff DICE artifacts to Hubris tasks. This
+        // pointer will always references a valid memory region.
```
kc8apf, line 40: Explain why you know it will always be valid

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,73 @@
...
+        hubpack::serialize(dst, alias_handoff).expect("serialize alias-handoff")
+    }
+}
+
+// Type to access the memory region use to hand DICE derived artifacts off
+// to the task acting as the root of trust for reporting.
+// TODO: This needs to be made generic to handle an arbitrary cert chain
+// instead of individual certs.
+#[derive(Deserialize, Serialize, SerializedSize)]
+pub struct AliasHandoff {
```
kc8apf, line 53: I'm confused why this is separate from Handoff.  The names probably aren't helping.

lib/dice/src/handoff.rs
flihp: This was a pretty bad safety comment. Apologies for that. I've updated both SAFETY comments in this file to include an enumeration of the preconditions that must be met for the call to be safe.

lib/dice/src/handoff.rs
flihp: I spent a bit of time looking at this and while I agree in general it would require adding some plumbing to xtask. The data we would need is in the 'chip' key from the top level toml table but AFAIK the currently available methods for getting at the app toml don't expose it:
- `build_util::config` exposes the app config that includes tables under `[config]`
- `build_util::task_config` exposes tables under the `[task_name]` table
So that's not to say it can't be done, but AFAIK doing so would require a new interface in build_util and some plumbing in xtask to expose the chip key / file path. This is definitely something that should be added but maybe it can wait till after this PR is merged? https://github.com/oxidecomputer/hubris/issues/731

lib/dice/src/handoff.rs
flihp: I can see how this would be confusing. I've gone back and renamed `AliasHandoff` to `AliasData` as it's a container to hold the Alias DICE credentials. I've also bulked up the comments on these two types to make clear their intended use.

stage0/src/dice.rs
```diff
@@ -0,0 +1,81 @@
...
+use unwrap_lite::UnwrapLite;
+
+fn get_deviceid_keypair(cdi: &Cdi) -> Keypair {
+    let devid_okm = DeviceIdOkm::from_cdi(cdi);
+
+    Keypair::from(devid_okm.as_bytes())
+}
+
+fn get_serial_number() -> SerialNumber {
+    // get serial number from somewhere
```
kc8apf, line 23: Mark this with a TODO (and file an issue) so we don't forget.

lib/dice/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+        &self.0
+    }
+}
+
+impl Cdi {
+    const REG_ADDR_NONSEC: u32 = 0x40000900;
+
+    // Read the CDI from registers & return to the caller.
+    // Clear the CDI after it's consumed to prevent future access.
+    pub fn new() -> Option<Self> {
```
kc8apf, line 44: This function has side effects that I wouldn't expect for new().  Namely, calling it multiple times is guaranteed to fail which is closer to a singleton than a normal struct.

stage0/src/dice.rs
```diff
@@ -0,0 +1,81 @@
...
+        .set_fwid(fwid.as_ref())
+        .sign(&deviceid_keypair);
+
+    let alias_data = AliasData {
+        seed: alias_okm,
+        alias_cert,
+        deviceid_cert,
+    };
+
+    handoff.alias(&alias_data);
```
kc8apf, line 80:
```suggestion
    handoff.store(&alias_data);
```

stage0/src/image_header.rs
```diff
@@ -56,6 +56,22 @@ impl Image {
...
 
+    #[cfg(feature = "dice")]
+    fn get_img_size(&self) -> Option<usize> {
+        use core::convert::TryFrom;
+
+        usize::try_from((unsafe { &*self.get_header() }).total_image_len).ok()
+    }
+
+    #[cfg(feature = "dice")]
+    pub fn as_bytes(&self) -> &[u8] {
```
kc8apf, line 67: This feels like a place where an attacker would try to confuse stage0 by tampering with the img_size.  If I set the img_size to 1, the Cdi will only include the first byte but we'll boot the image, which must be more than 1 byte long, anyway.  We probably need to hash the entire image slot, not just the used area so we aren't relying on attacker-provided data.

chips/lpc55/chip.toml
```diff
@@ -46,6 +46,12 @@ size = 4096
 address = 0x4003A000
 size = 4096
 
+# this is the start of the USB SRAM AHB peripheral (0x4000 bytes total)
+# we appropriate this SRAM for passing DICE artifacts
+[dice_alias]
```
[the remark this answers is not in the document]
flihp: Normally I'd agree with you here but eventually we'll be transferring other collections of artifacts across this boundary. In a future where we're transferring credentials to multiple recipients / hubris tasks we'll want guarantees that each task is only able to access the appropriate credentials. AFAIK, by dividing the USB SRAM into regions in the chip.toml we're able to grant access to each region using the existing mechanisms in the kernel. I wrote this up in RFD 282: https://rfd.shared.oxide.computer/rfd/0282#_ram

Use of the USB SRAM was a choice of convenience and so if the issue is with this memory being associated with a peripheral we could select any sufficiently large & unused region of memory for this purpose. This would still require that we list it in the chip.toml file as separate memory regions but it would keep us from labeling the USB SRAM as anything but what it is.

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,340 @@
...
+}
+
+#[derive(Deserialize, Serialize, SerializedSize)]
+pub struct DeviceIdCert(
+    #[serde(with = "BigArray")] [u8; deviceid_cert_tmpl::SIZE],
+);
+
+impl DeviceIdCert {
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
flihp: In this case the difference between 'new' and 'default' seemed arbitrary. I'm fine with using default instead.

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,340 @@
...
+    NoPubKey,
+    NoSig,
+    NoSignData,
+    TooSmall,
+    NotFound,
+    NoCn,
+}
+
+#[derive(Deserialize, Serialize, SerializedSize)]
+pub struct DeviceIdCert(
```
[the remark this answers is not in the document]
flihp: This cert is self signed and so we generate it locally. When we add the manufacturing protocol we'll generate a CSR, send it out to the mfg system and get a cert back. We'll then write it to flash where it will be effectively read-only as you expect.

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub fn sign(self, keypair: &Keypair) -> Self {
+        let signdata = self.get_signdata();
+        let sig = keypair.sign(signdata);
+
+        self.set_sig(&sig.to_bytes())
+    }
+}
+
+#[derive(Deserialize, Serialize, SerializedSize)]
+pub struct AliasCert(#[serde(with = "BigArray")] [u8; alias_cert_tmpl::SIZE]);
```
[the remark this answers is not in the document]
flihp: I wrestled with this quite a bit. The issue I ran into is that each function that's duplicated across the two types is operating on different offsets. Using `set_serial_number` as an example the impl for DeviceIdCert is:

```
pub fn set_serial_number(mut self, sn: u8) -> Self {
    self.0[deviceid_cert_tmpl::SERIAL_NUMBER_START
        ..deviceid_cert_tmpl::SERIAL_NUMBER_END]
        .copy_from_slice(&sn.to_be_bytes());

        self
}
```
while the impl for AliasCert is:
```
pub fn set_serial_number(mut self, sn: u8) -> Self {
    self.0[alias_cert_tmpl::SERIAL_NUMBER_START
        ..alias_cert_tmpl::SERIAL_NUMBER_END]
        .copy_from_slice(&sn.to_be_bytes());

    self
}
```

The only difference between these two functions are the different offsets used to create the slices. I may be understanding the Rust part of this wrong but AFAIK default implementations in traits cannot assume anything about the fields in the types so to get these offsets we'd have to define `get_*` functions in the trait for each offset. This will double the number of boilerplate functions (a `get` function for each start and end offset) though this may be an improvement since these `get` functions will only return an offset, and so as far as boilerplate goes that's pretty good.

I think after all this I've just talked myself into the idea. I'll hack something up to see how it looks.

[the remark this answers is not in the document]
flihp: This is actually better than I had expected. I didn't realize how useful associated constants in traits would be here.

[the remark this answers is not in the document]
flihp: I ended up banging my head against a known issue in my attempts to create a more generic interface with shared implementation: https://github.com/rust-lang/rust/issues/60551. I'm currently trying some of the suggested workarounds that primarily rely on using constant generic parameters.

stage0/src/dice.rs
flihp: https://github.com/oxidecomputer/hubris/issues/734

lib/dice/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+pub enum SNError {
+    BadSize,
+}
+
+impl FromStr for SerialNumber {
+    type Err = SNError;
+
+    fn from_str(sn: &str) -> Result<Self, Self::Err> {
+        let sn: [u8; SN_LENGTH] =
+            sn.as_bytes().try_into().map_err(|_| SNError::BadSize)?;
```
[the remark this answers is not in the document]
flihp: the serial number Attribute is defined as a PrintableString so it's a subset of ASCII: https://en.wikipedia.org/wiki/PrintableString. I'll create another ticket to add some filtering & error handling.

lib/dice/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+
+pub const SEED_LENGTH: usize = SECRETKEY_SEED_LENGTH;
+pub const SN_LENGTH: usize = deviceid_cert_tmpl::SN_LENGTH;
+
+pub trait SeedBuf {
+    fn as_bytes(&self) -> &[u8; SEED_LENGTH];
+}
+
+#[derive(Zeroize)]
+#[zeroize(drop)]
```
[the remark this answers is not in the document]
flihp: This library is using the older version of the Zeroize trait due to an incompatibility with the Rsa crate used by the lpc55_support crate. I've only upgraded the Rsa crate used by lpc55_support the other day. Moving that crate forward and using the newer version here shouldn't be too much work.

lib/dice/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+        // LPC55 DICE registers. This pointer will always reference valid
+        // memory as this is where the LPC55 maps the DICE registers.
+        let cdi = unsafe {
+            slice::from_raw_parts_mut(
+                Self::REG_ADDR_NONSEC as *mut u32,
+                SEED_LENGTH / mem::size_of::<u32>(),
+            )
+        };
+
+        cdi.fill(0);
```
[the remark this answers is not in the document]
flihp: In manual testing its not compiled out & the registers are cleared but that's not to say that the compiler may change its mind tomorrow. Using Zeroize is an interesting way to do this but it changes the behavior of the side effects discussed here: https://github.com/oxidecomputer/hubris/pull/698#discussion_r954399499.

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,90 @@
...
+        // handoff through USB SRAM requires we power it on
+        syscon.ahbclkctrl2.modify(|_, w| w.usb1_ram().enable());
+        syscon
+            .presetctrl2
+            .modify(|_, w| w.usb1_ram_rst().released());
+
+        Self(syscon)
+    }
+
+    pub fn turn_off(self) {
```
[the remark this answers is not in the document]
flihp: AFAIK it will eventually but YMMV depending on a number of variables like how long between cycling power off & back on.

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,90 @@
...
+// chips/lpc55/chip.toml
+// TODO: get from app.toml -> chip.toml at build time
+const MEM_START: usize = 0x4010_0000;
+const ALIAS_START: usize = MEM_START;
+
+/// The Handoff type is a thin wrapper over the memory region used to transfer
+/// DICE artifacts (seeds & certs) from stage0 to hubris tasks. It is intended
+/// for use by stage0 to write these artifacts to memory where they will later
+/// be read out by a hubris task.
+pub struct Handoff<'a>(&'a RegisterBlock);
```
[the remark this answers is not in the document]
flihp: It could. The use case described in https://rfd.shared.oxide.computer/rfd/0282#_ram is pretty limited though. I'm trying to keep things simple and explicit. I'm not against a more generic interface but deferring till there's a use case driving it seems prudent.

lib/dice/src/cert.rs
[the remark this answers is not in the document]
flihp: This wasn't as bad as expected. All of the implementation from the DeviceIdCert is now in a trait. The only additional code in the AliasCert is a get/set for the FWID (the only field in the AliasCert that's not in the DeviceIdCert). I had to compromise on a few sized arrays in the interface that are now slices but these can be added back once https://github.com/rust-lang/rust/issues/60551 moves forward.

stage0/src/image_header.rs
flihp: I've fleshed out my understanding of the threat and mitigations in an issue & have added it to the Trust Quorum backlog: https://github.com/oxidecomputer/hubris/issues/736

lib/dice/src/handoff.rs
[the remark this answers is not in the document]
flihp: I should note that this isn't explicitly documented in the user manual. This is just my general understanding of how SRAM works.

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    pub fn alias(&self, alias_data: &AliasData) -> usize {
+        let dst_ptr = ALIAS_START as *mut [u8; AliasData::MAX_SIZE];
+        // SAFETY: Dereferencing this raw pointer is necessary to write to the
+        // memory region used to handoff DICE artifacts to Hubris tasks. This
+        // pointer will references a valid memory region provided two
+        // conditions are met:
+        // 1) The associated memory region has been enabled / turned on if
+        // necessary. This happens in the constructor / 'turn_on' function.
+        // 2) The function call is made by code in stage0.
```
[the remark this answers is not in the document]
flihp: Accessing the SRAM when its powered off will land you in a fault handler. We could insure against this situation by enabling the memory unconditionally in stage0. It could still be disabled further down the line but this is partially mitigated by restricting access to the syscon driver. After this the only other condition is sufficient permissions to access the memory region which means either being in stage0 (where this interface should be used) or having been granted access by the hubris kernel. Whether or not the whole method becomes unsafe because of this is rightly a question for those you've tagged above.

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,340 @@
...
+        };
+
+        // well known seed
+        let seed: [u8; 32] = [42; 32];
+        let keypair: salty::Keypair = salty::Keypair::from(&seed);
+
+        let sn = SerialNumber::from_str("0123456789ab").expect("SN from_str");
+        let utc: DateTime<Utc> = Utc.ymd(2022, 07, 13).and_hms(6, 6, 6);
+        let utc = utc.format("%y%m%d%H%M%SZ").to_string();
+        let cert = DeviceIdCert::new()
```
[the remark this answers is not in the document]
flihp: I've taken a swing at simplifying this interface replacing the janky not-a-builder pattern with simple 'new' functions that return fully populated & signed certs.

lib/dice/src/lib.rs
flihp: Good point. I've renamed this function to 'from_reg' to indicate the constructor is populating the Cdi from related registers. The method used to clear the DICE registers has been simplified significantly as well. Finally I've documented the side effects more clearly / explicitly.

chips/lpc55/chip.toml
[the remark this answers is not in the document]
mkeeter: Could we use SRAM 4 instead (`0x20040000-0x20043FFF`)?  That seems less weird than using spare USB SRAM, and isn't otherwise used in the system.

(I'm not sure if we'd need to jump through any hoops at startup to turn it on, or whether it's on immediately)

lib/dice/src/alias_cert_tmpl.rs
```diff
@@ -0,0 +1,111 @@
...
+// DICE specific extensions and policy) using the openssl ca tools. The fields
+// that we need to operate on were identified and their offsets recorded. The
+// values in these regions (signatureValue, serialNumber, issuer, subject,
+// validity etc) are then removed.
+//
+// TODO: generate cert template DER from ASN.1 & text config
+#[allow(dead_code)]
+pub const SIZE: usize = 608;
+#[allow(dead_code)]
+pub const SERIAL_NUMBER_START: usize = 15;
```
mkeeter, line 16: Nit: these could be more compactly expressed as ranges, e.g.
```rust
pub const SERIAL_NUMBER_RANGE: core::ops::Range<usize> = 15..16;
```

Doing so would reduce the potential for copy-paste errors!

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,346 @@
...
+    TooSmall,
+    NotFound,
+    NoCn,
+}
+
+pub trait Cert {
+    fn as_bytes(&self) -> &[u8];
+    fn as_mut_bytes(&mut self) -> &mut [u8];
+
+    const SERIAL_NUMBER_START: usize;
```
mkeeter, line 32: (same comment here about using `core::ops::Range<usize>`)

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    const ISSUER_SN_START: usize;
+    const ISSUER_SN_END: usize;
+
+    fn get_issuer_sn(&self) -> SerialNumber {
+        SerialNumber::from_bytes(
+            &self.as_bytes()[Self::ISSUER_SN_START..Self::ISSUER_SN_END]
+                .try_into()
+                .unwrap_lite(),
```
mkeeter, line 60: Why are we using `unwrap_lite` here and `unwrap` at line 39?

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,346 @@
...
+}
+
+pub trait Cert {
+    fn as_bytes(&self) -> &[u8];
+    fn as_mut_bytes(&mut self) -> &mut [u8];
+
+    const SERIAL_NUMBER_START: usize;
+    const SERIAL_NUMBER_END: usize;
+
+    fn get_serial_number(&self) -> u8 {
```
mkeeter, line 35: Optional:

If you want to get fancy here, you could add a helper function of the form
```rust
    fn read_range<'a, T>(&'a self, r: core::ops::Range<usize>) -> T
    where T: TryFrom<&'a [u8]>
    {
        self.as_bytes()[r].try_into().unwrap_lite()
    }
```

Then, `get_serial_number` would become
```rust
fn get_serial_number(&self) -> u8 {
    u8::from_be_bytes(self.read_range(Self::SERIAL_NUMBER_RANGE))
}
```
(and similarly, the other getters would be a little less verbose)

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,346 @@
...
+    #[serde(with = "BigArray")] [u8; deviceid_cert_tmpl::SIZE],
+);
+
+impl DeviceIdSelfCert {
+    pub fn new(
+        cert_sn: u8,
+        dname_sn: &SerialNumber,
+        keypair: &Keypair,
+    ) -> Self {
+        let mut buf = [0u8; deviceid_cert_tmpl::SIZE];
```
mkeeter, line 171:
```rust
let mut buf = deviceid_cert_tmpl::CERT_TMPL.clone()
```
?

task/dice-data/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+enum Trace {
+    PubCheck(bool),
+    AliasSigCheck(bool),
+    DeviceIdSigCheck(bool),
+    Fwid([u8; 32]),
+    AllChecksPassed,
+    None,
+}
+
+ringbuf!(Trace, 64, Trace::None);
```
mkeeter, line 27: This will take up a moderate amount of RAM (~2K).  Probably not a big deal, but worth looking at if we run low.

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,346 @@
...
+        deviceid_cert_tmpl::fill(&mut buf);
+
+        Self(buf)
+            .set_serial_number(cert_sn)
+            .set_issuer_sn(dname_sn)
+            .set_subject_sn(dname_sn)
+            .set_pub(keypair.public.as_bytes())
+            .sign(keypair)
```
mkeeter, line 179: Vague architectural vibes (no need to take action): the API would be more robust if we split up the intermediate vs signed certificate, e.g. if we had `CertBuilder::sign(self, keypair: &Keypair) -> Cert`

This would protect us from editing the certificate after it was signed.  I'm not exactly sure how this would work with the generics – maybe the `Cert` type would be an associated type on `CertBuilder`, and it would store the various fields?

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    const PUB_START: usize;
+    const PUB_END: usize;
+
+    fn get_pub(&self) -> &[u8] {
+        &self.as_bytes()[Self::PUB_START..Self::PUB_END]
+    }
+
+    fn set_pub(mut self, pubkey: &[u8; PUBLICKEY_SERIALIZED_LENGTH]) -> Self
```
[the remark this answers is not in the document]
flihp: We could. This case is a bit weird since this `_LENGTH` is coming from a different library. But either way `_START`, `_END`, `_LENGTH` stuff from the `*_tmpl.rs` is mostly generated at this point so the `_LENGTH` values in these files are just calculated values from `_START` and `_END` behind the scenes.

lib/dice/src/alias_cert_tmpl.rs
flihp: TIL `core::ops::Range` exists. The code that generates these tmpl files can't go into the hubris tree cleanly since I'm using a bunch of external tools to do ASN.1 / x509 crimes but I'll queue this change up for the code gen. This may also help w/ your previous point about `_LENGTH` values.

flihp: https://github.com/oxidecomputer/hubris/issues/747

lib/dice/src/cert.rs
flihp: I think I only found `unwrap_lite` as I was doing the implementation here. I thought I had replaced all of the `unwrap`s with `unwrap_lite`s but looks like I missed at least one. I'll take a pass over the code to make sure there aren't others.

lib/dice/src/cert.rs
flihp: This one falls into the "I didn't know I could do that" category. I still qualify as a recovering C programmer I guess :disappointed:

lib/dice/src/cert.rs
flihp: I definitely like the idea of minimizing potential mistakes. I'm not sure how the generics would work either though. The two cert types have different fields (only one, but still different) so we'd need a different builder for each cert type?

mkeeter: Ah, I didn't notice that there was a difference in the Wall Of Fields.  In that case, we'd need separate builder types for each one (which might even let us kill the `Cert` trait entirely?)

lib/dice/src/handoff.rs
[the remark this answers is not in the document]
steveklabnik: I don't think that that means this becomes unsafe. The key question with unsafe is, "If I call this method with a certain parameter, will I introduce memory unsafety?" If the answer is, in all cases, that you end up in a fault handler, doesn't sound like memory unsafety to me.

lib/dice/src/handoff.rs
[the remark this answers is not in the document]
kc8apf: Is the expectation that stage0 never calls this and thus leaves it powered on when jumping into RoT image?

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn store(&self, alias_data: &AliasData) -> usize {
+        let dst_ptr = ALIAS_START as *mut [u8; AliasData::MAX_SIZE];
+        // SAFETY: Dereferencing this raw pointer is necessary to write to the
+        // memory region used to handoff DICE artifacts to Hubris tasks. This
+        // pointer will references a valid memory region provided two
+        // conditions are met:
+        // 1) The associated memory region has been enabled / turned on if
+        // necessary. This happens in the constructor / 'turn_on' function.
+        // 2) The function call is made by code in stage0.
+        let dst: &mut [u8] = unsafe { &mut *dst_ptr };
```
kc8apf, line 50:
```suggestion
        let dst: &mut [u8] = unsafe { core::slice::from_raw_parts_mut(ALIAS_START as *mut u8, AliasData::MAX_SIZE) };
```

lib/dice/src/handoff.rs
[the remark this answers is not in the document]
kc8apf: Feels worth clarifying in the safety comment that this is safe even if these conditions aren't met since it will cause a fault.

lib/dice/src/handoff.rs
kc8apf: Should also check that AliasData::MAX_SIZE fits in the dice_alias peripheral's range.  That should be a compile-time check of some sort.

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,90 @@
...
+        let src_ptr = ALIAS_START as *const [u8; AliasData::MAX_SIZE];
+        // SAFETY: Dereferencing this raw pointer is necessary to read from the
+        // memory region used to transfer the Alias DICE artifacts from stage0
+        // to a Hubris task. This pointer will reference a valid memory region
+        // provided two conditions are met:
+        // 1) The associated memory region has been enabled / turned on if
+        // necessary. This should be done by code in stage0.
+        // 2) The task making the call has been granted access to the memory
+        // region by the kernel.
+        let src: &[u8] = unsafe { &*src_ptr };
```
kc8apf, line 82:
```suggestion
        let src: &[u8] = unsafe { core::slice::from_raw_parts(ALIAS_START as *const u8, AliasData::MAX_SIZE) };
```

lib/dice/src/handoff.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn from_mem() -> Self {
+        let src_ptr = ALIAS_START as *const [u8; AliasData::MAX_SIZE];
+        // SAFETY: Dereferencing this raw pointer is necessary to read from the
+        // memory region used to transfer the Alias DICE artifacts from stage0
+        // to a Hubris task. This pointer will reference a valid memory region
+        // provided two conditions are met:
+        // 1) The associated memory region has been enabled / turned on if
+        // necessary. This should be done by code in stage0.
+        // 2) The task making the call has been granted access to the memory
+        // region by the kernel.
```
kc8apf, line 81: Same comment about noting that this is safe even if these conditions are violated as it will cause a fault.

lib/dice/src/handoff.rs
kc8apf: Same comment about checking that AliasData::MAX_SIZE is less than or equal to dice_alias peripheral size.

lib/dice/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+}
+
+impl CdiReg {
+    fn is_clear(&self) -> bool {
+        if self.0.iter().all(|&w| w == 0) {
+            true
+        } else {
+            false
+        }
```
kc8apf, line 65:
```suggestion
        self.0.iter().all(|&w| w == 0)
```

lib/dice/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        // When registers holding CDI have been cleared / zeroed return None
+        // to prevent unsuspecting consumers from deriving keys from 0's.
+        if cdi_reg.is_clear() {
+            return None;
+        }
+
+        for (dst, src) in cdi
+            .chunks_exact_mut(mem::size_of::<u32>())
+            .zip(cdi_reg.0.as_ref())
```
kc8apf, line 106: impl SeedBuf on CdiReg instead of cdi_reg.0.as_ref()?

stage0/src/dice.rs
```diff
@@ -0,0 +1,80 @@
...
+// TODO: get the legit SN from somewhere
+// https://github.com/oxidecomputer/hubris/issues/734
+fn get_serial_number() -> SerialNumber {
+    SerialNumber::from_str("0123456789ab").expect("SerialNumber::from_str")
+}
+
+pub fn run(image: &Image) {
+    let cdi = match Cdi::from_reg() {
+        Some(cdi) => cdi,
+        None => return,
```
kc8apf, line 31: If the CDI isn't available, this is going to skip turning on the USB SRAM.  That means that any task that attempts to access the DICE artifacts in stage1 will fault when they try to access it.  Probably better to enable the USB SRAM and write out a well-known pattern indicating that DICE was disabled in stage0 so that can reported upstream.

lib/dice/src/cert.rs
flihp: This will probably cause everything in the `Cert` trait to move into another trait for the builders. Just moves things around.

lib/dice/src/cert.rs
```diff
@@ -0,0 +1,340 @@
...
+    const SIG_EXPECTED: [u8; SIGNATURE_SERIALIZED_LENGTH] = [
+        0x26, 0x2A, 0x81, 0xE9, 0x1F, 0x06, 0xCF, 0xF0, 0x13, 0xEB, 0x33, 0x71,
+        0x5A, 0xB9, 0x5C, 0xC0, 0xC7, 0x40, 0x01, 0x83, 0x7C, 0xB6, 0x2F, 0x2E,
+        0x88, 0xE9, 0x95, 0xD9, 0x10, 0x9C, 0xD8, 0xF5, 0x33, 0x4B, 0x9B, 0xB1,
+        0x6A, 0xB3, 0x23, 0xDB, 0x3A, 0x1C, 0x35, 0x31, 0xE7, 0x38, 0xEC, 0x9B,
+        0xAA, 0x32, 0x36, 0x5A, 0xAA, 0x37, 0x4B, 0xF5, 0xE7, 0x7A, 0x2C, 0x4E,
+        0x88, 0x35, 0x50, 0x0E,
+    ];
+
+    // Each time this library is built the notBefore time in the validity
```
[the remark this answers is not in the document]
flihp: I had initially set this up such that the cert templates were generated at build time. I pulled this out because it had weird dependencies on external tools used to create the x.509 structures. In constructing the templates I was initially setting the notBefore field with the system time before I realized that this would break reproducible builds. This can be done in such a way it won't break reproducible builds but it requires support from xtask. I'll make a ticket and add a TODO.

lib/dice/src/lib.rs
flihp: CdiReg is a slice of u32s not bytes like the SeedBufs. The type exists pretty much exclusively to ensure the registers that it wraps are zeroed when it's dropped.

lib/dice/src/handoff.rs
flihp: https://github.com/oxidecomputer/hubris/issues/751

lib/dice/src/lib.rs
flihp: oof :rofl:

lib/dice/src/cert.rs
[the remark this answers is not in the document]
flihp: https://github.com/oxidecomputer/hubris/issues/752

stage0/src/dice.rs
flihp: This came up elsewhere in the review and I :100: agree. For now the well-known pattern is 0's but I'll do some testing to see how well this works in practice.

PR #704, opened 2022-08-10
Add Idol API to read Monorail MAC table

mkeeter:
Another step towards replacing `ringbuf` debugging with proper APIs!

### DIFF
--- a/drv/vsc7448/src/mac.rs
+++ b/drv/vsc7448/src/mac.rs
@@ -8,7 +8,7 @@ use vsc7448_pac::*;
 
 /// Represents an entry in the VSC7448's MAC tables
 #[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
-pub struct MacTableEntry {
+pub struct Vsc7448MacTableEntry {
     pub mac: [u8; 6],
     src_kill_fwd: bool,
     pub addr: u16,
@@ -34,7 +34,9 @@ pub fn count_macs(v: &impl Vsc7448Rw) -> Result<usize, VscError> {
     Ok(i)
 }
 
-pub fn next_mac(v: &impl Vsc7448Rw) -> Result<Option<MacTableEntry>, VscError> {
+pub fn next_mac(
+    v: &impl Vsc7448Rw,
+) -> Result<Option<Vsc7448MacTableEntry>, VscError> {
     // Trigger a FIND_SMALLEST action then wait for it to finish
     let ctrl = LRN().COMMON().COMMON_ACCESS_CTRL();
     v.write_with(ctrl, |r| {
@@ -55,7 +57,7 @@ pub fn next_mac(v: &impl Vsc7448Rw) -> Result<Option<MacTableEntry>, VscError> {
     if msb == 0 && lsb == 0 {
         Ok(None)
     } else {
-        let mut out = MacTableEntry::default();
+        let mut out = Vsc7448MacTableEntry::default();
         out.mac[0..2].copy_from_slice(&msb.to_be_bytes()[2..]);
         out.mac[2..6].copy_from_slice(&lsb.to_be_bytes());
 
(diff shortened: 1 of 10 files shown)


drv/vsc7448/src/mac.rs
```diff
@@ -23,7 +23,20 @@ pub struct MacTableEntry {
     valid: bool,
 }
 
-pub fn next_mac(v: &impl Vsc7448Rw) -> Result<Option<MacTableEntry>, VscError> {
+pub fn count_macs(v: &impl Vsc7448Rw) -> Result<usize, VscError> {
+    // Reset MAC table reader, so FIND_SMALLEST starts from 0
```
jgallagher, line 27: Possibly an ignorant question, not knowing the vsc7448 at all, but does the MAC table reader need to be reset in between calling `count_macs()` and iterating through with `next_mac()`?

task/monorail-server/src/server.rs
```diff
@@ -643,6 +643,37 @@ impl<'a, R: Vsc7448Rw> idl::InOrderMonorailImpl for ServerImpl<'a, R> {
...
+
+    fn read_vsc7448_next_mac(
+        &mut self,
+        _msg: &userlib::RecvMessage,
+    ) -> Result<MacTableEntry, RequestError<MonorailError>> {
+        let mac = vsc7448::mac::next_mac(self.vsc7448.rw)
+            .map_err(MonorailError::from)
+            .map_err(RequestError::from)?;
+        // Handle the case where there is no next MAC, e.g. because of a TOCTOU
+        // race with `read_vsc7448_mac_count`
```
jgallagher, line 664: Two questions:

* Is there a reason we can't return `Result<Option<MacTableEntry>, _>` instead translating a `None` into a sentinel value?
* A TOCTOU issue where the number of mac entries decreased will be visible with this code, but a TOCTOU issue where the number of mac entries increased won't, right? Do we care?

jgallagher (approved):
I left a few questions born from my ignorance of the VSC7448 and/or hiffy, but otherwise this LGTM :+1:.

drv/vsc7448/src/mac.rs
mkeeter: Good question!

Under the hood, we're doing a `FIND_SMALLEST` operation.  This operation reads the "current smallest" MAC address from `MAC_ACCESS_CFG0/1`, finds the next-smallest MAC address in the table, then writes it to those same registers.

This means if you do multiple `FIND_SMALLEST` calls, it will iterate over all MAC addresses in the table.  The final call will write the values back to 0.

This means that when `count_macs()` finishes, we've effectively reset the iterator, so there's no need to explicitly clear it again.

(If we have multiple tasks calling `next_mac`, then we'd want some kind of locking in place, but that's a problem for the future)

task/monorail-server/src/server.rs
mkeeter: - I don't trust `ssmarshal` to properly pack an `Option<MacTableEntry>`, since that's called out in the docs as unallowed.  
- Correct – if the system learns more MAC addresses in between `count_macs` and `next_mac`, they wouldn't be returned.  I think that's fine, since it's morally equivalent to having called the functions _slightly_ sooner (i.e. before the new address was learned).  The one downside is that it would leave the iterator pointing to a non-zero MAC entry, but `count_macs` clears it before running, so that should be okay.

drv/vsc7448/src/mac.rs
jgallagher: Ahh, I see. Thanks!

task/monorail-server/src/server.rs
jgallagher: * Should we tack a "TODO: switch to Option<_> if/when we fully switch from `ssmarshal` to `hubpack`" somewhere?
* Makes sense :+1:

mkeeter: Good idea, done!

PR #707, opened 2022-08-12
mgmt-gateway: add support for update and reset

jgallagher:
This also includes a change to move the update server block size into `update_api` (underneath a chip-specific module). I did not remove the `block_size` runtime API call, although it's no longer necessary for clients, since they can grab the constant directly from `update_api`.

### DIFF
--- a/drv/update-api/src/lib.rs
+++ b/drv/update-api/src/lib.rs
@@ -26,4 +26,17 @@ pub enum UpdateError {
     WriteProtErr = 13,
 }
 
+pub mod stm32h7 {
+    // RM0433 Rev 7 section 4.3.9
+    // Flash word is defined as 256 bits
+    pub const FLASH_WORD_BITS: usize = 256;
+
+    // Total length of a word in bytes (i.e. our array size)
+    pub const FLASH_WORD_BYTES: usize = FLASH_WORD_BITS / 8;
+
+    // Block is an abstract concept here. It represents the size of data the
+    // driver will process at a time.
+    pub const BLOCK_SIZE_BYTES: usize = FLASH_WORD_BYTES * 32;
+}
+
 include!(concat!(env!("OUT_DIR"), "/client_stub.rs"));
--- a/task/mgmt-gateway/src/main.rs
+++ b/task/mgmt-gateway/src/main.rs
@@ -242,10 +242,7 @@ impl NetHandler {
         }
     }
 
-    fn run_until_blocked(
-        &mut self,
-        mgs_handler: &mut MgsHandler,
-    ) {
+    fn run_until_blocked(&mut self, mgs_handler: &mut MgsHandler) {
         loop {
             // Try to send first.
             if let Some(meta) = self.packet_to_send.take() {
--- a/task/mgmt-gateway/src/main.rs
+++ b/task/mgmt-gateway/src/main.rs
@@ -92,8 +92,7 @@ const SOCKET: SocketName = SocketName::mgmt_gateway;
 fn main() {
     let usart = UsartHandler::new(configure_usart());
     let mut mgs_handler = MgsHandler::new(usart);
-    let mut net_handler = NetHandler::default();
-    let rx_buf = claim_request_buf_static();
+    let mut net_handler = NetHandler::new(claim_request_buf_static());
 
     // Enbale USART interrupts.
     sys_irq_control(USART_IRQ, true);
@@ -114,7 +113,7 @@ fn main() {
         }
 
         if (note & NET_IRQ) != 0 || mgs_handler.needs_usart_flush_to_mgs() {
-            net_handler.run_until_blocked(rx_buf, &mut mgs_handler);
+            net_handler.run_until_blocked(&mut mgs_handler);
         }
     }
 }
@@ -229,23 +228,22 @@ impl UsartHandler {
 struct NetHandler {
     net: Net,
     tx_buf: [u8; SpMessage::MAX_SIZE],
+    rx_buf: &'static mut [u8; Request::MAX_SIZE],
     packet_to_send: Option<UdpMetadata>,
 }
 
-impl Default for NetHandler {
-    fn default() -> Self {
+impl NetHandler {
+    fn new(rx_buf: &'static mut [u8; Request::MAX_SIZE]) -> Self {
         Self {
             net: Net::from(NET.get_task_id()),
             tx_buf: [0; SpMessage::MAX_SIZE],
+            rx_buf,
             packet_to_send: None,
         }
     }
-}
 
-impl NetHandler {
     fn run_until_blocked(
         &mut self,
-        rx_buf: &mut [u8; Request::MAX_SIZE],
         mgs_handler: &mut MgsHandler,
     ) {
         loop {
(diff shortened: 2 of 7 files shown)


drv/update-api/src/lib.rs
```diff
@@ -26,4 +26,17 @@ pub enum UpdateError {
...
+    pub const FLASH_WORD_BITS: usize = 256;
+
+    // Total length of a word in bytes (i.e. our array size)
+    pub const FLASH_WORD_BYTES: usize = FLASH_WORD_BITS / 8;
+
+    // Block is an abstract concept here. It represents the size of data the
+    // driver will process at a time.
+    pub const BLOCK_SIZE_BYTES: usize = FLASH_WORD_BYTES * 32;
+}
+
```
labbott, line 41: Ah! I like the module solution a lot!

task/mgmt-gateway/src/main.rs
```diff
@@ -215,7 +228,6 @@ impl UsartHandler {
 
 struct NetHandler {
     net: Net,
-    rx_buf: [u8; Request::MAX_SIZE],
```
mkeeter, line 218: Optional: if you liked having the member variable, you could do something like `rx_buf: &'static mut [u8]` (then pass the static buffer to the constructor as a slice)

task/mgmt-gateway/src/mgs_handler.rs
```diff
@@ -152,4 +221,104 @@ impl SpHandler for MgsHandler {
             Err(ResponseError::Busy)
         }
     }
+
+    fn sys_reset_prepare(
```
mkeeter, line 225: Nit: this name is slightly confusing, because functions beginning with `sys_` are normally syscalls (same for `sys_reset_trigger` below).

task/mgmt-gateway/src/mgs_handler.rs
```diff
@@ -152,4 +221,104 @@ impl SpHandler for MgsHandler {
...
+            self.bytes_written += to_copy;
+
+            // If the block is full or this is the final block, send it to the
+            // update task.
+            if self.current_block.len() == self.current_block.capacity()
+                || self.bytes_written == self.total_length
+            {
+                update_task
+                    .write_one_block(current_block_index, &self.current_block)
+                    .map_err(|err| ResponseError::UpdateFailed(err as u32))?;
```
mkeeter, line 296: If this returns with an error, then we exit before clearing `self.current_block`, meaning the assertion at line 281 could fire if this function is called again.  Should we be robust to that sequence of events?

task/mgmt-gateway/src/mgs_handler.rs
```diff
@@ -152,4 +221,104 @@ impl SpHandler for MgsHandler {
...
+struct UpdateBuffer {
+    total_length: usize,
+    bytes_written: usize,
+    current_block: ArrayVec<[u8; BLOCK_SIZE_BYTES]>,
+}
+
+impl UpdateBuffer {
+    fn ingest_chunk(
+        &mut self,
+        update_task: &Update,
```
mkeeter, line 262: Dumb question: why does this need to be passed in as an argument, rather than being owned by the `UpdateBuffer`?  (That would make the name confusing, to be fair, but you could rename to something like `UpdateWorker`)

task/mgmt-gateway/src/mgs_handler.rs
```diff
@@ -152,4 +221,104 @@ impl SpHandler for MgsHandler {
...
+            });
+        }
+
+        Ok(())
+    }
+}
+
+fn claim_update_buffer_static() -> &'static mut UpdateBuffer {
+    let update_buffer_array = mutable_statics! {
+        static mut BUF: [UpdateBuffer; 1] = [Default::default(); _];
```
mkeeter, line 321: Add a TODO to explain why we're doing an array of 1 element here (since it's a `mutable_statics` limitation)

task/mgmt-gateway/src/mgs_handler.rs
```diff
@@ -125,7 +141,60 @@ impl SpHandler for MgsHandler {
...
+            });
+        }
+
+        self.update_task
+            .prep_image_update()
+            .map_err(|err| ResponseError::UpdateFailed(err as u32))?;
+
+        // We can only call `claim_update_buffer_static` once; we bail out above
+        // if `self.update_progress` is already `Some(_)`, and after we claim it
+        // here, we store that into `self.update_progress` (and never clear it).
```
kc8apf, line 172: How do we abort the update process?  Resetting the SP seems a bit heavy handed.

task/mgmt-gateway/src/mgs_handler.rs
```diff
@@ -152,4 +221,104 @@ impl SpHandler for MgsHandler {
...
+        &mut self,
+        _sender: SocketAddrV6,
+        _port: SpPort,
+    ) -> Result<Infallible, ResponseError> {
+        if !self.reset_requested {
+            return Err(ResponseError::SysResetTriggerWithoutPrepare);
+        }
+
+        let jefe = task_jefe_api::Jefe::from(crate::JEFE.get_task_id());
+        jefe.request_reset();
```
kc8apf, line 245: Not thrilled that anyone who can deliver a UDP packet to an SP can cause the SP to reset.  If nothing else, add a TODO for "insert auth checks here"

task/mgmt-gateway/src/mgs_handler.rs
jgallagher: Will change; I'm trying to remember why I added the `sys_` prefix in the first place. This is indirectly triggering a system call (made by jefe); maybe that was it.

task/mgmt-gateway/src/mgs_handler.rs
jgallagher: https://github.com/oxidecomputer/hubris/issues/721

task/mgmt-gateway/src/mgs_handler.rs
jgallagher: This requires a corresponding change in omicron, where this trait is defined. I'm going to defer changing this name on this PR, but will include it in the next PR that bumps the omicron dep.

task/mgmt-gateway/src/mgs_handler.rs
jgallagher: Nice catch! Yes, we should, although as @kc8apf pointed out, failure mid update is currently a bit precarious. I'll update this to locally leave `UpdateBuffer` in a more consistent state so at least it can resume without panicking if given reasonable inputs.

task/mgmt-gateway/src/mgs_handler.rs
jgallagher: It could be owned by the `UpdateBuffer`, but it's also used directly by `MgsHandler` (in `update_start()`), so I think the options are:

1. Both own a copy
2. `UpdateBuffer` owns it, and `MgsHandler` reaches it via `self.update_buffer.update...` when it needs it
3. `MgsHandler` owns it and passes it to `UpdateBuffer` in this method

I lightly prefer option 3, but if you strongly prefer one of the other two I wouldn't argue.

mkeeter: Ah, I missed that it's also used by `MgsHandler`.  I don't feel strongly about it, so the current implementation is fine.

PR #708, opened 2022-08-12
README: Fix humility tasks example

mchodzikiewicz:
Just a tiny thing, but since it is first command documented after flashing it takes 5 minutes too long to figure out why it's not working :smile:


steveklabnik (approved):
Thank you!

PR #709, opened 2022-08-12
New `net` APIs

mkeeter:
This PR implements a set of new APIs for the `net` task, which allow us to
- Read the KSZ8463 MAC table
- Read registers on the KSZ8463
- Get the system's MAC address (from which the IP address can be inferred)
- Get the management network link status (when applicable)
- Read management network counters status (when applicable)

This is all infrastructure for the upcoming `humility net` subcommand.

### DIFF
--- a/drv/ksz8463/src/lib.rs
+++ b/drv/ksz8463/src/lib.rs
@@ -227,6 +227,13 @@ impl Ksz8463 {
         assert!(addr < 1024);
         self.write(Register::IACR, 0x1800 | addr)?;
         // Wait for the "not ready" bit to be cleared
+        //
+        // The IADR* registers together form a 72-bit value, which is packed
+        // into a set of u16 values; we use variables of the form `d_HI_LO`,
+        // where `HI` and `LO` are bit ranges in that value.
+        //
+        // Each `d_*` variable uses all 16 bits *except* `d_71_64`, which only
+        // uses the lower 8 bits.
         let d_71_64 = loop {
             let d = self.read(Register::IADR1)?;
             // Check bit 71 to see if the register is ready
--- a/drv/ksz8463/src/lib.rs
+++ b/drv/ksz8463/src/lib.rs
@@ -84,21 +84,21 @@ pub enum SourcePort {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub struct KszMacTableEntry {
+pub struct KszRawMacTableEntry {
     /// Number of valid entries in the table
-    count: u32,
+    pub count: u32,
 
     /// Two-bit counter for internal aging
     timestamp: u8,
 
     /// Source port where the FID + MAC is learned
-    source: SourcePort,
+    pub source: SourcePort,
 
     /// Filter ID
     fid: u8,
 
     /// MAC address from the table
-    addr: [u8; 6],
+    pub addr: [u8; 6],
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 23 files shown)


drv/ksz8463/src/lib.rs
```diff
@@ -222,13 +223,14 @@ impl Ksz8463 {
...
         addr: u16,
-    ) -> Result<Option<KszMacTableEntry>, Error> {
+    ) -> Result<Option<KszRawMacTableEntry>, Error> {
         assert!(addr < 1024);
         self.write(Register::IACR, 0x1800 | addr)?;
         // Wait for the "not ready" bit to be cleared
         let d_71_64 = loop {
             let d = self.read(Register::IADR1)?;
-            if d & (1 << 15) == 0 {
+            // Check bit 71 to see if the register is ready
```
jgallagher, line 232: Typo?
```suggestion
            // Check bit 7 to see if the register is ready
```

drv/vsc85xx/src/vsc85x2.rs
```diff
@@ -122,11 +129,26 @@ impl<'a, P: PhyRw> Vsc85x2Phy<'a, P> {
...
+    fn select_media_counters(&mut self) -> Result<(), VscError> {
+        // Configure the PHY to read fiber media SerDes counters
+        if self.phy_type == Vsc85x2Type::Vsc8562 {
+            self.phy.modify(
+                phy::EXTENDED_3::MEDIA_SERDES_TX_CRC_ERROR_COUNTER(),
+                |r| r.set_tx_select(0),
+            )?;
+            self.phy.modify(
+                phy::EXTENDED_3::MEDIA_MAC_SERDES_RX_CRC_CRC_ERR_COUNTER(),
+                |r| r.0 &= !(0b11 << 14),
```
jgallagher, line 141: I assume there's no convenient/correct helper method for this like there is for `set_tx_select()` above?

task/net-api/src/lib.rs
```diff
@@ -49,6 +50,116 @@ pub enum PhyError {
...
+#[repr(C)]
+pub struct ManagementCounters {
+    pub vsc85x2_tx: [ManagementCountersVsc85x2; 2],
+    pub vsc85x2_rx: [ManagementCountersVsc85x2; 2],
+
+    pub ksz8463_tx: [ManagementCountersKsz8463; 3],
+    pub ksz8463_rx: [ManagementCountersKsz8463; 3],
+
+    /// The MAC counters are only valid on the VSC8562
+    pub vsc85x2_mac_valid: bool,
```
jgallagher, line 150: I don't see this being set anywhere; is it always false?

jgallagher (approved):
I'm assuming all the low level register fiddling is correct and just reviewed for general comments; left a few nitpicks but LGTM. :+1:

task/net-api/src/lib.rs
mkeeter: Good catch, I forgot to set it!

(That's the downside of using `Default` then filling in fields later, although I'm not sure if there's a better way to initialize this struct...)

drv/vsc85xx/src/vsc85x2.rs
mkeeter: Unfortunately, no – these registers are all automatically extracted from the SDK, which only defines _some_ of the bits.

drv/ksz8463/src/lib.rs
mkeeter: This is actually intentional: the various `IADR*` registers form a 72-bit value, and everything is specified in terms of bits in that value.  I added a comment to clarify!

PR #710, opened 2022-08-15
WIP: Generic task restriction support

jgallagher:
This is a draft of support for generic support to restrict idol calls to specific tasks. The relevant change to idolatry is at https://github.com/jgallagher/idolatry/commit/2de1d7127998201a4bc741f254b7a2ec017ac64f, but I'm opening the PR here because I think it's more relevant. The specific changes are:

1. `idol::server::build_server_support()` now takes an extra arg of type `BTreeMap<String, Vec<String>>` which maps operation names to lists of tasks allowed to call them. If an operation name isn't present in the map (e.g., if the map is empty), there are no restrictions. Having to pass an empty map on every server stub is kind of annoying. Maybe this should be a separate function in idolatry, so callers that don't want to restrict ops don't need the extra arg?
2. For `jefe.get_state` (the only operation on master that currently has task restrictions, I think?), the explicit checks are removed in favor of adding the map for 1 to `jefe.config` in `app.toml`. The generated code looks like this:

```rust
    fn handle(
        &mut self,
        op: JefeOperation,
        incoming: &[u8],
        rm: &userlib::RecvMessage,
    ) -> Result<(), idol_runtime::RequestError<u16>> {
        // ... snip ...
        match op {
            // ... snip ...
            JefeOperation::request_reset => {
                if ![::hubris_num_tasks::Task::hiffy as usize,::hubris_num_tasks::Task::mgmt_gateway as usize,].contains(&rm.sender.index()) {
                    return Err(idol_runtime::RequestError::Fail(idol_runtime::ClientError::AccessViolation));
                }
```

which requires crates using this to depend on `hubris_num_tasks` with the `task-enum` feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1780,7 +1780,6 @@ checksum = "b9e0384b61958566e926dc50660321d12159025e767c18e043daf26b70104c39"
 [[package]]
 name = "idol"
 version = "0.2.0"
-source = "git+https://github.com/oxidecomputer/idolatry.git#285d92876b95f3f1c706e4a98aaa954d167b9de4"
 dependencies = [
  "indexmap",
  "quote",
@@ -1792,7 +1791,6 @@ dependencies = [
 [[package]]
 name = "idol-runtime"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/idolatry.git#285d92876b95f3f1c706e4a98aaa954d167b9de4"
 dependencies = [
  "userlib",
  "zerocopy",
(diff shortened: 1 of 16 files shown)


cbiffle:
I like the user-side ergonomics of this change a lot. We've talked about doing kernel-side MAC filtering, and we may still want to in the future because it'd avoid duplicating code in every server, but the kernel-side mechanisms we've discussed are closer to the MINIX facility of restricting which task can talk to which task, _not_ operation-level permissions. (That's not necessarily because operation level permissions in the kernel wouldn't work, we just haven't really discussed it.)

To reduce churn and allow an idolatry update without a careful dance across two repos, I'd suggest adding a new operation to the build library, something like `build_restricted_server_support`, that takes the new argument -- and then redefine the existing `build_server_support` to call that routine but pass `None`. That way we can make the change in idol whenever and then bump Hubris users incrementally.

jgallagher:
Superseded by #732

PR #712, opened 2022-08-17
xtask: Add dice configuration options to 'Signing' table.

flihp:
Names of variables / parameters have been kept as similar to those from
the spec as possible. These names are pretty awkward but we preserve
them as best we can for consistency.

NOTE: This PR depends on API changes in the lpc55_support repo. There's an open PR with the required changes in lpc55_support that must be merged first: https://github.com/oxidecomputer/lpc55_support/pull/19

I've marked this PR as an RFC since it configures xtask to use a topic branch from my fork of the lpc55_support repo with the API changes required to set these bits. It should not be merged until the associated changes in lpc55_support have been reviewed & merged.

### DIFF
--- a/app/lpc55xpresso/stage0.toml
+++ b/app/lpc55xpresso/stage0.toml
@@ -22,3 +22,7 @@ start = true
 [signing]
 priv-key = "../../support/fake_certs/fake_private_key.pem"
 root-cert = "../../support/fake_certs/fake_certificate.der.crt"
+enable-dice = true
+dice-inc-nxp-cfg = false
+dice-cust-cfg = false
+dice-inc-sec-epoch = false
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1902,7 +1902,7 @@ dependencies = [
 [[package]]
 name = "lpc55_sign"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/lpc55_support?rev=e1890a705e4fba8c4223af9fe9edad8be9d3ed98#e1890a705e4fba8c4223af9fe9edad8be9d3ed98"
+source = "git+https://github.com/oxidecomputer/lpc55_support?rev=edbc00749b708d5bc48c5fe0f1df0d1ed6f054c2#edbc00749b708d5bc48c5fe0f1df0d1ed6f054c2"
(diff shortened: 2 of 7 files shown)


app/lpc55xpresso/stage0.toml
```diff
@@ -22,3 +22,7 @@ start = true
 [signing]
 priv-key = "../../support/fake_certs/fake_private_key.pem"
 root-cert = "../../support/fake_certs/fake_certificate.der.crt"
+enable-dice = true
+dice-inc-nxp-cfg = false
+dice-cust-cfg = false
+dice-inc-sec-epoch = false
```
labbott, line 28: Does it make sense to put this on other targets? e.g. gemini-bu-rot. Those targets won't have TrustZone but secureboot and DICE should still work.

labbott (commented):
LGTM, thanks for doing this!

Is it worth documenting somewhere that if you flash the CMPA you need to have setup the CFPA and UDS?

app/lpc55xpresso/stage0.toml
flihp: Good point. I'll take a pass over the app configs with board set to 'lpcxpresso55s*' and make similar changes.

flihp:
> Is it worth documenting somewhere that if you flash the CMPA you need to have setup the CFPA and UDS?

I wouldn't want to do more than add short comment in the app.toml near these new keys. More detailed docs should probably go into the README in the lpc55_support crate if they're needed.

app/lpc55xpresso/stage0.toml
flihp: fixed by last push

flihp:
Just pushed the requested comment. To keep from duplicating this across the 3 app.tomls I've opted to put it in `build/xtask/src/dist.rs` at the call to `lpc55_support::create_cmpa`. LMK if you were asking for something else.

flihp:
The associated commit has been merged into the lpc55_support crate & this PR updated to use the appropriate git commit hash to get the new API. The title of the PR has been updated removing 'RFC' as it's now ready to merge.

kc8apf (approved):

PR #713, opened 2022-08-17
Improve RoT -> SP SWD performance

labbott:
Drastically improved from ~10 minutes to closer to 30 seconds

Ticks before:
```
    0   68        1        1 Start(0xc)
   1  130        1        1 End(0x9cb18)
```

Ticks after
```
    0   68        1        1 Start(0x3)
   1  130        1        1 End(0x54ae)
```

### DIFF
--- a/app/gemini-bu-rot/app.toml
+++ b/app/gemini-bu-rot/app.toml
@@ -99,7 +99,7 @@ pins = [
 name = "drv-lpc55-swd"
 priority = 4
 max-sizes = {flash = 16384, ram = 4096}
-uses = ["flexcomm3"]
+uses = ["flexcomm3", "iocon"]
 start = true
 stacksize = 1000
 task-slots = ["gpio_driver", "syscon_driver"]
--- a/drv/lpc55-swd/build.rs
+++ b/drv/lpc55-swd/build.rs
@@ -29,22 +29,49 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
 
     // The RoT -> SP SWD control requires setting the IO functions at runtime
     // as opposed to just startup.
+    //
+    // It turns out to be very expensive to call into the GPIO task for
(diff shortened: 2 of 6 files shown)


app/gemini-bu-rot/app.toml
```diff
@@ -99,7 +99,7 @@ pins = [
 name = "drv-lpc55-swd"
 priority = 4
 max-sizes = {flash = 16384, ram = 4096}
-uses = ["flexcomm3"]
+uses = ["flexcomm3", "iocon"]
```
mkeeter, line 102: This means that both `swd` and `gpio_driver` can use `iocon` – any concerns about that sharing?  It seems okay, since the only thing they do is write to an address derived from `device::IOCON::ptr()`

drv/lpc55-swd/build.rs
```diff
@@ -29,22 +29,67 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
...
+    // changing pin direction so these functions access the IOCON block
+    // directly.
     writeln!(
         &mut file,
         "{}",
         quote::quote! {
 
+        // This does the conversion of the pin configuration in a way
+        // that's compatible with the existing iocon APIs
+        fn to_conf(
```
mkeeter, line 43: Is there a convenient way to share code with `lpc55-gpio-api`, which has the same logic in `iocon_configure`?

drv/lpc55-swd/src/main.rs
```diff
@@ -53,6 +53,7 @@
 #![no_std]
 #![no_main]
 
+use drv_lpc55_gpio_api::*;
```
mkeeter, line 56: Is this import used?  I don't see any additions that would require it.

mkeeter (approved):
A few comments, but generally LGTM

PR #714, opened 2022-08-18
Add "reason for most recent reset"

jgallagher:
The stm32h7 branch is fully implemented. stm32g0 has primitive support (the raw reset register is reported, but it is not mapped to one of the logical `ResetReason`s).

`set_reset_reason` is (optionally) restricted to a single task using the same style as `Jefe.set_state()`, until something like #710 is available.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1352,6 +1352,7 @@ dependencies = [
 name = "drv-stm32xx-sys"
 version = "0.1.0"
 dependencies = [
+ "bitflags",
  "cfg-if",
  "drv-stm32xx-gpio-common",
  "drv-stm32xx-sys-api",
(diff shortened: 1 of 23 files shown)


cbiffle (commented):
We talked about the reset type exposed in the jefe API in chat -- I think it wants to be bitflags so we can faithfully describe combinations of circumstances. Otherwise this looks great.

labbott (commented):
LGTM, This will also be easy to add for the LPC55 (will need a few more enum items)

jgallagher:
> We talked about the reset type exposed in the jefe API in chat -- I think it wants to be bitflags so we can faithfully describe combinations of circumstances. Otherwise this looks great.

We talked about this more offline and decided this is okay as-is; each concrete `ResetReason` variant can be mapped to a specific set of chip-specific bits by looking at the code, and any chip-specific bits that don't map to a specific `ResetReason` can be represented verbatim as a `ResetReason::Other(raw_bits)`.

PR #715, opened 2022-08-18
UDP RPC task

mkeeter:
This PR adds a relatively minimal `task-udprpc`, which accepts specially-constructed UDP packets and executes them as Hiffy calls.

It still needs some clean-up, so I'm marking it as a draft for now.

The Humility side is implemented in [humility#216](https://github.com/oxidecomputer/humility/pull/216)

### DIFF
--- a/task/udprpc/src/main.rs
+++ b/task/udprpc/src/main.rs
@@ -86,7 +86,7 @@ fn main() -> ! {
                 let r = if (meta.size as usize) < HEADER_SIZE {
                     RpcReply::TooShort
                 } else if image_id != header.image_id {
-                    tx_data_buf[1..9].copy_from_slice(&image_id.as_bytes());
+                    tx_data_buf[1..9].copy_from_slice(image_id.as_bytes());
                     RpcReply::BadImageId
                 } else if meta.size as usize
                     != HEADER_SIZE + header.nbytes as usize
--- a/task/udprpc/src/main.rs
+++ b/task/udprpc/src/main.rs
@@ -1,0 +1,64 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#![no_std]
+#![no_main]
+
+use task_net_api::*;
+use userlib::*;
+
+task_slot!(NET, net);
+
+#[export_name = "main"]
+fn main() -> ! {
+    let net = NET.get_task_id();
+    let net = Net::from(net);
+
+    const SOCKET: SocketName = SocketName::rpc;
+
+    loop {
+        // Tiiiiiny payload buffer
+        let mut rx_data_buf = [0u8; 256];
+        let mut tx_data_buf = [0u8; 256];
+        match net.recv_packet(
+            SOCKET,
+            LargePayloadBehavior::Discard,
+            &mut rx_data_buf,
+        ) {
+            Ok(mut meta) => {
+                let task =
+                    u16::from_be_bytes(rx_data_buf[0..2].try_into().unwrap());
+                let op =
+                    u16::from_be_bytes(rx_data_buf[2..4].try_into().unwrap());
+                let nreply =
+                    u16::from_be_bytes(rx_data_buf[4..6].try_into().unwrap());
+                let nbytes =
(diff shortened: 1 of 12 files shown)


kc8apf (approved):

task/udprpc/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+                // in the conditional below).
+                const HEADER_SIZE: usize = core::mem::size_of::<RpcHeader>();
+                let header =
+                    RpcHeader::read_from(&rx_data_buf[..HEADER_SIZE]).unwrap();
+
+                const REPLY_PREFIX_SIZE: usize = 5;
+
+                // We deliberately assign to `r` here then manipulate it;
+                // otherwise, the compiler won't include RpcReply in DWARF data.
+                let r = if (meta.size as usize) < HEADER_SIZE {
```
jgallagher, line 86: I was going to leave a comment about not checking `meta.size` before constructing `header`, but I see it here; it's slightly strange to me for it to be in this order. Maybe throw a comment on the `read_from` call above that we'll check the size below (before reading any fields of the potentially-invalid `header`)?

task/udprpc/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+
+                tx_data_buf[0] = r as u8;
+                meta.size = match r {
+                    RpcReply::TooShort
+                    | RpcReply::NBytesMismatch
+                    | RpcReply::NReplyOverflow
+                    | RpcReply::NBytesOverflow
+                    | RpcReply::NReplyMismatch => 1,
+                    RpcReply::BadImageId => 9,
+                    RpcReply::Ok => header.nreply as u32 + 5,
```
jgallagher, line 129: Nit: maybe 
```suggestion
                    RpcReply::Ok => header.nreply as u32 + REPLY_PREFIX_SIZE,
```
?

task/udprpc/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+                    RpcReply::BadImageId
+                } else if meta.size as usize
+                    != HEADER_SIZE + header.nbytes as usize
+                {
+                    RpcReply::NBytesMismatch
+                } else if header.nbytes as usize + HEADER_SIZE
+                    > rx_data_buf.len()
+                {
+                    RpcReply::NBytesOverflow
+                } else if header.nreply as usize + 4 > tx_data_buf.len() {
```
jgallagher, line 99: Should this be

```suggestion
                } else if header.nreply as usize + REPLY_PREFIX_SIZE > tx_data_buf.len() {
```

(i.e., 5 instead of 4)?

jgallagher (approved):

task/udprpc/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+    NReplyOverflow,
+}
+
+/// Header for an RPC request
+///
+/// `humility` must cooperate with this layout, which is mirrored in `doppel.rs`
+#[derive(Copy, Clone, Debug, FromBytes)]
+#[repr(C)]
+struct RpcHeader {
+    image_id: u64,
```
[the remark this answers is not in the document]
mkeeter: Probably wise, done (and also done on the Humility side)

task/udprpc/src/main.rs
mkeeter: 👍🏻

task/udprpc/src/main.rs
mkeeter: I think so, good catch!

task/udprpc/src/main.rs
mkeeter: I ended up rearranging this a little bit to make it less weird!

task/udprpc/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+                    tx_data_buf[1..9].copy_from_slice(image_id.as_bytes());
+                    RpcReply::BadImageId
+                } else if meta.size as usize
+                    != HEADER_SIZE + header.nbytes as usize
+                {
+                    RpcReply::NBytesMismatch
+                } else if header.nbytes as usize + HEADER_SIZE
+                    > rx_data_buf.len()
+                {
+                    RpcReply::NBytesOverflow
```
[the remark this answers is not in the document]
mkeeter: I think you're right!

PR #716, opened 2022-08-18
Lpc55 lib usart

flihp:
This PR moves much of the lpc55-usart code into a library and adds a feature to support usart usage in hubris as well as stage0. The drv-lpc55-usart driver has been updated to use this library, and future work on the DICE certificate provisioning protocol will as well.


labbott:
I think a better example of a split between non-Hubris library and Hubris is the SPI setup. `drv/lpc55-spi` and `drv/stm32h7-spi` have all the non-Hubris specific parts and then the various servers implement the SPI `drv/lpc55-spi-server` (which will eventually have more things happening), `drv/lpc55-swd`, `drv/stm32h7-spi-server`. 

I'm also kind of leaning towards deleting the LPC55 UART driver for Hubris anyway. It was written during more of a bringup phase when we were learning what the LPC55 and Hubris could do. If we wanted to keep it around, it might be nice to re-write it to use more modern idol APIs.

flihp:
Thanks for the data here. I wouldn't blame my missteps here on the lpc55-usart driver. I had modeled my implementation after the stm32h7-usart library: https://github.com/oxidecomputer/hubris/blob/a04ad15af4f135bfdf18a86603a78f96edd2905c/drv/stm32h7-usart/src/lib.rs#L39 where the syscon interaction happens inside the library. I'll take a pass back over this after looking at the examples you've provided.

labbott (approved):
LGTM

PR #717, opened 2022-08-19
stm32h7: allow optional mapping into AXI SRAM

cbiffle:
By default, apps on stm32h7 are mapped into a 128 kiB RAM on the
processor's local bus. This is great because it's zero-wait-state.

However, there are a lot more RAMs than that. This adds an alternate
memory file for mapping all of an application into the 512 kiB AXI SRAM,
which has more wait states (though not many more) but is 4x larger.

This is in lieu of a smarter memory allocator that can distribute across
RAMs based on their properties, which we'll want to do eventually.

### DIFF
--- a/app/gimlet/rev-b.toml
+++ b/app/gimlet/rev-b.toml
@@ -2,6 +2,7 @@ name = "gimlet-b"
 target = "thumbv7em-none-eabihf"
 board = "gimlet-b"
 chip = "../../chips/stm32h7"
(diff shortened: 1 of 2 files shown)


mkeeter (approved):
Looks good, but we should switch Sidecar to `memory-large.toml` as well

mkeeter:
(I'm going to switch Sidecar's memory bank in  #715 , since there are a few other changes that go hand-in-hand)

PR #719, opened 2022-08-19
xtask: Pin to Cargo.toml to current HEAD of lpc55_support repo.

flihp:
We have some API changes to the lpc55_support repo pending that require
coordinated changes in xtask. To keep from breaking hubris during the
merge this commit pins xtask to the current HEAD. After the API changes
are merged into lpc55_support, another PR will go into hubris to use the
new API & update the pinned git commit accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,15 +29,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anyhow"
 version = "1.0.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


flihp:
related PR in lpc55_support: https://github.com/oxidecomputer/lpc55_support/pull/19

labbott (approved):
There are some other changes in `Cargo.lock` I don't know if we care about those changing but otherwise it LGTM

flihp:
I found the changes to Cargo.lock confusing / troubling as well. After rebasing on current hubris/master the number of changes in Cargo.lock have gone down but not away completely. I'm assuming that since this PR changes a Cargo.toml that some churn in the lock file is inevitable but this is more than I would expect.

jgallagher:
At a glance the `Cargo.lock` changes look reasonable to me, because of this change that's related to the PR:

```diff
 [[package]]
 name = "lpc55_sign"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/lpc55_support#98c44ba761f5a487c456e288d95859e7f6058d65"
+source = "git+https://github.com/oxidecomputer/lpc55_support?rev=e1890a705e4fba8c4223af9fe9edad8be9d3ed98#e1890a705e4fba8c4223af9fe9edad8be9d3ed98"
 dependencies = [
  "anyhow",
  "byteorder",
+ "clap",
  "crc-any",
  "ecdsa",
  "elliptic-curve",
  "hex",
  "p256",
  "packed_struct",
  "packed_struct_codegen",
  "rsa",
  "sha2",
- "structopt",
  "x509-parser",
]
```

`lpc55_sign` looks like it was the last crate in the dependency tree that was pulling in structopt, which itself was the last crate in the dependency tree that was pulling in clap 2.x. Pulling both of those out also pulled out all the deps they used that are no longer used by clap 3.x.

PR #724, opened 2022-08-22
Fill gaps in binaries with `0xff`

labbott:
By default, objcopy will fill gaps in sections with `0x00`. The flash
programming sequence on the STM32 will not program the gaps and will
leave them in the default state of `0xff`. This results in a mismatch
between the generated binary and flash. From a running program
perspective this is fine but it results in incorrect calculations of
expected values in flash. Fix this by filling the gap with `0xff`.


labbott:
From the objcopy man page:
```
       --gap-fill val
           Fill gaps between sections with val.  This operation applies to the
           load address (LMA) of the sections.  It is done by increasing the
           size of the section with the lower address, and filling in the
           extra space created with val.
```

flihp:
LGTM

PR #727, opened 2022-08-23
Sample task for measuring SP flash

labbott:


task/sp_measure/build.rs
```diff
@@ -0,0 +1,50 @@
...
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let out_dir = std::env::var("OUT_DIR")?;
+    let dest_path = std::path::Path::new(&out_dir).join("expected.rs");
+    let mut file = std::fs::File::create(&dest_path)?;
+
+    let task_config = build_util::task_config::<TaskConfig>()?;
+
+    let bin = match std::fs::read(&task_config.binary_path) {
+        Ok(b) => b,
+        Err(_) => vec![0; 256],
```
mkeeter, line 26: I'd prefer to error out here if the target `.bin` file doesn't exist, to avoid weird confusing situations – is there a reason we want to continue with an invalid checksum?

Also, should we be rerunning if the binary path changes, e.g.
```
    println!("cargo:rerun-if-changed={}", task_config.binary_path);
```
?

task/sp_measure/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+        ringbuf_entry!(Trace::ErrCnt(err_cnt));
+        if sha_out.as_slice() == EXPECTED.as_slice() {
+            ringbuf_entry!(Trace::ShaGood);
+        } else {
+            ringbuf_entry!(Trace::ShaBad);
+            panic!();
+        }
+
+        // Wait for a notification that will never come, politer than
+        // busy looping forever
```
mkeeter, line 107: Neat!  Should we be doing this elsewhere?

task/sp_measure/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+                loop {
+                    match sp_ctrl
+                        .read_transaction_start(addr, addr + TRANSACTION_SIZE)
+                    {
+                        Err(_) => {
+                            err_cnt += 1;
+                            let _ = sp_ctrl.setup();
+                            continue;
+                        }
+                        _ => break,
```
mkeeter, line 60: Nit: I would find this a little clearer if the `Ok` case was first, and without the explicit `continue`, i.e.
```suggestion
                        Ok(_) => break,
                        Err(_) => {
                            err_cnt += 1;
                            let _ = sp_ctrl.setup();
                        }
```

task/sp_measure/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                }
+            }
+
+            data.fill(0);
+            loop {
```
mkeeter, line 66: The triple-nested `loop { match { ...`, plus the previous loop make this all a bit hard to read.

Logically, I think we want something like this?

```
fn read_inner(...) -> Result<...> {
    read_transaction_start(...)?;
    read_transaction(...)?
}
fn read_checked -> Result<...> {
    loop {
        match read(...) {
            Ok(e) => return Ok(e),
            Err(...) => while sp_ctrl.setup().is_error() {
                // keep retrying setup until it works
            }
        }
    }
}
```

mkeeter: Hmmm, maybe not, because we only call `read_transaction_start` message on multiples of `TRANSACTION_SIZE`?

app/gimlet-rot/app.toml
```diff
@@ -118,3 +118,13 @@ pins = [
 { pin =  { port = 0, pin = 7 }, alt = 3 },
 ]
 spi_num = 5
+
+[tasks.sp_measure]
+name = "task-sp-measure"
+priority = 5
+max-sizes = {flash = 32768, ram = 8192}
+task-slots = ["swd"]
+stacksize = 2048
```
mkeeter, line 127: Do we want `start = false` here?

app/gimlet-rot/app.toml
```diff
@@ -93,7 +93,7 @@ name = "drv-lpc55-swd"
 priority = 4
 max-sizes = {flash = 16384, ram = 4096}
 uses = ["flexcomm5", "iocon"]
-start = false
+start = true
```
mkeeter, line 96: Is this intentional?  Does it conflict with debuggers plugged into the header?

task/sp_measure/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    Ok(_) => break,
+                }
+            }
+
+            sha.update(&data);
```
mkeeter, line 91: At some point, should we use the hardware HASH peripheral, or does SWD overhead eat most of our time?

task/sp_measure/build.rs
labbott: The problem is that for cases like CI we may not have a binary that exists. I didn't have a better idea of how to merge this code. There's also no guarantee what's flashed on the SP matches anyway. I do think there's value in continuing with an invalid hash to demonstrate how the RoT -> SP SWD control works and for testing anything related to that.

You are correct, I forgot to add the rerun line!

task/sp_measure/src/main.rs
labbott: Maybe? I didn't find any other tasks that don't want to loop. I expect this will actually be folded into real measurements once we figure out how that works.

task/sp_measure/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+drv-sp-ctrl-api = {path = "../../drv/sp-ctrl-api"}
+sha2 = {version = "0.9", default-features = false}
+
+[build-dependencies]
+build-util = {path = "../../build/util"}
+idol = {git = "https://github.com/oxidecomputer/idolatry.git"}
+anyhow = "1.0.31"
+serde = "1"
+quote = "1"
+sha2 = {version = "0.9", default-features = false}
```
andrewjstone, line 18: We're currently using `sha3-256` in sprockets and in the early-boot dice stuff that @flihp is doing. We should probably use that here also.

app/gimlet-rot/app.toml
mkeeter: I don't actually know what the default is 😆

So yeah, having it explicit for documentation purposes would be nice!

task/sp_measure/src/main.rs
flihp: I'm about to go update this data in RFD32. Initially we had planned on using sha512 so that we could share code with the salty library. The hash function built into salty doesn't implement any of the RustCrypto hash traits so it's not usable in an hmac. I've since switched to using sha3-256 but I never went back to the RFD to fix this.

task/sp_measure/Cargo.toml
flihp: I've just updated RFD36 to reflect this. I was using sha512 there to do "worst-case" storage analysis but hadn't updated it to include the determination we made recently w/r/t sha3-256: https://github.com/oxidecomputer/rfd/commit/a19b0e7187048afeab44f05b2bb55154a5542b24

app/gimlet-rot/app.toml
labbott: I added a comment stating that we're not starting the task

PR #728, opened 2022-08-23
Add reset reason for RoT targets

labbott:


drv/lpc55-syscon/src/main.rs
```diff
@@ -224,6 +229,35 @@ fn main() -> ! {
     }
 }
 
+fn set_reset_reason() {
+    let pmc = unsafe { &*device::PMC::ptr() };
+
+    // The Reset Reason is stored in the AOREG1 register in the power
+    // management block. This crypticly named register is set based
+    // on another undocumented register in the power management space.
+    // Official documentation for these bits is is available in 13.4.13
```
mkeeter, line 238: Nit: I found `AOREG1` in section 13.4.6 of [v2.3 of the datasheet](https://drive.google.com/file/d/1C_htkgvxVV74X1zpwz5xKna_1r8fmxnV/view?usp=sharing), not 13.4.13.

mkeeter (approved):

drv/lpc55-syscon/src/main.rs
labbott: ah, that's a typo, it's actually v2.4 from Oct 8 2021. v2.3 is from June 2021. I'll fix that.

PR #729, opened 2022-08-23
Extend the update API

labbott:
- Prep now requires an image type to indicate which image is being
  updated.
- Add an `abort_update` function if the calling task decides they
  need to stop an update.
- Added a few more error types to better specify exactly what may have
  gone wrong.

### DIFF
--- a/drv/update-api/src/lib.rs
+++ b/drv/update-api/src/lib.rs
@@ -6,6 +6,21 @@
 
 use derive_idol_err::IdolError;
 use userlib::{sys_send, FromPrimitive};
+use zerocopy::AsBytes;
+
+#[repr(u8)]
+#[derive(FromPrimitive, AsBytes, PartialEq, Clone, Copy)]
+pub enum UpdateTarget {
+    // Represents targets where we only ever write to a single
+    // alternate flash location. This is typically used in
+    // conjunction with a bank swap feature.
+    Alternate = 1,
+    // Represents targets where we must write to a specific range
+    // of flash.
+    ImageA = 2,
+    ImageB = 3,
+    Bootloader = 4,
+}
 
 #[derive(FromPrimitive, IdolError)]
 #[repr(u32)]
@@ -24,6 +39,9 @@ pub enum UpdateError {
     StrobeErr = 11,
     ProgSeqErr = 12,
     WriteProtErr = 13,
+    BadImageType = 14,
+    UpdateAlreadyFinished = 15,
+    UpdateNotStarted = 16,
 }
 
 pub mod stm32h7 {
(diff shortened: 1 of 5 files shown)


drv/update-api/src/lib.rs
```diff
@@ -6,6 +6,17 @@
 
 use derive_idol_err::IdolError;
 use userlib::{sys_send, FromPrimitive};
+use zerocopy::AsBytes;
+
+#[repr(u8)]
+#[derive(FromPrimitive, AsBytes, PartialEq, Clone, Copy)]
+pub enum ImageType {
+    // Represents targets where only a single image exists
```
jgallagher, line 14: This seems a bit misleading in the context of the SP, maybe, since it has two banks? What are the uses for the other cases going to be?

jgallagher (approved):
LGTM, thanks! Just a question on the docs/plans for `ImageType`.

drv/update-api/src/lib.rs
labbott: The SP has two banks of flash but from Hubris' perspective we're only ever writing to one bank. The RoT needs to write explicitly to different locations. I thought that just having the SP say "always use ImageA`" or the like wouldn't be as clear as a distinct image type. I'm open to suggestions for naming or other approaches to better represent the system.

jgallagher: I see. Would it be clearer to name this something like `FlashBank`? I'm not sure this is better, but something like

```rust
pub enum FlashBank {
    // Systems that only allow writing to the alternate bank (e.g., stm32h7)
    Alternate,
    // Systems that allow writing to explicit A/B/Bootloader banks
    A,
    B,
    Bootloader,
}
```

or possibly splitting that up, something like

```rust
pub enum FlashBank {
    Alternate,
    Explicit(BankName),
}

pub enum BankName {
    A,
    B,
    Bootloader,
}
```

but that makes the encoding/decoding more complicated.

rmustacc: One thing that I tripped over when I first read this is I thought it was trying to describe the target of the image (e.g. even if it's the same image in the SP, whether it was going to the A/B side) rather than the way that the thing was built.

labbott: I was trying to avoid Bank since we don't have true banked flash on the RoT. I do like the idea of a separate encoding though so I'll give this more thought and also add some more comments.

labbott: I ended up going with `UpdateTarget` for the name of the enum and added a few more comments.

jgallagher: LGTM :+1:

Issue #726, opened 2022-08-23
humility qspi failed: request failed: Some(Fault(ReturnStackOverflow))


rmustacc:
While trying to use humility to write to the qspi using -W, we end up failing with a return stack overflow:

```
rm@cadbury ~ $ pfexec /gimlet/bin/humility-0.8.8 -t sn9 qspi -W ./milan-gimlet-b.img 
humility: attached to 0483:374e:002F001A4741500520383733 via ST-Link V3
humility: erasing 33554432 bytes...
humility qspi failed: request failed: Some(Fault(ReturnStackOverflow))
rm@cadbury ~ $ pfexec /gimlet/bin/humility-0.8.8 -t sn9 ringbuf
humility: attached to 0483:374e:002F001A4741500520383733 via ST-Link V3
humility: ring buffer drv_gimlet_seq_server::__RINGBUF in gimlet_seq:
 NDX LINE      GEN    COUNT PAYLOAD
  56  437    62383        1 GetState
  57  364    62383        1 Status(0x0, 0x8, 0xff)
  58  437    62383        1 GetState
  59  364    62383        9 Status(0x0, 0x8, 0xff)
  60  437    62383        1 GetState
  61  364    62383        1 Status(0x0, 0x8, 0xff)
  62  437    62383        1 GetState
  63  364    62383        9 Status(0x0, 0x8, 0xff)
   0  437    62384        1 GetState
   1  364    62384        1 Status(0x0, 0x8, 0xff)
   2  437    62384        1 GetState
   3  364    62384        9 Status(0x0, 0x8, 0xff)
   4  437    62384        1 GetState
   5  364    62384        2 Status(0x0, 0x8, 0xff)
   6  437    62384        1 GetState
   7  364    62384        8 Status(0x0, 0x8, 0xff)
   8  437    62384        1 GetState
   9  364    62384        2 Status(0x0, 0x8, 0xff)
  10  437    62384        1 GetState
  11  364    62384        8 Status(0x0, 0x8, 0xff)
  12  437    62384        1 GetState
  13  364    62384        2 Status(0x0, 0x8, 0xff)
  14  437    62384        1 GetState
  15  364    62384        8 Status(0x0, 0x8, 0xff)
  16  437    62384        1 GetState
  17  364    62384        2 Status(0x0, 0x8, 0xff)
  18  437    62384        1 GetState
  19  364    62384        8 Status(0x0, 0x8, 0xff)
  20  437    62384        1 GetState
[...]

PR #730, opened 2022-08-24
host flash operations should fail explicitly when in wrong mux state

bcantrill:
This addresses the Hubris side of https://github.com/oxidecomputer/hubris/issues/726


cbiffle (approved):

bcantrill:
> Do we also want this check in `set_mux`? Swapping the selected flash IC while the host owns it seems like it could create some confusion.

Do you mean `set_dev`? If, so, yes, I think that would make a ton of sense -- will do...

mkeeter (approved):

Issue #699, opened 2022-08-09
System reset: Restrict which tasks are allowed to request a reset


jgallagher:
This should build on #684, which adds reset as a kipc and a corresponding idol call to jefe. We should be able to configure which task (possibly "tasks" plural?) are allowed to send a `reset_request` to jefe in app.toml.

jgallagher:
Implemented in #732

PR #732, opened 2022-08-25
Restrict operation callers in app.toml

jgallagher:
This allows us to remove the by-hand check for `state-owner` and `reset-reason-owner` in jefe's config and server implementation, in favor of passing the list of allowed callers to idolatry and letting it put the check in the generated code (via https://github.com/oxidecomputer/idolatry/pull/22).

While I was in here, I added restrictions on `request-reset` (to `hiffy`/`mgmt-gateway`), fixing #699.

Callers that attempt to perform operations that they aren't allowed to will now see `AccessViolation` errors instead of `UnknownOperation`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1809,7 +1809,7 @@ checksum = "b9e0384b61958566e926dc50660321d12159025e767c18e043daf26b70104c39"
 [[package]]
 name = "idol"
 version = "0.2.0"
-source = "git+https://github.com/oxidecomputer/idolatry.git#285d92876b95f3f1c706e4a98aaa954d167b9de4"
+source = "git+https://github.com/oxidecomputer/idolatry.git#cf366c0f8d3efc57eec6eb764eb5fc04bbf1a9fe"
 dependencies = [
  "indexmap",
  "quote",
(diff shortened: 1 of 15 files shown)


jgallagher:
I should note that attempting to call an operation from hiffy that isn't allowed to be called by hiffy doesn't have the best failure mode:

```
% humility hiffy -c Jefe.set_state -a state=1
humility hiffy failed: operation timed out
```

because the `hiffy` task faults:

```
% humility tasks
ID TASK                 GEN PRI STATE
...
 7 hiffy                  4   4 FAULT: reply fault: task id #0/gen0, reason AccessViolation (was: wait: reply from jefe/gen0)
```

kc8apf (approved):

mkeeter (approved):
Looks good!  Can you open a Humility ticket to read `allowed-callers` and error out immediately instead of letting `task-hiffy` get killed?

jgallagher:
> Can you open a Humility ticket to read `allowed-callers` and error out immediately instead of letting `task-hiffy` get killed?

I can, but to clarify: read it from `app.toml`? That might be a bit fragile; jefe happens to use that key, but a build script could construct the allowed callers using some other mechanism.

jgallagher:
https://github.com/oxidecomputer/humility/issues/222

PR #733, opened 2022-08-25
mgmt-gateway: Update to new MGS serialization of large messages

jgallagher:
Update payloads and serial console packets are now serialized as raw
data _after_ a hubpack-encoded message, instead of the data being
embedded as part of the hubpack message. Formerly we were required to
allocate several moderately-sized arrays on the stack and copy data
unnecessarily; now the raw data is borrowed from the receive buffer we
pass to `net`, which requires no extra copies or stack allocation of
large arrays.

This is the hubris half of https://github.com/oxidecomputer/omicron/pull/1645, which did the MGS side of this work.

We get significant improvements in flash and stack, and a modest improvement in ram (although it doesn't look like it at first glance). Before this PR:

```
# xtask sizes
PROGRAM        REGION  USED   SIZE   LIMIT
mgmt_gateway   flash   26764  32768  32768
               ram     6444   8192   8192

# stackmargin after performing an update, before resetting
11 mgmt_gateway       0x20004000       4096       3416        680
```

After this PR:

```
# xtask sizes
PROGRAM        REGION  USED   SIZE   LIMIT
mgmt_gateway   flash   19220  32768  32768
               ram     7456   8192   8192
               
# stackmargin after performing an update, before resetting
11 mgmt_gateway       0x20004000       1536       1048        488
```

Our absolute ram usage goes up by about 1k, but our buffers are significantly larger: we can now receive ~1k update chunks (previously 512 bytes), and we can send and receive ~1k serial console packets (previously 128 bytes). Handwaving a bit, I think we have about 1.8k more buffer space for only 1k more ram usage.

### DIFF
--- a/task/mgmt-gateway/src/main.rs
+++ b/task/mgmt-gateway/src/main.rs
@@ -7,10 +7,10 @@
 
 use drv_stm32h7_usart::Usart;
 use gateway_messages::{
-    sp_impl, sp_impl::Error as MgsDispatchError, sp_impl::SocketAddrV6,
-    IgnitionCommand, Request, SerialConsole, SerializedSize, SpMessage,
-    SpMessageKind, SpPort,
+    sp_impl, sp_impl::Error as MgsDispatchError, IgnitionCommand, SpComponent,
+    SpMessage, SpMessageKind, SpPort,
 };
+use mgs_handler::UsartFlush;
 use mutable_statics::mutable_statics;
 use ringbuf::{ringbuf, ringbuf_entry};
 use task_net_api::{
@@ -26,6 +26,8 @@ use userlib::{
 mod mgs_handler;
 
 use self::mgs_handler::MgsHandler;
+
+type SerializedMessageBuf = [u8; gateway_messages::MAX_SERIALIZED_SIZE];
 
 task_slot!(JEFE, jefe);
 task_slot!(NET, net);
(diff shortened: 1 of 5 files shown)


task/mgmt-gateway/src/main.rs
```diff
@@ -120,17 +122,20 @@ fn main() {
...
+    to_tx: &'static mut ArrayVec<SerializedMessageBuf>,
+    from_rx: &'static mut ArrayVec<SerializedMessageBuf>,
     from_rx_flush_deadline: Option<u64>,
 }
 
 impl UsartHandler {
-    fn new(usart: Usart) -> Self {
+    fn new(
+        usart: Usart,
+        [to_tx, from_rx]: &'static mut [ArrayVec<SerializedMessageBuf>; 2],
```
mkeeter, line 133: I'm not a huge fan of destructuring in function arguments, but that's just, like, _my opinion_, man.

jgallagher: Hah, me neither. IIRC this is a relic of trying to get `cargo fmt` to do something reasonable with a long line before I introduced the `SerializedMessageBuf` typedef; it was doing something heinous like

```rust
buffers: &'static mut [ArrayVec<[u8; SOME_LONG_CONSTANT]>;
                       2],
```

With the typedef I don't think that's a concern any more, so I'll switch this back.

mkeeter (approved):

PR #737, opened 2022-08-30
mgmt-gateway: Refine serial console messages

jgallagher:
This is the hubris half to https://github.com/oxidecomputer/omicron/pull/1651 that lets us address a few lingering TODOs:

1. Attaching and detaching an MGS instance to a particular SP's proxied serial console are now explicit messages; previously attaching was implicit when MGS sent data, and detaching was implicit when a new MGS instance attached, neither of which was ideal.
2. MGS -> SP serial console data packets now include an offset, allowing the SP to detect and ignore any resent UDP packets. Previously if MGS resent a serial console data packet because it missed an ack, the SP had no way to know it had already seen it, and it would forward it along the uart a second time.
3. SP -> MGS serial console data packets now include an offset. We do not ack SP -> MGS messages, both because the logic to do that on the SP would be tricky and because it has very little recourse for unacked messages (it can't apply backpressure on a uart, and it can't buffer data arbitrarily long). As of this PR MGS can at least know and log when data has been lost due to a networking or SP buffering issue.

### DIFF
--- a/task/mgmt-gateway/src/main.rs
+++ b/task/mgmt-gateway/src/main.rs
@@ -212,7 +212,9 @@ impl UsartHandler {
 
         // Clean up / ringbuf debug log after transmitting.
         if n_transmitted > 0 {
-            ringbuf_entry!(Log::UsartTx { num_bytes: n_transmitted });
+            ringbuf_entry!(Log::UsartTx {
+                num_bytes: n_transmitted
+            });
             self.to_tx.drain_front(n_transmitted);
         }
         if self.to_tx.is_empty() {
@@ -264,7 +266,9 @@ impl UsartHandler {
         }
 
         if n_received > 0 {
-            ringbuf_entry!(Log::UsartRx { num_bytes: n_received });
+            ringbuf_entry!(Log::UsartRx {
+                num_bytes: n_received
+            });
             self.start_flush_timer_if_needed();
         }
     }
--- a/task/mgmt-gateway/src/main.rs
+++ b/task/mgmt-gateway/src/main.rs
@@ -63,9 +63,12 @@ enum MgsMessage {
         command: IgnitionCommand,
     },
     SpState,
+    SerialConsoleAttach,
     SerialConsoleWrite {
+        offset: u64,
         length: u16,
     },
+    SerialConsoleDetach,
     UpdateStart {
         length: u32,
     },
--- a/task/mgmt-gateway/src/main.rs
+++ b/task/mgmt-gateway/src/main.rs
@@ -128,6 +128,7 @@ struct UsartHandler {
     to_tx: &'static mut ArrayVec<SerializedMessageBuf>,
     from_rx: &'static mut ArrayVec<SerializedMessageBuf>,
     from_rx_flush_deadline: Option<u64>,
+    from_rx_offset: u64,
 }
 
 impl UsartHandler {
@@ -141,6 +142,7 @@ impl UsartHandler {
             to_tx,
             from_rx,
             from_rx_flush_deadline: None,
+            from_rx_offset: 0,
         }
     }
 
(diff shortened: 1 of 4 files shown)


task/mgmt-gateway/src/main.rs
```diff
@@ -200,13 +221,27 @@ impl UsartHandler {
...
         }
 
         // Recv as much as we have space for.
         let mut n = 0;
         let mut available_rx_space =
             self.from_rx.capacity() - self.from_rx.len();
 
+        // Ensure we always have space for at least 16 bytes (the stm32h7 usart
+        // FIFO depth), even if that means discarding old data.
+        if available_rx_space < 16 {
```
mkeeter, line 239: Can we log an error here, even though MGS will also notice it?

task/mgmt-gateway/src/main.rs
```diff
@@ -173,6 +178,22 @@ impl UsartHandler {
         sys_set_timer(None, TIMER_IRQ);
     }
 
+    fn drain_flushed_data(&mut self, n: usize) {
+        self.from_rx.drain(..n);
```
mkeeter, line 182: Under the hood, this is going to `ptr::copy` the remaining data to shift it over by `n` ([source](https://docs.rs/arrayvec/latest/src/arrayvec/arrayvec.rs.html#983)), so it could do a bunch of extra work!

It seems like the "correct" data structure here is a circular buffer, although that makes `serialize_with_trailing_data` a little trickier, because the trailing data could be split over two slices.

Do you think it's worth fixing this now, or leaving a TODO?

task/mgmt-gateway/src/main.rs
```diff
@@ -200,13 +221,27 @@ impl UsartHandler {
...
 
         // Recv as much as we have space for.
         let mut n = 0;
         let mut available_rx_space =
             self.from_rx.capacity() - self.from_rx.len();
 
+        // Ensure we always have space for at least 16 bytes (the stm32h7 usart
+        // FIFO depth), even if that means discarding old data.
+        if available_rx_space < 16 {
+            self.from_rx.drain(..(16 - available_rx_space));
```
mkeeter, line 240: (same concern about `drain` being expensive here)

task/mgmt-gateway/src/mgs_handler.rs
```diff
@@ -206,19 +204,69 @@ impl SpHandler for MgsHandler {
...
         }
+
+        // Have we already received some or all of this data? (MGS may resend
+        // packets that for which it hasn't received our ACK.)
+        if self.serial_console_write_offset > offset {
+            let skip = self.serial_console_write_offset - offset;
+            // Have we already seen _all_ of this data? If so, just reply that
+            // we're ready for the data that comes after it.
+            if skip >= data.len() as u64 {
+                return Ok(offset + data.len() as u64);
```
mkeeter, line 247: Should we return `self.serial_console_write_offset` here instead?

e.g. if 
```
self.serial_console_write_offset = 128
offset = 64
data.len() = 32
```
we want the host to start sending us data at offset 128, not 96.

(or will the host get confused that we've managed to write more data than it thinks it's sent us?)

task/mgmt-gateway/src/main.rs
mkeeter: [`heapless::Deque`](https://docs.rs/heapless/latest/heapless/struct.Deque.html) looks promising (and already has an `as_slices` method!)

task/mgmt-gateway/src/main.rs
jgallagher: By "log" do you mean a ringbuf entry or something else?

task/mgmt-gateway/src/main.rs
jgallagher: Switched to `heapless::Deque` in 215a1a02a33f85858200b98a64e90fd99f56039d, replacing all the drains with loops over `pop_front()`s and rewriting the bit around `available_rx_space`. It's a big enough change to probably be worth a second look, if you don't mind.

task/mgmt-gateway/src/mgs_handler.rs
jgallagher: > (or will the host get confused that we've managed to write more data than it thinks it's sent us?)

Exactly this, at least as currently implemented. In practice we should never see `skip > data.len()` here, unless something has gone horribly wrong, because the host should never go backwards in time, and only moves forward when we ack its messages. We may see `skip == data.len()` if our ack gets lost and the host resends the same packet it most recently sent, though.

mkeeter (approved):

PR #739, opened 2022-08-30
Add mgmt-gateway to gimlet/sidecar/psc builds

jgallagher:
Builds on #738. The bulk of the code changes here are to break `mgs_handler.rs` up into gimlet-specific parts (serial console proxying) and common parts (updating the SP, etc.); the latter have moved to `mgs_common.rs`. What was `mgs_handler` is now `mgs_gimlet`, and this PR adds new `mgs_sidecar` and `mgs_psc` submodules (that only support the functionality provided by `mgs_common`). Exactly one of the three modules must be selected via the appropriate cargo feature. Failing to enable one feature (or enabling more than one) will result in a compilation error.

### DIFF
--- a/task/mgmt-gateway/src/main.rs
+++ b/task/mgmt-gateway/src/main.rs
@@ -16,6 +16,16 @@ use task_net_api::{
 };
 use userlib::{sys_recv_closed, task_slot, TaskId, UnwrapLite};
 
+mod mgs_common;
+
+// If the build system enables multiple of the gimlet/sidecar/psc features, this
+// sequence of `cfg_attr`s will trigger an unused_attributes warning. We can
+// turn this into a hard error via this `deny`, which will catch any such build
+// system misconfiguration.
+#[deny(unused_attributes)]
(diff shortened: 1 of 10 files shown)


task/mgmt-gateway/src/main.rs
```diff
@@ -16,6 +16,16 @@ use task_net_api::{
...
 use userlib::{sys_recv_closed, task_slot, TaskId, UnwrapLite};
 
+mod mgs_common;
+
+// If the build system enables multiple of the gimlet/sidecar/psc features, this
+// sequence of `cfg_attr`s will trigger an unused_attributes warning. We can
+// turn this into a hard error via this `deny`, which will catch any such build
+// system misconfiguration.
+#[deny(unused_attributes)]
+#[cfg_attr(feature = "gimlet", path = "mgs_gimlet.rs")]
```
mkeeter, line 26: 🤯 I never knew about this!

(and could use it to clean up BSPs in various tasks)

mkeeter (approved):

PR #740, opened 2022-08-30
Better warning message on kernel overflow

mkeeter:


build/xtask/src/dist.rs
```diff
@@ -1104,9 +1112,26 @@ fn build(
...
-        .status()
+        .output()
         .context(format!("failed to run rustc ({:?})", cmd))?;
-    if !status.success() {
+    std::io::stderr().write_all(&status.stderr);
+    if !status.status.success() {
+        // We've got a special case here: if the kernel memory is too small,
+        // then the build will fail with a cryptic linker error.  We can't
+        // necessarily do normal string search, because the `stderr` may contain
+        // ANSI control characters, so do a *very dumb* search here:
```
jgallagher, line 1122: Is the type-system-level reason here that `status.stderr` is a `Vec<u8>`, and we can't (and wouldn't want to) convert that to a `String`? Could we use https://docs.rs/memchr/latest/memchr/memmem/index.html instead of doing this thing that looks O(n*m) (although `m` is small)?

jgallagher (approved):
Thanks! This will definitely help me in the future. :-D

build/xtask/src/dist.rs
mkeeter: Yeah, we can't convert to a string because there could (and will!) be ANSI control characters in there.  We've already got a `memchr` dep, so I'll switch to that instead.

PR #745, opened 2022-08-31
Clean unused dep and BSP from Gimlet HF server

mkeeter:


flihp (approved):
LGTM - farewell gimlet_a

Issue #741, opened 2022-08-30
mutable_statics!() can put large data structures on the stack


jgallagher:
I ran into this trying to put a more-complex-than-plain-arrays type inside a `mutable_statics!()` call and noticed my `stackmargin` had jumped up by a few kilobytes. I made a small reproducer in this commit: https://github.com/oxidecomputer/hubris/commit/7910bd1d457f6c140400eda452b26aab25fdf4c5

On that commit, `humility stackmargin` reports

```
ID TASK                STACKBASE  STACKSIZE   MAXDEPTH     MARGIN
10 udpecho            0x20002000       4096       2624       1472
```

This definitely smells like compiler optimizer issues; if you apply this diff to that commit:

```diff
--- a/task/udpecho/src/main.rs
+++ b/task/udpecho/src/main.rs
@@ -64,7 +64,7 @@ fn main() -> ! {
 const STACK_SMASHER_BUF_SIZE: usize = 1024;

 struct StackSmasher {
-    counter1: Option<usize>,
+    counter1: usize,
     counter2: usize,
     data: [u8; STACK_SMASHER_BUF_SIZE],
 }
@@ -72,7 +72,7 @@ struct StackSmasher {
 impl Default for StackSmasher {
     fn default() -> Self {
         Self {
-            counter1: None,
+            counter1: 0,
             counter2: 0,
             data: [0; STACK_SMASHER_BUF_SIZE],
         }
```

the stack usage drops back down to the much more reasonable level it is without any of these changes:

```
ID TASK                STACKBASE  STACKSIZE   MAXDEPTH     MARGIN
10 udpecho            0x20002000       4096        656       3440
```

In the disassembly of the commit with the high stack usage, I see a couple large memcpys that aren't there in the lower stack usage variant:

```asm
0000007c <main>:
      7c:       b580            push    {r7, lr}
      7e:       466f            mov     r7, sp
      80:       f5ad 6d1b       sub.w   sp, sp, #2480   ; 0x9b0
      84:       f8df 0718       ldr.w   r0, [pc, #1816] ; 7a0 <main+0x724>
      88:       f240 31fe       movw    r1, #1022       ; 0x3fe
      8c:       8800            ldrh    r0, [r0, #0]
      8e:       4288            cmp     r0, r1
      90:       d105            bne.n   9e <main+0x22>
      92:       f8df 0758       ldr.w   r0, [pc, #1880] ; 7ec <main+0x770>
[...]

PR #746, opened 2022-08-31
Replace tinyvec::ArrayVec with heapless::{Vec,Deque}.

jgallagher:
Because `heapless` data structures' `::new()` methods are const, we can
put them into statics directly. This allows us to reduce all our
`mutable_statics!` usage to plain `[u8; N]` arrays, avoiding any stack
issues related to https://github.com/oxidecomputer/hubris/issues/741.

### DIFF
--- a/task/mgmt-gateway/src/main.rs
+++ b/task/mgmt-gateway/src/main.rs
@@ -30,8 +30,6 @@ mod mgs_handler;
 
 use self::mgs_handler::MgsHandler;
 
-type SerializedMessageBuf = [u8; gateway_messages::MAX_SERIALIZED_SIZE];
-
 task_slot!(JEFE, jefe);
 task_slot!(NET, net);
 task_slot!(SYS, sys);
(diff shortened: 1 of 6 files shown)


task/mgmt-gateway/src/main.rs
```diff
@@ -121,14 +119,21 @@ fn main() {
...
+    rx_buf: &'static mut [u8; gateway_messages::MAX_SERIALIZED_SIZE],
     packet_to_send: Option<UdpMetadata>,
 }
 
 impl NetHandler {
-    fn new(buffers: &'static mut [SerializedMessageBuf; 2]) -> Self {
-        let [tx_buf, rx_buf] = buffers;
+    fn claim_static_resources() -> Self {
+        let tx_buf = mutable_statics! {
+            static mut NET_TX_BUF: [u8; gateway_messages::MAX_SERIALIZED_SIZE] =
```
mkeeter, line 130: Nit: I think you can put both arrays into the same `mutable_statics!` call, saving an atomic.

jgallagher: Ooh, thanks. I was doing this as `[[u8; N]; 2]` before; didn't realize the macro supported multiple statics.

mkeeter (approved):

PR #748, opened 2022-09-01
Add `SendLeaseRead/Write` functions

mkeeter:
This is infrastructure for allowing `humility` to make _generic_ calls that require a lease, rather than having to special-case them (e.g. `qspi_page_program`).

### DIFF
--- a/task/hiffy/src/lpc55.rs
+++ b/task/hiffy/src/lpc55.rs
@@ -354,6 +354,8 @@ fn gpio_reset(
 pub(crate) static HIFFY_FUNCS: &[Function] = &[
     crate::common::sleep,
     crate::common::send,
+    crate::common::send_lease_read,
+    crate::common::send_lease_write,
     #[cfg(feature = "gpio")]
     gpio_input,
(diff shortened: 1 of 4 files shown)


kc8apf (approved):

bcantrill (approved):
Looks great -- ship it!

task/hiffy/src/lpc55.rs
```diff
@@ -354,6 +354,8 @@ fn gpio_reset(
 pub(crate) static HIFFY_FUNCS: &[Function] = &[
     crate::common::sleep,
     crate::common::send,
+    crate::common::send_lease_read,
```
bcantrill, line 357: This needs to be added to the `Functions` enum as well; as it stands, LPC55 Humliity-side consumers will call the wrong function.  And as long as this change is being made: it should also be added to the G0 array and enum.

bcantrill (commented):
The LPC55 issue should be fixed -- with my apologies that this mechanism is so brittle.

PR #753, opened 2022-09-06
Add sleeps to gimlet-hf-server's poll_for_write_complete

jgallagher:
`poll_for_write_complete()` is called by three writers (bulk_erase,
sector_erase, and page_program) that are writing vastly different
amounts of data (32MiB, 64KiB, 256 bytes). This commit adds an argument
to `poll_for_write_complete()` to control how long to sleep between
polls.

Based on experimental data, bulk_erase on an already-erased device (the
fastest scenario) takes about 400 ticks, sector_erase on an
already-erased sector takes about 5 ticks, and page_program takes 0-1
ticks. I added sleep times of 100 ticks, 1 tick, and no sleep (continue
busy waiting) for the three callers respectively, but am open to other
suggestions if folks disagree with those choices.

### DIFF
--- a/drv/gimlet-hf-server/src/main.rs
+++ b/drv/gimlet-hf-server/src/main.rs
@@ -202,7 +202,7 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         self.check_muxed_to_sp()?;
         set_and_check_write_enable(&self.qspi)?;
         self.qspi.bulk_erase();
-        poll_for_write_complete(&self.qspi);
+        poll_for_write_complete(&self.qspi, Some(100));
         Ok(())
     }
 
@@ -221,7 +221,7 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
 
         set_and_check_write_enable(&self.qspi)?;
         self.qspi.page_program(addr, &self.block[..data.len()]);
-        poll_for_write_complete(&self.qspi);
+        poll_for_write_complete(&self.qspi, None);
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


drv/gimlet-hf-server/src/main.rs
```diff
@@ -202,7 +202,7 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         self.check_muxed_to_sp()?;
         set_and_check_write_enable(&self.qspi)?;
         self.qspi.bulk_erase();
-        poll_for_write_complete(&self.qspi);
+        poll_for_write_complete(&self.qspi, Some(100));
```
jgallagher, line 205: I didn't note this in the PR description, but I chose this sleep time based on the _fastest_ possible bulk erase time (similarly for sector erase). bulk_erase can take much, much longer (order of minutes), but spinning while sleeping for 100 ticks between each poll for minutes is probably okay?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -221,7 +221,7 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
 
         set_and_check_write_enable(&self.qspi)?;
         self.qspi.page_program(addr, &self.block[..data.len()]);
-        poll_for_write_complete(&self.qspi);
+        poll_for_write_complete(&self.qspi, None);
```
jgallagher, line 224: This leaves page_program as a busywait; given my measurements saw this come back as 0 or 1 tick to complete, I _think_ that's the right decision, but I could easily be convinced to switch this to `Some(1)` if that makes anyone nervous.

mkeeter (approved):

PR #754, opened 2022-09-07
Add `auxflash` task and blob generation

mkeeter:
This is the first step of implementing RFD 311.  It does two things:

- Adds a small `auxflash` task to the Sidecar image, which can read, write, and validate checksums of slots in the auxiliary flash chip
- Inserts `[[auxflash.blobs]]` into the config file, which specifies files to be packed into `auxi.tlvc` in the resulting Hubris archive

### DIFF
--- a/build/xtask/src/auxflash.rs
+++ b/build/xtask/src/auxflash.rs
@@ -5,7 +5,6 @@
 use anyhow::{bail, Context, Result};
 use serde::Deserialize;
 use sha3::{Digest, Sha3_256};
-use zerocopy::AsBytes;
 
 /// List of binary blobs to include in the auxiliary flash binary shipped with
 /// this image.  The auxiliary flash is used to offload storage of large
@@ -23,42 +22,11 @@ pub struct AuxFlashBlob {
     pub tag: String,
 }
 
-/// Encode the given data as a tagged TLV-C chunk
-fn data_to_tlvc(tag: &str, data: &[u8]) -> Result<Vec<u8>> {
-    if tag.len() != 4 {
-        bail!("Tag must be a 4-byte value, not '{}'", tag);
-    }
-    let mut out = vec![];
-    let mut header = tlvc::ChunkHeader {
-        tag: tag.as_bytes().try_into().unwrap(),
-        len: 0.into(),
-        header_checksum: 0.into(),
-    };
-    out.extend(header.as_bytes());
-
-    let c = tlvc::compute_body_crc(data);
-
-    out.extend(data);
-    let body_len = out.len() - std::mem::size_of::<tlvc::ChunkHeader>();
-    let body_len = u32::try_from(body_len).unwrap();
-
-    // TLV-C requires the body to be padded to a multiple of four!
-    while out.len() & 0b11 != 0 {
-        out.push(0);
-    }
-    out.extend(c.to_le_bytes());
-
-    // Update the header.
-    header.len.set(body_len);
-    header.header_checksum.set(header.compute_checksum());
-
-    out[..std::mem::size_of::<tlvc::ChunkHeader>()]
-        .copy_from_slice(header.as_bytes());
-    Ok(out)
-}
-
 /// Packs a single blob into a TLV-C structure
-fn pack_blob(blob: &AuxFlashBlob) -> Result<Vec<u8>> {
+fn pack_blob(blob: &AuxFlashBlob) -> Result<tlvc_text::Piece> {
+    if blob.tag.len() != 4 {
+        bail!("Tag must be a 4-byte value, not '{}'", blob.tag);
+    }
     let data = std::fs::read(&blob.file)
         .with_context(|| format!("Could not read blob {}", blob.file))?;
     let data = if blob.compress {
@@ -66,7 +34,11 @@ fn pack_blob(blob: &AuxFlashBlob) -> Result<Vec<u8>> {
     } else {
         data
     };
-    data_to_tlvc(&blob.tag, &data)
+    let tag: [u8; 4] = blob.tag.as_bytes().try_into().unwrap();
+    Ok(tlvc_text::Piece::Chunk(
+        tlvc_text::Tag::new(tag),
+        vec![tlvc_text::Piece::Bytes(data)],
+    ))
 }
 
 /// Constructs an auxiliary flash image, based on RFD 311
@@ -75,12 +47,17 @@ fn pack_blob(blob: &AuxFlashBlob) -> Result<Vec<u8>> {
 pub fn build_auxflash(aux: &AuxFlash) -> Result<([u8; 32], Vec<u8>)> {
     let mut auxi = vec![];
     for f in &aux.blobs {
-        auxi.extend(pack_blob(f)?);
+        auxi.push(pack_blob(f)?);
     }
-    let sha = Sha3_256::digest(&auxi);
+    let sha = Sha3_256::digest(tlvc_text::pack(&auxi));
 
-    let mut out = vec![];
-    out.extend(data_to_tlvc("CHCK", &sha)?);
-    out.extend(data_to_tlvc("AUXI", &auxi)?);
-    Ok((sha.into(), out))
+    let out = [
+        tlvc_text::Piece::Chunk(
+            tlvc_text::Tag::new(*b"CHCK"),
+            vec![tlvc_text::Piece::Bytes(sha.to_vec())],
+        ),
+        tlvc_text::Piece::Chunk(tlvc_text::Tag::new(*b"AUXI"), auxi),
+    ];
+
+    Ok((sha.into(), tlvc_text::pack(&out)))
 }
--- a/build/xtask/src/auxflash.rs
+++ b/build/xtask/src/auxflash.rs
@@ -1,0 +1,86 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use anyhow::{bail, Context, Result};
+use serde::Deserialize;
+use sha3::{Digest, Sha3_256};
+use zerocopy::AsBytes;
+
+/// List of binary blobs to include in the auxiliary flash binary shipped with
+/// this image.  The auxiliary flash is used to offload storage of large
+/// configuration files (e.g. FPGA bitstreams)
+#[derive(Clone, Debug, Deserialize)]
+pub struct AuxFlash {
+    pub blobs: Vec<AuxFlashBlob>,
+}
+
+/// A single binary blob, encoded into the auxiliary flash file.
+#[derive(Clone, Debug, Deserialize)]
+pub struct AuxFlashBlob {
+    pub file: String,
+    pub compress: bool,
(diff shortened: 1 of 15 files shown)


build/xtask/src/dist.rs
```diff
@@ -573,6 +575,66 @@ fn write_gdb_script(cfg: &PackageConfig, image_name: &str) -> Result<()> {
...
 }
 
+/// Encode the given data as a tagged TLV-C chunk
+fn data_to_tlvc(tag: &str, data: &[u8]) -> Result<Vec<u8>> {
+    if tag.len() != 4 {
+        bail!("Tag must be a 4-byte value, not '{}'", tag);
+    }
+    let mut out = vec![];
+    let mut header = tlvc::ChunkHeader {
+        tag: tag.bytes().collect::<Vec<u8>>().try_into().unwrap(),
```
jgallagher, line 585: I think this could be 
```suggestion
        tag: tag.as_bytes().try_into().unwrap(),
```

build/xtask/src/dist.rs
```diff
@@ -573,6 +575,66 @@ fn write_gdb_script(cfg: &PackageConfig, image_name: &str) -> Result<()> {
...
+        len: 0.into(),
+        header_checksum: 0.into(),
+    };
+    out.extend(header.as_bytes());
+    let c = tlvc::begin_body_crc();
+    let mut c = c.digest();
+    c.update(data);
```
jgallagher, line 592: I think these three lines could be

```rust
let c = tlvc::compute_body_crc(data);
```

build/xtask/src/dist.rs
```diff
@@ -573,6 +575,66 @@ fn write_gdb_script(cfg: &PackageConfig, image_name: &str) -> Result<()> {
...
+    out.extend(data);
+    let body_len = out.len() - std::mem::size_of::<tlvc::ChunkHeader>();
+    let body_len = u32::try_from(body_len).unwrap();
+    while out.len() & 0b11 != 0 {
```
jgallagher, line 596: Maybe worth a comment that tlvc requires the body to be padded to a multiple of four (which is not included in the length)?

Nit: I think this would be marginally clearer as `while out.len() % 4 != 0`, but I could be in the minority.

build/xtask/src/dist.rs
```diff
@@ -573,6 +575,66 @@ fn write_gdb_script(cfg: &PackageConfig, image_name: &str) -> Result<()> {
...
+
+    out[..std::mem::size_of::<tlvc::ChunkHeader>()]
+        .copy_from_slice(header.as_bytes());
+    Ok(out)
+}
+
+/// Packs a single blob into a TLV-C structure
+fn pack_blob(blob: &AuxFlashBlob) -> Result<Vec<u8>> {
+    let data = std::fs::read(&blob.file)
+        .context(format!("Could not read blob {}", blob.file))?;
```
jgallagher, line 613: Nit:

```suggestion
        .with_context(|| format!("Could not read blob {}", blob.file))?;
```

would prevent allocating the error message if the read succeeds.

build/xtask/src/dist.rs
```diff
@@ -573,6 +575,66 @@ fn write_gdb_script(cfg: &PackageConfig, image_name: &str) -> Result<()> {
...
+
+/// Constructs an auxiliary flash image, based on RFD 311
+fn build_auxflash(aux: &AuxFlash) -> Result<Vec<u8>> {
+    let mut auxi = vec![];
+    for f in &aux.blobs {
+        auxi.extend(pack_blob(f)?);
+    }
+    let mut sha = Sha3_256::new();
+    sha.update(&auxi);
+    let sha_out = sha.finalize();
```
jgallagher, line 630: I think these three lines could be

```rust
let sha_out = Sha3_256::digest(&auxi);
```

drv/auxflash-server/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+
+    // Ensure hold time for reset in case we just restarted.
+    // TODO look up actual hold time requirement
+    hl::sleep_for(1);
+
+    // Release reset and let it stabilize.
+    sys.gpio_set(qspi_reset).unwrap();
+    hl::sleep_for(10);
+
+    // TODO: check the ID and make sure it's what we expect
```
jgallagher, line 125: Did you want to do this before merging?

drv/auxflash-server/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+
+impl ServerImpl {
+    fn poll_for_write_complete(&self) {
+        loop {
+            let status = self.qspi.read_status();
+            if status & 1 == 0 {
+                // ooh we're done
+                break;
+            }
+            hl::sleep_for(10);
```
jgallagher, line 151: Thank you! :grin:

drv/auxflash-server/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+        &mut self,
+        _: &RecvMessage,
+        slot: u32,
+    ) -> Result<AuxFlashChecksum, RequestError<AuxFlashError>> {
+        if slot >= SLOT_COUNT {
+            return Err(AuxFlashError::InvalidSlot.into());
+        }
+        let handle = SlotReader {
+            qspi: &self.qspi,
+            base: slot * 1024 * 1024,
```
jgallagher, line 201:
```suggestion
            base: slot * SLOT_SIZE,
```

drv/auxflash-server/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+                } else if chunk.len() != 32 {
+                    return Err(AuxFlashError::BadChckSize.into());
+                }
+                let mut out = [0; 32];
+                chunk
+                    .read_exact(0, &mut out)
+                    .map_err(|_| AuxFlashError::ChunkReadFail)?;
+                chck_expected = Some(out);
+            } else if &chunk.header().tag == b"AUXI" {
+                let mut sha = Sha3_256::new();
```
jgallagher, line 221: Should we check that `chck_actual.is_none()` here?

drv/auxflash-server/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+            }
+        }
+        if chck_expected.is_none() {
+            return Err(AuxFlashError::MissingChck.into());
+        } else if chck_actual.is_none() {
+            return Err(AuxFlashError::MissingAuxi.into());
+        } else if chck_expected != chck_actual {
+            return Err(AuxFlashError::ChckMismatch.into());
+        } else {
+            return Ok(AuxFlashChecksum(chck_expected.unwrap()));
```
jgallagher, line 245: I might write this as a `match` just to avoid the unwrap, something like:

```rust
match (chck_expected, chck_actual) {
    (None, _) => Err(AuxFlashError::MissingChck.into()),
    (_, None) => Err(AuxFlashError::MissingAuxi.into()),
    (Some(expected), Some(actual)) if expected == actual => Ok(AuxFlashChecksum(expected)),
    (Some(_), Some(_)) => Err(AuxFlashError::ChckMismatch.into()),
}
```

but not sure if it's really better :man_shrugging:

drv/auxflash-server/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+            return Err(AuxFlashError::InvalidSlot.into());
+        }
+        let mem_start = slot * SLOT_SIZE;
+        let mem_end = mem_start + SLOT_SIZE;
+
+        self.set_and_check_write_enable()?;
+        let mut addr = mem_start;
+        while addr < mem_end {
+            self.qspi.sector_erase(addr);
+            addr += 64 * 1024;
```
jgallagher, line 264: Using sector erase from `mgmt-gateway` I ended up adding a `SECTOR_SIZE_BYTES` constant to `gimlet-hf-api`, but maybe it should really be in `drv-stm32h7-qspi`?

drv/auxflash-server/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+        offset: u32,
+        data: LenLimit<Leased<R, [u8]>, 256>,
+    ) -> Result<(), RequestError<AuxFlashError>> {
+        let mut buf = [0u8; 256];
+        data.read_range(0..data.len(), &mut buf[..data.len()])
+            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+
+        self.set_and_check_write_enable()?;
+
+        let addr = slot * SLOT_SIZE + offset;
```
jgallagher, line 283: Should we check that `addr + data.len()` doesn't bleed into the next slot? (similarly for read)

drv/auxflash-server/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+        _: &RecvMessage,
+        slot: u32,
+        offset: u32,
+        dest: LenLimit<Leased<W, [u8]>, 256>,
+    ) -> Result<(), RequestError<AuxFlashError>> {
+        let mut buf = [0u8; 256];
+
+        let addr = slot * SLOT_SIZE + offset;
+        self.qspi.read_memory(addr, &mut buf[..dest.len()]);
+        dest.write_range(0..dest.len(), &mut buf[..dest.len()])
```
jgallagher, line 300:
```suggestion
        dest.write_range(0..dest.len(), &buf[..dest.len()])
```

jgallagher (approved):
Left a bunch of nitpicky comments, but this LGTM! :+1:

build/xtask/src/dist.rs
mkeeter: Added a comment!  (This is copy-pasta from the TLV-C `texttool` example, so I'll leave it be otherwise)

drv/auxflash-server/src/main.rs
mkeeter: Good idea, I ended up with
```
        match (chck_expected, chck_actual) {
            (None, _) => Err(AuxFlashError::MissingChck.into()),
            (_, None) => Err(AuxFlashError::MissingChck.into()),
            (Some(a), Some(b)) => {
                if a == b {
                    Err(AuxFlashError::ChckMismatch.into())
                } else {
                    Ok(AuxFlashChecksum(chck_expected.unwrap()))
                }
            }
        }
```
which is a pleasant middle ground

drv/auxflash-server/src/main.rs
mkeeter: Good idea, done

build/xtask/src/auxflash.rs
```diff
@@ -0,0 +1,86 @@
...
+    out.extend(c.to_le_bytes());
+
+    // Update the header.
+    header.len.set(body_len);
+    header.header_checksum.set(header.compute_checksum());
+
+    out[..std::mem::size_of::<tlvc::ChunkHeader>()]
+        .copy_from_slice(header.as_bytes());
+    Ok(out)
+}
```
labbott, line 58: This seems like a useful function to add to the tlv-c repo

mkeeter: Aha, it looks like I could use `tlvc-text` as a library instead of a command-line tool, so I'll try that in the morning.

mkeeter: Fixes are pending https://github.com/oxidecomputer/tlvc/pull/2

PR #755, opened 2022-09-07
lib-dice: Add magic value to handoff structure.

flihp:
This is intended as a more robust mechanism for the task consuming the DICE artifacts to identify valid values.

### DIFF
--- a/lib/dice/src/handoff.rs
+++ b/lib/dice/src/handoff.rs
@@ -70,13 +70,31 @@ impl<'a> Handoff<'a> {
 // instead of individual certs.
 #[derive(Deserialize, Serialize, SerializedSize)]
 pub struct AliasData {
+    pub magic: [u8; 16],
     pub seed: AliasOkm,
     pub alias_cert: AliasCert,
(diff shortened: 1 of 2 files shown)


lib/dice/src/handoff.rs
```diff
@@ -70,13 +70,31 @@ impl<'a> Handoff<'a> {
...
     pub alias_cert: AliasCert,
     pub deviceid_cert: DeviceIdSelfCert,
 }
 
 impl AliasData {
-    pub fn from_mem() -> Self {
+    const MAGIC: [u8; 16] = [
+        0x3e, 0xbc, 0x3c, 0xdc, 0x60, 0x37, 0xab, 0x86, 0xf0, 0x60, 0x20, 0x52,
+        0xc4, 0xfd, 0xd5, 0x58,
+    ];
```
labbott, line 83: a short comment if this is meaningful beyond "yes this is good" would be nice

labbott (approved):

kc8apf (approved):

PR #756, opened 2022-09-07
STM32G0 I2C support, VPD support, and potpourri

bcantrill:
Draft commit message:

```
This work:

- Adds i2c support for STM32G0-class parts, commonalizing the crate
  (which suports both H7 and G0) into stm32xx-i2c 
- Adds a VPD task for VPD reading/programming
- Fleshes out the Donglet application description to include the VPD
  programming board
- Adds Sharkfin and VPD descriptions to the Gimlet app definition
- Fixes i2c support for buses that have multiple muxes, which was more
  or less entirely broken
- Adds validation support for the MAX5970 hot-swap controller
- Fixes LED support for the G031 Nucleo board
```


build/i2c/src/lib.rs
```diff
@@ -342,10 +342,13 @@ impl ConfigGenerator {
...
             r##"
-    use drv_stm32h7_i2c::I2cController;
+    #[allow(dead_code)]
+    pub const NCONTROLLERS: usize = {ncontrollers};
+
+    use drv_stm32xx_i2c::I2cController;
 
-    pub fn controllers() -> [I2cController<'static>; {}] {{"##,
-            self.controllers.len()
+    pub fn controllers() -> [I2cController<'static>; {ncontrollers}] {{"##,
```
mkeeter, line 350: Nit: now that it's defined as a constant, could we do
```suggestion
    pub fn controllers() -> [I2cController<'static>; NCONTROLLERS] {{"##,
```
?

build/xtask/src/flash.rs
```diff
@@ -78,6 +80,14 @@ impl FlashConfig {
         self
     }
 
+    //
+    // Set the chip
+    //
+    fn chip<'a>(&'a mut self, val: &str) -> &'a mut Self {
```
mkeeter, line 86: We don't need explicit lifetimes here:
```suggestion
    fn chip(&mut self, val: &str) -> &mut Self {
```

mkeeter: (also, maybe name this `set_chip` instead of just `chip`?)

bcantrill: Fixed

build/i2c/src/lib.rs
bcantrill: Agreed -- fixed

PR #758, opened 2022-09-08
Add `capacity()` idol call to gimlet-hf-server.

jgallagher:
Also adds a couple constants to its API describing the sizes of `page_program()` and `sector_erase()` operations.

All of these additions will be used by `mgmt-gateway` shortly in updating the host boot flash.

### DIFF
--- a/drv/gimlet-hf-api/src/lib.rs
+++ b/drv/gimlet-hf-api/src/lib.rs
@@ -45,4 +45,16 @@ pub enum HfDevSelect {
     Flash1 = 1,
 }
 
+/// Size in bytes of a single page of data (i.e., the max length of slice we
+/// accept for `page_program()` and `read()`).
+// Note: There is no static check that this matches what's in our idl file in
+// terms of _client_ generation, but the server can use this constant in its
+// trait impl, which will produce a compile-time error if it doesn't match the
+// length in the idl file.
+pub const PAGE_SIZE_BYTES: usize = 256;
+
+/// Size in bytes of a single sector of data (i.e., the size of the data erased
+/// by a call to `sector_erase()`).
+pub const SECTOR_SIZE_BYTES: usize = 65_536;
+
 include!(concat!(env!("OUT_DIR"), "/client_stub.rs"));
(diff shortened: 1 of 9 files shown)


drv/gimlet-hf-api/src/lib.rs
```diff
@@ -45,4 +45,16 @@ pub enum HfDevSelect {
...
+/// accept for `page_program()` and `read()`).
+// Note: There is no static check that this matches what's in our idl file in
+// terms of _client_ generation, but the server can use this constant in its
+// trait impl, which will produce a compile-time error if it doesn't match the
+// length in the idl file.
+pub const PAGE_SIZE_BYTES: usize = 256;
+
+/// Size in bytes of a single sector of data (i.e., the size of the data erased
+/// by a call to `sector_erase()`).
+pub const SECTOR_SIZE_BYTES: usize = 65_536;
```
jgallagher, line 58: This came up [in another PR](https://github.com/oxidecomputer/hubris/pull/754#discussion_r965240718) - maybe this should actually be in `drv-stm32-qspi` and re-exported here instead?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -135,6 +135,8 @@ fn main() -> ! {
     let mut server = ServerImpl {
         qspi,
         block: [0; 256],
+        // Capacity is stored as log2 size
+        capacity: 1 << capacity,
```
mkeeter, line 139: Not your changes, but it would be nice if the first `capacity` was named `log2_capacity` here, for clarity

drv/gimlet-hf-api/src/lib.rs
mkeeter: I mostly agree, with a small asterisk that it may not be consistent for all QSPI flashes (e.g. there are variants of the Sidecar flash which use 256K sectors).  Putting it in `drv-stm32-qspi` is not _worse_, though, and would let us use it in multiple places.

mkeeter (approved):

drv/gimlet-hf-server/src/main.rs
jgallagher: Agreed, plus that would eliminate the need for the comment :+1:

PR #759, opened 2022-09-08
mgmt-gateway refactor: extract generic `UpdateBuffer`

jgallagher:
This PR extracts `UpdateBuffer` from `mgs_common.rs` (where it's used to buffer incoming an incoming SP image) and makes it generic over:

* the block size
* the function to write a single block
* the function to finalize after the last block is written

which will allow it to be used for both SP updates (in this PR) and host boot flash updates (in a subsequent PR), and hopefully other update components in the future.

One design choice that I'm not sure about - carving it out like this assumes separate buffers for every component we want to update. That theoretically allows multiple component updates to be streamed in simultaneously, but it also may end up being prohibitive in memory usage, because it means we set a side a component-update-block-sized buffer for each one. If we end up wanting to share a single or small number of shared update buffers, this will need some changes.

### DIFF
--- a/task/mgmt-gateway/src/update_buffer.rs
+++ b/task/mgmt-gateway/src/update_buffer.rs
@@ -3,10 +3,20 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use crate::{Log, __RINGBUF};
-use core::marker::PhantomData;
 use gateway_messages::ResponseError;
 use ringbuf::ringbuf_entry;
 use userlib::UnwrapLite;
+
+/// Type alias for a callback function that writes a single block of data.
+pub type WriteBlockFn<T> = fn(
+    user_data: &T,
+    block_index: usize,
+    block_data: &[u8],
+) -> Result<(), ResponseError>;
+
+/// Type alias for a callback function that finalizes an update after all blocks
+/// have been successfully written.
+pub type FinalizeFn<T> = fn(user_data: &T) -> Result<(), ResponseError>;
 
 /// `UpdateBuffer` provides common logic for apply updates over the management
 /// network, assuming a common pattern of:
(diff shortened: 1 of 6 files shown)


task/mgmt-gateway/src/update_buffer.rs
```diff
@@ -0,0 +1,148 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::{Log, __RINGBUF};
+use core::marker::PhantomData;
+use gateway_messages::ResponseError;
+use ringbuf::ringbuf_entry;
```
mkeeter, line 8: Nit: instead of importing `__RINGBUF`, you could
```
use ringbuf::ringbuf_entry_root as ringbuf_entry
```
(which logs to a ringbuf in the root of the crate)

task/mgmt-gateway/src/update_buffer.rs
```diff
@@ -0,0 +1,148 @@
...
+    write_block_fn: fn(&T, usize, &[u8]) -> Result<(), ResponseError>,
+    finalize_fn: fn(&T) -> Result<(), ResponseError>,
+    marker: PhantomData<fn() -> T>,
+}
+
+impl<T, const BLOCK_SIZE: usize> UpdateBuffer<T, BLOCK_SIZE> {
+    pub fn new(
+        buf: &'static mut heapless::Vec<u8, BLOCK_SIZE>,
+        write_block_fn: fn(
+            user_data: &T,
```
mkeeter, line 34: It might be worthwhile to extract `WriteBlockFn` and `FinalizeFn` types, then use them here and in the `struct` definition.

mkeeter (approved):

task/mgmt-gateway/src/update_buffer.rs
jgallagher: TIL! Thanks

PR #761, opened 2022-09-08
lib-dice: Replace START / END with Range in cert templates.

### LINKED ISSUE #747: use core::ops::Range<usize> in place of start / end / length values in DICE tmpls
https://github.com/oxidecomputer/hubris/pull/698#discussion_r960997191

flihp:
This simplifies the get / set functions in the Cert trait. This resolves #747

### DIFF
--- a/lib/dice/src/cert.rs
+++ b/lib/dice/src/cert.rs
@@ -3,8 +3,9 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use crate::{
-    alias_cert_tmpl, deviceid_cert_tmpl, SerialNumber, NOTBEFORE_LENGTH,
+    alias_cert_tmpl, deviceid_cert_tmpl, CertSerialNumber, SerialNumber,
 };
+use core::ops::Range;
 use hubpack::SerializedSize;
 use salty::constants::{
     PUBLICKEY_SERIALIZED_LENGTH, SIGNATURE_SERIALIZED_LENGTH,
@@ -13,6 +14,7 @@ use salty::signature::{Keypair, PublicKey};
 use serde::{Deserialize, Serialize};
 use serde_big_array::BigArray;
 use unwrap_lite::UnwrapLite;
+use zerocopy::AsBytes;
 
 #[derive(Debug)]
 pub enum CertError {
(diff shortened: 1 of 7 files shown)


labbott (approved):

lib/dice/src/cert.rs
```diff
@@ -29,121 +31,97 @@ pub trait Cert {
...
-                .try_into()
-                .unwrap_lite(),
-        )
-    }
-
-    fn set_serial_number(mut self, sn: u8) -> Self
+    fn read_range<'a, T>(&'a self, r: Range<usize>) -> T
     where
-        Self: Sized,
+        T: TryFrom<&'a [u8]>,
```
kc8apf, line 36: There's nothing inherent about TryFrom<&[u8]> that says it can only return Ok if the entire slice is used.  It happens that you use a pattern of `let x: [u8; X] = self.read_range(R)` which means T is TryFrom<&'a [u8]> for [u8; N] which does fail if the slice and array lengths are different (I had to go look that up: https://doc.rust-lang.org/src/core/array/mod.rs.html#177-186).  I'm not sure if that was an intentional choice here or whether T should always consume the entire slice provided.  This feels like an easy place to make a mistake but I can't think of any easy way to detect such an error.

lib/dice/src/cert.rs
```diff
@@ -29,121 +31,97 @@ pub trait Cert {
...
     {
-        self.as_mut_bytes()[Self::SERIAL_NUMBER_START..Self::SERIAL_NUMBER_END]
-            .copy_from_slice(&sn.to_be_bytes());
+        self.as_bytes()[r].try_into().unwrap_lite()
+    }
 
-        self
+    fn write_range<T: AsBytes>(&mut self, r: Range<usize>, t: &T) {
```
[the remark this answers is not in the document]
flihp: I'll do the same for 'read' vs 'get'

lib/dice/src/cert.rs
flihp: This pattern was intentional. It was the best option I could come up with that didn't rely on an unstable feature (generic_const_exprs). If you're OK with enabling incomplete / unstable features this can be updated, or we can wait for the feature to stabilize. I've created an issue to track this: https://github.com/oxidecomputer/hubris/issues/778

PR #762, opened 2022-09-08
mgmt-gateway: update host flash

jgallagher:
This tracks the corresponding changes to MGS messaging merged in https://github.com/oxidecomputer/omicron/pull/1684:

1. `UpdateStart` has been broken into `UpdatePrepare` and `UpdatePrepareStatus` (which MGS will continue to send periodically until we respond that preparation is done), allowing for updates that have a potentially-long running prep step (like updating host flash, which can take up to several minutes to erase!).
2. Update messages now include a stream-id that we use to correlate related messages; we reject update messages that don't match our current stream ID.
3. Add handling for the new `UpdateAbort` abort message to cancel an in-progress update.

By far the most complex bit of this is 1: I've moved setting the system timer out of `mgs_gimlet` (which previously set it only in relation to flushing serial console uart packets out to MGS) and into `main`: now the MGS handler only returns the deadline it wants `main` to set. If we're in the process of prepping for a host flash update (i.e., we need to erase the host flash), we'll set our deadline to 1 tick from now. When it fires, we'll erase 8 sectors (takes about 1 second, worst case), then return and allow `main` to check for other work. This allows us to continue to be responsive to incoming notifications, importantly network requests (allowing us to respond to the `UpdatePrepareStatus` messages in a timely way!).

### DIFF
--- a/task/mgmt-gateway/src/mgs_gimlet.rs
+++ b/task/mgmt-gateway/src/mgs_gimlet.rs
@@ -2,23 +2,29 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use crate::update_buffer::{UpdateBuffer, UpdateProgress};
 use crate::{
     mgs_common::MgsCommon, vlan_id_from_sp_port, Log, MgsMessage, SYS,
-    TIMER_IRQ, USART_IRQ,
+    USART_IRQ,
 };
 use core::convert::Infallible;
+use core::ops::Range;
 use core::sync::atomic::{AtomicBool, Ordering};
+use drv_gimlet_hf_api::{
+    HfDevSelect, HfError, HfMuxState, HostFlash, PAGE_SIZE_BYTES,
+    SECTOR_SIZE_BYTES,
+};
 use drv_stm32h7_usart::Usart;
 use gateway_messages::{
     sp_impl::SocketAddrV6, sp_impl::SpHandler, BulkIgnitionState,
     DiscoverResponse, IgnitionCommand, IgnitionState, ResponseError,
     SpComponent, SpMessage, SpMessageKind, SpPort, SpState, UpdateChunk,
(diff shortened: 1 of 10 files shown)


task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -543,3 +643,225 @@ fn claim_sp_to_mgs_usart_buf_static(
...
+        Ok(())
+    }
+}
+
+struct HostFlashSectorErase {
+    sectors_to_erase: Range<usize>,
+    most_recent_error: Option<HfError>,
+}
+
+impl Default for HostFlashSectorErase {
```
mkeeter, line 786: Do we need this `Default` implementation?  It seems a little non-obvious (e.g. I could imagine that the default would be to erase all sectors, instead of none!)

task/mgmt-gateway/src/update_buffer.rs
```diff
@@ -62,29 +67,53 @@ impl<T, const BLOCK_SIZE: usize> UpdateBuffer<T, BLOCK_SIZE> {
         }
     }
 
+    pub fn ensure_matching_stream_id(
+        &self,
+        stream_id: u64,
+    ) -> Result<(), ResponseError> {
+        match self.current_update_stream_id {
+            None => Err(ResponseError::UpdateNotPrepared),
```
mkeeter, line 75: Nit: is this the right error code?

Earlier in the PR, `UpdateNotPrepared` seems to mean "you sent a chunk while the update while still in the `UpdatePrepare` state", rather than "there is no current update happening at all".

mkeeter (approved):
A few small comments, but generally looks good

task/mgmt-gateway/src/mgs_gimlet.rs
jgallagher: Hmm, we need something to the effect of it unless I make `HostFlashUpdate::sector_erase` an `Option`. That might be clearer; I'll give it a shot.

Issue #769, opened 2022-09-13
Sidecar mainboard FPGA does not use checksum


mkeeter:
In order to detect when an FPGA does not need to be reconfigured, we bake a checksum into the Hubris image, then write this checksum to the FPGA's `CHECKSUM_SCRATCHPAD*` (or `CS*`) registers.

This is implemented for the Sidecar Front IO board FPGA(s), but not for the Sidecar mainboard FPGA.

PR #774, opened 2022-09-13
Use auxiliary flash on Sidecar

mkeeter:
This PR removes the FPGA images from the Sidecar image in favor of using the auxiliary flash!

### DIFF
--- a/build/xtask/src/auxflash.rs
+++ b/build/xtask/src/auxflash.rs
@@ -5,6 +5,7 @@
 use anyhow::{bail, Context, Result};
 use serde::Deserialize;
 use sha3::{Digest, Sha3_256};
+use std::collections::BTreeMap;
 
 /// List of binary blobs to include in the auxiliary flash binary shipped with
 /// this image.  The auxiliary flash is used to offload storage of large
@@ -22,8 +23,18 @@ pub struct AuxFlashBlob {
     pub tag: String,
 }
 
+#[derive(Clone, Debug)]
+pub struct AuxFlashData {
+    /// Main checksum
+    pub chck: [u8; 32],
+    /// Individual blob checksums
+    pub checksums: BTreeMap<String, [u8; 32]>,
+    /// Full serialized data
+    pub data: Vec<u8>,
+}
+
 /// Packs a single blob into a TLV-C structure
-fn pack_blob(blob: &AuxFlashBlob) -> Result<tlvc_text::Piece> {
+fn pack_blob(blob: &AuxFlashBlob) -> Result<(tlvc_text::Piece, [u8; 32])> {
     if blob.tag.len() != 4 {
         bail!("Tag must be a 4-byte value, not '{}'", blob.tag);
     }
@@ -34,20 +45,26 @@ fn pack_blob(blob: &AuxFlashBlob) -> Result<tlvc_text::Piece> {
     } else {
         data
     };
+    let blob_checksum = Sha3_256::digest(&data);
+
     let tag: [u8; 4] = blob.tag.as_bytes().try_into().unwrap();
-    Ok(tlvc_text::Piece::Chunk(
+    let piece = tlvc_text::Piece::Chunk(
         tlvc_text::Tag::new(tag),
         vec![tlvc_text::Piece::Bytes(data)],
-    ))
+    );
+    Ok((piece, blob_checksum.into()))
 }
 
 /// Constructs an auxiliary flash image, based on RFD 311
 ///
 /// Returns the checksum and the raw data to be saved
-pub fn build_auxflash(aux: &AuxFlash) -> Result<([u8; 32], Vec<u8>)> {
+pub fn build_auxflash(aux: &AuxFlash) -> Result<AuxFlashData> {
     let mut auxi = vec![];
+    let mut blob_checksums = BTreeMap::new();
     for f in &aux.blobs {
-        auxi.push(pack_blob(f)?);
+        let (piece, checksum) = pack_blob(f)?;
+        auxi.push(piece);
+        blob_checksums.insert(f.tag.clone(), checksum);
     }
     let sha = Sha3_256::digest(tlvc_text::pack(&auxi));
 
@@ -59,5 +76,9 @@ pub fn build_auxflash(aux: &AuxFlash) -> Result<([u8; 32], Vec<u8>)> {
         tlvc_text::Piece::Chunk(tlvc_text::Tag::new(*b"AUXI"), auxi),
     ];
 
-    Ok((sha.into(), tlvc_text::pack(&out)))
+    Ok(AuxFlashData {
+        chck: sha.into(),
+        checksums: blob_checksums,
+        data: tlvc_text::pack(&out),
+    })
 }
--- a/build/xtask/src/auxflash.rs
+++ b/build/xtask/src/auxflash.rs
@@ -23,18 +23,22 @@ pub struct AuxFlashBlob {
     pub tag: String,
 }
 
+pub type AuxFlashChecksum = [u8; 32];
+
 #[derive(Clone, Debug)]
 pub struct AuxFlashData {
     /// Main checksum
-    pub chck: [u8; 32],
+    pub chck: AuxFlashChecksum,
     /// Individual blob checksums
-    pub checksums: BTreeMap<String, [u8; 32]>,
+    pub checksums: BTreeMap<String, AuxFlashChecksum>,
     /// Full serialized data
     pub data: Vec<u8>,
 }
 
 /// Packs a single blob into a TLV-C structure
-fn pack_blob(blob: &AuxFlashBlob) -> Result<(tlvc_text::Piece, [u8; 32])> {
+fn pack_blob(
+    blob: &AuxFlashBlob,
+) -> Result<(tlvc_text::Piece, AuxFlashChecksum)> {
     if blob.tag.len() != 4 {
         bail!("Tag must be a 4-byte value, not '{}'", blob.tag);
     }
--- a/drv/auxflash-server/src/main.rs
+++ b/drv/auxflash-server/src/main.rs
@@ -506,18 +506,14 @@ impl idl::InOrderAuxFlashImpl for ServerImpl {
                 let mut inner_reader = outer_chunk.read_as_chunks();
                 while let Ok(Some(inner_chunk)) = inner_reader.next() {
                     if inner_chunk.header().tag == tag {
-                        // At this point, the outer reader is positioned *after*
-                        // the AUXI chunk, and the inner reader is positioned
-                        // *after* our target chunk.  To get to the data of the
-                        // inner chunk, we have to back off, then re-offset
-                        // ourselves by 2x chunk headers.
+                        // At this point, the inner reader is positioned *after*
+                        // our target chunk.  We back off by the full length of
+                        // the chunk (including the header), then offset by the
+                        // header size to get to the beginning of the blob data.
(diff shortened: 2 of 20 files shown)


build/xtask/src/auxflash.rs
```diff
@@ -34,20 +45,26 @@ fn pack_blob(blob: &AuxFlashBlob) -> Result<tlvc_text::Piece> {
...
 /// Returns the checksum and the raw data to be saved
-pub fn build_auxflash(aux: &AuxFlash) -> Result<([u8; 32], Vec<u8>)> {
+pub fn build_auxflash(aux: &AuxFlash) -> Result<AuxFlashData> {
     let mut auxi = vec![];
+    let mut blob_checksums = BTreeMap::new();
     for f in &aux.blobs {
-        auxi.push(pack_blob(f)?);
+        let (piece, checksum) = pack_blob(f)?;
+        auxi.push(piece);
+        blob_checksums.insert(f.tag.clone(), checksum);
```
jgallagher, line 67: Tiny nit - is `.clone()` necessary? I would expect tags to be `Copy`, I think

drv/auxflash-server/src/main.rs
```diff
@@ -441,17 +441,66 @@ impl idl::InOrderAuxFlashImpl for ServerImpl {
...
+                        // the chunk (including the header), then offset by the
+                        // header size to get to the beginning of the blob data.
+                        let (_, inner_offset, _) = inner_reader.into_inner();
+                        let pos = inner_offset
+                            - inner_chunk.header().total_len_in_bytes() as u64
+                            + core::mem::size_of::<tlvc::ChunkHeader>() as u64;
+                        return Ok(AuxFlashBlob {
+                            slot: active_slot,
+                            start: pos as u32,
+                            end: (pos + inner_chunk.len()) as u32,
```
jgallagher, line 478: Just checking my understanding:

* `inner_chunk.header().total_len_in_bytes()` is the full length of the chunk (including the header)
* `inner_chunk.len()` is the length of the data in the chunk

Does `total_len_in_bytes()` include more than header+data? If not, could the calculation of `pos` above be `let pos = inner_offset - inner_chunk.len()`?

drv/fpga-api/src/lib.rs
```diff
@@ -293,14 +307,51 @@ pub fn await_fpga_ready(
...
     bitstream_type: BitstreamType,
-    chunk_len: usize,
+    expected_checksum: [u8; 32],
 ) -> Result<(), FpgaError> {
-    let mut bitstream = fpga.start_bitstream_load(bitstream_type)?;
+    // Do an initial read to verify the checksum
+    let mut scratch_buf = [0u8; 128];
+    let mut pos = blob.start;
+    let mut sha = Sha3_256::new();
+    while pos != blob.end {
```
jgallagher, line 319: Tiny nit - maybe `<` instead of `!=`? Two reasons:

1. Slightly clearer that this loop is increasing `pos` until it gets to `blob.end`
2. If we screwed up the increment of `pos` and shoot past `blob.end`, we'll break out as soon as that happens

drv/fpga-api/src/lib.rs
```diff
@@ -293,14 +307,51 @@ pub fn await_fpga_ready(
...
+        let chunk = &mut scratch_buf[0..(amount as usize)];
+
+        auxflash
+            .read_slot_with_offset(blob.slot, pos, chunk)
+            .map_err(|_| FpgaError::AuxReadError)?;
+        pos += amount;
+
+        sha.update(&chunk);
+        bitstream.continue_load(&chunk)?;
+    }
```
jgallagher, line 351: Would it be slightly more maintainable to extract the "read and hash the slot" into a separate function, since this is almost identical to the above? Would need some kind of callback to handle the bitstream load. Untested, but something like

```rust
fn read_and_hash_slot(
    auxflash: &mut AuxFlash,
    blob: &AuxFlashBlob,
    scratch_buf: &mut [u8],
    callback: &mut dyn FnMut(&[u8]) -> Result<(), FpgaError>,
) -> Result<[u8; 32], FpgaError> {
    let mut pos = blob.start;
    let mut sha = Sha3_256::new();
    while pos < blob.end {
        let amount = (blob.end - pos).min(scratch_buf.len() as u32);
        let chunk = &mut scratch_buf[0..(amount as usize)];

        auxflash
            .read_slot_with_offset(blob.slot, pos, chunk)
            .map_err(|_| FpgaError::AuxReadError)?;
        pos += amount;

        sha.update(&chunk);
        callback(&chunk)?;
    }
    Ok(sha.finalize.into())
}
```

which would change most of this function to something like

```rust
    let mut scratch_buf = [0u8; 128];

    // Do an initial read to verify the checksum
    let sha_out = read_and_hash_slot(auxflash, &blob, &mut scratch_buf, &mut |_| Ok(()))?;
    if sha_out != expected_checksum {
        return Err(FpgaError::AuxChecksumMismatch);
    }

    // Then, stream the data into the FPGA, recalculating the checksum to
    // protect against TOCTOU attacks.
    let mut bitstream = fpga.start_bitstream_load(bitstream_type)?;
    let sha_out = read_and_hash_slot(auxflash, &blob, &mut scratch_buf,
        &mut |chunk| bitstream.continue_load(chunk))?;
    if sha_out != expected_checksum {
        return Err(FpgaError::AuxChecksumMismatch);
    }
```

drv/sidecar-front-io/build.rs
```diff
@@ -32,18 +19,13 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     )?;
 
     // Calculate a bitstream checksum and add it to the generated Rust file
```
jgallagher: Nit: this comment is slightly wrong now, since we're echoing what the build system has told us what the checksum is

drv/sidecar-mainboard-controller/build.rs
```diff
@@ -3,34 +3,27 @@
...
-    let fpga_bitstream = fs::read(ecp5_bitstream_name)?;
-    let compressed_fpga_bitstream = gnarle::compress_to_vec(&fpga_bitstream);
-
-    fs::write(out_dir.join("ecp5.bin.rle"), &compressed_fpga_bitstream)?;
-
-    // Make sure the app image is rebuilt if the bitstream file for this target
-    // changes.
-    println!("cargo:rerun-if-changed={}", ecp5_bitstream_name);
-
+    // Calculate a bitstream checksum and add it to the generated Rust file
```
jgallagher: Same nit as above on this comment

drv/sidecar-seq-server/src/main.rs
```diff
@@ -292,6 +296,8 @@ fn main() -> ! {
 
     ringbuf_entry!(Trace::FpgaInit);
 
+    server.mainboard_controller.reset(); // force reinit
```
jgallagher, line 299: Is this related to the aux flash changes? I have minimal context for this task so not sure why this would be needed.

jgallagher (approved):

build/xtask/src/auxflash.rs
mkeeter: They're `String` to make `#[derive(Deserialize)]` work, so `.clone()` is needed

drv/fpga-api/src/lib.rs
mkeeter: Yeah, this is great!

drv/fpga-api/src/lib.rs
mkeeter: 👍🏻  I ended up using your abstraction from below, which uses `<`

drv/auxflash-server/src/main.rs
jgallagher: Makes sense :+1:

drv/sidecar-seq-server/src/main.rs
mkeeter: Nope, this was for debug and related to #769.  I've removed it for now; next time we update the mainboard FPGA image, we'll need the checksum stuff to be available (otherwise powered-on Sidecars won't know to update their images).

build/xtask/src/auxflash.rs
```diff
@@ -22,8 +23,18 @@ pub struct AuxFlashBlob {
     pub tag: String,
 }
 
+#[derive(Clone, Debug)]
+pub struct AuxFlashData {
+    /// Main checksum
+    pub chck: [u8; 32],
+    /// Individual blob checksums
+    pub checksums: BTreeMap<String, [u8; 32]>,
```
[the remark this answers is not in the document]
mkeeter: Same, I decided against it since at this point, it really is just raw data that we're packing into the array.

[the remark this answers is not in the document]
arjenroodselaar: It's not just raw data though, it's a checksum? Even though the compile can't act on it, I would add a typedef for the human reading this, as it'll say something about the expectations on the data.

arjenroodselaar (changes requested):
[.. moved into comment on the file ..]

drv/fpga-api/src/lib.rs
```diff
@@ -290,17 +304,51 @@ pub fn await_fpga_ready(
...
+        pos += amount;
+
+        sha.update(&chunk);
+        callback(&chunk)?;
+    }
+    Ok(sha.finalize().into())
+}
+
 /// Load a bitstream.
 pub fn load_bitstream(
```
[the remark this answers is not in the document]
mkeeter: Sure, done in 50bcc3cf71c62a7621358de00016147a22664c01

drv/sidecar-seq-server/src/main.rs
arjenroodselaar: Yeah this has been on my TODO for too long and I got bitten by this the other day (updating the FPGA bitstream but the SP not reloading it on boot).

drv/fpga-api/src/lib.rs
```diff
@@ -306,6 +322,59 @@ pub fn load_bitstream(
...
+/// Load a bitstream from auxiliary flash
+#[cfg(feature = "auxflash")]
+pub fn load_bitstream_from_auxflash(
+    fpga: &mut Fpga,
+    auxflash: &mut AuxFlash,
+    blob: AuxFlashBlob,
+    bitstream_type: BitstreamType,
+    expected_checksum: [u8; 32],
+) -> Result<(), FpgaError> {
+    // Do an initial read to verify the checksum
```
arjenroodselaar, line 359: Reading this makes me feel we should add a `bitstream.cancel_load()`, so we can load in chunks, calculating the checksum as we go and simply canceling at the end if it doesn't match. That cancel would trigger an FPGA reset to get it ready for a retry.

I think changing the the `Bitstream` type in the API like so..
```
pub struct Bitstream(FpgaLock);

impl Bitstream {
    pub fn continue_load(&mut self, data: &[u8]) -> Result<(), FpgaError> {
        (*self.0).continue_bitstream_load(data)
    }

    pub fn finish_load(&mut self) -> Result<(), FpgaError> {
        (*self.0).finish_bitstream_load()
    }

    pub fn cancel_load(&mut self) -> Result<(), FpgaError> {
        (*self.0).reset_device()
    }
}
```
.. will do what you want. I think that gets rid of the `FpgaError::AuxTocTouChecksumMismatch` variant, and upstream code can take the mismatch error and decide what to do from there.

build/xtask/src/auxflash.rs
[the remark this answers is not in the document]
mkeeter: Alright, added an alias of `AuxFlashChecksum`

arjenroodselaar (approved):
Great! This is a significant functionality lift, thanks!

PR #779, opened 2022-09-14
Preliminary host/SP comms task

jgallagher:
This is a sketch intended to aid initial development and work its way toward a real task. It speaks the draft protocol of RFD 316 (corncobs-framed messages that themselves contain hubpack-encoded data), but doesn't do anything useful in response to those requests.

The most real part of this PR is that it also extends the stm32h7 uart driver to (optionally) enable CTS/RTS, which is a very minor addition.


app/gimlet/rev-b.toml
```diff
@@ -232,7 +243,7 @@ uses = [
     "system_flash", # TODO also used by `net`, both to read the stm32 uid
 ]
 task-slots = ["jefe", "net", "update_server", "sys"]
-features = ["gimlet", "usart1", "vlan"]
+features = ["gimlet", "usart1", "vlan", "baud_rate_3_000_000", "hardware_flow_control"]
```
mkeeter, line 246: Nit: maybe `baud_rate_3M` instead?  Or does that cause compiler complaints about not being `snake_case`?

lib/host-sp-messages/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "host-sp-messages"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+fletcher = "0.3"
+serde = { version = "1.0", default-features = false, features = ["derive"] }
+
+hubpack = { git = "https://github.com/cbiffle/hubpack", rev = "df08cc3a6e1f97381cd0472ae348e310f0119e25" }
```
mkeeter, line 10: Do we need to lock this to a commit here, versus in the `Cargo.lock` file?

lib/host-sp-messages/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    Ack,
+    DecodeFailure(DecodeFailureReason),
+    BootStorageUnit(u8),
+    Identity {
+        model: u8,
+        revision: u8,
+        serial: [u8; 11], // TODO serial format?
+    },
+    Status(u64), // TODO replace u64 with bitflags type?
+    MacAddresses([[u8; 6]; 16]), // TODO send as a data blob? how fixed is 16?
```
mkeeter, line 81: Vague thoughts:

It might be cleaner to have `GetMacAddress(MacAddressType)` in the `enum HostToSp`, with a separate
```rust
enum MacAddressType {
    SpManagement1,
    SpManagement2,
    HostNic1,
    ...
}
```
(basically forcing the caller to ask for a specific device's MAC address)

lib/host-sp-messages/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+/// `out[header_length..out.len() - 2]`.
+pub fn serialize<'a>(
+    out: &mut [u8; MAX_MESSAGE_SIZE],
+    header: &Header,
+    command: &impl Serialize,
+    data_blob: &'a [u8],
+) -> Result<(usize, &'a [u8]), HubpackError> {
+    let mut n = hubpack::serialize(out, header)?;
+
+    // We know `Header::MAX_SIZE` is much smaller than `out.len()`, so we know
```
mkeeter, line 113: This comment confuses me – it seems like we actually need to know that 
```
Header::MAX_SIZE + [type of command]::MAX_SIZE + 2 <= out.len()
```

mkeeter: (and it would be helpful to mention that the two bytes is for a checksum)

task/host-sp-comms/src/main.rs
```diff
@@ -0,0 +1,314 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum UartLog {
+    Tx(u8),
+    TxFull,
+    Rx(u8),
+    RxOverrun,
+}
+
+ringbuf!(UartLog, 64, UartLog::Rx(0));
```
mkeeter, line 33: Normally, we use a `None` variant for an uninitialized `ringbuf` entry

task/host-sp-comms/src/main.rs
```diff
@@ -0,0 +1,314 @@
...
+            ringbuf_entry!(UartLog::RxOverrun);
+        }
+
+        let maybe_response = 'response: loop {
+            // Receive until we find a message delimiter. Since we're using
+            // corncobs for framing, we're looking for 0x00.
+            while let Some(byte) = uart.try_rx_pop() {
+                ringbuf_entry!(UartLog::Rx(byte));
+
+                if byte == 0x00 {
```
mkeeter, line 66: It seems like this could fail if it turns on midway through the first packet, since we'd be trying to parse a partial `rx_buf`.

We should probably wait for the first `0x00`, _then_ start accumulating data in `rx_buf` and processing it on _subsequent_ `0x00` (either tracked with a boolean flag or a tiny state machine).

task/host-sp-comms/src/main.rs
```diff
@@ -0,0 +1,314 @@
...
+        if let Some(unframed_data) = maybe_response {
+            let mut iter = corncobs::encode_iter(unframed_data).peekable();
+
+            uart.enable_tx_fifo_empty_interrupt();
+            while let Some(&b) = iter.peek() {
+                if try_tx_push(&uart, b) {
+                    // Discard the byte we just peeked and successfully inserted
+                    // into the TX FIFO.
+                    iter.next().unwrap_lite();
+                    continue;
```
mkeeter, line 99: Nit: I'd rather have the "TX fifo is full" handling in an `else` block, instead of using `continue` here

task/host-sp-comms/src/main.rs
```diff
@@ -0,0 +1,314 @@
...
+                }
+
+                // TX fifo is full; wait for space.
+                sys_irq_control(USART_IRQ, true);
+                let _ = sys_recv_closed(&mut [], USART_IRQ, TaskId::KERNEL);
+            }
```
mkeeter, line 105: Do we need to send a `0x00` byte here to frame / terminate the packet?

task/uartecho/src/main.rs
```diff
@@ -182,17 +182,37 @@ fn configure_uart_device() -> Usart {
...
 
+    #[cfg(feature = "baud_rate_115_200")]
     const BAUD_RATE: u32 = 115_200;
+    #[cfg(feature = "baud_rate_3_000_000")]
+    const BAUD_RATE: u32 = 3_000_000;
+
+    #[cfg(feature = "hardware_flow_control")]
+    let hardware_flow_control = true;
+    #[cfg(not(feature = "hardware_flow_control"))]
+    let hardware_flow_control = false;
```
mkeeter, line 193:
```suggestion
    let hardware_flow_control = cfg!(feature = "hardware_flow_control");
```
perhaps?

lib/host-sp-messages/src/lib.rs
jgallagher: I'll transport this to a comment on RFD 316 :+1:

task/host-sp-comms/src/main.rs
jgallagher: Trying to parse a partial `rx_buf` ought to be "okay" in that we'll fail to parse and send an error back to the host; failing to respond at all (in this preliminary version) seems like it would be worse. We (cc @citrus-it) still need to figure out how we notify and get sync'd with the host when we start (or restart, in event of a task panic).

task/host-sp-comms/src/main.rs
jgallagher: No, corncobs::encode_* adds the terminating `0x00` itself.

task/uartecho/src/main.rs
jgallagher: TIL!

lib/host-sp-messages/Cargo.toml
jgallagher: Need to test; I thought that if we had a bare git dependency like

```toml
hubpack = { git = "https://github.com/cbiffle/hubpack" }
```

then cargo would check for updates more or less every time it's run, but maybe that's wrong?

app/gimlet/rev-b.toml
jgallagher: I get no complaints with `baud_rate_3M` :+1:

lib/host-sp-messages/src/lib.rs
citrus-it: In the prediscussion RFD316 I have:
> Could also send just first and last address, or start and length.
The reference I used to determine the block of 16 was https://rfd.shared.oxide.computer/rfd/0250#_mac_address_selection
this will drop out when the RFD goes to discussion and we can firm it up.

mkeeter (approved):

jgallagher:
RFD 316 has had a bunch of changes; I'm going to merge this as-is and open a followup PR to catch up to those changes.

PR #781, opened 2022-09-15
Dice cert builder + artifacts for sp-measure & rng task

flihp:
This PR addresses review feedback by refactoring the certificate trait / structs using the builder pattern. Additionally we add artifact generation for the sp-measure and lpc55-rng tasks.


lib/dice/src/lib.rs
```diff
@@ -238,3 +242,38 @@ impl AliasOkm {
         Self(okm_from_seed_no_extract(cdi, "attestation".as_bytes()))
     }
 }
+
+/// SwdspOkm is a type that represents the output keying material (OKM) used
+/// to create the Swdsp key. This key is used as an embedded CA by the task
+/// that owns the SWD interface to the service processor.
```
labbott, line 248: Suggest changing this and all structure names to refer to 'measurement' vs 'swd' since SWD is just the underlying protocol and I spent some time trying to figure out why SWD needed a certificate (SWD itself doesn't need it but we want it for measuring!)

app/lpc55xpresso/stage0.toml
```diff
@@ -8,9 +8,9 @@ external-images = ["a", "b"]
 
 [kernel]
 name = "stage0"
-requires = {flash = 0x7000, ram = 8192}
+requires = {flash = 0x7000, ram = 16000}
```
labbott, line 11: I'm guessing this bump is because we have yet another certificate?

flihp: Yes. Currently I'm cloning the whole cert chain to serialize certs into each handoff block since the serialization framework requires the serialized struct to own all serialized data. This probably won't be sustainable so I'm looking for ways to reduce the data duplication. The only thing I've come up with so far is to create a separate region in the handoff block to hold the common certs (the intermediate CA and the DeviceId). I'm :100: open to alternative solutions if you've got ideas though.

lib/dice/src/lib.rs
flihp: Reasonable. Might be best to use the same name as the task as the prefix (in UpperCamelCase) : SpMeasureOkm etc.

flihp: should be resolved

lib/dice/src/handoff.rs
```diff
@@ -38,26 +60,42 @@ impl<'a> Handoff<'a> {
...
+    const MAX_SIZE: usize;
+
+    fn get_magic(&self) -> [u8; 16];
+
+    fn from_mem() -> Option<Self>
+    where
+        Self: Sized,
+        for<'d> Self: Deserialize<'d>,
+    {
+        let src = slice_from_parts(Self::MEM_START, Self::MAX_SIZE);
```
kc8apf, line 87: Nothing verifies that MAX_SIZE <= MEM_SIZE.  I'd feel better if this did something like:
`let src = slice_from_parts(Self::MEM_START, Self::MEM_SIZE)[:MAX_SIZE]`
or has a compile-time assert on MAX_SIZE <= MEM_SIZE.

lib/dice/src/lib.rs
```diff
@@ -238,3 +246,58 @@ impl AliasOkm {
         Self(okm_from_seed_no_extract(cdi, "attestation".as_bytes()))
     }
 }
+
+/// SpMeasureOkm is a type that represents the output keying material (OKM) used
+/// to create the SpMeasure key. This key is used as an embedded CA by the task
+/// that owns the SWD interface to the service processor.
```
kc8apf, line 252:
```suggestion
/// that measures the service processor.
```

lib/dice/src/handoff.rs
```diff
@@ -13,6 +16,25 @@ use serde::{Deserialize, Serialize};
 // TODO: get from app.toml -> chip.toml at build time
 const MEM_START: usize = 0x4010_0000;
 const ALIAS_START: usize = MEM_START;
+const ALIAS_SIZE: usize = 0x800;
+const SPMEASURE_START: usize = ALIAS_START + ALIAS_SIZE;
+const SPMEASURE_SIZE: usize = 0x800;
+const RNG_START: usize = SPMEASURE_START + SPMEASURE_SIZE;
+const RNG_SIZE: usize = 0x100;
+
+pub fn slice_from_parts(start: usize, size: usize) -> &'static mut [u8] {
```
kc8apf, line 25: This seems to only be used in 2 places.  I'd much rather see the safety comment in those two places.

jgallagher: This function should also be marked `unsafe` itself. Currently it claims to present a safe interface, which is not correct since it has no way to check that `start`/`size` are reasonable.

lib/dice/src/handoff.rs
flihp: I've updated the PR with static assertions in each implementer of HandoffData to do this check.

lib/dice/src/handoff.rs
flihp: This is an interesting place where my efforts to consolidate implementation into traits / functions complicates things. I went back to the rustnomicon this morning to remind myself of the details and AFAIK the whole `HandoffData` trait here is unsafe since it can't determine whether types implementing the trait have overlapping memory regions. This is going to take some refactoring.

flihp: In refactoring the handoff module I've:
* converted all const START / SIZE values to be const Range<usize>
* added static assertions to prevent overlap of these regions as well as ensuring their size is sufficient
* clarified safety comments to better align with conventions from the standard library writers guide & the rustonomicon

lib/dice/src/handoff.rs
```diff
@@ -38,26 +56,68 @@ impl<'a> Handoff<'a> {
...
+
+// Types that can be transfered through the memory region used to pass DICE
+// artifacts from stage0 to hubris tasks.
+//
+// This trait cannot check the validity of the memory range selected by
+// implementers and so implementers of this trait are required to ensure that
+// the range denoted by Self::MEM_RANGE is:
+// - within the memory range used to hold DICE artifacts
+// - non-overlapping with the ranges of memory used by other implementers of
+// this trait
```
kc8apf, line 87: - large enough to contain a the largest serialized form of the implementing type

You already are checking this in your static asserts but didn't mention it here.

kc8apf (approved):

PR #785, opened 2022-09-16
Make `cargo clippy -p xtask` clean

mkeeter:


cbiffle (approved):
Pretty good Clippy feedback, imo.

PR #787, opened 2022-09-16
Gimlet SP shouldn't stop pay attention to SP3_TO_SP_NIC_PWREN_L

### LINKED ISSUE #786: Gimlet SP shouldn't stop pay attention to SP3_TO_SP_NIC_PWREN_L
Currently the `gimlet_seq` server pays attention to the `SP3_TO_SP_NIC_PWREN_L`to go from high to low while in the A0 state. When this happens, the SP does the following:

* Transitions to the A0+HP state
* Clears the Sequencer `NIC_CTRL` bit

After the system enters A0+HP state in the sequencer, it does not pay attention to any other changes here. This is a problem as part of the manufacturing mode to mission mode transition requires us to put the NIC back into reset to allow the GPIO change to take effect. Instead, we need to always pay attention to this transition, regardless of whether we're in A0 or A0+HP. If this is deasserted (e.g. transitions high) then we need to go back and set the appropriate bit again in `NIC_CTRL`.

rmustacc:
This is an attempt to fix #786, which describes the problem. While I've tested this in conjunction with the updated sequencer from #784 and this does work, I have several questions on what is appropriate:

* Should we continue to model this as a transition back to A0 from A0+HP when this happens?
* The polling logic changes its polling frequency based on the power state assuming that all it needs to look for is thermtrip. Should this instead be decreased in A0+HP to match A0?
* I definitely implemented this the max power way, I'm not sure stylistically if you'd prefer these be a single match ARM that just changes what we look for and transition or if this should honestly be done a different way entirely.

Please feel free to take this and redo it however it makes more sense.


kc8apf (approved):
1) Modeling as a transition from A0HP -> A0 makes sense to me
2) Only consideration with timeout I can see is minimum time that CLD_RST should be asserted
3) stylistically, this looks good to me.  Even though there is a fair amount of duplication between the two cases, they are also very distinct cases in the state diagram so having them be separate arms helps illustrate that.

PR #789, opened 2022-09-16
mgmt-gateway: Add update status and update IDs

jgallagher:
This is the companion to the larger MGS change in https://github.com/oxidecomputer/omicron/pull/1714. From the SP's side,
the changes are relatively minor:

1. We removed "update prepare status" in favor of the more general "update status"
2. Update messages now always include an update ID, and for "update abort" we only abort if the ID matches our current in-progress update.

### DIFF
--- a/task/mgmt-gateway/src/mgs_common.rs
+++ b/task/mgmt-gateway/src/mgs_common.rs
@@ -124,27 +124,23 @@ impl MgsCommon {
         //
         // We only want to return an error if we have a _different_ in-progress
         // update.
-        match self.update_buf.in_progress_update_id() {
-            Some(in_progress_id) if id != in_progress_id => {
+        if let Some(in_progress_id) = self.update_buf.in_progress_update_id() {
+            if id != in_progress_id {
                 return Err(ResponseError::UpdateInProgress(
                     self.update_buf.status(),
                 ));
             }
-            _ => (),
         }
 
         match self.update_task.abort_update() {
             // Aborting an update that hasn't started yet is fine; either way
             // our caller is clear to start a new update.
-            Ok(()) | Err(UpdateError::UpdateNotStarted) => (),
-            Err(other) => {
-                return Err(ResponseError::UpdateFailed(other as u32))
+            Ok(()) | Err(UpdateError::UpdateNotStarted) => {
+                self.update_buf.abort();
+                Ok(())
             }
+            Err(other) => Err(ResponseError::UpdateFailed(other as u32)),
         }
-
-        self.update_buf.abort();
-
-        Ok(())
     }
 
     pub(crate) fn reset_prepare(&mut self) -> Result<(), ResponseError> {
--- a/task/mgmt-gateway/src/mgs_common.rs
+++ b/task/mgmt-gateway/src/mgs_common.rs
@@ -5,10 +5,10 @@
 use crate::{update_buffer::UpdateBuffer, Log, MgsMessage};
 use core::convert::Infallible;
 use drv_update_api::stm32h7::BLOCK_SIZE_BYTES;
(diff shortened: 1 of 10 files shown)


task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -764,25 +770,28 @@ impl HostFlashUpdate {
...
+
+    fn abort(&mut self, id: &UpdateId) -> Result<(), ResponseError> {
+        // We will allow the abort if either:
+        //
+        // 1. We have an in-progress update that matches `id`
+        // 2. We do not have an in-progress update
+        //
+        // We only want to return an error if we have a _different_ in-progress
+        // update.
+        match self.buf.in_progress_update_id() {
```
mkeeter, line 785: Same suggestion as above (using a conditional + `matches!` instead of a `match`)

task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -674,6 +674,34 @@ impl HostFlashUpdate {
...
+                    // If we're still erasing sectors, we shouldn't have
+                    // ingested any chunks yet.
+                    assert!(sub_status.bytes_received == 0);
+
+                    sector_erase.progress().map(|progress| {
+                        UpdateStatus::Preparing(UpdatePreparationStatus {
+                            id: sub_status.id,
+                            progress: Some(progress),
+                        })
+                    })
```
mkeeter, line 694: Weakly-held opinion: since this function returns a `Result`, early-exiting may be cleaner than `map`, e.g.
```suggestion
                    let progress = sector_erase.progress()?;
                    UpdateStatus::Preparing(UpdatePreparationStatus {
                            id: sub_status.id,
                            progress: Some(progress),
                    })
```

task/mgmt-gateway/src/mgs_common.rs
```diff
@@ -94,46 +94,55 @@ impl MgsCommon {
...
+        }
+
+        match self.update_task.abort_update() {
+            // Aborting an update that hasn't started yet is fine; either way
+            // our caller is clear to start a new update.
+            Ok(()) | Err(UpdateError::UpdateNotStarted) => (),
+            Err(other) => {
+                return Err(ResponseError::UpdateFailed(other as u32))
+            }
+        }
```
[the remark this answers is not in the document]
jgallagher: I'm not sure what syntax should work since `let_chains` are still unstable (?!). What if I rolled the rest of the method into the successful branch:

```rust
        match self.update_task.abort_update() {
            // Aborting an update that hasn't started yet is fine; either way
            // our caller is clear to start a new update.
            Ok(()) | Err(UpdateError::UpdateNotStarted) => {
                self.update_buf.abort();
                Ok(())
            }
            Err(other) => {
                Err(ResponseError::UpdateFailed(other as u32))
            }
        }
```

task/mgmt-gateway/src/mgs_common.rs
```diff
@@ -94,46 +94,55 @@ impl MgsCommon {
...
+        // We only want to return an error if we have a _different_ in-progress
+        // update.
+        match self.update_buf.in_progress_update_id() {
+            Some(in_progress_id) if id != in_progress_id => {
+                return Err(ResponseError::UpdateInProgress(
+                    self.update_buf.status(),
+                ));
+            }
+            _ => (),
+        }
```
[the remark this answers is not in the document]
mkeeter: Totally fine by me!

task/mgmt-gateway/src/mgs_gimlet.rs
jgallagher: Yes, that's much better, thanks.

PR #790, opened 2022-09-19
Add support for PSC rev B

mkeeter:


app/psc/rev-b.toml
```diff
@@ -0,0 +1,356 @@
...
+
+[tasks.spi2_driver.config.spi]
+global_config = "spi2"
+
+[tasks.i2c_driver]
+name = "drv-stm32xx-i2c-server"
+features = ["h753", "itm"]
+priority = 2
+max-sizes = {flash = 16384, ram = 2048}
+uses = ["i2c2", "i2c3", "i2c4"]
```
rmustacc, line 76: Is this right? I only see two i2c ports on the SP being used:

* I2C_TO_SP_LOCAL
* I2C_TP_TO_PS

It doesn't appear I2C4 is actually used.

app/psc/rev-b.toml
```diff
@@ -0,0 +1,356 @@
...
+address = 0b1010_000
+device = "at24csw080"
+description = "FRU ID EEPROM"
+
+[[config.i2c.devices]]
+bus = "backplane"
+name = "psu1eeprom"
+address = 0b1010_000
+device = "psueeprom"
+description = "PSU 1 EEPROM"
```
rmustacc, line 220: Per RFD 200, rectifiers should be labeled starting at 0. I think we want to use our labeling as opposed to Murata's doc-only 1 based.

rmustacc (commented):
Thanks for doing this, Matt.

app/psc/rev-b.toml
isobering: Correct, there are only two I2C ports on the SP being used. The third port was previously used to talk to some I2C GPIO expanders, which we deleted.

mkeeter: 👍🏻  I deleted it lower down, but forgot to remove it from the driver task; fixed!

app/psc/rev-b.toml
rmustacc: There are other places where schematics and related diverge from what we ultimately settled on for nomenclature. I'm not sure if it's worth the churn in the Rev C schematic or not, I'll leave that up to folks. I think the important thing is that when the SP software reports up to the control plane we use the actual identifiers that we're labeling the system with and following those conventions. If that means we have a delta from the schematic to that (which we have in Gimlet due to changes in how we were labeling PCIe slots), then that's something we can probably live with.

mkeeter: Okay, I've updated both rev A and B files to use zero-indexing, with a note explaining that the schematic is different.

PR #791, opened 2022-09-19
Make `sidecar/rev-a.toml` clippy-clean

mkeeter:
Clippy doesn't want us to construct needless references, and _really_ wants you to derive `Eq` if you're deriving `PartialEq`.

### DIFF
--- a/drv/vsc85xx/src/atom.rs
+++ b/drv/vsc85xx/src/atom.rs
@@ -8,7 +8,9 @@ use vsc7448_pac::phy;
 use vsc_err::VscError;
 
 /// Based on `vtss_atom_patch_suspend` in the SDK
-pub fn atom_patch_suspend<P: PhyRw>(phy: &mut Phy<P>) -> Result<(), VscError> {
+pub fn atom_patch_suspend<P: PhyRw>(
+    phy: &mut Phy<'_, P>,
+) -> Result<(), VscError> {
     // We don't have VeriPHY running, so skip the first conditional
     let v = phy.read(phy::GPIO::MICRO_PAGE())?;
     if (v.0 & 0x4000) == 0 {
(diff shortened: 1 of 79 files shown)


drv/vsc85xx/src/atom.rs
```diff
@@ -8,9 +8,7 @@ use vsc7448_pac::phy;
 use vsc_err::VscError;
 
 /// Based on `vtss_atom_patch_suspend` in the SDK
-pub fn atom_patch_suspend<'a, 'b, P: PhyRw>(
-    phy: &'b mut Phy<'a, P>,
-) -> Result<(), VscError> {
+pub fn atom_patch_suspend<P: PhyRw>(phy: &mut Phy<P>) -> Result<(), VscError> {
```
cbiffle, line 11: I think you want `Phy<'_, P>` here, and probably also to set `deny(elided_lifetimes_in_paths)` at the crate root. I wish there was a way we could set this at the workspace level, but, there does not appear to be.

drv/vsc85xx/src/atom.rs
```diff
@@ -23,9 +21,7 @@ pub fn atom_patch_suspend<'a, 'b, P: PhyRw>(
 }
 
 /// Based on `vtss_atom_patch_suspend` in the SDK
-pub fn atom_patch_resume<'a, 'b, P: PhyRw>(
-    phy: &'b mut Phy<'a, P>,
-) -> Result<(), VscError> {
+pub fn atom_patch_resume<P: PhyRw>(phy: &mut Phy<P>) -> Result<(), VscError> {
```
cbiffle, line 24: (same)

cbiffle (approved):
Please fix the two elided lifetime cases I've identified in the comments below, but otherwise looks good.

drv/vsc85xx/src/atom.rs
mkeeter: I ended up turning it on as a Clippy lint (in `cargo xtask clippy`), then fixed it _everywhere_!

PR #793, opened 2022-09-20
Thermal loop updates

mkeeter:
This roughly implements the design from [RFD 276](https://github.com/oxidecomputer/rfd/tree/0276/rfd/0276).  I'd like to get feedback on the code before going through and tuning everything, so parameters (PID coefficients, thermal properties) are all made up right now.


task/thermal/src/control.rs
```diff
@@ -116,84 +142,216 @@ impl InputChannel {
...
 
+    /// Once we're in `Overheated`, how much does the temperature have to drop
+    /// by before we return to `Normal`
+    overheat_hysteresis: Celsius,
+
+    /// Most recent power mode mask
+    power_mode: u32,
+
+    /// Number of sensor reads which failed
     read_failed_count: u32,
```
lzrd, line 170: I haven't read where this is used yet. Given that I'm speculating on the implementation, maybe a sensor is marked as being mandatory for particular power states. A count here makes me think about there being a threshold, but this could just be a statistic.

task/thermal/src/control.rs
```diff
@@ -116,84 +142,216 @@ impl InputChannel {
...
+        cfg: &PidConfig,
+        error: f32,
+        limit: core::ops::Range<f32>,
+    ) -> f32 {
+        let p_contribution = cfg.gain_p * error;
+
+        // Only accumulate the integral term if we're not saturated, to prevent
+        // integral wind-up.
+        if limit.contains(&p_contribution) {
+            self.integral += error;
```
lzrd, line 217: Guarding against overflow/underflow? This is a float, so maybe we're always safe.

task/thermal/src/control.rs
```diff
@@ -116,84 +142,216 @@ impl InputChannel {
...
+    },
+
+    /// The system cannot control the temperature; power down and wait for
+    /// intervention from higher up the stack.
+    Uncontrollable,
+}
+
+enum ControlResult {
+    Pwm(PWMDuty),
+    PowerDown,
```
lzrd, line 277: A question from earlier, will we ever need an intermediate state "CoolDown" if we're overheated and are shutting down?

task/thermal/src/control.rs
```diff
@@ -259,51 +425,184 @@ impl<'a, B: BspT> ThermalControl<'a, B> {
...
+                    match v {
+                        Some(TemperatureReading::Valid { value, time_ms }) => {
+                            // Model the current temperature based on the last
+                            // reading and the worst-case slew rate.  This only
+                            // matters when samples are dropped; if we received
+                            // a reading on this control cycle, then time_ms ==
+                            // now_ms, so this becomes v.value (i.e. the most
+                            // recent reading).
+                            let temperature = value.0
+                                + (time_ms - now_ms) as f32
```
lzrd, line 453: nice.
Does this need to be capped to the sensor's maximum value?
We will want reports from thermal management on what outliers are causing fan pwm settings.

task/thermal/src/main.rs
```diff
@@ -175,38 +185,33 @@ impl<'a, B: BspT> idl::InOrderThermalImpl for ServerImpl<'a, B> {
...
+            match self.mode {
+                ThermalMode::Auto => {
+                    self.control.run_control(now);
+                }
+                ThermalMode::Manual => {
+                    // Read sensors and post them to the `sensors` task
+                    self.control.read_sensors(now);
+                }
+                ThermalMode::Off => {
+                    panic!("Mode must not be 'Off' when server is running")
```
lzrd, line 209: How would this panic affect the fans on the running system?
Are things ever going to be out of sync so that this condition needs to be tolerated? (e.g. engineer manually sets power state, thermal loop restarts due to panic?)

lzrd (approved):
Someday we may want different sensors sampled on different schedules. No need to think about that until there is data to support the effort though.

task/thermal/src/main.rs
mkeeter: I think this should be safe: `ThermalMode::Off` is only used during the brief period between constructing the `struct` and calling `set_mode_manual` on startup.

We only expose APIs to set the mode to manual or auto, so no one should be able to set it to `ThermalMode::Off` during operation, even during unusual debug circumstances.

task/thermal/src/control.rs
mkeeter: I'm going to remove `read_failed_count` and `post_failed_count`, since they aren't actually _read_ anywhere.

For now, we'll log read and post errors to the ringbuf; eventually, we'll need to come up with a way to ship that data to MGS / Nexus.

task/thermal/src/bsp/gimlet_b.rs
```diff
@@ -27,53 +25,56 @@ use i2c_config::sensors;
 
 task_slot!(SEQ, gimlet_seq);
 
+// We monitor the TMP117 air temperature sensors, but don't use them as part of
+// the control loop.
 const NUM_TEMPERATURE_SENSORS: usize = sensors::NUM_TMP117_TEMPERATURE_SENSORS;
-const NUM_TEMPERATURE_INPUTS: usize = sensors::NUM_SBTSI_TEMPERATURE_SENSORS
+
+// The control loop is driven by CPU, NIC, and DIMM temperatures
```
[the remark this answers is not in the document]
mkeeter: I'm working through temperature ranges in [RFD 276](https://github.com/oxidecomputer/rfd/tree/0276/rfd/0276#machine-specific-implementations), but there are still a bunch of blank spaces.

task/thermal/src/bsp/gimlet_b.rs
```diff
@@ -107,22 +107,47 @@ impl BspT for Bsp {
...
-        const MAX_DIMM_TEMP: Celsius = Celsius(60f32);
-        const MAX_CPU_TEMP: Celsius = Celsius(60f32);
-        const MAX_T6_TEMP: Celsius = Celsius(60f32);
+        // TODO these numbers are all made up
+        const DIMM_THERMALS: ThermalProperties = ThermalProperties {
+            target_temperature: Celsius(60f32),
+            critical_temperature: Celsius(70f32),
+            power_down_temperature: Celsius(80f32),
+            temperature_slew_deg_per_msec: 0.5 / 1000.0,
+        };
```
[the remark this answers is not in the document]
mkeeter: I think the JEDEC specs should be consistent (as long as we're consistently buying industrial temperature grade), but the slew rate could vary.  This is deliberately very conservative (high), so I think it's okay.

task/thermal/src/bsp/gimlet_b.rs
```diff
@@ -107,22 +107,47 @@ impl BspT for Bsp {
         // Handle for the sequencer task, which we check for power state
         let seq = Sequencer::from(SEQ.get_task_id());
 
-        const MAX_DIMM_TEMP: Celsius = Celsius(60f32);
-        const MAX_CPU_TEMP: Celsius = Celsius(60f32);
-        const MAX_T6_TEMP: Celsius = Celsius(60f32);
+        // TODO these numbers are all made up
```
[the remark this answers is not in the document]
mkeeter: That's for DDR5, rather than DDR4.  I added a note to look at RFD 276 for the details, and filled in more datasheets there.

task/thermal/src/control.rs
mkeeter: I've dramatically reworked this section, so it now clamps in a different way.

task/thermal/src/control.rs
mkeeter: I don't think so.  The `Overheating` state is meant to be emergency cooling; `PowerDown` is if all else has failed and it cuts power immediately.

task/thermal/src/control.rs
mkeeter: I don't think we want to cap this to the sensor's maximum value: if the sensor's maximum value is 127°, we have a reading of 120° ten seconds ago, and the temperature could slew by 1°/sec, we want to know that the worst-case temperature is 130°, instead of clamping to 127°.

(Kinda a moot point, because we'll be in one of the fail-safe modes well before then!)

PR #795, opened 2022-09-20
host-sp-comms: track RFD 316 changes

jgallagher:
The major changes here are, all of which are in the interest of matching the current draft of RFD 316:

1. Increase `MAX_MESSAGE_SIZE` 
2. Add / reorder some of the variants in the `HostToSp`/`SpToHost` enums
3. Add `magic` as the first field of the header

With the task as of this PR, we're able to do some very preliminary communications on a gimlet (talking to @citrus-it's work on the host side):

```
-----------> Sending command 0x4, attempt 1/10
     OUT: cc 19 de 01 01 00 00 00 01 00 00 00 00 00 00 00 04 cb 62
COBS OUT: 06 cc 19 de 01 01 01 01 02 01 01 01 01 01 01 01 04 04 cb 62
Packet sent, waiting for response...
 COBS IN: 06 cc 19 de 01 01 01 01 02 01 01 01 01 01 01 12 80 04 01 02 66 61 6b 65 2d 73 65 72 69 61 6c 97 af
      IN: cc 19 de 01 01 00 00 00 01 00 00 00 00 00 00 80 04 01 02 66 61 6b 65 2d 73 65 72 69 61 6c 97 af
Additional data length: 0xd
 DATA IN: 01 02 66 61 6b 65 2d 73 65 72 69 61 6c
../../oxide/os/fakebop.c:697: ipcc: retrieved serial number from SP - 'fake-serial'
```


lib/host-sp-messages/src/lib.rs
```diff
@@ -41,7 +51,11 @@ pub enum HostToSp {
     // Microoptimization: insert a dummy variant first, so we never serialize a
     // command value of `0` to make COBS's life slightly easier.
     _Unused,
+    RequestReboot,
```
mkeeter, line 54: Should we either (a) explicitly set values here or (b) use a shared crate with the host side?

lib/host-sp-messages/src/lib.rs
```diff
@@ -73,15 +91,46 @@ pub enum SpToHost {
...
         model: u8,
         revision: u8,
         serial: [u8; 11], // TODO serial format?
     },
-    Status(u64), // TODO replace u64 with bitflags type?
-    MacAddresses([[u8; 6]; 16]), // TODO send as a data blob? how fixed is 16?
-    RotResponse, // Followed by a binary data blob (the response)
+    MacAddresses {
+        base: [u8; 6],
+        count: u8,
```
mkeeter, line 102: It seems like it's up in the air whether `count` will be a `u8` or a `u16`, so maybe leave a note here to that effect?

lib/host-sp-messages/src/lib.rs
```diff
@@ -91,11 +140,21 @@ pub enum DecodeFailureReason {
     // Microoptimization: insert a dummy variant first, so we never serialize a
     // command value of `0` to make COBS's life slightly easier.
     _Unused,
-    CobsError,
-    HubpackError,
-    CrcFailure,
+    Cobs,
```
mkeeter, line 143: Same note here: should we explicitly lock these to numerical values?

lib/host-sp-messages/src/lib.rs
jgallagher: The host side is being written in C, which rules out (b). :-D

Explicitly listing values that can be cross referenced with RFD 316 / the C header is a good idea though, I'll do that.

mkeeter (approved):

lib/host-sp-messages/src/lib.rs
jgallagher: Added in d4e84d9f

lib/host-sp-messages/src/lib.rs
jgallagher: Noted in 1da97c8e

lib/host-sp-messages/src/lib.rs
jgallagher: Whoops, can't actually do that because some of these variants have associated data. I added a block comment and unit tests for both `HostToSp` and `SpToHost` in ebe90e9a instead.

PR #796, opened 2022-09-21
host-sp-comms: implement rebooting the host

jgallagher:
We reboot the host in response to a reboot request, and we power off the host in response to a power off request. In both cases, we do _not_ send a response on the uart.

The diff looks larger than it actually is because some things moved around and `git diff` doesn't handle it particularly well; it might be clearer to step through the commits individually.

This builds on #795 and should be merged after it.

A general comment - I'm not in love with the way the sequencer presents `PowerState` (and in particular the additive states beyond A0/A2). I asked about this in the hubris channel some, and am curious if we'd be better served if we modeled it more like

```rust
enum PowerState {
    A0(A0AdditiveState),
    A1,
    A2(A2AdditiveState),
}
```

where `AxAdditiveState` could be an enum or a bit mask? The implementation of `power_off_host()` on this PR is where this is most obvious. What's not obvious is that given the current gimlet-seq implementation, I believe `gimlet_seq.set_state()` will always fail if the current state is one of the A2 additive states. (Raw `PowerState::A2` is fine.)

### DIFF
--- a/drv/mock-gimlet-seq-server/src/main.rs
+++ b/drv/mock-gimlet-seq-server/src/main.rs
@@ -17,11 +17,7 @@ userlib::task_slot!(JEFE, jefe);
 #[export_name = "main"]
 fn main() -> ! {
     let mut buffer = [0; idl::INCOMING_SIZE];
-    let mut server = ServerImpl {
-        jefe: Jefe::from(JEFE.get_task_id()),
-    };
-
-    server.set_state_impl(PowerState::A2);
+    let mut server = ServerImpl::init(Jefe::from(JEFE.get_task_id()));
 
     loop {
         idol_runtime::dispatch(&mut buffer, &mut server);
@@ -33,6 +29,12 @@ struct ServerImpl {
 }
 
 impl ServerImpl {
+    fn init(jefe: Jefe) -> Self {
+        let me = Self { jefe };
+        me.set_state_impl(PowerState::A2);
+        me
+    }
+
     fn get_state_impl(&self) -> PowerState {
         // Only we should be setting the state, and we set it to A2 on startup;
         // this conversion should never fail.
--- a/drv/mock-gimlet-seq-server/src/main.rs
+++ b/drv/mock-gimlet-seq-server/src/main.rs
@@ -9,14 +9,19 @@
 
 use drv_gimlet_seq_api::{PowerState, SeqError};
 use idol_runtime::RequestError;
-use userlib::RecvMessage;
+use task_jefe_api::Jefe;
+use userlib::{FromPrimitive, RecvMessage, UnwrapLite};
+
+userlib::task_slot!(JEFE, jefe);
 
 #[export_name = "main"]
 fn main() -> ! {
     let mut buffer = [0; idl::INCOMING_SIZE];
     let mut server = ServerImpl {
-        state: PowerState::A2,
+        jefe: Jefe::from(JEFE.get_task_id()),
     };
+
+    server.set_state_impl(PowerState::A2);
 
     loop {
         idol_runtime::dispatch(&mut buffer, &mut server);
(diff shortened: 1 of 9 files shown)


drv/mock-gimlet-seq-server/src/main.rs
```diff
@@ -9,43 +9,60 @@
...
+}
+
+impl ServerImpl {
+    fn get_state_impl(&self) -> PowerState {
+        // Only we should be setting the state, and we set it to A2 on startup;
+        // this conversion should never fail.
+        PowerState::from_u32(self.jefe.get_state()).unwrap_lite()
+    }
+
+    fn set_state_impl(&self, state: PowerState) {
```
mkeeter, line 42: Nit: if we always want to set the state on startup, should we make `ServerImpl::new(jefe: Jefe)` do it?

jgallagher: I actually had it that way originally, then thought it was weird for `::new()` to have a side effect like that. Does that seem okay to you, or maybe there's a better name than `::new()`?

task/host-sp-comms/src/main.rs
```diff
@@ -114,45 +255,197 @@ impl ServerImpl {
...
+}
+
+// approximately idol_runtime::NotificationHandler, in anticipation of
+// eventually having an idol interface (at least for mgmt-gateway to give us
+// host phase 2 data)
+impl ServerImpl {
+    fn handle_notification(&mut self, bits: u32) {
+        if bits & USART_IRQ != 0 {
+            self.handle_usart_notification();
+            sys_irq_control(USART_IRQ, true);
```
mkeeter, line 423: Do we need to call this during startup to receive the initial interrupt?

task/host-sp-comms/src/main.rs
```diff
@@ -49,59 +64,185 @@ fn main() -> ! {
...
+                            continue;
+                        }
+                    } else {
+                        // Transitioning to A2 failed because we're already in
+                        // A2 - host is powered off.
+                        return;
+                    }
+                }
+                PowerState::A1 => {
+                    // A1 should be transitive; sleep then retry.
```
mkeeter, line 161: "transitory", perhaps?

task/host-sp-comms/src/main.rs
jgallagher: It was already there, just isn't showing in the diff (line 62).

task/host-sp-comms/src/main.rs
jgallagher: Hah, yes.

task/host-sp-comms/src/main.rs
```diff
@@ -49,59 +64,185 @@ fn main() -> ! {
...
-    fn handle_notification(&mut self, _bits: u32) {
-        // Clear any RX overrun errors. If we hit this, we will likely fail to
-        // decode the next message from the host, which will cause us to send a
-        // `DecodeFailure` response.
-        if self.uart.check_and_clear_rx_overrun() {
-            ringbuf_entry!(UartLog::RxOverrun);
+    // TODO is error handling in this method correct? I think we should
+    // basically only ever succeed in our initial set_state() request, so I
+    // don't know how we'd test it
+    fn power_off_host(&mut self, reboot: bool) {
```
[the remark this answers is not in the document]
mkeeter: (and in the first state, it's dependent on Jefe's state-change IRQ to turn stuff back on, which also warrants an explanation!)

mkeeter (commented):
One general question: when rebooting the host, do we have to hold in A2 for some amount of time, e.g. to let capacitors discharge?

task/host-sp-comms/src/main.rs
mkeeter: thanks, I was confused how this was working without it!

drv/mock-gimlet-seq-server/src/main.rs
mkeeter: Maybe `::init()` instead?  I normally associate that with more side effects.

task/host-sp-comms/src/main.rs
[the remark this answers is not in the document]
jgallagher: docstring added with the logic rework to sleep for a bit while in A2 in d44856d

mkeeter (approved):

drv/mock-gimlet-seq-server/src/main.rs
jgallagher: Sounds good; changed in c55fdaf

PR #804, opened 2022-09-26
simplify duplicate job removal

steveklabnik:
We would get duplicate builds for PRs because we want to check on master, but also on pull requests. Due to using 'push' for master, we'd end up with these duplicate jobs, and used this action to sort out the duplicates.

This change simplifies this by only doing the 'push' actions on master, and not on PR branches.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -5,24 +5,12 @@
 # * checkout PR head commit https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit
 name: documentation
 on:
+  pull_request:
   push:
-    branches:
-      - master
+    branches: [master]
+
 jobs:
-  skip_duplicate_jobs:
-    runs-on: ubuntu-latest
-    outputs:
-      should_skip: ${{ steps.skip_check.outputs.should_skip }}
-    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/documentation.yml
```diff
@@ -5,24 +5,12 @@
 # * checkout PR head commit https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit
 name: documentation
 on:
+  pull_request:
   push:
-    branches:
-      - master
+    branches: [master]
```
steveklabnik, line 10: we were even already doing this here! I changed the formatting to be identical across all of the workflows; I think this syntax is a bit nicer than the previous one.

.github/workflows/formatting.yml
```diff
@@ -1,5 +1,9 @@
 name: formatting
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches: [master]
```
steveklabnik, line 5: We were duplicating this job too, but since it's so quick nobody noticed.

steveklabnik:
Merging this after we discussed it at the huddle yesterday; if this causes problems for some reason I can revert/fix quickly, but I don't anticipate any issues.

PR #806, opened 2022-09-26
Use FRU ID EEPROM for MAC address

mkeeter:
This is pending deciding _exactly_ how we're going to store the MAC addresses in the EEPROM, but here's a first pass.

### DIFF
--- a/drv/local-vpd/build.rs
+++ b/drv/local-vpd/build.rs
@@ -39,6 +39,7 @@ pub fn get_vpd_eeprom(i2c_task: userlib::TaskId)
     assert_eq!(devs.len(), 1);
     drv_i2c_devices::at24csw080::At24Csw080::new(devs[0])
 }}
+include!(concat!(env!("OUT_DIR"), "/i2c_config.rs"));
 "#,
         cfg.vpd_bus
     )?;
--- a/drv/local-vpd/build.rs
+++ b/drv/local-vpd/build.rs
@@ -1,0 +1,47 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use serde::Deserialize;
+use std::io::Write;
+
+/// This represents our _subset_ of global config and _must not_ be marked with
+/// `deny_unknown_fields`!
+#[derive(Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct GlobalConfig {
+    pub local_vpd: LocalVpdConfig,
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub struct LocalVpdConfig {
+    /// Sockets known to the system, indexed by name.
+    pub vpd_bus: String,
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let cfg = build_util::config::<GlobalConfig>()?.local_vpd;
+
+    build_i2c::codegen(build_i2c::Disposition::Devices)?;
+
+    let out_dir = std::env::var("OUT_DIR")?;
+    let dest_path = std::path::Path::new(&out_dir).join("vpd_config.rs");
+    let mut out = std::fs::File::create(&dest_path)?;
+
+    write!(
+        out,
+        r#"
+include!(concat!(env!("OUT_DIR"), "/i2c_config.rs"));
+pub fn get_vpd_eeprom(i2c_task: userlib::TaskId)
+    -> drv_i2c_devices::at24csw080::At24Csw080
+{{
+    let devs = i2c_config::devices::at24csw080_{}(i2c_task);
+    assert_eq!(devs.len(), 1);
+    drv_i2c_devices::at24csw080::At24Csw080::new(devs[0])
+}}
+"#,
+        cfg.vpd_bus
+    )?;
+    Ok(())
+}
--- a/drv/local-vpd/build.rs
+++ b/drv/local-vpd/build.rs
@@ -32,7 +32,6 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     write!(
         out,
         r#"
-include!(concat!(env!("OUT_DIR"), "/i2c_config.rs"));
 pub fn get_vpd_eeprom(i2c_task: userlib::TaskId)
     -> drv_i2c_devices::at24csw080::At24Csw080
 {{
--- a/drv/local-vpd/build.rs
+++ b/drv/local-vpd/build.rs
@@ -25,7 +25,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     build_i2c::codegen(build_i2c::Disposition::Devices)?;
 
-    let out_dir = std::env::var("OUT_DIR")?;
+    let out_dir = build_util::out_dir();
     let dest_path = std::path::Path::new(&out_dir).join("vpd_config.rs");
     let mut out = std::fs::File::create(&dest_path)?;
 
--- a/drv/local-vpd/build.rs
+++ b/drv/local-vpd/build.rs
@@ -2,46 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use serde::Deserialize;
-use std::io::Write;
-
-/// This represents our _subset_ of global config and _must not_ be marked with
-/// `deny_unknown_fields`!
-#[derive(Deserialize)]
-#[serde(rename_all = "kebab-case")]
-pub struct GlobalConfig {
-    pub local_vpd: LocalVpdConfig,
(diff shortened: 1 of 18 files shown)


task/net/src/server_vlan.rs
```diff
@@ -202,10 +208,23 @@ impl<'a> ServerImpl<'a> {
...
+            // configuration block, so that each VLAN has a unique address.
+            //
+            // We take advantage of the fact that the top 3 octets are the OUI,
+            // and convert into a u32 to do the incrementing.
+            let next_mac = (u32::from_be_bytes(mac[2..].try_into().unwrap())
+                & 0xFFFFFF)
+                + mac_address_block.stride as u32;
+
+            // Per https://github.com/oxidecomputer/oana/#mac-addresses, we
+            // reserve `F0:00:00` and above for software stuff.
```
rmustacc, line 221: This is only specific to the current OUI, this shouldn't be universal I think. If we get another block, it'd be totally fine to use that range.

task/net/src/server_vlan.rs
```diff
@@ -202,10 +208,23 @@ impl<'a> ServerImpl<'a> {
...
-            ipv6_addr.0[15] += 1;
-            mac.0[5] += 1;
+            // Increment the MAC and IP addresses based on the stride in the
+            // configuration block, so that each VLAN has a unique address.
+            //
+            // We take advantage of the fact that the top 3 octets are the OUI,
+            // and convert into a u32 to do the incrementing.
+            let next_mac = (u32::from_be_bytes(mac[2..].try_into().unwrap())
+                & 0xFFFFFF)
+                + mac_address_block.stride as u32;
```
rmustacc, line 218: The comment and code disagree here. You're including a byte from the OUI. If the API requires four bytes and that's a constraint, then make that clear in the comment as otherwise it's confusing and I'd ask why is it being included and masked off.

drv/local-vpd/src/lib.rs
```diff
@@ -0,0 +1,75 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
rmustacc, line 1: A top-level block comment about what this is for or why it exists would be nice. But I can appreciate that may not be the hubris style. Just coming to this as an outsider it's not really clear what this is or isn't for.

Are we going to have a copy of this driver for each different i2c source in a board based on the build.rs?

What makes something "local"? I guess that it's on the same physical board as the SP and not a FRU like the fan modules or sharkfins?

task/net/src/main.rs
```diff
@@ -77,9 +91,33 @@ fn mac_address() -> &'static [u8; 6] {
...
+    // use the MAC address from the VPD in production.
+    buf[3] &= 0xEF;
+
+    MacAddressBlock {
+        base_mac: buf,
+        #[cfg(feature = "vlan")]
+        count: U16::new(crate::generated::VLAN_COUNT.try_into().unwrap()),
+
+        #[cfg(not(feature = "vlan"))]
+        count: U16::new(1),
```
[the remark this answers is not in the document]
mkeeter: This is the non-VLAN case (`#[cfg(not(feature = "vlan"))]`), which only supports a single address.

drv/local-vpd/src/lib.rs
mkeeter: Added a top-level comment explaining that this is the FRU ID EEPROM soldered to the board.

> Are we going to have a copy of this driver for each different i2c source in a board based on the build.rs?

I'm not sure what this means – each task that wants to read from the FRU ID will include this driver (as a library) and use it to extract the tags that the task cares about.

drv/local-vpd/build.rs
```diff
@@ -0,0 +1,47 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use serde::Deserialize;
+use std::io::Write;
+
+/// This represents our _subset_ of global config and _must not_ be marked with
```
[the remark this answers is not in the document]
mkeeter: Yeah, this is a pretty common Hubris-ism

task/net/src/main.rs
```diff
@@ -77,9 +91,33 @@ fn mac_address() -> &'static [u8; 6] {
...
+        stride: 1,
+    }
+}
+
+#[cfg(feature = "vpd-mac")]
+fn mac_address_from_vpd() -> Option<MacAddressBlock> {
+    let i2c_task = I2C.get_task_id();
+    drv_local_vpd::read_config(i2c_task, *b"MACS").ok()
```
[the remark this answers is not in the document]
mkeeter: Yup, this predated that decision – fixed!

task/net/src/main.rs
```diff
@@ -77,9 +91,33 @@ fn mac_address() -> &'static [u8; 6] {
 
     // Set the lower 32-bits based on the hashed UID
     buf[2..].copy_from_slice(&hash.to_be_bytes());
-    buf
+
+    // Mask the highest non-OUI octet, since we reserve `F0:00:00` and above for
```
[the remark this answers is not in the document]
mkeeter: Got it, that seems reasonable to me.

task/net/src/server_vlan.rs
mkeeter: Ah, it's a little funky: we need a `[u8; 4]` to call `u32::from_be_bytes`, so we include one byte of the OUI then mask it out.  I added a comment to that effect, because this confused me as well upon revisiting the code!

task/net/src/server_vlan.rs
mkeeter: Interesting – do you think I should hard-code the OUI, e.g.
```rust
            const OXIDE_OUI: [u8; 3] = [0xa8, 0x40, 0x25];
            if mac[..3] == OXIDE_OUI && next_mac > 0xEFFFFF {
                panic!("MAC overflow: {:?}", mac);
            }
```
or should I check the U/L bit and assume that we want this masking if it's a universally administered address?

drv/local-vpd/build.rs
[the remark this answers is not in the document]
rmustacc: OK. That's good to know. I'll resolve this, but I'll admit it's still strange.

task/net/src/main.rs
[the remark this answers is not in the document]
rmustacc: Sorry, I see now how I misread that. My apologies.

drv/local-vpd/src/lib.rs
rmustacc: > Added a top-level comment explaining that this is the FRU ID EEPROM soldered to the board.

Thank you. I guess it's a little weird that we'd rebuild this and use the same driver as a library in everything and are calling it local when the other's aren't local, but that's probably just me.

> > Are we going to have a copy of this driver for each different i2c source in a board based on the build.rs?
> 
> I'm not sure what this means – each task that wants to read from the FRU ID will include this driver (as a library) and use it to extract the tags that the task cares about.

So I was trying to sort through how this worked in practice. When the SP ultimately wants to do the same thing with the 12 different FRUID ROMs that we have, it sounds like we'll build this all (not just the build.rs) 12 different times even though the only difference is generally just which i2c to scan for the FRU block for and get the corresponding tagged data out of. Not knowing more about the overall design of the system that feels surprising, but if that's the general pattern then it is.

task/net/src/server_vlan.rs
rmustacc: Thanks, that looks good.

task/net/src/server_vlan.rs
rmustacc: Yeah, I think the hard coding approach makes sense. If we buy a second one or use something else for whatever reason then we won't want to apply that behavior. What you have looks good.

jclulow:
So I gave this a spin in the lab:

```
$ cat gimlet.ron
[
    ("FRU0", [
        ("MAC0", [
            [0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0x00, 0x08, 0x00, 0x08],
        ]),
    ]),
]

$ pfexec humility vpd -d 'Gimlet' -w gimlet.ron
humility: successfully wrote 44 bytes of VPD

$ pfexec humility vpd -d 'Gimlet' -r
[
    ("FRU0", [
        ("MAC0", [
            [
                170,
                187,
                204,
                221,
                238,
                0,
                8,
                0,
                8,
            ],
        ]),
    ]),
]

$ pfexec humility net ip
MAC address:  aa:bb:cc:dd:ee:00
IPv6 address: fe80::a8bb:ccff:fedd:ee00
```

Seems to work for me!

jclulow:
For completeness I tried this input VPD file as well:

```
[
    ("FRU0", [
        ("BARC", [
            "gimlet-sn08",
        ]),
        ("MAC0", [
            [0xA8, 0x40, 0x25, 0x01, 0x01, 0x05, 0x08, 0x00, 0x08],
        ]),
    ]),
]
```

(with a correctly allocated MAC address, and at least a string for the barcode, even though there is no barcode on this system)

This worked too.

PR #808, opened 2022-09-26
Faster MDIO!

mkeeter:
This switches MDIO to using a timer-based wait for sub-millisecond delays, bringing the VSC7448 boot sequence from 8 to 4 seconds!

### DIFF
--- a/task/net/src/main.rs
+++ b/task/net/src/main.rs
@@ -87,13 +87,11 @@ const RX_RING_SZ: usize = 4;
 /// Notification mask for our IRQ; must match configuration in app.toml.
 const ETH_IRQ: u32 = 1 << 0;
 
+/// Notification mask for MDIO timer; must match configuration in app.toml.
+const MDIO_TIMER_IRQ: u32 = 1 << 1;
+
 /// Notification mask for optional periodic logging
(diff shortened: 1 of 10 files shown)


task/net/src/main.rs
```diff
@@ -85,10 +85,15 @@ const TX_RING_SZ: usize = 4;
...
 
 /// Notification mask for our IRQ; must match configuration in app.toml.
-const ETH_IRQ: u32 = 1;
+const ETH_IRQ: u32 = 1 << 0;
 
 /// Notification mask for optional periodic logging
-const WAKE_IRQ: u32 = 2;
+const WAKE_IRQ: u32 = 1 << 1;
+
+// NOTE: bit 3 is used in the BSPs!
```
labbott, line 93: What about bit 2? Is there a nicer way to double check this since it looks like the TOML only shows the two IRQs?

mkeeter: I'm honestly not sure what's going on here – it dates back to @cbiffle 's original commit.

Consolidating to using bits 0,1,2 seems to work fine, so let's do that!

cbiffle (approved):

PR #809, opened 2022-09-27
Clippy fixes for RoT targets

labbott:


drv/lpc55-gpio-api/src/lib.rs
```diff
@@ -145,9 +145,13 @@ pub enum Digimode {
     Digital = 1,
 }
 
-impl Into<bool> for Digimode {
-    fn into(self) -> bool {
-        self == Self::Digital
+impl From<bool> for Digimode {
```
jgallagher, line 148: I'm a little confused about this change - this removed the ability to convert `Digimode -> bool` and added the ability to convert `bool -> Digimode`. Was clippy complaining about implementing `Into` instead of `From`? If so, should this be `impl From<Digimode> for bool` to preserve the existing functionality?

drv/lpc55-spi/src/lib.rs
```diff
@@ -195,7 +195,8 @@ impl Spi {
         rxignore: bool,
         len_bits: u8,
     ) {
-        if len_bits > 16 || len_bits < 4 {
+        // SPI hardware only supports lengths of range 4-16 bits
+        if !(4..=16).contains(&len_bits) {
```
jgallagher, line 199: This looks correct, but I'm not sure it's an improvement in readability, particularly because of the negation. :shrug:

drv/lpc55-update-server/src/main.rs
```diff
@@ -113,17 +113,25 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
-            HypoStatus::Success => Ok(()),
-            HypoStatus::OutOfBounds => Err(UpdateError::OutOfBounds.into()),
-            HypoStatus::RunningImage => Err(UpdateError::RunningImage.into()),
-            // Should probably encode the LPC55 flash status into the update
-            // error for good measure but that takes effort...
-            HypoStatus::FlashError(_) => Err(UpdateError::FlashError.into()),
+        unsafe {
+            // The write_to_flash API takes raw pointers due to TrustZone
+            // ABI requirements which makes this function unsafe.
+            match tz_table!().write_to_flash(
```
jgallagher, line 119: I think we could shrink the `unsafe` block a bit like this:

```rust
let result = unsafe {
    tz_table!().write_to_flash(...)
};
match result {
    ...
}
```

since it's only needed to call the function, not deal with the result.

drv/lpc55-spi-server/src/main.rs
```diff
@@ -19,7 +19,7 @@ task_slot!(GPIO, gpio_driver);
 
 #[derive(Copy, Clone, PartialEq)]
 enum Trace {
-    IRQ,
+    Irq,
```
mkeeter, line 22: If something is actually an acronym, I'd vote to use `#[allow(clippy::upper_case_acronyms)]` instead of messing with the capitalization

(this is what we do for `Device::IBC` in `power` and `Device::CPU` in `thermal`)

mkeeter (approved):

drv/lpc55-gpio-api/src/lib.rs
labbott: Turns out this code is actually unused so I removed it

drv/lpc55-update-server/src/main.rs
labbott: good suggestion

drv/lpc55-spi-server/src/main.rs
mkeeter: Okay, fine by me!

drv/lpc55-gpio-api/src/lib.rs
lzrd: I have some changes in this area that do impl From here.
I need to put that in a separate PR.

drv/lpc55-spi/src/lib.rs
lzrd: This can be an assert, right and get compiled out for production.

mkeeter: I don't think so – `assert!` remains in both debug and release builds; `debug_assert!` is only included in debug builds, but we never make debug builds 😆

PR #810, opened 2022-09-27
Move aux flash sizes to app.toml

jgallagher:
Allows clients to know the number and size of auxflash slots statically via constants exported from the API crate.

### DIFF
--- a/drv/auxflash-api/build.rs
+++ b/drv/auxflash-api/build.rs
@@ -2,10 +2,45 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use serde::Deserialize;
+use std::io::Write;
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let global_config = build_util::config::<GlobalConfig>()?;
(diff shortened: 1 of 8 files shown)


drv/auxflash-api/build.rs
```diff
@@ -2,10 +2,45 @@
...
+
+fn generate_auxflash_config(
+    config: &AuxFlashConfig,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let out_dir = std::env::var("OUT_DIR")?;
+    let dest_path = std::path::Path::new(&out_dir).join("auxflash_config.rs");
+
+    let mut out = std::fs::File::create(&dest_path)?;
+
+    writeln!(out, "pub const MEMORY_SIZE: u32 = {};", config.memory_size)?;
```
mkeeter, line 42: Now that we're doing this in `build.rs`, it would be nice to sprinkle in a few sanity checks:

- `SLOT_COUNT` should be >= 6
- `MEMORY_SIZE` should be evenly divisible by `SLOT_COUNT`
- `MEMORY_SIZE / SLOT_COUNT` should be page-aligned (I think assuming 64K pages is fine for now)

jgallagher: Added in a87e00e

PR #814, opened 2022-09-29
Allow for naming individual LPC55 GPIO pins.

lzrd:
Used for ROT_IRQ and SP_RESET in later PRs.

### DIFF
--- a/drv/lpc55-spi-server/src/main.rs
+++ b/drv/lpc55-spi-server/src/main.rs
@@ -8,6 +8,7 @@
 #![no_std]
 #![no_main]
 
+use drv_lpc55_gpio_api::Pin;
 use drv_lpc55_spi as spi_core;
 use drv_lpc55_syscon_api::{Peripheral, Syscon};
 use lpc55_pac as device;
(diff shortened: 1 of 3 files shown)


drv/lpc55-gpio-api/src/lib.rs
```diff
@@ -145,24 +145,64 @@ pub enum Digimode {
...
+impl From<bool> for Digimode {
+    fn from(state: bool) -> Self {
+        if state {
+            Digimode::Digital
+        } else {
+            Digimode::Analog
+        }
+    }
+}
+
```
labbott, line 157: Can you give me a pointer to where these functions actually get used in your code? I'm still not sure where they are actually needed.

labbott (commented):
This does make things much nicer.

drv/lpc55-gpio-api/src/lib.rs
lzrd: Bah.. I am not using them. They are leftover from some short-lived debugging.

lzrd: Removed

drv/lpc55-spi-server/src/main.rs
```diff
@@ -8,6 +8,7 @@
 #![no_std]
 #![no_main]
 
+use drv_lpc55_gpio_api::Pin;
```
labbott, line 11: Seem to have snuck in

labbott (approved):
One minor cleanup but otherwise LGTM

PR #817, opened 2022-09-30
Add update-buffer lib

jgallagher:
This is a library I intend to use in `mgmt-gateway` to deal with having multiple updateable components, each of which may have a different block/chunk/page size in which it wants updates staged (e.g., today we have SP updates that want 1024-byte chunks, and host flash / aux flash updates that want 256-byte chunks).

The intent is that one can declare a single static `BufferMutex` sized for the largest chunk of all updateable components, and then one can lock it _and record the component being updated_ (as the "owner"). This solves two issues:

1. We only need to allocate `max(component update chunk sizes)` memory to update buffers instead of `sum(component update chunk sizes)` as we're doing today.
2. If MGS attempts to start an update for component B while an update for component A is still in progress, we can include the name of component A in the error message we return.

I added it as a separate crate to make it easy to run the unit tests (and run them under Miri), since it has no direct dependency on anything hubris- or arm-specific, unlike `mgmt-gateway`.


lib/update-buffer/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+    cap: usize,
+}
+
+impl<T, const N: usize> Drop for UpdateBuffer<'_, T, N> {
+    fn drop(&mut self) {
+        // Avoiding deadlocks: We currently own the `data` lock and are
+        // attempting to acquire the `owner` lock. This is the reverse order
+        // from `BufferMutex::borrow()` above. See the comment in it above for
+        // the reasoning why this will not deadlock.
+        *self.owner.lock() = None;
```
mkeeter, line 114: Is it possible to explicitly drop `self.data` here?

That would make the ordering more obvious, but not sure if it's allowed.

jgallagher: It is not possible, no. At least not without wrapping `self.data` in an `Option` (or some moral equivalent), which would allow us to `.take()` it and explicitly drop it.

After rereading, I could change `borrow()` to not lock the data until after it unlocks the owner, which would eliminate basically all the comments regarding inverted lock ordering (since the only lock-within-a-lock would be right here: `borrow()` would acquire owner, release it, then acquire data).

mkeeter (approved):

PR #818, opened 2022-10-01
Document additional directories for the Hubris repo

stillinbeta:
It took me a while to figure out what `idl/` was about, so I decided to document it. I added `chip/` and `website/` mostly for completion's sake


mkeeter:
Thanks!

PR #820, opened 2022-10-03
Adds a `transceivers` task and `transceivers-api` for Sidecar

### LINKED ISSUE #765: Hubris APIs for transceiver communication & control
The transceivers on the QSFP board are connected to FPGA0/1 (via I2C and a few GPIOs).

These two FPGAs, in turn, are connected to the SP on Sidecar main board (via SPI, with a mux to select one or the other).

We should implement APIs for transceiver management on the Sidecar SP's Hubris image.

Aaron-Hartwig:
The goal of this PR is to add the basic needs of transceiver control and management to Hubris. This adds 

* control of enable (power supply), reset, and lpmode of individual modules
* readback of all low speed signals: the control signals above, in addition to power good, power good timeout, module presence, and module interrupt.
* ability to read/write up to 128 bytes on modules' I2C interface.

This does not yet add any SP monitoring for things like temperature and voltage.

Closes #765


drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -0,0 +1,201 @@
...
+        let mut r: [u32; 7] = [0; 7];
+        let f0: [u8; 14] = self.fpgas[0].read(Addr::QSFP_CTRL_EN_H)?;
+        let f1: [u8; 14] = self.fpgas[1].read(Addr::QSFP_CTRL_EN_H)?;
+
+        for i in 0..7 {
+            r[i] = ((f1[i * 2] as u32) << 24)
+                | ((f1[i * 2 + 1] as u32) << 16)
+                | ((f0[i * 2] as u32) << 8)
+                | (f0[i * 2 + 1] as u32);
+        }
```
arjenroodselaar, line 34: This is fine, but a bit bit-twiddly and hard to follow. If I understand the register map a slightly more functional/Rust approach, which expresses the idea of stitching together two half of an array, might be something along the lines of
```
let q0_q15: [u16; 7] = self.fpgas[0].read(Addr::QSFP_CTRL_EN_H)?;
let q16_q31: [u16; 7] = self.fpgas[1].read(Addr::QSFP_CTRL_EN_H)?;

let status = q0_q15.iter()
    .zip(q16_q31.iter())
    .map(|(q0_q15, q16_q31)| {
        ((q16_q31 as u32) << 16) | (q0_q15 as u32)
    });
    
Ok(status)
```
Depending on byte ordering you may have to do something like
```
((u16::from_be(q16_q31) as u32) << 16) | (u16::from_be(q0_q15) as u32)
```

arjenroodselaar (approved):
Seems good overall. @mkeeter may have some additional suggestions.

drv/fpga-api/src/lib.rs
```diff
@@ -265,6 +265,15 @@ impl FpgaUserDesign {
         Ok(v)
     }
 
+    pub fn read_bytes(
```
mkeeter, line 268: Now that you've got this function, you should modify `pub fn read<T>` above to use it.

drv/fpga-api/src/lib.rs
```diff
@@ -281,6 +290,16 @@ impl FpgaUserDesign {
             value.as_bytes(),
         )
     }
+
+    pub fn write_bytes(
```
mkeeter, line 294: Same here!

drv/sidecar-front-io/src/transceivers.rs
mkeeter: You could remove one layer of combination using `zerocopy::{U16, BE}` to read into U16s directly, then combine them with `zip`:
```rust
        let f0: [U16<BE>; 7] = self.fpgas[0].read(Addr::QSFP_CTRL_EN_H)?;
        let f1: [U16<BE>; 7] = self.fpgas[1].read(Addr::QSFP_CTRL_EN_H)?;

        let mut out: [u32; 7] = [0; 7];
        for (out, (lo, hi)) in out.iter_mut().zip(f0.iter().zip(f1.iter())) {
            *out = (lo.get() as u32) | ((hi.get() as u32) << 16);
        }

        Ok(out)
```
(you'll also need to add `U16` and `BE` to the `zerocopy` imports at the top)

More broadly, is there a reason this function wants to return raw bytes instead of `ModulesStatus`?

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    pub fn masked_op(
+        &self,
+        op: WriteOp,
+        mask: u32,
+        addr: Addr,
+    ) -> Result<(), FpgaError> {
+        let m = mask.as_bytes();
+
+        if (mask & 0xFFFF) > 0 {
```
mkeeter, line 47: Nit: use `!=` instead of `>` to make it more obvious (otherwise, I have to think about whether this is signed)

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -0,0 +1,201 @@
...
+            // Defaulting to RandomRead, rather than Read, because RandomRead
+            // sets the reg addr in the target device, then issues a restart to
+            // do the read at that reg addr. On the other hand, Read just starts
+            // a read wherever the reg addr is after the last transaction.
+            TransceiverI2COperation::RandomRead
+        } else {
+            TransceiverI2COperation::Write
+        };
+
+        if (mask & 0xFFFF) > 0 {
```
mkeeter, line 100: Same here (and below)

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub fn get_i2c_read_buffer(
+        &self,
+        port: u8,
+        buf: &mut [u8],
+    ) -> Result<(), FpgaError> {
+        let fpga_idx: usize = if port < 16 { 0 } else { 1 };
+        self.fpgas[fpga_idx].read_bytes(Self::read_buffer_address(port), buf)
+    }
+
+    pub fn set_i2c_write_buffer(&self, buf: &[u8]) -> Result<(), FpgaError> {
```
mkeeter, line 141: The asymmetry here is confusing: why don't we need to specify a `port` when setting the write buffer?

(I'm assuming it's because we always write the same buffer, but it would be good to clarify with a comment)

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -0,0 +1,201 @@
...
+            7 => Addr::QSFP_PORT7_READ_BUFFER,
+            8 => Addr::QSFP_PORT8_READ_BUFFER,
+            9 => Addr::QSFP_PORT9_READ_BUFFER,
+            10 => Addr::QSFP_PORT10_READ_BUFFER,
+            11 => Addr::QSFP_PORT11_READ_BUFFER,
+            12 => Addr::QSFP_PORT12_READ_BUFFER,
+            13 => Addr::QSFP_PORT13_READ_BUFFER,
+            14 => Addr::QSFP_PORT14_READ_BUFFER,
+            15 => Addr::QSFP_PORT15_READ_BUFFER,
+            _ => Addr::QSFP_PORT0_READ_BUFFER,
```
mkeeter, line 174: Make the default case here `unreachable!()`, instead of a dummy value.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -0,0 +1,201 @@
...
+        op as u8
+    }
+}
+
+#[derive(AsBytes, FromBytes, Unaligned)]
+#[repr(C)]
+pub struct TransceiversI2CRequest {
+    reg: u8,
+    num_bytes: u8,
+    mask: [u8; 2],
```
mkeeter, line 199: If this is philosophically a `u16` (and I think it is!), it would be cleaner to store as `U16<BigEndian>` (using `U16` and `BigEndian` from `zerocopy`.

(this will require fixing other places)

drv/transceivers-api/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+    InvalidNumberOfBytes = 3,
+}
+
+impl From<FpgaError> for TransceiversError {
+    fn from(_: FpgaError) -> Self {
+        Self::FpgaError
+    }
+}
+
+#[derive(Copy, Clone, zerocopy::FromBytes, zerocopy::AsBytes)]
```
mkeeter, line 26: Add a docstring explaining that each field is a bitmask of the 32 transceivers

drv/transceivers-api/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub enable: u32,
+    pub reset: u32,
+    pub lpmode_txdis: u32,
+    pub power_good: u32,
+    pub power_good_timeout: u32,
+    pub present: u32,
+    pub irq_rxlos: u32,
+}
+
+impl From<[u32; 7]> for ModulesStatus {
```
mkeeter, line 38: Because you've derived `FromBytes`, you could simplify this implementation to

```rust
ModulesStatus::read_from(data.as_bytes()).unwrap()
```

However, looking at the bigger picture, I'd consider
- removing the `impl From<...>` entirely
- make `Transceivers::get_modules_status` return a `ModuleStatus` instead of `[u32; 7]`
    - Do the `ModuleStatus::read_from(data.as_bytes()).unwrap()` in that function

(so we're _never_ passing around `[u32; 7]`, which is an unclear data type)

drv/transceivers-server/README.mkdn
```diff
@@ -0,0 +1,5 @@
+# Task Template
```
mkeeter, line 1: Delete this file

drv/transceivers-server/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+            .transceivers
+            .clear_lpmode(mask)
+            .map_err(TransceiversError::from)?)
+    }
+
+    fn setup_i2c_op(
+        &mut self,
+        _msg: &userlib::RecvMessage,
+        is_read: bool,
+        reg: u8,
```
mkeeter, line 104: Are there values of `reg` that are invalid?  If so, we should check and return an error.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -0,0 +1,201 @@
...
+            15 => Addr::QSFP_PORT15_READ_BUFFER,
+            _ => Addr::QSFP_PORT0_READ_BUFFER,
+        }
+    }
+}
+
+#[derive(AsBytes)]
+#[repr(u8)]
+pub enum TransceiverI2COperation {
+    Read = 0,
```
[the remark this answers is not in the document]
Aaron-Hartwig: Good point, I'll move the START bit into this struct and add a comment about those values.

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: Good point!

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: TIL `unreachable!()`, thanks!

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: You are correct. I will add a comment explaining that.

drv/transceivers-api/src/lib.rs
Aaron-Hartwig: Thanks for both of the recommended paths here. As you mention, given the bigger picture here, I will address this with your second recommendation.

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: Good point.

drv/transceivers-server/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+                .map_err(TransceiversError::from)?,
+        ))
+    }
+
+    fn set_power_enable(
+        &mut self,
+        _msg: &userlib::RecvMessage,
+        mask: u32,
+    ) -> Result<(), idol_runtime::RequestError<TransceiversError>> {
+        Ok(self
```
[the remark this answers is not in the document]
Aaron-Hartwig: Thanks for the style comment here, I agree with your suggestion being much more readable!

drv/transceivers-server/src/main.rs
Aaron-Hartwig: Not technically. `reg` represents a register address in the QSFP module, which is a 256-byte address space. I _could_ add some checking around `reg` and `num_bytes` as there is a little oddness around how the internal address register of the module works, but that could be starting to get a little cute.

Regardless, we don't expect more than a 128 byte operation here, so I will add a check for that.

drv/fpga-api/src/lib.rs
```diff
@@ -257,14 +257,19 @@ impl FpgaUserDesign {
         T: AsBytes + FromBytes,
     {
         let mut v = T::new_zeroed();
-        self.server.user_design_read(
-            self.device_index,
-            addr.into(),
-            v.as_bytes_mut(),
-        )?;
+        self.read_bytes(addr.into(), v.as_bytes_mut())?;
```
mkeeter, line 260: Nit: you don't need `addr.into()` here, since `read_bytes` will accept it as-is.

drv/fpga-api/src/lib.rs
```diff
@@ -274,12 +279,17 @@ impl FpgaUserDesign {
     where
         T: AsBytes + FromBytes,
     {
-        self.server.user_design_write(
-            self.device_index,
-            op,
-            addr.into(),
-            value.as_bytes(),
-        )
+        self.write_bytes(op, addr.into(), value.as_bytes())
```
mkeeter, line 282: Same here

mkeeter (approved):

drv/fpga-api/src/lib.rs
Aaron-Hartwig: Good point, I had not noticed that. Cleaned up in latest commit.

PR #822, opened 2022-10-04
mgmt-gateway: Cleanup and use one shared update buffer

jgallagher:
The bulk of the changes in this PR are to do some cleanup / reorganization of update handling. Previously, SP updates were handled in `mgs_common` and host flash updates were handled in `mgs_gimlet`. Now all update handling lives under the `update/` modules, and both SP updates and host flash updates use more explicit state machines. This PR does not add support for SP updates that have an auxflash image, but they will come in a follow PR I'll open immediately after opening this one.

Other changes in this PR:

* Track latest `gateway-messages` that separate the old `UpdatePrepare` message into two: `SpUpdatePrepare` for SP updates (includes the aux flash image size and CHCK value) and `ComponentUpdatePrepare` (what the old `UpdatePrepare` was - currently only used for host boot flash updates, but should be used for all other components moving forward)
* Removes the per-component `UpdateBuffer` in favor of using one static update buffer (introduced in #817) that is locked while an update is being applied. This removes our ability to update multiple components simultaneously, but reduces our update buffer memory usage from `SUM(all update component chunk sizes)` to `MAX(all update component chunk sizes)`.


task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -3,30 +3,43 @@
...
 use ringbuf::ringbuf_entry_root;
 use task_net_api::{Address, UdpMetadata};
 use userlib::{sys_get_timer, sys_irq_control, UnwrapLite};
 
+// Create type aliases that include our `BufferMutex` size (i.e., the size of
+// the largest update chunk of all the components we update). We could use some
+// kind of const max function over all the components, but we already have
+// static assertions for each component that they fit in this size, so we just
+// manually pick the component with the max (SP updates).
+pub(crate) type BufferMutex = ::update_buffer::BufferMutex<
```
mkeeter, line 30: I wish this could be an associated type of `UpdateBuffer`, but it looks like inherent associated types are unstable...

task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -3,30 +3,43 @@
...
+    SpComponent,
+    { drv_update_api::stm32h7::BLOCK_SIZE_BYTES },
+>;
+pub(crate) type UpdateBuffer = ::update_buffer::UpdateBuffer<
```
mkeeter, line 34: Nit: any reason to keep the `::` before `::update_buffer`?  (also true above)

task/mgmt-gateway/src/mgs_gimlet.rs
mkeeter: Oh, the naming actually confused me here: I expected `UpdateBuffer` to be the owner, and `BufferMutex` to be the borrowed thing.

They're just ambiguous enough to go either way; maybe use `UpdateBuffer` for the buffer itself, and `BorrowedUpdateBuffer` for the borrowed thing, to be super explicit?

task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -80,15 +95,18 @@ impl MgsHandler {
...
+        {
             Some(sys_get_timer().now + 1)
         } else {
             self.usart.from_rx_flush_deadline
         }
     }
 
     pub(crate) fn handle_timer_fired(&mut self) {
-        self.host_flash_update.erase_sectors_if_needed();
+        self.host_flash_update.step_preparation();
```
mkeeter, line 108: Add a comment here that these are no-ops if no update is in progress, and that only one will be running at a time.

task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -229,9 +263,8 @@ impl SpHandler for MgsHandler {
         }));
 
         match update.component {
-            SpComponent::SP_ITSELF => self.common.update_prepare(update),
             SpComponent::HOST_CPU_BOOT_FLASH => {
-                self.host_flash_update.prepare(update)
+                self.host_flash_update.prepare(&UPDATE_MEMORY, update)
```
mkeeter, line 267: It looks like `update_prepare` for the SP is moved to a separate MGS message type.  Do we need to remove `SP_ITSELF` from `enum SpComponent` [here](https://github.com/oxidecomputer/omicron/blob/7124e0a8381f26c767964664751d0edf7de32598/gateway-messages/src/lib.rs#L510-L511)?

mkeeter: Ah, I guess not, because we still use it in `update_chunk`.  In that case, out of curiosity, why did `sp_update_prepare` diverge from `component_update_prepare`?

task/mgmt-gateway/src/update/common.rs
```diff
@@ -0,0 +1,66 @@
...
+        self.update_state_with_result(|state| ((f)(state), ()));
+    }
+
+    #[inline(always)]
+    pub(super) fn update_state_with_result<F, T>(&mut self, f: F) -> T
+    where
+        F: FnOnce(State) -> (State, T),
+    {
+        let state = self.state.take().unwrap_lite();
+        let (new_state, t) = (f)(state);
```
mkeeter, line 62: Same here

task/mgmt-gateway/src/update/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+// same due to having to deal with both auxflash and SP images. It therefore
+// does not implement this trait. That's another wart that we might be able to
+// address once we start adding more implementors and tweak this trait.
+pub(crate) trait ComponentUpdater {
+    /// Size of one block / sector / page; whatever unit the underlying update
+    /// mechanism wants as a single chunk.
+    fn block_size(&self) -> usize;
+
+    /// Attempt to start preparing for an update, using `buffer` as the backing
+    /// store for incoming data.
```
mkeeter, line 33: Add a comment here that this records the `UpdateId` from `ComponentUpdatePrepare`, since that wasn't obvious to me

task/mgmt-gateway/src/update/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+        id: &UpdateId,
+        offset: u32,
+        data: &[u8],
+    ) -> Result<(), ResponseError>;
+
+    /// Abort the current update if it matches `id`.
+    ///
+    /// If no update is in progress, should return `Ok(())`; i.e., this should
+    /// only return an error if we attempted to abort the update with id `id`
+    /// and that abort failed.
```
mkeeter, line 61: Add comment about happens if an update is in progress, but with a different id – is that `Ok(())` or an error?

task/mgmt-gateway/src/update/host_flash.rs
```diff
@@ -0,0 +1,315 @@
...
+            .task
+            .capacity()
+            .map_err(|err| ResponseError::UpdateFailed(err as u32))?;
+
+        // How many total sectors do we need to erase? For gimlet, we know that
+        // capacity is an exact multiple of the sector size, which is probably
+        // a safe assumption for future parts as well. We'll assert here in case
+        // that ever becomes untrue, and we can update our math.
+        assert!(capacity % SECTOR_SIZE_BYTES == 0);
+        let num_sectors = (capacity / SECTOR_SIZE_BYTES) as u32;
```
mkeeter, line 102: Vague question: are we going to support differential writes to host flash (i.e. hashing sectors and only overwriting the ones that changed)?  Or is the network speed so much faster than SWD that it doesn't matter?

task/mgmt-gateway/src/update/host_flash.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub(crate) fn new() -> Self {
+        Self {
+            task: HostFlash::from(HOST_FLASH.get_task_id()),
+            current: None,
+        }
+    }
+}
+
+impl ComponentUpdater for HostFlashUpdate {
+    fn block_size(&self) -> usize {
```
mkeeter, line 35: This doesn't use `self`, so could be a `const fn block_size() -> usize`

(which would then make it easier to use in an explicit `const` calculation of buffer size)

task/mgmt-gateway/src/update/host_flash.rs
```diff
@@ -0,0 +1,315 @@
...
+                }
+                Err(err) => State::Failed(err),
+            }
+        });
+    }
+
+    fn status(&self) -> UpdateStatus {
+        let current = match self.current.as_ref() {
+            Some(current) => current,
+            None => return UpdateStatus::None,
```
mkeeter, line 172: Would it make sense to remove `UpdateStatus::None` and instead return an `Option<UpdateStatus>`?  

(Not at all necessary for this PR, since it would require `omicron` changes as well)

task/mgmt-gateway/src/mgs_gimlet.rs
jgallagher: Yeah, I like those names better. :+1:

task/mgmt-gateway/src/mgs_gimlet.rs
jgallagher: Nope! This is leftover from when `mgmt-gateway` still had its own `update_buffer` submodule, and the `::` were required. Will remove.

task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -80,15 +95,18 @@ impl MgsHandler {
         // avoid blocking while the entire erase happens. If we're in that case,
         // set our timer for 1 tick from now to give a window for other
         // interrupts/notifications to arrive.
-        if self.host_flash_update.needs_sectors_erased() {
+        if self.host_flash_update.is_preparing()
+            || self.sp_update.is_preparing()
+        {
             Some(sys_get_timer().now + 1)
```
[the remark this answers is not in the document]
jgallagher: That's intentional - if we're "preparing" (e.g., erasing flash), MGS is actively waiting for us to finish so it can start sending us an update. We only take a pause here to check for any new interrupts (e.g., MGS requesting status / progress). If we could remain responsive without this timer it wouldn't be here at all and we'd be doing all the prep work in a loop.

task/mgmt-gateway/src/mgs_gimlet.rs
jgallagher: `SpUpdatePrepare` grew fields that are auxflash specific (the auxflash image size and CHCK value), and does not have a component field (it's implicitly `SP_ITSELF`); `ComponentUpdatePrepare` has a `component` field but no auxflash fields.

task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -268,11 +303,13 @@ impl SpHandler for MgsHandler {
             component
         }));
 
-        match component {
-            SpComponent::SP_ITSELF => Ok(self.common.status()),
+        let status = match component {
+            SpComponent::SP_ITSELF => self.sp_update.status(),
```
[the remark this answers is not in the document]
jgallagher: No, but it's not obvious why that is. I'll add comments to explain. A first attempt to do so here:

MGS hides the fact that aux flash images exist from its clients - they only know about updates. When they ask for the status of or to abort an update, they will say "give me the status of the update being applied to `SP_ITSELF`" - at that level, we don't actually apply updates to `SP_AUX_FLASH`.

`update_chunk`, on the other hand, is only called directly by MGS as a part of delivering the update data, never by a client that doesn't know the details. Those chunks can include the detail of whether the data in that chunk is for the SP or for the aux flash. Strictly speaking even that isn't necessary, since the contract could be "the first `aux_flash_size` bytes are for the aux flash and the remaining `sp_image_size` bytes are for the SP", but it seemed slightly safer to include the component name along with the data.

task/mgmt-gateway/src/update/host_flash.rs
jgallagher: I was not planning to support differential writes because of the complexity that would add to update delivery. If update performance becomes a concern we can revisit, but I think it's probably fast enough.

task/mgmt-gateway/src/update/host_flash.rs
jgallagher: That's what it was originally, but in PR review on the `omicron` side we switched it for clarity / readability reasons (e.g., when skimming code, we always have an `UpdateStatus` available, but one of the possible status values is "no update in progress"). A minuscule technical benefit is that the hubpack encoding of `UpdateStatus` (with a `None` case) is one byte shorter than the encoding of `Option<UpdateStatus>`.

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -0,0 +1,296 @@
...
+            .map_err(|component| {
+                ResponseError::OtherComponentUpdateInProgress(component)
+            })?;
+
+        // TODO handle auxflash updates
+        if update.aux_flash_size > 0 {
+            return Err(ResponseError::RequestUnsupportedForSp);
+        }
+
+        // Attempt to prepare for an update (erases our flash).
```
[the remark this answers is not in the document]
jgallagher: It actually erases the flash - the SP update server does not currently expose erasing the alternate bank piecemeal, although maybe it should? This step takes 3-4 seconds on both my gimletlet and all the gimlet/sidecar SPs I've updated.

task/mgmt-gateway/src/mgs_gimlet.rs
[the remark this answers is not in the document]
mkeeter: Makes sense!

task/mgmt-gateway/src/update/sp.rs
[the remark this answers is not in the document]
mkeeter: That seems fine, just wanted to make sure the comment was correct

task/mgmt-gateway/src/update/mod.rs
jgallagher: I think I'm coming around to your position. _Someone_ has to decide what to do if we try to abort a completed update, and with the implementation as it is now, we swallow that information, which would force MGS to make more status calls (all inherently racy) to figure out what's going on. I'll change this to:

* If `id != current.id`, return an error no matter what our state is.
* If the ID matches and we're in state `Complete`, return an error.
* If the ID matches and we're in an active state or `Aborted`, succeed and return `Ok(())`.

I'll make similar changes to the SP update abort.

task/mgmt-gateway/src/mgs_gimlet.rs
jgallagher: Renamed in 606e9f46

task/mgmt-gateway/src/mgs_gimlet.rs
```diff
@@ -3,30 +3,43 @@
...
 use heapless::Deque;
 use ringbuf::ringbuf_entry_root;
 use task_net_api::{Address, UdpMetadata};
 use userlib::{sys_get_timer, sys_irq_control, UnwrapLite};
 
+// Create type aliases that include our `BufferMutex` size (i.e., the size of
+// the largest update chunk of all the components we update). We could use some
+// kind of const max function over all the components, but we already have
+// static assertions for each component that they fit in this size, so we just
+// manually pick the component with the max (SP updates).
```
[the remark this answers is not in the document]
jgallagher: Added in 50b40738 (and then reworked slightly in f4743daf)

task/mgmt-gateway/src/mgs_gimlet.rs
jgallagher: Added in 644a64a3

task/mgmt-gateway/src/mgs_gimlet.rs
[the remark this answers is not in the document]
jgallagher: Added comments to this and update_abort in 644a64a3

task/mgmt-gateway/src/update/mod.rs
jgallagher: Changed in d70746d5

task/mgmt-gateway/src/update/host_flash.rs
jgallagher: Traits can't have `const fn`s yet, but I changed it to an associated `const` instead in f4743daf.

I think originally I was hoping to stack an array of `&'static dyn ComponentUpdater` objects, in which casing having it as a `&self` function would still be useful. But for now we only have one implementer of this, so I went ahead with the change (plus I can use these constants when determining `UPDATE_BUFFER_SIZE`, which is handy). Can always revisit if we end up wanting trait objects at some point.

PR #823, opened 2022-10-04
mgmt-gateway: Handle SP updates that have auxflash images

jgallagher:
This builds on #822 and handles SP updates that have aux flash images.

### DIFF
--- a/drv/auxflash-api/src/lib.rs
+++ b/drv/auxflash-api/src/lib.rs
@@ -10,6 +10,7 @@ use derive_idol_err::IdolError;
 use sha3::{Digest, Sha3_256};
 use tlvc::{TlvcRead, TlvcReader};
 use userlib::*;
+use zerocopy::{AsBytes, FromBytes};
 
 pub use drv_qspi_api::{PAGE_SIZE_BYTES, SECTOR_SIZE_BYTES};
 
@@ -47,21 +48,19 @@ pub enum AuxFlashError {
     SlotActive,
 }
 
-#[derive(Copy, Clone, zerocopy::FromBytes, zerocopy::AsBytes)]
+#[derive(Copy, Clone, FromBytes, AsBytes)]
 #[repr(transparent)]
 pub struct AuxFlashId(pub [u8; 20]);
 
-#[derive(
-    Copy, Clone, PartialEq, Eq, zerocopy::FromBytes, zerocopy::AsBytes,
-)]
+#[derive(Copy, Clone, PartialEq, Eq, FromBytes, AsBytes)]
 #[repr(transparent)]
 pub struct AuxFlashChecksum(pub [u8; 32]);
 
-#[derive(Copy, Clone, zerocopy::FromBytes, zerocopy::AsBytes)]
+#[derive(Copy, Clone, FromBytes, AsBytes)]
 #[repr(transparent)]
 pub struct AuxFlashTag(pub [u8; 4]);
 
-#[derive(Copy, Clone, zerocopy::FromBytes, zerocopy::AsBytes)]
+#[derive(Copy, Clone, FromBytes, AsBytes)]
 #[repr(C)]
 pub struct AuxFlashBlob {
     pub slot: u32,
--- a/drv/auxflash-api/src/lib.rs
+++ b/drv/auxflash-api/src/lib.rs
@@ -51,7 +51,9 @@ pub enum AuxFlashError {
 #[repr(transparent)]
 pub struct AuxFlashId(pub [u8; 20]);
 
-#[derive(Copy, Clone, zerocopy::FromBytes, zerocopy::AsBytes)]
+#[derive(
+    Copy, Clone, PartialEq, Eq, zerocopy::FromBytes, zerocopy::AsBytes,
+)]
 #[repr(transparent)]
 pub struct AuxFlashChecksum(pub [u8; 32]);
 
--- a/task/mgmt-gateway/Cargo.toml
+++ b/task/mgmt-gateway/Cargo.toml
@@ -11,6 +11,7 @@ serde = {version = "1", default-features = false, features = ["derive"]}
 ssmarshal = {version = "1", default-features = false}
 static_assertions = "1.1"
 
+drv-auxflash-api = {path = "../../drv/auxflash-api", optional = true}
 drv-gimlet-hf-api = {path = "../../drv/gimlet-hf-api", optional = true}
 drv-gimlet-seq-api = {path = "../../drv/gimlet-seq-api", optional = true}
 drv-sidecar-seq-api = {path = "../../drv/sidecar-seq-api", optional = true}
@@ -36,3 +37,4 @@ vlan = ["task-net-api/vlan"]
 usart1 = []
 baud_rate_3M = []
 hardware_flow_control = []
+auxflash = ["drv-auxflash-api"]
--- a/task/mgmt-gateway/src/update/sp.rs
+++ b/task/mgmt-gateway/src/update/sp.rs
@@ -2,6 +2,58 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// State machine transitions for the "happy path" is below. From any state, we
+// could transition to `Aborted` (on request from MGS) or `Failed` (if an error
+// occurs on our side).
+
+/*
+               ┌─────────────────┐
+               │ SpUpdatePrepare │
+               └┬───────────────┬┘
+                │               │
+┌───────────────▼───┐       ┌───▼──────────────┐
+│aux_image_size == 0│     ┌─┤aux_image_size > 0│
+└┬──────────────────┘     │ └──────────────────┘
+ │                        │
+ │                 ┌──────▼───────────────────┐ No  ┌─────┐
+ │                 │#[cfg(feature="auxflash")]├────►│Error│
+ │                 └───────┬──────────────────┘     └─────┘
+ │                         │Yes
+ │   ┌─────────────────────▼────────────────────────────┐
+ │   │State::AuxFlash(AuxFlashState::ScanningForChck(_))│
+ │   └─────┬──────────────────────────┬─────────────────┘
+ │         │Match Found               │Match Not Found
+ │         │           ┌──────────────▼───────────────────────────────┐
+ │         │           │State::AuxFlash(AuxFlashState::ErasingSlot(_))│
+ │         │           └──────┬───────────────────────────────────────┘
+ │         │                  │
+ │ ┌───────┼──────────────────┼───────────────────────────────────────────┐
+ │ │       │                  │                                           │
+ │ │       │   ┌──────────────▼────────────────────────────────────────┐  │
+ │ │       │   │State::AuxFlash(AuxFlashState::FinishedErasingSlot(_)) ├──┼─┐
+ │ │       │   └───────────────────────────────────────────────────────┘  │ │
+ │ │       │                            MGS waits to see one of these two │ │
+ │ │ ┌─────▼──────────────────────────┐ states to decide whether to send  │ │
+ │ │ │State::FoundMatchingAuxFlashChck│ the aux flash image or skip it    │ │
+ │ │ └────┬───────────────────────────┘ and only send the SP image.       │ │
+ │ │      │                                                               │ │
+ │ └──────┼───────────────────────────────────────────────────────────────┘ │
+ │        │                                                                 │
+ │        │              ┌────────────────────────────────────────────────┐ │
+ │        │              │State::AuxFlash(AuxFlashState::AcceptingData(_))◄─┘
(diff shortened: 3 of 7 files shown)


drv/auxflash-api/src/lib.rs
```diff
@@ -51,7 +51,9 @@ pub enum AuxFlashError {
 #[repr(transparent)]
 pub struct AuxFlashId(pub [u8; 20]);
 
-#[derive(Copy, Clone, zerocopy::FromBytes, zerocopy::AsBytes)]
+#[derive(
```
mkeeter, line 54: Nit: since we use `zerocopy::{AsBytes, FromBytes}` a lot in this file, maybe add them as an import (and let this stay a single line)

task/mgmt-gateway/Cargo.toml
```diff
@@ -11,6 +11,7 @@ serde = {version = "1", default-features = false, features = ["derive"]}
 ssmarshal = {version = "1", default-features = false}
 static_assertions = "1.1"
 
+drv-auxflash-api = {path = "../../drv/auxflash-api", optional = true}
 drv-gimlet-hf-api = {path = "../../drv/gimlet-hf-api", optional = true}
 drv-gimlet-seq-api = {path = "../../drv/gimlet-seq-api", optional = true}
 drv-sidecar-seq-api = {path = "../../drv/sidecar-seq-api", optional = true}
```
mkeeter, line 17: I appreciate the perfectly increasing line lengths in this block 😂

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -54,7 +78,8 @@ impl SpUpdate {
                 ResponseError::OtherComponentUpdateInProgress(component)
             })?;
 
-        // TODO handle auxflash updates
+        // Can we handle an auxflash update?
+        #[cfg(not(feature = "auxflash"))]
         if update.aux_flash_size > 0 {
```
mkeeter, line 83: IMHO, this would be more clear as
```rust
if update.aux_flash_size > 0 && !cfg!(feature = "auxflash") {
```
or
```rust
if update.aux_flash_size > 0 && cfg!(not(feature = "auxflash")) {
```

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -11,19 +12,54 @@ use gateway_messages::{
     UpdateInProgressStatus, UpdateStatus,
 };
 
+#[cfg_attr(not(feature = "auxflash"), path = "sp/stub_auxflash.rs")]
+mod auxflash;
```
mkeeter, line 16: This is a little confusing – maybe move it into the `cfg_if!` block, e.g.
```rust
cfg_if! {
    if #[cfg(feature = "auxflash")] {
        mod auxflash;
    } else {
        mod stub_auxflash;
        use stub_auxflash as auxflash;
   }
```
(if that's allowed?)

task/mgmt-gateway/src/update/sp/stub_auxflash.rs
```diff
@@ -0,0 +1,101 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+/// Stub implementation that has the same types and methods as the real
+/// `auxflash` module (our sibling). All method implementations panic: none of
+/// them should ever be called.
+use super::AuxFlash;
+use crate::mgs_handler::UpdateBuffer;
+use gateway_messages::{ResponseError, UpdateId, UpdateStatus};
+
+#[allow(dead_code)] // we never construct any of these variants
```
mkeeter, line 12: Maybe move these to `auxflash_common.rs`, to avoid duplication between `auxflash.rs` and `stub_auxflash.rs`?

mkeeter: Oh, never mind, the important part is `State`, which has a diferent `impl`

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -11,19 +12,54 @@ use gateway_messages::{
...
+        use drv_auxflash_api::AuxFlash;
+
+        userlib::task_slot!(AUX_FLASH_SERVER, auxflash);
+    } else {
+        // If the auxflash feature isn't enabled, we can stub out a fake
+        // `AuxFlash` server that we never actually do anything with (other than
+        // construct it).
+        use userlib::TaskId;
+
+        struct AuxFlash;
```
mkeeter, line 33: Could this (and `struct FakeAuxFlashServer`) be defined in `stub_auxflash.rs`, instead of defining it in-line here?

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -11,19 +12,54 @@ use gateway_messages::{
...
+        impl From<TaskId> for AuxFlash {
+            fn from(_: TaskId) -> Self {
+                Self
+            }
+        }
+
+        struct FakeAuxFlashServer;
```
mkeeter, line 40: Nit: maybe name `FakeAuxFlashTaskSlot`, since that's a more accurate description of this type.

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -80,10 +126,52 @@ impl SpUpdate {
...
+
+        current.update_state(|state| match state {
+            // auxflash states that have prep work to do.
+            State::AuxFlash(AuxFlashState::ScanningForChck(scan)) => {
+                match scan.continue_scanning(&self.auxflash_task) {
+                    ChckScanResult::FoundMatch(mut buffer) => {
+                        // Take ownership of `buffer` back, and resize it for
+                        // our blocks.
+                        buffer
+                            .reborrow(SpComponent::SP_ITSELF, BLOCK_SIZE_BYTES);
```
mkeeter, line 150: Add a comment explaining why this isn't `SpComponent::SP_AUX_FLASH`?

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -131,8 +227,74 @@ impl SpUpdate {
...
+                AuxFlashState::ScanningForChck(_)
+                | AuxFlashState::ErasingSlot(_) => {
+                    return (
+                        State::AuxFlash(auxflash_state),
+                        Some(Err(ResponseError::UpdateNotPrepared)),
+                    );
+                }
+                AuxFlashState::Failed(err) => {
+                    return (
+                        State::AuxFlash(auxflash_state),
```
mkeeter, line 264: Do we want to fall into `State::Failing` here, instead of the nested fail state?

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -166,25 +333,29 @@ impl SpUpdate {
...
+                    return Err(ResponseError::UpdateInProgress(self.status()));
+                }
 
-        match self.sp_task.abort_update() {
-            // Aborting an update that hasn't started yet is fine;
-            // either way our caller is clear to start a new update.
-            Ok(()) | Err(UpdateError::UpdateNotStarted) => {
-                *current.state_mut() = State::Aborted;
-                Ok(())
+                match self.sp_task.abort_update() {
```
mkeeter, line 347: Do we need to call into `self.auxflash_task` instead of `self.sp_task` if we're in an auxflash state?

task/mgmt-gateway/src/update/sp/auxflash.rs
```diff
@@ -0,0 +1,335 @@
...
+        // Scan the slot at `index`.
+        match task.read_slot_chck(self.index) {
+            Ok(chck) => {
+                // If this matches, we're done; transition to the
+                // next state and return the buffer to our caller.
+                if chck == self.chck {
+                    return ChckScanResult::FoundMatch(self.buffer);
+                }
+            }
+            Err(AuxFlashError::MissingChck) => {
```
mkeeter, line 165: I think that `BadChckSize`, `MissingAuxi`, `MultipleChck`, `MultipleAuxi`, and `ChckMismatch` should all be in this section: treating corrupt slots as empty seems reasonable, since they're useless otherwise.

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -141,6 +303,11 @@ impl SpUpdate {
         // Handle SP image states.
         current.update_state_with_result(|state| {
             let accepting = match state {
+                State::AuxFlash(_) => unreachable!(), // handled above
+                State::FoundMatchingAuxFlashChck { buffer } => AcceptingData {
+                    buffer,
+                    next_write_offset: 0,
```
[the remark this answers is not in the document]
jgallagher: > I'm vaguely confused about the flow here: do we get to FoundMatchingAuxFlashChck if we have a valid checksum before or after writing the auxflash?
>
> If we skip writing when there's a matching checksum, how do we know to discard the first aux_flash_size bytes coming from the host?

Before, with an asterisk: if we find a match, we never write the auxflash. If we're in `State::AuxFlash(AuxFlashState::ScanningForChck(_))`, we can either transition to:

* `State::FoundMatchingAuxFlash` (if a match is found); if we're in this state, we expect the host to skip sending the aux flash image entirely, so we're waiting for `SP_ITSELF` update chunks.
* `State::AuxFlash(AuxFlashState::ErasingSlot(_)` then `State::AuxFlash(AuxFlashState::FinishedErasingSlot(_)`; if we're in this state, we expect the host to send us the aux flash image.

MGS will poll our `status()` until we return `UpdateStatus::SpUpdateAuxFlashChckScan`, which includes a boolean indicating which of the above two states we're in, so it knows whether or not to skip sending the aux flash data.

task/mgmt-gateway/src/update/sp.rs
jgallagher: No; there's nothing to abort on the aux flash side, but we do need to tell the SP `UpdateServer` to abort, even if we were still in one of the aux flash states (because we told it to prepare for an update back in `prepare()`).

task/mgmt-gateway/src/update/sp.rs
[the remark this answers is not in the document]
mkeeter: Got it, the missing piece was that MGS will decide to send / not send the auxflash data.

mkeeter (approved):

task/mgmt-gateway/src/update/sp.rs
```diff
@@ -11,19 +12,37 @@ use gateway_messages::{
     UpdateInProgressStatus, UpdateStatus,
 };
 
+cfg_if! {
+    if #[cfg(feature = "auxflash")] {
+        use drv_auxflash_api::AuxFlash;
+
+        mod auxflash;
+
+        use self::auxflash::State as AuxFlashState;
```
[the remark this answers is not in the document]
jgallagher: It's not; I thought it was clearer than `use auxflash::...` since we are depending on other auxflash-related crates (even though those aren't named `auxflash`). I don't feel strongly about it though.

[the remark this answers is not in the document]
mkeeter: I see `use self::` rarely enough that I did a double take, but could go either way.

task/mgmt-gateway/src/update/sp.rs
jgallagher: I don't think so, but it doesn't _really_ matter since the only way to recover from landing in a `Failed` state (of either variety) is to abort the update and restart it. `State::Failed(_)` holds an `UpdateError` but `AuxFlashState::Failed(_)` holds an `AuxFlashError`; to fall into `State::Failed(_)` we'd have to discard the aux flash error and pick an update error instead (probably `UpdateError::FlashError`, which is what I use in another place where we have to do this).

drv/auxflash-api/src/lib.rs
jgallagher: fixed in dfe90f63

task/mgmt-gateway/src/update/sp.rs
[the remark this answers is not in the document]
jgallagher: Removed as part of the stub import cleanup in 5c631443

task/mgmt-gateway/src/update/sp.rs
jgallagher: It is! Changed in 5c631443

task/mgmt-gateway/src/update/sp.rs
jgallagher: Yep: done in 5c631443

task/mgmt-gateway/src/update/sp.rs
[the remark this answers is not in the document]
jgallagher: Added a state machine diagram in 1fc90851 that should help

task/mgmt-gateway/src/update/sp/auxflash.rs
jgallagher: :+1: added in e966a578

task/mgmt-gateway/src/update/sp.rs
jgallagher: Noted this in 1fc90851

PR #824, opened 2022-10-04
(Re)load Sidecar mainboard controller based on bitstream checksum

### LINKED ISSUE #769: Sidecar mainboard FPGA does not use checksum

arjenroodselaar:
This diff uses the checksum soft fuses introduced in https://github.com/oxidecomputer/quartz/pull/29 to determine if a bitstream reload is needed.

Fixes #769


Aaron-Hartwig (changes requested):
In an effort to have more traceability around "what FPGA image is this running", I think we should do a few things.

1) Only use collateral generated from a `Quartz` Buildomat job
2) Link to that job in a README

This is all a bit manual for now, but it gives visibility into what exactly Hubris is trying to program (and confidence it is not some special 1-off build from someone's machine). I tried to model this in [drv/sidecar-front-io](https://github.com/oxidecomputer/hubris/tree/master/drv/sidecar-front-io)

Aaron-Hartwig (approved):
Thanks for adding that stuff! Like we've chatted about, we have a lot of work to do on the FPGA CI front, but for now this gets us the traceability we want.

PR #828, opened 2022-10-06
Use path attr instead of giant cfg_if trees for BSPs

mkeeter:
Also includes two drive-by fixes:
- Missing task slot in G030
- Warning about unused function in `gimlet/app.toml`

### DIFF
--- a/task/monorail-server/src/main.rs
+++ b/task/monorail-server/src/main.rs
@@ -5,6 +5,11 @@
 #![no_std]
 #![no_main]
 
+#[cfg_attr(target_board = "sidecar-a", path = "bsp/sidecar_1.rs")]
+#[cfg_attr(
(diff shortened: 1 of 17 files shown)


labbott (approved):
LGTM

task/monorail-server/src/main.rs
```diff
@@ -5,6 +5,11 @@
 #![no_std]
 #![no_main]
 
+#[cfg_attr(target_board = "sidecar-a", path = "bsp/sidecar_1.rs")]
```
jgallagher, line 8: Add a `#[deny(unused_attributes)]`

mkeeter: I ended up doing that globally in `build/dist.rs`

PR #832, opened 2022-10-06
host_sp_comms: Minor protocol updates

jgallagher:
Three small, mostly orthogonal changes to track the latest RFD 316 updates:

1. Ignore empty packets from the host instead of returning a decode error
2. Periodically send the host empty packets, if we're not otherwise receiving or transmitting (at least up until our TX FIFO fills)
3. Update to minor changes in a couple of the message variants

These are broken into three commits, and might be easier to review independently (although the total changes are not particularly large either way).

cc @citrus-it


task/host-sp-comms/src/main.rs
```diff
@@ -301,6 +334,54 @@ impl ServerImpl {
...
+        if let Some(deadline) = self.tx_zero_deadline {
+            if now >= deadline {
+                if self.tx_pkt_to_write.is_empty() && self.rx_buf.is_empty() {
+                    // If we have room in the FIFO, send a 0 byte and then reset
+                    // our timer to wake back up. If we can't, turn on the TX
+                    // FIFO empty interrupt instead, and wait for it to wake us
+                    // up.
+                    if try_tx_push(&self.uart, 0) {
+                        self.tx_zero_deadline = Some(now + UART_ZERO_DELAY);
+                        self.uart.disable_tx_fifo_empty_interrupt();
```
mkeeter, line 371: I'm confused why we disable the FIFO empty interrupt here.  We've just pushed data into the FIFO, do we not want to do anything when it's sent?

task/host-sp-comms/src/main.rs
```diff
@@ -301,6 +334,54 @@ impl ServerImpl {
...
+                    } else {
+                        self.tx_zero_deadline = None;
```
mkeeter, line 373: I _think_ this will be reset during the next call to `handle_usart_notification`, but it's not totally obvious – we might want another state machine diagram showing what interrupts are enabled when...

task/host-sp-comms/src/main.rs
jgallagher: I think this is just a mistake that happens to not really matter when hardware flow control is working (because the tx fifo stays full until the host sends us a packet, at which point we take a uart interrupt due to rx instead of tx). I'll remove it - nice catch.

task/host-sp-comms/src/main.rs
jgallagher: Added in 7d4df699

mkeeter (approved):

PR #833, opened 2022-10-06
Dice mfg cmds

flihp:


flihp:
AFAIK the most controversial part of this PR will be the new memory region where common artifacts are handed off. This requires that multiple tasks be able to read from the handoff memory region. Naturally write access will allow tasks to corrupt data consumed by another task. I've created an issue to track implementing a read-only mapping here: https://github.com/oxidecomputer/hubris/issues/835 and will take this work on if the design is acceptable.

lib/dice/src/mfg.rs
```diff
@@ -0,0 +1,272 @@
...
+pub enum Error {
+    MsgDecode,
+    MsgBufFull,
+    UsartRead,
+    UsartWrite,
+}
+
+// data returned to caller by MFG
+// typically required to use ECA post MFG
+// this is what gets written to persistent storage after successful mfg
```
labbott, line 24: Can you define ECA?

stage0/src/usart.rs
```diff
@@ -0,0 +1,96 @@
...
+    // digital mode
+    iocon
+        .pio0_29
+        .write(|w| w.func().alt1().digimode().digital());
+    iocon
+        .pio0_30
+        .write(|w| w.func().alt1().digimode().digital());
+
+    // disable IOCON clock
+    syscon.ahbclkctrl0.modify(|_, w| w.iocon().disable());
```
labbott, line 48: probably not worth turning off IOCON since the ROM ends up leaving it on for us anyway

stage0/src/usart.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    // reset pin 29 and 30 to defaults
+    iocon.pio0_29.reset();
+    iocon.pio0_30.reset();
+
+    syscon.presetctrl0.modify(|_, w| w.iocon_rst().asserted());
+    syscon.ahbclkctrl0.modify(|_, w| w.iocon().disable());
+
+    syscon.presetctrl0.modify(|_, w| w.gpio0_rst().asserted());
+    syscon.ahbclkctrl0.modify(|_, w| w.gpio0().disable());
```
labbott, line 67: again, probably not worth turning those off

lib/dice/src/mfg.rs
flihp: Apologies, this is a DICE-ism. ECA is "embedded certificate authority" and in this case it's the DeviceId cert. These comments probably should have been removed before the PR & they'll be replaced by code in my next PR.

stage0/src/usart.rs
flihp: I figured the safest thing to do was to return the platform to the default state. This was intended to avoid possibly breaking assumptions made by existing hubris code but also to prevent future code in hubris relying on behavior in stage0 that may change. This may be overly cautious but unless we need to reduce code size or speed up boot time it seemed like the prudent thing to do.

flihp: LMK if you're asking for this to be removed before merging.

flihp: fixed

stage0/src/usart.rs
flihp: fixed

labbott (approved):
Overall LGTM

lib/dice/src/alias_cert_tmpl.rs
```diff
@@ -2,75 +2,62 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// NOTE: This DER blob, offsets & lengths is mostly generated code. This was
```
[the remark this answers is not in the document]
flihp: I doubt it. Looks like I got sloppy. Will fix this up before merging.

stage0/src/dice_mfg_usart.rs
```diff
@@ -0,0 +1,216 @@
...
+macro_rules! flash_page_align {
+    ($size:expr) => {
+        if $size % lpc55_romapi::FLASH_PAGE_SIZE != 0 {
+            ($size & !(lpc55_romapi::FLASH_PAGE_SIZE - 1))
+                + lpc55_romapi::FLASH_PAGE_SIZE
+        } else {
+            $size
+        }
+    };
+}
```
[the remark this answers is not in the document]
flihp: they will be missed

stage0/src/dice.rs
```diff
@@ -3,122 +3,155 @@
...
 use lpc55_pac::Peripherals;
 use salty::signature::Keypair;
 use sha3::{Digest, Sha3_256};
 use unwrap_lite::UnwrapLite;
 
-fn get_deviceid_keypair(cdi: &Cdi) -> Keypair {
+#[cfg(feature = "dice-self")]
+use crate::dice_mfg_self::gen_mfg_artifacts;
+#[cfg(feature = "dice-mfg")]
+use crate::dice_mfg_usart::gen_mfg_artifacts;
```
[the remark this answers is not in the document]
flihp: yup

lib/dice/src/alias_cert_tmpl.rs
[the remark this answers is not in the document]
flihp: ROFL: The code that generates this file doesn't generate the comment. So the template was updated but I didn't preserve the comment that was added by hand.

PR #838, opened 2022-10-07
Move the thermal config from RAM into flash

mkeeter:
This brings the `thermal` task from 18648 to 17720 bytes of flash (a small improvement)

More importantly, it drops the stack usage from 7392 to 3560, meaning we can shrink the RAM usage by 8192!

### DIFF
--- a/app/gimlet/rev-b.toml
+++ b/app/gimlet/rev-b.toml
@@ -120,8 +120,8 @@ task-slots = ["sys", "i2c_driver"]
 name = "task-thermal"
 features = ["itm", "gimlet"]
 priority = 5
-max-sizes = {flash = 32768, ram = 16384 }
-stacksize = 8000
+max-sizes = {flash = 32768, ram = 8192 }
+stacksize = 4504
 start = true
 task-slots = ["i2c_driver", "sensor", "gimlet_seq", "jefe"]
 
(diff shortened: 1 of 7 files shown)


task/thermal/src/bsp/gimlet_b.rs
```diff
@@ -173,359 +120,374 @@ impl Bsp {
...
+            devices::nvmebmc_u2_n9,
+            sensors::NVMEBMC_U2_N9_TEMPERATURE_SENSOR,
+        ),
+        U2_THERMALS,
+        POWER_STATE_A0,
+        true,
+    ),
+];
+
+const MISC_SENSORS: [TemperatureSensor; NUM_TEMPERATURE_SENSORS] = [
```
jgallagher, line 462: Looks like the order of this array changed - does that matter?

jgallagher (approved):

app/gimlet/rev-b.toml
```diff
@@ -120,8 +120,8 @@ task-slots = ["sys", "i2c_driver"]
 name = "task-thermal"
 features = ["itm", "gimlet"]
 priority = 5
-max-sizes = {flash = 32768, ram = 16384 }
-stacksize = 8000
+max-sizes = {flash = 32768, ram = 8192 }
+stacksize = 4504
```
jgallagher, line 124: :tada: !

task/thermal/src/bsp/gimlet_b.rs
mkeeter: Good question – I don't think so, because we're tagging readings with `SensorId` when posting to the `sensors` task.

PR #840, opened 2022-10-10
Standardize on a safe wrapper around std::env::var

mkeeter:
This switches all uses of `std::env::var` to `build_env::env_var`, which ensures that Cargo will rebuild if the environment variable changes.

In addition, I added `build_util::out_dir`, which is a very common case.


cbiffle (approved):
You've introduced some unused import warnings with this change, please fix them -- otherwise looks good.

Issue #830, opened 2022-10-06
`thermal` task puts too much on the stack


mkeeter:
The `thermal` task's `Bsp` is too large.  It's awkward, because I2C handles are constructed through non-const functions (so it can't all live in `const` space), but it's _mostly_ configuration-style data.

mkeeter:
Fixed in #838

PR #841, opened 2022-10-11
max5970 + Rev C

bcantrill:
I'm not wild about the increase in memory footprint here. It should be possible to reduce in a similar way to #830, but it's a little messier because of some of the implementation details of the drivers (e.g., the adm1272) which cache state.  Memory footprint increase aside, I would like to get this in because of the Gimlet Rev C support.


drv/i2c-devices/src/max5970.rs
```diff
@@ -232,8 +240,81 @@ impl Max5970 {
...
+                self.read_reg(Register::adc_chx_mon_lsb_ch1)?,
+            )
+        } else {
+            (
+                self.read_reg(Register::adc_chx_mon_msb_ch2)?,
+                self.read_reg(Register::adc_chx_mon_lsb_ch2)?,
+            )
+        };
+
+        Ok(Volts(((((msb as u16) << 2) | (lsb as u16)) as f32) / 64.0))
```
mkeeter, line 263: I notice that `read_iout` checks the status register to decide on the divider, but this function locks it to 64.

Should we check the divider here, e.g. (untested)
```rust
        let mon_range = self.read_reg(Register::mon_range)?;

        let range = if self.rail == 0 {
            mon_range & 0b11
        } else {
            (mon_range >> 2) & 0b11
        };

        let full_scale_voltage = match range {
            0b00 => 16,
            0b01 => 8,
            0b10 => 4,
            0b11 => 2,
        };
        
        let scale = (1024 / full_scale_voltage) as f32;
        let value = ((msb as u16) << 2) | (lsb & 0b11 as u16);
        Ok(Volts(value / scale))
```

(I'm also explicitly masking `lsb`, to make it obvious that we're intentionally only using two bits)

task/power/src/main.rs
```diff
@@ -215,8 +237,8 @@ fn controllers() -> [PowerController; 13] {
     ]
 }
 
-#[cfg(target_board = "gimlet-b")]
-fn controllers() -> [PowerController; 15] {
+#[cfg(any(target_board = "gimlet-b", target_board = "gimlet-c"))]
+fn controllers() -> [PowerController; 37] {
```
mkeeter, line 241: 🤯

mkeeter (approved):

drv/i2c-devices/src/max5970.rs
bcantrill: Done -- thank you!

PR #843, opened 2022-10-11
Add files for gimlet-rot RevC

labbott:
Featuring chips with TrustZone

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        build: [stm32f3, stm32f4, lpc55, stm32h743, stm32h753, gemini, rot-carrier, gimlet-b, sidecar-a, psc-a, stm32g0, gimlet-rot]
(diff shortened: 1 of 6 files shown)


mkeeter:
To match the naming scheme used by other boards, TOML files should be named
- `app-b/c.toml`
- `stage0-b/c.toml`

(i.e. renaming the previous `app.toml` and `stage0.toml`)

I'd also like to see the previous `board` renamed from `gimlet-rot-1` to `gimlet-rot-b`, which matches all the other board names.

We'll need to add `gimlet-rot-c` to [`cargo xtask flash`](https://github.com/oxidecomputer/hubris/blob/8c5db0da00912b368a2295c49c0f13ae4a46aefc/build/xtask/src/flash.rs#L148) (multiple locations in the file, search for "gimlet-rot-1")

app/gimlet-rot/stage0-revc.toml
```diff
@@ -0,0 +1,28 @@
+name = "gimlet-rot"
+target = "thumbv8m.main-none-eabihf"
+board = "gimlet-rot-1"
```
mkeeter, line 3: This should be `gimlet-rot-c`

PR #844, opened 2022-10-12
Add startup code to configure TMP451 THERM limit

mkeeter:
In the discussions at https://github.com/oxidecomputer/quartz/issues/24, Arjen suggested configuring the TMP451 to a lower cutoff point so that the sequencer can cut power if the Tofino overheats.  This PR does that configuration as part of Sidecar sequencer setup.


arjenroodselaar (approved):
Looks good. I think we need to do some work clearing this signal when it trips before restarting the sequencer, but I may be misremembering this. Either way this needs testing so we can consider the restart case at that point.

PR #846, opened 2022-10-13
Add correct memory map for LPC55S28 variant

labbott:
The LPC55S28 variant has a smaller amount of flash/ram. Add an appropriate map.


flihp (approved):
Tested successfully w/ latest DICE certification branch. Thanks Laura!

PR #849, opened 2022-10-14
Reduce stack and flash usage of `power` task

mkeeter:
Same basic idea as #830 , except that many power devices have a local state cache, so we keep the devices in RAM (moving the configuration to flash).

It's not _as big_ of a win, but is still non-trivial:
- Flash goes from 17804 -> 16588 bytes
- Stack depth goes from 4160 -> 3880

Alas, neither of these changes pushes us below a power of two, but they're still probably worthwhile.

### DIFF
--- a/lib/mutable-statics/src/lib.rs
+++ b/lib/mutable-statics/src/lib.rs
@@ -11,8 +11,9 @@
 /// &mut references to statics, which are declared within the macro.
 ///
 /// The macro accepts definitions of one or more mutable static arrays. It will
-/// arrange for them to be initialized a per-array lambda function, and return a
-/// tuple containing mutable references to each, in the order they're declared.
+/// arrange for them to be initialized by a per-array lambda function, and
+/// return a tuple containing mutable references to each, in the order they're
+/// declared.
 #[macro_export]
 macro_rules! mutable_statics {
     (
--- a/lib/mutable-statics/src/lib.rs
+++ b/lib/mutable-statics/src/lib.rs
@@ -11,8 +11,8 @@
 /// &mut references to statics, which are declared within the macro.
 ///
 /// The macro accepts definitions of one or more mutable static arrays. It will
-/// arrange for them to be initialized, and return a tuple containing mutable
-/// references to each, in the order they're declared.
+/// arrange for them to be initialized a per-array lambda function, and return a
+/// tuple containing mutable references to each, in the order they're declared.
 #[macro_export]
 macro_rules! mutable_statics {
     (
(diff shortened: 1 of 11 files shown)


task/power/src/main.rs
```diff
@@ -39,117 +39,100 @@ include!(concat!(env!("OUT_DIR"), "/i2c_config.rs"));
...
+            Device::Max5970(dev) => dev.read_vout()?,
+        };
+        Ok(r)
+    }
+}
+
+impl PowerControllerConfig {
+    fn get_device(&self, task: TaskId) -> Device {
+        let (dev, rail) = (self.builder)(task);
+        use DeviceType::*;
```
jgallagher, line 113: Style nit - I'm not a fan of importing enum variants like this; match arms with just a single identifier open the door for typos to bind new variable names (e.g., https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7d0461a815e49ffbfe83c0f6a9b48bea). Usually (maybe always?) this produces warnings, but it will still compile.

task/power/src/main.rs
```diff
@@ -349,14 +297,17 @@ fn get_state() -> PowerState {
 fn main() -> ! {
     let sensor = sensor_api::Sensor::from(SENSOR.get_task_id());
 
-    let mut controllers = controllers();
+    let i2c_task = I2C.get_task_id();
+
+    let mut devices: [Device; CONTROLLER_CONFIG.len()] =
+        array_init::array_init(|i| CONTROLLER_CONFIG[i].get_device(i2c_task));
```
jgallagher, line 303: Do we need all the devices on the stack, or could we create them on demand in the loop below? Would only be a win if the creation is very cheap (in both text and runtime), but at a glance it looks like it might be, and would let us buy back a little chunk of stack (although maybe not enough to bother).

jgallagher (approved):

task/power/src/main.rs
cbiffle: You might also consider using `lib/mutable-statics` to put this in static memory, where its usage is more cleanly accounted for. (This would also require some dancing with either a default value or MaybeUninit; I should add an `array_init`-style version of the mutable static types....)

mkeeter: Some of the devices have local caches of values, so we do need them to be persistent.

(only _some_ of them, but separating them into persistent vs transient would be a bit of a mess)

task/power/src/main.rs
mkeeter: Okay, I'm persuaded and will use full names here

lib/mutable-statics/src/lib.rs
```diff
@@ -11,8 +11,8 @@
 /// &mut references to statics, which are declared within the macro.
 ///
 /// The macro accepts definitions of one or more mutable static arrays. It will
-/// arrange for them to be initialized, and return a tuple containing mutable
-/// references to each, in the order they're declared.
+/// arrange for them to be initialized a per-array lambda function, and return a
```
cbiffle, line 14: +by

PR #852, opened 2022-10-15
Add M.2 NVMe devices

mkeeter:


mkeeter:
Disregard in favor of #851

Issue #668, opened 2022-07-27
Improve flexibility of clock generator configs


cbiffle:
Currently, gimlet's clock gen config is in Rust literals in `drv/gimlet-seq-server/src/payload.rs`. It can be regenerated by running a Humility command, which is a little odd since it has nothing to do with debugging.

At minimum, we need to alter this so that we can have a different clock gen config _per revision of the board._

I would argue that, ideally, we would also move the code generation out of Humility and into the build system, so that we could just check in the Aardvark file that defines the sequence of config writes. This would make it dramatically easier for folks doing bringup to substitute different configurations in Aardvark format, which is a need @refugeesus ran into this afternoon.

If we did this, we should switch Sidecar to use the same mechanism, as it contains a similar canned file.

cbiffle:
We have at least extended this to support multiple revisions in #853, so I think the time pressure's off on cleaning this up.

PR #853, opened 2022-10-15
Add revC clock config and change to support different configs per build

nathanaelhuffman:
This is the quick and dirty way of supporting two different clock configs for different boards without pulling in all the humility rencm generate stuff.  See #668  as a reference for where we might like to land eventually.

Credit to @mkeeter who provided the suggestion which is wholesale cribbed here.

The original config was renamed and the new revC config added supporting the oscillator frequency used on revC gimlets.


mkeeter (approved):
Assuming the payload is correct, this seems simple enough!

PR #860, opened 2022-10-19
Save and restore timer deadline in sleep_until

mkeeter:
Without this fix, `hl::sleep_for` will (confusingly) wipe out pending deadlines.


cbiffle (approved):
This is an improvement, I think. It's too bad that `sleep_for` -- the variant most code appears to use -- is now five syscalls. We may want to rethink the timer syscall API if this use case is common.

PR #863, opened 2022-10-19
Enable Tofino debug port, allow flashing SPI EEPROM using Humility

arjenroodselaar:
This diff adds the plumbing to use the Tofino I2C debug port and program the attached SPI EEPROM. There is ample opportunity for improvements, but this is a first pass which works.


drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,435 @@ impl Sequencer {
         })
     }
 }
+
+bitfield! {
+    #[derive(Copy, Clone, PartialEq, FromPrimitive, AsBytes)]
+    #[repr(C)]
+    pub struct DebugPortState(u8);
+    pub send_buffer_empty, set_send_buffer_empty: 0;
```
mkeeter, line 240: Do we need any of these setters (`set_send_buffer_empty`, `set_receive_buffer_empty`, etc)?

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,435 @@ impl Sequencer {
...
+
+#[derive(Copy, Clone, PartialEq, FromPrimitive, AsBytes)]
+#[repr(u8)]
+pub enum DebugRequestOpcode {
+    LocalWrite = 0,
+    LocalRead = 32,
+    DirectWrite = 128,
+    DirectRead = 160,
+    IndirectWrite = 192,
+    IndirectRead = 224,
```
mkeeter, line 257: Maybe put these in binary to make the pattern more clear?
```suggestion
    LocalWrite = 0b0000_0000,
    LocalRead = 0b0010_0000,
    DirectWrite = 0b1000_0000,
    DirectRead =  0b1010_0000,
    IndirectWrite = 0b1100_0000,
    IndirectRead = 0b1110_0000,
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,435 @@ impl Sequencer {
...
+    fn from(opcode: DebugRequestOpcode) -> Self {
+        opcode as u8
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, FromPrimitive, AsBytes)]
+#[repr(u8)]
+pub enum DirectBarSegment {
+    Bar0 = 0,
+    Msi = 1,
```
mkeeter, line 270: It's weird to me that we explicitly specify values, but those values _aren't_ what we use for `impl From<DirectBarSegment> for u32`.

I'd vote to either
(a) make this a simple `enum` (no `repr` or assigned values), _or_
(b) make this `#[repr(u32)]` and use the `u32` values below

(leaning slightly towards the latter)

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,435 @@ impl Sequencer {
...
+            DirectBarSegment::Bar0 => 0,
+            DirectBarSegment::Msi => (1 as u32) << 28,
+            DirectBarSegment::Cfg => (2 as u32) << 28,
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, FromPrimitive, AsBytes)]
+#[repr(u32)]
+pub enum TofinoRegisters {
```
mkeeter, line 286: Add a comment explaining that this comes from the Tofino register datasheet

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,435 @@ impl Sequencer {
...
+
+impl From<TofinoRegisters> for u32 {
+    fn from(r: TofinoRegisters) -> Self {
+        r as u32
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, FromPrimitive, AsBytes)]
+#[repr(u8)]
+pub enum SpiEepromInstruction {
```
mkeeter, line 304: Add a comment explaining that this comes from the EEPROM, and maybe add the command acronyms (`WREN`, etc) in comments?

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,435 @@ impl Sequencer {
...
+    Write = 0x2,
+}
+
+impl From<SpiEepromInstruction> for u8 {
+    fn from(i: SpiEepromInstruction) -> Self {
+        i as u8
+    }
+}
+
+impl From<SpiEepromInstruction> for u32 {
```
mkeeter, line 319: (I'm going to continue reading, but am curious if we need both `u8` and `u32` conversions here)

drv/sidecar-mainboard-controller/src/tofino2.rs
arjenroodselaar: I used these while debugging the FPGA logic. They're currently not used, but we may have some clean-up logic in the future which may want to use these.

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,435 @@ impl Sequencer {
...
+
+    pub fn state(&self) -> Result<DebugPortState, FpgaError> {
+        self.fpga
+            .read::<u8>(Addr::TOFINO_DEBUG_PORT_STATE)
+            .map(DebugPortState)
+    }
+
+    pub fn set_state(&self, state: DebugPortState) -> Result<(), FpgaError> {
+        self.fpga
+            .write(WriteOp::Write, Addr::TOFINO_DEBUG_PORT_STATE, state.0)
```
mkeeter, line 347: We should remove the `FromBytes` constraint [here](https://github.com/oxidecomputer/hubris/blob/5c9390113461db06e0ec88ffbab8cb0a295afdff/drv/fpga-api/src/lib.rs#L289), which doesn't really make sense for `write`.

Then, we wouldn't need `state.0` (we could just pass `state` instead)

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,435 @@ impl Sequencer {
...
+                task_id,
+                MainboardController::DEVICE_INDEX,
+            ),
+        }
+    }
+
+    pub fn state(&self) -> Result<DebugPortState, FpgaError> {
+        self.fpga
+            .read::<u8>(Addr::TOFINO_DEBUG_PORT_STATE)
+            .map(DebugPortState)
```
mkeeter, line 342: If you add `FromBytes` to the `#[derive(...)]` block of `DebugPortState`, this could become simply
```rust
        self.fpga.read(Addr::TOFINO_DEBUG_PORT_STATE)
```
(since it would derive the return type from the function signature)

drv/sidecar-mainboard-controller/src/tofino2.rs
arjenroodselaar: Yes, this is only used in the context where you want that base address. So using the desired address as the representation makes sense.

drv/sidecar-mainboard-controller/src/tofino2.rs
arjenroodselaar: The `u32` `From` is only used in one place, so I replaced that with a `u8::from(..) as u32`.

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+        offset: impl Into<u32>,
+    ) -> Result<u32, FpgaError> {
+        let state = self.state()?;
+
+        if !state.send_buffer_empty() || !state.receive_buffer_empty() {
+            return Err(FpgaError::InvalidState);
+        }
+
+        // Add the segement base address to the given read offset.
+        let address = segment as u32 | offset.into();
```
mkeeter, line 356: We should check that `offset.into()` is less than `(1 << 28)`, so it can't hit any bits in `segment` (and return an error code if it's too big)

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+        )?;
+
+        // Write the address.
+        for b in address.as_bytes().iter() {
+            self.fpga.write(
+                WriteOp::Write,
+                Addr::TOFINO_DEBUG_PORT_BUFFER,
+                *b,
+            )?;
+        }
```
mkeeter, line 372: `u32` already implements `as_bytes`, so I think we can avoid an explicit loop:
```suggestion
        // Write the address.
        self.fpga.write(
            WriteOp::Write,
            Addr::TOFINO_DEBUG_PORT_BUFFER,
            address,
        )?;
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+        // Wait for the request to complete.
+        while self.state()?.request_in_progress() {
+            userlib::hl::sleep_for(1);
+        }
+
+        // Read the response.
+        let mut v: u32 = 0;
+        for b in v.as_bytes_mut().iter_mut() {
+            *b = self.fpga.read(Addr::TOFINO_DEBUG_PORT_BUFFER)?;
+        }
```
mkeeter, line 390: Similarly,
```suggestion
        let v: u32 = self.fpga.read(Addr::TOFINO_DEBUG_PORT_BUFFER)?;
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+
+        Ok(v)
```
mkeeter, line 392: We can replace the loop here with a single `read` call, since it will derive the correct type:
```suggestion
        self.fpga.read(Addr::TOFINO_DEBUG_PORT_BUFFER)
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+        }
+
+        // Add the segement base address to the given read offset.
+        let address = segment as u32 | offset.into();
+
+        // Write the opcode.
+        self.fpga.write(
+            WriteOp::Write,
+            Addr::TOFINO_DEBUG_PORT_BUFFER,
+            u8::from(DebugRequestOpcode::DirectRead),
```
mkeeter, line 362: `DebugRequestOpcode` already derives `AsBytes`, so we don't need to cast it directly:
```suggestion
            DebugRequestOpcode::DirectRead,
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+        }
+
+        // Add the segement base address to the given read offset.
+        let address = segment as u32 | offset.into();
+
+        // Write the opcode to the queue.
+        self.fpga.write(
+            WriteOp::Write,
+            Addr::TOFINO_DEBUG_PORT_BUFFER,
+            u8::from(DebugRequestOpcode::DirectWrite),
```
mkeeter, line 414:
```suggestion
            DebugRequestOpcode::DirectWrite,
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+        }
+
+        // Write the value to the queue.
+        for b in value.as_bytes().iter() {
+            self.fpga.write(
+                WriteOp::Write,
+                Addr::TOFINO_DEBUG_PORT_BUFFER,
+                *b,
+            )?;
+        }
```
mkeeter, line 433: Same as before, we can factor out the explicit loops
```suggestion
        // Write the address to the queue.
        self.fpga.write(
            WriteOp::Write,
            Addr::TOFINO_DEBUG_PORT_BUFFER,
            address,
        )?;

        // Write the value to the queue.
        self.fpga.write(
            WriteOp::Write,
            Addr::TOFINO_DEBUG_PORT_BUFFER,
            values,
        )?;
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+        // Wait for the request to complete.
+        while self.state()?.request_in_progress() {
+            userlib::hl::sleep_for(1);
+        }
+
+        Ok(())
+    }
+
+    /// Generate the SPI command Tofino needs to complete a SPI request.
+    fn spi_command(n_bytes_to_write: usize, n_bytes_to_read: usize) -> u32 {
```
mkeeter, line 451: Add some assertions that `n_bytes_to_read/write` are correctly sized here

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -231,3 +232,429 @@ impl Sequencer {
...
+        )?;
+        // Initiate the SPI transaction.
+        self.write_direct(
+            DirectBarSegment::Bar0,
+            TofinoRegisters::SpiCommand,
+            Self::spi_command(1, 1),
+        )?;
+
+        // Wait for the SPI command to complete.
+        while self
```
mkeeter, line 477: We do this enough that it should probably be a helper function, e.g. `self.spi_wait()?;`

drv/sidecar-mainboard-controller/src/tofino2.rs
arjenroodselaar: Good idea, thanks.

drv/sidecar-mainboard-controller/src/tofino2.rs
arjenroodselaar: I've added comments explaining why this is done in a loop. Similar for all the cases below.

PR #864, opened 2022-10-19
Only import Pin when needed

mkeeter:
Fixes stuff like [this wall of warnings](https://github.com/oxidecomputer/hubris/pull/861/files#file-drv-lpc55-spi-server-src-main-rs-L11)


labbott (approved):
Thanks for cleaning this up!

PR #865, opened 2022-10-19
Hubris versions

labbott:


drv/lpc55-update-server/src/main.rs
```diff
@@ -158,6 +158,16 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
         Ok(BLOCK_SIZE_BYTES)
     }
+
+    fn current_version(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<ImageVersion, RequestError<UpdateError>> {
+        Ok(ImageVersion {
+            epoch: HUBRIS_BUILD_EPOCH,
+            version: HUBRIS_BUILD_EPOCH,
```
andrewjstone, line 168: Is this a typo?

lzrd (approved):
This looks good. I see with these changes that versions as epoch and version exist. No policies are implemented, but versions can be queried through the necessary low level update_server API.

drv/lpc55-update-server/src/main.rs
labbott: *sigh* yes

PR #867, opened 2022-10-20
net: RX activity watchdog

cbiffle:
We're hitting an intermittent bug on the stm32h7 where the Ethernet MAC
stops delivering packets from the MTL, without any status bits to
indicate it or any obvious way to get it unstuck. Resetting the device
works, however!

This changes the netstack to reset itself (by panicking) if no traffic
has emerged from the MAC within a time window (currently 10 seconds).
Note that this should cover any traffic on the wire, _not just_ traffic
addressed to the node, so 10 seconds isn't as aggressive as it might
sound.

It would be nice to reset just the Ethernet driver instead of losing
netstack state, which includes any packets received to sockets but not
yet handed out to tasks. Maybe later.


lib/multitimer/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+    pub fn set_timer(
+        &mut self,
+        which: E,
+        deadline: u64,
+        repeat: Option<Repeat>,
+    ) {
+        // If the timer has previously fired without us noticing it, preserve
+        // that across set.
+        let fired_but_not_observed = self.timers[which].fired_but_not_observed;
+        self.timers[which] = Timer {
```
mkeeter, line 72: Vague question: is it better to save `fired_but_not_observed` then build a whole new `Timer`, vs modifying the deadline on the existing timer, e.g.
```rust
self.timers[which].deadline = Some((deadline, repeat))
```

lib/multitimer/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Repeat {
+    AfterWake(u64),
+    AfterDeadline(u64),
+}
+
+// Syscall fakes for testing!
+
+#[cfg(not(target_os = "none"))]
```
mkeeter, line 206: We could put all this into a single `mod fakes` (then import `super::fakes::*` in the test suite), to reduce repeated `#[cfg(not(target_os = "none"))]`

mkeeter (approved):

lib/multitimer/src/lib.rs
cbiffle: I could go either way -- my current approach feels more robust if Timer grows new fields in the future, while your approach is simpler with the current definition.

lib/multitimer/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+            if core::mem::replace(&mut timer.fired_but_not_observed, false) {
+                Some(e)
+            } else {
+                None
+            }
+        })
+    }
+}
+
+trait SomeTimer {
```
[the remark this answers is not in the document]
cbiffle: Good catch! That's a leftover from my early prototyping. Will remove.

lib/multitimer/src/lib.rs
cbiffle: Required a top-level conditional `use`, but, I think this is cleaner. Will change.

PR #868, opened 2022-10-20
Initial SP inventory reporting to MGS

jgallagher:
Adds functionality for `control_plane_agent` to send a list of devices to MGS. This is analogous to running `humility validate` remotely (and much of the same information is present), with some differences in the details available per device and that we report a small number of high-level devices that aren't present in `humility validate` (all the `SpComponent::*` constants we use in updates: `SP_ITSELF`, `SP_AUX_FLASH`, etc.). This does mean we're baking in a bunch of human-readable strings (a few KiB for gimlets, at the moment), but that seems worth it for the SP to be fully self-reporting.

This PR also adds a `validate` task to the PSC app.toml.


task/control-plane-agent/src/inventory.rs
```diff
@@ -0,0 +1,201 @@
...
+        // If `index` is in `0..self.ours.num_devices()`, defer to it;
+        // otherwise, subtract `self.ours.num_devices()` to shift it into the
+        // range `0..VALIDATE_DEVICES.len()` and ask `validate`.
+        let index = match index.checked_sub(self.ours.num_devices()) {
+            // Subtraction failure; `index` is in `0..self.ours.num_devices()`.
+            None => return self.ours.device_description(index),
+            // Subtraction success; `index` has been shifted down to be in the
+            // range `0..VALIDATE_DEVICES.len()`.
+            Some(index) => index,
+        };
```
mkeeter, line 55: This seems too clever – maybe just a regular conditional instead?

```suggestion
        // If `index` is in `0..self.ours.num_devices()`, defer to it;
        // otherwise, subtract `self.ours.num_devices()` to shift it into the
        // range `0..VALIDATE_DEVICES.len()` and ask `validate`.
        let index = if index < self.ours.num_devices() {
            return self.ours.device_description(index);
        } else {
            index - self.ours.num_devices()
        };
```

(this calls `self.ours.num_devices()` twice, but the compiler will probably strip it out)

task/control-plane-agent/src/inventory.rs
```diff
@@ -0,0 +1,201 @@
...
+            Err(ValidateError::Unavailable | ValidateError::DeviceOff) => {
+                DevicePresence::Unavailable
+            }
+            Err(ValidateError::DeviceTimeout) => DevicePresence::Timeout,
+            Err(ValidateError::InvalidDevice | ValidateError::DeviceError) => {
+                DevicePresence::Error
+            }
+        };
+
+        let mut component = FmtComponentId::default();
```
mkeeter, line 76: Add a comment here explaining why this won't fail (I assume it's checked by `ASSERT_EACH_DEVICE_FITS_IN_ONE_PACKET` above?)

task/control-plane-agent/src/inventory.rs
```diff
@@ -0,0 +1,201 @@
...
+        component
+            .write_fmt(format_args!(
+                "{}{}",
+                SpComponent::GENERIC_DEVICE_PREFIX,
+                index
+            ))
+            .unwrap_lite();
```
mkeeter, line 83: The docs say that you don't usually need to call `write_fmt` directly; sure enough, this compiles with `write!` instead:

```suggestion
        write!(
            &mut component,
            "{}{}",
            SpComponent::GENERIC_DEVICE_PREFIX,
            index
        )
        .unwrap_lite();
```

task/control-plane-agent/src/inventory.rs
```diff
@@ -0,0 +1,201 @@
...
+
+#[derive(Default)]
+struct FmtComponentId {
+    pos: usize,
+    id: [u8; SpComponent::MAX_ID_LENGTH],
+}
+
+impl fmt::Write for FmtComponentId {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        let remaining = &mut self.id[self.pos..];
```
mkeeter, line 107: This is very elegant!

task/control-plane-agent/src/inventory/our_inventory.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// List of logical or high-level components that this task is responsible for
+// (or at least responds to in terms of MGS requests for status / update, even
+// if another task is actually responsible for lower-level details).
+//
+// TODO: Are our device names and descriptions good enough, or are there more
+//       specific names we should use? This may be answered when we expand
+//       DeviceDescription with any VPD / serial numbers.
+const DEVICES: &'static [DeviceDescription<'static>] = &[
```
mkeeter, line 39: We don't need `'static` here.  Oddly, we _do_ need it above (in `device_description`), despite `DeviceDescription` not having an associated lifetime.

task/control-plane-agent/src/inventory.rs
```diff
@@ -0,0 +1,201 @@
...
+    loop {
+        if i == VALIDATE_DEVICES.len() {
+            break;
+        }
+        assert_device_tlv_fits_in_one_packet(
+            VALIDATE_DEVICES[i].device,
+            VALIDATE_DEVICES[i].description,
+        );
+        i += 1;
+    }
```
mkeeter, line 200: I'd vote to also check `DEVICES` here, since I was confused when initially reading this function.

This requires making `our_inventory::DEVICES` into `pub(super)`, but then you can delete the checks from `our_inventory` and have them all here, e.g. something like this:

```suggestion
    use our_inventory::DEVICES;
    let mut i = 0;
    loop {
        if i == DEVICES.len() {
            break;
        }
        assert_device_tlv_fits_in_one_packet(
            DEVICES[i].device,
            DEVICES[i].description,
        );
        i += 1;
    }
```

task/control-plane-agent/src/inventory/our_inventory.rs
```diff
@@ -0,0 +1,111 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Submodule that defines logical components we claim to have (as opposed to
+//! components entirely managed by other tasks, like I2C devices).
+
+use gateway_messages::{
+    sp_impl::DeviceDescription, DeviceCapabilities, DevicePresence, SpComponent,
+};
+
+pub(super) struct OurInventory {}
```
mkeeter, line 12: If we're making `DEVICES` `pub(super)` to consolidate the compile-time length checks, then this `struct` seems superfluous versus storing `DEVICES` directly in the `struct Inventory`.

```rust
ours: OurInventory,
```
becomes
```rust
ours: &'static [DeviceDescriptor], // do we need `<'static>` here?
```

task/control-plane-agent/src/inventory.rs
jgallagher: Almost; it's checked by the stuff in the `max_num_devices` submodule. Will add a comment to that effect.

task/control-plane-agent/src/inventory.rs
jgallagher: Nice! For some reason I thought `write!()` wasn't available in `core`, but it definitely is.

task/control-plane-agent/src/inventory/our_inventory.rs
jgallagher: We can remove the `'static` here because the [elided lifetimes in paths](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#elided-lifetimes-in-paths) lint is "allowed by default". I don't like elided lifetimes in paths, so I made it explicit here. :)

mkeeter: Okay, I'm fine keeping it here!

mkeeter (approved):

PR #870, opened 2022-10-24
Use a worker thread to constantly echo stderr when building

mkeeter:
This fixes an issue introduced in #740 where we didn't see `cargo` output until a task _finished_ building, which made it less obvious where the time was being spent.


steveklabnik (approved):
Looks great and works well here!

PR #878, opened 2022-10-26
netstack refactoring

cbiffle:
These commits...

- Formalize the relationship of the BSP modules with the rest of the netstack by defining a `Bsp` trait
- Simplify the `server_vlan` setup code to do less parallel array iteration and asserting
- Make the `server_vlan` implementation the sole implementation, by treating the VLAN-less case as a degenerate single-network case of the VLAN server.

This reduces code duplication and opportunities for divergence, and makes it easier to reason about whether non-target-specific code (like `src/main.rs`) will work with all the various platform-specific bits (since it only needs to be checked against traits and concrete implementations now).

### DIFF
--- a/task/net/build.rs
+++ b/task/net/build.rs
@@ -49,10 +49,14 @@ fn generate_net_config(
         writeln!(
             out,
             "{}",
-            generate_socket_state(name, socket, config.vlan.map(|v| v.count))?
+            generate_socket_state(
+                name,
+                socket,
+                config.vlan.map(|v| v.count).unwrap_or(1)
+            )?
         )?;
     }
-    writeln!(out, "{}", generate_state_struct(config)?)?;
+    writeln!(out, "{}", generate_state_struct(config))?;
     writeln!(out, "{}", generate_constructor(config)?)?;
     writeln!(out, "{}", generate_owner_info(config)?)?;
     writeln!(out, "{}", generate_port_table(config)?)?;
@@ -112,14 +116,14 @@ fn generate_owner_info(
 fn generate_socket_state(
     name: &str,
     config: &SocketConfig,
-    vlan_count: Option<usize>,
+    vlan_count: usize,
 ) -> Result<TokenStream, Box<dyn std::error::Error>> {
     if config.kind != "udp" {
         return Err("unsupported socket kind".into());
     }
 
-    let tx = generate_buffers(name, "TX", &config.tx, vlan_count)?;
-    let rx = generate_buffers(name, "RX", &config.rx, vlan_count)?;
+    let tx = generate_buffers(name, "TX", &config.tx, vlan_count);
+    let rx = generate_buffers(name, "RX", &config.rx, vlan_count);
     Ok(quote::quote! {
         #tx
         #rx
@@ -130,8 +134,8 @@ fn generate_buffers(
     name: &str,
     dir: &str,
     config: &BufSize,
-    vlan_count: Option<usize>,
-) -> Result<TokenStream, Box<dyn std::error::Error>> {
+    vlan_count: usize,
+) -> TokenStream {
     let pktcnt = config.packets;
     let bytecnt = config.bytes;
     let upname = name.to_ascii_uppercase();
@@ -139,41 +143,25 @@ fn generate_buffers(
         syn::parse_str(&format!("SOCK_{}_HDR_{}", dir, upname)).unwrap();
     let bufname: syn::Ident =
         syn::parse_str(&format!("SOCK_{}_DAT_{}", dir, upname)).unwrap();
-    Ok(if let Some(vlan_count) = vlan_count {
-        quote::quote! {
-            static mut #hdrname: [[UdpPacketMetadata; #pktcnt]; #vlan_count] = [
-                [UdpPacketMetadata::EMPTY; #pktcnt]; #vlan_count
-            ];
-            static mut #bufname: [[u8; #bytecnt]; #vlan_count] = [[0u8; #bytecnt]; #vlan_count];
-        }
-    } else {
-        quote::quote! {
-            static mut #hdrname: [UdpPacketMetadata; #pktcnt] =
-                [UdpPacketMetadata::EMPTY; #pktcnt];
-            static mut #bufname: [u8; #bytecnt] = [0u8; #bytecnt];
-        }
-    })
+    quote::quote! {
+        static mut #hdrname: [[UdpPacketMetadata; #pktcnt]; #vlan_count] = [
+            [UdpPacketMetadata::EMPTY; #pktcnt]; #vlan_count
+        ];
+        static mut #bufname: [[u8; #bytecnt]; #vlan_count] = [[0u8; #bytecnt]; #vlan_count];
+    }
 }
 
-fn generate_state_struct(
-    config: &NetConfig,
-) -> Result<TokenStream, Box<dyn std::error::Error>> {
+fn generate_state_struct(config: &NetConfig) -> TokenStream {
     let n = config.sockets.len();
-    Ok(if let Some(vlan_count) = config.vlan.map(|v| v.count) {
-        quote::quote! {
-            pub(crate) struct Sockets<'a>(pub [[UdpSocket<'a>; #n]; #vlan_count]);
-        }
-    } else {
-        quote::quote! {
-            pub(crate) struct Sockets<'a>(pub [UdpSocket<'a>; #n]);
-        }
-    })
+    quote::quote! {
+        pub(crate) struct Sockets<'a, const N: usize>(pub [[UdpSocket<'a>; #n]; N]);
+    }
 }
 
 fn generate_constructor(
     config: &NetConfig,
 ) -> Result<TokenStream, Box<dyn std::error::Error>> {
-    let name_to_sockets = |name: &String, i: Option<usize>| {
+    let name_to_sockets = |name: &String, i: usize| {
         let upname = name.to_ascii_uppercase();
         let rxhdrs: syn::Ident =
             syn::parse_str(&format!("SOCK_RX_HDR_{}", upname)).unwrap();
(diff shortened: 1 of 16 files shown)


task/net/build.rs
```diff
@@ -49,10 +49,14 @@ fn generate_net_config(
         writeln!(
             out,
             "{}",
-            generate_socket_state(name, socket, config.vlan.map(|v| v.count))?
+            generate_socket_state(
+                name,
+                socket,
+                config.vlan.map(|v| v.count).unwrap_or(1)
```
mkeeter, line 55: How do you feel about `config.vlan.map_or(1, |v| v.count)`?

John introduced me to it, and it's a bit shorter, but doesn't read quite as naturally (maybe just because I haven't seen it as often).

task/net/build.rs
```diff
@@ -184,59 +172,37 @@ fn generate_constructor(
...
-            })
-            .collect::<Vec<_>>()
-    } else {
-        config
-            .sockets
-            .keys()
-            .map(|n| name_to_sockets(n, None))
-            .collect::<Vec<_>>()
-    };
+    let vlan_count = config.vlan.map(|v| v.count).unwrap_or(1);
```
mkeeter, line 188: same here (using `map_or`)

task/net/src/bsp/gimlet_bc.rs
```diff
@@ -2,6 +2,9 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+#[cfg(any(not(feature = "ksz8463"), not(feature = "mgmt")))]
```
mkeeter, line 5: I don't feel strongly about it, but this could be De Morgan'd into
```suggestion
#[cfg(not(all(feature = "ksz8463", feature = "mgmt"))]
```
which is slightly shorter

task/net/src/bsp/gimletlet_mgmt.rs
```diff
@@ -2,7 +2,10 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use crate::{mgmt, miim_bridge::MiimBridge, pins};
+#[cfg(any(not(feature = "ksz8463"), not(feature = "mgmt")))]
```
mkeeter, line 5: Same here
```suggestion
#[cfg(not(all(feature = "ksz8463", feature = "mgmt"))]
```

task/net/src/bsp_support.rs
```diff
@@ -0,0 +1,86 @@
...
+//! # How the BSP stuff works
+//!
+//! The netstack defines a `bsp` module, but chooses which source file to use
+//! based on the board ID/rev, selecting among options in `src/bsp/`.
+//!
+//! A BSP module is expected to export a single type, called `BspImpl`, which
+//! implements the `Bsp` trait from this module.
+
+use drv_stm32h7_eth as eth;
+// This is not a typo: we use the vsc7448 type on all BSPs.
```
mkeeter, line 18: This is probably meant to be a few lines down

task/net/src/main.rs
```diff
@@ -241,7 +245,7 @@ fn main() -> ! {
                     Timers::Watchdog => panic!("MAC RX watchdog"),
                 }
             }
-            let mut msgbuf = [0u8; ServerImpl::INCOMING_SIZE];
+            let mut msgbuf = [0u8; ServerImpl::<BspImpl>::INCOMING_SIZE];
```
mkeeter, line 248: We could use `idl::INCOMING_SIZE` here, then remove `INCOMING_SIZE` from the `ServerImpl` entirely.

task/net/src/server.rs
```diff
@@ -240,3 +233,343 @@ impl<T: NetServer> idl::InOrderNetImpl for T {
...
+    iface: Interface<'static, E>,
+}
+
+impl<'a, B, E, const N: usize> GenServerImpl<'a, B, E, N>
+where
+    B: bsp_support::Bsp,
+    E: DeviceExt,
+{
+    /// Size of buffer that must be allocated to use `dispatch`.
+    pub const INCOMING_SIZE: usize = idl::INCOMING_SIZE;
```
mkeeter, line 276: Since this isn't server-dependent, we could remove it here (same suggestion as above)

task/net/src/server.rs
```diff
@@ -240,3 +233,343 @@ impl<T: NetServer> idl::InOrderNetImpl for T {
...
+
+        Ok(crate::Activity { ip, mac_rx })
+    }
+
+    /// Iterate over sockets, waking any that can do work.  A task can do work
+    /// if all of the (internal) VLAN sockets can receive a packet, since
+    /// we don't know which VLAN it will write to.
+    pub fn wake_sockets(&mut self) {
+        for i in 0..SOCKET_COUNT {
+            if (0..N).any(|v| {
```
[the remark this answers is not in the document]
cbiffle: Hmmm. So this is the existing logic... and I can't really make sense of it. It mashes together the recv logic (in which case you want `any`) and the send logic (in which case you want `all`).

I'm guessing the send side of this got added fairly recently and may have a logic bug.

Lemme see what I can do here.

task/net/src/server.rs
```diff
@@ -240,3 +233,343 @@ impl<T: NetServer> idl::InOrderNetImpl for T {
...
+            .cloned()
+    }
+
+    /// Gets the socket `index`. If `index` is out of range, returns
+    /// `BadMessage`. Panics if `vlan_index` is out of range, which should
+    /// never happen (because messages with invalid VIDs are dropped in
+    /// RxRing).
+    ///
+    /// Sockets are currently assumed to be UDP.
+    pub(crate) fn get_socket_mut(
```
mkeeter, line 398: Now that we've consolidated `iface` and `socket_handles` into the `VLanState`, it seems like `get_handle` and `get_socket_mut` could be functions on that struct, e.g.
```rust
impl VLanState
    fn get_handle(&self, index: usize) -> Option<SocketHandle> {
        self.socket_handles.get(index).cloned()
    }
    fn get_socket_mut(
        &mut self,
        index: usize,
    ) -> Option<&mut UdpSocket<'static>> {
        self.iface.get_socket::<UdpSocket<'_>>(self.get_handle(index)?)
    }
}
```

This removes repeated indexing of `self.vlan_state`

task/net/src/server.rs
```diff
@@ -240,3 +233,343 @@ impl<T: NetServer> idl::InOrderNetImpl for T {
...
+    ) -> Option<SocketHandle> {
+        self.vlan_state
+            .get(vlan_index)?
+            .socket_handles
+            .get(index)
+            .cloned()
+    }
+
+    /// Gets the socket `index`. If `index` is out of range, returns
+    /// `BadMessage`. Panics if `vlan_index` is out of range, which should
```
mkeeter, line 393: Comment is out of date, since this now returns `Option`

task/net/src/bsp/gimlet_bc.rs
cbiffle: Yeah, I agree that DeMorganing this seems right.

task/net/build.rs
cbiffle: `map_or` has its arguments in the wrong order and I find it jarring.

task/net/src/server.rs
```diff
@@ -240,3 +233,343 @@ impl<T: NetServer> idl::InOrderNetImpl for T {
...
+        match socket.send(payload.len(), metadata.into()) {
+            Ok(buf) => {
+                payload
+                    .read_range(0..payload.len(), buf)
+                    .map_err(|_| RequestError::went_away())?;
+                self.client_waiting_to_send[socket_index] = false;
+                Ok(())
+            }
+            Err(smoltcp::Error::Exhausted) => {
+                self.client_waiting_to_send[socket_index] = false;
```
mkeeter, line 528: I think this should be `true`!

task/net/src/server.rs
cbiffle: Fixing.

task/net/src/server_basic.rs
```diff
@@ -8,282 +8,43 @@
...
+use smoltcp::wire::{EthernetAddress, Ipv6Address};
+use task_net_api::UdpMetadata;
 
 /// Grabs references to the server storage arrays.  Can only be called once!
-pub fn claim_server_storage_statics() -> (
-    &'static mut [NeighborStorage; NEIGHBORS],
-    &'static mut [SocketStorage<'static>; SOCKET_COUNT],
-    &'static mut [IpCidr; 1],
-) {
+fn claim_server_storage_statics() -> &'static mut [Storage; 1] {
```
mkeeter, line 20: I attempted to make this generic, i.e. `fn claim_server_storage_statics<const N: usize>()`, but ran into issues with the capture:
```
error[E0401]: can't use generic parameters from outer function
   --> task/net/src/server.rs:583:39
    |
581 | fn claim_server_storage_statics<const N: usize>() -> &'static mut [Storage; N] {
    |                                       - const parameter from outer function
582 |     mutable_statics! {
583 |         static mut STORAGE: [Storage; N] = [|| Default::default(); _];
    |                                       ^ use of generic parameter from outer function
```

Too bad...

task/net/src/bsp_support.rs
mkeeter: Ah yeah; I think that you could get away with it if you included a whitespace line separating this import from the others, but deleting it is fine by me.

task/net/src/server.rs
cbiffle: If only rustfmt would let those routines be so vertically compact. :-)

The result of this is better than you implied -- not only does it remove repeated indexing, it actually removes _the indexing_ in most cases. e.g.

```diff
         // Iterate over all of the per-VLAN sockets, returning the first
         // available packet with a bonus `vid` tag attached in the metadata.
-        for i in 0..N {
-            let socket = self
-                .get_socket_mut(socket_index, i)
+        for vlan in &mut self.vlan_state {
+            let socket = vlan
+                .get_socket_mut(socket_index)
                 .ok_or(RequestError::Fail(ClientError::BadMessageContents))?;
```

Change made in new tip commit.

task/net/src/server_basic.rs
cbiffle: Yeah, statics are not cloned once per monomorphization of their containing generic function, unfortunately.

task/net/src/server.rs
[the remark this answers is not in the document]
cbiffle: Think I got it, PTAL.

task/net/src/server.rs
cbiffle: Done

mkeeter (approved):

PR #884, opened 2022-10-27
Redesign build system kernel config structs.

cbiffle:
This insulates xtask dist from a bunch of kernel implementation decisions, removing (for example) the region table construction code from xtask and moving it into kern/build.rs. This should let us iterate on the kernel's internal representations more easily.

### DIFF
--- a/build/kconfig/src/lib.rs
+++ b/build/kconfig/src/lib.rs
@@ -1,0 +1,143 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use serde::{Deserialize, Serialize};
+use std::collections::{BTreeMap, BTreeSet};
+
+/// Application configuration passed into the kernel build.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct KernelConfig {
+    /// Tasks in the app image. The order of tasks is significant.
+    pub tasks: Vec<TaskConfig>,
+
+    /// Regions that tasks have shared access to, keyed by the name the task
+    /// config used to grant access (often peripheral name). These are typically
+    /// memory mapped peripherals.
+    pub shared_regions: BTreeMap<String, RegionConfig>,
+
+    /// Interrupts hooked by the application, keyed by IRQ number.
+    pub irqs: BTreeMap<u32, InterruptConfig>,
+}
+
+/// Configuration for a single hooked interrupt.
+#[derive(
+    Copy,
+    Clone,
+    Debug,
+    Serialize,
+    Deserialize,
+    Eq,
+    PartialEq,
+    Hash,
+    Ord,
+    PartialOrd,
+)]
+pub struct InterruptConfig {
+    /// Index of task (in the application task array) that receives this
+    /// interrupt.
+    pub task_index: usize,
+    /// Notification bits that are posted to the task when the interrupt fires.
+    /// Note that this is a mask and can have multiple (or zero!) bits set; the
+    /// kernel doesn't really care.
+    pub notification: u32,
+}
+
+/// Record describing a single task.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct TaskConfig {
+    /// Named memory regions that this task has exclusive access to, keyed by
+    /// name.
+    ///
+    /// The name is the "output" assignment that generated this region,
+    /// typically (but not necessarily!) either `"ram"` or `"flash"`.
+    pub owned_regions: BTreeMap<String, RegionConfig>,
+
+    /// Names of regions (in the app-level `shared_regions`) that this task
+    /// needs access to.
+    pub shared_regions: BTreeSet<String>,
+
+    /// Address of the task's entry point. This is the first instruction that
+    /// will be executed whenever the task is (re)started.
+    pub entry_point: OwnedAddress,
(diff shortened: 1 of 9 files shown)


build/kconfig/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// Application configuration passed into the kernel build.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct KernelConfig {
+    /// Tasks in the app image. The order of tasks is significant.
+    pub tasks: Vec<TaskConfig>,
+
+    /// Regions that tasks have shared access to, keyed by the name the task
+    /// config used to grant access (often peripheral name). These are typically
+    /// memory mapped peripherals (and should probably not be anything else).
```
mkeeter, line 16: Two notable exceptions to "should probably not be anything else" that may be worth calling out: `bank2` and `system_flash`

build/kconfig/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Address of the task's entry point. This is the first instruction that
+    /// will be executed whenever the task is (re)started.
+    pub entry_point: OwnedAddress,
+
+    /// Address of the task's initial stack pointer, to be loaded at (re)start.
+    /// It must be pointing into or *just past* one of the task's memory
+    /// regions.
+    pub initial_stack: OwnedAddress,
+
+    /// Initial priority of this task.
```
mkeeter, line 69: "Initial" suggests priority may change?

build/xtask/src/dist.rs
```diff
@@ -1712,198 +1711,156 @@ fn allocate_one(
...
     }
-
-    for (name, p) in toml.extratext.iter() {
-        if power_of_two_required && !p.size.is_power_of_two() {
-            panic!("Memory region for extratext '{}' is required to be a power of two, but has size {}", name, p.size);
+    for (name, p) in &toml.extratext {
+        if p2_required && !p.size.is_power_of_two() {
+            bail!(
+                "memory region for extratext '{}' is required to be \
+                   a power of two, but has size {}",
```
mkeeter, line 1757: Nit: weird indentation

build/xtask/src/dist.rs
```diff
@@ -1712,198 +1711,156 @@ fn allocate_one(
...
-
-        // skip periperhals that aren't in at least one task's `uses`
-        if !used_peripherals.contains(name) {
-            continue;
+    let mut flat_shared = BTreeMap::new();
+    for (name, p) in &toml.peripherals {
+        if p2_required && !p.size.is_power_of_two() {
+            bail!(
+                "memory region for peripheral '{}' is required to be \
+                   a power of two, but has size {}",
```
mkeeter, line 1733: Nit: weird indentation

build/xtask/src/dist.rs
```diff
@@ -1712,198 +1711,156 @@ fn allocate_one(
...
+        let shared_regions: std::collections::BTreeSet<String> =
+            task.uses.iter().cloned().collect();
+
+        let owned_regions = task_allocations[name].iter()
+            .map(|(out_name, range)| {
+                // Look up region for this image
+                let mut regions = toml.outputs[out_name].iter()
+                    .filter(|o| &o.name == image_name);
+                let out = regions.next().expect("no region for name");
+                if regions.next().is_some() {
```
mkeeter, line 1824: Nit: it's a little inefficient to do this check repeatedly in the inner-most loop, versus checking `toml.outputs` once elsewhere.

sys/kern/build.rs
```diff
@@ -80,14 +401,14 @@ fn generate_statics() -> Result<()> {
     /////////////////////////////////////////////////////////
     // Basic constants and empty space
 
-    let task_count = kconfig.tasks.len();
+    let task_count = gen.tasks.len();
```
mkeeter, line 404: This is the length of the `TokenStream`, which seems... odd.

mkeeter: oh wait, never mind

mkeeter (approved):

build/kconfig/src/lib.rs
cbiffle: TBH neither of those should be represented using shared peripheral regions, we sort of need a new mechanism -- but in terms of this file, you are correct.

build/kconfig/src/lib.rs
cbiffle: Yeah, doing priority ceiling protocol may require us to change the dynamic priority of tasks, if we decide to do that.

build/xtask/src/dist.rs
cbiffle: Huh, interesting, I did not write it that way -- rustfmt's wrapped string handling seems hard to predict.

build/xtask/src/dist.rs
cbiffle: At this point, I have _no idea_ what invariants are being maintained on the toml datastructure (i.e. which checks I can rely upon having done elsewhere), so while I agree that it would be more efficient to "parse don't validate" this, I feel like we're kind of in debt here and I'm not sure where to do it.

sys/kern/build.rs
cbiffle: It is the length of the `Vec<TokenStream>` where we generate one stanza per task.

cbiffle:
- fixed indentation of wrapped strings
- removed ideological statement on `shared_regions` that is currently not correct
- rebased

Issue #872, opened 2022-10-24
PSC inventory support


mkeeter:
The power shelf has two I2C addresses (`pmbus` and FRU ID), and GPIOs.

We should expose them through the `humility validate` and/or inventory APIs.

The PMBus device is already in the `pmbus` crate [here](https://github.com/oxidecomputer/pmbus/blob/master/src/mwocp68.ron), but is missing a Hubris driver.

For details, see this [ACAN-114 Application Note](https://www.murata.com/-/media/webrenewal/products/power/appnote/acan-114.ashx)

PR #885, opened 2022-10-27
Integrate Murata power shelf sensors

### LINKED ISSUE #873: Power shelf sensor support
Once we can take inventory of the power shelves (#872), we should monitor voltage / current / fan speed / temperature (?) and log it to the `sensors` task.

mkeeter:
This PR adds communication to the actual Murata power shelves!

It includes a few pieces:

- Modifications to the build system to support sensors that are decoupled from rails.  This is analogous to https://github.com/oxidecomputer/humility/pull/252 , and requires Humility 0.9.3 (enforced by a Hubris build archive bump)
- I2C drivers for the M24C02 EEPROM (used as a FRU ID in the power shelf) and the MWOCP68 itself
- Support for the power shelves in the `power` task, which means they work with `humility pmbus` (they were already in the `pmbus` crate, thanks @bcantrill!)
- A new `thermal-sensors` task, for boards which read temperature / fan sensors, but do not have a control loop.  <del>I'm open to bikeshedding about this name!</del> **Edit:** renamed to `sensor-polling`

Closes #873, #872.


build/i2c/src/lib.rs
```diff
@@ -1037,7 +1046,26 @@ impl ConfigGenerator {
...
+            // isn't always true: in the power shelf, for example, there are two
+            // rails and three (uncorrelated) temperature sensors.
+            //
+            // This is indicated with the `sensors` array, which allows us to
+            // specify only certain kinds of sensors being tied to rails.
+            //
+            // If the `sensors` array is `None`, then we fall back to the
+            // default case of all sensors being one-to-one associated with
+            // rails.
+            let name: Option<String> = if d.power.is_some()
```
jgallagher, line 1060: I took a bunch of stabs at making this easier to read and mostly failed. One possibility that might work: what if we extract this into a separate function? Something like

```rust
        // In most cases, ... rest of that comment ...
        fn power_for_kind(kind: Sensor, d: &I2cDevice) -> Option<&I2cPower> {
            d.power.as_ref().and_then(|power| {
                if let Some(sensors) = power.sensors.as_ref() {
                    if !sensors.contains(&kind) {
                        return None;
                    }
                }
                Some(power)
            })
        }
```

at which point I think this could be

```rust

let name: Option<String> = if let Some(power) = power_for_kind(kind, d) {
    if let Some(rails) = &power.rails {
        // .. rest unchanged ..
```

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -0,0 +1,127 @@
...
+    InvalidData { err: pmbus::Error },
+}
+
+impl From<Error> for ResponseCode {
+    fn from(err: Error) -> Self {
+        match err {
+            Error::BadRead { code, .. } => code,
+            Error::BadWrite { code, .. } => code,
+            Error::BadValidation { code, .. } => code,
+            _ => panic!(),
```
jgallagher, line 41: This seems a little questionable in isolation. `From` seems like an easy impl to call without looking at the details; is a caller responsible for knowing the error is in one of the valid cases? Should this be `TryFrom` and let the caller panic?

task/sensor-polling/src/main.rs
```diff
@@ -0,0 +1,186 @@
...
+use drv_i2c_devices::mwocp68::{Error as Mwocp68Error, Mwocp68};
+use ringbuf::*;
+use task_sensor_api::{Sensor, SensorError, SensorId};
+use userlib::*;
+
+task_slot!(I2C, i2c_driver);
+task_slot!(SENSOR, sensor);
+
+/// Type containing all of our temperature sensor types, so we can store them
+/// generically in an array.  Right now, we only support
```
jgallagher, line 19: Are we missing the end of this comment? (If not and this was intentional, maybe add a `:` to the end?)

jgallagher (approved):

build/i2c/src/lib.rs
mkeeter: Yup, brought this into a helper function under `I2cDevice` and cleaned it up _substantially_.

drv/i2c-devices/src/mwocp68.rs
mkeeter: Good point – this matches previous code, but the previous code could probably be improved as well.  I switched to returning `ResponseCode::BadDeviceState` as a catch-all.

task/sensor-polling/src/main.rs
mkeeter: Not intentional; fixed!

PR #888, opened 2022-10-28
Support NVME BMC devices with bad mux

mkeeter:
Due to https://github.com/oxidecomputer/hardware-gimlet/issues/1804, we can't talk to M.2s when they are unpowered.  We work around this in the `thermal` task by asking the `MAX5970` for its status; however, `humility validate` doesn't have any support for that kind of special-casing.

This PR adds a new `nvme_bmc_bad_mux`, which is ignored during `humility validate`.

In addition, it renames `nvmebmc` -> `nvme_bmc` so that the existing driver is used for `humility validate` (on good devices).

### DIFF
--- a/drv/i2c-devices/src/nvme_bmc_bad_mux.rs
+++ b/drv/i2c-devices/src/nvme_bmc_bad_mux.rs
@@ -1,0 +1,44 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::Validate;
+use drv_i2c_api::{I2cDevice, ResponseCode};
+use userlib::units::Celsius;
+
+pub use crate::nvme_bmc::{Error, NvmeBmc};
+
+/// Wrapper for an NVME BMC device on the far end of an I2C mux that exhibits
+/// lock-up behavior; see `hardware-gimlet#1804`
+pub struct NvmeBmcBadMux {
+    dev: NvmeBmc,
+}
+
+impl NvmeBmcBadMux {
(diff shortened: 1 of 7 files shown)


drv/i2c-devices/src/nvme_bmc_bad_mux.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+impl Validate<ResponseCode> for NvmeBmcBadMux {
+    fn validate(
+        _device: &drv_i2c_api::I2cDevice,
+    ) -> Result<bool, ResponseCode> {
+        // Due to a hardware limitation, we can only *attempt* to communicate
+        // with the M.2s when they are powered; otherwise, the entire I2C bus
+        // locks up, which is bad for business.
```
bcantrill, line 35: 🤣

bcantrill (approved):
This looks good, but the bit that isn't sitting quite right is the nomenclature.  Aside from overflowing the `humility validate` column width, I also think it's a bit of a misnomer: it's not a *bad* mux, it's just potentially an unpowered segment.  What about `nvmc_bmc_a0` or `nvme_bmc_m2`?

mkeeter:
Hmmm, I don't like `nvmc_bmc_a0` or `nvme_bmc_m2` because it's still a generic NVME BMC device, and it _should_ work even when the device is unpowered (i.e. return `NotFound` but not lock up entirely).

In the next Gimlet hardware revision – when this is fixed – they'll still be powered only in A0+HP (and will remain M2 drives), but we will switch them back to the regular `nvme_bmc` driver, so both of the suggested names seem misleading...

I'll think about the name a little more and ask for suggestions at Hubris Huddle today.

mkeeter:
We chatted in #oxide-hubris, and settled on `m2_hp_only`

PR #891, opened 2022-10-30
host-sp-comms interrupt line to SP3 should be open drain

### LINKED ISSUE #890: host-sp-comms set_status_impl() not taking power domain into account
The `host-sp-comms` tasks understandably exists across power domains. In particular, it can exist in A2 as it is part of the host normal path for executing commands like a reboot and related. However, this puts us in a particularly gnarly power domain case where we can possibly cross power domains. Normally the UART is protected by a buffer for the actual data and ready to send style pins that are treated as push-pull. However, the interrupt pin that we're using is not protected by that buffer. Unfortunately, all of the UART pins should only be driven from the ACPI state S0 according to the docs, which we treat as A0.

Because the task can operate across domains the current construct in `set_status_impl()` is liable to cause back driving. UART backriving has been a serious concern and we've proven issues on Ethanol-X. While we do have isolation between the main UART signals, we didn't for the SP->SP3 interrupt line because at the time use was less clear and we figured the SP could do the right thing semantically.

There are two approaches to this to consider:

1. `set_status_impl()` needs to be aware of the power domain and if we are transitioning away from A0 at all (more specifically the sequencer would toggle the uart enable lines) then we need to disable output on the GPIO.
2. Alternatively, because there is a pull up resistor on this, we may be able to get away with a bit of cheating and treating this like an open-drain pin on the SP side. Technically this is a push-pull from the SP3 side, but depending on if the interrupt is edge or level triggered and the rise times of this, that would allow us to potentially pretend the power domain doesn't exist as if the line is driven to 3.3V the pull up will cover it if it's in A0 and will not if we're not in A0 as the pull is in that domain.

citrus-it:
Here's a proposed PR to fix #890 by configuring the SP-to-host GPIO as open drain.

### DIFF
--- a/task/host-sp-comms/src/main.rs
+++ b/task/host-sp-comms/src/main.rs
@@ -857,9 +857,9 @@ fn sp_to_sp3_interrupt_enable(sys: &sys_api::Sys) {
 
     sys.gpio_configure_output(
         SP_TO_SP3_INT_L,
-        sys_api::OutputType::PushPull,
+        sys_api::OutputType::OpenDrain,
         sys_api::Speed::Low,
-        sys_api::Pull::Down,
+        sys_api::Pull::None,
     )
     .unwrap();
 }


task/host-sp-comms/src/main.rs
```diff
@@ -853,15 +853,15 @@ cfg_if::cfg_if! {
...
         SP_TO_SP3_INT_L,
-        sys_api::OutputType::PushPull,
+        sys_api::OutputType::OpenDrain,
         sys_api::Speed::Low,
-        sys_api::Pull::Down,
+        sys_api::Pull::None,
     )
     .unwrap();
+
+    sys.gpio_set(SP_TO_SP3_INT_L).unwrap();
```
jgallagher, line 864: Why do we move the `gpio_set` to after the configure call? (FWIW I originally had it this way, but @mkeeter corrected me https://github.com/oxidecomputer/hubris/pull/861#discussion_r999810734)

jgallagher (approved):

task/host-sp-comms/src/main.rs
citrus-it: We never want to actively drive this pin high from the SP end and I presumed that setting it before configuring as open drain would cause the SP to do that. I'll defer to @mkeeter here.

mkeeter: I think the `gpio_set` call should be called **before** `gpio_configure_output`.  @citrus-it the missing bit of info is that pins come out of reset in analog mode, rather than as outputs.

Here's what happens when `gpio_set` is called after `gpio_configure_output` (the bad order):

- When the SP comes out of reset, this pin is configured in analog mode, meaning the pin is High-Z (and therefore pulled up by the external resistor).
- Calling `gpio_configure_output` switches to output, open-drain mode.  Because the output data register `ODRx` is be 0 (its default value), the output driver will enable the N-MOS, pulling the pin down to ground.
- Then, the `gpio_set` call will set `ODRx` to 1, the output driver will go High-Z, and we'll be pulled back up by the external resistor.

The net effect will be a High → Low → High glitch.

If you call `gpio_set` beforehand (the good order):

- When the SP comes out of reset, this pin is configured in analog mode, meaning the pin is High-Z (and therefore pulled up by the external resistor).
- Calling `gpio_set` will set `ODRx` to 1, which is ignored in analog mode
- Calling `gpio_configure_output` switches us to open-drain mode; because `ODRx` is already high, we stay High-Z and there is no glitch.

One corner case: if the task restarts, then the pin is _already_ configured in open-drain mode, which is also fine.

citrus-it: Thanks for the detail, I've put the order back at it was, rebased and pushed.

PR #892, opened 2022-10-31
thermal: add PID term to apply offset to output values

kc8apf:
When output value is used as fan duty cycle directly, only positive error values are meaningful.  This prevents the PID loop from engaging until the set point is already reached.  By offsetting the output values, negative errors are meaningful and the PID loop can engage control during the initial rise toward the set point.


mkeeter (approved):
Thanks!

PR #895, opened 2022-11-01
Implement host-sp-comms debug register and new status reply format

citrus-it:


lib/host-sp-messages/src/lib.rs
```diff
@@ -336,11 +348,7 @@ mod tests {
             version: 123,
             sequence: 456,
         };
-        let host_to_sp = HostToSp::HostPanic {
-            status: 78,
-            cpu: 90,
-            thread: 12345,
-        };
+        let host_to_sp = HostToSp::HostPanic { code: 78 };
```
jgallagher, line 351: Thanks for fixing these tests; I'll see if I can get them added to CI so they don't bitrot moving forward.

task/host-sp-comms/src/main.rs
```diff
@@ -175,6 +179,12 @@ impl ServerImpl {
         }
     }
 
+    fn set_debug_impl(&mut self, debug: DebugReg) {
+        if debug != self.debug {
```
jgallagher, line 183: Tiny nitpick - we don't need this conditional here; this can be just `self.debug = debug`. `set_status_impl()` has a check on whether its value changed to avoid needlessly calling into the `sys` task to set the GPIO to what it already is.

jgallagher (approved):
Thanks!

citrus-it:
Closing in favour of https://github.com/oxidecomputer/hubris/pull/899

PR #896, opened 2022-11-01
build/util: make env var lookup failure message better

cbiffle:
Hey I found another case where we were leaking the crappy default env var lookup failure message.

### DIFF
--- a/build/util/src/lib.rs
+++ b/build/util/src/lib.rs
@@ -9,9 +9,9 @@ use std::collections::BTreeMap;
 /// Reads the given environment variable and marks that it's used
 ///
 /// This ensures a rebuild if the variable changes
-pub fn env_var(key: &str) -> Result<String, std::env::VarError> {
+pub fn env_var(key: &str) -> Result<String> {
     println!("cargo:rerun-if-env-changed={}", key);
-    std::env::var(key)
+    std::env::var(key).with_context(|| format!("reading env var ${key}"))
 }
 
 /// Reads the `OUT_DIR` environment variable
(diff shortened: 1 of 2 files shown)


build/util/src/lib.rs
```diff
@@ -9,9 +9,9 @@ use std::collections::BTreeMap;
 /// Reads the given environment variable and marks that it's used
 ///
 /// This ensures a rebuild if the variable changes
-pub fn env_var(key: &str) -> Result<String, std::env::VarError> {
+pub fn env_var(key: &str) -> Result<String> {
     println!("cargo:rerun-if-env-changed={}", key);
-    std::env::var(key)
+    std::env::var(key).with_context(|| format!("reading ${key}"))
```
faithanalog, line 14: i swear im not just trying to nitpick, but i think this should maybe be `"reading env var ${key}"` instead of `"reading ${key}"`. here's why:

```
vi@localhost ~/p/testshit (main)> cat src/main.rs
use anyhow::{Context};
fn main() {
    println!(
        "{:?}",
        std::env::var("BLIT")
            .context("Get owned n00b: BLIT")
            .err()
            .unwrap()
    );
    println!(
        "{}",
        std::env::var("BLORT")
            .context("Get owned n00b: BLORT")
            .err()
            .unwrap()
    );
}
vi@localhost ~/p/testshit (main)> ./target/release/testshit
Get owned n00b: BLIT

Caused by:
    environment variable not found
Get owned n00b: BLORT
```

see like in the second one it didn't print the "environment variable not found" bit. we should probablyyy be printing all the errors with `{:?}` but just in case some error from this code (or some copy pasted version of this code) ends up in a `{}` itd be nice to have it in the context as backup.

just my 2 bytes!

faithanalog (changes requested):

faithanalog:
hell yes give me error messages i crave error messages

build/util/src/lib.rs
cbiffle: Yeah, I kinda wish `anyhow` would always print the good error message, making this less likely. Ah well.

Fixed!

faithanalog (approved):

PR #897, opened 2022-11-01
Make TaskDesc/RegionDesc kernel implementation details

cbiffle:
Since the Early Days, we've had these `TaskDesc`/`RegionDesc` structs in the `abi` crate, giving the static definition of tasks in a Hubris application. Originally this was because they got deposited, by a separate tool, into an array at a predictable location in Flash; the kernel would then interpret them at runtime to set things up. This made it almost impossible to notice when you were running out of kernel RAM, since it depended on the results of that interpretation, rather than (say) the linker.

So about a year ago I fixed that.

The fact that the structs were still public has been gnawing at me, and recently I ran into some cases where I wanted to make changes to the structs that simply wouldn't work if they were in abi. This is because the remaining out-of-kernel use of the structs was by the build system, and it needed `Serialize`/`Deserialize` on the structs, which limits which types they can contain. It also introduced weird coupling between the build system (specifically the ever-growing `build/xtask/src/dist.rs`) and the _internal task structure used by the kernel,_ which seemed grody. Anyway, that part was fixed recently with the introduction of the `build/kconfig` crate.

These commits finish the job, by removing the `Desc` structs from `abi` _completely,_ recognizing them as the kernel-internal implementation detail that they are. I then take advantage of this fact to optimize their layout a bit, saving 40 bytes of RAM per task and excising a considerable chunk of unsafe code from the startup routine.

### DIFF
--- a/sys/kern/src/arch/arm_m.rs
+++ b/sys/kern/src/arch/arm_m.rs
@@ -76,6 +76,7 @@ use core::sync::atomic::{AtomicBool, AtomicPtr, AtomicU32, Ordering};
 use zerocopy::FromBytes;
 
 use crate::atomic::AtomicExt;
+use crate::descs::RegionAttributes;
 use crate::startup::with_task_table;
 use crate::task;
 use crate::time::Timestamp;
@@ -344,12 +345,12 @@ pub fn apply_memory_protection(task: &task::Task) {
             | (1 << 4)  // honor the region number
             | region.base;
         let ratts = region.attributes;
-        let xn = !ratts.contains(abi::RegionAttributes::EXECUTE);
+        let xn = !ratts.contains(RegionAttributes::EXECUTE);
         // These AP encodings are chosen such that we never deny *privileged*
         // code (i.e. us) access to the memory.
-        let ap = if ratts.contains(abi::RegionAttributes::WRITE) {
+        let ap = if ratts.contains(RegionAttributes::WRITE) {
             0b011
-        } else if ratts.contains(abi::RegionAttributes::READ) {
+        } else if ratts.contains(RegionAttributes::READ) {
             0b010
         } else {
             0b001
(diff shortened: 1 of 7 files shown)


sys/kern/src/arch/arm_m.rs
```diff
@@ -278,25 +279,21 @@ pub fn reinitialize(task: &mut task::Task) {
...
-
-        if initial_stack > region.base + region.size {
-            continue;
-        }
-
+    if let Some(region) = task
+        .region_table()
+        .iter()
+        .find(|region| region.contains(initial_stack))
+    {
```
mkeeter, line 293: There's the possibility of a behavior change here: the previous code would zap **every** region that contains the stack pointer, while the new code stops after finding one region.  It seems unlikely to matter – are there any cases where this is significant?

sys/kern/src/descs.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+/// Record describing a single task.
+#[derive(Clone, Debug)]
+pub struct TaskDesc {
+    /// Identifies memory regions this task has access to, by index in the
+    /// `RegionDesc` table. If the task needs fewer than `REGIONS_PER_TASK`
+    /// regions, it should use remaining entries to name a region that confers
+    /// no access; by convention, this region is usually entry 0 in the table.
```
mkeeter, line 41: This comment seems out of date, because the data is now here!

sys/kern/src/descs.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub size: u32,
+    /// Flags describing what can be done with this region.
+    pub attributes: RegionAttributes,
+}
+
+impl RegionDesc {
+    /// Tests whether `self` contains `addr`.
+    pub fn contains(&self, addr: usize) -> bool {
+        let next_addr = addr.wrapping_add(1);
+        if next_addr < addr {
```
mkeeter, line 107: Is this equivalent to `addr == usize::MAX`?

sys/kern/src/arch/arm_m.rs
cbiffle: No, multiple regions containing the stack pointer would have been rejected by the region overlap check in the build system. I changed this code for sort of a peripheral reason: the loop held a borrow on the region table, but my switch to using `.region_table()` caused that to be a borrow on the entire task. Inserting a `break;` into the loop would also have fixed the borrowing issue, but at that point this seemed like a clearer statement of intent.

sys/kern/src/descs.rs
cbiffle: It is _in this commit_ but I have an upcoming commit switching this to `*const T`, where the meaning of the addition will now depend on sizeof T. So that's why it's phrased this way.

sys/kern/src/descs.rs
cbiffle: Hmmm, will rewrite, thanks

cbiffle: Fixed.

cbiffle: We talked about this in chat, but, that is in fact the fixed comment -- references to indices removed.

cbiffle:
Turns out this also knocks 160 bytes off kernel text, which is nice.

PR #899, opened 2022-11-02
Implement host-sp-comms debug register and new status reply format

jgallagher:
This is re-opening #895 with a source branch in `hubris` for github-PR reasons :-|


jgallagher:
This is now a rollup of several related PRs all working toward phase 2 recovery via the management network: #901 #902 #921 #922 #927

PR #901, opened 2022-11-02
host-sp-comms: fetch (NYI) host phase 2 blocks from control-plane-agent

jgallagher:
Major changes in this PR include:

* `control-plane-agent` now has an idol interface and associated API crate. It currently has two functions ("fetch phase 2 data from MGS, notify me when it arrives" and "get the phase 2 data you just notified me about"). These functions are stubbed out with dummy implementations for now; actual communication to MGS will come in a later PR.
* `host-sp-comms`'s handler for the host requesting phase 2 data now uses the above-mentioned `control-plane-agent` idol interface instead of immediately returning dummy data. This requires nontrivial changes to the uart state machine / interrupt handler, because while previously we were in one of the three states:

1. Receiving an incoming request
2. Transmitting a response
3. Periodically transmitting 0x00 bytes while waiting for a new request

we now have an additional state:

4. Not yet transmitting a response, but also not periodically transmitting 0x00 bytes, because we're waiting for another task (`control-plane-agent`) to give us data we need to formulate our response

I moved some of this state machinery into the new `tx_buf` submodule. That submodule also has a bit of dead code in it that will be applicable in the next PR (which I'll open momentarily).

This PR builds on #899 and should be merged after it.


task/control-plane-agent-api/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "task-control-plane-agent-api"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+derive-idol-err = {path = "../../lib/derive-idol-err" }
+num-traits = { version = "0.2.12", default-features = false }
+ssmarshal = {version = "1", default-features = false}
```
mkeeter, line 9: Is this import used?

task/host-sp-comms/src/main.rs
```diff
@@ -391,40 +400,37 @@ impl ServerImpl {
...
-                        // Encode our outgoing packet.
-                        let pkt_len = corncobs::encode_buf(
-                            &self.tx_msg_buf[..msg_len],
-                            &mut self.tx_pkt_buf[..],
-                        );
-                        self.tx_pkt_to_write = 0..pkt_len;
+                // Process message and set up `self.tx_buf` with our response
+                // (or intermediate state if we don't have a response yet).
+                self.process_message();
+                self.rx_buf.clear();
```
mkeeter, line 428: Do we ever call `self.process_message()` without then clearing `rx_buf`?  If not, then we may want to put `self.rx_buf.clear()` into `process_message`.

mkeeter (approved):

task/control-plane-agent-api/Cargo.toml
jgallagher: Nope, good catch. Removed

task/host-sp-comms/src/main.rs
jgallagher: This took a bit of refactoring to keep the borrow checker happy, but I think still ended up net positive? 6ee65fb4 if you disagree, let me know!

mkeeter: Yeah, looks better to me!

PR #902, opened 2022-11-02
host-sp-comms: Avoid deadlocking if host sends back-to-back requests

jgallagher:
Previously, the uart interrupt handler would only attempt to read from the RX FIFO after it had fully drained any data it needed to send to the TX FIFO. This works assuming a well-behaved host that only ever sends one request and always waits for one response. However, a host that sends two requests back to back for any reason (bug, malicious, spontaneous reboot, ...) could, if the messages were longer than our small uart FIFOs, land us in a hardware-flow-control-based deadlock, where both sides were attempting to transmit and neither side was attempting to receive.

This PR changes our handler to check for incoming data any time we fill the TX FIFO (and would have previously returned); if we receive an out-of-sync request, we'll abandon any remaining data left in our TX FIFO and start sending a response to the new request. We rely on the protocol and the host's error handling to handle the partially-written response in this case (which it can do based on cobs, crcs, and sequence numbers).


task/host-sp-comms/src/main.rs
```diff
@@ -492,51 +534,24 @@ impl ServerImpl {
...
+    // should only be set to true if we're being called in an "out of sync"
+    // path; see the comments in `handle_usart_notification()` where we check
+    // for an incoming request while we're still trying to send a previous
+    // response.
+    fn process_message(
+        &mut self,
+        reset_tx_buf: bool,
+    ) -> Result<(), DecodeFailureReason> {
+        let (header, request) = decode_packet(&mut self.rx_buf)
+            .map(|(header, request, _data)| (header, request))?;
```
mkeeter, line 549: Nit:
```suggestion
        let (header, request, _data) = decode_packet(&mut self.rx_buf)?;
```

mkeeter (approved):

Issue #640, opened 2022-07-05
Test Sidecar front IO board LEDs


mkeeter:
At bare minimum, hit them through Humility and see if they're present.

Ideally, write a device driver for them.

PR #904, opened 2022-11-02
Adds PCA9956B as an I2C device, adds LED control to transceivers

Aaron-Hartwig:
Closes https://github.com/oxidecomputer/hubris/issues/640


app/sidecar/rev-a.toml
```diff
@@ -740,6 +740,22 @@ device = "pca9538"
 description = "Front IO GPIO expander"
 removable = true
 
+[[config.i2c.devices]]
+bus = "front_io"
+address = 0b0001_010
+device = "pca9956b_left"
```
mkeeter, line 746: `device` should be the device name (without qualifiers), i.e. `pca9956b`.  Add a `name` field to distinguish them (here and below), e.g.
```toml
name = "front_leds_left"
```

After this, you'll need to edit `drv-transceivers-server` to use the new names, i.e.
```diff
diff --git a/drv/transceivers-server/src/main.rs b/drv/transceivers-server/src/main.rs
index a9b1a5fa..52f7f67d 100644
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -235,8 +235,8 @@ fn main() -> ! {

         let transceivers = Transceivers::new(FRONT_IO.get_task_id());
         let leds = Leds::new(
-            &i2c_config::devices::pca9956b_left_front_io(I2C.get_task_id())[0],
-            &i2c_config::devices::pca9956b_right_front_io(I2C.get_task_id())[0],
+            &i2c_config::devices::pca9956b_front_leds_left(I2C.get_task_id()),
+            &i2c_config::devices::pca9956b_front_leds_right(I2C.get_task_id()),
         );

         let mut server = ServerImpl {
```

app/sidecar/rev-b.toml
```diff
@@ -718,6 +718,22 @@ device = "pca9538"
 description = "Front IO GPIO expander"
 removable = true
 
+[[config.i2c.devices]]
+bus = "front_io"
+address = 0b0001_010
+device = "pca9956b_left"
```
mkeeter, line 724: Same here

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+    EFLAG0 = 0x41,
+    EFLAG1 = 0x42,
+    EFLAG2 = 0x43,
+    EFLAG3 = 0x44,
+    EFLAG4 = 0x45,
+    EFLAG5 = 0x46,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum LedErr {
```
mkeeter, line 91: Add a comment explaining where these values come from (Table 21 of the datasheet?)

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+            _ => LedErr::Invalid,
+        }
+    }
+}
+
+/// LedErrSummary is used to summarize the types of errors seen
+/// overtemp - true if MODE2_OVERTEMP was 1
+/// short_circuit - number of LEDs with LedErr::ShortCircuit
+/// open_circuit - number of LEDs with LedErr::OpenCircuit
+/// invald - number of LEDs with LedErr::Invalid
```
mkeeter, line 119:
```suggestion
/// invalid - number of LEDs with LedErr::Invalid
```

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+pub struct LedErrSummary {
+    pub overtemp: bool,
+    pub short_circuit: u8,
+    pub open_circuit: u8,
+    pub invalid: u8,
+}
+
+#[derive(Copy, Clone, Debug, Default)]
+struct Pca9956BErrorState {
```
mkeeter, line 129: I'm not sure why we need both `LedErrSummary` and `Pca9956BErrorState` here, versus always using `Pca9956BErrorState` (or something equivalent, e.g. using 3x `u32` bitmasks for Invalid / Open / Short, like I mentioned above)

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+    InvalidLED(u8),
+}
+
+impl From<ResponseCode> for Error {
+    fn from(err: ResponseCode) -> Self {
+        Error::I2cError(err)
+    }
+}
+
+impl From<Error> for ResponseCode {
```
mkeeter, line 189: I know that this pattern appears in a bunch of places, but we're trying to move away from it.

I'd recommend deleting this `impl From` and instead doing the `Error -> ResponseCode` conversion in `impl Validate`, e.g.
```diff
diff --git a/drv/i2c-devices/src/pca9956b.rs b/drv/i2c-devices/src/pca9956b.rs
index f9bda765..1c2bf743 100644
--- a/drv/i2c-devices/src/pca9956b.rs
+++ b/drv/i2c-devices/src/pca9956b.rs
@@ -186,15 +186,6 @@ impl From<ResponseCode> for Error {
     }
 }

-impl From<Error> for ResponseCode {
-    fn from(err: Error) -> Self {
-        match err {
-            Error::I2cError(code) => code,
-            _ => panic!(),
-        }
-    }
-}
-
 impl Pca9956B {
     pub fn new(device: &I2cDevice) -> Self {
         Self { device: *device }
@@ -323,9 +314,15 @@ impl Pca9956B {
 // which is the type of information we typically like to validate against.
 // MODE2[2:0] are set to read only an initialized to b101, so use that to
 // validate.
-impl Validate<Error> for Pca9956B {
-    fn validate(device: &I2cDevice) -> Result<bool, Error> {
-        let mode = Pca9956B::new(device).read_reg(Register::MODE2)?;
+impl Validate<ResponseCode> for Pca9956B {
+    fn validate(device: &I2cDevice) -> Result<bool, ResponseCode> {
+        let mode = Pca9956B::new(device).read_reg(Register::MODE2).map_err(
+            |e| match e {
+                // read_reg can only return Error::I2cError
+                Error::I2cError(e) => e,
+                _ => panic!(),
+            },
+        )?;

         Ok(mode & MODE2_RSVD_MASK == MODE2_RSVD)
     }
```

That way, the only possible `panic!` is in a well-known place; future users can't accidentally shoot themselves in the foot.

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    /// Write a number of Registers into `buf`
+    /// Note that CTRL_AUTO_INCR_MASK is set here to enable the device's
+    /// auto-increment feature, which has varying behaviors and limitations that
+    /// must be accounted for.
+    fn write_buffer(&self, reg: Register, buf: &[u8]) -> Result<(), Error> {
+        let mut data: [u8; 25] = [0; 25];
+        data[0] = (reg as u8) | CTRL_AUTO_INCR_MASK;
+        data[1..=buf.len()].copy_from_slice(buf);
```
mkeeter, line 238: This will panic if the incoming `buf` is too long.  Where does the `25` limitation come from?

It looks like the longest possible write is `0h-40h`, i.e. 65 bytes.

Should we use that size for `data`, and preemptively return an error if `reg as u8 + buf.len() >= EFLAG0`?

(along with a comment explaining that `EFLAG0` is the first read-only register)

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+        let reg = FromPrimitive::from_u8((Register::PWM0 as u8) + led).unwrap();
+        self.write_reg(reg, val)
+    }
+
+    /// Sets the PWMx register a number of LEDs, up to NUM_LEDS, beginning with
+    /// PWM0.
+    pub fn set_all_led_pwm(&self, vals: &[u8]) -> Result<(), Error> {
+        if vals.len() > NUM_LEDS {
+            return Err(Error::InvalidLED(
```
mkeeter, line 270: What about returning `Error::InvalidLED(NUM_LEDS)` (i.e. reporting the _first_ bad LED, instead of trying to report the _last_ bad LED and maybe falling back to 0xFF)?

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+            ));
+        }
+        self.write_buffer(Register::PWM0, &vals)
+    }
+
+    /// Queries the MODE2 register and to check the OVERTEMP and ERROR bits
+    /// If ERROR is set, each EFLAGx register will be read and parsed.
+    /// An important thing to note about this device is that in order to do
+    /// error detection (reflected in EFLAGx) the PWMx associated with the LED
+    /// must be higher than 0x08. So if the PWMx is set to zero, not error can
```
mkeeter, line 281:
```suggestion
    /// must be higher than 0x08. So if the PWMx is set to zero, no error can
```

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+        let mut err_state = Pca9956BErrorState {
+            overtemp: overtemp,
+            ..Default::default()
+        };
+
+        if error {
+            let mut eflags: [u8; 6] = [0; 6];
+            for i in 0..eflags.len() {
+                // Notably, the auto-increment function does not apply to these
+                // registers, so they must be fetched individually
```
mkeeter, line 297: 😱

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,320 @@
...
+    /// Set the current to whatever DEFAULT_LED_CURRENT is
+    pub fn initialize_current(&self) -> Result<(), Error> {
+        self.set_current(DEFAULT_LED_CURRENT)?;
+
+        Ok(())
+    }
+
+    /// Set the current to `value`
+    pub fn set_current(&self, value: u8) -> Result<(), Error> {
+        for (_i, controller) in self.controllers.iter().enumerate() {
```
mkeeter, line 249: Not sure why we're `.enumerate()`'ing here if we're ignoring the index

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,320 @@
...
+            controller.set_iref_all(value)?;
+        }
+
+        Ok(())
+    }
+
+    /// Turns on the System LED to a PWM value of DEFAULT_LED_PWM
+    pub fn turn_on_system_led(&self) -> Result<(), Error> {
+        self.controllers[LED_MAP[SYSTEM_LED_IDX].controller as usize]
```
mkeeter, line 258: Instead of looking it up repeatedly, I'd do
```rust
        const SYSTEM_LED: LedLocation = LED_MAP[SYSTEM_LED_IDX];
        self.controllers[SYSTEM_LED.controller as usize]
            .set_a_led_pwm(SYSTEM_LED.output, DEFAULT_LED_PWM)?;
```
(requires `#[derive(Clone)]` on `LedLocation`)

This isn't any more or less efficient, but is slightly more readable IMHO

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,320 @@
...
+            .set_a_led_pwm(LED_MAP[SYSTEM_LED_IDX].output, DEFAULT_LED_PWM)?;
+
+        Ok(())
+    }
+
+    /// Takes a `mask` of which ports need their LEDs turned on
```
mkeeter, line 264: Add one more sentence here explaining that this sets `DEFAULT_LED_PWM` for any port with a bit set in the mask.

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,320 @@
...
+            } else {
+                0
+            };
+
+            if LED_MAP[i].controller == LedController::Left {
+                data_l[LED_MAP[i].output as usize] = pwm_value;
+            } else {
+                data_r[LED_MAP[i].output as usize] = pwm_value;
+            }
+        }
```
mkeeter, line 282: Since we initialized the arrays to 0, this can be

```suggestion
        for i in 0..32 {
            let bit_mask: u32 = 1 << i;
            if (mask & bit_mask) != 0 {
                let index = LED_MAP[i].output as usize;
                match LED_MAP[i].controller {
                    LedController::Left => data_l[index] = DEFAULT_LED_PWM,
                    LedController::Right => data_r[index] = DEFAULT_LED_PWM,
                }
            }
        }
```

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,320 @@
...
+        let summary = FullErrorSummary {
+            left: match self.controllers[LedController::Left as usize]
+                .check_errors()
+            {
+                Ok(errs) => errs,
+                Err(_) => LedErrSummary {
+                    ..Default::default()
+                },
+            },
+            right: match self.controllers[LedController::Right as usize]
```
mkeeter, line 308: After seeing this `as usize` cast a bunch, I think we should add a helper function

```rust
fn controller(&self, c: LedController) -> &Pca9956B {
    &self.controllers[c as usize]
}
```
and use it everywhere

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,320 @@
...
+    /// Query device registers and return a summary of observed errors
+    pub fn error_summary(&self) -> Result<FullErrorSummary, Error> {
+        // Errors are being temporarily suppressed here due to a miswiring of
+        // the I2C bus at the LED controller parts. They will not be accessible
+        // without rework to older hardware, and newer (correct) hardware will
+        // be replacing the hold stuff very soon.
+        // TODO: remove error suppression here once Rev B hardware is in
+        let summary = FullErrorSummary {
+            left: match self.controllers[LedController::Left as usize]
+                .check_errors()
```
mkeeter, line 301: This could be `.check_errors().unwrap_or_default()`

drv/transceivers-server/src/main.rs
```diff
@@ -160,16 +187,87 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
     }
 }
 
+impl NotificationHandler for ServerImpl {
+    fn current_notification_mask(&self) -> u32 {
+        TIMER_NOTIFICATION_MASK
+    }
+
+    fn handle_notification(&mut self, _bits: u32) {
```
mkeeter, line 195: Add a comment explaining we only have one source of notification, so we don't need to check bits here.

drv/transceivers-server/src/main.rs
```diff
@@ -160,16 +187,87 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
...
+            Err(_) => 0,
+        };
+
+        if presence != self.modules_present {
+            // Errors are being temporarily suppressed here due to a miswiring of
+            // the I2C bus at the LED controller parts. They will not be accessible
+            // without rework to older hardware, and newer (correct) hardware will
+            // be replacing the hold stuff very soon.
+            // TODO: remove error suppression here once Rev B hardware is in
+            let _ = self.leds.update_led_state(presence);
```
mkeeter, line 215: Same as before, conditionally build code with / without error handling depending on `#[cfg(board_name = "sidecar-[a or B]")]`

drv/transceivers-server/src/main.rs
```diff
@@ -160,16 +187,87 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
...
+    }
+}
+
 #[export_name = "main"]
 fn main() -> ! {
     loop {
-        let mut buffer = [0; idl::INCOMING_SIZE];
+        // This is a temporary workaround that makes sure the FPGAs are up
+        // before we start doing things with them. A more sophisticated
+        // notification system will be put in place.
```
mkeeter, line 230: yup, don't like this 😆

drv/transceivers-server/src/main.rs
```diff
@@ -160,16 +187,87 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
...
+            &i2c_config::devices::pca9956b_right_front_io(I2C.get_task_id())[0],
+        );
+
+        let mut server = ServerImpl {
+            transceivers,
+            leds,
+            modules_present: 0,
+            led_error: FullErrorSummary {
+                ..Default::default()
+            },
```
mkeeter, line 248:
```suggestion
            led_error: Default::default(),
```

drv/transceivers-server/src/main.rs
```diff
@@ -160,16 +187,87 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
...
+
+        ringbuf_entry!(Trace::LEDInit);
 
-        let mut server = ServerImpl { transceivers };
+        server.transceivers.enable_led_controllers().unwrap();
+        // Errors are being temporarily suppressed here due to a miswiring of
+        // the I2C bus at the LED controller parts. They will not be accessible
+        // without rework to older hardware, and newer (correct) hardware will
+        // be replacing the hold stuff very soon.
+        // TODO: remove error suppression here once Rev B hardware is in
```
mkeeter, line 258: Add conditional compilation here as well

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,332 @@
...
+
+/// LedErrSummary is used to summarize the types of errors seen
+/// overtemp - true if MODE2_OVERTEMP was 1
+/// short_circuit - number of LEDs with LedErr::ShortCircuit
+/// open_circuit - number of LEDs with LedErr::OpenCircuit
+/// invald - number of LEDs with LedErr::Invalid
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+pub struct LedErrSummary {
+    pub overtemp: bool,
+    pub short_circuit: u8,
```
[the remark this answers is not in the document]
Aaron-Hartwig: Yeah, good point. I wasn't really happy with where I ended up on this but I like your suggestion.

drv/i2c-devices/src/pca9956b.rs
Aaron-Hartwig: Yeah, I was not thrilled with where I landed on this initial implementation. I think turning them into bitmasks is the right move and follows the existing pattern of passing around `u32`s as masks for the transceivers.

drv/i2c-devices/src/pca9956b.rs
Aaron-Hartwig: Yeah...Table 6 in the datasheet outlines the different configurations for auto-incrementing, none of which operate above `0x3E`. It makes a bit of sense, since after that is just `PWMALL`, `IREFALL`, and `EFLAGn` anyway.

drv/i2c-devices/src/pca9956b.rs
Aaron-Hartwig: Thanks for pointing this out! The 25 was somewhat arbitrary assuming a stride of all PWMx and an address byte. I'll size `data` at `63` (`0h-3Eh`, plus an address byte) and add in the requisite check and comment.

drv/sidecar-front-io/src/leds.rs
Aaron-Hartwig: 🧠

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    /// Query device registers and return a summary of observed errors
+    pub fn error_summary(&self) -> Result<FullErrorSummary, Error> {
+        // Errors are being temporarily suppressed here due to a miswiring of
+        // the I2C bus at the LED controller parts. They will not be accessible
+        // without rework to older hardware, and newer (correct) hardware will
+        // be replacing the hold stuff very soon.
+        // TODO: remove error suppression here once Rev B hardware is in
+        let summary = FullErrorSummary {
```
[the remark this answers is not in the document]
Aaron-Hartwig: 👍🏻 That is the certainly the better path here. I will rework to do that.

drv/transceivers-server/src/main.rs
mkeeter: More helpfully, could you use the `Sequencer.front_io_phy_ready` API (like [this](https://github.com/oxidecomputer/hubris/blob/master/task/monorail-server/src/bsp/sidecar_b.rs#L138-L146)) instead of manually recreating it here?

Aaron-Hartwig: Yeah, me either! We should chat about this, with @arjenroodselaar if possible.

Aaron-Hartwig: Dang, I had this window open so I never saw your second comment before I responded. I'll push another commit which addresses this.

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,311 @@
...
+            .map_err(Error::I2cError)
+    }
+
+    /// Read a number of Registers into `buf`
+    /// Note that CTRL_AUTO_INCR_MASK is set here to enable the device's
+    /// auto-increment feature, which has varying behaviors and limitations that
+    /// must be accounted for.
+    #[allow(dead_code)]
+    fn read_buffer(&self, reg: Register, buf: &mut [u8]) -> Result<(), Error> {
+        self.device
```
mkeeter, line 188: Now that we've got errors for bad autoincrement choices, should we check that stuff here as well?

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,311 @@
...
+    pub fn set_pwm_all(&self, val: u8) -> Result<(), Error> {
+        self.write_reg(Register::PWMALL, val)
+    }
+
+    /// Sets the PWMx register to `val` for a given `led`
+    pub fn set_a_led_pwm(&self, led: u8, val: u8) -> Result<(), Error> {
+        if led >= NUM_LEDS as u8 {
+            return Err(Error::InvalidLED(led));
+        }
+        let reg = FromPrimitive::from_u8((Register::PWM0 as u8) + led).unwrap();
```
mkeeter, line 238: You can use `from_u8` on the `Register` type here, which makes it more obvious:
```suggestion
        let reg = Register::from_u8((Register::PWM0 as u8) + led).unwrap();
```

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,311 @@
...
+            ..Default::default()
+        };
+
+        if error {
+            let mut eflags: [u8; 6] = [0; 6];
+            for (i, eflagx) in eflags.iter_mut().enumerate() {
+                // Notably, the auto-increment function does not apply to these
+                // registers, so they must be fetched individually
+                let reg =
+                    FromPrimitive::from_u8(Register::EFLAG0 as u8 + i as u8)
```
mkeeter, line 273:
```suggestion
                let reg = Register::from_u8(Register::EFLAG0 as u8 + i as u8)
```

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,356 @@
...
+    output: u8,
+}
+
+/// FullErrorSummary takes the errors reported by each individual PCA9956B and
+/// maps them to a by-transceiver-port representation for the masks
+#[derive(Copy, Clone, Default, PartialEq, Eq)]
+pub struct FullErrorSummary {
+    pub overtemp_left: bool,
+    pub overtemp_right: bool,
+    pub system_led_err: bool,
```
mkeeter, line 48: hmmm, do we want this to be an `LedErr` instead, to avoid losing information?

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    fn controller(&self, c: LedController) -> &Pca9956B {
+        &self.controllers[c as usize]
+    }
+
+    /// Set the current to whatever DEFAULT_LED_CURRENT is
+    pub fn initialize_current(&self) -> Result<(), Error> {
+        self.set_current(DEFAULT_LED_CURRENT)?;
+
+        Ok(())
```
mkeeter, line 255:
```suggestion
        self.set_current(DEFAULT_LED_CURRENT)
```

drv/sidecar-front-io/src/leds.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    /// Turns on the System LED to a PWM value of DEFAULT_LED_PWM
+    pub fn turn_on_system_led(&self) -> Result<(), Error> {
+        const SYSTEM_LED: LedLocation = LED_MAP[SYSTEM_LED_IDX];
+        self.controllers[SYSTEM_LED.controller as usize]
+            .set_a_led_pwm(SYSTEM_LED.output, DEFAULT_LED_PWM)?;
+
+        Ok(())
```
mkeeter, line 273:
```suggestion
        self.controller(SYSTEM_LED.controller)
            .set_a_led_pwm(SYSTEM_LED.output, DEFAULT_LED_PWM)
```

drv/transceivers-server/src/main.rs
```diff
@@ -160,16 +190,117 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
...
+            // Errors are being suppressed here due to a miswiring of the I2C bus at
+            // the LED controller parts. They will not be accessible without rework
+            // to older hardware, and newer (correct) hardware will be replacing the
+            // hold stuff very soon.
+            // TODO: remove conditional compilation path once sidecar-a is sunset
+            cfg_if::cfg_if! {
+                if #[cfg(target_board = "sidecar-a")] {
+                    let _ = self.leds.update_led_state(presence);
+                } else {
+                    self.leds.update_led_state(presence).unwrap();
```
mkeeter, line 223: I'm not sure if panicking the task is the right call when communication fails.  Maybe logging an error in the ringbuf instead, unless there's a good reason to restart the whole task?

drv/transceivers-server/src/main.rs
```diff
@@ -160,16 +190,117 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
...
+        // to older hardware, and newer (correct) hardware will be replacing the
+        // hold stuff very soon.
+        // TODO: remove conditional compilation path once sidecar-a is sunset
+        cfg_if::cfg_if! {
+            if #[cfg(target_board = "sidecar-a")] {
+                let _ = server.leds.initialize_current();
+                let _ = server.leds.turn_on_system_led();
+            } else {
+                server.leds.initialize_current().unwrap();
+                server.leds.turn_on_system_led().unwrap();
```
mkeeter, line 291: (For context, I'm fine panicking here, because if startup fails, there's not much else we can do...)

drv/i2c-devices/src/pca9956b.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    /// Write a number of Registers into `buf`
+    /// Note that CTRL_AUTO_INCR_MASK is set here to enable the device's
+    /// auto-increment feature, which has varying behaviors and limitations that
+    /// must be accounted for.
+    fn write_buffer(&self, reg: Register, buf: &[u8]) -> Result<(), Error> {
+        if buf.len() > MAX_BUF_SIZE {
+            return Err(Error::WriteBufferTooLarge(buf.len()));
+        } else if reg > MAX_AUTO_INC_REG {
```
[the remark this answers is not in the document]
Aaron-Hartwig: I was a bit conflicted on how opinionated to be with these error conditions because while there is a small subset of registers where auto-increment doesn't work, if you begin reading/writing _outside_ that range, the PCA9956B rolls its internal counter over. So, for instance, if you begin writing to `0x3E`, the next write will occur at `0x0` (in the default auto-increment configuration). However if you try to begin an auto-incrementing transaction at `0x3F` the behavior is undefined. So my decision here was to prevent someone from directly doing something undefined, while leaving open the opportunity to leverage the internal address rollover if they chose to.

drv/sidecar-front-io/src/leds.rs
Aaron-Hartwig: I agree. I actually had it that way at first but convinced myself last night that I didn't need that. Today Aaron disagrees with that choice.

drv/transceivers-server/src/main.rs
Aaron-Hartwig: Good point. I agree that there is really no point in crashing the whole task due to problems with the LEDs. It will be good if we can recognize that an error has occurred in order to inform an operator (if necessary) that LEDs may not be functional, but the transceivers themselves may still be operational.

drv/transceivers-server/src/main.rs
Aaron-Hartwig: Thinking on this more, I don't think we should panic here. I would just as soon hold a piece of state indicating that LEDs were not initialized (and therefore we should not try to update them) and then proceed as normal. We could then build out a recovery strategy if we so choose, but in reality if we fail to initialize the LEDs but can still talk to the FPGAs, we likely have a problem limited in scope to the PCA9956Bs. Following my logic above, I don't think LED failure should not impede transceiver activities.

mkeeter (approved):
Looks good, once you rebase + fix the build check

PR #905, opened 2022-11-02
Introduce new command, xtask print

steveklabnik:
This command can print out information about the build. Currently, it only supports one option, --archive, which prints out the final build archive. This is useful in scripting.

Example:

```
〉cargo xtask print .\app\demo-stm32f4-discovery\app.toml --archive
    Finished dev [optimized + debuginfo] target(s) in 1.68s
     Running `target\debug\xtask.exe print .\app\demo-stm32f4-discovery\app.toml --archive`
target\demo-stm32f4-discovery\dist\default\build-demo-stm32f4-discovery.zip
```

Additonally, I wanted to make the errors nice for this, and so I ended up adding some contexts inside of the other commands, which will also make them slightly better. Example:

```
〉cargo xtask print .\app\demo-stm32f4-discovery\app.tom --archive
    Finished dev [optimized + debuginfo] target(s) in 1.52s
     Running `target\debug\xtask.exe print .\app\demo-stm32f4-discovery\app.tom --archive`
Error: could not print information about the build

Caused by:
    0: could not create build configuration
    1: could not read .\app\demo-stm32f4-discovery\app.tom
    2: The system cannot find the file specified. (os error 2)
```

This feature was requested by @rmustacc and @jclulow , so I'd love to make sure that this satisfies their needs.

I could make the command just `print-archive` (suggested by @mkeeter) and remove some of the machinery here, but I feel like we've wanted to print other stuff in the past? IIRC @cbiffle was talking about something with environment variables? maybe that was just setting them, not examining them.

### DIFF
--- a/build/xtask/src/print.rs
+++ b/build/xtask/src/print.rs
@@ -1,0 +1,34 @@
+use std::path::Path;
+
+use anyhow::{bail, Context, Error, Result};
+
+use crate::dist::PackageConfig;
+
+pub fn run(
+    cfg: &Path,
+    archive: bool,
+    image_name: Option<String>,
+) -> Result<()> {
+    if archive {
+        let config = PackageConfig::new(cfg, false, false)
+            .context("could not create build configuration")?;
+
+        let image_name = image_name.unwrap_or(String::from("default"));
+
+        let image_name = config
+            .toml
+            .image_names
+            .iter()
+            .find(|name| name == &&image_name)
+            .ok_or(Error::msg(format!("cannot find image {}", image_name)))?;
+
+        let final_path = config
+            .img_file(format!("build-{}.zip", config.toml.name), image_name);
+
+        println!("{}", final_path.display());
+    } else {
+        bail!("I'm not sure what to print. Currently supported: --archive");
+    }
+
+    Ok(())
+}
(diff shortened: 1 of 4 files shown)


build/xtask/src/print.rs
```diff
@@ -0,0 +1,38 @@
...
+            .iter()
+            .find(|name| name == &&image_name)
+            .ok_or(Error::msg(format!("cannot find image {}", image_name)))?;
+
+        let final_path = config
+            .img_file(format!("build-{}.zip", config.toml.name), image_name);
+
+        println!("{}", final_path.display());
+    } else {
+        bail!("I'm not sure what to print. Currently supported: --archive");
```
steveklabnik, line 34: I'd prefer to use an ArgGroup here, but it seems like it's not fully baked in the derive API yet https://github.com/clap-rs/clap/issues/2621

No big deal.

build/xtask/src/print.rs
```diff
@@ -0,0 +1,38 @@
...
+
+        let image_name = config
+            .toml
+            .image_names
+            .iter()
+            .find(|name| name == &&image_name)
+            .ok_or(Error::msg(format!("cannot find image {}", image_name)))?;
+
+        let final_path = config
+            .img_file(format!("build-{}.zip", config.toml.name), image_name);
```
steveklabnik, line 30: I am a *little* not happy about this one, as it's duplicating the logic that's already been duplicated twice:

* https://github.com/oxidecomputer/hubris/blob/0b29eb59b5910c0279bfb4c16913565711005c34/build/xtask/src/humility.rs#L29
* https://github.com/oxidecomputer/hubris/blob/60ecfe39308288478914c35ffbf96ba826e9fe7c/build/xtask/src/dist.rs#L579

Not 100% sure what the best way to solve this would be, though...

mkeeter (approved):

PR #906, opened 2022-11-02
xtask: make allowed-features list easier to maintain

cbiffle:
Rather than a wrapped string constant that doesn't allow any internal whitespace, the features are now listed in arrays and assembled into a string when needed. This (IMO) makes it way easier to scan and add/remove features.


steveklabnik (approved):

steveklabnik:
This is great, kinda embarrassed I didn't do it this way the first time, hah.

PR #912, opened 2022-11-03
Implement logical to physical mapping for transceiver ports

Aaron-Hartwig:


drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -22,74 +234,111 @@ impl Transceivers {
...
-            self.fpgas[1].read(Addr::QSFP_CTRL_EN_H)?;
+            self.fpga(FpgaController::Left).read(Addr::QSFP_CTRL_EN_H)?;
+        let f1: [U16<byteorder::BigEndian>; 7] = self
+            .fpga(FpgaController::Right)
+            .read(Addr::QSFP_CTRL_EN_H)?;
+
+        let mut status_masks: [u32; 7] = [0; 7];
+
+        // loop through the 7 different fields we need to map
+        for (i, mask) in status_masks.iter_mut().enumerate() {
```
mkeeter, line 254: The iteration order was a little confusing to me.  How about switching the loop order (and extracting constants into the outer loop), e.g.

```rust
        // loop through each port
        for (port, port_loc) in PORT_MAP.iter().enumerate() {
            // get a mask for where the current logical port is mapped
            // locally on the FPGA
            let local_port_mask = 1 << port_loc.port;

            // get the relevant data from the correct FPGA
            let local_data = match port_loc.controller {
                FpgaController::Left => &f0,
                FpgaController::Right => &f1,
            };
            // loop through the 7 different fields we need to map
            for (word, out) in local_data.iter().zip(status_masks.iter_mut()) {
                // if the bit is set, update our status mask at the correct
                // logical position
                let word: u16 = (*word).into();
                if (word & local_port_mask) != 0 {
                    *out |= 1 << port;
                }
            }
        }
```

mkeeter (approved):

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: Yeah, good suggestion. I was in too deep at the moment and the weirdness there did not occur to me.

PR #916, opened 2022-11-09
Add minimal Ignition driver and task

arjenroodselaar:
This diff adds a first cut of the Ignition Controller driver and task. A minimal API is provided by the Idol and `ignition-api` crate. Currently we're behind on documentation for this sub-system (near the top of my list of things to work on), so happy to answer questions. The basics;

- Call `port_count` to determine how many ports are in the system (35 for a rack, in some test setups this may be smaller)
- Poll `presence_summary` for a bit-vector which indicates which ports have a device/Target present
- Call `state` for a given port and use the `PortState` and `Target` types to learn about the specifics of a Target
- Use `send_request` to send a `SystemPowerOn`, `SystemPowerOff` or `SystemReset` request, which will return an appropriate error if no Target is present or a request is already in progress

I will be expanding on this API for some more in-dept logging and diagnostics of link events, but that'll be in follow-up diffs.


drv/ignition-api/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+bitfield! {
+    #[derive(Copy, Clone, Debug, PartialEq, FromPrimitive, ToPrimitive, FromBytes, AsBytes)]
+    #[repr(C)]
+    pub struct PortState(u64);
+    pub target_present, _: 0;
+}
+
+impl PortState {
```
mkeeter, line 47: You can move some of this block into the `bitfield!` type, e.g.
```rust
bitfield! {
    // ...
    pub u8, raw_receiver_status, _: 15, 8;
}
```

Alternatively, if you implement `impl From<u8> for ReceiverStatus`, you could do _everything_ in the `bitfield!` macro, removing `pub fn receiver_status` entirely:
```rust
bitfield! {
    // ...
    pub u8, into ReceiverStatus, raw_receiver_status, _: 15, 8;
}
```

You can't do exactly the same thing for `target`, since it's conditional on bit 0, but maybe adding `raw_target, _: 63, 16` then calling `raw_target()` in `pub fn target`?

(Note that this would add a shift; you'd need to adjust bit offsets in `struct Target` below)

mkeeter: Looking at this a little more, I'm not sure why we're using `FromPrimitive` and `ToPrimitive` for newtypes in this file **in general**.  I'd suggest removing them whenever possible, and instead using `derive_more::{From, Into}`, i.e.

```rust
use derive_more::{From, Into};
// ...
bitfield! {
    #[derive(Copy, Clone, Debug, PartialEq, From, Into)]
    #[repr(C)]
    pub struct ReceiverStatus(u8);
    pub aligned, _: 0;
    pub locked, _: 1;
    pub polarity_inverted, _: 2;
}
```

That means that you'd be able to use the `pub u8, into ReceiverStatus, receiver_status, _: 15, 8;` form and remove a bunch of hand-written functions.

drv/ignition-api/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub reserved_fault1, _: 34;
+    pub reserved_fault2, _: 35;
+    pub sp_fault, _: 36;
+    pub rot_fault, _: 37;
+    pub system_power_off_in_progress, _: 40;
+    pub system_power_on_in_progress, _: 41;
+    pub system_reset_in_progress, _: 42;
+}
+
+impl Target {
```
mkeeter, line 88: Same idea here: we should use `derive(From)` (from `derive_more`) on `SystemType` and `ReceiverStatus`, then could do
```
bitfield! {
    // ...
    pub u8, into SystemType, system_type, _: 23, 16;
}
```
and delete this function entirely (same for `link0/1_receiver_status`)

drv/ignition-api/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub fn system_type(&self) -> SystemType {
+        SystemType(self.0.as_bytes()[2])
+    }
+
+    pub fn system_power_state(&self) -> PowerState {
+        match self.0.as_bytes()[3] & 0x4 != 0 {
```
mkeeter, line 94: I'd vote to put this into the bitfield as `raw_system_power_state`, to keep all the bit math in one place.

drv/ignition-api/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+    Copy, Clone, Debug, PartialEq, FromPrimitive, ToPrimitive, AsBytes,
+)]
+#[repr(u8)]
+pub enum Request {
+    SystemPowerOff = 1,
+    SystemPowerOn = 2,
+    SystemReset = 3,
+}
+
+impl From<Request> for u8 {
```
mkeeter, line 141: Using `derive_more::Into` on `Request` gets this for free!

mkeeter, line 141: I'm mad that we can't derive this automatically (using either `derive_more` or `num_derive`, but that appears to be the case!

drv/ignition-server/src/main.rs
```diff
@@ -0,0 +1,243 @@
...
+                drv_sidecar_seq_api::Sequencer::from(SEQUENCER.get_task_id());
+
+            // Poll the sequencer to determine if the mainboard controller is
+            // ready.
+            ringbuf_entry!(Trace::AwaitingMainboardControllerReady);
+            while !sequencer.mainboard_controller_ready().unwrap_or(false) {
+                hl::sleep_for(25);
+            }
+        }
+    }
```
mkeeter, line 59: Using `cfg_if` is overkill here, since it's a single feature-based conditional:
```suggestion
    #[cfg(feature = "sequencer")]
    {
        let sequencer =
            drv_sidecar_seq_api::Sequencer::from(SEQUENCER.get_task_id());

        // Poll the sequencer to determine if the mainboard controller is
        // ready.
        ringbuf_entry!(Trace::AwaitingMainboardControllerReady);
        while !sequencer.mainboard_controller_ready().unwrap_or(false) {
            hl::sleep_for(25);
        }
    }
```

drv/ignition-server/src/main.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    fn handle_notification(&mut self, _bits: u32) {
+        let start = sys_get_timer().now;
+
+        // Only poll the presence summary if the port count seems reasonable. A
+        // count of 0xff may occur if the FPGA is running an incorrect
+        // bitstream.
+        if self.port_count > 0 && self.port_count != 0xff {
+            if let Err(_e) = self.poll_presence() {}
```
mkeeter, line 220: Maybe log this to the ringbuf, since we've got one?

drv/ignition-server/src/main.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        // We now know when we were notified and when any work was completed.
+        // Note that the assumption here is that `start` < `finish` and that
+        // this won't hold if the system time rolls over. But, the system timer
+        // is a u64, with each bit representing a ms, so in practice this should
+        // be fine. Anyway, armed with this information, find the next deadline
+        // some multiple of `TIMER_INTERVAL` in the future.
+
+        let delta = finish - start;
+        let next_deadline = finish + TIMER_INTERVAL - (delta % TIMER_INTERVAL);
```
mkeeter, line 233: This is a little funky – what's the advantage versus doing `let next_deadline = finish + TIMER_INTERVAL`?

drv/sidecar-mainboard-controller/src/ignition.rs
```diff
@@ -0,0 +1,143 @@
...
+            + u16::from(offset)
+    }
+
+    fn read_port_register<T>(
+        &self,
+        port: u8,
+        offset: IgnitionAddr,
+    ) -> Result<T, FpgaError>
+    where
+        T: AsBytes + Default + FromBytes,
```
mkeeter, line 38:
```suggestion
        T: AsBytes + FromBytes,
```

drv/sidecar-mainboard-controller/src/ignition.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    fn write_port_register<T>(
+        &self,
+        port: u8,
+        offset: IgnitionAddr,
+        value: T,
+    ) -> Result<(), FpgaError>
+    where
+        T: AsBytes + Default + FromBytes,
```
mkeeter, line 50:
```suggestion
        T: AsBytes,
```

mkeeter:
This generally looks good, with a few comments.

You should also run `cargo xtask clippy app/sidecar/rev-a.toml` and fix whatever Clippy complains about!

drv/ignition-server/src/main.rs
```diff
@@ -0,0 +1,243 @@
...
+        TIMER_NOTIFICATION_MASK
+    }
+
+    fn handle_notification(&mut self, _bits: u32) {
+        let start = sys_get_timer().now;
+
+        // Only poll the presence summary if the port count seems reasonable. A
+        // count of 0xff may occur if the FPGA is running an incorrect
+        // bitstream.
+        if self.port_count > 0 && self.port_count != 0xff {
```
jgallagher, line 219: Should the `port_count()` idol function return an error if the port count is `0xff` too?

drv/sidecar-mainboard-controller/src/ignition.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    #[inline]
+    fn port_addr(&self, port: u8, offset: IgnitionAddr) -> u16 {
+        u16::from(Addr::IGNITION_CONTROLLERS_COUNT)
+            + 0x100
+            + (0x100 * port as u16)
+            + u16::from(offset)
+    }
+
+    fn read_port_register<T>(
```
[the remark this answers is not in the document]
jgallagher: The methods these call on `FpgaUserDesign` may also need `inline` to see a difference (they're small generic wrappers around non-generic implementations).

drv/sidecar-mainboard-controller/src/ignition.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Clear the events for the given port, link.
+    pub fn clear_link_events(
+        &self,
+        port: u8,
+        link: LinkSelect,
+    ) -> Result<(), FpgaError> {
+        self.write_port_register(
+            port,
+            Self::link_events_addr(link),
+            LinkEvents(0b111111), // Clear all events
```
jgallagher, line 114: Could this be a constant on `LinkEvents`?

drv/ignition-server/src/main.rs
arjenroodselaar: I did not realize this syntax was an option. But thinking about how decorators work it makes sense, thanks!

drv/ignition-server/src/main.rs
arjenroodselaar: `let next_deadline = finish + TIMER_INTERVAL` guarantees `TIMER_INTERVAL` between notifications while the method above stays close to a specific periodic point in time, irrespective of how long work takes, at the expense of skipping a "frame" if work pushes past what would have been the next deadline.

This code comes from the sequencer where I feel this matters more. What's here will be replaced by interrupts soon, so it's probably fine either way.

drv/sidecar-mainboard-controller/src/ignition.rs
arjenroodselaar: Done, thanks!

arjenroodselaar:
> > Call port_count to determine how many ports are in the system (35 for a rack, in some test setups this may be smaller)
> 
> Just double-checking my understanding: 35 is for 32 sleds + 2 sidecars + 1 psc?

32 sleds + 2 PSCs (currently only one in use, but the rack has an open slot for another power shelve) + 1 Sidecar (a Sidecar only connect to it's peer, not itself)

A Sidecar could technically connect to itself and issue a `SystemReset` and that would work, because the Target would handle the power up. But lets keep this sane :)

drv/ignition-server/build.rs
```diff
@@ -0,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    build_util::expose_target_board();
```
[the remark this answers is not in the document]
arjenroodselaar: We don't. This is copy/pasta since most tasks do seem to need it. Removing.

drv/ignition-api/src/lib.rs
arjenroodselaar: This cleans up nicely, thanks!

drv/sidecar-mainboard-controller/src/ignition.rs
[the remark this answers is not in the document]
arjenroodselaar: The compiler may have already reached the conclusion that all this can be inlined since explicitly requesting it doesn't seem to change the size. But either way it seems like a reasonable thing to do.

mkeeter (approved):

PR #920, opened 2022-11-10
Ignition API refinements, add state_dump(..) function

arjenroodselaar:
This diff refines the Ignition API after working on some UI. A `state_dump(..)` function is added to make rendering a status overview more economical. And finally the `userlib` dependency is put behind the `idol-client` feature flag, allowing the API crate to be used as a Humility dependency (`userlib` depends on `armv6m-atomic-hack`, which can't compile on non-small-ARM platforms).


mkeeter (approved):

drv/ignition-server/src/main.rs
```diff
@@ -29,6 +29,7 @@ ringbuf!(Trace, 16, Trace::None);
 
 const TIMER_NOTIFICATION_MASK: u32 = 1 << 0;
 const TIMER_INTERVAL: u64 = 1000;
+const PORT_MAX: usize = 40;
```
Aaron-Hartwig, line 32: Why is this 40 instead of 35? Perhaps I'm misunderstanding the purpose here or am lacking knowledge of where the other 5 Ignition devices come from. 

32 sleds
2  PSCs
1  other switch
35 total devices

drv/ignition-api/src/lib.rs
```diff
@@ -66,6 +61,24 @@ bitfield! {
...
+    pub reserved1, _: 2;
+    pub reserved2, _: 3;
+    pub sp, _: 4;
+    pub rot, _: 5;
+}
+
+impl SystemFaults {
+    pub fn count(&self) -> usize {
+        self.0 as usize
+    }
```
Aaron-Hartwig, line 79: I think this just returns the value of `self.power_a3`. I believe you want to count the ones in the bitfield.

Aaron-Hartwig, line 79: I think this just returns the value of `self.power_a3`. I believe you want to count the ones in the bitfield.

Aaron-Hartwig (changes requested):

drv/ignition-server/src/main.rs
```diff
@@ -210,6 +205,30 @@ impl idl::InOrderIgnitionImpl for ServerImpl {
             .map_err(IgnitionError::from)
             .map_err(RequestError::from)
     }
+
+    fn state_dump(
```
mkeeter, line 209: You _could_ implement this as a HIF loop on the Humility side, e.g. something like [this](https://github.com/oxidecomputer/humility/blob/master/cmd/monorail/src/lib.rs#L922-L948).

drv/ignition-server/src/main.rs
arjenroodselaar: The number of controllers is actually variable, determined at RTL synthesis time and published through a register (see `port_count()` at https://github.com/oxidecomputer/hubris/blob/090ad7345c1252ce032a32980d5ffca94bddb61b/drv/sidecar-mainboard-controller/src/ignition.rs#L59). The reason for this is so I can have smaller builds for development (the emulator top has 4 controller ports for example) and scale the Hubris driver accordingly. Currently the `presence_summary` vector allows for up to 40 ports (35 rounded up to a multiple of 8). This `PORT_MAX` constant mirrors that limit and is used below to statically allocate the array used to shuttle the state dump between Ignition and the control plane and/or Humility.

I'll add a comment.

drv/ignition-api/src/lib.rs
arjenroodselaar: Good catch, thanks. It was late last night :)

drv/ignition-server/src/main.rs
arjenroodselaar: That was actually my first idea, but after some back and forth with @bcantrill about the intent/limitations of HIF I decided to add this function which can also be used by the control plane/technician port functions down the road.

drv/ignition-api/src/lib.rs
```diff
@@ -162,4 +170,24 @@ pub enum LinkSelect {
     TargetLink1 = 3,
 }
 
-include!(concat!(env!("OUT_DIR"), "/client_stub.rs"));
+cfg_if::cfg_if! {
+    if #[cfg(feature = "idol-client")] {
+        use drv_fpga_api::FpgaError;
+        use idol_runtime::ServerDeath;
+        use userlib::sys_send;
```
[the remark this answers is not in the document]
arjenroodselaar: Yes, the include of `client_stubs.rs` a few lines further down relies on `sys_send`.

PR #922, opened 2022-11-10
Fetch host phase2 blocks from MGS

jgallagher:
The bulk of this PR is in the new `host_phase2` module, which is responsible for sending requests to MGS (potentially on both ports) to ask for phase 2 data in response to `host-sp-comms` asking `control-plane-agent` to do so. There are a handful of minor changes to `host-sp-messages` to track the latest RFD 316 updates (some of which may be so new to have not even landed yet!).

### DIFF
--- a/task/control-plane-agent/src/main.rs
+++ b/task/control-plane-agent/src/main.rs
@@ -9,6 +9,7 @@ use gateway_messages::{
     sp_impl, IgnitionCommand, MgsError, PowerState, SpComponent, SpPort,
     UpdateId,
 };
+use host_sp_messages::HostStartupOptions;
 use idol_runtime::{Leased, NotificationHandler, RequestError};
 use mutable_statics::mutable_statics;
 use ringbuf::{ringbuf, ringbuf_entry};
@@ -104,6 +105,8 @@ enum MgsMessage {
         message_id: u32,
         err: MgsError,
     },
+    GetStartupOptions,
+    SetStartupOptions(gateway_messages::StartupOptions),
 }
 
 ringbuf!(Log, 16, Log::Empty);
@@ -195,6 +198,24 @@ impl idl::InOrderControlPlaneAgentImpl for ServerImpl {
     ) -> Result<usize, RequestError<ControlPlaneAgentError>> {
         self.mgs_handler
             .get_host_phase2_data(image_hash, offset, data)
+    }
+
+    fn get_startup_options(
+        &mut self,
+        _msg: &userlib::RecvMessage,
+    ) -> Result<HostStartupOptions, RequestError<ControlPlaneAgentError>> {
+        self.mgs_handler.startup_options()
+    }
+
+    fn set_startup_options(
+        &mut self,
+        _msg: &userlib::RecvMessage,
+        startup_options: u64,
+    ) -> Result<(), RequestError<ControlPlaneAgentError>> {
+        let startup_options = HostStartupOptions::from_bits(startup_options)
+            .ok_or(ControlPlaneAgentError::InvalidStartupOptions)?;
+
+        self.mgs_handler.set_startup_options(startup_options)
     }
 }
 
@@ -290,15 +311,13 @@ impl NetHandler {
         // `MgsHandler` implementation, and serializing the response we should
         // send into `self.tx_buf`.
         assert!(self.packet_to_send.is_none());
-        let n = sp_impl::handle_message(
+        if let Some(n) = sp_impl::handle_message(
             sender,
             sp_port_from_udp_metadata(&meta),
             &self.rx_buf[..meta.size as usize],
             mgs_handler,
             self.tx_buf,
-        );
-
-        if n > 0 {
+        ) {
             meta.size = n as u32;
             self.packet_to_send = Some(meta);
         }
@@ -338,6 +357,8 @@ const fn usize_max(a: usize, b: usize) -> usize {
 }
 
 mod idl {
-    use task_control_plane_agent_api::ControlPlaneAgentError;
+    use task_control_plane_agent_api::{
+        ControlPlaneAgentError, HostStartupOptions,
+    };
     include!(concat!(env!("OUT_DIR"), "/server_stub.rs"));
 }
--- a/task/control-plane-agent/src/main.rs
+++ b/task/control-plane-agent/src/main.rs
@@ -6,7 +6,8 @@
 #![no_main]
 
 use gateway_messages::{
-    sp_impl, IgnitionCommand, PowerState, SpComponent, SpPort, UpdateId, MgsError,
+    sp_impl, IgnitionCommand, MgsError, PowerState, SpComponent, SpPort,
+    UpdateId,
 };
 use idol_runtime::{Leased, NotificationHandler, RequestError};
 use mutable_statics::mutable_statics;
--- a/task/control-plane-agent/src/main.rs
+++ b/task/control-plane-agent/src/main.rs
@@ -6,7 +6,7 @@
 #![no_main]
 
 use gateway_messages::{
-    sp_impl, IgnitionCommand, PowerState, SpComponent, SpPort, UpdateId,
+    sp_impl, IgnitionCommand, PowerState, SpComponent, SpPort, UpdateId, MgsError,
 };
 use idol_runtime::{Leased, NotificationHandler, RequestError};
 use mutable_statics::mutable_statics;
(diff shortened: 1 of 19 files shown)


task/control-plane-agent/src/main.rs
```diff
@@ -290,8 +298,10 @@ impl NetHandler {
             self.tx_buf,
         );
 
-        meta.size = n as u32;
-        self.packet_to_send = Some(meta);
+        if n > 0 {
```
mkeeter, line 301: The correctness here isn't obvious to me.  Should `handle_message` return an `Option<usize>` instead, to indicate whether or not to send a reply?

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
```diff
@@ -0,0 +1,285 @@
...
+
+// We only support a single in-flight host phase 2 request at a time (i.e., we
+// do not support any pipelining). When the host makes a request of us (via
+// `host-sp-comms`, which calls our idol interface on its behalf), we may not
+// know _which_ MGS instance is able to provide the data it wants, and given
+// we're using UDP, we need to support a small number of retries. Our constants
+// here control how long we wait before we try "the other" MGS (assuming we've
+// picked one to start with), and then how long we wait before switching back
+// and retrying both. Assuming no successful replies from either MGS instance,
+// our timeline for a request will be:
```
mkeeter, line 27: Per my usual feedback, it would be helpful to have a block diagram showing
a) Which Hubris tasks are sending data in which direction, and
b) Which interfaces on the host side are sending / receiving that data

I think the request comes from the serial port and the data comes from the network, but am not 100% sure!

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
```diff
@@ -0,0 +1,285 @@
...
+const MAX_ATTEMPTS: u8 = 3;
+
+pub(crate) struct HostPhase2Requester {
+    current: Option<CurrentRequest>,
+    last_responsive_mgs: SpPort,
+    buffer: &'static mut Vec<u8, { gateway_messages::MAX_SERIALIZED_SIZE }>,
+}
+
+impl HostPhase2Requester {
+    pub(crate) fn claim_static_resources() -> Self {
```
mkeeter, line 50: Add a comment saying this can only be called once

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
```diff
@@ -0,0 +1,285 @@
...
+        offset: u64,
+        data: &[u8],
+    ) {
+        let current = match self.current.as_mut() {
+            Some(current)
+                if hash == current.hash && offset == current.offset =>
+            {
+                current
+            }
+            _ => return,
```
mkeeter, line 185: Is this... bad?  Or okay?

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
```diff
@@ -0,0 +1,285 @@
...
+        };
+
+        self.buffer.clear();
```
mkeeter, line 188: Is it possible for there to be data in the buffer that has yet to be used?

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
```diff
@@ -0,0 +1,285 @@
...
+        let n = usize::min(self.buffer.capacity(), data.len());
+        self.buffer.extend_from_slice(&data[..n]).unwrap_lite();
```
mkeeter, line 190: Is it okay to discard the tail of `data` here?

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
```diff
@@ -0,0 +1,285 @@
...
+                } else {
+                    None
+                }
+            }
+        }
+    }
+
+    fn timer_deadline(&self) -> Option<u64> {
+        match self {
+            State::NeedToSendFirstMgs(_) => Some(sys_get_timer().now + 1),
```
mkeeter, line 262: Why not `sys_get_timer().now`, to trigger an immediate send?

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
jgallagher: It's okay, because we're all UDP with retries and resends; it just means MGS sent us data that we don't currently want (presumably it's data we used to want). Imagine something like:

0. Host requests data at offset 0
1. We send a request for offset 0 to MGS 0
2. No response within our timeout window, so we send a request for offset 0 to MGS 1
3. MGS 1 responds with data at offset 0; we forward it back to the host
4. Host requests data at offset N
5. We send a request for offset N to MGS 1
6. MGS 0 finally replies to the request for data at offset 0 we sent back in step 1; we no longer care about this data, so we ignore it here.

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
jgallagher: Good catch; I was copy/pasting from somewhere else where the + 1 was useful. Will change.

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
jgallagher: It shouldn't be possible to lose anything from `data` here, assuming it came from an MGS-sent packet (which technically this function doesn't know for sure, but in practice is true). `self.buffer` is sized large enough to hold a full UDP packet worth of data plus a bit, since packets all have some kind of metadata at the front but `buffer` is only given the payload. We could change this to assert that `data.len() <= self.buffer.capacity()`.

However, it's also _okay_ if we did end up losing data here somehow; the host requests for data only include the offset where the data should start, not a length. If we ended up dropping the tail of data here, we would have wasted the bandwidth spent to transfer that data, but nothing else would break. Something like:

1. Host requests data at offset 0
2. MGS sends us 1000 bytes starting at offset 0
3. Somehow `self.buffer` is undersized, so we only copy in 900 bytes
4. We give 900 bytes to the host
5. Host requests data at offset 900

would cause MGS to send the data we dropped (bytes 900-1000) twice but would otherwise function correctly.

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
jgallagher: It's possible we could get multiple packets from MGS that all contain the same data, but otherwise no. Maybe it would be clearer to do something like:

* if our state is already `Fetched`, just return
* if our state is not `Fetched`, assert that `self.buffer.is_empty()`

? The current implementation will always copy in the latest data from MGS (although it should only be sending exact copies, if we get multiples like this).

task/control-plane-agent/src/main.rs
jgallagher: Agreed, but `handle_message` is defined in another crate; I'll fix this in a followup PR.

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
jgallagher: Added a comment explaining this

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
jgallagher: Made that change in 77e52be88c6ae4563c0dccc98efaea18a2cf9728

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
jgallagher: Added this to a comment in 77e52be88c6ae4563c0dccc98efaea18a2cf9728

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
jgallagher: I didn't go full block diagram, but added a step-by-step in 77e52be88c6ae4563c0dccc98efaea18a2cf9728 - is that good enough?

task/control-plane-agent/src/main.rs
jgallagher: This is fixed by #927.

mkeeter (approved):

PR #925, opened 2022-11-11
lpc55-sprot-server: Pin repo hash to current HEAD.

flihp:
This should allow us to make incompatible changes upstream w/o breaking stuff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3512,7 +3512,7 @@ dependencies = [
 [[package]]
 name = "sprockets-common"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/sprockets.git#77df31efa5619d0767ffc837ef7468101608aee9"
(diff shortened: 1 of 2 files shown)


flihp:
The end goal is to get all use of the salty library in hubris on our mirror (for zeroize). To do so we need to make incompatible changes upstream and coordinate here. The order of operations should be:
* merge this PR
* migrate sprockets to oxide fork: https://github.com/oxidecomputer/sprockets/pull/39
* migrate lpc55-sprot-server to oxide fork of salty & sprockets rev post merge of https://github.com/oxidecomputer/sprockets/pull/39

mkeeter:
Because we've committed the `Cargo.lock` file (which already pinned the version), I don't think this is actually necessary!

flihp:
Missed that secton of the cargo book (https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html#cargotoml-vs-cargolock). The other dependencies (hubpack) in this `Cargo.toml` manually pinned probably aren't necessary either.

PR #926, opened 2022-11-11
Get our hubpacks from one place: the crates.io registry.

flihp:
This is minor cleanup. These toml files were probably created before hubpack was published to the registry.


mkeeter:
I'd suggest _not_ merging this, because the workspace dependency inheritance PR #932 does the same thing, and I don't want to rebase it again 😆

mkeeter:
Yup, it's real broken now; closing!

PR #931, opened 2022-11-14
Transceiver management API over UDP

mkeeter:
This PR adds a UDP API for transceiver management.

I'm opening it early as a draft, since both the FPGA image and higher-level APIs are still in flux.


drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+                    .map_err(|_e| Error::ReadFailed)?;
+                let lpmode: U16<BigEndian> = fpga
+                    .read(Addr::QSFP_CTRL_LPMODE_H)
+                    .map_err(|_e| Error::ReadFailed)?;
+                let present: U16<BigEndian> = fpga
+                    .read(Addr::QSFP_STATUS_PRESENT_H)
+                    .map_err(|_e| Error::ReadFailed)?;
+                let irq: U16<BigEndian> = fpga
+                    .read(Addr::QSFP_STATUS_IRQ_H)
+                    .map_err(|_e| Error::ReadFailed)?;
```
Aaron-Hartwig, line 185: `transceivers` also has a `get_module_status` function that may be of use to you here. It will get all of these things in one larger read. Then you can go about masking them off below. https://github.com/oxidecomputer/hubris/blob/master/drv/sidecar-front-io/src/transceivers.rs#L241-L276

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+                // TODO: check equality here and return a different error?
+                if data_size as usize > data.len() {
+                    return Err(Error::RequestTooLarge);
+                }
+                self.write(mem, modules, data)?;
+                Ok((HostResponse::Write(mem), 0))
+            }
+            HostRequest::SetPowerMode(mode) => {
+                let mask = get_mask(modules)?;
+                // TODO: do we need delays in between any of these operations?
```
Aaron-Hartwig, line 231: Given that the FPGA will soon manage the details of these transitions (per RFD 244), I don't think so.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+        let mask = get_mask(modules)?;
+        match upper_page {
+            UpperPage::Cmis(page) => {
+                self.transceivers
+                    .set_i2c_write_buffer(&[page.page()])
+                    .map_err(|_e| Error::ReadFailed)?;
+                self.transceivers
+                    .setup_i2c_write(0x7F, 1, mask)
+                    .map_err(|_e| Error::ReadFailed)?;
+                sleep_for(5); // TODO: Poll instead
```
Aaron-Hartwig, line 290: As of https://github.com/oxidecomputer/hubris/commit/ae111d6b5a027c1c987cc7ce8e8eb5c663cb82ca, you can now poll busy registers.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+
+                if let Some(bank) = page.bank() {
+                    self.transceivers
+                        .set_i2c_write_buffer(&[bank])
+                        .map_err(|_e| Error::ReadFailed)?;
+                    self.transceivers
+                        .setup_i2c_write(0x7E, 1, mask)
+                        .map_err(|_e| Error::ReadFailed)?;
+                    sleep_for(5); // TODO: Poll instead
```
Aaron-Hartwig, line 299: Same here.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+                }
+            }
+            UpperPage::Sff8636(page) => {
+                self.transceivers
+                    .set_i2c_write_buffer(&[page.page()])
+                    .map_err(|_e| Error::ReadFailed)?;
+                self.transceivers
+                    .setup_i2c_write(0x7F, 1, mask)
+                    .map_err(|_e| Error::ReadFailed)?;
+                sleep_for(5); // TODO: Poll instead
```
Aaron-Hartwig, line 309: Same here.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+        // Copy data into the FPGA write buffer
+        self.transceivers
+            .set_i2c_write_buffer(&data[..mem.len() as usize])
+            .map_err(|_e| Error::WriteFailed)?;
+
+        // Trigger a multicast write to all transceivers in the mask
+        self.transceivers
+            .setup_i2c_write(mem.offset(), mem.len(), get_mask(modules)?)
+            .map_err(|_e| Error::WriteFailed)?;
+        sleep_for(5); // TODO: wait for completion
```
Aaron-Hartwig, line 406: There are busy registers to poll now.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+
+        Ok(())
+    }
+}
```
Aaron-Hartwig, line 410: As I noted in a few places above, as of https://github.com/oxidecomputer/hubris/commit/ae111d6b5a027c1c987cc7ce8e8eb5c663cb82ca we have registers to poll for I2C bus busy per port. Additionally, there are error registers we should also be checking to make sure the transaction has not only completed, but completed successfully.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -298,46 +306,108 @@ impl Transceivers {
...
+        mask: M,
+    ) -> Result<(), FpgaError> {
+        self.setup_i2c_op(false, reg, num_bytes, mask)
     }
 
     /// Initiate an I2C operation on all ports per the specified `mask`. When
     /// `is_read` is true, the operation will be a random-read, not a pure I2C
     /// read. The maximum value of `num_bytes` is 128.
-    pub fn setup_i2c_op(
+    pub fn setup_i2c_op<M: Into<FpgaPortMasks>>(
```
jgallagher, line 403: Does this still need to be `pub` given the addition of `setup_i2c_{read,write}`?

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+                    .unwrap() as u32,
+                };
+
+                self.net
+                    .send_packet(
+                        SOCKET,
+                        meta,
+                        &tx_data_buf[..meta.size as usize],
+                    )
+                    .unwrap();
```
jgallagher, line 105: Is this temporary / part of the draft? Handling `QueueFull` in particular might be slightly annoying, so maybe worth doing sooner rather than later.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
+//! Hardware revision-independent server code for UDP interactions
```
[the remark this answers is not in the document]
mkeeter: Yes, I'm surprised the bot didn't complain about it!

drv/transceivers-server/src/net_api.rs
mkeeter: I looked at that, but it's not quite what I want: `get_module_status` returns _logical_ port status, and here we want _physical_ port status (since the host side is doing the physical -> logical decoding itself).

drv/sidecar-front-io/src/transceivers.rs
mkeeter: It did because it was exposed by the Idol API, but I'm going to change that as well.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,414 @@
...
+        &mut self,
+        mem: MemoryRegion,
+        modules: ModuleId,
+        out: &mut [u8],
+    ) -> Result<(), Error> {
+        // Reading beyond a 128-byte page boundary will wrap around; return an
+        // error if that's going to happen.
+        let page_end = (mem.offset() as u32 / 128 + 1) * 128;
+        if mem.offset() as u32 + mem.len() as u32 > page_end {
+            // TODO use a more descriptive error name here?
```
bnaecker, line 329: Yeah, I've got a [PR out now](https://github.com/oxidecomputer/transceiver-control/pull/11) that reworks the `MemoryRegion` into length-checked `Memory{Read,Write}`. A better name or more descriptive variants would be helpful, and I'm open to suggestions.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,414 @@
...
+            return Err(Error::InvalidMemoryAccess {
+                offset: mem.offset(),
+                len: mem.len(),
+            });
+        }
+        // CMIS spec only guarantee an 8-byte read (?!)
+        // (see CMIS 5.0, 5.2.2.1)
+        if matches!(mem.upper_page(), UpperPage::Cmis(..)) && mem.len() > 8 {
```
bnaecker, line 337: This check should be obviated by the type system once the rework PR I mentioned above is merged (assuming it _is_ merged 😄).

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,414 @@
...
+        self.transceivers
+            .setup_i2c_read(mem.offset(), mem.len(), get_mask(modules)?)
+            .map_err(|_e| Error::ReadFailed)?;
+        sleep_for(5); // TODO: wait for completion
+
+        let controller = match modules.fpga_id {
+            0 => FpgaController::Left,
+            1 => FpgaController::Right,
+            i => return Err(Error::InvalidFpga(i)),
+        };
```
bnaecker, line 352: It seems like this should be at the top of the function, but I'm not sure how much state `self.select_page` or `self.transceivers.setup_i2c_read` mutate along the way.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,414 @@
...
+                }
+            }
+            // SFF-8636, §5.3.3: max of 4 bytes in one write transaction
+            UpperPage::Sff8636(..) => {
+                if mem.len() > 4 {
+                    // TODO: more specialized error
+                    return Err(Error::RequestTooLarge);
+                }
+            }
+        }
```
bnaecker, line 396: Actually, _all_ of these checks should be obviated by the type system (once the PR mentioned above is merged). It won't be possible to construct a read for greater than 128 bytes (SFF-8636) or 8 bytes (CMIS), or a read greater than 4 (SFF) or 8 bytes (CMIS).

bnaecker (commented):
Nice work, looks solid overall! I've a few questions about error-handling and some comments related to updates that are in the works on the message API itself. Thanks!

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,410 @@
...
+                            msg_len,
+                        );
+                        (msg_len + data_len) as u32
+                    }
+                    Err(_e) => hubpack::serialize(
+                        tx_data_buf,
+                        &Message {
+                            header: Header {
+                                version: 1,
+                                message_id: u64::MAX,
```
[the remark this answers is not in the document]
mkeeter: Yeah, this branch is only taken when deserialization fails (which shouldn't happen, considering UDP has checksums, but 🤷🏻 ).

I think it makes sense to try deserializing just the `Header`, since we should be able to deserialize that independently of the rest of the message; if that fails _too_, any opinions on what to do?

(We could certainly drop it, rather than replying at all, but if we want to report something to the host, we'd have to pick a message ID with nothing to go on)

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,414 @@
...
+                    return Err(Error::RequestTooLarge);
+                }
+                self.write(mem, modules, data)?;
+                Ok((HostResponse::Write(mem), 0))
+            }
+            HostRequest::SetPowerMode(mode) => {
+                let mask = get_mask(modules)?;
+                // TODO: do we need delays in between any of these operations?
+                match mode {
+                    PowerMode::Off => {
```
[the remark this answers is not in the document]
Aaron-Hartwig: The next step is for the FPGA to actually handle these transitions. See [RFD 244](https://rfd.shared.oxide.computer/rfd/0244#_fpga_module_sequencing) for details.

This is just a temporary sketch for Matt's testing.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,414 @@
...
+                            .map_err(|_e| Error::ReadFailed)?;
+                    }
+                    PowerMode::High => {
+                        // Power enabled, LpMode disabled
+                        self.transceivers
+                            .clear_lpmode(mask)
+                            .map_err(|_e| Error::ReadFailed)?;
+                        self.transceivers
+                            .set_power_enable(mask)
+                            .map_err(|_e| Error::ReadFailed)?;
```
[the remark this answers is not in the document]
Aaron-Hartwig: LpMode really only seems to matter when the module is both powered on and out of reset. Given that, changing it prior to those conditions seems fine.

Also see my comment above about the FPGA handling these details in the near future, not the SP. So while in the mode we are doing `set_power_enable` after `clear_lpmode`, the power should already be enabled as part of `PowerMode::Low` (aka A2).

drv/transceivers-server/src/net_api.rs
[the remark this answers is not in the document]
bnaecker: Agreed it's all pretty unlikely!

I also think that it makes sense to drop the packet if the `Header` can't be deserialized. The host will currently keep retrying until it gets a positive ACK from the SP, so dropping it on the SP-side does not affect that behavior. The alternative, such as sending a message ID of `u64::MAX`, seems no better. Even _if_ the host could match that up with its outstanding request, it would just retry it anyway.

drv/transceivers-server/src/net_api.rs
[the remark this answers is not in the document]
bnaecker: Thanks Aaron, that's helpful. Does that mean that the SP will not even issue multiple commands like this? That instead, there'd be a requested transition from say A2 -> A0, and the FPGA would set these pins?

drv/transceivers-server/src/net_api.rs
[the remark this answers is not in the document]
jgallagher: > Yeah, this branch is only taken when deserialization fails (which shouldn't happen, considering UDP has checksums, but 🤷🏻 ).

I've never seen a "packet was corrupt" deserialization failure, but in development I see deserialization failures _all the time_ when I've added or changed message types and only updated or rebuilt one side. Splitting the deserialization of the header handles all of those cases cleanly, since the header never changes.

FWIW in MGS I went with "reply with a bogus message ID" if deserializing the header fails; MGS can't actually do anything with that to pair it up with the request, so it'll eventually resend its request anyway, but again in development in particular it's nice to be able to tell the difference between "I'm not getting the response I expect, but I am seeing packets with bogus message IDs" from "the SP isn't responding".

drv/transceivers-server/src/net_api.rs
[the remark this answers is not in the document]
Aaron-Hartwig: Correct, it would issue a single "go to power state `X`" command. The intention here will be for the SP to request a module to get to a certain state and then the FPGA would handle all the signal-level details of what that means. If the FPGA were to fail to comply with the request (due to a power fault for example), it will report as such so when the SP realizes that the port is not in the state it requested it could go get details as to why that is the case.

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,414 @@
...
+                            data,
+                            &mut tx_data_buf[Message::MAX_SIZE..],
+                        ) {
+                            Ok(r) => r,
+                            Err(e) => (HostResponse::Error(e), 0),
+                        };
+                        let out = Message {
+                            header: msg.header,
+                            modules: msg.modules,
+                            body: MessageBody::HostResponse(reply),
```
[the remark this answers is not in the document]
mkeeter: Thanks!  I'm going to wait for https://github.com/oxidecomputer/transceiver-control/pull/11, then rebase onto both of them (heading over to review that one now)

drv/transceivers-server/src/net_api.rs
mkeeter: Done in the recent rebase!

drv/transceivers-server/src/net_api.rs
mkeeter: Yeah, I cleaned this up a little bit to make the `ModulesId` -> local types happen at the top of each function.

drv/transceivers-server/src/net_api.rs
mkeeter: Yup, fixed post-rebase

drv/transceivers-server/src/net_api.rs
```diff
@@ -0,0 +1,414 @@
...
+use transceiver_messages::{
+    message::*,
+    mgmt::{MemoryRegion, UpperPage},
+    Error, ModuleId,
+};
+use userlib::hl::sleep_for;
+
+/// Convert from the over-the-network type to our local port mask type
+fn get_mask(m: ModuleId) -> Result<FpgaPortMasks, Error> {
+    let fpga_ports: u16 = m.ports.0;
```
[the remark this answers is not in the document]
mkeeter: I'd rather keep the `transceiver-messages` types isolated to this one file.

(Otherwise, we'd need to add a dependency on `transceiver-messages` to the low-level `sidecar-front-io` crate, which is a bit of a leak across abstraction layers)

drv/transceivers-server/src/net_api.rs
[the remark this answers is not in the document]
mkeeter: I think this is now changed to match [the host behavior](https://github.com/oxidecomputer/transceiver-control/pull/13#discussion_r1024496259), but double-checking would be appreciated!

drv/transceivers-server/src/net_api.rs
mkeeter: Fixed to explicitly drop packets on `QueueFull` (still panicking on other errors, since they're not expected)

drv/transceivers-server/src/net_api.rs
```diff
@@ -80,68 +85,51 @@ impl ServerImpl {
...
-                            msg_len,
-                        );
-                        (msg_len + data_len) as u32
+                        self.handle_message(msg, data, tx_data_buf)
                     }
                     Err(e) => {
+                        // At this point, deserialization has failed, so we
+                        // can't handle the packet.  We'll attempt to
+                        // deserialize *just the header* (which should never
+                        // chnage), in the hopes of logging a more detailed
```
bnaecker, line 96: Nit: typo in `change`

drv/transceivers-server/src/net_api.rs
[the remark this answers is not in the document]
bnaecker: GH is so hard for me to navigate, that I can't actually find this comment thread relative to the source. It looks like this is handled by returning an error message body from `handle_host_request` and the response `out` in `net_api.rs:192` always uses the incoming message header. If that's true, then I agree it looks good!

drv/transceivers-server/src/net_api.rs
mkeeter: Added a comment to that effect

drv/transceivers-server/src/udp.rs
```diff
@@ -0,0 +1,503 @@
...
+            self.wait_and_check_i2c(mask)?;
+        }
+        Ok(())
+    }
+
+    fn wait_and_check_i2c(
+        &mut self,
+        mask: FpgaPortMasks,
+    ) -> Result<(), HwError> {
+        // TODO: use a better error type here
```
bnaecker, line 413: This todo now seems resolved?

bnaecker (approved):
💯  Looks great to me!

drv/transceivers-server/src/udp.rs
```diff
@@ -0,0 +1,503 @@
...
+                                    header.version
+                                ));
+                            }
+                            Err(e) => {
+                                ringbuf_entry!(Trace::DeserializeHeaderError(
+                                    e
+                                ));
+                            }
+                        }
+                        None
```
jgallagher, line 114: Should the SP send back a protocol error message here? Maybe only in the "`header` deserialized successfully" branch, because then we have the message ID to tell the host it's not going to get an answer.

drv/transceivers-server/src/udp.rs
```diff
@@ -0,0 +1,503 @@
...
+                    }
+                };
+
+                if let Some(out_len) = out_len {
+                    meta.size = out_len;
```
jgallagher, line 119: Idle thought not relevant to this PR - should `net.send_packet()` fill in `meta.size` on its own based on the buffer it gets as the third arg? Or assert that they match (if it doesn't already)?

drv/transceivers-server/src/udp.rs
```diff
@@ -0,0 +1,503 @@
...
+        &mut self,
+        msg: Message,
+        data: &[u8],
+        tx_data_buf: &mut [u8],
+    ) -> Option<u32> {
+        // If the version is mismatched, then we can't trust the deserialization
+        // (even though it nominally succeeded); don't reply.
+        if msg.header.version != 1 {
+            ringbuf_entry!(Trace::WrongVersion(msg.header.version));
+            return None;
```
jgallagher, line 161: Similar comment as above - should this send a protocol error message back to the host?

drv/transceivers-server/src/udp.rs
mkeeter: Yup, fixed!

jgallagher (approved):

drv/transceivers-server/src/udp.rs
mkeeter: We decided to silently drop cases where the version doesn't match, since it means that we can't trust much of anything.  Also, if we did reply, our serialized message _back_ would necessarily be encoded with **our** version, meaning the host would see a version mismatch and try to reply with a protocol error, etc.

PR #932, opened 2022-11-15
Switch to using workspace dependencies for _everything_

mkeeter:
This PR switches to [workspace dependency inheritance](https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace), which is a way to define dependency versions _once_ at the root of a workspace.

After this PR, every (remote) crate is declared as `{ workspace = true }`, without any version or additional features*.  Local crates are still declared as `{ path = "../, ... }` for now, because I find it useful to distinguish between remote and local crates at a glance; I could be persuaded otherwise.

Doing this minimizes build churn, because all Hubris crates use the same versions + features of dependencies.  In addition, it makes it easier to bump dependencies in a single place!

*the hardware PAC crates are an exception to the "no features in crate-level `Cargo.toml` rule, because their features aren't additive.

In addition, this PR
- Switches every Hubris crate to `edition = "2021"`, fixing minor breakages
- Enables `default-features = false` for all (remote) dependencies, then selectively adds features that are needed
    - Goblin is the one exception, because it requires all features to unlock `goblin::Object`
- Removes `default-features = false` for all path (within-Hubris) dependencies that don't _have_ any default features
- Removed explicit versions for stuff that's already tracked in `Cargo.lock`

This is the rare PR where reviewing the `Cargo.lock` changes is useful, since that's ground truth for actual changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -712,7 +712,7 @@ dependencies = [
  "lib-lpc55-usart",
  "lpc55-pac",
  "nb 1.0.0",
- "salty 0.2.0 (git+https://github.com/oxidecomputer/salty?rev=eb3c31858f631a7fb9934246c8efdef080d05726)",
+ "salty 0.2.0 (git+https://github.com/oxidecomputer/salty)",
  "serde",
  "serde-big-array",
  "sha3",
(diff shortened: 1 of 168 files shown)


jgallagher (approved):
Beautiful

cbiffle (approved):

PR #937, opened 2022-11-16
Ignition API refinements, improve presence tracking

arjenroodselaar:
Some additional refinement (naming of things, pub fields in structs) of the Ignition API after more work on Humility UI. Added arrival/departure hooks for Targets to allow for additional initialization when Targets are first marked present.


drv/ignition-api/src/lib.rs
```diff
@@ -68,29 +69,11 @@ bitfield! {
...
-        self.0.count_ones().try_into().unwrap()
-    }
-}
-
 bitfield! {
     #[derive(Copy, Clone, Debug, PartialEq, Eq, From, FromBytes, AsBytes)]
     #[repr(C)]
     pub struct Target(u64);
-    pub u8, into SystemType, system_type, _: 7, 0;
+    pub u8, into SystemId, id, _: 7, 0;
```
mkeeter, line 76: What's the difference between `SystemId` and `SystemType`?

drv/ignition-api/src/lib.rs
```diff
@@ -144,39 +189,58 @@ impl From<Request> for u8 {
...
+    PartialEq,
+    Eq,
+    From,
+    FromPrimitive,
+    ToPrimitive,
+    AsBytes,
 )]
 #[repr(u8)]
-pub enum LinkSelect {
+pub enum TransceiverSelect {
```
mkeeter, line 233: Can you add a comment explaining what this is?  I thought the main FPGA had a single link to each Ignition FPGA, but this is showing three separate links?

drv/ignition-api/src/lib.rs
```diff
@@ -144,39 +189,58 @@ impl From<Request> for u8 {
...
     Controller = 1,
     TargetLink0 = 2,
     TargetLink1 = 3,
 }
 
+impl TransceiverSelect {
+    pub const ALL: [Self; 3] =
+        [Self::Controller, Self::TargetLink0, Self::TargetLink1];
```
mkeeter, line 241: Should this be a bitfield instead?

(hard to say for certain without understanding what it does 😆)

drv/ignition-server/src/main.rs
```diff
@@ -112,10 +190,7 @@ impl idl::InOrderIgnitionImpl for ServerImpl {
         &mut self,
         _: &userlib::RecvMessage,
     ) -> Result<u64, RequestError> {
-        self.controller
```
mkeeter, line 115: Since this can never fail, you can use `err: ServerDeath` (instead of `err: CLike("IgnitionError")`) in the Idol API, and `Result<u64, RequestError<core::convert::Infallible>>` as the return type.

drv/ignition-server/src/main.rs
```diff
@@ -209,21 +284,23 @@ impl idl::InOrderIgnitionImpl for ServerImpl {
...
-        for i in 0..core::cmp::min(state.len(), self.port_count as usize) {
+        for i in 0..min(state.len(), self.port_count as usize) {
             // Check if the present bit is set in the summary.
             if summary & 0x1 != 0 {
                 state[i] =
                     self.port_state(i as u8).map_err(RequestError::from)?.0;
             }
             // Advance to the next port.
-            summary = summary >> 1;
+            summary >>= 1;
```
mkeeter, line 303: Same suggestion here: explicitly calculate `mask = 1 << i;`, then leave `summary` immutable.

drv/ignition-api/src/lib.rs
```diff
@@ -8,16 +8,16 @@
...
-// the) mainboard controller. This constant is used to conservatively allocate
-// an array type which can contain the port state for all ports. The actual
-// number of pors configured in the system can be learned through the
-// `port_count()` function below.
+// The `presence_summary` vector (see `ignition-server`) is implicitly capped at
+// 40 bits by (the RTL of the) mainboard controller. This constant is used to
+// conservatively allocate an array type which can contain the port state for
+// all ports. The actual number of pors configured in the system can be learned
+// through the `port_count()` function below.
 pub const PORT_MAX: usize = 40;
```
mkeeter, line 21: I'd vote for this to be a `u8`: in different places, we use it as both a `u8` and a `usize`; making the constant a `u8` is the safer choice.

drv/ignition-server/src/main.rs
```diff
@@ -78,20 +81,95 @@ struct ServerImpl {
...
+                    Err(e) => ringbuf_entry!(Trace::TargetError(port, e)),
+                }
+            }
+            // Advance to the next port.
+            ports >>= 1;
+            success >>= 1;
+        }
+
+        success
+    }
```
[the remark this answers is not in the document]
arjenroodselaar: Using a mask makes this more readable. Thanks.

drv/ignition-api/src/lib.rs
arjenroodselaar: Yes, this is the right thing to do. The places where it's used as `usize` are pretty minimal.

mkeeter:
One thing I just noticed: `enum IgnitionError` should specify that the first member is `= 1`; otherwise, RPC calls will mistake `ServerDied (0)` as "okay".

drv/ignition-server/src/main.rs
```diff
@@ -78,20 +81,95 @@ struct ServerImpl {
...
+    fn map_ports<F>(mut ports: u64, mut f: F) -> u64
+    where
+        F: FnMut(u8) -> Result<(), IgnitionError>,
+    {
+        let mut success = 0u64;
+
+        for port in 0..PORT_MAX as u8 {
+            if ports & 0x1 != 0 {
+                match f(port) {
+                    Ok(()) => success |= 1 << PORT_MAX,
```
Aaron-Hartwig, line 126: This sets the same bit each time, I think you mean:
```suggestion
                    Ok(()) => success |= 1 << port,
```

arjenroodselaar:
Abandoning this PR after discussion with @mkeeter, @cbiffle and @jgallagher about how to handle Hubris/Humility/MGS ABIs. The `ignition-api` crate is undergoing some change and I'll put up a new PR.

PR #939, opened 2022-11-16
host-sp-comms: give the host real MAC addresses from the FRU ID EEPROM

### LINKED ISSUE #930: host-sp-comms should report MAC addresses from FRU ID EEPROM
As of #807, the MAC addresses are already being read and used by the `net` task. We should add a way for `host-sp-comms` to ask `net` for the MAC addresses it should give to the host instead of returning fake data.

jgallagher:
I'm on the fence on where `NUM_HOST_MAC_ADDRESSES` lives; I could see an argument for putting it in `net` and changing the idol call to `get_host_mac_addresses()`. Happy to make that change if it seems cleaner.

Fixes #930.


task/host-sp-comms/src/main.rs
```diff
@@ -600,12 +618,23 @@ impl ServerImpl {
...
+                let response = if block.count.get() > 0 {
+                    let count =
+                        u16::min(block.count.get(), NUM_HOST_MAC_ADDRESSES);
+                    SpToHost::MacAddresses {
+                        base: block.base_mac,
+                        count,
+                        stride: block.stride,
+                    }
+                } else {
+                    SpToHost::MacAddresses {
```
mkeeter, line 631: Is it important to return zeros for `base` and `stride` here?  Otherwise, I'm not sure whether this block is needed: the host should notice `count: 0` either way (and freak out accordingly).

idl/net.idol
```diff
@@ -110,10 +110,13 @@ Interface(
...
             doc: "Reports the MAC address of port 0",
-            reply: Result(
-                ok: "MacAddress",
-                err: ServerDeath,
-            ),
+            reply: Simple("MacAddress"),
+            idempotent: true,
+        ),
+        "get_leftover_mac_addresses": (
+            doc: "Reports any remaining MAC addresses after the net task has claimed its own; e.g., leftover macs from local VPD for use by the host on a gimlet",
```
mkeeter, line 117: Same as below, I'm not sure this is Gimlet-specific!

task/host-sp-comms/src/main.rs
jgallagher: I think I'd classify this as "nice to have" and not "critical" - I agree the host _should_ notice `count: 0` either way, but it seems safer in some sense to give it all zeroes in this case: even if the host is buggy and assumes at least one address exists, that address will be nonsense.

idl/net.idol
jgallagher: Agreed; I'm happy to either remove the example from the doc comment or extend it ("for use by the host on a gimlet or the tofino on a sidecar" or somesuch). Do you have a preference between those?

task/host-sp-comms/src/main.rs
```diff
@@ -45,6 +47,20 @@ const A2_REBOOT_DELAY: u64 = 5_000;
 // response to send, and we haven't yet started to receive a request).
 const UART_ZERO_DELAY: u64 = 200;
 
+// How many MAC addresses should we report to the host? Per RFD 320, a gimlet
```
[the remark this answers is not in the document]
jgallagher: I think this is a solid argument for this constant being in the correct place, and _not_ moving it to net:

1. `host-sp-comms` only runs on gimlets, so it knows that a gimlet needs 3 MACs
2. Some other task that talks UDP (for now, based on our tentative plan around RFD 329, and i2c in the future) is responsible for deciding how many MAC addresses to give to the scrimlet
3. There is no task that calls net's `get_spare_mac_addresses()` on PSCs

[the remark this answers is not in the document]
mkeeter: Yup, I forgot that this is in `host-sp-comms`, rather than `control-plane-agent`

task/host-sp-comms/src/main.rs
mkeeter: Okay, seems reasonable

idl/net.idol
mkeeter: I'd vote to remove the example entirely

idl/net.idol
```diff
@@ -110,10 +110,13 @@ Interface(
...
         "get_mac_address": (
             doc: "Reports the MAC address of port 0",
-            reply: Result(
-                ok: "MacAddress",
-                err: ServerDeath,
-            ),
+            reply: Simple("MacAddress"),
+            idempotent: true,
+        ),
+        "get_leftover_mac_addresses": (
```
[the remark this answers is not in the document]
jgallagher: Renamed in ed0e851

idl/net.idol
jgallagher: Removed in ed0e851

mkeeter (approved):

PR #941, opened 2022-11-16
Allow MGS to query for VSC7448 port status 

jgallagher:
Adds handlers for a new "component details" message type from MGS. Currently the only meaningful component that can return details is the vsc7448 on a sidecar, but in the future we can slot in other components that have status (or sensor measurements, etc.) in this same handler.

This PR is smaller than it appears: the bulk of it is converting between the (nearly identical) monorail types and the corresponding `gateway-messages` types. It also includes some minor shuffling around of `monorail-api` (including adding generation of an idol client).

### DIFF
--- a/drv/monorail-api/Cargo.toml
+++ b/drv/monorail-api/Cargo.toml
@@ -1,15 +1,17 @@
 [package]
-name = "monorail-api"
+name = "drv-monorail-api"
 version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-derive-idol-err = { path = "../../lib/derive-idol-err"  }
+derive-idol-err = { path = "../../lib/derive-idol-err" }
 userlib = { path = "../../sys/userlib" }
 vsc7448 = { path = "../vsc7448" }
+vsc85xx = { path = "../vsc85xx" }
 
 num-traits = { workspace = true }
 serde = { workspace = true }
+ssmarshal = { workspace = true }
 zerocopy = { workspace = true }
 
 # This section is here to discourage RLS/rust-analyzer from doing test builds,
@@ -17,3 +19,6 @@ zerocopy = { workspace = true }
 [lib]
 test = false
 bench = false
+
+[build-dependencies]
+idol = {git = "https://github.com/oxidecomputer/idolatry.git"}
--- a/drv/monorail-api/Cargo.toml
+++ b/drv/monorail-api/Cargo.toml
@@ -21,4 +21,4 @@ test = false
 bench = false
 
 [build-dependencies]
-idol = {git = "https://github.com/oxidecomputer/idolatry.git"}
+idol = { workspace = true }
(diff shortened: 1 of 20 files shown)


drv/monorail-api/Cargo.toml
```diff
@@ -1,19 +1,24 @@
...
 zerocopy = { workspace = true }
 
 # This section is here to discourage RLS/rust-analyzer from doing test builds,
 # since test builds don't work for cross compilation.
 [lib]
 test = false
 bench = false
+
+[build-dependencies]
+idol = {git = "https://github.com/oxidecomputer/idolatry.git"}
```
mkeeter, line 24:
```suggestion
idol = { workspace = true }
```

task/control-plane-agent/src/inventory.rs
```diff
@@ -168,6 +168,15 @@ const OUR_DEVICES: &[DeviceDescription<'static>] = &[
...
         presence: DevicePresence::Present, // TODO: ok to assume always present?
     },
+    // If we're building for sidecar, we always claim to have a VSC7448.
+    #[cfg(feature = "sidecar")]
+    DeviceDescription {
+        component: SpComponent::VSC7448,
+        device: SpComponent::VSC7448.const_as_str(),
+        description: "Management network switch",
+        capabilities: DeviceCapabilities::HAS_MEASUREMENT_CHANNELS,
+        presence: DevicePresence::Present, // TODO: ok to assume always present?
```
mkeeter, line 178: Kinda – it takes ~4 seconds to boot, but also, while booting, we can't access the management network 😂

task/control-plane-agent/src/mgs_sidecar/monorail_port_status.rs
```diff
@@ -0,0 +1,171 @@
...
+    Ok(PortStatus {
+        port,
+        cfg: PortConfigConvert(port_status.cfg).into(),
+        link_status: LinkStatusConvert(port_status.link_up).into(),
+        phy_status: phy_status.map(|s| PhyStatusConvert(s).into()),
+        counters: PortCountersConvert(counters).into(),
+    })
+}
+
+struct PortConfigConvert(drv_monorail_api::PortConfig);
```
mkeeter, line 48: Vague question: is having conversion structs and implementing `Into` on them cleaner than free functions?

e.g.
```rust
fn convert_port_config(p: drv_monorail_api::PortConfig) -> PortConfig {
    PortConfig {
        mode: convert_port_mode(s.mode),
        dev: (convert_port_dev(s.dev.0), s.dev.1),
        serdes: (convert_port_serdes(s.serdes.0), s.serdes.1),
    }
}
```
(I've got a weak preference for free functions, unless there's a benefit I'm not seeing)

jgallagher: No benefit, just a weak preference in the other direction. :man_shrugging: Without orphan rules I'd obviously implement `From` directly, so this feels like "I'm just working around orphan rules to do these conversions"; using free functions seems slightly less obvious to me?

mkeeter: Okay, that seems fine

mkeeter (approved):

PR #942, opened 2022-11-17
Fix errors starting at 0

mkeeter:
Error values of 0 are mistaken for "no error" by Hiffy IPC.

### DIFF
--- a/task/net-api/src/lib.rs
+++ b/task/net-api/src/lib.rs
@@ -54,7 +54,7 @@ pub enum PhyError {
 #[repr(u32)]
 pub enum KszError {
     /// This functionality is not available on the given board
-    NotAvailable,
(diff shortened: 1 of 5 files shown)


task/net-api/src/lib.rs
```diff
@@ -54,7 +54,7 @@ pub enum PhyError {
 #[repr(u32)]
 pub enum KszError {
     /// This functionality is not available on the given board
-    NotAvailable,
+    NotAvailable = 1,
```
lzrd, line 57: Should you add this for documentation for each of the enums?:
enum Foo {
  _InterpretendAsOk = 0,
 ...

mkeeter: John added a check to the `IdolError` macro!

PR #946, opened 2022-11-17
lib/derive-idol-error: improve compile errors

cbiffle:
This now prints nice compile error messages highlighting any invalid discriminants, and also catches attempts to use non-C-style enums.

### DIFF
--- a/lib/derive-idol-err/Cargo.toml
+++ b/lib/derive-idol-err/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 syn = { workspace = true }
 quote = { workspace = true }
+proc-macro2 = {workspace = true}
 
 [lib]
 proc-macro = true
--- a/lib/derive-idol-err/src/lib.rs
+++ b/lib/derive-idol-err/src/lib.rs
@@ -24,29 +24,62 @@ pub fn derive(input: TokenStream) -> TokenStream {
     let data = match data {
         syn::Data::Enum(data) => data,
         syn::Data::Struct(_) | syn::Data::Union(_) => {
-            panic!("IdolError can only be derived on enums")
+            return compile_error(
+                ident.span(),
+                "IdolError can only be derived on enums",
+            )
+            .into();
         }
     };
 
-    // Assert that each variant is nonzero when cast to a `u32`; zero is
-    // reserved for success!
-    let variant_nonzero_assertions = data.variants.into_iter().map(|variant| {
-        let v = variant.ident;
-
-        // Inline version of
-        // ```
-        // static_assertions::const_assert_ne!(#ident::#v, 0)
-        // ```
-        quote! {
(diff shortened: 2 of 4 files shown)


lib/derive-idol-err/src/lib.rs
```diff
@@ -28,25 +28,62 @@ pub fn derive(input: TokenStream) -> TokenStream {
...
+                if value == 0 {
+                    variant_errors.push(compile_error(
+                        lit.span(),
+                        "error enums must not contain zero",
+                    ));
+                }
+                discriminant = Some(value);
+            } else {
+                variant_errors.push(
+                    compile_error(v.ident.span(), "idol errors must use simple positive integer discriminants"),
```
mkeeter, line 62: I'm surprised rustfmt didn't shorten this for you

lib/derive-idol-err/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 syn = { workspace = true }
 quote = { workspace = true }
+proc-macro2 = {workspace = true}
```
mkeeter, line 9: Nit: to match the other lines,
```suggestion
proc-macro2 = { workspace = true }
```

lib/derive-idol-err/src/lib.rs
```diff
@@ -28,25 +28,62 @@ pub fn derive(input: TokenStream) -> TokenStream {
         }
```
jgallagher, line 28: Can't see it in the diff, but one line above this is

```rust
panic!("IdolError can only be derived on enums")
```

Would that have a nicer error if returned as a `compile_error` like the new ones below?

jgallagher (approved):

lib/derive-idol-err/src/lib.rs
cbiffle: Not only did it not shorten this, it's forcing it onto one line. I've managed to get it to stop by explicitly breaking the string across lines.

lib/derive-idol-err/Cargo.toml
cbiffle: Is there a tool to maintain this? I've made the change in this commit, but I'm not excited about maintaining non-semantic whitespace through code review going forward.

lib/derive-idol-err/src/lib.rs
cbiffle: It would; I initially didn't do that because it seemed hard, but it was not hard. Fixed.

lib/derive-idol-err/Cargo.toml
mkeeter: I did some searching for an automatic TOML formatter a while ago, and came up empty.  It looks like IntelliJ has [something](https://github.com/intellij-rust/intellij-rust/pull/5536), but not sure whether it's usable as a standalone tool...

PR #948, opened 2022-11-19
control-plane-agent: Report PowerState from `sp_state`

andrewjstone:
Calls to return the SpState of a given SP now return the PowerState as well. This is to minimize calls from MGS when interacting with wicketd.

Individual `SpHandler::power_state` methods had the logic put into `power_state_impl` methods directly on the MgsHandler to prevent logging an extra `RingBuf` entry and the need to pass unused parameters.

This change depends upon
https://github.com/oxidecomputer/management-gateway-service/pull/15 and must be merged after that, once the cargo changes in here are also fixed.


jgallagher (approved):
LGTM; just needs the `Cargo.toml` change back to point to MGS's main branch once that PR is merged.

PR #949, opened 2022-11-19
Ignition API overhaul, ignition-server improvements

arjenroodselaar:
- Data types abstracted from hardware representation
- Add API documentation
- Track Target presence, update state on arrival and departure
- send_request(..) reports requests discarded by Target
- Improved batch operations


Cargo.toml
```diff
@@ -49,7 +49,7 @@ cortex-m-rt = { version = "0.6.12", default-features = false }
 cortex-m-semihosting = { version = "0.5.0", default-features = false }
 crc = { version = "3.0.0", default-features = false }
 ctrlc = { version = "3.1.5", default-features = false }
-derive_more = { version = "0.99", default-features = false, features = ["from"] }
+derive_more = { version = "0.99", default-features = false, features = ["from", "display"] }
```
mkeeter, line 52: Is `display` needed?

drv/ignition-api/build.rs
```diff
@@ -2,10 +2,22 @@
...
         "../../idl/ignition.idol",
         "client_stub.rs",
     )?;
+
+    let out_dir = build_util::out_dir();
+    let mut reg_map = fs::File::create(out_dir.join("ignition_controller.rs"))?;
+    write!(
+        &mut reg_map,
+        "{}",
+        fpga_regs(include_str!("ignition_controller.json"))?,
```
mkeeter, line 19: Our other FPGA images also commit the bitstream file and a README indicating which Buildomat build it came from.  Even though Hubris isn't responsible for flashing the Ignition target FPGA, I think this would be a good idea – if only so folks know where to find the image!

drv/ignition-server/src/main.rs
```diff
@@ -24,6 +24,13 @@ enum Trace {
     PortCount(u8),
     PresenceUpdate(u64),
     PresencePollError(IgnitionError),
+    TargetError(u8, IgnitionError),
+    TargetArrive(u8),
+    TargetDepart(u8),
+    SystemPowerRequest(u8, Request),
+    SystemPowerRequestError(u8, IgnitionError),
+    Foo,
```
mkeeter, line 32: `Foo` / `Bar` are probably debug leftovers?

drv/ignition-server/src/main.rs
```diff
@@ -78,19 +85,126 @@ struct ServerImpl {
...
+
+            ringbuf_entry!(Trace::PresenceUpdate(self.last_presence_summary));
+        }
+
+        Ok(())
+    }
+
+    /// Apply the given function to each port for which a bit in the `ports`
+    /// vector is set. Returns a bit vector with bits set for ports for which
+    /// the operation was succesfull. Under normal circumstances this output
```
mkeeter, line 128:
```suggestion
    /// the operation was succesful. Under normal circumstances this output
```

drv/ignition-server/src/main.rs
```diff
@@ -187,47 +301,55 @@ impl idl::InOrderIgnitionImpl for ServerImpl {
...
-            RequestError::from(IgnitionError::NoTargetPresent)
-        })?;
+        ringbuf_entry!(Trace::SystemPowerRequest(port, request));
 
-        if target_state.system_power_reset_in_progress()
-            || target_state.system_power_off_in_progress()
-            || target_state.system_power_on_in_progress()
-        {
-            return Err(RequestError::from(IgnitionError::RequestInProgress));
+        match self.target_request(port, request) {
```
mkeeter, line 306: Since we leave the `Ok` option unchanged, this could be something like
```rust
map_err(|e| {
    ringbuf_entry!(Trace::SystemPowerRequestError(port, e));
    Err(RequestError::from(e))
});
```

drv/ignition-api/build.rs
jgallagher, line 19: Tiny nit - is `include_str!()` necessary? It's probably fine, but I've had issues with compiler times / memory usage in the past with the `include_*!()` macros, and since this is a build script, we could probably do `std::fs::read_to_string()` instead.

drv/ignition-api/src/lib.rs
```diff
@@ -33,105 +36,395 @@ pub const PORT_MAX: usize = 40;
...
+        self.controller.presence_summary()
+    }
+
+    /// Return the raw state for the given port. The internals of `PortState`
+    /// are purposefully opague since they are closely tied to the RTL
+    /// implementation and as such are not intended for direct consumption. The
+    /// `port(..)` and `target(..)` functions below are probably what you want
+    /// to use.
+    #[inline]
+    pub fn raw_port_state(&self, port: u8) -> Result<PortState, IgnitionError> {
```
jgallagher, line 103: Similar to a comment Matt left elsewhere - if the details of `PortState` are opaque, does this method need to exist at all? When would a caller want this over `port()`?

drv/ignition-api/src/lib.rs
```diff
@@ -33,105 +36,395 @@ pub const PORT_MAX: usize = 40;
...
+        txr: TransceiverSelect,
+    ) -> Result<(), IgnitionError> {
+        self.controller.clear_link_events(port, txr)
+    }
+
+    /// Return the state of all ports in a single operation. Be aware that this
+    /// reply is fairly large and may require enlarging the stack of the caller.
+    pub fn all_ports(
+        &self,
+    ) -> Result<[Port; PORT_MAX as usize], IgnitionError> {
```
jgallagher, line 163: Thinking out loud - this (and `all_link_events()` below) always return a `PORT_MAX`-length array, which the caller must then shorten by knowing the actual port count. Would it make sense to return an iterator instead? If so, I think this could even return a concrete iterator type without having to write one by hand. Untested, but roughly:

```rust
pub fn all_ports(&self) -> Result<core::iter::Take<core::array::IntoIter<Port, PORT_MAX>>, IgnitionError> {
    let port_count = usize::from(self.port_count()?);
    let all_ports = self.controller.all_port_state()?.map(Port::from);
    Ok(all_ports.into_iter().take(port_count))
}
```

drv/ignition-api/src/lib.rs
```diff
@@ -33,105 +36,395 @@ pub const PORT_MAX: usize = 40;
...
+    Default,
+    PartialEq,
+    Eq,
+    FromPrimitive,
+    From,
+    FromBytes,
+    AsBytes,
+)]
+#[repr(C)]
+pub struct PortState(u64);
```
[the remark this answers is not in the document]
jgallagher: It may have to be public to be part of the idol interface?

drv/ignition-api/src/lib.rs
```diff
@@ -33,105 +36,395 @@ pub const PORT_MAX: usize = 40;
...
+        &self,
+    ) -> Result<[Port; PORT_MAX as usize], IgnitionError> {
+        Ok(self.controller.all_port_state()?.map(Port::from))
+    }
+
+    /// Return the `LinkEvents` for all ports and transceivers in a single
+    /// operation.
+    pub fn all_link_events(
+        &self,
+    ) -> Result<[[LinkEvents; 3]; PORT_MAX as usize], IgnitionError> {
```
jgallagher, line 171: `LinkEvents` doesn't have a field on it that indicates which transceiver it was for (and I think that's correct!); does that mean the caller has to infer the order of the returned `[LinkEvents; 3]` to tell which goes with which receiver? Would it make sense to switch this to a 3-field struct instead, something like

```rust
pub struct AllLinkEvents {
    pub controller: LinkEvents,
    pub target_link0: LinkEvents,
    pub target_link1: LinkEvents,
}
```

It's the same size/layout as `[LinkEvents; 3]` but maybe easier to understand.

drv/ignition-server/src/main.rs
```diff
@@ -78,19 +85,126 @@ struct ServerImpl {
...
+
+        // Determine if the request was accepted by matching the (updated)
+        // Target power state with the request.
+        match (request, self.target(port)?.power_state) {
+            (Request::SystemPowerOff, SystemPowerState::PoweringOff)
+            | (Request::SystemPowerOff, SystemPowerState::Off)
+            | (Request::SystemPowerOn, SystemPowerState::PoweringOn)
+            | (Request::SystemPowerOn, SystemPowerState::On)
+            | (Request::SystemPowerReset, SystemPowerState::PoweringOff)
+            | (Request::SystemPowerReset, SystemPowerState::PoweringOn) => {
```
jgallagher, line 203: I'm on the fence whether this would be an improvement, but patterns can be nested if it would be clearer to only have one arm per `Request` variant:

```rust
            (
                Request::SystemPowerOff,
                SystemPowerState::PoweringOff | SystemPowerState::Off,
            )
            | (
                Request::SystemPowerOn,
                SystemPowerState::PoweringOn | SystemPowerState::On,
            )
            | (
                Request::SystemPowerReset,
                SystemPowerState::PoweringOff | SystemPowerState::PoweringOn,
            ) => Ok(()),
```

jgallagher:
Thanks for these changes; they're nearly identical to how I'd modeled the MGS representation of this information based on the raw API and will make converting between the two very easy.

drv/ignition-api/build.rs
arjenroodselaar: This is from the same build which generates the pieces in `sidecar-mainboard-controller`. I'll copy over the `README` and put a sentence in to describe it. This did originally live in that crate, but there is a circular dependency between them if we want to keep it there. This felt like a reasonable compromise.

Cargo.toml
arjenroodselaar: It's not and I thought I deleted it, let me check what's up.

drv/ignition-api/src/lib.rs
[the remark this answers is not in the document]
arjenroodselaar: Yes, I do want to use it in the Idol interface (because I want the `-server` to serve up a New Type rather than just a u64 and enforce that in the `From` implementations). The compromise here is that while visible it's opaque to the rest of the system.

drv/ignition-server/src/main.rs
arjenroodselaar: Ah yes, thanks.

drv/ignition-api/src/lib.rs
arjenroodselaar: Well, `Port` initially did not had all the information found in `PortState`, so I wanted to use with https://github.com/oxidecomputer/hubris/blob/f065736a3502e07f072422af60db141308536e91/drv/ignition-api/src/lib.rs#L602 in an upcoming Hiffy function. But I can get that data from `Port` now, so there indeed is no need for this. I'll rework this.

drv/ignition-api/build.rs
mkeeter: Okay, that makes sense.

drv/ignition-api/src/lib.rs
arjenroodselaar: Hmm.. I like this a lot and should have thought of this since it avoids copying from one array to another as data goes from `PortState` to `Port`. It should reduce stack pressure in the caller and I may even double down on this in a follow up and allow requesting these in two or four batches instead of one large array. This would fit better with the model I need for Hiffy.

jgallagher: The version I wrote wouldn't reduce stack pressure on the caller, since the `all_ports` array gets moved into the iterator that's returned. But if you do some rework to use iterators or batched calls all the way down, it certainly could.

drv/ignition-api/src/lib.rs
arjenroodselaar: I came to the same conclusion and reworked this a bit:

`LinkEvents` are now `TransceiverEvents`, which is much closer to what they are (since these are observed by a transceiver). And from there the set of transceiver events observed by a port are now `LinkEvents`, with a name for each event set.

Apologies as I struggle through a consistent naming convention for the thing I built :)

drv/ignition-server/src/main.rs
arjenroodselaar: Sure, I had this in an earlier iteration. It ends up visually grouping these, which is may aid understanding. I don't feel strongly one way or another.

jgallagher: I don't either. Leaving it the way it is is fine with me.

mkeeter (approved):

jgallagher (approved):

PR #951, opened 2022-11-22
Use larger HIFFY_DATA for Sidecar A/B and Gimlet C

mkeeter:
This will speed up auxflash on Sidecar (pending https://github.com/oxidecomputer/humility/pull/265) and QSPI on Gimlet rev C.


arjenroodselaar (approved):
Shippit!

PR #952, opened 2022-11-22
Make pmbus_validate a function instead of a macro

jgallagher:
This allows us to size the amount of data we're reading based on the expected value, which fixes a failure to validate mwocp68 (mwocp68's expected value is 17 bytes long, one larger than the max size assumed by the macro).


drv/i2c-devices/src/tps546b24a.rs
```diff
@@ -64,8 +76,9 @@ impl Tps546B24A {
 
 impl Validate<Error> for Tps546B24A {
     fn validate(device: &I2cDevice) -> Result<bool, Error> {
-        let expected = [0x54, 0x49, 0x54, 0x6B, 0x24, 0x41];
-        pmbus_validate!(device, IC_DEVICE_ID, expected)
+        let expected = &[0x54, 0x49, 0x54, 0x6B, 0x24, 0x41];
```
mkeeter, line 79: This is _technically_ ASCII, but dunno if
```suggestion
        let expected = b"TITk$A";
```
is any clearer 😆

mkeeter (approved):

drv/i2c-devices/src/tps546b24a.rs
jgallagher: Lol yeah I checked that one too and opted to leave it :joy:

PR #956, opened 2022-11-23
Add input current and voltage sensors on MWOCP68

jgallagher:
This adds new sensor types (`InputCurrent`, `InputVoltage`), implements them for the mwocp68, and requires a corresponding humility change (https://github.com/oxidecomputer/humility/pull/266).

### DIFF
--- a/app/psc/rev-a.toml
+++ b/app/psc/rev-a.toml
@@ -298,8 +298,8 @@ name = "psu0mcu"
 address = 0b1011_000
 device = "mwocp68"
 description = "PSU 0 MCU"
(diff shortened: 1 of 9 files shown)


mkeeter:
Because this is a backwards-incompatible change, it should include a Hubris archive version bump; that way, Humility will give a clean error instead of cryptically failing to load the archive.

(and then we should do _another_ Humility version bump to add support for the new Hubris archive version, sorry for not calling this out earlier!)

jgallagher:
Good call. 1b1a5018 bumps us to v5, and https://github.com/oxidecomputer/hubris/pull/969 is the corresponding humility PR. Will wait to merge this until that can be pushed out everywhere.

mkeeter (approved):

PR #958, opened 2022-11-28
Slice RX buffer based on packet length

mkeeter:
This fixes an issue where we falsely return `WrongDataSize`, because we passed in the _entire_ rx buffer instead of just the packet.


bnaecker (approved):
Nice catch!

PR #963, opened 2022-11-29
QSFP: Fix i2c write buffer bug

Aaron-Hartwig:


bnaecker:
I've verified that this fixes https://github.com/oxidecomputer/transceiver-control/issues/27, so it seems good to me!

bnaecker (approved):
🤘

PR #965, opened 2022-11-29
[sprot] Try to make error handling more idiomatic

andrewjstone:
Overall, the behavior of the protocol should not have changed!

The overriding goal of this commit is to simplify error handling while
making it more idiomatic. The primary methodology used was "parse, don't
validate", meaning encoding knowledge about what was already discovered
in the type system, so that we don't have to redo checks about length,
or reparse the `MsgHeader`. The mechanisms to implement this methodology
are the two types: `RxMsg` and `TxMsg`, and the corresponding types
`VerifiedRxMsg` and `VerifiedTxMsg`, which ensure that a complete
message with a valid `MsgHeader` has been read into `rx_buf` or a valid
message  has been serialized into `tx_buf`, respectively.

The above changes rippled through the rest of the code, and allowed
removing some extra checks around message/buffer sizes. There were a
handful of other changes including:

 * Some function signatures changed to use the new types
 * `MsgError` was renamed to `SprotError`
 * `UpdateError` variants are now mapped 1:1 into `SprotError` variants.
While this adds a bunch of extra variants, it gives us more information
about what went wrong for debugging, and also allows us to expose a
single error type from Sprot.
 * Some extra ringbuf trace messages for debugging were added.
 * Some standalone functions related to buffer access were superceded
   by their `TxMsg`/`RxMsg` methods.
 * `ErrorRsp` messages now always include an underlying 1-byte code

### DIFF
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -39,6 +39,7 @@ pub fn new() -> Handler {
 #[derive(Copy, Clone, PartialEq)]
 enum Trace {
     None,
+    HeaderSizeMismatch(MsgType, u16, usize),
     Prev(usize, PrevMsg),
     ErrHeader(usize, PrevMsg, u8, u8, u8, u8),
     Overrun(usize),
@@ -59,11 +60,11 @@ impl Handler {
         &mut self,
         tx_prev: bool,
         iostat: IoStatus,
-        rx_buf: &[u8],
-        tx_buf: &mut [u8],
+        rx_buf: &RxMsg,
+        rx_bytes: usize,
+        tx_buf: &mut TxMsg,
         status: &mut Status, // for responses and updating
-    ) -> Option<usize> {
-        let tx_payload = payload_buf_mut(None, tx_buf);
+    ) -> Option<VerifiedTxMsg> {
         self.count = self.count.wrapping_add(1);
 
         // Before looking at the received message, check for explicit flush or
@@ -84,18 +85,22 @@ impl Handler {
                 // That includes the protocol identifier.
                 // If it is not an ignored protocol, then send an error.
                 if overrun {
-                    if !rx_buf.is_empty() {
-                        if Protocol::from(rx_buf[0]) != Protocol::Ignore {
+                    if rx_bytes != 0 {
+                        if Protocol::from(rx_buf.as_slice()[0])
+                            != Protocol::Ignore
+                        {
+                            let rx_buf = rx_buf.as_slice();
                             status.rx_overrun =
                                 status.rx_overrun.wrapping_add(1);
-                            tx_payload[0] = MsgError::FlowError as u8;
                             ringbuf_entry!(Trace::Prev(self.count, self.prev));
                             self.prev = PrevMsg::Overrun;
                             ringbuf_entry!(Trace::ErrHeader(
                                 self.count, self.prev, rx_buf[0], rx_buf[1],
                                 rx_buf[2], rx_buf[3]
                             ));
-                            return compose(MsgType::ErrorRsp, 1, tx_buf).ok();
+                            return Some(
+                                tx_buf.error_rsp(SprotError::FlowError),
+                            );
                         }
                     } else {
                         ringbuf_entry!(Trace::Prev(self.count, self.prev));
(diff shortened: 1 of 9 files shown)


andrewjstone:
Note: This PR requires more testing before merge. Specifically all the tests in the sprot readme

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
-    {
-        let end = MIN_MSG_SIZE + (header.len as usize);
-        if end > buffer.len() {
-            Err(MsgError::BadMessageLength)
-        } else if valid_bytes < end {
-            Ok(&mut buffer[valid_bytes..end])
+
+    pub fn validate_crc(&self, header: &MsgHeader) -> Result<(), SprotError> {
+        // The only way to get a `MsgHeader` is to call parse_header, which
+        // already validated the payload size.
```
lzrd, line 565: s/validated/sanity checked/

We don't know if that length is correct, only that it fits.
The CRC will validate the length (and the rest of the message).

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -116,217 +121,194 @@ impl Handler {
...
-        let (msgtype, rx_payload) = match parse(rx_buf) {
-            Ok((msgtype, payload)) => {
-                self.prev = PrevMsg::Good(msgtype);
-                (msgtype, payload)
+        let rxmsg = match rx_buf.parse_header(rx_bytes) {
+            Ok(header) => {
+                // We want to ensure the number of bytes received matches the header's
+                // expected payload size.
+                let expected_payload = rx_bytes - MIN_MSG_SIZE;
+                if header.payload_len as usize != rx_bytes - MIN_MSG_SIZE {
```
jgallagher, line 134:
```suggestion
                if header.payload_len as usize != expected_payload {
```

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -84,18 +85,22 @@ impl Handler {
                 // That includes the protocol identifier.
                 // If it is not an ignored protocol, then send an error.
                 if overrun {
-                    if !rx_buf.is_empty() {
-                        if Protocol::from(rx_buf[0]) != Protocol::Ignore {
+                    if rx_bytes != 0 {
+                        if Protocol::from(rx_buf.as_slice()[0])
+                            != Protocol::Ignore
+                        {
+                            let rx_buf = rx_buf.as_slice();
```
jgallagher, line 92: This might become clearer as I read other bits of the code, but: what is the length of this slice? In isolation I think I would expect it to be `rx_bytes`, but if that's the case why is `rx_bytes` a separate argument?

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
+const CRC16: Crc<u16> = Crc::<u16>::new(&CRC_16_XMODEM);
+pub const HEADER_SIZE: usize = <MsgHeader as SerializedSize>::MAX_SIZE;
+const PAYLOAD_CMD_SIZE: usize = 64; // Allow for struct accompanying block data.
+const PAYLOAD_BLOCK_SIZE: usize = 1024; // Allow for bulk data (arbitrary).
+const PAYLOAD_SIZE_MAX: usize = PAYLOAD_CMD_SIZE + PAYLOAD_BLOCK_SIZE;
+const CRC_SIZE: usize = <u16 as SerializedSize>::MAX_SIZE;
+pub const BUF_SIZE: usize = HEADER_SIZE + PAYLOAD_SIZE_MAX + CRC_SIZE;
+pub const MIN_MSG_SIZE: usize = HEADER_SIZE + CRC_SIZE;
+// XXX ROT FIFO size should be discovered.
+pub const ROT_FIFO_SIZE: usize = 8;
```
lzrd, line 657: TODO: Put ROT_FIFO_SIZE size in Status message or similar.

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -116,217 +121,194 @@ impl Handler {
...
-                | MsgType::UpdAbortUpdateRsp
-                | MsgType::UpdFinishImageUpdateRsp
-                | MsgType::UpdCurrentVersionRsp
-                | MsgType::Unknown => {
-                    status.rx_invalid = status.rx_invalid.wrapping_add(1);
-                    Err(MsgError::BadMessageType)
-                }
-            }
-        };
         // The above cases either enqueued a message and returned size
```
jgallagher, line 166: This comment is out of date now, I think. Now it's referring to what `self.run()` returns?

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -66,7 +70,7 @@ const TIMEOUT_WRITE_ONE_BLOCK: u32 = 2_000;
 const PART1_DELAY: u64 = 0;
```
lzrd, line 71: PART1_DELAY should be non-zero to work around issue 929 at the cost of some performance.

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -116,217 +121,194 @@ impl Handler {
...
+        | MsgType::SprocketsRsp
+        | MsgType::StatusRsp
+        | MsgType::UpdBlockSizeRsp
+        | MsgType::UpdPrepImageUpdateRsp
+        | MsgType::UpdWriteOneBlockRsp
+        | MsgType::UpdAbortUpdateRsp
+        | MsgType::UpdFinishImageUpdateRsp
+        | MsgType::UpdCurrentVersionRsp
+        | MsgType::Unknown => {
+            panic!("MsgType is not a request: {}", msgtype as u8)
```
jgallagher, line 311: Needing to duplicate all of these here and on lines 260-273 above is pretty unfortunate. Could we "inline" this function by having the match in `self.run()` give back a tuple of response type and length?

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
-#[repr(C)]
-pub enum UpdateRspKind {
-    Ok = 1,
-    Value = 2,
-    Error = 3,
-    Unknown = 0xff,
-}
+impl TxMsg {
+    pub fn new() -> TxMsg {
+        TxMsg { buf: [0; BUF_SIZE] }
```
jgallagher, line 435: This puts a relatively large buffer on the stack, I think? Large buffers in other tasks often end up with `&'static [u8; N]` references to move them to flash (where we have more tooling to keep track of size requirements than we do stack size); e.g., https://github.com/oxidecomputer/hubris/blob/master/task/net/src/buf.rs, https://github.com/oxidecomputer/hubris/blob/master/task/control-plane-agent/src/main.rs#L234-L240.

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -59,11 +60,11 @@ impl Handler {
         &mut self,
         tx_prev: bool,
         iostat: IoStatus,
-        rx_buf: &[u8],
-        tx_buf: &mut [u8],
+        rx_buf: &RxMsg,
+        rx_bytes: usize,
```
mkeeter, line 64: Dumb question: should `rx_bytes` live in the `RxMsg` (indicating what subset of the buffer is valid), or is it independent?

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -84,18 +85,22 @@ impl Handler {
                 // That includes the protocol identifier.
                 // If it is not an ignored protocol, then send an error.
                 if overrun {
-                    if !rx_buf.is_empty() {
-                        if Protocol::from(rx_buf[0]) != Protocol::Ignore {
+                    if rx_bytes != 0 {
+                        if Protocol::from(rx_buf.as_slice()[0])
```
mkeeter, line 89: These could be folded into a single conditional, I think!

(also, run `cargo xtask clippy` and see what it complains about; I think it would have caught this)

drv/stm32h7-sprot-server/src/main.rs
lzrd: ...but that should be carefully tested and in a separate PR.

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
-            _ => UpdateRspKind::Unknown,
+    pub fn as_slice(&self) -> &[u8] {
+        &self.buf[..]
+    }
+
+    pub fn as_mut(&mut self) -> &mut [u8] {
+        &mut self.buf[..]
+    }
+
+    pub fn payload_mut(&mut self) -> &mut [u8] {
```
jgallagher, line 446: These three methods feel a little questionable to me, but maybe they're fine. Just thinking out loud - `TxBuf` is sort of a higher-level thing to assemble messages (e.g., `error_rsp()`, `no_payload()`, and `from_lease()` all build a message without the caller having to deal with the details of the buffer), but these methods (and `from_existing()`) require the caller to have relatively intricate knowledge of how the message is assembled to use correctly (either to write to the correct place, or to know where to stop reading).

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -227,7 +237,7 @@ impl IO {
     /// Returns false on spurious interrupt.
     fn pio(&mut self, transmit: bool) {
         ringbuf_entry!(Trace::Pio(transmit));
-        let tx_end = self.tx_buf.len(); // Available bytes and trailing zeros
+        let tx_end = self.tx_buf.as_slice().len(); // Available bytes and trailing zeros
```
mkeeter, line 240: Why does `RxBuf` have a `.len()` function, but `TxBuf` does not?

drv/sprot-api/src/lib.rs
```diff
@@ -83,49 +82,138 @@ pub enum MsgError {
...
+            UpdateError::RunningImage => SprotError::UpdateRunningImage,
+            UpdateError::FlashError => SprotError::UpdateFlashError,
+            UpdateError::SpRotError => SprotError::UpdateSpRotError,
+            UpdateError::Unknown => SprotError::UpdateUnknown,
+        }
+    }
+}
+
+impl From<u8> for SprotError {
+    fn from(byte: u8) -> SprotError {
```
mkeeter, line 151: Because we derived `FromPrimitive`, this can simply be
```rust
    fn from(byte: u8) -> SprotError {
        Self::from_u8(byte).unwrap_or(SprotError::Unknown)
    }
```

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
-}
 
-impl From<UpdateRspKind> for u8 {
-    fn from(kind: UpdateRspKind) -> Self {
-        match kind {
-            UpdateRspKind::Ok => UpdateRspKind::Ok as u8,
-            UpdateRspKind::Value => UpdateRspKind::Value as u8,
-            UpdateRspKind::Error => UpdateRspKind::Error as u8,
-            _ => UpdateRspKind::Unknown as u8,
+    /// Serialize a request from a buffer with an already written payload
```
mkeeter, line 485: Clarify that the already-written payload is in **our** buffer, i.e. `self.buf`

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
-                    kind: UpdateRspKind::Value,
-                    value,
-                },
-            },
-            Some(error) => Self {
-                kind: UpdateRspKind::Error,
-                value: error,
-            },
-        }
+    fn write_crc(&mut self, payload_size: usize) -> VerifiedTxMsg {
```
mkeeter, line 505: I'd be tempted to merge this with `write_header`; otherwise, you're not sure that the `payload_size` argument necessarily matches what was written to the header.

(of course, we're doing it correctly at the moment, but it seems like a cleaner API to do them simultaneously)

mkeeter: At that point, I'd also want to rename it something like `finalize` and _maybe_ have it consume `self` by value?

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -250,7 +260,7 @@ impl IO {
             if self.txcount >= tx_end || !self.spi.can_tx() {
                 break;
             }
-            let b = self.tx_buf[self.txcount];
+            let b = self.tx_buf.as_slice()[self.txcount];
```
mkeeter, line 263: Would it be worth it to implement
```
impl<T, I: SliceIndex> Index/IndexMut for TxBuf
```
?

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
+    pub fn as_slice(&self) -> &[u8] {
+        &self.buf[..]
+    }
+
+    pub fn as_mut(&mut self) -> &mut [u8] {
+        &mut self.buf[..]
+    }
+
+    pub fn payload_mut(&mut self) -> &mut [u8] {
+        &mut self.buf[HEADER_SIZE..BUF_SIZE - CRC_SIZE]
```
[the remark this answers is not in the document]
mkeeter: I think it's correct as written, since `BUF_SIZE` is the full buffer size (not just the payload size) – but agree that testing would be good!

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
-const PAYLOAD_SIZE_MAX: usize = PAYLOAD_CMD_SIZE + PAYLOAD_BLOCK_SIZE;
-const CRC_SIZE: usize = <u16 as SerializedSize>::MAX_SIZE;
-pub const BUF_SIZE: usize = HEADER_SIZE + PAYLOAD_SIZE_MAX + CRC_SIZE;
-pub const MIN_MSG_SIZE: usize = HEADER_SIZE + CRC_SIZE;
-// XXX ROT FIFO size should be discovered.
-pub const ROT_FIFO_SIZE: usize = 8;
+/// A type indicating that a complete message has been successfully serialized and therefore
+/// fits in the allocated buffer.
+#[derive(Clone, Copy)]
+pub struct VerifiedTxMsg(pub usize);
```
mkeeter, line 519: Aha, this doesn't contain the data, just its size.

Would we be struck down by the Lifetime Gods if this _also_ borrowed the `&mut TxMsg`, to make sure no one else could mess with it?

drv/sprot-api/src/lib.rs
```diff
@@ -333,232 +422,238 @@ impl From<u8> for MsgType {
...
-        let end = crc_begin + CRC_SIZE;
-        if let Some(crc_buf) = buffer.get_mut(crc_begin..end) {
-            if let Ok(_n) = hubpack::serialize(crc_buf, &crc) {
-                return Ok(end);
-            } else {
-                return Err(MsgError::Serialization);
-            }
+/// A type indicating that the message header has been parsed and the CRC has
+/// been successfully verified
+pub struct VerifiedRxMsg(pub MsgHeader);
```
mkeeter, line 603: Same question here: should this borrow `&'a RxMsg` to ensure that no one else can modify it?

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -124,12 +129,24 @@ cfg_if::cfg_if! {
...
+    ($expected:expr, $actual:expr) => {
+        if $expected != $actual {
+            ringbuf_entry!(Trace::WrongMsgType($actual));
+            Err(SprotError::BadMessageType)
+        } else {
+            Ok(())
+        }
+    };
+}
+
```
mkeeter, line 143: Not sure why this is a macro, e.g.
```suggestion
fn expect_msg(expected: MsgType, actual: MsgType) -> Result<(), SprotError> {
    if expected != actual {
        ringbuf_entry!(Trace::WrongMsgType(actual));
        Err(SprotError::BadMessageType)
    } else {
        Ok(())
    }
}
```

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -124,12 +129,24 @@ cfg_if::cfg_if! {
...
 pub struct ServerImpl {
     sys: sys_api::Sys,
     spi: drv_spi_api::SpiDevice,
     // Use separate buffers so that retries can be generic.
-    pub tx_buf: [u8; BUF_SIZE],
-    pub rx_buf: [u8; BUF_SIZE],
+    pub tx_buf: TxMsg,
```
mkeeter, line 148: Nit: maybe name these `tx/rx_msg`, since they're not bare buffers?

(same thing below)

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -275,164 +288,147 @@ impl ServerImpl {
...
-        // be longer than ROT_FIFO_SIZE and get ROT_FIFO_SIZE
-        // instead of MIN_MSG_SIZE.
-        //
-        // TODO: Use DMA on RoT to avoid this dance.
+        // Fill in rx_buf with a complete message and validate its crc
+        let res = self.do_read_response();
+
+        // We must release the SPI bus before we return
+        ringbuf_entry!(Trace::CSnDeassert);
+        self.spi.release().map_err(|_| SprotError::SpiServerError)?;
```
mkeeter, line 309: Not required for this PR, but I sure would like to see an RAII-ish handle for the SPI bus...

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -275,164 +288,147 @@ impl ServerImpl {
...
-                    match msgtype {
-                        MsgType::ErrorRsp if payload_len > 0 => {
-                            let payload = payload_buf(
-                                Some(payload_len),
-                                &self.rx_buf[..],
-                            );
-                            errcode =
-                                MsgError::from_u8(payload[0]).unwrap_lite();
-                            ringbuf_entry!(Trace::MsgError(errcode));
+                Ok(rxmsg @ VerifiedRxMsg(header)) => {
```
mkeeter, line 401: cursed syntax, but okay

mkeeter: Actually, is there a reason to explicitly unpack the header, versus using `rxmsg.header` in the code below?

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -522,59 +497,31 @@ impl ServerImpl {
...
-        if msgtype == rsp {
-            let buf = payload_buf(Some(payload_len), &self.rx_buf[..]);
-            let (rsp, _) = hubpack::deserialize::<UpdateRspHeader>(buf)
-                .map_err(|_| {
-                    idol_runtime::RequestError::Runtime(UpdateError::SpRotError)
-                })?;
+    ) -> Result<Option<u32>, SprotError> {
+        let txmsg = self.tx_buf.from_existing(req, payload_len)?;
+        ringbuf_entry!(Trace::TxSize(txmsg.0));
+        let rxmsg @ VerifiedRxMsg(header) =
```
mkeeter, line 503: Same question here: could we dodge the cursed syntax and use `rxmsg.header`?

idl/sprot.idol
```diff
@@ -89,7 +89,7 @@ Interface(
             },
             reply : Result(
                 ok: "()",
-                err: CLike("UpdateError"),    // MsgError and UpdateError
+                err: CLike("SprotError"),    // SprotError and SprotError
```
mkeeter, line 92: "SprotError and SprotError"?

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -273,10 +283,12 @@ impl IO {
...
         let mut inframe = false;
         'outer: loop {
+            ringbuf_entry!(Trace::PioOuterLoop);
             // restart here on the one expected spurious interrupt.
-            sys_irq_control(1, true);
-            sys_recv_closed(&mut [], 1, TaskId::KERNEL).unwrap_lite();
+            sys_irq_control(SPI_IRQ, true);
+            sys_recv_closed(&mut [], SPI_IRQ, TaskId::KERNEL).unwrap_lite();
             loop {
+                ringbuf_entry!(Trace::PioInnerLoop);
```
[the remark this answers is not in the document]
andrewjstone: done

drv/sprot-api/src/lib.rs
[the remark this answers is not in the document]
andrewjstone: Yeah, I'm pretty sure it's correct as is.

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: It's not a dumb question at all. I very much considered doing this, but it felt wrong, since RxMsg owns the buffer and would require a reset on the next transaction. Right now all the "size" information lives outside this type. 

I will think more about it as I go through the rest of the comments and refactor.

drv/lpc55-sprot-server/src/main.rs
andrewjstone: Because I needed to get the length of the inner buffer for `RxBuf` :) These types definitely aren't symmetrical, but maybe they should be.

drv/sprot-api/src/lib.rs
andrewjstone: Excellent!

drv/sprot-api/src/lib.rs
andrewjstone: So... yeah. That's also a strategy I considered. It doesn't work here because the buffer is owned by `TxMsg`. However, in light of other parts of this review, I'm questioning if `TxMsg` should be created to wrap an `&mut [u8]` and consumed for each write. That seems like the better option, and I had a version of that that I replaced that wasn't as full featured.

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: It's the size of the entire buffer... Along with other comments, I'm considering that `RxMsg` should instead wrap the slice and keep track of the size read rather than owning the array on the stack. That would likely make some of the calling code a bit safer.

drv/sprot-api/src/lib.rs
andrewjstone: The buffer was already on the stack. I just wrapped it in this type. Presumably this is necessary for performance in the hot path. @lzrd any thoughts here?

drv/sprot-api/src/lib.rs
andrewjstone: I totally agree @jgallagher. This was sort of a compromise with the code as it already existed. However, I think I can tighten this up some more.

drv/lpc55-sprot-server/src/main.rs
andrewjstone: Maybe? I'm actually thinking of changing how these types work a bit... Will keep it in mind.

drv/sprot-api/src/lib.rs
andrewjstone: Most likely... yes. I don't think the current structure of the code lends itself well to that.  I'm not super concerned about that problem, and future improvements may make this clearer.

drv/sprot-api/src/lib.rs
andrewjstone: As above, there's something to this, but I'm worried about lifetime issues. I think some restructuring with an `RxMsg` and `TxMsg` that get consumed to create these values would make this clearer, even if references to the buffers weren't held.... We'd make sure that any places where the buffer gets mutated can't be called unless it has a `TxMsg` or `RxMsg` already, which it won't if it was consumed and the code moved on.

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: It's a historical arfifact!

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: They might go back to being bare buffers :)

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: Totally agreed. Stoltz and I already discussed that before I let this PR fly. It looks like Cliff anticipated this with `pub fn lock_auto(&self, device_index: u8, assert_cs) -> Result<ControllerLock<'_>, SpiError> {}`. I just didn't notice it because I didn't look at the spi code at all.

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: It's confusing. I'll do what you suggest.

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: yup

idl/sprot.idol
andrewjstone: lol, nice automated replacement

drv/sprot-api/src/lib.rs
```diff
@@ -3,6 +3,15 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 //! Client API for SP to RoT messages over SPI.
+//!
+//! An SP/RoT SPI message is:
+//!   1. A hubpack encoded `MsgHeader` containing the protocol, payload length,
+//!      and message type.
+//!   2. Payload according to MessageType, typically hubpack encoded
+//!      structure(s) and/or bulk data.
+//!   3. A CRC16 parameters that covers all of the bytes from the protocol ID to
```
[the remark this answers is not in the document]
andrewjstone: Fixed in [f86812d](https://github.com/oxidecomputer/hubris/pull/965/commits/f86812d56e41435275dab8b4cf32982f88d4c35f)

drv/sprot-api/src/lib.rs
andrewjstone: Fixed in [f86812d](https://github.com/oxidecomputer/hubris/pull/965/commits/f86812d56e41435275dab8b4cf32982f88d4c35f)

drv/sprot-api/src/lib.rs
andrewjstone: Fixed!

drv/sprot-api/src/lib.rs
andrewjstone: Fixed in [f86812d](https://github.com/oxidecomputer/hubris/pull/965/commits/f86812d56e41435275dab8b4cf32982f88d4c35f)

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: Fixed in [f86812d](https://github.com/oxidecomputer/hubris/pull/965/commits/f86812d56e41435275dab8b4cf32982f88d4c35f)

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: Comment removed in [f86812d](https://github.com/oxidecomputer/hubris/pull/965/commits/f86812d56e41435275dab8b4cf32982f88d4c35f)

idl/sprot.idol
andrewjstone: Fixed in [3a0f61a](https://github.com/oxidecomputer/hubris/pull/965/commits/3a0f61ac0031aaed46524561bf56eb41f20f085f)

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: Fixed in [3a0f61a](https://github.com/oxidecomputer/hubris/pull/965/commits/3a0f61ac0031aaed46524561bf56eb41f20f085f)

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: Fixed in [3a0f61a](https://github.com/oxidecomputer/hubris/pull/965/commits/3a0f61ac0031aaed46524561bf56eb41f20f085f)

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: Fixed in [3a0f61a](https://github.com/oxidecomputer/hubris/pull/965/commits/3a0f61ac0031aaed46524561bf56eb41f20f085f)

lzrd:
Go ahead and commit this. I will fix my bug re sink with small payloads.

andrewjstone:
There's a bug (unrelated to this change) that @lzrd is going to fix on top of this. I'm going to merge this in for now, and go back with a separate commit to resolve the more fuzzy comments relating to structure of RxMsg/TxMsg etc...

PR #966, opened 2022-11-29
Add CMPA and CFPA to build archive

labbott:


build/xtask/src/dist.rs
```diff
@@ -682,6 +682,17 @@ fn build_archive(cfg: &PackageConfig, image_name: &str) -> Result<()> {
     archive
         .copy(chip_dir.join("openocd.gdb"), debug_dir.join("openocd.gdb"))?;
 
+    if cfg.img_file("CMPA.bin", image_name).exists() {
```
kc8apf, line 685: Probably should test for CMPA and CFPA separately.  I realize they _should_ either _both_ be present or absent but accidents happen.  Right now this will only fail if CMPA is present and CFPA is missing.

kc8apf (approved):

PR #970, opened 2022-11-30
Add separate monitoring for PHY link flapping

mkeeter:
In some cases, the VSC7448 link is going to a PHY, so link status is less meaningful.  This is mostly relevant for Sidecar ports 40-43, which use a VSC8504 as a QSGMII → 4x SGMII converter before going off-board; the link between the VSC7448 and VSC8504 should _always be up_, because they're on the same board.

This PR adds a separate `phy_link_down_sticky: bool` to the `struct PortCounters`, which returns the **PHY's** opinion on whether the link has gone done.  We want this to be stateless (i.e. explicitly cleared by `reset_port_counters`), but the PHY registers are self-clearing; to work around this, we add a separate cache in the Monorail `ServerImpl`.

(this cache will be lost if the `monorail` task restarts, but it would reconfigure the PHYs anyways in that case, so that seems fine)

### DIFF
--- a/drv/vsc7448/src/lib.rs
+++ b/drv/vsc7448/src/lib.rs
@@ -25,6 +25,9 @@ pub use vsc_err::VscError;
 
 use crate::dev::Dev10g;
 
+/// Maximum port count
+pub const PORT_COUNT: usize = 53;
+
 /// This trait abstracts over various ways of talking to a VSC7448.
 pub trait Vsc7448Rw {
     /// Writes to a VSC7448 register.  Depending on the underlying transit
--- a/task/monorail-server/src/server.rs
+++ b/task/monorail-server/src/server.rs
@@ -11,7 +11,7 @@ use idol_runtime::{NotificationHandler, RequestError};
 use userlib::{sys_get_timer, sys_set_timer};
 use vsc7448::{
     config::{PortMap, PortMode},
-    DevGeneric, Vsc7448, Vsc7448Rw,
+    DevGeneric, Vsc7448, Vsc7448Rw, PORT_COUNT,
 };
 use vsc7448_pac::{types::PhyRegisterAddress, *};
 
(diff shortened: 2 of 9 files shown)


drv/vsc7448/src/lib.rs
```diff
@@ -25,6 +25,9 @@ pub use vsc_err::VscError;
 
 use crate::dev::Dev10g;
 
+/// Maximum port count
+pub const PORT_COUNT: usize = 53;
```
jgallagher, line 29: I also have this constant [in `control-plane-agent`](https://github.com/oxidecomputer/hubris/blob/master/task/control-plane-agent/src/mgs_sidecar.rs#L33-L34); could we move it to `monorail-api`? (Or maybe control-plane-agent could depend on this crate to grab it?)

task/monorail-server/src/server.rs
```diff
@@ -224,7 +232,32 @@ impl<'a, R: Vsc7448Rw> idl::InOrderMonorailImpl for ServerImpl<'a, R> {
...
+                            // If there is no PHY present, then the PHY link
+                            // down indication is always false.
+                            None => false,
+                            // Otherwise, bit 13 is "Link state change mask"
+                            Some(r) => r
+                                .map(|r| r.0 & (1 << 13) != 0)
+                                .map_err(MonorailError::from)?,
+                        };
+                        if v {
+                            self.phy_link_down_sticky[port as usize] = true;
```
jgallagher, line 256: I'd probably have written this as

```rust
self.phy_link_down_sticky[port as usize] |= v;
```

I think it's slightly clearer and avoids the conditional (although maybe the compiler would rewrite it to avoid the condition anyway).

jgallagher (approved):

task/monorail-server/src/server.rs
mkeeter: Yup, I refactored this to always check the PHY, which is more obviously correct.

drv/vsc7448/src/lib.rs
mkeeter: Done!

PR #971, opened 2022-11-30
gimlet_seq + thermal ringbuf cleanup

bcantrill:
Draft comment:

```
This very modestly cleans up gimlet_seq by making its A0 and A1 state
machines symbolic, as well as logging a bit more in the ring buffer on
a state change in an attempt to help debug some transient sequencing
issues (e.g.  https://github.com/oxidecomputer/quartz/issues/57).
This also very modestly cleans up the ring buffering in the thermal
task to always include the sensor ID rather than an internal index:
when the thermal loop fails due to a device issue, the sensor ID is
important for determining what device is failing.
```


drv/gimlet-seq-server/src/seq_spi.rs
```diff
@@ -21,6 +21,51 @@ pub enum Cmd {
...
+    GROUPC_PG = 0x07,
+    DELAY_1MS = 0x08,
+    ASSERT_PG = 0x09,
+    WAIT_PWROK = 0x0a,
+    WAIT_RESET_L = 0x0b,
+    DONE = 0x0c,
+    SAFE_DISABLE = 0x0d,
+}
+
+impl From<A0StateMachine> for u8 {
```
mkeeter, line 45: Optional: if you add a dependency on `derive_more`, you could `derive(From)` and skip this implementation.

mkeeter: hmmm, maybe not (see [this comment](https://jeltef.github.io/derive_more/derive_more/into.html#enums))

bcantrill: Oh, love that! Thank you!

PR #974, opened 2022-11-30
Restore Clippy cleanliness

mkeeter:
The compiler bump brought along a bunch of new Clippy lints!

I've gone through and fixed everything for Gimlet / Sidecar / PSC.

### DIFF
--- a/drv/sprot-api/src/lib.rs
+++ b/drv/sprot-api/src/lib.rs
@@ -161,13 +161,13 @@ impl From<hubpack::Error> for SprotError {
 
 // Return true if the error is recoverable, otherwise return false
 pub fn is_recoverable_error(err: SprotError) -> bool {
-    match err {
+    matches!(
+        err,
         SprotError::InvalidCrc
-        | SprotError::EmptyMessage
-        | SprotError::RotNotReady
-        | SprotError::RotBusy => true,
-        _ => false,
-    }
+            | SprotError::EmptyMessage
+            | SprotError::RotNotReady
+            | SprotError::RotBusy
+    )
 }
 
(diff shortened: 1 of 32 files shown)


task/thermal/src/control.rs
```diff
@@ -339,17 +339,11 @@ impl OneSidedPidState {
         } else {
             (-out_pd, output_limit - out_pd)
         };
-        self.integral = self.integral.max(integral_min).min(integral_max);
+        self.integral = self.integral.clamp(integral_min, integral_max);
```
jgallagher, line 342: TIL about `.clamp()`!

andrewjstone (approved):
Thanks for fixing this and cleaning up my mess Matt!

drv/sprot-api/src/lib.rs
```diff
@@ -481,23 +489,27 @@ pub struct RxMsg {
...
+
 impl RxMsg {
     pub fn new() -> RxMsg {
         RxMsg { buf: [0; BUF_SIZE] }
     }
 
-    pub fn len(&self) -> usize {
-        self.buf.len()
-    }
-
```
jgallagher, line 492: I don't think `len()` is provided by `AsMut<_>` is it?

jgallagher: Nevermind, you fixed this already

mkeeter: No, I'm removing this entirely to fix a warning about the lack of `is_empty`.

(you can still get the length with `as_slice().len()`, which is what we do in `TxBuf` already)

PR #976, opened 2022-12-01
Fix test code, minor efficiency change, minor refactor.

lzrd:
RoT: Fix array out of bounds in test code (SpRot.rot_sink).

SP: Follow @ajs's lead and factor out do_send_request() similar to do_read_response() to simplify chip select signal management.

SP: In do_read_response(), always read one full RoT FIFO of data to start which may exceed the minimum message size by 2-bytes but saves a second read for any respose read from the RoT that has only a one or two byte payload.

SP: To support possibly reading extra bytes, make the calculation of the size of the second read more careful.

SP and RoT: Replace multiple `core::mem::size_of::<u16>()` with a local const U16_SIZE.

### DIFF
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -252,8 +252,12 @@ impl Handler {
             MsgType::SinkReq => {
                 // The first two bytes of a SinkReq payload are the U16
                 // mod 2^16 sequence number.
-                tx_payload[0..2].copy_from_slice(&rx_payload[0..2]);
-                2
+                if rx_payload.len() >= core::mem::size_of::<u16>() {
+                    tx_payload[0..2].copy_from_slice(&rx_payload[0..core::mem::size_of::<u16>()]);
+                    core::mem::size_of::<u16>()
+                } else {
+                    0
+                }
             }
             // All of the unexpected messages
             MsgType::Invalid
--- a/drv/stm32h7-sprot-server/src/main.rs
+++ b/drv/stm32h7-sprot-server/src/main.rs
@@ -176,8 +176,6 @@ impl ServerImpl {
         txmsg: VerifiedTxMsg,
         timeout: u32,
     ) -> Result<VerifiedRxMsg, SprotError> {
-        let size = txmsg.0;
-        ringbuf_entry!(Trace::SendRecv(size));
         // Polling and timeout configuration
         // TODO: Use EXTI interrupt and just a timeout, no polling.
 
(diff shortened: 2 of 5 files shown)


drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -252,8 +252,12 @@ impl Handler {
             MsgType::SinkReq => {
                 // The first two bytes of a SinkReq payload are the U16
                 // mod 2^16 sequence number.
-                tx_payload[0..2].copy_from_slice(&rx_payload[0..2]);
-                2
+                if rx_payload.len() >= core::mem::size_of::<u16>() {
+                    tx_payload[0..2].copy_from_slice(&rx_payload[0..core::mem::size_of::<u16>()]);
```
andrewjstone, line 256: Just a note: You don't actually need the leading `0` for array indexing, but there's no problem leaving them in.

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -309,8 +265,48 @@ impl ServerImpl {
...
+        let buf = &self.tx_buf.as_slice()[..size];
+
+        // In order to improve reliability, start by sending only the
+        // first ROT_FIFO_SIZE bytes and then delaying a short time.
+        // If the RoT is ready, those first bytes will always fit
+        // in the RoT receive FIFO. Eventually, the RoT FW will respond
+        // to the interrupt and enter a tight loop to receive.
+        // The short delay should cover most of the lag in RoT interrupt
+        // handling.
+        let part1 = if let Some(part1) = buf.get(0..ROT_FIFO_SIZE.min(size)) {
```
andrewjstone, line 279: I know this is just moved code, but I believe you can get rid of the `if let` here and use indexing directly. You already know size fits in `buf` and you know `ROT_FIFO_SIZE` also fits in `buf`.

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -309,8 +265,48 @@ impl ServerImpl {
...
+            part1
+        } else {
+            return Err(SprotError::BadMessageLength);
+        };
+        let part2 = buf.get(part1.len()..).unwrap_lite(); // empty or not
+        ringbuf_entry!(Trace::TxPart1(part1.len()));
+        ringbuf_entry!(Trace::TxPart2(part2.len()));
+        ringbuf_entry!(Trace::CSnAssert);
+        self.spi
```
andrewjstone, line 288: We can simplify this to remove the 3 `map_err` calls by implementing `From<SpiError> for SprotError`

drv/stm32h7-sprot-server/src/main.rs
lzrd: done

drv/lpc55-sprot-server/src/handler.rs
lzrd: done

drv/stm32h7-sprot-server/src/main.rs
lzrd: done

PR #979, opened 2022-12-02
Tofino sequencer improvements

arjenroodselaar:
The Tofino sequencer broke after we changed the output ramp rate for some of the regulators. While debugging this it became painfully obvious that the previous version of the sequencer was rather aggressive in keeping rails disabled, discarding fault information in the process. This diff includes a new sequencer which attempts to do better, preserving more state when it aborts, some of which is logged to `ringbuf` and can be queried through `hiffy`.

There is room for improvement of the sequencer API, including the ability to report these aborts up stack, which unfortunately will have to wait until I/we have more time.


drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -8,29 +8,246 @@ use drv_fpga_api::{FpgaError, FpgaUserDesign, WriteOp};
...
+
+    fn try_from(v: u8) -> Result<Self, Self::Error> {
+        match v {
+            0 => Ok(PowerRailState::Disabled),
+            1 => Ok(PowerRailState::RampingUp),
+            2 => Ok(PowerRailState::GoodTimeout),
+            3 => Ok(PowerRailState::Aborted),
+            4 => Ok(PowerRailState::Enabled),
+            _ => Err(FpgaError::InvalidValue),
+        }
```
mkeeter, line 166:
```suggestion
        Self::from_u8(v).ok_or(FpgaError::InvalidValue)
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -8,29 +8,246 @@ use drv_fpga_api::{FpgaError, FpgaUserDesign, WriteOp};
...
+        })
+    }
+}
+
+impl PowerRail {
+    /// Convert power rail state register data to an array of `PowerRail`
+    /// structs.
+    pub fn from_raw(data: [u8; 6]) -> Result<[PowerRail; 6], FpgaError> {
+        #[inline]
+        const fn offset(addr: Addr) -> usize {
```
mkeeter, line 218: Nit: you could capture `data` here, using the same pattern as in `TofinoSeqStatus`:
```rust
let mut value = |addr: Addr| -> u8 {
    data[(addr as usize) - (Addr::TOFINO_POWER_VDD18_STATE as usize)]
};
```

mkeeter (approved):

PR #980, opened 2022-12-02
Bootrom alias

labbott:


flihp (approved):
LGTM: tested against gimlet-c w/ DICE mfg flow

PR #984, opened 2022-12-05
Allowing getting and setting the host boot flash device via control-plane-agent/MGS

jgallagher:
Also adds a mock `gimlet-hf` server for use on gimletlets that don't have a qspi for the real `gimlet-hf` to use.

### DIFF
--- a/app/gimletlet/app.toml
+++ b/app/gimletlet/app.toml
@@ -146,7 +146,11 @@ start = true
 task-slots = ["hash_driver", "hf", "i2c_driver", "rng_driver", "sprot", "sys", "update_server", "user_leds"]
 
 [tasks.hf]
+# If you do not have a gimletlet qspi-let adapter but want to test the hf API
+# (e.g., from control-plane-agent / MGS), switch to `drv-mock-gimlet-hf-server`.
 name = "drv-gimlet-hf-server"
+#name = "drv-mock-gimlet-hf-server"
+
 features = ["h753", "hash"]
 priority = 6
 max-sizes = {flash = 16384, ram = 4096}
--- a/app/gimletlet/app.toml
+++ b/app/gimletlet/app.toml
@@ -149,7 +149,6 @@ task-slots = ["hash_driver", "hf", "i2c_driver", "rng_driver", "sprot", "sys", "
 # If you do not have a gimletlet qspi-let adapter but want to test the hf API
 # (e.g., from control-plane-agent / MGS), switch to `drv-mock-gimlet-hf-server`.
 name = "drv-gimlet-hf-server"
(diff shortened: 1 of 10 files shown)


drv/mock-gimlet-hf-server/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// Selects between QSPI flash chips 1 and 2 (if present)
+    dev_state: HfDevSelect,
+}
+
+impl idl::InOrderHostFlashImpl for ServerImpl {
+    fn read_id(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<[u8; 20], RequestError<HfError>> {
+        Ok(*b"mockmockmockmockmock")
```
mkeeter, line 47: incredible code 💯

task/control-plane-agent/src/update/host_flash.rs
```diff
@@ -29,6 +29,42 @@ impl HostFlashUpdate {
...
+            .map_err(|err| SpError::UpdateFailed(err as u32))?
+        {
+            HfMuxState::SP => (),
+            HfMuxState::HostCPU => return Err(SpError::UpdateSlotBusy),
+        }
+
+        // Swap to the chosen slot.
+        self.task
+            .set_dev(slot)
+            .map_err(|err| SpError::UpdateFailed(err as u32))?;
```
mkeeter, line 64: It seems like there's a TOCTOU race here.

Luckily, `self.task.set_dev(slot)` _also_ checks that the mux is valid; we can probably eliminate the first check and just its its result directly

```suggestion
        // Swap to the chosen slot, returning an appropriate error if we don't
        // have control over the host flash.
        self.task
            .set_dev(slot)
            .map_err(|err| match err{
                HfError::NotMuxedToSP => SpError::UpdateSlotBusy,
                _ => SpError::UpdateFailed(err as u32)
            })?;
```

(untested)

jgallagher: Nice catch, fixed.

app/gimletlet/app.toml
```diff
@@ -146,7 +146,10 @@ start = true
 task-slots = ["hash_driver", "hf", "i2c_driver", "rng_driver", "sprot", "sys", "update_server", "user_leds"]
 
 [tasks.hf]
+# If you do not have a gimletlet qspi-let adapter but want to test the hf API
+# (e.g., from control-plane-agent / MGS), switch to `drv-mock-gimlet-hf-server`.
 name = "drv-gimlet-hf-server"
+
```
mkeeter, line 152: Extreme nit: extra newline here

mkeeter (approved):

PR #989, opened 2022-12-06
Seq fix

nathanaelhuffman:
This fixes logic bugs in the sequencer FPGA that were preventing A0 sequence state machine from turning off power properly.

This also adds a small delay between telling the sequencer to turn off and commanding pmbus off so that we deterministically notify the AMD processor that power is dropping per the EDS recomendation.


drv/gimlet-seq-server/src/main.rs
```diff
@@ -549,8 +549,23 @@ impl ServerImpl {
...
+                // so as not to trip a MAPO fault.
+                //
                 let a1a0 = Reg::PWR_CTRL::A1PWREN | Reg::PWR_CTRL::A0A_EN;
                 self.seq.clear_bytes(Addr::PWR_CTRL, &[a1a0]).unwrap();
+
+                //
+                // FPGA de-asserts PWR_GOOD for 2 ms before yanking enables, we wait
+                // for a tick here to make sure the SPI command to the FPGA
+                // propagated and the FPGA has had time to act. AMD's EDS doesn't
+                // give a minimum time so we'll givem them 1 ms.
```
mkeeter, line 566:
```suggestion
                // give a minimum time so we'll give them 1 ms.
```

mkeeter (approved):

PR #991, opened 2022-12-06
Add support for a `status` message to update-api

andrewjstone:
Currently the status only reports on the currently active slot. The value is correct for the lpc55, but there is a placeholder for the stm32h7. Future commits will provide more detail in the `UpdateStatus` value returned such as which version is in which slot, the image ids, and image hashes. We expect to remove the `current_version` API, such that it is subsumed as part of `status`.

The API has been plumbed through the update servers and sprot servers on both RoT and SP.

### DIFF
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -246,8 +246,11 @@ impl Handler {
             }
             MsgType::UpdCurrentVersionReq => {
                 let version = self.update.current_version();
-                let rsp: Result<ImageVersion, u32> = Ok(version);
-                hubpack::serialize(tx_payload, &rsp)?
+                hubpack::serialize(tx_payload, &version)?
+            }
+            MsgType::UpdStatusReq => {
+                let status = self.update.status();
+                hubpack::serialize(tx_payload, &status)?
             }
             MsgType::SinkReq => {
                 // The first two bytes of a SinkReq payload are the U16
@@ -268,6 +271,7 @@ impl Handler {
             | MsgType::UpdAbortUpdateRsp
             | MsgType::UpdFinishImageUpdateRsp
             | MsgType::UpdCurrentVersionRsp
+            | MsgType::UpdStatusRsp
             | MsgType::Unknown => {
                 status.rx_invalid = status.rx_invalid.wrapping_add(1);
                 return Err(SprotError::BadMessageType);
@@ -290,6 +294,7 @@ fn req_msgtype_to_rsp_msgtype(msgtype: MsgType) -> MsgType {
         MsgType::UpdAbortUpdateReq => MsgType::UpdAbortUpdateRsp,
         MsgType::UpdFinishImageUpdateReq => MsgType::UpdFinishImageUpdateRsp,
         MsgType::UpdCurrentVersionReq => MsgType::UpdCurrentVersionRsp,
+        MsgType::UpdStatusReq => MsgType::UpdStatusRsp,
         MsgType::SinkReq => MsgType::SinkRsp,
 
         // All of the unexpected messages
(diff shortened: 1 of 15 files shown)


drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -246,8 +246,11 @@ impl Handler {
             }
             MsgType::UpdCurrentVersionReq => {
                 let version = self.update.current_version();
-                let rsp: Result<ImageVersion, u32> = Ok(version);
-                hubpack::serialize(tx_payload, &rsp)?
+                hubpack::serialize(tx_payload, &version)?
```
lzrd, line 249: The information we will need includes:
Current version in stage0,
Current version in A, // if this is known, B is unknown
Current version in B, // if this is known, A is unknown
Currently executing bank, // given by active_slot() which you have added.
Currently executing version, // given by current_version()
Next executing bank on reset. // unknown

current_version() tells you the version of the currently executing code but doesn't tell you which bank you are executing out of.

We should establish the APIs for these data early.

lib/lpc55-flash/src/lib.rs
```diff
@@ -156,3 +156,15 @@ pub unsafe extern "C" fn __write_block(
...
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn __active_slot() -> UpdateTarget {
+    let current = this_image!();
+    if image_a_base!() == current {
+        UpdateTarget::ImageA
+    } else if image_b_base!() == current {
+        UpdateTarget::ImageB
+    } else {
```
lzrd, line 167: Test against image_stage0_base

drv/update-api/src/lib.rs
```diff
@@ -53,6 +54,22 @@ pub struct ImageVersion {
...
+    PartialEq,
+    Eq,
+    Serialize,
+    Deserialize,
+    SerializedSize,
+)]
+#[repr(C)]
+pub struct UpdateStatus {
+    pub active_slot: u8,
+}
```
labbott, line 71: This looks like it is being sent back and forth between RoT and SP. This also seems like a structure that could change as we discover new data we need to send back and forth. Do we need a structure version on this to be able to handle changes in fields?

labbott (commented):
Adding the new function to the access table LGTM!

lib/lpc55-flash/src/lib.rs
andrewjstone: Fixed in https://github.com/oxidecomputer/hubris/pull/991/commits/34fa9b3c7bfefaf747c7161db49937c79615f8c6

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: Thank you for the details @lzrd. I will implement this in a follow up commit immediately.

drv/update-api/src/lib.rs
andrewjstone: That's a great question @labbott. Perhaps. This structure itself will be embedded in a sprot structure before actually being sent over the wire, and so maybe that's where the versioning should happen. There's already a header with a [Protocol field](https://github.com/oxidecomputer/hubris/blob/master/drv/sprot-api/src/lib.rs#L260-L268), but that's just a u8 and really is meant to version the whole sprot API. Maybe there should be a separate per message version? I dealt with this in [sprockets](https://github.com/oxidecomputer/sprockets/blob/main/common/src/msgs.rs#L24-L36) by versioning the whole interface, but sprockets is expected to change much less than other things, and we want to change the whole protocol atomically.

It's certainly easy enough to add a version number here, and I'm open to it. @lzrd @jgallagher what do you think?

lzrd: It needs to be stabilized. There is a version on the protocol as a whole which must imply a consistent set of operations used to manage firmware updates (if nothing else).
We could add to the message set at any time while old implementations would return "not implemented".
Andrew is making some of these commits when they are still works in progress.

andrewjstone:
Closed in favor of #1009

PR #993, opened 2022-12-06
Enable DICE on Gimlet rot & create lab images

flihp:
Enables DICE our gimlet b & c builds. The new lab images configure DICE to create self-signed identity certs. Non-lab images configure DICE for the manufacturing process.

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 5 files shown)


labbott:
Are the Hubris images the same for the lab and not lab machines?

flihp:
As part of this work I took a look at the existing toml inheritence mechanism from 61bf972e6a but it doesn't support the mechanisms we'd need (changing kernel features). Adding this to the implementation may be possible but it didn't seem like the right thing to do. I'm happy to do this work if reviewers think it's necessary.

flihp:
> Are the Hubris images the same for the lab and not lab machines?

Yes. That might be necessary in the future but based on my work with @jclulow last week we only need a lab variant for stage0.

labbott:
> As part of this work I took a look at the existing toml inheritence mechanism from [61bf972](https://github.com/oxidecomputer/hubris/commit/61bf972e6a91af25c79daac9cbe175373754002c) but it doesn't support the mechanisms we'd need (changing kernel features). Adding this to the implementation may be possible but it didn't seem like the right thing to do. I'm happy to do this work if reviewers think it's necessary.

It would be nice to have but I don't think it's necessary for this PR to merge. We might be looking at TOML alternatives long term anyway.

labbott (approved):
LGTM

PR #1000, opened 2022-12-07
move QSFP module sequencing from SP -> FPGA

Aaron-Hartwig:
This PR is dependent on the merging of https://github.com/oxidecomputer/quartz/pull/62 and https://github.com/oxidecomputer/transceiver-control/pull/34


drv/transceivers-api/src/lib.rs
```diff
@@ -25,16 +27,60 @@ impl From<FpgaError> for TransceiversError {
...
+
+impl From<u8> for PowerState {
+    fn from(v: u8) -> PowerState {
+        match v {
+            0 => PowerState::A4,
+            1 => PowerState::A3,
+            2 => PowerState::A2,
+            3 => PowerState::A0,
+            4 => PowerState::Fault,
+            _ => unreachable!(),
```
mkeeter, line 69: I'm skeptical about non-total implementations of `from`, since they lead to Surprise Crashes.

The cleanest API design would be to make this `TryFrom`, e.g.
```rust
impl TryFrom<u8> for PowerState {
    type Error = ();
    fn try_from(v: u8) -> Result<PowerState, Self::Error> {
        match v {
            0 => Ok(PowerState::A4),
            1 => Ok(PowerState::A3),
            2 => Ok(PowerState::A2),
            3 => Ok(PowerState::A0),
            4 => Ok(PowerState::Fault),
            _ => Err(()),
        }
    }
}
```

(you'd then need to add an `unwrap` to `all_power_states`, which is the only place that uses this conversion)

drv/transceivers-server/src/main.rs
```diff
@@ -125,70 +126,49 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
...
         &mut self,
         _msg: &userlib::RecvMessage,
         mask: u32,
+        state: PowerState,
     ) -> Result<(), idol_runtime::RequestError<TransceiversError>> {
-        self.transceivers
-            .clear_reset(mask)
-            .map_err(TransceiversError::from)?;
-        Ok(())
+        Ok(self
```
mkeeter, line 146: Style nit: I don't love returning `Ok(some_function()?)` in cases when `some_function` returns a `Result<(), ...>`, because it's less obvious that we're not returning anything.

(this also applies to the two functions below)

drv/transceivers-server/src/udp.rs
```diff
@@ -270,53 +271,33 @@ impl ServerImpl {
...
+                        self.transceivers
+                            .set_power_state(PowerState::A0, mask)
+                            .map_err(|_e| {
                                 Error::PowerModeFailed(
-                                    HwError::SetPowerEnableFailed,
+                                    HwError::ControlPortWriteFailed,
                                 )
-                            },
-                        )?;
+                            })?;
```
mkeeter, line 300: Should we wait / check here that it succeeds in going to A0?  That would match the behavior of the main sequencers.

Aaron-Hartwig: We could. My reasoning in not doing so here is that we would need a timeout policy or something on how long to wait until we give up on a `Fault`ing module before we return. Since we are sequencing up to 32 things in parallel here, it felt a bit different than our other sequencers. I tried to give the host more information via `Status` now to communicate power faults should they occur. Would it make sense to implement a `HostRequest::GetPowerMode` that would read the actual `STATUS_PORTx` and let the host do the comparison / fault handling?

Cargo.toml
```diff
@@ -128,5 +128,5 @@ sprockets-rot = { git = "https://github.com/oxidecomputer/sprockets.git", defaul
 stm32g0 = { git = "https://github.com/oxidecomputer/stm32-rs-nightlies", branch = "stm32g0b1-update", default-features = false }
 tlvc = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
 tlvc-text = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
-transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control", default-features = false }
+transceiver-messages = { path = "../transceiver-control/messages", default-features = false }
```
Aaron-Hartwig, line 131: This will get backed out to point at git again once https://github.com/oxidecomputer/transceiver-control/pull/34 is merged and I don't need to keep switching it to test locally.

mkeeter (approved):

drv/transceivers-server/src/udp.rs
mkeeter: I'm on the fence, but am generally fine with keeping Hubris simple and punting this upstream to the host.

PR #1005, opened 2022-12-08
control-plane-agent: Allow Humility to proxy the serial console uart

jgallagher:
This PR adds a set of idol calls to `control-plane-agent` that allows Humility to step in and make itself the "client" for the host serial console uart (instead of the typical client: MGS). This is only useful for bench debugging, and will be accompanied by an associated humility PR to add a new command to exercise these functions.

The changes are _slightly_ more invasive than I expected / would have liked, because our default behavior is always read as much as we can from the RX FIFO, discarding it if necessary (and forwarding data to MGS as our buffers fill, so hopefully we don't _need_ to discard anything). However, Humility as a polling-on-about-1-Hz client is _so slow_ that we really need to make use of flow control to avoid losing data, so this PR also adds a couple 1-line functions to the uart driver allowing clients to enable or disable the RX FIFO interrupt (which until now was always on).

The flow from Humility's point of view is:

1. Call `set_humility_uart_client(attach=true)`. This will fail if there is a currently-attached MGS, but if there is, why are you trying to use Humility?
2. In a polling loop, call `uart_read()` to get data from the SP.
3. Simultaneously, watch stdin and forward any data written to it via `uart_write()`.
4. When disconnecting, call `set_humility_uart_client(attach=false)` to restore the normal "constantly read from the RX FIFO" behavior.

(If Humility disconnects uncleanly, one can always call the `attach=false` from step 4 at any time.)


task/control-plane-agent/src/main.rs
```diff
@@ -315,6 +312,89 @@ impl idl::InOrderControlPlaneAgentImpl for ServerImpl {
 
         Ok(id)
     }
+
+    #[cfg(feature = "gimlet")]
+    fn get_uart_client(
```
mkeeter, line 317: Optional: I usually put groups of operations like this into a separate `impl` block, to avoid copy-pasting the `#[cfg(feature...)]` many times.

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -235,6 +257,65 @@ impl MgsHandler {
...
+            };
+
+            if data.write_at(i, b).is_err() {
+                break;
+            }
+
+            i += 1;
+        }
+
+        if !self.usart.from_rx.is_full() {
```
mkeeter, line 280: Should we also check that `i > 0` here?

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -235,6 +257,65 @@ impl MgsHandler {
...
+        let mut i = 0;
+
+        while self.usart.tx_buffer_remaining_capacity() > 0 && i < data.len() {
+            // Min of all three: remaining buffer, remaining data, sizeof(chunk)
+            let n = usize::min(
+                self.usart.tx_buffer_remaining_capacity(),
+                usize::min(CHUNK_SIZE, data.len() - i),
+            );
+            if data.read_range(i..i + n, &mut chunk[..n]).is_err() {
+                return Ok(i);
```
mkeeter, line 310: Why does this return `Ok` instead of `Err`?  I'd expect a lease read to never fail here unless the lender died, which seems like it should be an error.

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -235,6 +257,65 @@ impl MgsHandler {
...
+        self.set_uart_client(UartClient::Humility)?;
+
+        let mut i = 0;
+        while i < data.len() {
+            let Some(b) = self.usart.from_rx.pop_front() else {
+                break;
+            };
+
+            if data.write_at(i, b).is_err() {
+                break;
```
mkeeter, line 274: Under what conditions should this fail?  We checked `i < data.len()` earlier, so I feel like this failing is an actual eror that should be reported.

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -885,22 +998,49 @@ impl UsartHandler {
...
+
+        // If our client is Humility, we only want to read from the usart if we
+        // have room in our buffer (i.e., we want to make use of flow control
+        // because Humility is a _very slow_ reader).
+        //
+        // If our client is MGS (the default), we always recv as much as we can
+        // from the USART FIFO, even if that means discarding old data. If an
+        // MGS instance is actually attached, we should be flushing packets out
+        // to it fast enough that we don't see this in practice.
+        if self.client == UartClient::Humility {
```
mkeeter, line 1012: Nit: use a `match` here just in case someone adds a _third_ client type?

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -960,6 +1097,29 @@ fn configure_usart() -> Usart {
...
+            // From thin air, pluck a pointer to the USART register block.
+            //
+            // Safety: this is needlessly unsafe in the API. The USART is
+            // essentially a static, and we access it through a & reference so
+            // aliasing is not a concern. Were it literally a static, we could
+            // just reference it.
+            let usart = unsafe { &*device::USART1::ptr() };
+            let peripheral = Peripheral::Usart1;
+            let pins = PINS;
+        } else if #[cfg(feature = "usart1-gimletlet")] {
```
mkeeter, line 1109: This conditional should probably begin with a check that no one has enabled both `usart` and `usart1-gimletlet`, since they're mutually exclusive.

task/control-plane-agent/src/mgs_gimlet.rs
jgallagher: I don't think it's necessary to; if someone calls us and we have no buffered data or they gave us a 0-length lease, we'll (unnecessarily) re-enable the RX FIFO interrupt, but the same is true if someone calls us when `from_rx` wasn't full. The call to `enable_rx_interrupt()` is just a register write, not an IPC call to a separate driver task, so I didn't think I needed to worry too much about avoiding calling it unnecessarily. Do you think it's more confusing this way?

task/control-plane-agent/src/mgs_gimlet.rs
jgallagher: Mmm yeah that's fair, this should really map to `ClientError::WentAway`, I guess? Out of curiosity - who receives that error if the client is gone?

task/control-plane-agent/src/mgs_gimlet.rs
jgallagher: This is also `ClientError::WentAway`. Will fix.

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -885,22 +998,49 @@ impl UsartHandler {
...
-                    // item, allowing us to unwrap `pop_front`. At that point we
-                    // know there's space for at least one, allowing us to
-                    // unwrap a subsequent `push_back`.
-                    self.from_rx.pop_front().unwrap_lite();
-                    self.from_rx.push_back(b).unwrap_lite();
-                    discarded_data += 1;
+
+        // If our client is Humility, we only want to read from the usart if we
+        // have room in our buffer (i.e., we want to make use of flow control
+        // because Humility is a _very slow_ reader).
```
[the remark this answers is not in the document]
jgallagher: It does as long as we've enabled hardware flow control. Technically that's a cargo feature, but we currently refuse to compile if it isn't set. Maybe I should just remove the feature? It was handy when I was first doing uart stuff with a cheap USB uart that didn't support flow control, but has probably outlived its usefulness.

task/control-plane-agent/src/mgs_gimlet.rs
mkeeter: Okay, I agree it's not necessary.  I'm on the fence whether it's more or less confusing...

task/control-plane-agent/src/mgs_gimlet.rs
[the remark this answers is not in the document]
mkeeter: Yeah, I'd vote to remove it, especially because the `cfg_if!` conditional is getting pretty hairy.

task/control-plane-agent/src/mgs_gimlet.rs
mkeeter: Good question!  I'd still like this to return an error instead of `Ok(i)`, but am not exactly sure where that error goes; maybe it's dropped by the kernel?

task/control-plane-agent/src/main.rs
jgallagher: Hmm, I don't think I can do this here, because these methods are required by `idl::InOrderControlPlaneAgentImpl`. Some options:

1. Leave it as-is
2. Split out separate `ServerImpl::get_uart_client_impl()` etc methods (that can be defined in separate impl blocks), then call them from here
3. Add methods to `mgs_sidecar`+`mgs_psc` so we can always call `self.mgs_handler.uart_client()` etc from here (i.e., no features required), and have those return errors instead

I think my preference of those is 1,3,2, but it's not particularly strong. What do you think?

task/control-plane-agent/src/mgs_gimlet.rs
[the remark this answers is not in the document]
jgallagher: Removed in b7de27e2

task/control-plane-agent/src/mgs_gimlet.rs
jgallagher: Fixed in 24ec457d

task/control-plane-agent/src/mgs_gimlet.rs
jgallagher: For posterity: in practice there is no difference here; the error / Ok is indeed dropped if the client is gone. However, it's certainly clearer to return an appropriate error; fixed in 24ec457d.

task/control-plane-agent/src/mgs_gimlet.rs
jgallagher: Fixed in e0eefae4

task/control-plane-agent/src/mgs_gimlet.rs
jgallagher: Added in a35001ca

task/control-plane-agent/src/main.rs
mkeeter: Huh, I thought that you could split the trait impl into multiple blocks (some gated with features), but it looks like that's not possible:
```
error[E0119]: conflicting implementations of trait `idl::InOrderControlPlaneAgentImpl` for type `ServerImpl`
   --> task/control-plane-agent/src/main.rs:318:1
    |
258 | impl idl::InOrderControlPlaneAgentImpl for ServerImpl {
    | ----------------------------------------------------- first implementation here
...
318 | impl idl::InOrderControlPlaneAgentImpl for ServerImpl {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `ServerImpl`

For more information about this error, try `rustc --explain E0119`.
error: could not compile `task-control-plane-agent` due to previous error
Error: failed to build control_plane_agent

Caused by:
    command failed, see output for details
```

In that case, I'm fine leaving this as-is.

mkeeter (approved):

PR #1009, opened 2022-12-09
Plumb RotBootState from Stage0 to ControlPlaneAgent

andrewjstone:
This change dumps boot time information about images into USB RAM using the
  `Handoff` mechanism created for dice. In order to allow types outside stage0
  to implement `HandoffData`, this type was moved into `lib/stage0-handoff`. That
  library also contains the definitions for the `RotBootState` that gets handed
  off. A `HandoffHeader` type was also created to allow versioning handoff data
  and standardizing on a format across types. The `fits_in_ram` macro replaces a
  bunch of longer static assertions in the DICE code.

  Additionally, since `HandoffData` uses `hubpack` for serialization, we end
  up with some optimized code that writes unaligned data to USB SRAM, which is
  treated as device memory and only allows aligned reads. This caused a usage
  fault. To fix this, the MPU is now enabled in stage0 with a region covering all
  of USB SRAM, that treats it as regular memory and therefore allows unaligned
  writes.

  The `RotBootState` is read from USB RAM by the `lpc55-update-server` and
  plumbed all the way to the `control-plane-agent` via `sprot`. The
  `control-plane-agent` plumbing required a change to `gateway-messages` that
  will be open in a separate PR to the `management-gateway-service` repo and
  which must be merged before this one. The plumbing through sprot also required
  some changes, with the most important  being the change to split the `status`
  API into a `status` api that returns information to be utilized across boots,
  and one less important one (io_stats) that may change, and currently just
  stores metrics.

  A bash script that uses the RoT Update hiffy API to load images, and was used
  for testing is also included.


app/rot-carrier/app.toml
```diff
@@ -34,7 +34,7 @@ priority = 3
 max-sizes = {flash = 16384, ram = 4096}
 stacksize = 2048
 start = true
-uses = ["rom", "secure_syscon", "syscon", "flash"]
+uses = ["rom", "secure_syscon", "syscon", "flash", "update_details"]
```
andrewjstone, line 37: It's unclear why we don't get alignment errors on read here. Probably just luck. This should be changed from a `uses` section to a memory array using `MaybeUninit` similar to https://github.com/oxidecomputer/hubris/blob/d9cf6c70d5cb505797e4d78e150db318ec463c8c/task/net/src/buf.rs#L17-L19

Cargo.toml
```diff
@@ -116,7 +116,7 @@ zeroize = { version = "1.5.7", default-features = false, features = ["zeroize_de
 
 # Oxide forks and repos
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
-gateway-messages = {git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = {git = "https://github.com/oxidecomputer/management-gateway-service", branch = "update-rot-state-msg", default-features = false, features = ["smoltcp"] }
```
jgallagher, line 119: Nit - let's merge the MGS branch and put this back on `main` before merging this PR.

idl/sprot.idol
```diff
@@ -40,14 +40,22 @@ Interface(
...
-            ok: "Status",
-            err: CLike("SprotError"),
+        "status": (
+            doc: "Return untrusted status from RoT",
+            reply : Result(
+              ok: "SprotStatus",
+              err: CLike("SprotError"),
+            ),
+            encoding: Hubpack
+		),
```
jgallagher, line 50: Nit - Indentation is off here; maybe tabs vs spaces? Where is my `idolfmt`

lib/stage0-handoff/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+mod rot_update_details;
+
+pub use rot_update_details::{
+    ImageVersion, RotBootState, RotImageDetails, RotSlot,
+};
+
+// This memory is the USB peripheral SRAM that's 0x4000 bytes long. Changes
+// to this address must be coordinated with the [dice_*] tables in
+// chips/lpc55/chip.toml
+// TODO: get from app.toml -> chip.toml at build time
```
jgallagher, line 23: Do you want to do this as part of this PR?

jgallagher (approved):

drv/update-api/hiffy-rot-flash.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/zsh
+
+# Load an image using the humility hiffy interface for the lpc55-update-server.
+#
+# Usage Example:
+# ARCHIVE=hubris/target/rot-carrier/dist/a/build-rot-carrier.zip rot-update-tools/hiffy-rot-flash.sh rot-carrier-v2/b/final.bin ImageB
```
labbott, line 6: I think this should work on the SP as well, might be worth testing just so we have this as another reference

stage0/src/image_header.rs
```diff
@@ -157,6 +161,17 @@ impl Image {
         header.version
     }
 
+    // TODO(AJS): Replace get_version with this?
```
labbott, line 164: the reason why the hubris `get_version`  doesn't access this directly is because if we're not running with TrustZone (i.e. on an s28) we don't have full access to be able to access the header the way it's currently designed. It probably would be much better to have it access the same data.

stage0/src/main.rs
```diff
@@ -145,6 +150,64 @@ unsafe fn branch_to_image(image: Image) -> ! {
     );
 }
 
+// Setup the MPU so that we can treat the USB RAM as normal RAM, and not as a
+// peripheral. Specifically we want to clear the `DEVICE` attributes, so that
+// we can allow unaligned access.
+//
+// NB: Portions opied from `sys/kern/src/arch/arm_m.rs:apply_memory_proteciton`
```
labbott, line 157: Might be nice to have a set of functions under `lib`?

stage0/src/main.rs
```diff
@@ -145,6 +150,64 @@ unsafe fn branch_to_image(image: Image) -> ! {
...
+    const USB_RAM_REGION_NUMBER: u32 = 0;
+
+    // Subtract 32 because the `LIMIT` field in the `rlar` register range is inclusive
+    // and then enable the region (bit 0).
+    let rlar = (USB_RAM_BASE + USB_RAM_SIZE - 32) | (1 << 0);
+
+    let ap = 0b01; // read-write by any privilege level
+    let sh = 0b00; // non-shareable - we only use one core with no DMA here
+    let xn = 1; // Don't execute out of this region
+    let rbar = USB_RAM_BASE | (sh as u32) << 3 | ap << 1 | xn;
```
labbott, line 182: hmmmm, I realize stage0 is not exactly Hubris but I wonder if we could generate this at build time as well. If we did it right we could add more regions to potentially catch stage0 touching unexpected parts of the system

lib/stage0-handoff/src/lib.rs
andrewjstone: TLDR: Let's leave it as is for this PR.

Longer answer:

This is a left over TODO from @flihp. However, as we've discovered alignment of 16 bits is required when the USB SRAM is treated as a peripheral, and that doesn't always play nicely with hubpack serialization and compiler optimizations. This PR works around that by setting the USB SRAM region as normal memory in the MPU in stage0, which has the effect of removing alignment requirements. I then took the advice of @cbiffle and @labbott and made the mapping in hubris its own linker section and pulled it in as a static with `#[link_section = ".bootstate]`. That actually already does fix the issue for my bootstate code here. The `UPDATE_RANGE` no longer lives in `chip.toml`, but in `memory.toml` and `256k.toml` and no hardcoding of addresses is used. This particular constant is left over to show what ranges exist while we still have things in `chip.toml`. 

All that being said, I *think* we'll want to do the same thing for the `DICE_RANGE` sections in chip.toml, but I'm not 100% sure. That's all @flihp's code and I don't want to make that change without him.

stage0/src/main.rs
andrewjstone: I'm up for doing this with guidance, but probably not as part of this PR.

stage0/src/main.rs
andrewjstone: Good idea! I think the 2 relevant functions are likely the disabling and enabling of the MPU, but it's possible I can also add one for the middle where we config the MPU. I'll take a stab at that.

stage0/src/image_header.rs
andrewjstone: @labbott Are you saying to just go ahead and do that now in this PR?

app/rot-carrier/app.toml
andrewjstone: Fixed.

idl/sprot.idol
andrewjstone: Fixed.

stage0/src/main.rs
andrewjstone: Done

drv/update-api/hiffy-rot-flash.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/zsh
+
```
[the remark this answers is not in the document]
lzrd: https://gist.github.com/5dca82a807cfa737bd22ee2805e0810a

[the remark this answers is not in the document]
andrewjstone: Updated, with a one line adjustment, so it runs on OSX where `stat` is lacking the `-c` option.

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -158,11 +193,9 @@ fn main() -> ! {
         underrun: false,
     };
 
-    let mut status = Status {
+    let mut state = LocalState {
         supported: 1_u32 << (Protocol::V1 as u8),
```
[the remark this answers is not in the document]
andrewjstone: Fixed.

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -193,7 +192,30 @@ impl Handler {
                 dst.copy_from_slice(rx_payload);
                 dst.len()
             }
-            MsgType::StatusReq => hubpack::serialize(tx_payload, &status)?,
+            MsgType::StatusReq => {
+                let rot_updates = match self.update.status() {
+                    UpdateStatus::Rot(rot_updates) => rot_updates,
+                    UpdateStatus::LoadError(e) => return Err(e.into()),
+                    UpdateStatus::Sp => panic!(),
```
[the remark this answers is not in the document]
andrewjstone: Fixed.

drv/update-api/hiffy-rot-flash.sh
andrewjstone: It does appear to work on the SP as well, although it's extremely slow.

drv/lpc55-update-server/src/main.rs
```diff
@@ -169,6 +177,19 @@ impl idl::InOrderUpdateImpl for ServerImpl {
             version: HUBRIS_BUILD_VERSION,
         })
     }
+
+    fn status(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<UpdateStatus, RequestError<Infallible>> {
+        // Safety: Data is published by stage0
+        let addr: &[u8] = unsafe { BOOTSTATE.assume_init_ref() };
```
labbott, line 186: I mentioned this before but I had bad luck with the compiler still trying to optimize calls to `assume_init` out. Maybe it works better with `assume_init_ref` because it's shared?

stage0/src/image_header.rs
labbott: I don't think it necessarily needs to do it in this PR but we should make sure it happens sooner rather than later to avoid confusion when the two inevitably end up out of sync.

andrewjstone: Ok. Got it. I'll likely do it in this PR.

app/gimletlet/app.toml
```diff
@@ -5,11 +5,12 @@ chip = "../../chips/stm32h7"
 memory = "memory-large.toml"
 stacksize = 896
 epoch = 0
-version = 0
+version = 2
```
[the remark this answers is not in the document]
andrewjstone: Good catch. This was an accidental commit. I think I've reverted this one twice already. I'll do it again.

drv/lpc55-update-server/src/main.rs
andrewjstone: I will talk to Cliff about it next week. I don't plan to merge before then anyway.

app/gimletlet/app.toml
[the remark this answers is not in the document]
andrewjstone: Will do. This was just an accidental commit while performing updates to test. It's definitely not required.

drv/update-api/hiffy-rot-flash.sh
```diff
@@ -0,0 +1,72 @@
...
+    fi
+  done
+}
+
+humility -a "${ARCHIVE}" hiffy -c Update.prep_image_update -a image_type="${image_type}"
+
+# Begin by invalidating the header which resides at offset 0x130 (in block zero).
+write_blocks 0 0 /dev/zero
+write_blocks 1 "${loopend}" "${image}"
+humility -a "${ARCHIVE}" hiffy -c Update.write_one_block -a block_num="${numblocks}" -i <(dd if="${image}" bs=1 count="${lastblocksize}" skip="${total_before_last_block}")
```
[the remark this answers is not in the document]
andrewjstone: Because github is confusing, I assume you tagged line 68. The purpose is because the last block may not be of block size, and we only want to write. the correct number of bytes.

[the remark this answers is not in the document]
lzrd: In the case of the LPC55, only 512 blocks can be written. For STM32, you could conceivably have a partial block.
There should probably be a warning on writing of a partial block.
This command won't be necessary if there are an integral number of blocks.

In later work, we need to enforce the state of all the blocks in the bank.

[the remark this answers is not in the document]
andrewjstone: The lpc55-update-server pads smaller blocks with zeros, which prevents the need to edit the binary or pad the last block directly during the update. Now this may cause problems if an updater tried to send an intermediate block smaller than 512b, and would prevent the update from booting this is also detectable with a hash / signature check which we need to do anyway.

app/gimlet-rot/app-b.toml
```diff
@@ -11,6 +11,7 @@ version = 0
 [kernel]
 name = "gimlet-rot"
 requires = {flash = 32768, ram = 4096}
+features = []
```
flihp, line 14: This doesn't look necessary? If it is feel free to disregard.

lib/stage0-handoff/src/lib.rs
flihp: No objections here. I'll get on changing the way the DICE regions are mapped once this is merged.

stage0/Cargo.toml
```diff
@@ -4,41 +4,43 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-dice-mfg = ["dice_crate", "digest", "hubpack", "salty", "serde", "sha3",
-            "static_assertions", "unwrap-lite", "lib-lpc55-usart"]
-dice-self = ["dice_crate", "digest", "salty", "sha3", "unwrap-lite",]
+dice-mfg = [ "salty", 
+            "static_assertions",  "lib-lpc55-usart"]
+dice-self = ["salty" ]
```
flihp, line 9: NIT: inconsistent whitespace

stage0/src/dice_mfg_self.rs
```diff
@@ -3,13 +3,14 @@
...
-use dice_crate::{CertData, DeviceIdSelfMfg, DiceMfg, Handoff};
+use crate::Handoff;
+use dice_crate::{CertData, DeviceIdSelfMfg, DiceMfg};
 use lpc55_pac::Peripherals;
 use salty::signature::Keypair;
 
 pub fn gen_mfg_artifacts(
     deviceid_keypair: &Keypair,
-    _peripherals: &Peripherals,
+    _: &Peripherals,
```
flihp, line 13: NIT: Seem unrelated to this PR. This was a style issue?

stage0/src/handoff.rs
```diff
@@ -0,0 +1,61 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
flihp, line 1: NIT: Stage0 is the only place where this half of the handoff code is used but when I was working on this initially I found it far easier to debug from a hubris task. Once I had working code in a library it was simple enough to just use the library from stage0. I'm still pretty new to rust so this may have been the wrong way to structure things? If so there's a lot of other stuff in `lib-dice` that needs to be moved down into stage0. If this is a style thing I've tended to put as much into libraries as possible and keeping this close to the consumer side of the handoff code seemed logical (to me at least).

drv/lpc55-update-server/src/main.rs
cbiffle: I'd be happy to walk you through how to check for this condition.

stage0/src/handoff.rs
andrewjstone: @flihp Great comment! I fully agree with the strategy you describe for most things. I actually had this part of the code in the `stage0-handoff` library which lives in `/lib` I moved it back into stage0 because the `stage0-handoff` library also contains the type definitions that get used on the SP and this code uses the `lpc55_pac` crate which makes it platform dependent and ends up in a symbol conflict when linking. 

I could have, I suppose, created a feature for `stage0-handoff` that only compiles this part for stage0, but that seems somewhat silly given this code only runs in stage0 ever. I also could have created two separate crates, or translation of types between the RoT and SP and kept the whole stage0-handoff crate only compiling for the lpc55, but that seemed like a touch overkill.

app/gimlet-rot/app-b.toml
andrewjstone: This is very unfortunate. Once I added a new section for `sram3` to `tasks.update_server`, it appeared that I had to add this line everywhere to get the containing images to build.

stage0/src/dice_mfg_self.rs
andrewjstone: That was a due to a couple of changes. I removed that line as I didn't think it was necessary, but then the dice code wouldn't build. I didn't realize the function is essentially part of an implicit trait. When I added it back, I just never added the rest of the name I guess, as it seemed redundant.

app/gimlet-rot/app-b.toml
andrewjstone: Turns out this is unnecessary. I just needed to bump humility locally.

PR #1010, opened 2022-12-09
Add transceivers to thermal control loop

mkeeter:
This PR adds the 32x Sidecar transceivers to the thermal loop managed by `task-thermal`.

This is non-trivial: until now, the thermal loop assumed that all of its sensors would be known I2C devices, and that it would be responsible for polling them itself.

Transceivers break these assumptions:
- A given transceiver isn't fixed; users may insert hardware with different specifications and thermal limits
- Transceivers, while _technically_ I2C devices, aren't read through our normal I2C code; they're read through a pair of FPGAs on the front IO board.

The fix is to add a new category of sensors to the thermal loop.  "Dynamic inputs" represent sensors which are both
- Configured at runtime with a particular thermal model, and
- Fed data by a **different task**

The different task is our `transceivers` task, which is the liaison to the FPGAs.  When a transceiver is detected, this task picks a thermal model and interface*, then periodically polls and sends temperature data to `thermal`.

There are also a few other drive-by changes:

- Switching to strong types for `Physical` vs `Logical` transceiver ports
- Polling the _actual_ power state of Sidecar, instead of the sequencer configuration
- **Enabling the thermal loop for Sidecar**, with the PID tuning from Gimlet (as a temporary value)

This PR _does not_ integrate the transceiver temperature measurements into the `sensors` task.  I'm going to investigate how feasible that is, but it shouldn't block reviewing the PR.

*Right now, the interface is selected based on byte 0 of the memory map, which contains an identifier code; the thermal model is picked arbitrarily and somewhat conservatively.  Eventually, we expect host software to send us this data before enabling a transceiver.


drv/transceivers-server/src/main.rs
```diff
@@ -112,6 +158,178 @@ impl ServerImpl {
...
+    }
+
+    /// Trigger a read from the given port's given register, which is assumed to
+    /// be an `i16` containing 1/256 °C.
+    fn read_temperature_from_i16(
+        &self,
+        port: LogicalPort,
+        reg: u8,
+    ) -> Result<Celsius, FpgaError> {
+        self.transceivers.setup_i2c_read(reg, 2, port.as_mask())?;
```
jgallagher, line 191: At first glance I was surprised this was 2 when `StatusAndTemperature` has 3 bytes worth of data. Is status treated separately even though we pass a `StatusAndTemperature` as a `&mut [u8]`?

drv/transceivers-server/src/main.rs
```diff
@@ -112,6 +158,178 @@ impl ServerImpl {
...
+
+        #[derive(Copy, Clone, FromBytes, AsBytes)]
+        #[repr(C)]
```
jgallagher, line 194: Does this also need `packed`? I think it's currently 4 bytes with a pad byte after `status`; not sure how that interacts with `zerocopy`.

drv/transceivers-server/src/main.rs
```diff
@@ -112,6 +158,178 @@ impl ServerImpl {
...
+
+    fn get_transceiver_interface(
+        &mut self,
+        port: LogicalPort,
+    ) -> Result<ManagementInterface, FpgaError> {
+        self.transceivers.setup_i2c_read(0, 1, port.as_mask())?;
+        let mut out = [0u8; 2]; // [status, SFF8024Identifier]
+
+        // Wait for the I2C transaction to complete
+        loop {
```
jgallagher, line 225: If this gets stuck it'll starve lower priority tasks. Could it have a max count or start sleeping after some number of retries?

drv/transceivers-server/src/main.rs
```diff
@@ -112,6 +158,178 @@ impl ServerImpl {
...
+                self.thermal_models[i] = None;
+            }
+        }
+
+        for (i, m) in self.thermal_models.iter().enumerate() {
+            let port = LogicalPort(i as u8);
+            let m = match m {
+                Some(m) => m,
+                None => continue,
+            };
```
jgallagher, line 293: Might be too clever, but I think you could squish this down into the iterator:

```rust
        for (port, m) in self
            .thermal_models
            .iter()
            .enumerate()
            .filter_map(|(i, m)| m.map(|m| (LogicalPort(i as u8), m)))
        {
```

It's not any shorter, but arguably might be slightly more obvious that the for loop is only operating on a subset of `self.thermal_models`?

drv/transceivers-server/src/udp.rs
```diff
@@ -168,6 +100,81 @@ impl ServerImpl {
         }
     }
 
+    fn handle_packet(
+        &mut self,
+        mut meta: UdpMetadata,
+        rx_data_buf: &[u8],
+        tx_data_buf: &mut [u8],
+    ) {
+        // Modify meta.size based on the output packet size
```
jgallagher, line 109: I think this comment should move down to line 119 or 120?

task/thermal/src/bsp/sidecar_ab.rs
```diff
@@ -22,13 +22,26 @@ use i2c_config::sensors;
...
+// Air temperature sensors, which aren't used in the control loop
 const NUM_TEMPERATURE_SENSORS: usize = sensors::NUM_TMP117_TEMPERATURE_SENSORS;
+
+// Temperature inputs (I2C devices), which are used in the control loop.
 pub const NUM_TEMPERATURE_INPUTS: usize =
     sensors::NUM_TMP451_TEMPERATURE_SENSORS;
+
+// External temperature inputs, which are provided to the task over IPC
+// In practice, these are our transceivers.
+pub const NUM_DYNAMIC_TEMPERATURE_INPUTS: usize = 32;
```
jgallagher, line 37: Could this be `transceivers_api::NUM_PORTS`?

task/thermal/src/bsp/sidecar_ab.rs
```diff
@@ -22,13 +22,26 @@ use i2c_config::sensors;
...
+
 const NUM_FANS: usize = sensors::NUM_MAX31790_SPEED_SENSORS;
 
 // The Sidecar controller hasn't been tuned yet, so boot into manual mode
```
jgallagher, line 41: Remove/change comment? Still isn't tuned, maybe, but isn't booting into manual mode.

task/thermal/src/control.rs
```diff
@@ -357,15 +360,21 @@ impl Default for OneSidedPidState {
 }
 
 /// This corresponds to states shown in RFD 276
+///
+/// All of our temperature arrays contain, in order
+/// - I2C temperature inputs (read by this task)
+/// - Dynamic temperature inputs (read by another task and passed in)
 enum ThermalControlState {
     /// Wait for each sensor to report in at least once
```
jgallagher, line 368: Is this comment still accurate? I.e., do we wait for all the dynamic inputs to report at least once?

mkeeter:
Update: this now also integrates the new sensors into the `sensors` task!  There's an extra block added to the configuration which lets you define non-I2C sensors, which are then bundled into the rest of the `sensors` configuration.

The `sensors` configuration is also moved into `sensors_api`, so that tasks don't need to use `build_i2c` directly.

app/sidecar/rev-b.toml
```diff
@@ -274,13 +274,14 @@ task-slots = ["sys", "i2c_driver", {spi_driver = "spi1_driver"}]
 name = "drv-transceivers-server"
 features = ["vlan"]
 priority = 6
-max-sizes = {flash = 65536, ram = 8192}
-stacksize = 2048
+max-sizes = {flash = 65536, ram = 16384}
+stacksize = 4096
```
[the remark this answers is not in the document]
mkeeter: Yes, done in e396d578cf33b0b472a5569789fefad8824f8755

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -274,13 +313,13 @@ impl Transceivers {
...
-            if mask != 0 {
+            if !mask.is_empty() {
                 let fpga = self.fpga(fpga_index);
                 for port in 0..16 {
-                    if mask & (1 << port) != 0 {
+                    if mask.is_set(PhysicalPort(port)) {
                         fpga.write(
                             WriteOp::Write,
-                            Addr::QSFP_CONTROL_PORT0 as u16 + port,
+                            Addr::QSFP_CONTROL_PORT0 as u16 + port as u16,
```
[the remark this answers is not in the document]
mkeeter: Sure, done in 3cd7a8b359bbb713f9d304b37592021ae2585bd5

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -507,6 +547,20 @@ impl Transceivers {
             .read_bytes(Self::read_buffer_address(port_loc.port), buf)
     }
 
+    /// Get `buf.len()` bytes of data, where the first byte is port status and
+    /// trailing bytes are the I2C read buffer for a `port`. The buffer stores
+    /// data from the last I2C read transaction done and thus only the number of
+    /// bytes read will be valid in the buffer.
```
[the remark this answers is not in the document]
mkeeter: Presumably because it's the one that ordered the last transaction!

drv/transceivers-server/src/main.rs
mkeeter: [`zerocopy::byteorder::I16`](https://docs.rs/zerocopy/latest/zerocopy/byteorder/struct.I16.html#) is unaligned, so this ends up being tightly packed!  I could switch to `repr(packed)` if you think it would be more clear, though.

drv/transceivers-server/src/main.rs
mkeeter: You had me doubting myself, but I think it's correct:
- `setup_i2c_read` tells us how many bytes to read over I2C (i.e. from the transceiver).  These bytes are written to the I2C read buffer.
- `get_i2c_status_and_read_buffer` reads bytes from the FPGA, so it reads the status byte then two bytes from the I2C read buffer (which is right after it in FPGA register space).

drv/transceivers-server/src/udp.rs
mkeeter: Yes, done in 2f7c6bf9b6de413d63fcaaa7005d3a7b78319c4a

task/thermal/src/bsp/sidecar_ab.rs
mkeeter: Yeah, done in 1829494e21c133fc6dfcf5e26bc9b9c480ba94a8

(although this task would still need edits if it changed, because of the new `DYNAMIC_INPUTS` array below, which arrived after your review)

task/thermal/src/control.rs
mkeeter: Every `dynamic_input` that's not `None` has to report in at least once; if they are `None`, then they are skipped during `Boot` and become `TemperatureReading::Inactive` during `Running`.

Double-checking this logic would be appreciated!

jgallagher: Yep, I read over that later and forgot to come back to this comment. It looks good to me.

drv/transceivers-server/src/main.rs
jgallagher: Ah, I didn't realize that! I should have, though - w.r.t. "not sure how [padding] interacts with zerocopy": it doesn't. If you were using regular `i16` here (which would have a pad byte) zerocopy would fail to derive `AsBytes`.

I don't think adding `repr(packed)` is necessary given all that.

bnaecker (approved):
Looks good! A couple of small suggestions, but the transceiver details seem solid.

task/thermal/src/bsp/sidecar_ab.rs
mkeeter: Reverted to booting in manual mode!

task/sensor-api/build.rs
```diff
@@ -2,7 +2,132 @@
...
+        for (i, d) in config_sensor.devices.iter().enumerate() {
+            by_device.entry(d.device.clone()).or_default().push(i);
+            if !names.insert(d.name.clone()) {
+                bail!("Duplicate sensor name: {}", d.name);
+            }
+        }
+
+        let mut sensors_text = String::new();
+        let mut sensor_id = 0;
+        for d in &config_sensor.devices {
```
jgallagher, line 56: I'm not sure this is the right place for this code - in particular, I assume we (eventually?) want to be able to report these sensor readings up to MGS, which means `control-plane-agent` needs to be able to step over this list at runtime. For i2c sensors I mucked with `build-i2c` to add sensor descriptions to `validate-api`, but it seems likely that's not really the right place either? 

I don't have a concrete suggestion here and leave this comment primarily as a data point going into the upcoming discussion on organizing sensors.

drv/transceivers-server/src/main.rs
jgallagher: I'm sold; not sure how to suggest making this better. Maybe just a comment that the count here is specific to the i16 field and the status is separate?

drv/sidecar-front-io/src/transceivers.rs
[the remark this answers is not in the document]
jgallagher: Fair enough, hah.

jgallagher (approved):

drv/transceivers-server/src/main.rs
```diff
@@ -112,6 +158,180 @@ impl ServerImpl {
...
+
+            let temperature = match m.interface {
+                ManagementInterface::Cmis => self.read_cmis_temperature(port),
+                ManagementInterface::Sff8636 => {
+                    self.read_sff8636_temperature(port)
+                }
+                ManagementInterface::Unknown(..) => {
+                    // TODO: what should we do here?
```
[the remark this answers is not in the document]
mkeeter: Once the host is responsible for setting up the thermal loop, I'd like to consider removing `ManagementInterface::Unknown` entirely – if you don't know what it is, you aren't allowed to turn it on 😆

PR #1012, opened 2022-12-10
Upgrade Hmac crate to latest version, use it in CDI derivation.

flihp:
This resolves a TODO in lib/dice

### DIFF
--- a/lib/dice/src/lib.rs
+++ b/lib/dice/src/lib.rs
@@ -6,6 +6,7 @@
 
 use core::{mem, slice};
 use hkdf::Hkdf;
+use hmac::{Hmac, Mac};
 use hubpack::SerializedSize;
 use salty::constants::SECRETKEY_SEED_LENGTH;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


lib/dice/src/lib.rs
```diff
@@ -189,18 +189,13 @@ impl SeedBuf for CdiL1 {
...
-        // should be the IKM? Does it matter?
-        let hk = Hkdf::<Sha3_256>::new(Some(tcb_tci), cdi.as_bytes());
-        // No info provided to 'expand', see RFC 5869 §3.2.
-        // TODO: return error instead of expect
-        hk.expand(&[], &mut okm).expect("failed to expand");
-
-        CdiL1(okm)
+    pub fn new(cdi: &Cdi, tci: &[u8; SEED_LENGTH]) -> Self {
+        let mut hmac = Hmac::<Sha3_256>::new_from_slice(cdi.as_bytes())
+            .expect("HmacSha3_256");
```
cbiffle, line 194: What would cause this condition to fail? It looks like sizeof(Cdi) being wrong?

I ask because `expect` (and `unwrap`) are both relatively expensive in terms of text size. Changing this to `unwrap` would improve things a bit. Using `unwrap_lite` would improve things a lot but that's only available in Hubris right now.

cbiffle (approved):
One comment on potential text size of generated code below.

PR #1013, opened 2022-12-12
Mux forensics + sensor error counters

bcantrill:
Draft commit message:

```
This adds some forensics to better aid debuggability of an I2C hang on
a selected segment of a mux.  Specifically this work:

- Adds an interface to the I2C server to get the last selected
  mux/segment
- Modifies the I2C server to clear the mux/segment state *after* that
  state has been cleared successfully to preserve mux/segment state in
  the event of an unresponsive mux or hung bus
- Adds some rough error counters to the Sensor API

```


drv/stm32xx-i2c-server/src/main.rs
```diff
@@ -121,16 +121,18 @@ fn configure_mux(
...
-            // it point to the old mux+segment.
-            map.remove((controller.controller, port));
         }
 
-        // If we're here, our mux is valid, but the current segment is
-        // not the specfied segment; we will now call upon our
-        // driver to enable this segment.
+        //
+        // If we're here, our mux is valid, but the current segment is not the
+        // specfied segment; we will now call upon our driver to enable this
```
mkeeter, line 128: pre-existing typo:
```suggestion
        // specified segment; we will now call upon our driver to enable this
```

task/sensor/src/main.rs
```diff
@@ -69,11 +70,38 @@ impl idl::InOrderSensorImpl for ServerImpl {
...
+
+            //
+            // We pack per-`NoData` counters into a u32.
+            //
+            let (nbits, shift) = nodata.counter_encoding::<u32>();
+            let mask = (1 << nbits) - 1;
+            let bitmask = mask << shift;
+            let incr = 1 << shift;
+
+            if self.nerrors[index] & bitmask != bitmask {
```
mkeeter, line 82: This isn't obvious at first glance; maybe a comment here explaining that it's checking for saturating?

task/sensor/src/main.rs
```diff
@@ -98,6 +126,7 @@ fn main() -> ! {
 
     let mut server = ServerImpl {
         data: [Reading::Absent; NUM_SENSORS],
+        nerrors: [0; NUM_SENSORS],
```
mkeeter, line 129: This is going to conflict with https://github.com/oxidecomputer/hubris/pull/1018 (should be an easy fix)

mkeeter (approved):

drv/stm32xx-i2c-server/src/main.rs
bcantrill: Fixed -- thank you!

task/sensor/src/main.rs
bcantrill: Definitely; thanks, and fixed.

PR #1014, opened 2022-12-12
Add a README about where to find the RoT image for psc and sidecar

labbott:


flihp (approved):

app/psc/ROT_README.md
```diff
@@ -0,0 +1,56 @@
...
+humility:         chip => LPC55, ROM revision 1, device revision 0x1 (1B), ROM patch 0x8
+```
+
+*The expectation is that all new boards going forward will have the fixed
+LPC55S69 and that `gimlet-rot-c` should be used unless otherwise specified*
+
+The following setups should use `gimlet-rot-b`:
+
+- All Gimlet Rev B
+- Unreworked PSC (the first pass of these incorrectly had LPC55S28)
```
rmustacc, line 36: So, I think I would clarify this a bit. There are no Rev B PSCs that exist with the incorrect BOM entry. However, PSC Rev A is in this state. So I would juse phrase it as 'All Gimlet Rev A'.

app/sidecar/ROT_README.md
```diff
@@ -0,0 +1,56 @@
...
+humility:         chip => LPC55, ROM revision 1, device revision 0x1 (1B), ROM patch 0x8
+```
+
+*The expectation is that all new boards going forward will have the fixed
+LPC55S69 and that `gimlet-rot-c` should be used unless otherwise specified*
+
+The following setups should use `gimlet-rot-b`:
+
+- All Gimlet Rev B
+- Unreworked PSC (the first pass of these incorrectly had LPC55S28)
```
rmustacc, line 36: Presumably this should say Sidecar Rev A?

rmustacc (commented):
Thanks for putting this together @labbott.

PR #1017, opened 2022-12-12
Fix off-by-one-ish error in MPU RLAR programming.

### LINKED ISSUE #1016: ARMv8-M MPU setup appears to have an off-by-32 error
@flihp has been having a hard time making abutting regions on the LPC55; reviewing the kernel code I noticed this:

```rust
        let rlar = (region.base + region.size)
                | (i as u32) << 1 // AttrIndx
                | (1 << 0); // enable
```

The RLAR LIMIT field is described by ARM as containing...

> Ending address (upper inclusive limit)of MPU region address (bits [31:5] - the address must be aligned to multiple of 32 bytes). Bit [4:0] of the address value is assigned with 0x1F to provide the limit address to be checked against.

This means I think we've got an off-by-1-32-byte-chunk error on all regions right now, that needs to have a `- 32` added.

flihp:
The error stems from the RLAR value being inclusive which typically produces an off by one error. In this case the region is 32 byte aligned and so it's an off by one 32 byte chunk error instead.

This resolves #1016

### DIFF
--- a/sys/kern/src/arch/arm_m.rs
+++ b/sys/kern/src/arch/arm_m.rs
@@ -471,7 +471,7 @@ pub fn apply_memory_protection(task: &task::Task) {
         };
 
         // RLAR = our upper bound
-        let rlar = (region.base + region.size)
+        let rlar = (region.base + region.size - 32)
                 | (i as u32) << 1 // AttrIndx
                 | (1 << 0); // enable
 


flihp:
probably worth noting: this was tested on an lpc55s69 dev board and a gimletlet rot carrier carrier

cbiffle (approved):

PR #1020, opened 2022-12-13
Enable VLANs on the management network, with one exception

mkeeter:
The difference between `configure_vlan_strict` and `configure_vlan_semistrict` is the handling of technician ports: in this PR, the technician ports remain able to talk to every SP.  This makes development easier, until MGS is ready to go (and booting autonomously).

I tested this on the `niles` rev B Sidecar, and confirmed that the SPs are now isolated:
```
matt@niles ~ (sidecar-b) $ h monorail mac
Reading 4 MAC addresses...
 PORT |        MAC
------|-------------------
   13 | 0e:1d:80:ba:b8:c6
   15 | 0e:1d:80:ba:b8:c7
   44 | 00:0e:c6:46:43:d9
   48 | 0e:1d:dc:cf:27:34
matt@niles ~ (sidecar-b) $ h net mac
humility: Reading 3 MAC addresses...
 PORT |        MAC
------|-------------------
    1 | 00:0e:c6:46:43:d9
    3 | 0e:1d:dc:cf:27:34
      | 0e:1d:dc:cf:27:35
```

The first command shows that the VSC7448 switch is receiving packets from 3 SPs (ports 13, 15, and 48; MAC addresses beginning with `0e:1d`).  The second command shows that the SP's local switch is only seeing packets from the SP itself (on PORT 3) and the host (on PORT 1); this indicates that SP-to-SP communication is no longer possible.

### DIFF
--- a/drv/vsc7448/src/lib.rs
+++ b/drv/vsc7448/src/lib.rs
@@ -704,9 +704,9 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
     /// - The uplink port (49) sends and receives packets with one VLAN tag, and
     ///   uses that packet to select which VLAN (i.e. which downstream port)
     ///   should receive that packet.  The VLAN tag is stripped on egress.
-    fn configure_vlan_with_mask<F: Fn(u8) -> u64>(
+    fn configure_vlan_with_mask(
         &self,
-        f: F,
+        f: fn(u8) -> u64,
     ) -> Result<(), VscError> {
         const UPLINK: u8 = 49; // DEV10G_0, uplink to the Tofino 2
 
--- a/drv/vsc7448/src/lib.rs
+++ b/drv/vsc7448/src/lib.rs
@@ -691,8 +691,23 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
         Ok(())
     }
 
-    /// Implements the VLAN scheme described in RFD 250.
-    pub fn configure_vlan_strict(&self) -> Result<(), VscError> {
+    /// Configures VLANs for a production-style system.
+    ///
+    /// For details, see RFD 250, but in brief:
+    ///
+    /// - VLANs in the form 0x1YY are configured for each downstream port.
+    ///   Ports active on a given VLAN are selected by the given function `f`;
+    ///   typically, each VLAN will contain a downstream port and the uplink
+    ///   port.
+    /// - Downstream ports send and receive untagged frames, and apply their
+    ///   VLAN tag on packet ingress.
+    /// - The uplink port (49) sends and receives packets with one VLAN tag, and
+    ///   uses that packet to select which VLAN (i.e. which downstream port)
+    ///   should receive that packet.  The VLAN tag is stripped on egress.
+    fn configure_vlan_with_mask<F: Fn(u8) -> u64>(
+        &self,
+        f: F,
+    ) -> Result<(), VscError> {
         const UPLINK: u8 = 49; // DEV10G_0, uplink to the Tofino 2
 
         // Enable the VLAN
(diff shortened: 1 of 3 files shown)


drv/vsc7448/src/lib.rs
```diff
@@ -766,6 +766,98 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
...
+                (1 << p)
+                    | (1 << UPLINK)
+                    | (1 << TECHNICIAN_1)
+                    | (1 << TECHNICIAN_2)
+            };
+
+            // Configure the 0x1YY VLAN for this port
+            self.write_port_mask(
+                ANA_L3().VLAN(0x100 + p as u16).VLAN_MASK_CFG(),
+                mask,
```
jgallagher, line 807: Is this function identical to `configure_vlan_strict()` except for this mask? If so, could we split it into a common function that takes a `port -> mask` closure?

task/monorail-server/src/bsp/sidecar_a.rs
```diff
@@ -194,6 +194,7 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
         self.phy_vsc8504_init()?;
         self.phy_vsc8562_init()?;
         self.vsc7448.configure_ports_from_map(&PORT_MAP)?;
+        self.vsc7448.configure_vlan_semistrict()?;
```
jgallagher, line 197: How do we call `configure_vlan_strict()` instead? (Wondering if `configure_vlan_strict()` should be the default, with the `-lab` app.tomls setting a feature to switch to `configure_vlan_semistrict()`.)

jgallagher (approved):

task/monorail-server/src/bsp/sidecar_a.rs
mkeeter: Right now, we don't call `configure_vlan_strict()` ever!

I suspect we want to keep using `semistrict` in **both** the regular and lab images for the moment; we'll be using the regular image for compliance, and I imagine folks may want technician port access to arbitrary SPs.

drv/vsc7448/src/lib.rs
```diff
@@ -766,6 +766,98 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
...
+        // The uplink port requires one VLAN tag, and pops it on ingress
+        //
+        // It has a default VID of 0x1, but we removed all ports from
+        // that VLAN, so it will only accept our desired set of VIDs.
+        let port = ANA_CL().PORT(UPLINK);
+        self.modify(port.VLAN_CTRL(), |r| {
+            r.set_vlan_pop_cnt(1);
+            r.set_vlan_aware_ena(1);
+        })?;
+        // Only accept 0x8100 as a valid TPID, to keep things simple,
```
bnaecker, line 832: What's the relationship between `0x8100` and the bits written below? Maybe a quick comment here would be helpful.

bnaecker (approved):
I don't pretend to understand the details of operating against the VSC7448 itself, but generally looks ok to me!

drv/vsc7448/src/lib.rs
mkeeter: I was suspicious of this, but did end up liking it (3f0c314ae696b2a8df73b05306682a72f4cb9038)

drv/vsc7448/src/lib.rs
mkeeter: `0x8100` is the 802.1Q EtherType for "customer VLAN", so it's baked into the VSC7448 as one of the default options for "what TPID do I accept".  I added a little more detail in d728b905c196b1b43d77b13e822d781bef087356

drv/vsc7448/src/lib.rs
```diff
@@ -691,8 +691,23 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
...
+    ///   typically, each VLAN will contain a downstream port and the uplink
+    ///   port.
+    /// - Downstream ports send and receive untagged frames, and apply their
+    ///   VLAN tag on packet ingress.
+    /// - The uplink port (49) sends and receives packets with one VLAN tag, and
+    ///   uses that packet to select which VLAN (i.e. which downstream port)
+    ///   should receive that packet.  The VLAN tag is stripped on egress.
+    fn configure_vlan_with_mask<F: Fn(u8) -> u64>(
+        &self,
+        f: F,
```
jgallagher, line 709: I think because the closures passed here don't actually close over anything, it doesn't need to be generic and can be

```rust
f: fn(u8) -> u64,
```

with no changes at the call sites.

mkeeter: Neat! I had assumed that wouldn't work (because it was a closure), but apparently [non-capturing closures will automatically be coerced into plain `fn`'s](https://rust-lang.github.io/rfcs/1558-closure-to-fn-coercion.html).

PR #1023, opened 2022-12-14
Add timestamps to sensor readings

mkeeter:
This PR adds timestamps to readings within the `sensors` task, which must be supplied when calling `Sensor.post`.

In addition, sensor data is reorganized into a struct-of-arrays which stores the most recent reading **and** the current error (if present).  Using a struct-of-arrays saves a significant amount of RAM: `sidecar/rev-b.toml` goes from 2296 to 2868 bytes, but we've added 760 bytes of timestamps (95 sensors * 8 bytes each), so we're coming out ahead.

The existing `Sensor.get` API is unchanged to preserve compatibility with Humility; the PR adds `Sensor.get_reading` which returns both the value and the timestamp.


task/power/src/main.rs
```diff
@@ -429,6 +429,7 @@ fn main() -> ! {
         hl::sleep_for(1000);
 
         let state = get_state();
+        let now = sys_get_timer().now;
```
cbiffle, line 432: I feel like we ought to record this closer to where the measurement happens (i.e. down by `read_temperature` et al)

task/sensor-api/src/lib.rs
```diff
@@ -26,11 +26,22 @@ impl From<SensorId> for usize {
...
-pub enum Reading {
-    Absent,
-    Value(f32),
-    NoData(NoData),
+#[derive(Copy, Clone, Debug, zerocopy::FromBytes, zerocopy::AsBytes)]
+#[repr(C)]
+pub struct Reading {
+    pub timestamp: u64,
+    pub value: f32,
+    _pad: u32, // Required for FromBytes / AsBytes
```
cbiffle, line 34: Consider using ssmarshal/hubpack to avoid having to do this.

task/sensor-polling/src/main.rs
```diff
@@ -68,14 +68,14 @@ impl TemperatureSensor {
...
-    fn poll(&self, i2c_task: TaskId, sensor_api: &Sensor) {
+    fn poll(&self, i2c_task: TaskId, sensor_api: &Sensor, now: u64) {
         let dev = (self.builder)(i2c_task);
         match &self.device {
             Device::Mwocp68 => {
                 for (i, &s) in self.temperature_sensors.iter().enumerate() {
                     let m = Mwocp68::new(&dev, i.try_into().unwrap());
                     let post_result = match m.read_temperature() {
-                        Ok(v) => sensor_api.post(s, v.0),
+                        Ok(v) => sensor_api.post(s, v.0, now),
```
cbiffle, line 78: (should probably read the time here-ish)

task/sensor/src/main.rs
```diff
@@ -49,11 +66,15 @@ impl idl::InOrderSensorImpl for ServerImpl {
...
+        timestamp: u64,
     ) -> Result<(), RequestError<SensorError>> {
         let index = id.0;
 
         if index < NUM_SENSORS {
-            self.data[index] = Reading::Value(value);
+            self.has_data[index] = true;
+            self.data_value[index] = value;
+            self.data_time[index] = timestamp;
+            self.nodata[index] = None;
```
cbiffle, line 77: If you timestamp the nodata reading, I would suggest removing this line so we can see the last error even if there's been a success since then.

task/sensor/src/main.rs
```diff
@@ -128,16 +149,19 @@ fn main() -> ! {
...
 
-    let data = mutable_statics::mutable_statics! {
-        static mut SENSOR_DATA: [Reading; NUM_SENSORS] = [|| Reading::Absent; _];
-    };
-
-    let nerrors = mutable_statics::mutable_statics! {
-        static mut SENSOR_NERRORS: [u32; NUM_SENSORS] = [|| 0; _];
+    let (has_data, data_value, data_time, nodata, nerrors) = mutable_statics::mutable_statics! {
+        static mut HAS_DATA: [bool; NUM_SENSORS] = [|| false; _];
+        static mut DATA_VALUE: [f32; NUM_SENSORS] = [|| f32::NAN; _];
```
cbiffle, line 154: I admit I was briefly considering using `is_nan` to remove the need for your bool array.... ;-)

cbiffle (commented):
If you're doing a series of I2C transactions, tens or hundreds of milliseconds can easily pass; I would recommend sampling the timer near when you make your reading in each case to avoid misstating timestamps. Because our I2C access is nondeterministic (everyone fights over RPC access to the driver) the time you _start_ trying to access I2C is less predictable, with respect to the sensor reading, then the time you _finish_, so concretely I'd suggest sampling the timer _after_ the operation returns. Since there's nothing in the i2c driver that would cause it to wait a long time to return to you, but there definitely are things that could cause it to wait before serving you.

task/sensor/src/main.rs
jgallagher: If we do that, do we need separate functions for "get last valid reading (if any)", "get last error (if any)", and "get last reading (if any, of either kind)"?

task/sensor/src/main.rs
```diff
@@ -14,7 +14,16 @@ use userlib::*;
 use task_sensor_api::config::NUM_SENSORS;
 
 struct ServerImpl {
-    data: &'static mut [Reading; NUM_SENSORS],
+    // We're using structure-of-arrays packing here because otherwise padding
+    // eats up a considerable amount of RAM; for example, Sidecar goes from 2868
+    // to 4200 bytes of RAM!
```
jgallagher, line 19: Would it be worth hiding these behind a type that provides a getter (returning an `Option<(f32, u64)>`) / setter (taking a `f32, u64`)?

task/power/src/main.rs
mkeeter: Done!

task/sensor-api/src/lib.rs
mkeeter: Done!

task/sensor/src/main.rs
mkeeter: I decided to change this in two ways:

- Adding timestamps to the error readings (and not clearing the error readings when writing data)
- changing `has_data` to an `enum` representing whether the `data` or `err` arrays were last written

`get()` and `get_reading()` still have the semantics of "get last reading (of either kind)", chosen by this enum.

I'd vote to add functions with the other semantics once they're needed (in MGS work or elsewhere).

task/sensor/src/main.rs
mkeeter: I ended up with a slightly more complex structure-of-arrays, but I don't think getters/setter would make it more clear.  Take a look and let me know what you think.

jgallagher: I think the slightly-more-complex version is actually clearer than what was there before; I agree that I don't think getters/setters would help anymore.

task/sensor/src/main.rs
mkeeter: Same, but I've now swapped stuff around to track the "most recent" reading (whether it's an error or data value), in 9cdbb2503ef75b495ffc0e576f04190106986ffa

jgallagher (approved):

PR #1029, opened 2022-12-15
xtask: fix 'build kernel'

cbiffle:
When building just the kernel, we wind up with a default bogus memory map for the tasks. This would cause an arithmetic underflow when the "find highest address" loop produced 0.

Since we don't care about the contents of the header in the build-kernel case, this substitutes a placeholder.


labbott (approved):
LGTM

PR #1030, opened 2022-12-16
kern: record failure message in predictable place.

cbiffle:
This should make kernel crashes and panics easier to inspect using a debugger, by recording state in predictably-named statics. This replaces the formerly-configurable panic handler, which in practice was always set to itm, and which in practice was rarely useful because of inconsistent ITM support. This approach has the distinct advantage of working on systems without ITM, like M0s.


cbiffle:
To save time wading through all the TOML deletions, the actual code starts here: https://github.com/oxidecomputer/hubris/pull/1030/files#diff-6d1571708c07b0c842b1c0d0682beeca0b0aa63ebefc69e57be793e36468b6f7

app/donglet/app-g030.toml
```diff
@@ -6,8 +6,8 @@ board = "donglet-g030"
 
 [kernel]
 name = "app-donglet"
-requires = {flash = 11702, ram = 1536}
-features = ["g030", "panic-halt"]
+requires = {flash = 18000, ram = 1328}
```
labbott, line 9: Intentional change?

sys/kern/src/fail.rs
```diff
@@ -0,0 +1,94 @@
...
+impl Write for Eulogist {
+    fn write_str(&mut self, s: &str) -> core::fmt::Result {
+        let s = s.as_bytes();
+        let n = s.len().min(self.dest.len());
+        let (dest, leftovers) = {
+            let taken = core::mem::replace(&mut self.dest, &mut []);
+            taken.split_at_mut(n)
+        };
+        dest.copy_from_slice(&s[..n]);
+        self.dest = leftovers;
```
andrewjstone, line 86: Is the purpose of this line just to give self.dest something to hold onto when it goes out of scope?

andrewjstone (approved):

app/donglet/app-g030.toml
cbiffle: (Answered this in chat, but, yes, the g0 was using `panic-halt` before, so it takes a flash hit to bring in formatting code.)

sys/kern/src/fail.rs
cbiffle: So on line 82 we replace self.dest with an empty slice. Line 86 fixes that by setting the leftovers, so that the next call to write_str will write into the leftovers.

Why go to all this trouble? Because `self.dest.split_at_mut(n)` produces two slices with the lifetime of self, since they alias `self.dest` and thus `self`. This means we can't do this:

```rust
let (dest, leftovers) = self.dest.split_at_mut(n);
self.dest = dest;
```
...because `dest` has lifetime `'static` and `self` does not. By moving it _out of_ self.dest and temporarily substituting self.dest with a placeholder, we unlink its lifetime from that of self.

This pattern probably has a name somewhere, but, you'll find it in several places in the Hubris kernel.

PR #1031, opened 2022-12-16
Be consistent with specifying images in linker script

labbott:
It turns out we have two disjoint ways of specifying approximately the same thing w.r.t. other images. Switch to a single method.

### DIFF
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -450,20 +450,6 @@ impl Config {
         Ok(memories)
     }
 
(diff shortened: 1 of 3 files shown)


labbott:
What this actually means is that each section of flash specified in the TOML gets a definition for us to do as we want. IT also makes the `external-images` field redundant. I think I had some sort of over engineered idea of how `external-images` would work but I think this is just as clean.

andrewjstone (approved):
LGTM!

PR #1035, opened 2022-12-19
Fix subtraction order

mkeeter:
Previously, this subtraction would panic if we were in the `Overheating` state and missed a sample.  I don't think this is what we're seeing in the field, but it's definitely incorrect, so let's fix it.


cbiffle (approved):
Nice catch.

kc8apf (approved):

PR #1036, opened 2022-12-20
Enums from RDL generated in generator

nathanaelhuffman:
There are a lot changes to the upstream tools etc to get this in place, so I'd like to review the changes here and iterate if necessary.

This uses num-derive per @cbiffle's suggestion and removes the hand-written enum for state decoding now that it is generated. We can change most everything here, so feedback appreciated!
This works fine with the legacy maps too, since the encoding is optional.

I've also not been a serious consumer of the FPGA registers API so feedback from those that have would be appreciated.


nathanaelhuffman:
This generates something that looks like this if you want to see the code without digging through the target/ directory:

```
    #[allow(non_snake_case)]
     pub mod A0SMSTATUS {
        #[allow(dead_code)]
        #[allow(non_upper_case_globals)]
        pub const A0SM: u8 = 0b11111111;

        use num_derive::{ToPrimitive, FromPrimitive};
        #[derive(FromPrimitive, ToPrimitive)]
        #[allow(dead_code)]
        #[allow(non_camel_case_types, clippy::upper_case_acronyms)]
        pub enum Encoded {
            IDLE = 0x00,
            PBTN = 0x01,
            WAITSLP = 0x02,
            GROUPB1_EN = 0x03,
            GROUPB1_PG = 0x04,
            GROUPB2_EN = 0x05,
            GROUPB2_PG = 0x06,
            GROUPC_PG = 0x07,
            DELAY_1MS = 0x08,
            ASSERT_PG = 0x09,
            WAIT_PWROK = 0x0a,
            WAIT_RESET_L = 0x0b,
            DONE = 0x0c,
            SAFE_DISABLE = 0x0d,
        }
    }
```

arjenroodselaar:
I have the same comment here as in https://github.com/oxidecomputer/cobalt/pull/47; let's leave the enum strings as specified by the design in the RDL. That'll leave it up to them whether no not to use camel case or upper snake case. I do suggest we leave the `#[allow(non_camel_case_types, clippy::upper_case_acronyms)]`.

@mkeeter do you think we could/should go directly to enum for these registers? Or maybe change the `pub enum Encoded` into `pub enum Values` so you can write something like `let foo = Reg::FOO::Values::SomeValue`?

nathanaelhuffman:
> I have the same comment here as in [oxidecomputer/cobalt#47](https://github.com/oxidecomputer/cobalt/pull/47); let's leave the enum strings as specified by the design in the RDL. That'll leave it up to them whether no not to use camel case or upper snake case. I do suggest we leave the `#[allow(non_camel_case_types, clippy::upper_case_acronyms)]`.

On the Json side, these are currently passed through directly without case-changing them or otherwise messing with them.

PR #1037, opened 2022-12-20
jefe: rework fault handling and command impl

cbiffle:
I happened to read through this code while preparing it for fault logging, and I noticed some things.

1. The dispositions were kind of redundant. If you set one to Fault the task effectively got held (in that it would not restart), but also there's Hold that causes it to get held?

2. As a side effect, issuing the "start" command would set the disposition to Start, which is a weird disposition that ... causes the task not to get restarted after faults?

3. The fault handling loop was complex.

4. There was no convenient way to restart a running task.

5. There was no way to atomically restart a task but catch it at the next fault, because the "start" and "release" commands both cleared the fault hold state.

It turns out these were all related to each other. The "External Command" facility, used by debuggers to poke the supervisor, was trying to communicate with the main logic _exclusively_ through the Disposition table. This includes achieving side effects like restarting a task. So, the Disposition enum wound up containing cases that aren't really restart-logic dispositions (like Start and Fault), with strange results.

This commit reduces the Disposition enum to two cases: Restart (the default) and Hold (configurable or on request). Tasks with dispo Restart will get restarted on faults (after it's logged); tasks with dispo Hold will not.

I've simplified the external command implementation by just having it produce the side effects it wants -- it'll deliver faults to tasks, or restart them, or whatever, only changing the disposition table if it actually wants to change the disposition.

I think this makes the code easier to follow, and by pulling the external command side effects out of the fault-handling main supervisor loop, that gets simpler too. This knocks 272 bytes of text off on ARMv6-M, just to sweeten the deal.

### DIFF
--- a/task/jefe/src/external.rs
+++ b/task/jefe/src/external.rs
@@ -43,7 +43,7 @@
 //! sands...
 //!
 
-use crate::Disposition;
+use crate::{Disposition, TaskStatus};
 use core::sync::atomic::{AtomicU32, Ordering};
 
 #[cfg(armv6m)]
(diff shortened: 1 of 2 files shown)


bcantrill (approved):
This looks good! Note primarily to myself: the docs for `humility jefe` need to be updated.  This is already true (they have the old option for hold, for example) -- but now especially true with this change.

PR #1038, opened 2022-12-20
add Validate.selected_mux_segment()

bcantrill:
In debugging the rack, one of the things we were missing was a way to get the last mux/segment over Idol. This adds an Idol interface to Validate that take a device index, and returns the last mux/segmented selected on the bus that the specified device is on.


task/validate/src/main.rs
```diff
@@ -50,6 +50,33 @@ impl idl::InOrderValidateImpl for ServerImpl {
...
+        use i2c_config::devices::{lookup_controller, lookup_port};
+
+        let c = lookup_controller(index).ok_or(ValidateError::InvalidDevice)?;
+        let p = lookup_port(index).ok_or(ValidateError::InvalidDevice)?;
+
+        let task = I2C.get_task_id();
+        let device = drv_i2c_api::I2cDevice::new(task, c, p, None, 0);
+
+        match device.selected_mux_segment() {
+            Ok(None) => Ok([0xff, 0xff]),
```
jgallagher, line 72: I don't object to using `0xff` as sentinel values here, but if you want to use `encoding: Hubpack`, we could make this return `Result<Option<[u8; 2]>, _>` to avoid it.

jgallagher (approved):

task/validate/src/main.rs
bcantrill: Fixed -- and then some!

task/validate-api/src/lib.rs
```diff
@@ -45,6 +47,12 @@ pub enum ValidateOk {
     Removed = 3,
 }
 
+#[derive(Copy, Clone, Debug, SerializedSize, Serialize, Deserialize)]
+pub struct MuxSegment {
+    pub mux: drv_i2c_api::Mux,
+    pub segment: drv_i2c_api::Segment,
```
jgallagher, line 53: Not strictly necessary since all these crates live in our workspace, but I'd publicly re-export these types so a client of `validate-api` doesn't have to also explicitly depend on `drv-i2c-api` to dig into a `MuxSegment`.

jgallagher (approved):

PR #1040, opened 2022-12-22
Add `Power.pmbus_read` idol operation

jgallagher:
The bulk of this PR adds the `Power.pmbus_read()` operation for remote use via `humility rpc` during compliance, and the 50+ variant enum describing all the ACAN-114 pmbus commands we want to read. This command (for now?) only supports the MWOCP68.

There are a handful of other changes included as a part of this work:

1. Adds gimletlet to CI and fixes its build
2. Removes the `NReplyMismatch` error from `udprpc`: for idol calls with `encoding: Hubpack` or `encoding: Ssmarshal`, the reply size is expected to be _at most_ the reply size requested by `humility rpc`, but is not guaranteed to be exactly that value. Removing this error allows smaller replies to come through okay.
3. The `power` task is now an idol server, so its previous main loop (sleep 1 second, then read data and post it to the `sensors` task) lives inside the idol server notification system.

### DIFF
--- a/task/power/src/main.rs
+++ b/task/power/src/main.rs
@@ -35,6 +35,9 @@ enum PowerState {
     A2,
 }
 
+const TIMER_MASK: u32 = 1 << 0;
+const TIMER_INTERVAL: u64 = 1000;
+
 task_slot!(I2C, i2c_driver);
 task_slot!(SENSOR, sensor);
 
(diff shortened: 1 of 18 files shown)


task/power/src/main.rs
```diff
@@ -140,6 +145,22 @@ impl Device {
...
+        op: task_power_api::Operation,
+    ) -> Result<PmbusValue, ResponseCode> {
+        let v = match &self {
+            Device::Mwocp68(dev) => dev.pmbus_read(op)?,
+            Device::Bmr491(_)
+            | Device::Raa229618(_)
+            | Device::Isl68224(_)
+            | Device::Tps546B24A(_)
+            | Device::Adm1272(_)
+            | Device::Max5970(_) => return Err(ResponseCode::NoDevice),
```
bcantrill, line 160: Do we want to return a distinct error code here to indicate that the operation is not supported rather than the device is not present?

bcantrill (approved):
Looks great!

task/power/src/main.rs
jgallagher: Good call - added in 4ec47222

PR #1041, opened 2022-12-22
control-plane-agent: Fill in new `SpState` with VPD identity

jgallagher:


task/control-plane-agent/src/mgs_common.rs
```diff
@@ -33,30 +40,43 @@ impl MgsCommon {
...
+            model: [0; 32],
+            revision: id.revision,
+            hubris_archive_id: kipc::read_image_id().to_le_bytes(),
             version: update.current_version(),
             power_state,
             rot: rot_state(update.sprot_task()),
-        })
+        };
+
+        let n = usize::min(state.serial_number.len(), id.serial.len());
```
andrewjstone, line 73: It would be nice if we could somehow report that we are truncating the serial number.

andrewjstone (approved):
LGTM. Selfishly I want this to go in, because I'm sick of rebasing gateway-messages on old versions :)

task/control-plane-agent/src/mgs_common.rs
jgallagher: Hmm. How about instead I add a static assertion that we haven't? Both these field sizes are known at compile time.

jgallagher: Added these in 72fc2343

PR #1044, opened 2023-01-02
Fix incorrect comment

mkeeter:


task/power-api/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-//! API crate for the Gimlet Host Flash server.
+//! API crate for the Gimlet `power` server.
```
jgallagher, line 5: Should we remove "Gimlet" too, I think?

```suggestion
//! API crate for the `power` server.
```

jgallagher (approved):

PR #1052, opened 2023-01-04
Use sensors in thermal task

mkeeter:
This PR refactors the `thermal` task to read temperature data from the `sensors`
task.  It still **writes** (most of) the data itself, but then reads it back, to
improve support for cases where other tasks provide temperature data.

The `sequencer` task now provides data to `sensors` directly, rather than
passing values through `thermal`.

In addition, it adds new APIs to the sensors task: `post_now` and `nodata_now`
send a reading and the current timestamp.  These APIs are used to reduce code in
a few unrelated tasks, e.g. `power` and `sensor-polling`.

### DIFF
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -87,7 +87,7 @@ struct ServerImpl {
     sensor_api: Sensor,
 
     /// Thermal models are populated by the host
-    thermal_models: [Option<ThermalModel>; NUM_PORTS],
+    thermal_models: [Option<ThermalModel>; NUM_PORTS as usize],
 }
 
 #[derive(Copy, Clone)]
@@ -343,7 +343,7 @@ impl ServerImpl {
 
 // This must be in the same order as the `DYNAMIC_INPUTS` array in
 // `thermal/src/bsp/sidecar_ab.rs`
-const SENSOR_IDS: [SensorId; NUM_PORTS] = [
+const SENSOR_IDS: [SensorId; NUM_PORTS as usize] = [
     other_sensors::QSFP_XCVR0_TEMPERATURE_SENSOR,
     other_sensors::QSFP_XCVR1_TEMPERATURE_SENSOR,
     other_sensors::QSFP_XCVR2_TEMPERATURE_SENSOR,
@@ -607,7 +607,7 @@ fn main() -> ! {
             leds_initialized: false,
             thermal_api,
             sensor_api,
-            thermal_models: [None; NUM_PORTS],
+            thermal_models: [None; NUM_PORTS as usize],
         };
 
         ringbuf_entry!(Trace::LEDInit);
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -87,7 +87,7 @@ struct ServerImpl {
     sensor_api: task_sensor_api::Sensor,
 
     /// Thermal models are populated by the host
-    thermal_models: [Option<ThermalModel>; 32],
+    thermal_models: [Option<ThermalModel>; NUM_PORTS],
 }
 
 #[derive(Copy, Clone)]
@@ -343,7 +343,7 @@ impl ServerImpl {
 
 // This must be in the same order as the `DYNAMIC_INPUTS` array in
 // `thermal/src/bsp/sidecar_ab.rs`
-const SENSOR_IDS: [SensorId; 32] = [
+const SENSOR_IDS: [SensorId; NUM_PORTS] = [
     other_sensors::QSFP_XCVR0_TEMPERATURE_SENSOR,
     other_sensors::QSFP_XCVR1_TEMPERATURE_SENSOR,
     other_sensors::QSFP_XCVR2_TEMPERATURE_SENSOR,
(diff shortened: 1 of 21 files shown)


task/thermal/src/control.rs
```diff
@@ -402,20 +406,18 @@ enum ControlResult {
...
-    fn write_temperature(
-        &mut self,
-        index: usize,
-        time_ms: u64,
-        value: Celsius,
-    ) {
+    fn write_temperature(&mut self, index: usize, reading: Reading) {
+        let r = TemperatureReading::Valid {
+            time_ms: reading.timestamp,
+            value: Celsius(reading.value),
```
jgallagher, line 412: Not related to your changes, but - should `reading.value` be typed instead of a raw `f32`?

task/thermal/src/control.rs
```diff
@@ -658,13 +636,52 @@ impl<'a> ThermalControl<'a> {
...
+        }
+        for (i, sensor_id) in self
+            .bsp
+            .dynamic_inputs
+            .iter()
+            .enumerate()
+            .filter(|(i, _s)| self.dynamic_inputs[*i].is_some())
+        {
+            if let Ok(r) = self.sensor_api.get_reading(*sensor_id) {
+                self.state.write_temperature(i + self.bsp.inputs.len(), r);
```
jgallagher, line 676: Do we ever `write_temperature_inactive` for dynamic inputs?

drv/transceivers-server/src/main.rs
```diff
@@ -319,6 +341,43 @@ impl ServerImpl {
     }
 }
 
+// This must be in the same order as the `DYNAMIC_INPUTS` array in
+// `thermal/src/bsp/sidecar_ab.rs`
```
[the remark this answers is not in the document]
mkeeter: Good idea, done in 346dd4917c027dde2b8de4e70a1a0a92c8ac333e

drv/transceivers-server/src/main.rs
```diff
@@ -298,11 +321,10 @@ impl ServerImpl {
             match temperature {
                 Ok(t) => {
                     // We got a temperature! Send it over to the thermal task
-                    if let Err(e) = self
-                        .thermal_api
-                        .update_dynamic_input(i, now, m.model, t)
+                    if let Err(e) =
+                        self.sensor_api.post(SENSOR_IDS[i], t.0, now)
```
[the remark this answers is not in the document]
mkeeter: Oversight, fixed in 80014f656dcbca4186d63b8c117fe67cc55c54dd

task/thermal/src/control.rs
mkeeter: Not in this case – `Reading` is used by every sensor type, so it's a raw value (and you need to know which sensor populated it to know the units).

task/thermal/src/control.rs
mkeeter: Not in the current code.  It's a good question, though; I wonder whether calling `write_temperature_inactive` in cases where a dynamic input is missing (i.e. has no model in `self.dynamic_inputs`) would make the code more clear...

mkeeter: Okay, this makes things a little cleaner (see 7fe465c616cf2b1e9d627e2f2b4495187b6b6273, although it also includes other changes).

In particular, we now don't have to special-case `None` sensors when going from `Boot` to `Running`!

task/control-plane-agent/src/inventory.rs
```diff
@@ -419,6 +419,8 @@ impl From<SensorErrorConvert> for MeasurementError {
             SensorError::DeviceUnavailable => Self::DeviceUnavailable,
             SensorError::DeviceTimeout => Self::DeviceTimeout,
             SensorError::DeviceOff => Self::DeviceOff,
+
+            SensorError::ServerDied => panic!(),
```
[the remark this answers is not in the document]
jgallagher: Hm, I think we've mostly gone by the "handwavey-idempotent" definition, or perhaps an even looser definition (just "harmless to retry" and not "no side effects")? E.g., `Jefe.get_state` is idempotent but its return value may change between calls; most of the functions in `stm32xx-sys.idol` are idempotent even though they have side effects (`Sys.enable_clock_raw`, `Sys.enter_reset_raw`, ...).

If we _don't_ mark `.get_reading` as idempotent, then control-plane-agent really ought to retry if it gets back `ServerDied`, but that's exactly what the effect of marking it idempotent would be.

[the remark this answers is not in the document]
mkeeter: Okay, you've persuaded me 😄 

Marked `Sensor.get` and `Sensor.get_reading` as idempotent in fd1cd496d3df501fffdb3d698a6f3d2811e26f2a

task/sensor-polling/src/main.rs
```diff
@@ -91,13 +89,12 @@ impl TemperatureSensor {
                 for (i, &s) in self.speed_sensors.iter().enumerate() {
                     let m = Mwocp68::new(&dev, i.try_into().unwrap());
                     let r = m.read_speed();
```
jgallagher, line 91: Github actions points out `r` is unused

jgallagher (approved):

Issue #1004, opened 2022-12-08
May want to move the SP/RoT image header


cbiffle:
Currently, we deposit an image header near the start of flash -- specifically, right after the vector table, since the CPU has opinions about where the vector table lives.

This has some drawbacks:

- It's useful to treat the image header as a "commit" point for writing an image, meaning that the update process may want to write the image header as the final sector written. Our current approach doesn't ensure that the image header is page-aligned, and requires any update logic to have intimate knowledge of the size of the vector table (which varies on every ARM implementation, meaning, the lpc55 and stm32h7 have very different header addresses).
- We may want to be able to conveniently extract the image header from a file on disk from generic tools. Currently this requires knowing which _processor variant_ the image is intended for. (Or heuristically scanning for things that look like a header). Such tools would be simplified if the header were in a predictable location.

I suggest that the _last page in flash_ is the right place. Our A:B partition layouts are fixed size for a part, so we can find the last page unambiguously. It's also easy to find the last block in a file on disk, for tooling. Finally, the last page in the flash address space is the last one written by a naive sequential update algorithm, removing special knowledge from updating tools.

We should be able to generate images with the header in the last sector as follows:
- Deduct one flash programming page from the FLASH `MEMORY` definition of the linker script for the part.
- Create an additional `MEMORY` definition containing just that one page, call it (say) `HEADER`
- Edit the existing linker script fragment that assigns image headers to right after the vector table, to instead put them in the new reserved area, by switching `>FLASH` to `>HEADER`.

This will produce an ELF file with a "hole" between the firmware and the header, which is probably what we want for right now -- in particular, this keeps SWD tools from flashing the unused space between the two, wasting programmer time during iteration. Eventually we want the hole to contain a predictable bit pattern for attestation reasons, and we probably want to do that with a post-processing tool either way; this is not a new concern from this bug, so I won't go into it further here.

@andrewjstone @lzrd @labbott @flihp

lzrd:
Is there any reason to have a compatibility phase where both the vector table header and the header trailer exists?
I'm fine with the answer "No".

lzrd:
An update that is destined to fail (wrong SKU, invalid epoch, linked for wrong bank, etc.) should be rejected by the target without erasing anything in the destination bank. That helps prevent excessive wear due to a misunderstanding with an upper layer that is retrying with an invalid image. That argues for showing a copy of the header to the target device at the start of the update process.
[...]

PR #1057, opened 2023-01-05
Write header block last on lpc55

andrewjstone:
Zero out the header block in flash, cache it in the lpc55-update-server, then write it during update finish.

This strategy prevents boot loops during partial RoT updates by invalidating the header magic.

### DIFF
--- a/drv/lpc55-update-server/src/main.rs
+++ b/drv/lpc55-update-server/src/main.rs
@@ -36,6 +36,7 @@ enum UpdateState {
 }
 
 struct ServerImpl {
+    block0: Option<[u8; BLOCK_SIZE_BYTES]>,
     state: UpdateState,
     image: Option<UpdateTarget>,
 }
@@ -43,6 +44,7 @@ struct ServerImpl {
 const BLOCK_SIZE_BYTES: usize = FLASH_PAGE_SIZE;
 
 const MAX_LEASE: usize = 1024;
+const HEADER_BLOCK: usize = 0;
 
 impl idl::InOrderUpdateImpl for ServerImpl {
     fn prep_image_update(
@@ -110,34 +112,27 @@ impl idl::InOrderUpdateImpl for ServerImpl {
             return Err(UpdateError::BadLength.into());
         }
 
-        let mut flash_page: [u8; BLOCK_SIZE_BYTES] = [0; BLOCK_SIZE_BYTES];
+        let mut flash_page = [0u8; BLOCK_SIZE_BYTES];
 
-        block
-            .read_range(0..len as usize, &mut flash_page)
-            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+        if block_num == HEADER_BLOCK {
+            self.block0 = Some([0u8; BLOCK_SIZE_BYTES]);
+            let block0 = &mut self.block0.as_mut().unwrap_lite()[..];
+            block
+                .read_range(0..len as usize, block0)
+                .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
 
-        flash_page[len..].fill(0);
+            block0[len..].fill(0);
+        } else {
+            block
+                .read_range(0..len as usize, &mut flash_page)
+                .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
 
-        let img = self.image.unwrap_lite();
-
-        let result = unsafe {
-            // The write_to_flash API takes raw pointers due to TrustZone
-            // ABI requirements which makes this function unsafe.
-            tz_table!().write_to_flash(
-                img,
-                block_num as u32,
-                flash_page.as_mut_ptr(),
-            )
-        };
-
-        match result {
-            HypoStatus::Success => Ok(()),
-            HypoStatus::OutOfBounds => Err(UpdateError::OutOfBounds.into()),
-            HypoStatus::RunningImage => Err(UpdateError::RunningImage.into()),
-            // Should probably encode the LPC55 flash status into the update
-            // error for good measure but that takes effort...
-            HypoStatus::FlashError(_) => Err(UpdateError::FlashError.into()),
+            flash_page[len..].fill(0);
         }
+
+        do_block_write(self.image.unwrap_lite(), block_num, &mut flash_page)?;
+
+        Ok(())
     }
 
     fn finish_image_update(
@@ -153,6 +148,16 @@ impl idl::InOrderUpdateImpl for ServerImpl {
             }
             UpdateState::InProgress => (),
         }
+
+        if self.block0.is_none() {
+            return Err(UpdateError::MissingHeaderBlock.into());
+        }
+
+        do_block_write(
+            self.image.unwrap_lite(),
+            HEADER_BLOCK,
+            self.block0.as_mut().unwrap_lite(),
+        )?;
 
         self.state = UpdateState::Finished;
         self.image = None;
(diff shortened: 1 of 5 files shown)


drv/lpc55-update-server/src/main.rs
```diff
@@ -110,34 +112,27 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
-            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+        if block_num == HEADER_BLOCK {
+            self.block0 = Some([0u8; BLOCK_SIZE_BYTES]);
+            let block0 = &mut self.block0.as_mut().unwrap_lite()[..];
+            block
+                .read_range(0..len as usize, block0)
+                .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
 
-        flash_page[len..].fill(0);
+            block0[len..].fill(0);
```
lzrd, line 124: Now we should do a sanity check on the contents.
The correct MAGIC should be there.
There should be a SKU and bank indication.
The vector table and entry point should be within the range of blocks being written.

lzrd: Doing the sanity check in the prepare call would leave the bank intact; an argument for including the header as a parameter to the prepare call.
If the prepare call does no bank erase, and the header containing block is the first we see, then we're still ok.

drv/lpc55-update-server/src/main.rs
```diff
@@ -36,13 +36,15 @@ enum UpdateState {
...
 struct ServerImpl {
+    block0: Option<[u8; BLOCK_SIZE_BYTES]>,
     state: UpdateState,
     image: Option<UpdateTarget>,
 }
 
 const BLOCK_SIZE_BYTES: usize = FLASH_PAGE_SIZE;
 
 const MAX_LEASE: usize = 1024;
+const HEADER_BLOCK: usize = 0;
```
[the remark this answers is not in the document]
andrewjstone: This is short term fix for our immediate problem. I expect we'll implement https://github.com/oxidecomputer/hubris/issues/1004 before we ship. If not, for some reason, we will install an update that changes things, then install another update I suspect.

drv/lpc55-update-server/src/main.rs
```diff
@@ -110,34 +112,27 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
+            self.block0 = Some([0u8; BLOCK_SIZE_BYTES]);
+            let block0 = &mut self.block0.as_mut().unwrap_lite()[..];
+            block
+                .read_range(0..len as usize, block0)
+                .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
 
-        flash_page[len..].fill(0);
+            block0[len..].fill(0);
+        } else {
+            block
```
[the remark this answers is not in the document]
andrewjstone: Agreed. I will add one.

drv/lpc55-update-server/src/main.rs
andrewjstone: Do we have a SKU? 

I can add the contents sanity check to look for the MAGIC and other things.  I don't want to change the API in this PR, because there are other PRs in flight that use it and I'd prefer not to have this interfere. I agree we should revisit though.

drv/lpc55-update-server/src/main.rs
```diff
@@ -36,13 +36,15 @@ enum UpdateState {
 }
 
 struct ServerImpl {
+    block0: Option<[u8; BLOCK_SIZE_BYTES]>,
```
labbott, line 39: This is only 512 bytes more but how close are we to running out of stack space?

drv/lpc55-update-server/src/main.rs
```diff
@@ -110,34 +112,27 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
 
-        let mut flash_page: [u8; BLOCK_SIZE_BYTES] = [0; BLOCK_SIZE_BYTES];
+        let mut flash_page = [0u8; BLOCK_SIZE_BYTES];
 
-        block
-            .read_range(0..len as usize, &mut flash_page)
-            .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
+        if block_num == HEADER_BLOCK {
+            self.block0 = Some([0u8; BLOCK_SIZE_BYTES]);
+            let block0 = &mut self.block0.as_mut().unwrap_lite()[..];
```
labbott, line 119: This seems odd to assign the block and then immediately unwrap it, is this necessary?

andrewjstone: I agree that it looks ridiculous. I think if I don't do this though, the compiler may perform a stack copy resulting in double size. I've seen a stack overflow due to something similar before.

drv/lpc55-update-server/src/main.rs
andrewjstone: Great question! How can I tell?

labbott: `humility stackmargin`. For reference on my xpressoboard without this change
```
 4 update_server      0x20007e80       2048        968       1080
```
512 bytes would eat into a good bit of our margin. It's certainly not terrible but something to be mindful of.

andrewjstone: TIL! Thank you.

drv/lpc55-update-server/src/main.rs
lzrd: Agreed.

drv/lpc55-update-server/src/main.rs
andrewjstone: Ok, I will experiment and see if it makes a difference with stack margin if I do the copy before inserting into self.

drv/lpc55-update-server/src/main.rs
lzrd: This kind of concern is why I was thinking of writing the header containing block with MAGIG ^ 0xFFFF_FFFF.
In that case, stage0 could also(someday) recognize a disabled image but still compute its hash if needed.

drv/lpc55-update-server/src/main.rs
lzrd: So a separate boolean may be better than Option<>?

drv/lpc55-update-server/src/main.rs
andrewjstone: I agree that's the preferred approach that we should take in the future.

drv/lpc55-update-server/src/main.rs
andrewjstone: I think we can actually just do it with a map without the unwrap. Using a bool makes me feel icky.

drv/lpc55-update-server/src/main.rs
andrewjstone: Update on this. Moving the array into self did cause a stack overflow. I'm playing around still.

drv/lpc55-update-server/src/main.rs
andrewjstone: Update on this, copying the array into self did cause a stack overflow. I'm still playing around.

andrewjstone: Ah, there is an entry style API, which is exactly what we want. https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert

drv/lpc55-update-server/src/main.rs
andrewjstone: ID TASK                STACKBASE  STACKSIZE   MAXDEPTH     MARGIN
 2 update_server      0x20007ba0       2048       1552        496

drv/lpc55-update-server/src/main.rs
andrewjstone: fixed

drv/lpc55-update-server/src/main.rs
andrewjstone: I added a check for header magic and that the image for the right slot is being sent via an entry-point check in the header.

drv/lpc55-update-server/src/main.rs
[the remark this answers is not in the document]
andrewjstone: Fixed

drv/lpc55-update-server/src/main.rs
```diff
@@ -197,6 +225,44 @@ impl idl::InOrderUpdateImpl for ServerImpl {
     }
 }
 
+// Perform some sanity checking on the header block.
+fn validate_header_block(
+    target: UpdateTarget,
+    block: &[u8; BLOCK_SIZE_BYTES],
+) -> Result<(), UpdateError> {
+    let reset_vector = u32::from_le_bytes(
+        block[RESET_VECTOR_OFFSET..][..4].try_into().unwrap_lite(),
```
lzrd, line 234: nit: 4 is obvious and understandable and keeps the line short, but I don't like to see constants other than -1, 0, 1, 2 in code and sometimes not those either.

core::mem::size_of::<u32>()

lzrd (approved):

PR #1059, opened 2023-01-05
Fix panic cuased by taking lpc55_pac Peripheral twice w/ unwrap.

flihp:
Pass the peripheral to the dice module as a parameter instead.


flihp:
np, easy fix

PR #1072, opened 2023-01-09
When starting an LPC55 update, erase block 0, but do not write any data.

lzrd:
Some testing and research is needed, but here is a draft PR.

On the LPC55, stage0 handles the case where block 0 is erase, i.e. block 0 is not zero or any other pattern.

It's important to invalidate the header as the first step in updating the bank. However, it's the erase/write cycle that causes wear in most flash parts. LPC55's data sheet claims 10k erase/write cycles (IIRC). So, a pathological update fail/retry loop where each iteration takes 10 seconds could burn out the flash in just over a day.

If the first step is only erase, and the LPC55 will not erase an erased block, or we can test for that, or erase on top of erase doesn't cause wear, then not writing any data to block
zero would eliminate wear in this pathological case.

### DIFF
--- a/drv/lpc55-update-server/src/main.rs
+++ b/drv/lpc55-update-server/src/main.rs
@@ -143,6 +143,7 @@ impl idl::InOrderUpdateImpl for ServerImpl {
                 self.header_block = None;
                 return Err(e.into());
             }
+            do_block_erase(target, block_num)?;
         } else {
             // The header block is currently block 0. We should ensure
             // we've seen and cached it before proceeding with other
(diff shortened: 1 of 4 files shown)


drv/lpc55-update-server/src/main.rs
```diff
@@ -156,10 +157,8 @@ impl idl::InOrderUpdateImpl for ServerImpl {
                 .map_err(|_| RequestError::Fail(ClientError::WentAway))?;
 
             flash_page[len..].fill(0);
+            do_block_write(target, block_num, &mut flash_page)?;
```
andrewjstone, line 160: With this change, `flash_page` can move into the else clause.

drv/lpc55-update-server/src/main.rs
```diff
@@ -288,6 +287,26 @@ fn do_block_write(
...
+    img: UpdateTarget,
+    block_num: usize,
+) -> Result<(), UpdateError> {
+    let result = unsafe {
+        // The write_to_flash API takes raw pointers due to TrustZone
+        // ABI requirements which makes this function unsafe.
+        tz_table!().erase_flash(img, block_num as u32)
+    };
+
+    match result {
```
andrewjstone, line 300: Now that this is used in two places, we should probably `impl From<HypoStatus> for Result<(), UpdateError>`

PR #1074, opened 2023-01-10
Idol error types: represent server death (or be intentional with panics) 

jgallagher:
This PR updates to the latest idolatry, which requires error types used in non-idempotent operations to implement the (empty) marker trait `IHaveConsideredServerDeathWithThisErrorType`. `idol-derive-error` now adds this implementation for any error type that includes an `#[idol(server_death)]` variant, and the rest of the PR makes a pass through all error types used in non-idempotent operations that did not have an `#[idol(server_death)]` variant to either add one or manually implement `IHaveConsideredServerDeathWithThisErrorType` (indicating an intention to cause clients to panic if the server dies). In most cases I opted for adding an `#[idol(server_death)]` variant.

This is an annoying PR to review, and I suggest reading more carefully than the diff appears to warrant. The changes are repetitive, similar, and small, but that does not mean they're all correct! Please consider whether I made the right decision with respect to the options of:

1. Add a `server_death` variant
2. Manually implement `IHaveConsideredServerDeathWithThisErrorType` indicating we want the client to panic
3. Changing operations to idempotent to avoid the need for `IHaveConsideredServerDeathWithThisErrorType`


drv/ignition-api/src/lib.rs
```diff
@@ -52,6 +52,8 @@ pub enum IgnitionError {
     /// Indicates the given request conflicts with the Target system power
     /// state. Poll the Target state and retry if desired.
     RequestDiscarded,
+    #[idol(server_death)]
+    ServerRestarted,
```
mkeeter, line 56: We already have `ServerDied = 1` up at the top; should that be our `#[idol(server_death)]` flavor?

drv/lpc55-gpio-api/src/lib.rs
```diff
@@ -205,6 +205,9 @@ pub enum Value {
 #[repr(u32)]
 pub enum GpioError {
     BadArg = 2,
+
+    #[idol(server_death)]
+    ServerRestarted,
```
mkeeter, line 210: I'd vote for this to manually implement `IHaveConsideredServerDeathWithThisErrorType` or become idempotent (probably the latter?); GPIO operations really shouldn't fail.

mkeeter: In fact, this error appears to never be constructed!

mkeeter:
Formatting nit – I personally find it hard to parse something like
```rust
    SpiError = 1,
    #[idol(server_death)]
    ServerDied,
    ProxyError,
    BadChipId,
```

I'd suggest consistently putting `#[idol(server_death)]` at the end of the enumeration, with a blank line above it (to make it clear that it's binding to the line below)

```rust
    SpiError = 1,
    ProxyError,
    BadChipId,

    #[idol(server_death)]
    ServerDied,
```

drv/ignition-api/src/lib.rs
jgallagher: Almost certainly - nice catch!

drv/lpc55-gpio-api/src/lib.rs
jgallagher: I think the problem with `idempotent` on GPIO calls is `toggle()`, which does not seem like something you want an idol client to retry some unknown number of times. The STM32 GPIO idol interface marks all the operations except `toggle` as idempotent. I can do the same for `lpc55-pins` but will still need this error for `toggle`.

mkeeter: `toggle` appears to never actually construct an error:
```rust
    fn toggle(
        &mut self,
        _: &RecvMessage,
        pin: Pin,
    ) -> Result<(), RequestError<GpioError>> {
        let (port, pin) = gpio_port_pin_validate(pin);

        self.gpio.not[port].write(|w| unsafe { w.notp().bits(1 << pin) });
        Ok(())
    }
```
(which is why I could delete the members of `GpioError` and stuff still compiled)
 
I think this means we could return `Simple("()")`?

mkeeter: Ah right, I was thinking of `err: ServerDeath` instead.

This causes Idolatry to generate a function which returns `Result<(), idol_runtime::ServerDeath>`, where the latter is an empty marker:

```diff
diff --git a/drv/lpc55-gpio/src/main.rs b/drv/lpc55-gpio/src/main.rs
index a3c58ca9..0654b9fd 100644
--- a/drv/lpc55-gpio/src/main.rs
+++ b/drv/lpc55-gpio/src/main.rs
@@ -111,7 +111,7 @@ impl idl::InOrderPinsImpl for ServerImpl<'_> {
         &mut self,
         _: &RecvMessage,
         pin: Pin,
-    ) -> Result<(), RequestError<GpioError>> {
+    ) -> Result<(), RequestError<core::convert::Infallible>> {
         let (port, pin) = gpio_port_pin_validate(pin);
 
         self.gpio.not[port].write(|w| unsafe { w.notp().bits(1 << pin) });
diff --git a/idl/lpc55-pins.idol b/idl/lpc55-pins.idol
index 36b89e06..cdd3c646 100644
--- a/idl/lpc55-pins.idol
+++ b/idl/lpc55-pins.idol
@@ -46,7 +46,7 @@ Interface(
 		},
 		reply: Result(
 			ok: "()",
-			err: CLike("GpioError"),
+			err: ServerDeath,
 		)
 	),
 
diff --git a/task/hiffy/src/lpc55.rs b/task/hiffy/src/lpc55.rs
index 63c098fc..9164077e 100644
--- a/task/hiffy/src/lpc55.rs
+++ b/task/hiffy/src/lpc55.rs
@@ -273,7 +273,7 @@ fn gpio_toggle(
 
     match gpio.toggle(pin) {
         Ok(_) => Ok(0),
-        Err(err) => Err(Failure::FunctionError(err.into())),
+        Err(_) => panic!(), // only error possible is server death
     }
 }
```

drv/ignition-api/src/lib.rs
jgallagher: Done in d09d063

drv/lpc55-gpio-api/src/lib.rs
jgallagher: Good call; I forgot about `ServerDeath`. I made the change to lpc55 gpios in 4ea0f07 and stm32 gpios in 69a720d.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -283,18 +282,13 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
-        };
-
-        match rv {
-            Err(_) => Err(HfError::MuxFailed.into()),
-            Ok(_) => {
-                self.mux_state = state;
-                Ok(())
-            }
         }
+
```
mkeeter, line 289: `HfError::MuxFailed` is now never constructed and can be deleted!

drv/gimlet-hf-server/src/main.rs
```diff
@@ -315,18 +309,13 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
-        };
-
-        match rv {
-            Err(_) => Err(HfError::DevSelectFailed.into()),
-            Ok(_) => {
-                self.dev_state = state;
-                Ok(())
-            }
         }
+
```
mkeeter, line 316: Same for `HfError::DevSelectFailed`

task/hiffy/src/lpc55.rs
```diff
@@ -38,13 +38,13 @@ pub enum Functions {
     SendLeaseRead((Task, u16, Buffer, usize, usize), u32),
     SendLeaseWrite((Task, u16, Buffer, usize, usize), u32),
     #[cfg(feature = "gpio")]
-    GpioInput(drv_lpc55_gpio_api::Pin, drv_lpc55_gpio_api::GpioError),
+    GpioInput(drv_lpc55_gpio_api::Pin, u32),
```
mkeeter, line 41: Does this break `humility gpio`?

mkeeter (approved):

PR #1082, opened 2023-01-12
Rename and rework secure rot-carrier build.

cbiffle:
Currently, the rot-carrier/stage0-fixed-rom bootloader builds an image that isn't correctly configured for trustzone and won't actually boot on an LPC55S69.

This commit fixes that by harmonizing its configuration with the (working) bootloader from app/lpc55xpresso.

While debugging this, I noticed that
- rot-carrier nonsecure build produced a file called rot-carrier
- rot-carrier secure build also produced a file called rot-carrier
- rot-carrier stage0 also produced a file called rot-carrier

This seemed really error prone so I've renamed them all to be distinct.

I've also renamed "fixed-rom" to "secure" to more accurately reflect its purpose.

### DIFF
--- a/app/rot-carrier/app-secure.toml
+++ b/app/rot-carrier/app-secure.toml
@@ -1,10 +1,11 @@
-name = "rot-carrier"
+name = "rot-carrier-secure"
 target = "thumbv8m.main-none-eabihf"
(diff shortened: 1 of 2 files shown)


flihp (approved):
LGTM

PR #1096, opened 2023-01-20
Use named notifications everywhere

mkeeter:
This PR changes how we declare notifications in Hubris tasks to reduce the possibility of human error.

Every task now declares a `notifications = [...]` list (in the `app.toml`); these names are used in the `interrupts = { ... }` table and injected into the code through a new `notifications` crate.

(The few tasks which need to know about _other_ task notifications – `net` and `jefe` – have also been adjusted to use this crate)

TODO:
- [x] Merge https://github.com/oxidecomputer/humility/pull/319
- [x] Update `test/*.toml`
- [ ] Test on real hardware
    - [x] PSC
    - [x] Gimlet
    - [x] Sidecar
    - [x] Gimletlet
    - [x] G0-Nucleo
    - [x] RoT


build/util/src/lib.rs
```diff
@@ -215,3 +235,52 @@ fn toml_from_env<T: DeserializeOwned>(var: &str) -> Result<Option<T>> {
...
+        bail!(
+            "Too many notifications; \
+             overlapping with `INTERNAL_TIMER_NOTIFICATION`"
+        );
+    }
+    if full_task_config.name == "task-jefe"
+        && full_task_config.notifications.get(0).cloned()
+            != Some("fault".to_string())
+    {
+        bail!("`jefe` must have \"fault\" as its first notification");
```
jgallagher, line 256: Is this for compatibility with `humility jefe`?

build/util/src/lib.rs
```diff
@@ -215,3 +235,52 @@ fn toml_from_env<T: DeserializeOwned>(var: &str) -> Result<Option<T>> {
...
+
+    write_task_notifications(&mut out, &full_task_config.notifications)?;
+
+    for task in env_var("HUBRIS_TASKS")
+        .expect("missing HUBRIS_TASKS")
+        .split(",")
+    {
+        let full_task_config = other_task_full_config_toml(task)?;
+        writeln!(&mut out, "pub mod {task} {{")?;
+        write_task_notifications(&mut out, &full_task_config.notifications)?;
```
jgallagher, line 270: Should we repeat the check for the max number of notifications here? If we add a task with too many and compile _that task_ we'll get the good error message above, but every other task will fail to compile with a less-useful message.

build/util/src/lib.rs
```diff
@@ -215,3 +235,52 @@ fn toml_from_env<T: DeserializeOwned>(var: &str) -> Result<Option<T>> {
...
 }
+
+pub fn build_notifications() -> Result<()> {
+    let out_dir = out_dir();
+    let dest_path = out_dir.join("notifications.rs");
+    let mut out = std::fs::File::create(dest_path)?;
+
+    let full_task_config = task_full_config_toml()?;
+
+    if full_task_config.notifications.len() >= 32 {
```
jgallagher, line 246: Should this be 31? The 32nd notification will collide with `INTERNAL_TIMER_NOTIFICATION`.

drv/stm32xx-i2c-server/src/main.rs
```diff
@@ -476,3 +476,5 @@ fn configure_muxes(
         }
     }
 }
+
+include!(concat!(env!("OUT_DIR"), "/notifications.rs"));
```
jgallagher, line 480: Are the notifications not used by name in this server? (Are there implicit assumptions about i2c notification numbers/masks somewhere else?)

task/net/src/bsp_support.rs
```diff
@@ -14,6 +14,7 @@
 //! A BSP module is expected to export a single type, called `BspImpl`, which
 //! implements the `Bsp` trait from this module.
 
+use crate::notifications;
```
jgallagher, line 17: Github actions points out this is unused in several build configs. Maybe move into the cfg_if below?

jgallagher (approved):
This is a really nice change, thanks!

build/util/src/lib.rs
```diff
@@ -215,3 +235,52 @@ fn toml_from_env<T: DeserializeOwned>(var: &str) -> Result<Option<T>> {
...
+    }
+    if full_task_config.name == "task-jefe"
+        && full_task_config.notifications.get(0).cloned()
+            != Some("fault".to_string())
+    {
+        bail!("`jefe` must have \"fault\" as its first notification");
+    }
+
+    writeln!(&mut out, "#[allow(dead_code)]")?;
+    writeln!(&mut out, "pub mod notifications {{")?;
```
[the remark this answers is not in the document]
mkeeter: Yeah, I've seen it done both ways, but don't have a strong preference one way or the other.

build/util/src/lib.rs
mkeeter: Nope, it's for compatibility with the kernel, which always posts faults to the supervisor's bit 0.

build/util/src/lib.rs
mkeeter: I think this is correct, because it's `>=` (so 32 notification exactly will trigger the conditional)

drv/stm32xx-i2c-server/src/main.rs
mkeeter: They're baked into the I2C configuration, e.g.
```rust
        [
            I2cController {
                controller: Controller::I2C2,
                peripheral: Peripheral::I2c2,
                notification: crate::notifications::I2C2_IRQ_MASK,
                registers: unsafe { &*device::I2C2::ptr() },
            },
        ]
```

build/util/src/lib.rs
jgallagher: Oy, yep, you're right. :man_facepalming:

task/net/src/bsp_support.rs
mkeeter: good catch, done

build/util/src/lib.rs
mkeeter: Good idea, I added a fallback check in 6b273242a936d5f3cc80102f25d62f00b5c18e23

(this will only matter for _other task IRQs_, so I'm just protecting against `u32` overflow, rather than collision with the default timer IRQ)

lib/toml-task/src/lib.rs
```diff
@@ -46,13 +49,29 @@ pub struct Task<T = ordered_toml::Value> {
...
     #[serde(default)]
     pub sections: IndexMap<String, String>,
     #[serde(default)]
     pub max_sizes: IndexMap<String, u32>,
 }
 
+impl<T> Task<T> {
+    pub fn notification_bit(&self, name: &str) -> Result<u8> {
+        match self.notifications.iter().position(|n| n == name) {
+            Some(i) => return Ok(i.try_into().unwrap()),
```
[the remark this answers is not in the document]
mkeeter: Sure, done in 1d04a48bc56a9bc3153b93415f98e13bf1f9354e (only checking that it fits in a `u32`, since "does it mask the `userlib::hl` timer notification" isn't really the kernel's problem)

PR #1103, opened 2023-01-24
Fail on oversized host flash update

### LINKED ISSUE #1098: control-plane-agent: immediately reject host phase1 updates larger than the flash capacity

mkeeter:
Depends on https://github.com/oxidecomputer/management-gateway-service/pull/53 ; we should switch the dependency back to `main` once that's merged.

Closes #1098


jgallagher (approved):
Thanks!

PR #1104, opened 2023-01-25
Plumb SPI features through tasks to improve rebuild

mkeeter:
Previously, Sidecar took 19 seconds to rebuild, because `drv-stm32h7-spi-server-core` was being

- instantiated more than once
- with identical features
- but dependent on environmental variables that are changing

This caused it to be rebuilt – and more importantly, the tasks using it to be re-linked – multiple times per full Hubris build.

Because Cargo caches builds based on features, we can fix this by activating a `spiX` feature in this crate, based on the SPI peripheral being used.

This requires a little plumbing to propagate the feature through other crates which depend on `drv-stm32h7-spi-server-core`.

This brings a Sidecar rebuild down to **7 seconds**, which is much more tolerable.

In addition, the feature is used to select the global SPI configuration block, eliminating the old `global_config`.  This couples the global configuration block name to SPI peripherals, which seems... fine?  We were already naming those blocks based on peripherals, and this removes a bunch of copy-pasta.

### DIFF
--- a/drv/stm32h7-spi-server-core/Cargo.toml
+++ b/drv/stm32h7-spi-server-core/Cargo.toml
@@ -32,7 +32,8 @@ build-util = { path = "../../build/util" }
 call_rustfmt = { path = "../../build/call_rustfmt" }
 
 [features]
-# These features are used to select a global SPI peripheral block; making this
+# These features are used in `build.rs` to select a global SPI peripheral block,
+# which modifies the generated code that's compiled in the crate.  Making this
 # choice based on features prevents extra rebuilds when building a complete
 # Hubris app.
 spi1 = []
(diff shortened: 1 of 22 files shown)


drv/stm32h7-spi-server-core/Cargo.toml
```diff
@@ -32,5 +32,14 @@ build-util = { path = "../../build/util" }
 call_rustfmt = { path = "../../build/call_rustfmt" }
 
 [features]
+# These features are used to select a global SPI peripheral block; making this
+# choice based on features prevents extra rebuilds when building a complete
```
cbiffle, line 36: Personally I'd make this longer by adding something like

`These features are *not* referenced anywhere in the code, they exist only at build time, but should not be removed without measuring impact on build times.`

cbiffle (approved):

drv/stm32h7-spi-server-core/Cargo.toml
mkeeter: Yup, I updated this to clarify that the features modify code-generation (since we're using them to select the global block), but aren't used by name anywhere in the code.

PR #1105, opened 2023-01-25
Provide direct control of transceiver low speed signals

Aaron-Hartwig:
This branch needs two dependent PRs to land first:
FPGA: https://github.com/oxidecomputer/quartz/pull/69
messages: https://github.com/oxidecomputer/transceiver-control/pull/54


drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -300,44 +301,148 @@ impl Transceivers {
...
+        mask: M,
+    ) -> Result<(), FpgaError> {
+        self.enable_power(mask)?;
+        self.assert_lpmode(mask)?;
+        self.deassert_reset(mask)?;
+        Ok(())
+    }
+
+    /// Sequence of actions to have modules enter high power mode per the
+    /// specified `mask
```
mkeeter, line 426:
```suggestion
    /// specified `mask`
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -300,44 +301,148 @@ impl Transceivers {
...
+        mask: M,
+    ) -> Result<(), FpgaError> {
+        self.assert_reset(mask)?;
+        userlib::hl::sleep_for(1);
+        self.deassert_reset(mask)?;
+        userlib::hl::sleep_for(2000);
+        Ok(())
+    }
+
+    /// Sequence of actions to turn off modules per the specified `mask
```
mkeeter, line 398:
```suggestion
    /// Sequence of actions to turn off modules per the specified `mask`
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -300,44 +301,148 @@ impl Transceivers {
...
+        // because it will leak into the main power rail if
+        // actively driven in the absence of power for at least
+        // some modules.
+        self.deassert_lpmode(mask)?;
+        self.assert_reset(mask)?;
+        Ok(())
+    }
+
+    /// Sequence of actions to have modules enter low power mode per the
+    /// specified `mask
```
mkeeter, line 414:
```suggestion
    /// specified `mask`
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -300,44 +301,148 @@ impl Transceivers {
...
+    }
+
     /// Get the current status of all low speed signals for all ports. This is
     /// Enable, Reset, LpMode/TxDis, Power Good, Power Good Timeout, Present,
     /// and IRQ/RxLos.
     pub fn get_modules_status(&self) -> Result<ModulesStatus, FpgaError> {
-        let f0: [U16<byteorder::BigEndian>; 8] =
-            self.fpga(FpgaController::Left).read(Addr::QSFP_CTRL_EN_H)?;
-        let f1: [U16<byteorder::BigEndian>; 8] = self
+        let f0: [U16<byteorder::LittleEndian>; 8] =
```
mkeeter, line 441: Was this just wrong before?

drv/transceivers-server/src/main.rs
```diff
@@ -503,11 +536,11 @@ impl NotificationHandler for ServerImpl {
...
 
-            if status.present != self.modules_present {
-                self.led_update(status.present);
+            if !status.modprsl != self.modules_present {
+                self.led_update(!status.modprsl);
 
-                self.modules_present = status.present;
-                ringbuf_entry!(Trace::ModulePresenceUpdate(status.present));
+                self.modules_present = !status.modprsl;
+                ringbuf_entry!(Trace::ModulePresenceUpdate(!status.modprsl));
```
mkeeter, line 543: How about assigning `!status.modprsl` to a local variable here, e.g.
```rust
let modules_present = !status.modprsl;
 if modules_present != self.modules_present {
    self.led_update(modules_present);

    self.modules_present = modules_present;
    ringbuf_entry!(Trace::ModulePresenceUpdate(modules_present));
 }
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -300,44 +301,148 @@ impl Transceivers {
         &self.fpgas[c as usize]
     }
 
-    // helper to write a value to CONTROL_PORTx registers given a mask
-    fn write_control_port<M: Into<FpgaPortMasks>>(
+    /// Executes a specified WriteOp (`op`) at `addr` for all ports specified by
+    /// the `mask`.
+    pub fn masked_port_op<M: Into<FpgaPortMasks>>(
```
mkeeter, line 306: Nit: if this function is only ever called here, then it could be (a) not `pub` and (b) take `FpgaPortmasks` directly, leaving the generic conversions to the user-facing APIs.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -300,44 +301,148 @@ impl Transceivers {
...
+    /// (t_reset_init), but the module will not be ready for normal operation
+    /// until 2 seconds (t_reset) after reset is released.
+    pub fn module_reset<M: Into<FpgaPortMasks> + Copy>(
+        &self,
+        mask: M,
+    ) -> Result<(), FpgaError> {
+        self.assert_reset(mask)?;
+        userlib::hl::sleep_for(1);
+        self.deassert_reset(mask)?;
+        userlib::hl::sleep_for(2000);
```
mkeeter, line 394: Hmmm, I'm not sure I love having this long wait on the Hubris side.

It makes it harder to misuse the API, but I'm worried about something else (e.g. network communication) hitting a timeout.

@bnaecker any opinions?  I'm wondering if we should put this on the host side instead.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -300,44 +301,148 @@ impl Transceivers {
...
+        &self,
+        mask: M,
+    ) -> Result<(), FpgaError> {
+        self.disable_power(mask)?;
+        // lpmode is being deasserted while the module is off to
+        // because it will leak into the main power rail if
+        // actively driven in the absence of power for at least
+        // some modules.
+        self.deassert_lpmode(mask)?;
+        self.assert_reset(mask)?;
```
mkeeter, line 409: What happens if one of these operations fails and we return early with the `Err`? Is there a danger of leaving the FPGA in an inconsistent state?

drv/sidecar-front-io/src/transceivers.rs
bnaecker: Doing that on the host side seems like a good idea. I know for sure that the Tofino SDE will do that manually, and we can always build a host-side method ourselves should we need it.

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: No, since I was making some invasive changes to the FPGA's register map I standardized all the multi-byte fields to be little endian as I had been inconsistent on that before.

drv/transceivers-server/src/main.rs
Aaron-Hartwig: :+1:

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: Good catch

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: I'm good with this as well. Admittedly I should have yanked this when we pulled out all the `PowerMode` stuff into the host, as those are higher order functions than the SP cares about anymore.

Given we now have `AssertReset` and `DeassertReset` messages, I'm going to remove the `Reset` message (the only caller of this function).

mkeeter (approved):

PR #1107, opened 2023-01-25
app.toml: Fix priority inversions in RoT app tomls.

flihp:
Most of these changes come from the `gpio_driver` and `swd` tasks operating with the same priority. `swd` talks  to `gpio_driver` so it must be at a lower (numerically higher) priority. This trickles down to other tasks that use `swd` like `dumper`, and tasks that use `dumper` like `sprot`.


cbiffle (approved):
Thank you for fixing this! The priority inversions are going to bite us if left unattended.

PR #1109, opened 2023-01-26
Add 'attest' task acting as the root of trust for reporting (RoT-R)

flihp:
This is a DRAFT. The 'attest' task is currently limited to recording measurements and dumping out the DICE certs that make up the cert chain starting from the attestation signing key (the DICE 'Alias' key) back to either the self signed identity cert, or the identity manufacturing intermediate cert.


flihp:
The associated humility command to poke this task is: https://github.com/oxidecomputer/humility/pull/324

PR #1110, opened 2023-01-26
Support stage0 update from MGS

andrewjstone:


task/control-plane-agent/src/update/rot.rs
```diff
@@ -72,15 +72,22 @@ impl ComponentUpdater for RotUpdate {
...
+                0 => UpdateTarget::ImageA,
+                1 => UpdateTarget::ImageB,
+                _ => return Err(SpError::InvalidSlotForComponent),
+            },
+            SpComponent::STAGE0 => match update.slot {
+                0 => UpdateTarget::Bootloader,
+                _ => return Err(SpError::InvalidSlotForComponent),
+            },
+            _ => unreachable!(),
         };
```
jgallagher, line 91: This might be a little cleaner if we match on the tuple instead of nesting matches:

```rust
        let target = match (update.component, update.slot) {
            (SpComponent::ROT, 0) => UpdateTarget::ImageA,
            (SpComponent::ROT, 1) => UpdateTarget::ImageB,
            (SpComponent::STAGE0, 0) => UpdateTarget::Bootloader,
            _ => return Err(SpError::InvalidSlotForComponent),
        };
```

This discards the unreachable case of "someone called us with a component other than ROT/STAGE0" and would return `InvalidSlotForComponent`. We could add a

```rust
(SpComponent::STAGE0 | SpComponent::ROT, _) => unreachable!()`
```

case if we think that's important? I could go either way on that.

jgallagher (approved):

PR #1113, opened 2023-01-27
host-sp-comms: Implement KeyLookup RPC

jgallagher:
The driving motivation for this RPC is to be able to pass installinator the image ID it should fetch, but this PR just implements the basic message with a ping/pong key. Adding and handling the `InstallinatorImageId` key will come in a subsequent PR that will include `control-plane-agent` changes to allow MGS to set it.

`KeyLookup` is exposed to userland via an ioctl; demo code from @citrus-it with a few annotations:

```c
        ipcc_keylookup_t kl;
     
        bzero(&kl, sizeof (kl));
        kl.ik_key = 0;  // "Ping"
        kl.ik_buflen = 1024; // must be at least 4 for Ping

        kl.ik_buf = calloc(kl.ik_buflen, 1);

        if (ioctl(ipcc_fd, IPCC_KEYLOOKUP, &kl) < 0)
                err(EXIT_FAILURE, "IPCC_KEYLOOKUP ioctl failed");

        // ik_result is the `KeyLookupResult` we send back
        // assuming a successful Ping call:
        // * `ik_datalen` has been set to 4
        // * `ik_buf[..4]` has been set to b"pong"
        (void) printf("Result: %u (length %u)\n", kl.ik_result, kl.ik_datalen);
        hexdump(kl.ik_buf, kl.ik_datalen);

        free(kl.ik_buf);
```


citrus-it:
Excerpt from he `ipcc(4I)` manual page:

image

task/host-sp-comms/src/main.rs
```diff
@@ -731,6 +748,39 @@ impl ServerImpl {
...
+
+                if max_response_len < PONG.len() {
+                    return Err(KeyLookupResult::MaxResponseLenTooShort);
+                }
+
+                self.tx_buf.encode_response(
+                    sequence,
+                    &SpToHost::KeyLookupResult(KeyLookupResult::Ok),
+                    |buf| {
+                        buf[..PONG.len()].copy_from_slice(PONG);
```
mkeeter, line 775: Not specific to this PR, but how do we handle cases where the encoded response is too large for the `tx_buf`?  Is it statically checked anywhere, or would it panic in `encode_response`?

jgallagher: It would panic in `encode_response`, yeah. The length of `buf` is statically guaranteed to be at least:

```
use host_sp_messages::{MAX_MESSAGE_SIZE, CHECKSUM_SIZE, Header, SpToHost};

MAX_MESSAGE_SIZE - Header::MAX_SIZE - CHECKSUM_SIZE - SpToHost::MAX_SIZE
```

which is a little less than 4 KiB currently. (The exact size only depends on which `SpToHost` variant we're in; `Header` always serializes to its `MAX_SIZE`.) Clients should either:

a) guarantee they're writing less than this amount of data (which `PONG` technically should, but it's omitted because it's only 4 bytes)
b) limit their writes whatever size `buf` actually is (which is what `get_host_phase2_data` does, the only other message type to date that appends data via this closure)

mkeeter (approved):

PR #1114, opened 2023-01-30
Switch to names for SPI devices

mkeeter:
This isn't as strongly-typed as possible, but it's an improvement over the status quo!

I pulled the `GlobalConfig` from `stm32h7-drv-spi-server-core` to a new `build/spi` crate, then reused it in `spi-api`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -229,6 +229,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "build-spi"
+version = "0.1.0"
+dependencies = [
+ "indexmap",
+ "proc-macro2",
+ "quote",
+ "serde",
+ "syn",
+]
+
+[[package]]
 name = "build-util"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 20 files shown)


jgallagher (approved):
> This isn't as strongly-typed as possible, but it's an improvement over the status quo!

The generated `spi_devices.rs` for sidecar/rev-b (as an example) looks like this:

```rust
pub mod devices {
    // spi1 (2 devices)
    pub const ECP5_FRONT_IO_FPGA: u8 = 0;
    pub const ECP5_FRONT_IO_USER_DESIGN: u8 = 1;
    // spi2 (1 devices)
    pub const VSC7448: u8 = 0;
    // spi3 (1 devices)
    pub const KSZ8463: u8 = 0;
    // spi4 (1 devices)
    pub const ROT: u8 = 0;
    // spi5 (2 devices)
    pub const ECP5_MAINBOARD_FPGA: u8 = 0;
    pub const ECP5_MAINBOARD_USER_DESIGN: u8 = 1;
}
```

Is the "not as strongly-typed as possible" because we can use a constant from the wrong SPI controller, or because we can still use raw `u8`s, or both? Regardless of the answer to that, agree this is a definite improvement!

mkeeter:
Both!  The ideal would be an `enum` that only contains valid devices for the selected SPI bus, but that requires plumbing a SPI bus feature through everything that uses `spi-api`, which seems very awkward...

PR #1115, opened 2023-01-30
Prefer static over const for device description arrays.

jgallagher:
We keep the const versions around (with `_CONST` appended to their names) for use by static assertions, but use the `static` variant in all real code.

### DIFF
--- a/task/control-plane-agent/src/inventory.rs
+++ b/task/control-plane-agent/src/inventory.rs
@@ -26,7 +26,7 @@ pub(crate) struct Inventory {
 
 impl Inventory {
     pub(crate) fn new() -> Self {
-        let () = ASSERT_EACH_DEVICE_FITS_IN_ONE_PACKET;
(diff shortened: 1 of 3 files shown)


task/control-plane-agent/src/inventory.rs
```diff
@@ -327,31 +336,30 @@ mod max_num_devices {
...
+        let encoded_len = tlv::tlv_len(
+            gateway_messages::DeviceDescriptionHeader::MAX_SIZE
+                + device.len()
+                + description.len(),
+        );
 
-    if encoded_len > MIN_TRAILING_DATA_LEN {
-        panic!(concat!(
+        if encoded_len > MIN_TRAILING_DATA_LEN {
+            panic!(concat!(
```
mkeeter, line 362: Want to also indent the inner text here?

mkeeter (approved):

task/control-plane-agent/src/inventory.rs
jgallagher: Nice catch, fixed in 664ccbcf

PR #1116, opened 2023-01-31
Use zero sized types for extern image locations

andrewjstone:
To more clearly indicate that IMAGE addresses should not be dereferenced in the lpc55-update-server, make them arrays of 0 length. This is based on current recommendations from unsafe code wg. Hat tip to @cbiffle for the suggestion.


cbiffle (approved):
Looks right to me! :+1:

PR #1117, opened 2023-02-01
PSC SP needs the VPD programming task for manufacturing

jclulow:
This is analogous to #1011 but for the PSC instead of Sidecar.  I just copied and pasted, here, what I did there; I'm not sure if these are correct -- especially the priority!


mkeeter:
Done in #1120

PR #1118, opened 2023-02-01
Add Installinator Image ID plumbing

jgallagher:
Adds:

* Implementations for the new `set_ipcc_key_lookup_value()` handler for incoming MGS messages
  * sidecar/psc return errors - they have no IPCC
  * gimlet's implementation allows MGS to set the installinator image ID to any binary blob of at most 512 bytes
* The `host_sp_messages::Key::InstallinatorImageId` variant
* An idol call allowing `host_sp_comms` to ask `control_plane_agent` for the current installinator image ID when a host process requests it via the IPCC ioctl described on #1113 

Todo:

- [ ] Restore the `main` branch reference after https://github.com/oxidecomputer/management-gateway-service/pull/58 is merged

### DIFF
--- a/idl/control-plane-agent.idol
+++ b/idl/control-plane-agent.idol
@@ -69,6 +69,14 @@ Interface(
             reply: Simple("UartClient"),
             idempotent: true,
         ),
+        "get_installinator_image_id": (
+            doc: "Get the current installinator image ID; writes 0 bytes if we have no ID.",
+            leases: {
+                "data": (type: "[u8]", write: true, max_len: Some(512)),
+            },
+            reply: Simple("usize"),
+            idempotent: true,
+        ),
         "set_humility_uart_client": (
             doc: "Attach or detach Humility as the host serial console client.",
             args: {
--- a/idl/control-plane-agent.idol
+++ b/idl/control-plane-agent.idol
@@ -70,7 +70,7 @@ Interface(
             idempotent: true,
         ),
         "get_installinator_image_id": (
-            doc: "Get the current installinator image ID; writes 0 bytes if we have no ID.",
+            doc: "Get the current installinator image ID, returning the number of bytes written. If we have no installinator image ID, does not modify the leased buffer and returns 0.",
             leases: {
                 "data": (type: "[u8]", write: true, max_len: Some(512)),
             },
--- a/task/control-plane-agent/src/mgs_gimlet.rs
+++ b/task/control-plane-agent/src/mgs_gimlet.rs
@@ -20,15 +20,16 @@ use gateway_messages::{
     PowerState, SpComponent, SpError, SpPort, SpRequest, SpState,
     SpUpdatePrepare, UpdateChunk, UpdateId, UpdateStatus,
 };
-use heapless::Deque;
+use heapless::{Deque, Vec};
 use host_sp_messages::HostStartupOptions;
 use idol_runtime::{Leased, RequestError};
 use ringbuf::ringbuf_entry_root as ringbuf_entry;
 use task_control_plane_agent_api::{
     ControlPlaneAgentError, UartClient, VpdIdentity,
+    MAX_INSTALLINATOR_IMAGE_ID_LEN,
 };
 use task_net_api::{Address, MacAddress, UdpMetadata};
-use userlib::{sys_get_timer, sys_irq_control, UnwrapLite};
+use userlib::{sys_get_timer, sys_irq_control, FromPrimitive, UnwrapLite};
 
 // We're included under a special `path` cfg from main.rs, which confuses rustc
 // about where our submodules live. Pass explicit paths to correct it.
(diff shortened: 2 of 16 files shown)


idl/control-plane-agent.idol
```diff
@@ -69,6 +69,14 @@ Interface(
             reply: Simple("UartClient"),
             idempotent: true,
         ),
+        "get_installinator_image_id": (
+            doc: "Get the current installinator image ID; writes 0 bytes if we have no ID.",
```
mkeeter, line 73: Nit: "writes 0 bytes" is confusing (is it writing no bytes, or bytes with the value 0?)

How about "does not modify the leased buffer" or "fills the buffer with 0s" respectively?

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -908,6 +917,46 @@ impl SpHandler for MgsHandler {
...
+        match Key::from_u8(key) {
+            Some(Key::InstallinatorImageId) => {
+                if value.len() > self.installinator_image_id.capacity() {
+                    return Err(SpError::SetIpccKeyLookupValueFailed(
+                        IpccKeyLookupValueError::ValueTooLong {
+                            max_len: self.installinator_image_id.capacity()
+                                as u16,
+                        },
+                    ));
+                }
```
mkeeter, line 945: Nit: maybe extract `let max_len = self.installinator_image_id.capacity()` before this conditional, which helps with the wacky indentation below?

```suggestion
                let max_len = installinator_image_id.capacity();
                if value.len() > max_len {
                    return Err(SpError::SetIpccKeyLookupValueFailed(
                        IpccKeyLookupValueError::ValueTooLong {
                            max_len: max_len as u16,
                        },
                    ));
                }
```

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -908,6 +917,46 @@ impl SpHandler for MgsHandler {
...
+
+                self.installinator_image_id.clear();
+                self.installinator_image_id
+                    .extend_from_slice(value)
+                    .unwrap_lite();
```
mkeeter, line 950: You could also remove the check above entirely and do something like
```
        self.installinator_image_id
            .extend_from_slice(value)
            .ok_or_else(|| {
                let max_len = self.installinator_image_id.capacity();
                SpError::SetIpccKeyLookupValueFailed(
                    IpccKeyLookupValueError::ValueTooLong {
                        max_len: max_len as u16,
                    },
                )
            })?;
```
(dunno if this passes the borrow checker; you might need to save `max_len` beforehand)

task/host-sp-comms/src/main.rs
```diff
@@ -777,6 +784,55 @@ impl ServerImpl {
...
+                                >= MAX_INSTALLINATOR_IMAGE_ID_LEN
+                        );
+                        buf = &mut buf[..MAX_INSTALLINATOR_IMAGE_ID_LEN];
+
+                        response_len = cp_agent.get_installinator_image_id(buf);
+                        response_len
+                    },
+                );
+
+                if response_len == 0 {
```
mkeeter, line 828: Is it possible for `control-plane-agent` to load a zero-length installinator image ID, or is that nonsensical?

jgallagher: I think we want some way of unsetting the value, so the control plane can clear out the image ID after recovery completes for cleanliness. Given an image ID has to have _something_ in it, I was using "the image ID vec is empty" as the sentinel value for "we have no image ID", meaning the control plane would intentionally clear the vec out (leading to `response_len == 0` here) after recovery completes.

If this is too clever I could use `Option`s instead; `Option<Vec<u8>>` always feels weird to me though when `Some([])` is effectively the same as `None`.

mkeeter: Okay, that plan seems reasonable to me!

task/control-plane-agent/src/mgs_gimlet.rs
mkeeter: Yeah, I don't feel too strongly about it either way.

idl/control-plane-agent.idol
jgallagher: Good call; reworded to

> Get the current installinator image ID, returning the number of bytes written. If we have no installinator image ID, does not modify the leased buffer and returns 0.

task/control-plane-agent/src/mgs_gimlet.rs
jgallagher: Added a couple short comments explaining why it's structured this way in 68473bc3

task/host-sp-comms/src/main.rs
jgallagher: Added a comment in 68473bc3 explaining this conversion.

mkeeter (approved):

PR #1123, opened 2023-02-03
Remove transceivers task panic on LED issues, enforce reset pulse

### LINKED ISSUE #981: `transceivers` task should not crash when it cannot talk to the PCA9956Bs
The LED controllers (PCA9956B) provide cosmetic functions for the product and are not mission-critical to normal operation. If for some reason we cannot communicate with them, we should add a `ringbuf` message as such and keep moving.

### LINKED ISSUE #1069: `transceivers-server` shouldn't trash an entire request if one transceiver errors
Right now we [`Err` an entire request](https://github.com/oxidecomputer/hubris/blob/master/drv/transceivers-server/src/udp.rs#L480) if a single port has an error. This is a bad user experience and does not really give much actionable info back to the host.

We should put additional thought into this interaction and how we want it to work. I'd like to give all the good data that we can and then provide an error message for each port that failed, even if it would just be "Port X had an I2C error", which is better than the current state where we just error out completely:

```
BRM42220028-switch # xcvradm -i sidecar0 -f 0 identify
Error: Failed to identify transceiver modules

Caused by:
    0: Network protocol error
    1: Failure during read: FPGA reported an I2C error, module may not be present
```

Aaron-Hartwig:
This code more gracefully handles `Err` conditions when we have an I2C failure with the PCA9956B's rather than just `.unwrap()`-ing and potentially causing the `transceivers` task to panic. Since the LEDs are not critical to function, we will now log a `LEDReadError` or `LEDEnableError` to the `ringbuf` as appropriate.

Fixes #981

Edit: Fix incorrect issue linkage.

### DIFF
--- a/drv/sidecar-front-io/src/transceivers.rs
+++ b/drv/sidecar-front-io/src/transceivers.rs
@@ -638,18 +638,42 @@ impl Transceivers {
         }
     }
 
-    /// Releases the LED controller from reset and enables the output
-    pub fn enable_led_controllers(&mut self) -> Result<(), FpgaError> {
+    /// Apply reset to the LED controller
+    ///
+    /// Per section 7.6 of the datasheet the minimum required pulse width here
+    /// is 2.5 microseconds. Given the SPI interface runs at 3MHz, the
+    /// transaction to clear the reset would take ~10 microseconds on its own,
+    /// so there is no additional delay here.
+    pub fn assert_led_controllers_reset(&mut self) -> Result<(), FpgaError> {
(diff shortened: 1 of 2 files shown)


Aaron-Hartwig:
I will test this next week when I can make use of the `niles` setup. With these changes I think I can get rid of the conditional compile around the initialization code, but I don't think we have a `sidecar-a` for me to even verify that on. Regardless, we should be removing `sidecar-a` as a target soon enough (🤞🏻).

mkeeter (approved):

PR #1124, opened 2023-02-06
Add LSP / rust-analyzer support

mkeeter:
This is messy but shockingly functional!

Here's the main idea: given a source file (e.g. `task/monorail-server/src/main.rs`), we need to find a Hubris `app.toml` and task within that app which uses that source file.

We can then generate the Hubris build configuration for that task (`HUBRIS_TASK_CONFIG`, etc) and serialize it to a JSON blob.  Code on the other side (e.g. the Neovim configuration added to our README) can use that blob to set up a custom `rust-analyzer` client environment.

Finding the app + task is the hard part here!  I ended up re-implementing limited dependency resolution + feature unification, based on `cargo_metadata`.  It seems to work!

I'm not sure how easy this will be to integrate into the VSCode `rust-analyzer` plugin.  With Neovim, I had to hook into `on_attach` to avoid defaulting to a single workspace client for all of Hubris; I haven't looked at the VSCode plugin source to see whether that's feasible.


build/xtask/src/lsp.rs
```diff
@@ -0,0 +1,390 @@
...
+                            }
+                        }
+                    }
+                }
+
+                // Check to see if we're also enabling a feature here
+                if feat.is_none() {
+                    continue;
+                }
+                let feat = feat.unwrap();
```
jgallagher, line 160: Maybe

```rust
let Some(feat) = feat else {
    continue;
};
```

build/xtask/src/lsp.rs
```diff
@@ -0,0 +1,390 @@
...
+        let mut build_override_command: Vec<String> =
+            "cargo check --message-format=json"
+                .split(' ')
+                .map(|s| s.to_string())
+                .collect();
+        build_override_command.extend(build_cfg.args.iter().cloned());
+        build_override_command
+            .extend(dependencies.keys().map(|p| format!("-p{p}")));
+        build_override_command.push("--target-dir".to_owned());
+        build_override_command.push("target/rust-analyzer".to_owned());
```
jgallagher, line 278: I also use this target dir for rust-analyzer, but is this personal-config-specific?

build/xtask/src/lsp.rs
```diff
@@ -0,0 +1,390 @@
...
+        {
+            return Ok(out);
+        }
+    }
+
+    let preferred_apps = [
+        "app/gimlet/rev-c.toml",
+        "app/sidecar/rev-b.toml",
+        "app/psc/rev-b.toml",
+    ];
```
jgallagher, line 358: Is there a way to tell from within nvim which app it picked? When I open `control-plane-agent` I'm seeing some errors that imply none of the `gimlet`/`sidecar`/`psc` features are active, and I think I expected `gimlet` to be active based on this list.

mkeeter: `LspInfo` should show you what app + task is selected.

README.mkdn
```diff
@@ -180,6 +180,140 @@ more tasks in the context of a particular image:
...
+    on_new_config = function(new_config, new_root_dir)
+      local bufnr = vim.api.nvim_get_current_buf()
+      local bufname = vim.api.nvim_buf_get_name(bufnr)
+      local dir = new_config.root_dir()
+      if string.find(dir, "hubris") then
+        -- Run `xtask lsp` for the target file, which gives us a JSON
+        -- dictionary with bonus configuration.
+        local prev_cwd = vim.fn.getcwd()
+        vim.cmd("cd " .. dir)
+        local cmd = dir .. "/target/debug/xtask lsp "
```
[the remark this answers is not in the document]
mkeeter: That was intentional, because running through `cargo xtask` takes ~580 ms; running the binary directly takes ~200 ms, which is very noticeable when you change a file.

I added a comment in 98474fc361b61930243b183b4389c4f0f6fc1a11 and also tweaked the suggested config to print an obvious warning.

build/xtask/src/lsp.rs
mkeeter: Good idea, done in b609f504bcb05a04b55235342a3c8be0d44004ad

build/xtask/src/lsp.rs
mkeeter: On second thought, it's useful, but let's do it on the editor side (to make it more flexible); done in bfb8e1ed52e934cd0038ac8a4e8967d9e1b24a35

PR #1139, opened 2023-02-10
arm_m: clear other fault pend bits if task faults.

### LINKED ISSUE #1134: Stack overflow on SVC entry still performs SVC, but from the wrong task
This is very mysterious.

If you build `sidecar/rev-b.toml` on commit `6465f006f3aacc7c51b6a0b8114438044d703d06`, the `control_plane_agent` task has very little stack margin – so little, in fact, that you can trigger a stack overflow by talking to it:
```
$ faux-mgs --interface axf2 --discovery-addr [fe80::0c1d:dcff:fecf:2734]:11111 state
```
(this will time out)

After this failure, the system should be in an odd state: `control_plane_agent` will be faulted, and `jefe` will be waiting on the `fault` bit (1) but not its timer bit (2), which should never happen.

Normally, `jefe` should be notified by the kernel when a task faults, and will restart it.  It's unclear why this isn't happening.

Adding an infinite loop to `configurable_fault` [here](https://github.com/oxidecomputer/hubris/blob/a12d38092dade5e7c7ab91c6f5581c3d525d79b5/sys/kern/src/arch/arm_m.rs#L1211) (right before the return) shows a system that **should** return to `jefe`:

```
matt@niles ~ (sidecar-b) $ h registers
   R0 = 0x240004c0 <- kernel: HUBRIS_TASK_TABLE_SPACE+0x28
   R1 = 0x00000008
   R2 = 0x110b0009
   R3 = 0x00000008
   R4 = 0x00000000
   R5 = 0x0000ffff
   R6 = 0x00000001
   R7 = 0x00000000
   R8 = 0x00000000
   R9 = 0x00000000
  R10 = 0x24040710 <- jefe: JEFE_EXTERNAL_READY+0x0
  R11 = 0x00000001
  R12 = 0x24040488 <- jefe: 0x24040000+0x488
   SP = 0x24000398 <- kernel: 0x24000000+0x398
   LR = 0xffffffed
   PC = 0x08003bc2 <- kernel: configurable_fault+0x36
  PSR = 0x61000004 <- 0110_0001_0000_0000_0000_0000_0000_0100
                      |||| | ||         |       |           |
                      |||| | ||         |       |           + Exception = 0x4
                      |||| | ||         |       +------------ IC/IT = 0x0
                      |||| | ||         +-------------------- GE = 0x0
                      |||| | |+------------------------------ T = 1
                      |||| | +------------------------------- IC/IT = 0x0
                      |||| +--------------------------------- Q = 0
                      |||+----------------------------------- V = 0
                      ||+------------------------------------ C = 1
                      |+------------------------------------- Z = 1
                      +-------------------------------------- N = 0

  MSP = 0x24000398 <- kernel: 0x24000000+0x398
  PSP = 0x24040488 <- jefe: 0x24040000+0x488
  SPR = 0x05000000 <- 0000_0101_0000_0000_0000_0000_0000_0000
                            |||         |         |         |
                            |||         |         |         + PRIMASK = 0
                            |||         |         +---------- BASEPRI = 0x0
                            |||         +-------------------- FAULTMASK = 0
                            ||+------------------------------ CONTROL.nPRIV = 1
                            |+------------------------------- CONTROL.SPSEL = 0
                            +-------------------------------- CONTROL.FPCA = 1

FPSCR = 0x00000000
```

Here's the saved `jefe` state, which looks like the kernel sending a notification of 1:
```
matt@niles ~ (sidecar-b) $ h tasks -v jefe
system time = 29858
 0 jefe                         0   0 RUNNING
   |
   +-----------> 0x24000498 Task {
                    save: SavedState {
                        r4: 0x0,
                        r5: 0xffff,
                        r6: 0x1,
                        r7: 0x0,
                        r8: 0x0,
                        r9: 0x0,
                        r10: 0x24040710,
                        r11: 0x1,
                        psp: 0x24040488,
                        exc_return: 0xffffffed,
                        s16: 0x0,
                        s17: 0x0,
                        s18: 0x0,
                        s19: 0x0,
                        s20: 0x0,
                        s21: 0x0,
                        s22: 0x0,
                        s23: 0x0,
                        s24: 0x0,
                        s25: 0x0,
                        s26: 0x0,
                        s27: 0x0,
                        s28: 0x0,
                        s29: 0x0,
                        s30: 0x0,
                        s31: 0xffffffff
                    },
                    priority: Priority(0x0),
                    state: Healthy(Runnable),
                    timer: TimerState {
                        deadline: Some(Timestamp(0x74cc)),
                        to_post: NotificationSet(0x2)
                    },
                    generation: 0x0,
                    notifications: 0x0,
                    descriptor: 0x8004d28 (&kern::descs::TaskDesc)
                }
```

Because `LR` is `0xFFFFFFED`, it will return by looking at PSP.  We, too, can look at PSP:
```
matt@niles ~ (sidecar-b) $ h readmem 0x24040488 -w
                    0        4       \/        c
0x24040480 |                   240405a0 00000008 |         ...$....
0x24040490 | 00000003 240405c8 240405a8 080083f1 | .......$...$....
0x240404a0 | 08009692 61000000 00000000 00000000 | .......a........
0x240404b0 | 00000000 00000000 00000000 00000000 | ................
0x240404c0 | 00000000 00000000 00000000 00000000 | ................
0x240404d0 | 00000000 00000000 00000000 00000000 | ................
0x240404e0 | 00000000 00000000 00000000 00000000 | ................
0x240404f0 | 00000000 00000017 
```
This shows that it's about to return to `08009692`, which is our `sys_recv_stub` function:
```
arm-none-eabi-addr2line -e jefe -i 0x08009692
/hubris/sys/userlib/src/lib.rs:387
```

However, it never seems to make it: adding a jefe-specific trap in `userlib::sys_recv_stub`, it never seems to be entered.

One theory is something about interrupt chaining going wrong: if there are pending interrupts when the final `bx lr` in `configurable_fault` is evaluated, then it will handle them instead, and that... somehow... eventually prevents jumping to `jefe`?  A piece of evidence for this theory: in the infinite-loop-before-`bx lr` code, we see `jefe` listed as `Runnable`; however, if we remove that loop and let the `bx lr` execute, `jefe` ends up in `Healthy(InRecv(None))`.

One more observation: the `net` task will panic itself every 60 seconds if it doesn't see any traffic.  When that happens, `jefe` is woken up and restarts `control_plane_agent`, as it should.

If I artificially lower the `control_plane_agent` stack size on Gimletlet, this does not reproduce; haven't tried yet on Gimlet.

cbiffle:
This fixes an unusual but vexing misbehavior in fault handling. If a task ran out of stack as a side effect of executing an SVC instruction, instead of entering the kernel at SVCall, it would enter at MemManageFault. However, _the SVC exception triggered by the SVC instruction is still pending when this happens._

This means that the flow of control goes like this:

1. Task faults.
2. Kernel does bookkeeping.
3. Kernel notifies supervisor and switches to supervisor task.
4. Kernel attempts to return to taskyland.
5. ARM M-profile tail chaining kicks in, converting the return-from-interrupt operation into a branch to SVCall.
6. Because we've already changed the current task, this simulates a surprise syscall _from the supervisor task._
7. In practice, the syscall is recv, and the supervisor then issues a bogus-shaped recv instead of handling the fault.

We can fix this by clearing the SVCALLPENDED bit in the interrupt controller. In practice, we can do this unconditionally when we're in the kernel handling any task fault -- since a fault in the kernel while servicing an SVC would follow a different (and fatal) code path.

And so that's what I've done here.

Thanks to @mkeeter for working out the test reduction case for reproducing this on any H7 board!

Fixes #1134.

### DIFF
--- a/sys/kern/src/arch/arm_m.rs
+++ b/sys/kern/src/arch/arm_m.rs
@@ -1383,6 +1383,29 @@ unsafe extern "C" fn handle_fault(task: *mut task::Task) {
         panic!("Kernel fault");
     }
 
+    // Okay, now that we're confident we came from a task, we need to deal with
+    // the case where the fault occurred while stacking an SVC exception frame.
+    // In this case, the SVC exception will still be set as pending, which means
+    // when we try to return to the supervisor to handle this fault, it'll
+    // generate a spurious SVC. Even in the best of cases, this breaks the
+    // supervisor.
+    //
+    // SVCALLPENDED is in the System Handler Control and State Register, bit 15,
+    // and we need to clear its bit. We do this unconditionally because it
+    // doesn't hurt, and it's slightly faster/smaller.
+    //
+    // (If you're comparing this to the ARMv7/v8-M equivalent, note that v6-M
+    // lacks the usage/mem/bus faults present in v7/8.)
+    //
+    // Safety: the cortex-m crate makes all these registers blanket-unsafe
+    // without documenting the required preconditions. From the ARMv7-M spec, we
+    // can infer that the main risk here is if SVC were higher priority than
+    // this handler, which it is not.
+    unsafe {
+        let scb = &*cortex_m::peripheral::SCB::PTR;
+        scb.shcsr.modify(|bits| bits & !(1 << 15));
+    }
+
     // ARMv6-M, to reduce complexity, does not distinguish fault causes.
     let fault = FaultInfo::InvalidOperation(0);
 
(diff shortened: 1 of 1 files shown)


cbiffle:
Updated to also clear usage/bus/memmang fault pend bits, because otherwise you could reproduce this problem by substituting any illegal instruction for SVC.

PR #1144, opened 2023-02-13
Use Option to hold intermediate cert in CertData.

flihp:
Across the DICE mfg options the intermediate cert is only present if the identity cert has been manufactured (not self signed). Using an Option to represent this makes code on the consumer side less bad.


andrewjstone (approved):
Ship it!

Issue #985, opened 2022-12-05
The SP should remember which host flash slot is active


jgallagher:
Currently, when `host-flash` starts, it [always chooses slot 0](https://github.com/oxidecomputer/hubris/blob/ee3787efcac4b1a3c90bdf11504aeb345bf9b1ce/drv/gimlet-hf-server/src/main.rs#L148), which is then [reported to the host](https://github.com/oxidecomputer/hubris/blob/ee3787efcac4b1a3c90bdf11504aeb345bf9b1ce/task/host-sp-comms/src/main.rs#L582-L599) when it asks. However, if the control plane has previously told the SP it should be booting flash slot 1, the SP should remember that choice if the `host-flash` task or the SP itself restarts. Currently changing the host flash slot only persists until `host-flash` restarts.

andrewjstone:
This information should be tracked and made available to the `stm32h7-update-server` similar to what is done in #1009 for the RoT.

mkeeter:
Should we also be saving the host startup options?

jgallagher:
Probably so, yeah.

mkeeter:
I had some good discussion in [#oxide-host-software here](https://matrix.to/#/!rHMNCYjVQiQsVlGkGP:oxide.computer/$HT2w6zmGIvB56wCc6NBmX8J2R-HW6_l-zolJeI5RjTQ?via=oxide.computer&via=unix.house&via=timecube.club), and have the core of a proposal.

The AMD boot process doesn't care about anything other than a few particular memory addresses in the SPIROM, which contain user-defined pointers to other data structures in the SPIROM.  None of these magic addresses are below 0x10000, so we could reserve the lowest sector (64 KiB) for the SP's use.

To remain robust against power loss at any point, we should use a ping-pong strategy with a monotonically increasing epoch value.  This is similar to the strategy in [RFD 311](https://rfd.shared.oxide.computer/rfd/0311), but uses an epoch instead of a checksum to select which slot to boot.  Note that unlike in RFD 311, where we ping-pong between slots within a single flash chip, here we're ping-ponging between the _two different flash ICs_!

Consider storing the following data structure at address 0 in the QSPI:
```rust
#[derive(AsBytes, FromBytes)]
#[repr(C)]
struct SpiHeader {
    empty: u32,   // must always be 0xFFFF_FFFF
    magic: u32,   // must always be 0x01de_abcd or something
    version: u64, // header version format, must always be 1 (for now)
[...]

Issue #994, opened 2022-12-07
`hf` should at least respond to inquiries when it's upset


cbiffle:
Currently, if the hf driver fails to recognize the JEDEC ident info it reads from the QSPI host flash, it does this:

```rust
    if log2_capacity.is_none() {
        loop {
            // We are dead now.
            hl::sleep_for(1000);
        }
    }
```

This means that anyone who attempts to IPC the `hf` server will just wait forever, which is not great. It would probably be better to have the server come up in a known-degraded state, unable to (say) do programming but willing to respond to status inquiries or re-load the JEDEC info.

mkeeter:
Also, the `hf` task always muxes the flash to the SP when the task starts up.  @jgallagher points out in #1145 that if the task has died for some reason – but the rest of the system is in A0 – this could steal the flash away from the host!

nathanaelhuffman:
just to link this: https://github.com/oxidecomputer/hubris/issues/1172 this is the issue about expected muxing away

cbiffle:
> Also, the `hf` task always muxes the flash to the SP when the task starts up. @jgallagher points out in #1145 that if the task has died for some reason – but the rest of the system is in A0 – this could steal the flash away from the host!

Since the mux state is maintained by a GPIO, it should be straightforward to make `hf` respect its current state on task restart.

labbott:
Cosmo introduced the idea of a `FailServer` which returns an error each time. We now have that on gimlet after #2116. It's not quite " ...have the server come up in a known-degraded state, unable to (say) do programming but willing to respond to status inquiries or re-load the JEDEC info." but I think it's enough of an improvement over the existing behavior we can close out this issue and file a new one when we decide we want different behavior.

PR #1145, opened 2023-02-13
Initial support for host flash persistence

mkeeter:
This PR implements the scheme described in https://github.com/oxidecomputer/hubris/issues/985 for storing persistent host flash data (current slot and boot flags).

It reserves the lowest sector (64 KiB) of QSPI flash for Hubris, adding new HIF functions if you _really_ want to overwrite it and returning an error code otherwise.  Within sector 0, we store up to 512 copies of `RawPersistentData`; the one with the largest `epoch` field is our current value.

A new Idol API is added to write this persistent data, by scanning sector 0 on both flash ICs, finding the highest `epoch`, and incrementing it.

`control-plane-agent` is modified to avoid writing sector 0, but checks that any bytes that we _try_ to write in that sector are `0xFF` and returns an error otherwise.

### DIFF
--- a/drv/gimlet-hf-server/src/main.rs
+++ b/drv/gimlet-hf-server/src/main.rs
@@ -131,13 +131,12 @@ fn main() -> ! {
         }
     };
 
-    if log2_capacity.is_none() {
+    let Some(log2_capacity) = log2_capacity else {
         loop {
             // We are dead now.
             hl::sleep_for(1000);
         }
-    }
-    let log2_capacity = log2_capacity.unwrap();
+    };
     qspi.configure(cfg.clock, log2_capacity);
 
     let mut buffer = [0; idl::INCOMING_SIZE];
@@ -150,6 +149,8 @@ fn main() -> ! {
         mux_select_pin: cfg.sp_host_mux_select,
         dev_select_pin: cfg.flash_dev_select,
     };
+
+    server.ensure_persistent_data_is_redundant().unwrap(); // TODO: log this?
 
     loop {
         idol_runtime::dispatch(&mut buffer, &mut server);
@@ -376,6 +377,7 @@ impl ServerImpl {
         }
         (best, empty_slot)
     }
+
     fn raw_sector0_erase(&mut self) -> Result<(), HfError> {
         set_and_check_write_enable(&self.qspi)?;
         self.qspi.sector_erase(0);
@@ -396,6 +398,59 @@ impl ServerImpl {
             }
         };
         Self::page_program_raw(&self.qspi, addr, raw_data.as_bytes())
+    }
+
+    /// Checks that the persistent data is consistent between the two flash ICs.
+    fn ensure_persistent_data_is_redundant(&mut self) -> Result<(), HfError> {
+        // This should only be called on startup, at which point we're always
+        // muxed to the SP.
+        self.check_muxed_to_sp().unwrap();
+
+        // We can't have redundant data if we've only got a single flash
+        if self.dev_select_pin.is_none() {
+            return Ok(());
+        }
+
+        // Load the current state of persistent data from flash
+        let prev_slot = self.dev_state;
+        self.set_dev(!prev_slot).unwrap();
+        let (a_data, a_next) = self.persistent_data_scan();
+
+        self.set_dev(prev_slot).unwrap();
+        let (b_data, b_next) = self.persistent_data_scan();
+
+        match (a_data, b_data) {
+            (Some(a), Some(b)) => match a.epoch.cmp(&b.epoch) {
+                core::cmp::Ordering::Less => {
+                    self.set_dev(!prev_slot).unwrap();
+                    let out =
+                        self.write_raw_persistent_data_to_addr(a_next, &b);
+                    self.set_dev(prev_slot).unwrap();
+                    out
+                }
+                core::cmp::Ordering::Greater => {
+                    self.write_raw_persistent_data_to_addr(b_next, &a)
+                }
+                core::cmp::Ordering::Equal => {
+                    // Redundant data is consistent
+                    // TODO: should we have a special case if they don't agree?
+                    Ok(())
+                }
+            },
+            (Some(a), None) => {
+                self.write_raw_persistent_data_to_addr(b_next, &a)
+            }
+            (None, Some(b)) => {
+                self.set_dev(!prev_slot).unwrap();
+                let out = self.write_raw_persistent_data_to_addr(a_next, &b);
+                self.set_dev(prev_slot).unwrap();
+                out
+            }
+            (None, None) => {
+                // No persistent data recorded; nothing to do here
+                Ok(())
+            }
+        }
     }
 }
 
@@ -607,6 +662,7 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         })
     }
 
+    /// Writes the given persistent data to host flash
     fn write_persistent_data(
         &mut self,
         _: &RecvMessage,
@@ -617,12 +673,12 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
             let prev_slot = self.dev_state;
 
             self.set_dev(!prev_slot).unwrap();
-            let (a_slot, a_next) = self.persistent_data_scan();
+            let (a_data, a_next) = self.persistent_data_scan();
 
             self.set_dev(prev_slot).unwrap();
-            let (b_slot, b_next) = self.persistent_data_scan();
+            let (b_data, b_next) = self.persistent_data_scan();
 
-            let prev_epoch = match (a_slot, b_slot) {
+            let prev_epoch = match (a_data, b_data) {
                 (Some(a), Some(b)) => a.epoch.max(b.epoch),
                 (Some(a), None) => a.epoch,
                 (None, Some(b)) => b.epoch,
@@ -631,7 +687,6 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
             let epoch = prev_epoch + 1;
             let raw = RawPersistentData::new(data, epoch);
 
-            // ----------------------------------------------------------------
             // Write the persistent data to the currently inactive flash.
             //
             // At this point, all of the ways that `set_dev` could fail have
@@ -641,15 +696,14 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
             self.set_dev(!prev_slot).unwrap();
             let out_a = self.write_raw_persistent_data_to_addr(a_next, &raw);
 
-            // ----------------------------------------------------------------
             // Write the persistent data to the currently active flash
             self.set_dev(prev_slot).unwrap();
-            let out_b = self.write_raw_persistent_data_to_addr(b_next, &raw);
 
             // Now that we've restored the current active flash, check whether
             // we should propagate errors.
             out_a?;
-            out_b?;
+
+            self.write_raw_persistent_data_to_addr(b_next, &raw)?;
         } else {
             // Single-flash case is less complicated, because we don't have to
             // track and maintain the active device.
--- a/drv/gimlet-hf-server/src/main.rs
+++ b/drv/gimlet-hf-server/src/main.rs
@@ -183,7 +183,7 @@ struct RawPersistentData {
     /// Must always be `HF_PERSISTENT_DATA_MAGIC`.
     oxide_magic: u32,
 
-    /// Must always be 1 (for now)
+    /// Must always be `HF_PERSISTENT_DATA_HEADER_VERSION` (for now)
     header_version: u32,
 
     /// Monotonically increasing counter
--- a/drv/gimlet-hf-server/src/main.rs
+++ b/drv/gimlet-hf-server/src/main.rs
@@ -735,7 +735,9 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
                 return Ok(());
             }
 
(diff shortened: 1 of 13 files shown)


drv/gimlet-hf-server/src/main.rs
```diff
@@ -147,11 +151,96 @@ fn main() -> ! {
         dev_select_pin: cfg.flash_dev_select,
     };
 
+    server.ensure_persistent_data_is_redundant().unwrap(); // TODO: log this?
```
jgallagher, line 154: There's also #994; maybe not as part of this PR, but should we take Cliff's suggestion of "server starts in a degraded state" for both this failure and the capacity failure above?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -147,11 +151,96 @@ fn main() -> ! {
...
+        c.update(&self.as_bytes()[..size]);
+        c.finalize()
+    }
+
+    fn is_valid(&self) -> bool {
+        self.amd_reserved_must_be_all_ones == u64::MAX
+            && self.oxide_magic == HF_PERSISTENT_DATA_MAGIC
+            && self.header_version == HF_PERSISTENT_DATA_HEADER_VERSION
+            && self.dev_select <= 1
+            && self.checksum == self.expected_checksum()
```
jgallagher, line 234: Do we also want to check that all set bits of `self.host_startup_options` are valid `HostStartupOptions`? I suspect _not_, because that feels like some fragile over-coupling, but am not completely sure so wanted to ask.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -180,6 +269,190 @@ impl ServerImpl {
...
+        let mut best: Option<RawPersistentData> = None;
+        let mut empty_slot: Option<u32> = None;
+        for i in 0..SECTOR_SIZE_BYTES / HF_PERSISTENT_DATA_STRIDE {
+            let addr = (i * HF_PERSISTENT_DATA_STRIDE) as u32;
+            let mut data = RawPersistentData::new_zeroed();
+            self.qspi.read_memory(addr, data.as_bytes_mut());
+            if data.is_valid()
+                && best.map(|b| b.epoch < data.epoch).unwrap_or(true)
+            {
+                best = Some(data)
```
jgallagher, line 356: Typo:

```suggestion
                best = Some(data);
```

drv/gimlet-hf-server/src/main.rs
```diff
@@ -180,6 +269,190 @@ impl ServerImpl {
...
+            }
+            if empty_slot.is_none()
```
jgallagher, line 358: This might be slightly clearer as an `else if` - if `data.is_valid()`, we know that some of the bytes are not `0xff` already.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -180,6 +269,190 @@ impl ServerImpl {
...
+                0
+            }
+        };
+        Self::page_program_raw(&self.qspi, addr, raw_data.as_bytes())
+    }
+
+    /// Checks that the persistent data is consistent between the two flash ICs.
+    fn ensure_persistent_data_is_redundant(&mut self) -> Result<(), HfError> {
+        // This should only be called on startup, at which point we're always
+        // muxed to the SP.
```
jgallagher, line 415: Is this comment true even if this task panics and restarts while the host is in A0?

drv/gimlet-hf-server/src/main.rs
jgallagher: Separate thought - since we're scanning the flashes on startup anyway, should we remember the current data in `server`? I think that would allow `get_persistent_data()` to be idempotent (just return the current data), and `write_persistent_data()` could no-op out if there were no changes. The latter seems like it might be useful to protect against spurious flash writes if we had an automated process that was (e.g.) setting startup options on every boot (but changing them infrequently).

drv/gimlet-hf-server/src/main.rs
mkeeter: Yes – in `Config::init` in this same file, we ensure that the flash is muxed to the SP.

(Whether this is a good idea is questionable!  If the task restarts while we're in A0, that would be bad!)

drv/gimlet-hf-server/src/main.rs
mkeeter: Hmmm, that's logically correct, but I feel like it's more clear as-is, since the two variables that we're populating (`best` and `empty_slot`) are philosophically distinct.

drv/gimlet-hf-server/src/main.rs
mkeeter: I considered storing persistent data in the server, but decided it was extra complexity versus always doing the scans (which is Fast Enough).

However, "don't write if the persistent data matches" seems like a good plan; done in d6c8d269b9b45ce0c5c157e56d17a22e0056642b

drv/gimlet-hf-server/src/main.rs
mkeeter: Fixed in 7af4bb1c9463209515fa1db11e3f015ca4dc47b5

drv/gimlet-hf-server/src/main.rs
jgallagher: My "don't write if the data matches" suggest was intended for the non-persistent writing function (i.e., the idol call). I'm not sure who the client is going to be - probably control-plane-agent? - but it seems well within reason we could have something setting startup options to the same thing they're already set to, without realizing that could cause flash wear.

drv/gimlet-hf-server/src/main.rs
jgallagher: Huh, I thought of them more closely related - it's not possible for one slot to be both `best` and `empty` at the same time (hence thinking "else if" would make that clear). But this is not a hill I'm dying on - either is fine!

drv/gimlet-hf-server/src/main.rs
mkeeter: I considered it, but agree that it would be a little too much coupling.

drv/gimlet-hf-server/src/main.rs
mkeeter: Now that we're not storing host flash options at all (per discussions in #oxide-host-software), I think this is good to go?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -147,11 +151,114 @@ fn main() -> ! {
...
+    amd_reserved_must_be_all_ones: u64,
+
+    /// Must always be `HF_PERSISTENT_DATA_MAGIC`.
+    oxide_magic: u32,
+
+    /// Must always be `HF_PERSISTENT_DATA_HEADER_VERSION` (for now)
+    header_version: u32,
+
+    /// Monotonically increasing counter
+    epoch: u64,
```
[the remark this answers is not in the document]
lzrd: Sanity check on this field. Will the MSB ever be set?
4 billion values, flash wear for the part on erase/write cycles, number of pages in use.
Something like 10,000 cycles times 128 headers (that's from span) about 2^21. Unless there is a bug, the highest possible value is around 0x00080000 or 0x00800000 if the part allows 100,000 cycles.
(check my math and data sheets)

drv/gimlet-hf-server/src/main.rs
mkeeter: There is implicit coordination with the sequencer – we'd have a bad time if the `hf` task restarts while the host is reading its phase1 image from QSPI – but this PR doesn't make things any worse than before.  I added some comments to that effect in the relevant ticket (#994)

drv/gimlet-hf-server/src/main.rs
```diff
@@ -376,36 +698,95 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
+                None => 0,
+            };
+
+            // Early exit if the previous persistent data matches
+            let prev_raw = RawPersistentData::new(data, prev_epoch);
+            if prev_data == Some(prev_raw) {
+                return Ok(());
+            }
+
+            let epoch = prev_epoch + 1;
```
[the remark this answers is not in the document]
mkeeter: It's a `u64`, rather than a `u32`, so it _definitely_ will not roll over unless someone manually reprograms the flash to have data with an artificially high epoch.  I added an overflow check in 53d6b530c914c8f0f6c862f4305f75ab89475d0e which will return an error on that acse.

drv/gimlet-hf-server/src/main.rs
lzrd: The host has ownership of the flash some of the time which means that the host can write to it. If the host is compromised, what mischief is possible?

jgallagher (approved):

drv/gimlet-hf-server/src/main.rs
```diff
@@ -180,6 +290,226 @@ impl ServerImpl {
...
+    /// - newest persistent data record
+    /// - address of first empty slot
+    fn persistent_data_scan(&self) -> (Option<RawPersistentData>, Option<u32>) {
+        let mut best: Option<RawPersistentData> = None;
+        let mut empty_slot: Option<u32> = None;
+        for i in 0..SECTOR_SIZE_BYTES / HF_PERSISTENT_DATA_STRIDE {
+            let addr = (i * HF_PERSISTENT_DATA_STRIDE) as u32;
+            let mut data = RawPersistentData::new_zeroed();
+            self.qspi.read_memory(addr, data.as_bytes_mut());
+            if data.is_valid()
```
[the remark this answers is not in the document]
mkeeter: During normal operation, we _should_ proceed through the header locations basically in order.  The code doesn't assume that, though, since it picks the header with the highest monotonic counter (no matter where it is).

Scanning through the whole flash takes ~7 ms on Gimletlet, so double that for Gimlet itself.  I think that's fast enough that it's not worth speeding up.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -180,6 +290,226 @@ impl ServerImpl {
...
+            let (a, _) = self.persistent_data_scan();
+
+            // Then switch back to our current slot, so that the resulting state
+            // is unchanged.
+            self.set_dev(prev_slot).unwrap();
+            let (b, _) = self.persistent_data_scan();
+
+            match (a, b) {
+                (Some(a), Some(b)) => {
+                    if a.epoch > b.epoch {
```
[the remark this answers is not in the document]
mkeeter: Renamed `epoch`.  These should be equal unless something went wrong, because (a) the task ensure that they're equal on startup and (b) always writes them in pairs.  They could be unequal in practice if the second write of the pair fails and we haven't yet restarted the task.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -147,11 +151,114 @@ fn main() -> ! {
...
+    /// Monotonically increasing counter
+    epoch: u64,
+
+    /// Equivalent to `host_sp_messages::HostStartupOptions`
+    host_startup_options: u64,
+
+    /// Either 0 or 1; directly translatable to `gimlet_hf_api::HfDevSelect`
+    dev_select: u32,
+
+    /// CRC-32 over the rest of the data using the iSCSI polynomial
```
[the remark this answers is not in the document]
mkeeter: I picked this to match TLV-C, since that's what we use for auxflash.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -147,11 +151,114 @@ fn main() -> ! {
...
+/// with a valid checksum and the highest `epoch` across both flash ICs.
+///
+/// When writing new data, we increment the epoch and write to both ICs, one by
+/// one.  This ensures robustness in case of power loss.
+#[derive(Copy, Clone, Eq, PartialEq, AsBytes, FromBytes)]
+#[repr(C)]
+struct RawPersistentData {
+    /// Reserved field, because this is placed at address 0, which PSP firmware
+    /// may look at under certain circumstances.
+    amd_reserved_must_be_all_ones: u64,
```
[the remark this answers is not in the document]
mkeeter: Not sure; it would certainly be surprising to me if this changed!

drv/gimlet-hf-server/src/main.rs
mkeeter: If the host could modify the flash, then it can already modify the phase1 image, so there's limited _new_ mischief that it could do.  A malicious host could write a header with a very high monotonic counter, which would supersede all other headers when deciding which header to use.  We'd see that represented by a `MonotonicCounterOverflow` error and could respond accordingly (e.g. erasing both flashes to reset `monotonic_counter = 0`).

drv/gimlet-hf-server/src/main.rs
```diff
@@ -180,6 +290,226 @@ impl ServerImpl {
...
+
+    fn get_raw_persistent_data(
+        &mut self,
+    ) -> Result<RawPersistentData, HfError> {
+        self.check_muxed_to_sp()?;
+        let best = if self.dev_select_pin.is_some() {
+            let prev_slot = self.dev_state;
+
+            // At this point, all of the ways that `set_dev` could fail have
+            // been checked: we already called both `check_muxed_to_sp` and that
```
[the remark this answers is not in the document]
mkeeter: It's hard to put this in a type, since it would capture `&mut self`...

PR #1149, opened 2023-02-14
xtask graph: fix priority ranking

cbiffle:
The graphs we were generating did not reliably put different priorities on different tiers. As far as I can tell, this turns out to be due to backwards (priority inversion) edges causing it to throw up its hands and collapse ranks.

This marks backwards edges as constraint=false to exempt them from the rank calculation.


lzrd (approved):
This is better.

PR #1151, opened 2023-02-15
spontaneous reset support + sequencer fixes + SB-RMI support

### LINKED ISSUE #1135: STM32 I2C driver busy-wait spin count is too short on H7
The stm32xx-i2c crate contains an optimistic busy-wait routine for sleeping until the BUSY flag is cleared, as ST does not appear to provide an interrupt for that condition. Currently, it loops 3 times before sleeping for 2 ms.

This sleep is happening on basically every mux-changed I2C access according to my traces here, because the busy loop itself takes about 2.66 µs, or a small fraction of one bit on I2C. This is causing mux changes to take 2x longer than they should.

The idea of a hardcoded busy-wait repeat count probably doesn't make sense now that the crate has been generalized to a range of STM32 variants, from the relatively slow stm32g0 up to the stm32h7. No hardcoded count is going to be correct on all these platforms.

Currently this is causing us slightly more than 2 ms of idle time between mux selection and the next I2C transaction. The frequency with which this happens depends on how often the mux is changed, which in our current architecture is impossible to predict (since we're not scheduling I2C activity and instead doing thundering-herd).

### LINKED ISSUE #1160: sequencer/power race results in lost disables of RAA229618 rails on transition to A2
We have -- very rarely -- seen machines fail to transition from A2 to A0, seemingly because the A0 state machine skipped ahead, indicating that rails were enabled before we had in fact enabled them.  

This has been seen a couple of different times over the last few months.  For example, here is the ring buffer from [a dump of sn9](https://github.com/oxidecomputer/hubris/files/10775281/hubris.core.sn9-hang.0.zip), a machine that so hung after 5 days of non-stop rebooting (!):

```
  19  467    36495        1 SetState(A0PlusHP, A2)
  20  552    36495        1 A2
  21  458    36495        1 GetState
  22  366    36495        1 Status(0x0, 0x8, 0xff)
  23  458    36495       13 GetState
  24  467    36495        1 SetState(A2, A0)
  25  488    36495      101 A1Power(0x3, 0xc)
  26  488    36495 76088280 A1Power(0x5, 0xc)
```

In looking at this, @nathanaelhuffman observed that should only be possible if the rails turned themselves on -- or if they were never turned off in the earlier transition to A2.  We took the opportunity to both improve the instrumentation to record a little more sequencer state, and [to improve the sequencer itself](https://github.com/oxidecomputer/quartz/issues/57), but were unable to get this reproducible enough for further study.

Later, in [a dump from sn6](https://github.com/oxidecomputer/hubris/files/10775284/core.hubris.sn6.20230124.zip), a machine was found by @rmustacc to have had [a seeming manifestation of hardware-gimlet#1815](https://github.com/oxidecomputer/hardware-gimlet/issues/1815#issuecomment-1402993834") -- but it is in fact this same mystery of the enabled rails:

```
   2  446    21616        1 SetState(A2, A0)
   3  448    21616        1 PGStatus(0x0, 0x0, 0x0)
   4  454    21616        1 SMStatus(0x0, 0x0)
   5  459    21616        1 PowerControl(0x0)
   6  463    21616        1 InterruptFlags(0x0)
   7  492    21616        2 A1Power(0x2, 0x0)
   8  492    21616      100 A1Power(0x3, 0x0)
   9  492    21616       10 A1Power(0x5, 0x1)
  10  492    21616      336 A1Power(0x5, 0x2)
  11  492    21616        1 A1Power(0x5, 0x4)
  12  492    21616        2 A1Power(0x5, 0x6)
  13  492    21616        9 A1Power(0x5, 0xa)
  14  492    21616        4 A1Power(0x5, 0xb)
  15  492    21616   202923 A1Power(0x5, 0xc)
```

More recently, as part of the compliance preparation, c23 (serial BRM42220023) was found to be [spontaneously in A2](https://github.com/oxidecomputer/hardware-gimlet/issues/1859).  We do not have a dump from this (it predates our ability to get a dump from a racked Gimlet), but this Gimlet was removed and put in a reboot loop. After ~12 hours, it hit the issue; from 
[the dump](https://github.com/oxidecomputer/hubris/files/10775287/hubris.core.BRM42220023-hung-sequencer.zip):

```
  21  491       55        1 SetState(A2, A0)
  22  493       55        1 PGStatus { b_pg: 0x0, c_pg: 0x0, nic: 0x0 }
  23  499       55        1 SMStatus { a1: 0x0, a0: 0x0 }
  24  504       55        1 PowerControl(0x0)
  25  508       55        1 InterruptFlags(0x0)
  26  538       55        2 A1Power(0x2, 0x0)
  27  538       55      100 A1Power(0x3, 0x0)
  28  538       55       10 A1Power(0x5, 0x1)
  29  538       55      292 A1Power(0x5, 0x2)
  30  538       55        1 A1Power(0x5, 0x4)
  31  538       55        2 A1Power(0x5, 0x6)
  32  538       55        9 A1Power(0x5, 0xa)
  33  538       55        4 A1Power(0x5, 0xb)
  34  538       55   730752 A1Power(0x5, 0xc)
```

Additional instrumentation was added (specifically, to record the `STATUS_WORD` on the rail whenever we saw the state machine advance); this was seen again on c23 ([dump](https://github.com/oxidecomputer/hubris/files/10775288/hubris.core.BRM42220023-hang.2.zip)):

```
humility: attached to dump
  72  578        3        1 A0Power(0xc)
  73  597        3        1 UartEnabled
  74  486        3        1 UpdateState(A0)
  75  408        3    10806 Status { ier: 0x0, ifr: 0x0, amd_status: 0x4, amd_a0: 0x4 }
  76  439        3        1 NICPowerEnableLow(false)
  77  486        3        1 UpdateState(A0PlusHP)
  78  408        3      690 Status { ier: 0x0, ifr: 0x0, amd_status: 0x4, amd_a0: 0x4 }
  79  408        3        1 Status { ier: 0x0, ifr: 0x60, amd_status: 0x4, amd_a0: 0x8 }
  80  672        3        1 ResetCounts { rstn: 0x1, pwrokn: 0x1 }
  81  486        3        1 UpdateState(A0Reset)
  82  495        3        1 SetState(A0Reset, A2)
  83  497        3        1 PGStatus { b_pg: 0xff, c_pg: 0x3, nic: 0x7f }
  84  503        3        1 SMStatus { a1: 0x5, a0: 0xc }
  85  508        3        1 PowerControl(0x3)
  86  512        3        1 InterruptFlags(0x0)
  87  486        3        1 UpdateState(A2)
  88  643        3        1 A2
  89  912        3        1 RailStatusCore(0x840)
  90  920        3        1 RailStatusSoc(0x840)
  91  495        3        1 SetState(A2, A0)
  92  497        3        1 PGStatus { b_pg: 0x0, c_pg: 0x0, nic: 0x0 }
  93  503        3        1 SMStatus { a1: 0x0, a0: 0x0 }
  94  508        3        1 PowerControl(0x0)
  95  512        3        1 InterruptFlags(0x0)
  96  543        3        2 A1Power(0x2, 0x0)
  97  543        3      100 A1Power(0x3, 0x0)
  98  543        3        1 A1Power(0x5, 0x1)
  99  912        3        1 RailStatusCore(0x840)
   0  920        4        1 RailStatusSoc(0x840)
   1  543        4        8 A1Power(0x5, 0x1)
   2  543        4        1 A1Power(0x5, 0x2)
   3  912        4        1 RailStatusCore(0x840)
   4  920        4        1 RailStatusSoc(0x840)
   5  543        4      296 A1Power(0x5, 0x2)
   6  543        4        1 A1Power(0x5, 0x4)
   7  913        4        1 RailErrorCore(BadRead { cmd: 0x79, code: BusReset })
   8  921        4        1 RailErrorSoc(BadWrite { cmd: 0x0, code: BusReset })
   9  543        4        1 A1Power(0x5, 0x6)
  10  912        4        1 RailStatusCore(0x842)
  11  920        4        1 RailStatusSoc(0x842)
  12  543        4        1 A1Power(0x5, 0x7)
  13  912        4        1 RailStatusCore(0x842)
  14  920        4        1 RailStatusSoc(0x842)
  15  568        4        1 RailsOn
  16  912        4        1 RailStatusCore(0x0)
  17  920        4        1 RailStatusSoc(0x0)
  18  578        4       10 A0Power(0xa)
  19  578        4        3 A0Power(0xb)
  20  578        4        1 A0Power(0xc)
  21  597        4        1 UartEnabled
  22  486        4        1 UpdateState(A0)
  23  408        4    10826 Status { ier: 0x0, ifr: 0x0, amd_status: 0x4, amd_a0: 0x4 }
  24  439        4        1 NICPowerEnableLow(false)
  25  486        4        1 UpdateState(A0PlusHP)
  26  408        4      641 Status { ier: 0x0, ifr: 0x0, amd_status: 0x4, amd_a0: 0x4 }
  27  408        4        1 Status { ier: 0x0, ifr: 0x60, amd_status: 0x4, amd_a0: 0x4 }
  28  672        4        1 ResetCounts { rstn: 0x1, pwrokn: 0x1 }
  29  486        4        1 UpdateState(A0Reset)
  30  495        4        1 SetState(A0Reset, A2)
  31  497        4        1 PGStatus { b_pg: 0xff, c_pg: 0x3, nic: 0x7f }
  32  503        4        1 SMStatus { a1: 0x5, a0: 0xc }
  33  508        4        1 PowerControl(0x3)
  34  512        4        1 InterruptFlags(0x0)
  35  486        4        1 UpdateState(A2)
  36  643        4        1 A2
  37  912        4        1 RailStatusCore(0x840)
  38  920        4        1 RailStatusSoc(0x840)
  39  495        4        1 SetState(A2, A0)
  40  497        4        1 PGStatus { b_pg: 0x0, c_pg: 0x0, nic: 0x0 }
  41  503        4        1 SMStatus { a1: 0x0, a0: 0x0 }
  42  508        4        1 PowerControl(0x0)
  43  512        4        1 InterruptFlags(0x0)
  44  543        4        2 A1Power(0x2, 0x0)
  45  543        4      100 A1Power(0x3, 0x0)
  46  543        4        1 A1Power(0x5, 0x1)
  47  912        4        1 RailStatusCore(0x840)
  48  920        4        1 RailStatusSoc(0x840)
  49  543        4        8 A1Power(0x5, 0x1)
  50  543        4        1 A1Power(0x5, 0x2)
  51  912        4        1 RailStatusCore(0x840)
  52  920        4        1 RailStatusSoc(0x840)
  53  543        4      289 A1Power(0x5, 0x2)
  54  543        4        1 A1Power(0x5, 0x4)
  55  912        4        1 RailStatusCore(0x840)
  56  920        4        1 RailStatusSoc(0x840)
  57  543        4        1 A1Power(0x5, 0x6)
  58  912        4        1 RailStatusCore(0x0)
  59  920        4        1 RailStatusSoc(0x0)
  60  543        4        1 A1Power(0x5, 0xa)
  61  912        4        1 RailStatusCore(0x0)
  62  920        4        1 RailStatusSoc(0x0)
  63  543        4        8 A1Power(0x5, 0xa)
  64  543        4        1 A1Power(0x5, 0xb)
  65  912        4        1 RailStatusCore(0x0)
  66  920        4        1 RailStatusSoc(0x0)
  67  543        4        2 A1Power(0x5, 0xb)
  68  543        4        1 A1Power(0x5, 0xc)
  69  912        4        1 RailStatusCore(0x0)
  70  920        4        1 RailStatusSoc(0x0)
  71  543        4   910250 A1Power(0x5, 0xc)
```

Note that `0x840` for the `STATUS_WORD` denotes what we can expect when we're off; from `humility pmbus` output:

```
0x79 STATUS_WORD               0x0840
     |
     | b15    0b0 = no fault                 <= OutputVoltageFault
     | b14    0b0 = no fault                 <= OutputCurrentFault
     | b13    0b0 = no fault                 <= InputFault
     | b12    0b0 = no fault                 <= ManufacturerFault
     | b11    0b1 = POWER_GOOD not set       <= PowerGoodStatus
     | b10    0b0 = no fault                 <= FanFault
     | b9     0b0 = no fault                 <= OtherFault
     | b8     0b0 = no fault                 <= UnknownFault
     | b7     0b0 = no fault                 <= Busy
     | b6     0b1 = power is off             <= Off
     | b5     0b0 = no fault                 <= OutputOvervoltageFault
     | b4     0b0 = no fault                 <= OutputOvercurrentFault
     | b3     0b0 = no fault                 <= InputUndervoltageFault
     | b2     0b0 = no fault                 <= TemperatureFault
     | b1     0b0 = no fault                 <= CMLFault
     | b0     0b0 = no fault                 <= NoneOfTheAbove
     +-----------------------------------------------------------------------
```

This shows that the rail turned on when the FPGA hit the enable, which should only be possible if it were directed on -- or had never been properly directed off as part of the transition to A2. 

Additional instrumentation was added to record `OPERATION`, and the loop was tightened to a A2/A0 loop instead of a full reboot loop.  After about an hour, the transition to A0 failed ([dump](https://github.com/oxidecomputer/hubris/files/10775289/hubris.core.BRM42220023-hang.3.zip)):

```
...
  26  499      235        1 PGStatus { b_pg: 0xff, c_pg: 0x3, nic: 0x7f }
  27  505      235        1 SMStatus { a1: 0x5, a0: 0xc }
  28  510      235        1 PowerControl(0x3)
  29  514      235        1 InterruptFlags(0x0)
  30  914      235        1 RailOperationCore(0x88)
  31  919      235        1 RailStatusCore(0x840)
  32  927      235        1 RailOperationSoc(0x88)
  33  932      235        1 RailStatusSoc(0x840)
  34  488      235        1 UpdateState(A2)
  35  646      235        1 A2
...
```

When a rail is enabled, we expect its `OPERATION` to be 0x88; again from `humility pmbus` output:

```
0x01 OPERATION                 0x88
     |
     | b7     0b1 = output on                <= OnOffState
     | b6     0b0 = powers down immediately  <= TurnOffBehavior
     | b5:4   0b0 = set by VOUT_COMMAND      <= VoltageCommandSource
     | b3:2   0b10 = act upon margin faults  <= MarginFaultResponse
     | b1     0b0 = not set                  <= TransitionControl
     +-----------------------------------------------------------------------
```

The ring buffer very clearly shows that when we disabled the rail (which should have set `OPERATION` from 0x88 to 0x8), we did, in fact, disable the rail. (!)

Putting a logic analyzer on the mid bus, and then disabling all other activity on that bus did not reproduce the problem after ~8 hours.  This itself was very telling; was the other bus activity related? Indeed, keeping `thermal` disabled but running the `power` task at 1000X its normal rate of 1 Hz resulted in the problem being hit after only a few minutes.

The logic analyzer tells all; here is our (failed) enabling of the rail, which points immediately to the problem:

2023-02-18-11-52-32_3258x494

We are reading 0x88 from `OPERATION` (command 0x01) for the SoC power controller (device 0x5b) and then writing it -- but in between the read and the write, there is a `PAGE` command to select rail 1, which is very much **not** the rail we are powering off! So instead of powering off `VDDCR_SOC` (rail 0), we are in fact powering off `VDD_MEM_EFGH` (rail 1).  Now, this is meaningless because this rail is not set to power by the `OPERATION` command -- but it also means that our disable of `VDDCR_SOC` (and, through similar fashion, `VDD_VCORE`) is being dropped.

Why is rail 1 being selected?  Because we have interleaved with the `power` task:  the fundamental problem is that both `power` and `gimlet_seq` are operating on the RAA229618 VRs, but with no coordination between them. In particular, the selection of the rail and the subsequent operation upon it must be atomic -- and they very much aren't. To quote [Robert Ledru](https://medium.com/exploring-history/the-police-detective-who-caught-himself-7b7bcda1caf2): "you need look no further for your killer. He is right here!"

Knowing all of this, it is hardly surprising that this happens so rarely:  it needs a pretty precise interleaving between the `power` task running at 1 Hz and the A0 to A2 transition which cannot run at even that frequency.

In terms of fixing this, there are several ways to do it.  But given that the `power` task already must ask the sequencer about the current power state, it seems the simplest to introduce a time-based lease for the RAA229618s as administered by the sequencer:  for the (relatively common) querying of properties, this will add minimal overhead to the `power` task as it will need to ask the sequencer periodically for a lease.  For the (much less common) change in power state, this will add a fixed delay as the sequencer will need to wait for the lease to expire before manipulating the RAA229618s itself.  (This only really needs to be done in the A0 to A2 transition:  the `power` task will not query the RAA229618s when the system is in A2.)

bcantrill:
Fixes #1160 (among several other issues) -- and adds the necessary SP support oxidecomputer/stlouis#321.

Draft commit comment:

```
spontaneous reset support + sequencer fixes + SB-RMI support

- Adds logic to recognize a reset from host software -- along with a
  new state, A0Reset, that denotes that the host has reset.

- Fixes #1160 by having the RAA229618 driver always form its rail
  selection and subsequent read or write as a single transaction with
  respect to the I2C server

- Fixes #1135 by greatly increasing the busy spin count

- Fixes an issue whereby we were inadvertently backdriving via
  SP_TO_SP3_NMI_SYNC_FLOOD_L when in A2, preventing reset of the GPIO
  expander (manifesting itself as being in A0PlusHP on an
  A0PlusHP->A2->A0 transition rather than the expected A0).

- Cleans up the sequncer to always set CLT_RST in NIC_CTRL when in A2

- Adds an SB-RMI driver, interface and task to be able to query
  host CPU state 
```


drv/gimlet-seq-server/src/main.rs
```diff
@@ -46,25 +47,42 @@ enum Trace {
...
+        rstn: u8,
+        pwrokn: u8,
+    },
     PowerControl(u8),
     InterruptFlags(u8),
     None,
 }
 
-ringbuf!(Trace, 64, Trace::None);
+ringbuf!(Trace, 100, Trace::None);
```
mkeeter, line 85: Nit: making this a power-of-two may be a little more efficient (mask versus division).

drv/i2c-devices/src/sbrmi.rs
```diff
@@ -0,0 +1,339 @@
...
+    BadRdmsr { code: ResponseCode },
+    RdmsrFailed { code: StatusCode },
+}
+
+impl From<Error> for ResponseCode {
+    fn from(err: Error) -> Self {
+        match err {
+            Error::BadRegisterRead { code, .. } => code,
+            Error::BadCpuidRead { code } => code,
+            _ => ResponseCode::BadResponse,
```
mkeeter, line 80: Doing a quick audit, it seems like different drivers do many different things when this conversion fails:
- `bmr491` (and many others) will panic
- `nvme_bmc` (and a few others) will return `ResponseCode::BadDeviceState`
- Here, we're returning `ResponseCode::BadResponse`

Philosophically, I think this should be `TryFrom`, but would require modifying I2C validation codegen.  I opened #1161 to track this.

drv/i2c-devices/src/sbrmi.rs
```diff
@@ -0,0 +1,339 @@
...
+impl Sbrmi {
+    pub fn new(device: &I2cDevice) -> Self {
+        Self { device: *device }
+    }
+
+    fn read_reg(&self, reg: Register) -> Result<u8, Error> {
+        match self.device.read_reg::<u8, u8>(reg.into()) {
+            Ok(buf) => Ok(buf),
+            Err(code) => Err(Error::BadRegisterRead { reg, code }),
+        }
```
mkeeter, line 132: Nit:
```suggestion
        self.device
            .read_reg::<u8, u8>(reg.into())
            .map_err(|code| Error::BadRegisterRead { reg, code })
```

drv/i2c-devices/src/sbrmi.rs
```diff
@@ -0,0 +1,339 @@
...
+
+            roffs += banksize;
+        }
+
+        Ok(rval)
+    }
+
+    pub fn alert(&self) -> Result<[u8; 16], Error> {
+        //
+        // Have you ever played 52-bit Pickup?  Unlike the enabled bits, the
```
mkeeter, line 165: 🤣

drv/i2c-devices/src/sbrmi.rs
```diff
@@ -0,0 +1,339 @@
...
+        //   0x50      MceStat[3:0] = Threads[112,96,80,64]
+        //   0x51      MceStat[3:0] = Threads[113,97,81,65]
+        //   0x52      MceStat[3:0] = Threads[114,98,82,66]
+        //   0x53      MceStat[3:0] = Threads[115,99,83,67]
+        //   0x54      MceStat[3:0] = Threads[116,100,84,68]
+        //
+        let mut rval = [0u8; 16];
+        let banksize = 16usize;
+
+        for (bank, base) in [0x10, 0x50].iter().enumerate() {
```
mkeeter, line 196: Is it harmless to read extra registers that aren't in the table above (`0x55-0x5F`)?

If not, we could do something like
```suggestion
        for (bank, (base, banksize0)) in
            [(0x10, 16), (0x50, 5)].iter().enumerate()
        {
```

drv/sbrmi-api/src/lib.rs
```diff
@@ -0,0 +1,59 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Client API for the MeanWell driver.
```
mkeeter, line 5: 🤔

drv/sbrmi/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+    Rdmsr(u32),
+    RdmsrError(drv_i2c_devices::sbrmi::Error),
+    RdmsrOk,
+}
+
+ringbuf!(Trace, 16, Trace::None);
+
+fn sbrmi() -> Sbrmi {
+    let devs = i2c_config::devices::sbrmi(I2C.get_task_id());
+    Sbrmi::new(&devs[0])
```
mkeeter, line 37: It would be a little more efficient to store this in the `ServerImpl`, so you don't need to rebuild it every time.

mkeeter:
I don't _love_ the timed-lease strategy ("timeouts, timeouts, always wrong...").

However, doing explicit lock + release is tricky: the release would naturally be part of the Idol API, but we want the `gimlet-seq` task to pause mid-transition, so we can't naturally return to the Idol dispatch.

Here's a quick sketch of an alternative:
- Replace `lease_devices` with `lock`, which stores the calling task ID in an `Option<TaskId>` within `gimlet-seq`
- Add a new `unlock` API, which sets that value to `None`
- During the critical period, if `gimlet-seq` is locked, then do a closed receive, _only_ listening for the `unlock` operation coming from the locking task (and faulting the caller if it tries to call any other IPCs, using `REPLY_FAULT`)
- Once we hear the `unlock` operation, then continue with the power-up sequence.  We know that other tasks won't be able to interfere with power-up, because they'll call `lock()`, and `gimlet-seq` won't return to Idol dispatch until it's done powering up.

If this seems plausible, [here's a `diff`](https://gist.github.com/mkeeter/ae085ae13a0c9207bb1e60cb37adb1a2) that implements this strategy.  It's totally untested, but I think it has the right ideas.

bcantrill:
> I don't _love_ the timed-lease strategy ("timeouts, timeouts, always wrong...").

While I agree in the abstract, we already have a bunch of time-based waiting in the sequencer path -- and I really want to avoid additional complexity here. So in this case, I'm not sure that the timeouts *are* wrong -- and in this case, I think they are quite a bit simpler than the alternative

drv/gimlet-seq-server/src/main.rs
bcantrill: It does -- but the deeper history has been really, really helpful for debugging sequencer issues.

cbiffle:
We had a brief call to talk through the requirements here, and I've suggested a different approach.

The core problem here, and the reason why the tasks are capable of achieving a destructive interleaving, is that operations on the Renesas VR require two I2C transactions:

1. A write to choose which rail to affect in a subsequent transaction.
2. Another write or write-read or read to affect that rail.

Our current I2C IPC API only exposes a single operation, write-then-read. However, we've hit requirements for more complex I2C transaction sequencing in the past, particularly around the Konami Code (which thankfully, afaict, we are not shipping). If, however, we were to extend the API to support a more general notion of uninterruptible transaction consisting of some combination of write and read phases, this would be incredibly general, and would fix the power-sequencer interleaving issue as a side effect -- as I understand it.

As a sketch of a totally-general operation to either replace or supplement write-then-read, consider:

- The new IPC message contains the target address, as it does today, but does not explicitly indicate write or read.
- The IPC can take an _arbitrary_ number of leases.
- The write bit on each lease determines whether it is a read (note: this appears backwards, because on an I2C read we must be able to write the lease, and vice versa).
- The operations described by the sequence of leases will be carried out one after the other without yielding the bus or switching muxes.

This would let you encode arbitrarily complex sequences of operations to the same I2C address as long as you don't require a STOP condition anywhere in the middle -- for instance, you could do "write 3 bytes, repeated start, write 2 bytes, repeated start, read 4 bytes, repeated start, write 1 byte, stop." The current write-then-read operation falls out of this as a special case.

Idol cannot currently express this operation, unfortunately, but the I2C server isn't implemented using Idol, so no big deal. (In fact, I was pretty sure we _used to have_ an operation like this, which is why the I2C server doesn't use Idol, but I may have been making that up.)

drv/i2c-devices/src/sbrmi.rs
bcantrill: Ha -- the comment was incomplete!  I had neglected to include the second page of the table; thanks for spotting that!

drv/i2c-devices/src/sbrmi.rs
```diff
@@ -0,0 +1,339 @@
...
+    pub fn cpuid(
+        &self,
+        thread: u8,
+        eax: u32,
+        ecx: u32,
+    ) -> Result<CpuidResult, Error> {
+        let eax = eax.to_le_bytes();
+        let mut rval = CpuidResult {
+            ..Default::default()
+        };
```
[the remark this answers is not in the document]
bcantrill: Thank you!

drv/sbrmi/src/main.rs
bcantrill: Done

drv/sbrmi-api/src/lib.rs
bcantrill: 🤦‍♂️

drv/i2c-devices/src/sbrmi.rs
bcantrill: Done

drv/gimlet-seq-server/src/main.rs
bcantrill: We had not had room for 128 originally, but with some of the other changes, now we do; bumping this to 128.

drv/i2c-devices/src/sbrmi.rs
bcantrill: Thanks for filing that!

bcantrill:
> We had a brief call to talk through the requirements here, and I've suggested a different approach.
> 
> The core problem here, and the reason why the tasks are capable of achieving a destructive interleaving, is that operations on the Renesas VR require two I2C transactions:
> 
> 1. A write to choose which rail to affect in a subsequent transaction.
> 2. Another write or write-read or read to affect that rail.
> 
> Our current I2C IPC API only exposes a single operation, write-then-read. However, we've hit requirements for more complex I2C transaction sequencing in the past, particularly around the Konami Code (which thankfully, afaict, we are not shipping). If, however, we were to extend the API to support a more general notion of uninterruptible transaction consisting of some combination of write and read phases, this would be incredibly general, and would fix the power-sequencer interleaving issue as a side effect -- as I understand it.
> 
> As a sketch of a totally-general operation to either replace or supplement write-then-read, consider:
> 
> * The new IPC message contains the target address, as it does today, but does not explicitly indicate write or read.
> * The IPC can take an _arbitrary_ number of leases.
> * The write bit on each lease determines whether it is a read (note: this appears backwards, because on an I2C read we must be able to write the lease, and vice versa).
> * The operations described by the sequence of leases will be carried out one after the other without yielding the bus or switching muxes.
> 
> This would let you encode arbitrarily complex sequences of operations to the same I2C address as long as you don't require a STOP condition anywhere in the middle -- for instance, you could do "write 3 bytes, repeated start, write 2 bytes, repeated start, read 4 bytes, repeated start, write 1 byte, stop." The current write-then-read operation falls out of this as a special case.
> 
> Idol cannot currently express this operation, unfortunately, but the I2C server isn't implemented using Idol, so no big deal. (In fact, I was pretty sure we _used to have_ an operation like this, which is why the I2C server doesn't use Idol, but I may have been making that up.)

I stopped short of making this truly arbitrary, but instead phrased it in terms of write/read pairs -- and allowing any number of these to be in a single I2C server transaction.  (This brought me straight into the grip of #1135, so I fixed that as well.)  I modified the RAA229618 and ISL68224 driver (the former being the source of the problem and the latter being a very similar driver), but stopped short of updating the mwocp68 driver:  we do not currently share this device across tasks (indeed, it's only used by the PSC) and it's quite a bit more involved to convert this.

 The draft commit comment has been updated accordingly!

drv/i2c-api/src/lib.rs
```diff
@@ -510,6 +512,91 @@ impl I2cDevice {
...
+            &Marshal::marshal(&(
+                self.address,
+                self.controller,
+                self.port,
+                self.segment,
+            )),
+            response.as_bytes_mut(),
+            &[
+                Lease::from(buffer),
+                Lease::from(&empty[0..0]),
```
mkeeter, line 544: It would be nice to be able to write
```suggestion
                Lease::from(&[]),
```

To make that work, [insert this code right here](https://github.com/oxidecomputer/hubris/blob/master/sys/userlib/src/lib.rs#L94)
```rust
impl<'a> From<&'a [u8; 0]> for Lease<'a> {
    fn from(x: &'a [u8; 0]) -> Self {
        Self::read_only(x)
    }
}
```

mkeeter: Alternatively,
```suggestion
                Lease::read_only(&[]),
```
works already, without need to add a new `impl From`

drv/i2c-devices/src/isl68224.rs
```diff
@@ -81,23 +81,16 @@ impl Isl68224 {
...
-    }
-
     pub fn turn_off(&self) -> Result<(), Error> {
-        self.set_rail()?;
-        let mut operation = pmbus_read!(self.device, OPERATION)?;
-        operation.set_on_off_state(OPERATION::OnOffState::Off);
-        pmbus_write!(self.device, OPERATION, operation)
+        let mut op = pmbus_rail_read!(self.device, self.rail, OPERATION)?;
+        op.set_on_off_state(OPERATION::OnOffState::Off);
+        pmbus_write!(self.device, OPERATION, op)
```
mkeeter, line 87: Should this be `pmbus_rail_write!`?

drv/i2c-devices/src/isl68224.rs
```diff
@@ -81,23 +81,16 @@ impl Isl68224 {
...
     }
 
     pub fn turn_on(&self) -> Result<(), Error> {
-        self.set_rail()?;
-        let mut operation = pmbus_read!(self.device, OPERATION)?;
-        operation.set_on_off_state(OPERATION::OnOffState::On);
-        pmbus_write!(self.device, OPERATION, operation)
+        let mut op = pmbus_rail_read!(self.device, self.rail, OPERATION)?;
+        op.set_on_off_state(OPERATION::OnOffState::On);
+        pmbus_write!(self.device, OPERATION, op)
```
mkeeter, line 93: Same here

drv/i2c-devices/src/lib.rs
```diff
@@ -105,6 +150,40 @@ macro_rules! pmbus_write {
...
+        payload[0] = $dev::$cmd::CommandData::code();
+        $data.to_slice(&mut payload[1..]);
+
+        match $device.write_write(&rpayload, &payload) {
+            Err(code) => Err(Error::BadWrite {
+                cmd: $dev::$cmd::CommandData::code(),
+                code,
+            }),
+            Ok(_) => Ok(()),
+        }
```
mkeeter, line 167: I spent too much time trying to figure out way to remove this duplication:
```suggestion
        use $dev::{$cmd, PAGE};
        pmbus_rail_write!($device, $rail, $cmd, $data)
```

(the double `{{...}}` already make this is a local scope, so the `use` doesn't escape)

drv/i2c-devices/src/lib.rs
```diff
@@ -75,6 +76,50 @@ macro_rules! pmbus_read {
...
+            Err(code) => Err(Error::BadRead {
+                cmd: $dev::$cmd::CommandData::code(),
+                code,
+            }),
+        }?) {
+            Some(data) => Ok(data),
+            None => Err(Error::BadData {
+                cmd: $dev::$cmd::CommandData::code(),
+            }),
+        }
```
mkeeter, line 119:
```suggestion
        use $dev::{$cmd, PAGE};
        pmbus_rail_read!($device, $rail, $cmd)
```
(see explanation below; this could apply to every `pmbus_xyz!` macro)

drv/i2c-devices/src/isl68224.rs
bcantrill: God, yes -- thank you!

drv/i2c-api/src/lib.rs
bcantrill: I love that alternative -- will change this throughout that file.

drv/i2c-devices/src/lib.rs
bcantrill: That is a huge improvement!

drv/i2c-devices/src/isl68224.rs
bcantrill: Got it

mkeeter (approved):

drv/i2c-devices/src/lib.rs
bcantrill: Done -- thank you again!

PR #1164, opened 2023-02-21
hubris#1163 gimlet NMI output GPIO back-powers A0 rail

wesolows:
Testing notes: measured `SP3_VDD33_SENSE` at TP454 in A2 and A0 on rev B s/n 15.  Before this change, it was at 1.87 V in A2 and 3.34 V in A0.  After this change, it is at 0.036 V in A2 and 3.34 V in A0.  Verified that no NMI is asserted by default when the host boots.  Verified default disposition of pin J:2 with `hiffy gpio -i` is 1.  Then asserted NMI via hiffy (`gpio -p J:2 -r`) which worked as expected.


mkeeter (approved):
LGTM.

@bcantrill also has the same change staged in #1151, but this can go in first and that will rebase cleanly.

Issue #1157, opened 2023-02-17
Hubris component-level version


mkeeter:
This is roughly based on [a 2023-02-16 meeting](https://docs.google.com/document/d/1a6Ah_dI2GKbDc8pyip1bdJO7qg4jD9HFH9xkoFamC8Q/edit#), plus my proposal at the end.

## Existing version-ish things
We have a few version-ish things in the Hubris archive already:
- `HEADER: ImageHeader` is placed at a known location in flash (`.header`,  directly after the vector table; `0x298` on STM32H7, `0x130` on LPC55).
    - It includes a `version` and `epoch` (both `u32`), which will be used for boot selection (depending on TBD policy)
    - At rest, it can be read either by knowing its address (MCU-dependent) or by searching for a magic number (`HEADER_MAGIC`); the latter _could_ be error-prone, but is probably fine.
    - It is not ever read from a running system, though it _could_ be.
    - This header also always includes `sau_entries`, which is only relevant for the RoT.  This is a little odd.
- `HUBRIS_IMAGE_ID` is a `u64` built into the `kconfig.rs` at compile time.
    - It is tagged with `no_mangle`, but is not placed at any particular address in flash memory.
    - At rest, it can only be read if you have the associated Hubris archive (since you need to know its address in flash).
    - In a running system, it can be read with the `read_image_id` kernel IPC operation.  At the moment, it is broadcast by `udpbroadcast` and included in the SP state returned by `control-plane-agent` (`hubris_archive_id`).
    - This image ID is built from a hash of many things:
        - Output sections, which includes object code for all tasks
        - Kernel configuration
        - Allocations
    - The image ID **does not include** the kernel's object code; changes to the kernel are invisible to this ID unless they change allocations or memory layout.
        - ⚠️ We should probably fix this.
- The Git commit is included in the Hubris archive as a file named `git-rev`
    - This is a full Git hash, with an optional `-dirty` suffix
    - It is not included anywhere in the image.

## Conditions under which we may want to read a version
There are a few situations where we want to read a version-ish thing, roughly in order of difficulty:
- Asking a running system to report it through a well-known API
- Getting it from a Hubris archive (offline, without a running system)
- Extracting it from a binary file (or running SP), given access to the original Hubris archive
- Extracting it from a binary file (or running SP), **without** access to the original Hubris archive

(Think of "attached-with-an-ST-Link" and "RoT-debugging-SP" as cases of "extracting it from a binary file", since we can read arbitrary memory locations)

[...]

PR #1165, opened 2023-02-22
Add caboose to build system

mkeeter:
This PR does ~two~ ~three~ four things:

- Reserves space for a `caboose` (#1157) when building an image
- Adds the caboose size to the `ImageHeader`
- Pulls a bunch of generic Hubris-archive-manipulatey stuff into the new `hubtools` repo.  We'll soon be manipulating Hubris archives from `permission-slip`, so want to move this code out.
- Adds a new (very small) G0 image to test the caboose in its `app.toml`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2295,6 +2295,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "hubtools"
+version = "0.1.0"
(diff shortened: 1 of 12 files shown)


mkeeter:
Upon further discussion, I'm going to cancel this – we've got some good ideas about how to avoid relying on `ImageHeader`, which is already on the chopping block.

PR #1169, opened 2023-02-27
Add the Hubris Caboose

### LINKED ISSUE #1157: Hubris component-level version

mkeeter:
Closes #1157, which has discussion of this design.

This PR does several things!

Primarily, it adds an optional `caboose` to Hubris images.  The caboose is located at the end of flash and contains (1) two words indicating that it's a caboose and (2) extra space for user data.

It is documented in a newly added section of the [Hubris manual](https://github.com/oxidecomputer/hubris/blob/caboose/doc/guide/caboose.adoc).

This also accomplishes _some_ of the goals of #1004 , namely having a "commit point" at the end of flashing.  For the moment, this implementation still requires having an `ImageHeader` after the vector table.  The caboose is after everything else in flash, but is not at the last page; this means that we need a way to find it, which is specified in the `ImageHeader`.

(We could work around this by using [NXP-style overloading spare vector table entries](https://github.com/oxidecomputer/hubris/issues/1157#issuecomment-1440878674), but this ends up feeling hacky on the STM32 and tightly couples us to a Cortex-M-flavored MCU)

In addition, this PR moves common Hubris archive functionality to a new [`hubtools` repository](https://github.com/oxidecomputer/hubtools).  This is because both Hubris `xtask` and `permission-slip` want to do certain manipulations of Hubris archives – most notably, going from `srec` to all the other binary formats.

`hubtools` contains this common functionality, as well as a minimal CLI to edit the caboose.

Finally, the PR adds a new `caboose-reader` task, which is a minimal task to read from the caboose.  It's mostly useful for development and debugging.

### DIFF
--- a/app/demo-stm32g0-nucleo/app-g070-mini.toml
+++ b/app/demo-stm32g0-nucleo/app-g070-mini.toml
@@ -1,0 +1,33 @@
+# Tiny G0 image, useful for making small Humility archives
+name = "demo-stm32g070-nucleo"
+target = "thumbv6m-none-eabi"
+chip = "../../chips/stm32g0"
+memory = "memory-g070.toml"
+board = "stm32g070"
+stacksize = 944
+
+[kernel]
+name = "demo-stm32g0-nucleo"
+requires = {flash = 18048, ram = 1632}
+features = ["g070"]
+stacksize = 640
+
+[caboose]
+region = "flash"
+size = 128
+
+[tasks.jefe]
+name = "task-jefe"
+priority = 0
+max-sizes = {flash = 4096, ram = 512}
+start = true
+features = ["log-null"]
+stacksize = 352
+notifications = ["fault", "timer"]
+
+[tasks.idle]
+name = "task-idle"
+priority = 5
+max-sizes = {flash = 128, ram = 64}
+stacksize = 64
+start = true
--- a/app/demo-stm32g0-nucleo/app-g070-mini.toml
+++ b/app/demo-stm32g0-nucleo/app-g070-mini.toml
@@ -14,7 +14,7 @@ stacksize = 640
 
 [caboose]
 region = "flash"
-size = 128
+size = 256
 
 [tasks.jefe]
 name = "task-jefe"
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -338,12 +338,16 @@ pub fn package(
         // Add an empty output section for the caboose
         if let Some(caboose) = &cfg.toml.caboose {
             let (_, caboose_range) = allocs.caboose.as_ref().unwrap();
-            // BTreeMap<u32, LoadSegment>,
+            // The final word in the caboose is the caboose length, so that we
+            // can decode the caboose start by looking at it
+            let mut caboose_data = vec![0xFF; caboose.size as usize];
+            caboose_data[caboose.size as usize - 4..]
+                .copy_from_slice(&caboose.size.to_le_bytes());
             all_output_sections.insert(
                 caboose_range.start,
                 LoadSegment {
                     source_file: "caboose".into(),
-                    data: vec![0xFF; caboose.size as usize],
+                    data: caboose_data,
                 },
             );
         }
@@ -1045,12 +1049,6 @@ fn update_image_header(
                     epoch: cfg.toml.epoch,
                     magic: abi::HEADER_MAGIC,
                     total_image_len: len as u32,
-                    caboose_size: cfg
-                        .toml
-                        .caboose
-                        .as_ref()
-                        .map(|c| c.size)
-                        .unwrap_or(0),
                     ..Default::default()
                 };
 
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -293,7 +293,7 @@ pub fn package(
         // Build all relevant tasks, collecting entry points into a HashMap.  If
         // we're doing a partial build, then assign a dummy entry point into
         // the HashMap, because the kernel kconfig will still need it.
-        let entry_points: HashMap<_, _> = cfg
+        let mut entry_points: HashMap<_, _> = cfg
             .toml
             .tasks
             .keys()
@@ -326,6 +326,16 @@ pub fn package(
         // is included in the total image size that's patched into the kernel
         // header.
         if let Some(caboose) = &cfg.toml.caboose {
+            if (caboose.size as usize) < std::mem::size_of::<u32>() * 2 {
+                bail!("caboose is too small; must fit at least 2x u32");
+            }
+
+            for t in &caboose.tasks {
+                if !cfg.toml.tasks.contains_key(t) {
+                    bail!("caboose specifies invalid task {t}");
+                }
+            }
+
             let (_, caboose_range) = allocs.caboose.as_ref().unwrap();
             // The caboose has the format
             // [CABOOSE_MAGIC, ..., MAX_LENGTH]
@@ -349,6 +359,7 @@ pub fn package(
                     data: caboose_data,
                 },
             );
+            entry_points.insert("caboose".to_string(), caboose_range.start);
         }
 
         // Build the kernel!
@@ -1943,6 +1954,22 @@ pub fn make_kconfig(
         );
     }
 
+    if let Some(c) = &toml.caboose {
+        flat_shared.insert(
+            "caboose".to_string(),
+            build_kconfig::RegionConfig {
+                base: entry_points["caboose"],
+                size: c.size,
+                attributes: build_kconfig::RegionAttributes {
+                    read: true,
+                    write: false,
+                    execute: false,
+                    special_role: None,
+                },
+            },
+        );
+    }
+
     let mut used_shared_regions = BTreeSet::new();
 
     for (i, (name, task)) in toml.tasks.iter().enumerate() {
@@ -1961,12 +1988,20 @@ pub fn make_kconfig(
 
         // Mark off the regions this task uses.
         for region in &task.uses {
-            used_shared_regions.insert(region);
+            used_shared_regions.insert(region.as_str());
         }
 
         // Prep this task's shared region name set.
-        let shared_regions: std::collections::BTreeSet<String> =
+        let mut shared_regions: std::collections::BTreeSet<String> =
             task.uses.iter().cloned().collect();
+
+        // Allow specified tasks to use the caboose
+        if let Some(caboose) = &toml.caboose {
+            if caboose.tasks.contains(&name) {
+                used_shared_regions.insert("caboose");
+                shared_regions.insert("caboose".to_owned());
+            }
+        }
 
         let owned_regions = task_allocations[name].iter()
             .map(|(out_name, range)| {
@@ -2077,7 +2112,7 @@ pub fn make_kconfig(
     }
 
     // Pare down the list of shared regions.
-    flat_shared.retain(|name, _v| used_shared_regions.contains(name));
+    flat_shared.retain(|name, _v| used_shared_regions.contains(name.as_str()));
 
     Ok(build_kconfig::KernelConfig {
         irqs,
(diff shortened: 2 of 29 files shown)


plotnick:
I'm not qualified to comment on this part of the implementation, but hubtools looks like exactly the thing that `permslip-server` will need to do version stamping. Thanks for splitting that out and making it easy!

idl/caboose.idol
```diff
@@ -0,0 +1,43 @@
...
+            leases: {
+                "data": (type: "[u8]", write: true),
+            },
+            reply: Result(
+                ok: "u32",
+                err: CLike("CabooseError"),
+            ),
+            idempotent: true,
+        ),
+    }
```
lzrd, line 42: Can the API be used to iterate over the entire contents of the caboose?

lzrd (approved):
Looks good. Couple of minor questions that don't stop merge.

app/demo-stm32g0-nucleo/app-g070-mini.toml
```diff
@@ -0,0 +1,33 @@
...
+board = "stm32g070"
+stacksize = 944
+
+[kernel]
+name = "demo-stm32g0-nucleo"
+requires = {flash = 18048, ram = 1632}
+features = ["g070"]
+stacksize = 640
+
+[caboose]
```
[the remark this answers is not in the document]
mkeeter: Yup, that's correct!

idl/caboose.idol
mkeeter: Yes, in that it returns an address range.

doc/guide/caboose.adoc
```diff
@@ -0,0 +1,77 @@
...
+The caboose's length is included in the `total_image_len` field of
+`abi::ImageHeader`.  Because the caboose is located at the end of flash, its
+presence and size can be determined as follows:
+
+* Read total image length from the `ImageHeader`
+** At runtime, this is a variable that can be read by the kernel
+** At rest, the image header is at a known location (depending on
+   microcontroller) and includes a distinctive magic number
+   (`abi::HEADER_MAGIC`)
+* Read the final word of the image, which may be the caboose length
```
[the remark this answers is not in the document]
mkeeter: I think I want to defer that to the caboose contents, e.g. if we're storing TLV-C (which is our expected format), there's a checksum built into the data format.

build/xtask/src/dist.rs
```diff
@@ -318,6 +320,48 @@ pub fn package(
...
+            for t in &caboose.tasks {
+                if !cfg.toml.tasks.contains_key(t) {
+                    bail!("caboose specifies invalid task {t}");
+                }
+            }
+
+            let (_, caboose_range) = allocs.caboose.as_ref().unwrap();
+            // The caboose has the format
+            // [CABOOSE_MAGIC, ..., MAX_LENGTH]
+            // where all words in between are initialize to u32::MAX
```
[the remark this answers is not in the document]
mkeeter: Yup, `u32::MAX` is because we have to pick _something_, and that's free on the SP after the section is erased.

It's worth noting that the caboose will _still_ be written even if no one has modified the archive to put data in there; it will just be all `0xFF`.  This means that on the LPC55, we won't be leaving this region in the troublesome erased-but-not-written state.

build/xtask/src/sizes.rs
```diff
@@ -136,13 +136,25 @@ fn build_memory_map<'a>(
...
+            toml.kernel.requires.clone(),
+            allocs.kernel.clone(),
         )))
+        .chain(allocs.caboose.iter().map(|(region, size)| {
+            let mut alloc = BTreeMap::new();
+            alloc.insert(region.clone(), size.clone());
+            let mut requires = IndexMap::new();
+            requires
+                .insert(region.clone(), toml.caboose.as_ref().unwrap().size);
+            ("-caboose-", requires, alloc)
```
jgallagher, line 156: Other references to this region just call it `"caboose"` - does it matter that this one is `"-caboose-"`?

doc/guide/caboose.adoc
```diff
@@ -0,0 +1,77 @@
+[#caboose]
```
jgallagher, line 1: This is a great doc :+1:

idl/caboose.idol
```diff
@@ -0,0 +1,43 @@
...
+                "data": (type: "[u8]", write: true),
+            },
+            reply: Result(
+                ok: "u32",
+                err: CLike("CabooseError"),
+            ),
+            idempotent: true,
+        ),
+        "get_key_by_u32": (
+            doc: "Scans the caboose for a key with the given tag",
```
jgallagher, line 29: Is calling this equivalent to calling `get_key_by_tag(tag.to_le_bytes())`? If so, do we need both?

task/caboose-reader/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+            base: self.pos.start,
+            size: self.pos.len() as u32,
+        };
+
+        let mut reader = TlvcReader::begin(reader)
+            .map_err(|_| CabooseError::TlvcReaderBeginFailed)?;
+        while let Ok(Some(chunk)) = reader.next() {
+            if chunk.header().tag == name {
+                // TODO: verify checksum
+                // TODO: make this not one byte at a time
```
jgallagher, line 88: Just double-checking - did you want to handle these TODOs before merging?

jgallagher (approved):

idl/caboose.idol
mkeeter: Yup, and that's how it's implemented under the hood.  We need both because Humility doesn't know how to send strings yet 😬

build/xtask/src/sizes.rs
mkeeter: This is for pretty-printing, so it shows up in the table as a special region:
```
flash:
      ADDRESS  |    PROGRAM     |  USED |  SIZE | LIMIT
    0x08000000 | kernel         | 18048 | 18048 | (fixed)
    0x08004680 | idle           |    68 |   128 | 128
    0x08004700 | -padding-      |    -- |  2304 |
    0x08005000 | jefe           |  3424 |  4096 | 4096
    0x08006000 | caboose_reader |  7424 |  8192 | 8192
    0x08008000 | hiffy          |  6568 |  8192 | 8192
    0x0800a000 | -caboose-      |   256 |   256 | 256
    0x0800a100 | --end--        |       |       |
```

task/caboose-reader/src/main.rs
mkeeter: Good question.  I think they're fine for now, since this is a debug task that isn't actually shipped; once we start using the caboose (presumably in `control-plane-agent`?), we can make better APIs for it.

PR #1184, opened 2023-03-06
SP default retries=3, fix CSn pulse to allow RoT to respond in time.

lzrd:
- Retries for all SpRot messages is set to three now.
  - There was a bug in the SP SpRot server where a CSn pulse to clear a pending ROT_IRQ (and RoT queued message) was done with ~20ms of CSn asserted and no wait time afterwards for the RoT to react to the CSn deassert. That has been fixed to be ~10ms assert followed by ~10ms delay before checking for ROT_IRQ deassert. Previously this could use up one of the delivery attempts and when attempts was set to 1 would result in message delivery failure.


andrewjstone (approved):
Nice find and fix!

PR #1185, opened 2023-03-06
Remove PSC rev A support.

cbiffle:
Do not upgrade this to a full PR and commit it until we're confident we've got all the rev As rounded up.


mkeeter:
This ended up being done in #1424

PR #1186, opened 2023-03-07
Switch RoT builds over to RFD374 stage0

cbiffle:
**Updated description:** This flips the switch and moves all RoT builds over to the new stage0. This implies

- Removes the stage0/ program and its build configs, including CI

- Moves the DICE logic from stage0 into an `lpc55-rot-startup` library,
  which is in turn called by the RoT main function(s) before starting
  the kernel. This performs the same logic as the previous code,
  including reading a header out of the binary; we can improve this
  later by making more stuff happen at build time.

- Revs the header magic number, ensuring that the original stage0 will
  not attempt to boot one of these images (since it would potentially do
  so in nonsecure-mode, which wouldn't work). This also has the
  potential to allow in-place upgrades, though this is not currently
  being tested.

Once this lands, new RoT builds will not be able to run with old stage0. However, old stage0 will also not _boot into_ new builds, so it's possible to update existing boards without it being _truly_ infuriating. I hope.

**Original description:** This is a sketch of pulling the DICE computation into RoT main, which is a requirement for collapsing the TZ separation.


flihp (approved):
Thanks for taking this on Cliff. Looks exactly like what we discussed the other day.

flihp (approved):
Read through this on my flight home. Apologies for holding this up, hopefully the rebase won't be too bad. The DICE stuff looks right. I'm less certain about how this integrates into the build. I'll dig through that for my own education tomorrow but no need to hold up this PR while I do that.

flihp:
I guess my confusion over integration into the build is explained by those pieces of code no longer being in stage0 :upside_down_face:

PR #1191, opened 2023-03-08
Hubtools update and archive cleanup

mkeeter:
Depends on https://github.com/oxidecomputer/hubtools/pull/2; once that's merged, we need to update the branch to point to `hubtools:main`.

This PR updates the `hubtools` dependency to use the new `RawHubrisImage` data structure.  In doing so, it eliminates the SREC and IHEX files from the Hubris archive.  This is fine; no one was using them anyways.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,7 +122,7 @@ dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-fe
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
 humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false }
-hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false }
+hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false, branch = "bin-mode" }
 idol = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
 idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
 lpc55_areas = { git = "https://github.com/oxidecomputer/lpc55_support", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,7 +122,7 @@ dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-fe
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
 humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false }
-hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false, branch = "bin-mode" }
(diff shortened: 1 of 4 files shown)


jgallagher (approved):

Cargo.toml
```diff
@@ -122,7 +122,7 @@ dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-fe
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
 humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false }
-hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false }
+hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false, branch = "bin-mode" }
```
cbiffle, line 125: So, by using a branch here, we're committing to retain that branch forever or the commits between here and any future switch to a `main` branch become unreproduceable. Comfortable with that?

build/xtask/src/dist.rs
```diff
@@ -439,21 +448,18 @@ pub fn package(
...
-                "final",
+            let image_bin =
+                std::fs::read(&cfg.img_file("combined.bin", image_name))?;
+            let start_addr = cfg
+                .toml
+                .memories(image_name)?
+                .get(&"flash".to_string())
+                .ok_or_else(|| anyhow!("failed to get flash region"))?
+                .start;
+            let raw_image = hubtools::RawHubrisImage::from_binary(
```
cbiffle, line 459: I think the comment above that talks about srec is wrong now?

build/xtask/src/dist.rs
```diff
@@ -808,6 +814,12 @@ fn check_rebuild(toml: &Config) -> Result<()> {
     Ok(())
 }
 
+#[derive(Debug, Hash)]
```
cbiffle, line 817: What's the purpose of the hash impl on this type? Will it be factored into image ID?

If yes, I'm concerned about including a machine-specific path in it.

Cargo.toml
mkeeter: I'm planning to switch back to `main` once the `hubtools` PR is merged!

build/xtask/src/dist.rs
mkeeter: This isn't changing anything; I'm just moving the data structure from `hubtools` back into Hubris proper, to simplify `hubtools`.

The `Hash` derivation is unchanged from before, e.g. [here's a commit predating `hubtools`](https://github.com/oxidecomputer/hubris/blob/f2bcd883b084c0481d3608c9e48209ac72e71360/build/xtask/src/dist.rs#L775-L779).

This _is_ included in image ID (and always have been), which is seeded with
```rust
fn build_kernel(
    cfg: &PackageConfig,
    allocs: &Allocations,
    all_output_sections: &mut BTreeMap<u32, LoadSegment>,
    all_memories: &IndexMap<String, Range<u32>>,
    entry_points: &HashMap<String, u32>,
    image_name: &str,
    secure: &Option<SecureData>,
) -> Result<(u32, BTreeMap<String, u32>)> {
    let mut image_id = fnv::FnvHasher::default();
    all_output_sections.hash(&mut image_id);
```
so we're not any worse than before!

build/xtask/src/dist.rs
mkeeter: Yup, fixed in 8ad21ecd7562001aa0510fe36746efbdc549e2e4

PR #1192, opened 2023-03-08
record V3P3_SYS_A0 Vout when in A2

### LINKED ISSUE #1188: SP could check readable SP3 rails in A2 and log them/report rails that don't bleed down.
This would help detect and regressions of pins accidentally bleeding across power domains and possibly give us useful information for debugging other issues.

bcantrill:
Fixes #1188.


nathanaelhuffman (approved):
lgtm

PR #1194, opened 2023-03-09
Add MGS API to read Caboose messages

mkeeter:
To read the Caboose – without necessarily creating a dedicated task – I moved the `CabooseReader` into the caboose API crate, renamed to `drv-caboose` (for symmetry with `drv-local-vpd`).

Then, I added support for the new MGS messages introduced in https://github.com/oxidecomputer/management-gateway-service/pull/66 ; we'll want to switch back to `main` after that PR is merged.

The implementation is a _little_ spicy!  The easiest API for `SpHandler::get_caboose_value` has the implementer returning a `&'static [u8]`, which is a slice of caboose data.  We can do that: `kipc::read_caboose_pos` will give us the caboose range, then we can search for the key and cast it into a slice using `core::slice::from_raw_parts`!

Still, I'd appreciate a few more eyes on that code.


task/caboose-reader/src/main.rs
```diff
@@ -72,33 +46,19 @@ impl idl::InOrderCabooseImpl for ServerImpl {
         name: [u8; 4],
         data: Leased<W, [u8]>,
     ) -> Result<u32, RequestError<CabooseError>> {
-        if self.pos.is_empty() {
-            return Err(CabooseError::MissingCaboose.into());
+        let reader = self
+            .caboose
```
jgallagher, line 50: Tiny nit - could `self` hold a `CabooseReader` instead? (Looks like `CabooseReader::new()` is a no-op so this doesn't really matter.)

jgallagher (approved):

Issue #1196, opened 2023-03-10
SP still comes up with self-assigned MAC address after #1171


rmustacc:
While doing testing with rack 1 (and older RoTs) we've seen this while doing a loop across the fleet of the form: `for f in BRM42220060 BRM42220041 BRM42220020 BRM42220035 BRM42220052 BRM42220087 BRM42220077 BRM42220019 BRM42220029 BRM42220075 BRM42220003 BRM44220002 BRM42220089 BRM42220053 BRM42220012 BRM42220068 BRM42220064 BRM42220048 BRM42220056 BRM42220054 BRM42220007 BRM42220042 BRM42220062 BRM42220076 BRM42220044 BRM42220010 BRM42220005 BRM42220024 BRM42220030 BRM42220061; do pfexec pilot sp exec -e 'update sp 0 /data/local/rack1/gimlet/c/sp.10ms/build-gimlet-c.zip' $f && pfexec pilot sp exec -e 'reset' $f; done`

I'm grabbing dumps and will upload them to catacomb, but recording this so we don't lose track. Assigning to MVP milestone right now. Feel free to move at your discretion.

cbiffle:
#1171 doesn't look like a fix for this, it changes the I2C driver to be more tolerant of errors selecting muxes. To avoid coming up with a self-assigned address, the netstack would need to block startup until VPD is available, which could be forever if the bus is locked.

Which in turn kind of suggests to me that a bus lock is happening.

rmustacc:
Dumps are on catacomb from 3 different occurrences in `/data/staff/core/rack1/` as `hubris.core.<sn>.badvpd`. Two spot checked had a similarly suspicious ringbuf entry:

```
humility: ring buffer drv_stm32xx_i2c_server::__RINGBUF in i2c_driver:
 NDX LINE      GEN    COUNT PAYLOAD
   1  162        1        1 Reset(I2C2, PortIndex(0x1))
   2  536        1        2 SegmentFailed(BusResetMux)
   3  162        1        1 Reset(I2C2, PortIndex(0x1))
   4  536        1        1 SegmentFailed(BusResetMux)
   5  536        1        1 SegmentFailed(BadMuxAddress)
   6  334        1        1 Error
   7  162        1        1 Reset(I2C3, PortIndex(0x0))
   0  334        2      952 Error
```

bcantrill:
[...]

PR #1204, opened 2023-03-13
add read_fpga_regs() entry point + sequencer register descriptions

bcantrill:
This is the prep work for https://github.com/oxidecomputer/humility/issues/334 and adds additional instrumentation per the discussion in #1196.


build/xtask/src/dist.rs
```diff
@@ -737,6 +737,73 @@ fn build_archive(cfg: &PackageConfig, image_name: &str) -> Result<()> {
...
+                Some(config) => match config.get(c) {
+                    Some(ordered_toml::Value::String(s)) => {
+                        //
+                        // This is a bit of a heavy hammer:  we need the
+                        // directory name for the task to find the file to be
+                        // copied into the archive, so we're going to iterate
+                        // over all packages to find the crate assocated with
+                        // this task.
+                        //
+                        use cargo_metadata::MetadataCommand;
```
mkeeter, line 767: Since this is so expensive, how about caching it, e.g. putting
```rust
let mut metadata = None;
for (name, task) in &cfg.toml.tasks {
    // ....
    let metadata = match metadata.as_ref() {
        Some(m) => m,
        None => {
            let d = MetadataCommand::new()
                .manifest_path("./Cargo.toml")
                .exec()?;
            metadata.get_or_insert(d)
        }
    };
```

drv/gimlet-seq-server/src/seq_spi.rs
```diff
@@ -142,6 +142,10 @@ impl<S: SpiServer> SequencerFpga<S> {
         let header = CmdHeader { cmd, addr };
         let header = header.as_bytes();
 
+        if data_out.len() > data.len() - header.len() {
+            return Err(spi_api::SpiError::BadTransferSize);
+        }
```
mkeeter, line 147: Does this check make the conditional below at L154 superfluous?

build/xtask/src/dist.rs
bcantrill: Yeah, that's a good idea. I went so far as to clock it (~160 ms), and decided that caching wasn't practical -- but it's just too easy to do and too offensive to not do.

drv/gimlet-seq-server/src/seq_spi.rs
bcantrill: Yes -- changed to an `assert!`

mkeeter (approved):

PR #1209, opened 2023-03-15
Sign images with hubtools

mkeeter:
This PR removes the Hubris-specific signing code in favor of calling into `hubtools`, which knows how to sign Hubris archives.

Depends on https://github.com/oxidecomputer/hubtools/pull/3/files and should be merged after it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,6 @@ ctrlc = { version = "3.1.5", default-features = false }
 derive_more = { version = "0.99", default-features = false, features = ["from", "into"] }
 digest = { version = "0.10", default-features = false }
 dunce = { version = "1.0.2", default-features = false }
-ecdsa = { version = "0.12.4", default-features = false, features = ["der"] }
 embedded-hal = { version = "0.2", default-features = false }
 enum-map = { version = "2.4.1", default-features = false }
 filetime = { version = "0.2.12", default-features = false }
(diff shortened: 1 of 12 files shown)


Cargo.toml
mkeeter, line 1: TODO: revert `hubtools`, `lpc55_areas`, and `lpc55_sign` repositories back to `main` after https://github.com/oxidecomputer/lpc55_support/pull/31 and https://github.com/oxidecomputer/hubtools/pull/3/files are merged.

Cargo.toml
```diff
@@ -43,7 +43,7 @@ byteorder = { version = "1.3.4", default-features = false }
 cargo_metadata = { version = "0.12.0", default-features = false }
 cfg-if = { version = "1", default-features = false }
 chrono = { version = "0.4", default-features = false }
-clap = { version = "3.0.14", default-features = false, features = ["derive"] }
+clap = { version = "3.0.14", default-features = false, features = ["std", "derive"] }
```
cbiffle, line 46: Passing default-features=false for the crates used by the std build system is kinda weird, if we removed that you wouldn't need std here.

cbiffle (approved):

Cargo.toml
mkeeter: After chatting in #oxide-hubris, I decided to leave this as-is; we have the build system and embedded crates all jumbled up, so it's easier to be consistent if we _always_ set `default-features = false`

PR #1219, opened 2023-03-18
support for task dumps

bcantrill:
Draft commit comment:

```
This work:

- Adds support for task dumps, including support for many H7 boards
  and one LPC55 board (the LPCXpresso55S69).  Extracting task dumps
  necessitates Humility v0.10.0 or later.

- Retools dump regions to be in terms of extern-regions

- Fixes an ambiguity in the Idol definition of the sequencer, whereby
  Humility did not know which definition of PowerState was being used.

- Because resolving the above ambiuguity also required some Humility
  work, this bumps the archive version to v8, which also requires
  Humility v0.10.0 or later.

- Polishes the build output to make clear how extern-regions are being
  used (i.e., by which task).
```


Cargo.lock
```diff
@@ -2281,9 +2281,11 @@ dependencies = [
 [[package]]
 name = "humpty"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/humpty#bcb02bd530beadd296a3a80b0ec2c4dbc00403c1"
+source = "git+https://github.com/oxidecomputer/humpty?branch=taskdumpy#aa0c374cfae8b772e21a23e8e3991a98a3046652"
```
mkeeter, line 2284: TODO: merge this first?

app/demo-stm32g0-nucleo/Cargo.toml
```diff
@@ -16,7 +16,7 @@ cortex-m-rt = { workspace = true }
 cfg-if = { workspace = true }
 stm32g0 = { workspace = true, features = ["rt"] }
 
-kern = { path = "../../sys/kern" }
+kern = { path = "../../sys/kern", features = ["no-dump"] }
```
mkeeter, line 19: This seems to violate the "features should be additive" rule of thumb; however, changing it would require editing a bunch of TOML files to add `features = ["dump"]`...

idl/jefe.idol
```diff
@@ -33,5 +33,32 @@ Interface(
...
+            encoding: Hubpack,
+            idempotent: true,
+        ),
+        "claim_dump_area": (
+            reply: Result(
+                ok: "DumpArea",
+                err: CLike("DumpAgentError"),
+            ),
+            encoding: Hubpack,
+            idempotent: true,
```
mkeeter, line 61: Is this actually idempotent?

sys/abi/src/lib.rs
```diff
@@ -479,6 +487,8 @@ pub enum Kipcnum {
     ReadImageId = 4,
     Reset = 5,
     ReadCaboosePos = 6,
+    GetTaskDumpRegion = 7,
```
mkeeter, line 490: It would be helpful to add these to `kipc.adoc`

task/dump-agent/src/main.rs
```diff
@@ -20,83 +20,48 @@ use userlib::*;
...
-
-        unsafe {
-            let header = area.address as *mut DumpAreaHeader;
-
-            if (*header).magic != DUMP_MAGIC {
-                panic!("bad dump magic!");
-            }
-
-            let nsegments = (*header).nsegments;
+        let area = self.dump_area(0)?;
```
mkeeter, line 51: This does an IPC call to `jefe`.

At the end of the IPC call, can the following sequence occur?

- `jefe` decides to return control to a different (higher-priority) task
-  That task crashes (returning to `jefe`)
- `jefe` saves a dump of the crashed task
- `area` is invalidated in this function

(I'm still reading, so this may not be possible!)

task/dump-agent/src/main.rs
```diff
@@ -148,9 +108,15 @@ impl idl::InOrderDumpAgentImpl for ServerImpl {
         let sprot = drv_sprot_api::SpRot::from(SPROT.get_task_id());
         let mut buf = [0u8; 4];
 
+        let area = self.dump_area(0)?;
+
+        if area.agent != humpty::DumpAgent::Task {
```
mkeeter, line 113: When would this happen?  Is this if `jefe` happened to use the 0th dump area?

task/dump-agent/src/main.rs
```diff
@@ -20,83 +20,48 @@ use userlib::*;
...
-                    written: size_of::<DumpAreaHeader>() as u32,
-                    length: area.length,
-                    agent_version: DUMP_AGENT_VERSION,
-                    dumper_version: DUMPER_NONE,
-                    next,
-                }
-            }
+    fn initialize(&self) -> Result<(), DumpAgentError> {
+        let jefe = Jefe::from(JEFE.get_task_id());
+        jefe.initialize_dump_areas()
```
mkeeter, line 33: Right now, this means that `jefe` can't dump crashing tasks until `dump-agent` starts up.

How about letting `jefe` initialize the dump areas itself, then call `Jefe.get_state` here (which will only return once `jefe` enters the Idol runtime loop)?

bcantrill: We also use this entry point to reinitialize the dump areas -- but you're absolutely right that Jefe should do this on its own without relying on the dump agent.  Will change the name of the Jefe entry point to make clear that a caller can reinitialize the dump areas.

task/dump-agent/src/main.rs
bcantrill: Adding a comment to explain this -- and at the same time fixing the agent nomenclature which has made this more unclear than it should be.

app/demo-stm32g0-nucleo/Cargo.toml
bcantrill: Yeah, it *does* violate the rule of thumb, but it's also unfortunate to require it ever TOML simply because we are so extraordinarily flash constrained on the G0s.  How important is the rule of thumb?

mkeeter: It wouldn't be the worst abuse of features in Hubris; I'm fine letting it stand unless anyone else is especially offended.

task/jefe/src/dump.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    let base = match area {
+        Ok(Some(area)) => area.address,
+        _ => return,
+    };
+
+    let mut ndx = 0;
+
+    loop {
+        match kipc::get_task_dump_region(task, ndx) {
```
mkeeter, line 119: This is accidentally quadratic: each call to `kipc::get_task_dump_region(task, ndx)` iterates over every region from 0 to `ndx`.  One option would be to order task regions in the build system so that dumpable regions are always first; then we could use normal array indexing instead of the current iterator.

(It may not matter in practice, but I'm [allergic to secretly quadratic behavior](https://www.mattkeeter.com/blog/2021-03-01-happen/); if you leave it as-is, a comment would be appreciated)

task/jefe/src/dump.rs
```diff
@@ -0,0 +1,168 @@
...
+        &crate::generated::DUMP_AREAS,
+        Some(0x1000),
+    );
+
+    ringbuf_entry!(Trace::Initialized);
+
+    areas
+}
+
+pub fn get_dump_area(
```
mkeeter, line 44: `get_dump_area` and `claim_dump_area` would be cleaner if they accepted `base: u32` instead of `Option<u32>`, moving the `NoDumpAreas` error back into `jefe`'s `ServerImpl`.

task/jefe/src/main.rs
```diff
@@ -145,13 +152,15 @@ struct ServerImpl<'s> {
     task_states: &'s mut [TaskStatus; NUM_TASKS],
     deadline: u64,
     reset_reason: ResetReason,
+    #[cfg(feature = "dump")]
+    dump_areas: Option<u32>,
```
mkeeter, line 156: If `jefe` initializes its own dump regions (per discussion [here](https://github.com/oxidecomputer/hubris/pull/1219/files#r1142706351)), could this just be a `u32`?

task/dump-agent/src/main.rs
```diff
@@ -20,83 +20,48 @@ use userlib::*;
...
-                    nsegments: 0,
-                    written: size_of::<DumpAreaHeader>() as u32,
-                    length: area.length,
-                    agent_version: DUMP_AGENT_VERSION,
-                    dumper_version: DUMPER_NONE,
-                    next,
-                }
-            }
+    fn initialize(&self) -> Result<(), DumpAgentError> {
+        let jefe = Jefe::from(JEFE.get_task_id());
```
mkeeter, line 32: Nit: it would be slightly more efficient to get one `Jefe` then store it in the `ServerImpl`

task/dump-agent/src/main.rs
```diff
@@ -20,83 +20,48 @@ use userlib::*;
...
-            let saddr = area.address as usize + offset;
-            let segment = saddr as *mut DumpSegmentHeader;
-
-            (*segment).address = addr;
-            (*segment).length = length;
-
-            (*header).nsegments = nsegments + 1;
-            (*header).written = need;
+        if area.agent != humpty::DumpAgent::Task {
+            self.claim_dump_area()?;
```
mkeeter, line 54: Looking at the `humpty` code, it seems like claiming a dump area only modifies the `agent` field, but doesn't clear anything else.  If dump area 0 was previously used as a `Jefe` dump site, then it will have non-zero `nsegments` / `written`, which seems bad.

app/demo-stm32g0-nucleo/Cargo.toml
bcantrill: Agreed; flipped this around and then added `dump` as a default feature.

build/xtask/src/dist.rs
```diff
@@ -1261,6 +1267,45 @@ fn update_image_header(
...
+    if let Some(task) = toml.tasks.get("dump_agent") {
+        if task.task_slots.get("jefe").is_some() {
+            //
+            // We have a dump agent, and it has a slot for Jefe, denoting that
+            // it is configured for task dumps; now we want to check that Jefe
+            // (1) has the dump feature enabled and (2) uses everything that
+            // the dump agent is using.
+            //
+            let jefe = toml.tasks.get("jefe").context("missing jefe?")?;
+
```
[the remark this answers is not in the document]
bcantrill: Agreed; done.

sys/abi/src/lib.rs
bcantrill: Done!

task/dump-agent/src/main.rs
bcantrill: Done

task/jefe/src/dump.rs
bcantrill: Agreed; done

task/jefe/src/main.rs
bcantrill: Done

sys/userlib/src/kipc.rs
```diff
@@ -24,6 +24,46 @@ pub fn read_task_status(task: usize) -> abi::TaskState {
...
+        Kipcnum::GetTaskDumpRegion as u16,
+        &buf,
+        &mut response,
+        &[],
+    );
+    assert_eq!(rc, 0);
+    ssmarshal::deserialize(&response[..len]).unwrap_lite().0
+}
+
+pub fn read_task_dump_region(
```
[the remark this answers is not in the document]
bcantrill: Preventing this by not allowing the supervisor to read its own regions (which makes no sense anyway).

task/dump-agent/src/main.rs
bcantrill: Done.

Cargo.lock
bcantrill: Done!

task/jefe/src/dump.rs
bcantrill: The absolute number here are very small, so this amount to an O(~7) scan O(~twice). Added a comment to this effect.

idl/jefe.idol
bcantrill: It is indeed not idempotent -- thank you!

task/dump-agent/src/main.rs
bcantrill: Added a comment to clarify what's going on here.

task/dump-agent/src/main.rs
bcantrill: This wasn't possible, but has been obviated anyway by moving the initialization code into Jefe.

build/xtask/src/dist.rs
```diff
@@ -1213,6 +1217,9 @@ fn update_image_header(
                 // Alias for the NS peripherals
                 sau_ranges.insert(0x4000_0000..=0x4fff_ffff, false);
 
+                // Alias for NS SRAMs
+                sau_ranges.insert(0x2002_0000..=0x2004_ffff, false);
```
mkeeter, line 1221: Is this change intentional?

build/xtask/src/dist.rs
```diff
@@ -1306,6 +1313,74 @@ fn update_image_header(
...
+        for u in &task.extern_regions {
+            if jefe.extern_regions.iter().find(|&j| j == u).is_none() {
+                bail!(
+                    "dump agent/jefe misconfiguration: dump agent has \
+                    {u} as an extern-region and depends on jefe, but jefe \
+                    does not have {u} as an extern-region"
+                );
+            }
+        }
+    } else {
```
mkeeter, line 1375: Nit: Clippy will probably tell you to flatten this conditional.

build/xtask/src/dist.rs
bcantrill: Yes, otherwise you can't use `sram2` for a dump area. (It's being used in the LPCXpresso image.)

mkeeter: 👍🏻

chips/lpc55/memory.toml
```diff
@@ -51,28 +51,59 @@ read = true
 write = true
 execute = false
 
+[[sram2]]
```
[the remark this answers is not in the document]
bcantrill: Agreed with the opportunity for improvement (this is definitely not introduced by this work though).

sys/kern/src/kipc.rs
```diff
@@ -262,3 +272,138 @@ fn read_caboose_pos(
...
+        )));
+    }
+
+    let rval = if rindex == 0 {
+        let size: u32 = core::mem::size_of::<Task>() as u32;
+
+        Some(abi::TaskDumpRegion {
+            base: tasks.as_ptr() as u32 + index * size,
+            size: size,
+        })
```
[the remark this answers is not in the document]
bcantrill: I like it more too! Thank you.

build/xtask/src/dist.rs
bcantrill: It does! It also tells me a *lot* of other things on this file -- may be worth making `build` Clippy-clean at some point... ;)

Anyway, will fix this one -- thank you!

mkeeter (approved):

PR #1224, opened 2023-03-21
re-enable testing with humility

steveklabnik:


steveklabnik:
Woo! Guess the temporary issue has passed.

@bcantrill indicated support for turning this back on, and this PR is trivial, so I'm going to merge it. Standard "if this messes something up I'm happy to do the reverting work" applies.

PR #1227, opened 2023-03-21
stm32h7-update-server: array handling cleanups

cbiffle:
So, this feels cleaner to me because it removes a bunch of runtime error cases in favor of static checks -- but that might just be me, given my bias against runtime errors. Feedback wanted.


drv/stm32h7-update-server/src/main.rs
```diff
@@ -163,36 +169,29 @@ impl<'a> ServerImpl<'a> {
...
-        }
+        let addresses = (start..start + FLASH_WORD_BYTES as u32).step_by(4);
 
         self.flash.bank2().cr.write(|w| {
             // SAFETY
             // The `psize().bits(_)` function is marked unsafe in the stm32
             // crate because it allows arbitrary bit patterns. `0b11`
-            // corresponds to 64-bit parallelism.
+            // corresponds to 64-bit internal parallelism during the write cycle
+            // (not to be confused with 64-bit accesses below).
```
mkeeter, line 183: Nit:
```suggestion
            // (not to be confused with accesses below, which are 32-bit).
```

drv/stm32h7-update-server/src/main.rs
```diff
@@ -333,24 +332,33 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
-        if len < BLOCK_SIZE_BYTES {
-            flash_page[len..].fill(0);
+        // While our input arrives as unstructured borrowed bytes, we want to
+        // ensure that we've got it aligned to 32-bits for internal reasons, and
+        // we make the internal structure of the flash "page" apparent: from the
+        // hardware's perspective, it is actually an array of flash words,
+        // grouped (by our arbitrary choice) into units of
+        // FLASH_WORDS_PER_BLOCK.
+        let mut flash_page: [[u32; FLASH_WORD_WORDS]; FLASH_WORDS_PER_BLOCK] =
+            [[0; FLASH_WORD_WORDS]; FLASH_WORDS_PER_BLOCK];
```
jgallagher, line 342: This puts 1 KiB on the stack, I think - do we need to bump the stacksize in app.toml?

jgallagher: Oh nevermind, we already had that before!

mkeeter (approved):

drv/stm32h7-update-server/src/main.rs
cbiffle: Whew, I was worried when Github only showed me your first comment for some reason. :-) Yeah, my intent here was to leave the actual size of the buffers unchanged but to alter their internal structure. If I got that _wrong_ then I would like to fix it.

drv/stm32h7-update-server/src/main.rs
cbiffle: Fixed

PR #1229, opened 2023-03-22
Treat `bank2` as an `extern` instead of a peripheral

mkeeter:
Regions specified in the `uses` array are meant to be peripherals, and have extra requirements in the MMU configuration (specifically, 32-bit aligned writes).  Our update server configured `bank2` as a peripheral; it happened to work because we were always doing 32-bit aligned writes, but isn't the right design.

This PR adds a new `extern-regions` configuration block to tasks.  If a task specifies an `extern-region`, it is given (exclusive) access to the entire region in its MMU configuration, and the build system injects symbols to indicate the region's start and end in the linker script.  These symbols can be used in an `extern` block.

We then use the new `extern-regions` mechanism to access `bank2` of flash in `stm32h7-update-server`, instead of defining it as a peripheral and using hard-coded addresses.


drv/stm32h7-update-server/src/main.rs
```diff
@@ -163,13 +163,23 @@ impl<'a> ServerImpl<'a> {
...
         ringbuf_entry!(Trace::WriteStart);
 
-        if word_number > BANK_WORD_LIMIT {
+        // These variables are _philosophically_ constants, but since they're
+        // generated by taking the address of a linker-generated symbol, we
+        // can't define them as `const` values.
+        //
+        // SAFETY: these are symbols populated by the linker.
+        let bank_addr = unsafe { __REGION_BANK2_BASE.as_ptr() } as u32;
+        let bank_end = unsafe { __REGION_BANK2_END.as_ptr() } as u32;
```
jgallagher, line 172: Tiny nit - would `as usize` here prevent the `as u32` casts below it?

mkeeter: Good idea, done in 82544d4c20c1f3bb9e3b9f521b75378d6f8144cd

jgallagher (approved):

PR #1230, opened 2023-03-22
Introduce `packrat` task for caching VPD

### LINKED ISSUE #1196: SP still comes up with self-assigned MAC address after #1171

### LINKED ISSUE #1197: It's possible for net and control-plane-agent to disagree on our IP, MAC, etc.
`control-plane-agent` gets our address from `net`. If `net` restarts, it will attempt to synchronously contact the VPD EEPROM to retrieve the MAC address. If for whatever reason that is not possible -- because I2C -- `net` will self-assign an address rather than becoming unresponsive. This is probably good.

However, `control-plane-agent` won't discover that the address has changed. This is probably bad.

I would argue that the problem is that we're synchronously accessing the VPD EEPROM well after boot and revisiting decisions that we thought we'd already made (like what our MAC address is). One way we could address this would be:

1. Access the VPD EEPROM one (1) time early in boot, probably from the sequencer task.
2. Store its results in a cache. This cache might be served by a dedicated task (probably the most general option), or there might be another appropriate place for it, I'm not sure off the top of my head.
3. Have other tasks that need access to the information ask the cache.
4. Treat the contents of the cache as immutable once loaded.

There's a subtle but (IMO) important detail in my proposal above: note that something like the sequencer, and _not_ the cache task, is responsible for accessing I2C. The cache task should probably not send any IPCs, period, because sending IPCs gives other tasks the right to fault you, and we don't want this task to ever fault. (Jefe follows this same principle.) So in my imagination here it's a load-once caching server that can answer questions once loaded. It might also distribute a notification when it gets loaded in case other tasks want to wait for VPD data to become valid.

This raises the question, of course, of what happens if the party responsible for loading the cache _can't access the VPD EEPROM._ This scenario is possible because I2C, and would likely require a power cycle to clear. I would think that `net` would probably want to wait on both notification of cache-load, _and some arbitrary timeout,_ so that it can self-assign and appear on the network in the event of total VPD failure.

Unless we want a machine with inaccessible VPD to act like a brick so that it gets power cycled -- this doesn't seem right to me, but is an option!

FYI @mkeeter @jgallagher @rmustacc

jgallagher:
I still need to test this on real hardware, but I think it's ready for review (and I want to run it through CI!). Some notes on changes:

1. Adds a `packrat` task that only receives messages from other tasks but makes no calls into tasks on its own, and should therefore never crash (assuming reasonable stack size etc, which I still need to confirm).
2. Moves the VPD reading of mac addresses out of `net` and into `gimlet-seq`, `sidecar-seq`, and `psc-seq`, the latter of which is new (and only does VPD reading, at least for now!).
3. Moves the VPD reading of the board identity/barcode out of `control-plane-agent` and into the same sequencer tasks.
4. Adds a sleep/retry loop for VPD reading in an attempt to handle #1196. It's possible this is not necessary and/or the wrong way to go about this; open to suggestions/feedback.

A couple things I don't love about the current implementation:

1. `read_vpd()` is copy/pasted between the three seq tasks. This was expedient because I wanted to ringbuf-log errors in the retry loops, but it seems long enough that have 3 copies is not great.
2. The aforementioned retries - do we even want this at all? If so, are the values I chose reasonable? (On a board without a programmed VPD we will be waiting at least that long for the netstack to come up on first boot.)

Some things that aren't in this implementation but could be done as more or less immediate followup PRs:

1. Updating udpbroadcast to pull the board identity from packrat and include it in its packets
2. Moving host startup options from c-p-a to packrat
3. packrat sending notifications when values are set - I'm not sure we need this today, but Cliff mentioned it in #1197 and it seems like a perfectly reasonable thing to expect.

Fixes #1197. Might fix #1196, but I'm not sure we understand the root cause there so I can't say either way with much confidence.


drv/gimlet-seq-server/src/main.rs
```diff
@@ -386,6 +396,63 @@ fn main() -> ! {
     }
 }
 
+fn read_vpd() {
```
mkeeter, line 399: In the PR, you mentioned disliking that this is copy-pasted into every sequencer, which I agree with!

How about moving it into the `task-packrat-api` crate, so everyone does
```rust
let i2c = I2C.get_task_id();
task_packrat_api::read_vpd(i2c, packrat);
```
instead?

task/control-plane-agent/src/main.rs
```diff
@@ -71,7 +71,7 @@ enum Log {
     },
 
     #[cfg(feature = "vpd-identity")]
-    VpdReadError(drv_local_vpd::LocalVpdError),
+    VpdReadError(drv_local_vpd::VpdIdentityError),
```
mkeeter, line 74: This feature no longer exists, so presumably this variant should be deleted?

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -47,14 +45,8 @@ impl MgsCommon {
...
     pub(crate) fn identity(&self) -> VpdIdentity {
-        if let Some(identity) = *self.identity.borrow() {
-            return identity;
-        }
-
-        let id = identity();
-        let mut cached = self.identity.borrow_mut();
-        *cached = Some(id);
-        id
+        let packrat = Packrat::from(PACKRAT.get_task_id());
```
mkeeter, line 48: What prevents us from talking to `packrat` before it's ready?  Are we waiting on `jefe` state somewhere in `control-plane-agent` startup?

task/net/src/bsp/psc_a.rs
```diff
@@ -48,6 +53,32 @@ impl bsp_support::Bsp for BspImpl {
...
+                    // This happens before we're in a valid power state.
+                    //
+                    // Only listen to our Jefe notification. Discard any error
+                    // since this can't fail but the compiler doesn't know that.
+                    let _ = sys_recv_closed(
+                        &mut [],
+                        notifications::JEFE_STATE_CHANGE_MASK,
+                        TaskId::KERNEL,
+                    );
+                }
```
mkeeter, line 77: Missing case statement here for `None`?

task/net/src/main.rs
```diff
@@ -114,8 +114,9 @@ fn mac_address_from_uid(sys: &Sys) -> MacAddressBlock {
 
 #[cfg(feature = "vpd-mac")]
 fn mac_address_from_vpd() -> Option<MacAddressBlock> {
-    let i2c_task = I2C.get_task_id();
-    drv_local_vpd::read_config(i2c_task, *b"MAC0").ok()
+    use task_packrat_api::Packrat;
+    let packrat = Packrat::from(PACKRAT.get_task_id());
```
mkeeter, line 118: Same question here – how does `net` know to wait until `packrat` has been loaded?

task/packrat/src/main.rs
```diff
@@ -0,0 +1,140 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! VPD manipulation
```
mkeeter, line 5: This isn't the best description of the task (and indeed appears copied from `task/vpd/src/main.rs`)

drv/gimlet-seq-server/src/main.rs
jgallagher: The problem with moving it is the `ringbuf_entry!` calls - those don't make sense from `task-packrat-api`. I could:

* remove them
* add callbacks to report errors
* remove the retry loop and just return an error, although this would complicate the caller enough that we might still have 3 moderately-sized functions copy-pasted

task/control-plane-agent/src/mgs_common.rs
jgallagher: This might be too roundabout for comfort, but: we only get here if we received a packet on the network, and `net` waits for `jefe` to populate packrat (by waiting for the transition to A2).

task/net/src/bsp/psc_a.rs
jgallagher: No, it's on line 67:

```rust
Some(PowerState::Init) | None => {
```

task/packrat/src/main.rs
jgallagher: It is indeed! Will fix.

task/net/src/main.rs
jgallagher: It's implicit in the `BspImpl::preinit()` call where we wait for the sequencer - in the changes to both gimlet-seq and sidecar-seq, we should be loading packrat before transitioning to the state that releases net's `preinit()`.

The PSC didn't have a `preinit()` before this PR, but now it does and it works the same way: we wait until the psc-seq transitions to A2, which it does immediately after loading `packrat`.

jgallagher:
Running this branch on the niles psc-b (which does not have programmed VPD) works as expected; we come up with a self-assigned IP address, and have errors in the ringbuf logs from the sequencer trying to read the VPD:

```
john@niles ~ $ pfexec humility -t psc -a ./build-psc-b.zip ringbuf sequencer
humility: attached to 0483:374e:001C00175553500F20393256 via ST-Link V3
humility: ring buffer drv_local_vpd::__RINGBUF in sequencer:
   0  149        1       10 Error(NoRootChunk)
humility: ring buffer drv_psc_seq_server::__RINGBUF in sequencer:
   0   72        1        1 LocalVpdError(NoRootChunk)
   1   89        1        1 LocalVpdError(NoRootChunk)
   2   72        1        1 LocalVpdError(NoRootChunk)
   3   89        1        1 LocalVpdError(NoRootChunk)
   4   72        1        1 LocalVpdError(NoRootChunk)
   5   89        1        1 LocalVpdError(NoRootChunk)
   6   72        1        1 LocalVpdError(NoRootChunk)
   7   89        1        1 LocalVpdError(NoRootChunk)
   8   72        1        1 LocalVpdError(NoRootChunk)
   9   89        1        1 LocalVpdError(NoRootChunk)
```

jgallagher:
Running on niles's sidecar-b (which does have programmed VPD) looks good; macs and identity are loaded into packrat, and both `net` and `control-plane-agent` see them:

```
john@niles ~ $ pfexec humility -t sidecar-b -a ./build-sidecar-b-lab.zip ringbuf packrat
   0   84        1        1 MacAddressBlockSet(Set(MacAddressBlock { base_mac: [ 0xa8, 0x40, 0x25, 0x5, 0xff, 0x0 ], count: U16<byteorder::LittleEndian>([ 0x0, 0x1 ], PhantomData<byteorder::LittleEndian>), stride: 0x1 }))
   1   84        1        1 VpdIdentitySet(Set(VpdIdentity { part_number: [ 0x39, 0x31, 0x33, 0x2d, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30, 0x36 ], revision: 0x4, serial: [ 0x42, 0x52, 0x4d, 0x34, 0x34, 0x32, 0x32, 0x30, 0x30, 0x30, 0x31 ] }))
john@niles ~ $ pfexec humility -t sidecar-b -a ./build-sidecar-b-lab.zip net ip
MAC address:  a8:40:25:05:ff:00
IPv6 address: fe80::aa40:25ff:fe05:ff00
john@niles ~ $ ./faux-mgs --interface axf4 --discovery-addr '[fe80::aa40:25ff:fe05:ff00]:11111' state 2>/dev/null
hubris archive: 2ff4bd04e42ccd37
serial number: BRM44220001
model: 913-0000006
revision: 4
base MAC address: a8:40:25:05:ff:00
hubris version: ImageVersion { epoch: 0, version: 0 }
power state: A0
RoT state: Err(MessageError { code: 19 })
```

drv/gimlet-seq-server/src/main.rs
mkeeter: Ah, I see.

You _could_ put a ringbuf into `task-packrat-api` if you wanted; having ringbufs in library crates isn't totally standard, but we do it in a few drivers already.

mkeeter: (although then every task that calls _into_ packrat would get the ringbuf allocated; maybe an extra layer of indirection with a `drv-packrat-writer` crate that includes just this function and ringbuf?)

jgallagher:
Confirmed working on a gimlet-c with programmed VPD as well:

```
john@lurch ~ $ pfexec humility -t sn66 -a ./build-gimlet-c-lab.zip ringbuf packrat
humility: attached to 0483:3754:003F00164741500920383733 via ST-Link V3
   0  106        1        1 MacAddressBlockSet(Set(MacAddressBlock { base_mac: [ 0xa8, 0x40, 0x25, 0x4, 0x2, 0x81 ], count: U16<byteorder::LittleEndian>([ 0x8, 0x0 ], PhantomData<byteorder::LittleEndian>), stride: 0x8 }))
   1  106        1        1 VpdIdentitySet(Set(VpdIdentity { part_number: [ 0x39, 0x31, 0x33, 0x2d, 0x30, 0x30, 0x30, 0x30, 0x30, 0x31, 0x39 ], revision: 0x6, serial: [ 0x42, 0x52, 0x4d, 0x34, 0x32, 0x32, 0x32, 0x30, 0x30, 0x36, 0x36 ] }))
john@lurch ~ $ ./faux-mgs --interface axf1 --discovery-addr '[fe80::aa40:25ff:fe04:0281]:11111' state 2>/dev/null
hubris archive: 01b26ef98a5c4e36
serial number: BRM42220066
model: 913-0000019
revision: 6
base MAC address: a8:40:25:04:02:81
hubris version: ImageVersion { epoch: 0, version: 0 }
power state: A2
RoT state: Err(MessageError { code: 19 })
```

jgallagher:
@mkeeter this is ready for another look whenever convenient; I think all your initial comments are addresses (maybe modulo the "how do packrat clients know packrat has data for them" issue).

app/gimlet/rev-b.toml
```diff
@@ -100,6 +100,14 @@ notifications = ["i2c1-irq", "jefe-state-change"]
 "i2c1.event" = "i2c1-irq"
 "i2c1.error" = "i2c1-irq"
 
+[tasks.packrat]
+name = "task-packrat"
+priority = 3
+max-sizes = {flash = 8192, ram = 2048}
+start = true
+# task-slots is explicitly empty: packrat should send not send IPCs!
```
mkeeter, line 108:
```suggestion
# task-slots is explicitly empty: packrat should not send IPCs!
```
(also needs fixing in other `app.toml` files)

drv/local-vpd/src/lib.rs
```diff
@@ -163,4 +184,21 @@ pub fn read_config_into(
     Err(err(LocalVpdError::NoRootChunk))
 }
 
+/// Read the Oxide barcode tag and parse it.
+///
+/// Supports `0XV1` and `0XV2` formats.
+pub fn read_oxide_barcode(
```
mkeeter, line 190: Now that we've got `drv-packrat-vpd-loader`, I wonder if this function should be moved into there.

Moving the function would remove the dependency on `oxide_barcode` and keep `drv-local-vpd` as a thin raw-byte-reader.

drv/psc-seq-api/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! API crate for the Gimlet Sequencer server.
```
mkeeter, line 5:
```suggestion
//! API crate for the PSC Sequencer server.
```

task/packrat-api/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+
+//! Client API for the VPD task.
+
+#![no_std]
+
+use derive_idol_err::IdolError;
+use userlib::*;
+use zerocopy::{AsBytes, FromBytes, LittleEndian, U16};
+
+pub use oxide_barcode::VpdIdentity;
```
mkeeter, line 13: Is re-exporting this necessary?

jgallagher: Not strictly necessary, but I think it should be - `VpdIdentity` is part of this crate's public API (although it's hidden here because it's part of the generated idol client), and I don't like when crates require users to depend on _another_ crate for types they use.

mkeeter: Seems reasonable!

app/gimlet/rev-b.toml
jgallagher: thanks, fixed in 72f64225

drv/psc-seq-api/src/lib.rs
jgallagher: Fixed this and a similar copy/paste error in 17a2d8c5

task/net/src/main.rs
jgallagher: :+1: Added in 52ac248c

task/control-plane-agent/src/mgs_common.rs
jgallagher: :+1: Added in 52ac248c

mkeeter (approved):

PR #1221, opened 2023-03-20
Add support for rev C Sidecar

mkeeter:
(still in progress)

- [ ] Add new bitstream file
- [ ] Add to CI

Closes https://github.com/oxidecomputer/hubris/issues/1211

PR #1231, opened 2023-03-23
Deprecate Sidecar rev A, add support for rev C

arjenroodselaar:
This diff deprecates support for Sidecar rev A (🫡) but adds support for Sidecar rev C. As part of rev C support there are some slight tweaks of the `ignition-server`.


arjenroodselaar:
@Aaron-Hartwig, @bnaecker, there is some conditional code for `sidecar-a` in the LED and transceiver code which I think can be removed as part of this. Mind taking a quick peek to see if I did that correct?

mkeeter:
I started this in https://github.com/oxidecomputer/hubris/pull/1221, but this looks more complete, so I'll close that PR and review this one instead!

drv/ignition-server/src/main.rs
```diff
@@ -64,6 +64,25 @@ fn main() -> ! {
...
+    //
+    // Methods provided by this server use the count as an upper bound when
+    // iterating over the ports, causing port 35 to be ignored on rev B
+    // systems.
+    #[cfg(target_board = "sidecar-b")]
+    {
+        if server.port_count == 36 {
+            server.port_count = 35;
+        }
+    }
```
mkeeter, line 84:
```suggestion
    if cfg!(target_board = "sidecar-b") && server.port_count == 36 {
        server.port_count = 35;
    }
```

Also, why does `port_count` return 36 on the rev B bitstream, rather than 35?

task/thermal/src/main.rs
```diff
@@ -15,8 +15,8 @@
 #[cfg_attr(target_board = "gimlet-b", path = "bsp/gimlet_bc.rs")]
 #[cfg_attr(target_board = "gimlet-c", path = "bsp/gimlet_bc.rs")]
 #[cfg_attr(
-    any(target_board = "sidecar-a", target_board = "sidecar-b"),
-    path = "bsp/sidecar_ab.rs"
+    any(target_board = "sidecar-b", target_board = "sidecar-c"),
```
mkeeter, line 18: Nit: we should make the Sidecar and Gimlet `cfg_attr` consistent here; either
```rust
#[cfg_attr(
    any(target_board = "gimlet-b", target_board = "gimlet-c"),
    path = "bsp/gimlet_bc.rs"
)]
#[cfg_attr(
    any(target_board = "sidecar-b", target_board = "sidecar-c"),
    path = "bsp/sidecar_bc.rs"
)]
```
or
```rust
#[cfg_attr(target_board = "gimlet-b", path = "bsp/gimlet_bc.rs")]
#[cfg_attr(target_board = "gimlet-c", path = "bsp/gimlet_bc.rs")]
#[cfg_attr(target_board = "sidecar-b", path = "bsp/sidecar_bc.rs")]
#[cfg_attr(target_board = "sidecar-c", path = "bsp/sidecar_bc.rs")]
```

I like the upper option more, but don't feel too strongly about it.

mkeeter:
@arjenroodselaar This generally looks good!

You'll also want to bring in this [`main.rs`](https://github.com/oxidecomputer/hubris/blob/dc2cb82c3c8e075ada5db09a873497811e77b7fd/app/sidecar/src/main.rs), which adds support for checking resistors on startup.

drv/ignition-server/src/main.rs
arjenroodselaar: Thanks, this is much cleaner.

I also added to the comment to try and explain better why this is done here. The tl;dr is that this uses a set of templates to generate the RTL which involve some shenanigans to make work if there are 36 sets of device pins but only 35 channels. Performing a small three line workaround here seems like a reasonable trade-off.

arjenroodselaar:
> @arjenroodselaar This generally looks good!
> 
> You'll also want to bring in this [`main.rs`](https://github.com/oxidecomputer/hubris/blob/dc2cb82c3c8e075ada5db09a873497811e77b7fd/app/sidecar/src/main.rs), which adds support for checking resistors on startup.

I pulled this in and removed the if arm for `sidecar-a`.

drv/ignition-server/src/main.rs
arjenroodselaar: I moved this into `drv/sidecar-mainboard-controller` since its cleaning up a hardware detail. This leaves the `ignition-server` task with less implementation details.

mkeeter (approved):

PR #1232, opened 2023-03-23
Read caboose from bank2 and refuse to update mismatched boards

mkeeter:
(only implemented on the SP side for the moment)


Cargo.toml
```diff
@@ -117,7 +117,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 
 # Oxide forks and repos
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"], branch = "image-board-check" }
```
jgallagher, line 120: (I think this is merged and can point back to `main` now)

drv/stm32h7-update-server/src/main.rs
```diff
@@ -421,6 +422,100 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+        let image_start = unsafe { __REGION_BANK2_BASE.as_ptr() } as u32;
+
+        // If all is going according to plan, there will be a valid Hubris image
+        // flashed into the other slot, which is delimited by
+        // `__REGION_BANK2_BASE` and `__REGION_BASE2_END` (which are symbols
+        // injected by the linker).
+        //
+        // We'll first want to read the image header, which is at a fixed
+        // location at the end of the vector table.
+        const HEADER_OFFSET: u32 = 0x298;
```
jgallagher, line 443: Is this offset determined by our build system or chip hardware or something else?

task/control-plane-agent/src/update/sp.rs
```diff
@@ -523,11 +524,36 @@ impl AcceptingData {
...
-                Err(err) => {
-                    (State::Failed(err), Err(SpError::UpdateFailed(err as u32)))
+            // Confirm that the image written is targeting the same board as our
+            // current image.  If the current image doesn't have a caboose or a
+            // `BORD` key, then we'll accept anything, but the incoming image
+            // **must** have a valid caboose and `BORD`.
+            const BOARD_KEY: [u8; 4] = *b"BORD";
+            let ours = userlib::kipc::get_caboose()
+                .map(CabooseReader::new)
+                .and_then(|reader| reader.get(BOARD_KEY).ok());
```
jgallagher, line 534: Two questions:

1. Can this fail in a way that makes us think we don't have a `BORD`, but actually we do and reading it at runtime failed? (I.e., should we be checking the kind of error we get here and only proceeding on specific ones?)
2. Would it make sense to read the caboose and stuff it into packrat? (Not suggesting this happen as part of this PR, but just a general question.)

Cargo.toml
mkeeter: Yes, done

drv/stm32h7-update-server/src/main.rs
mkeeter: A little from column A, a little from column B: our build system places the `ImageHeader` directly after the vector table, which is determined by the chip hardware.  I added a little more detail to this comment.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -421,6 +422,100 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+        // See the detailed comment above `can_access` in `sys/kern/src/task.rs`
+        // for details!
+        const BUF_SIZE: usize = 16;
+        let mut buf = [0u8; BUF_SIZE];
+        let mut pos = 0;
+        for c in chunk.chunks(BUF_SIZE) {
+            let buf = &mut buf[..c.len()];
+            buf.copy_from_slice(c);
+            data.write_range(pos..pos + c.len(), buf)
+                .map_err(|_| RequestError::Fail(ClientError::BadLease))?;
```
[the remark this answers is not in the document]
mkeeter: Yes, good catch!

jgallagher (approved):

PR #1233, opened 2023-03-24
Update reset with intent

lzrd:
This adds `ResetComponent` messages to MGS, SpRot, and Update APIs.
The function currently plumbed through is a simple reset of the RoT.
The API also provides a semantic for modifying RoT boot image selection and delivering a blob of authentication data.
The authentication blob could be used for online key revocation, debug enable, or other operations that involve RoT stage0 or hypovisor and are associated with resetting the RoT.
These forward compatible features of the messages are incorporated now because some will be needed very soon.

### DIFF
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -189,8 +189,6 @@ impl Handler {
                             .reset_component(intent, target)
                             .map(|_| None)
                             .map_err(|e| e.into());
-                        // TODO: Some sort of error if we didn't reset.
-
                         tx_buf.serialize(MsgType::UpdResetComponentRsp, rsp)
                     }
                     Err(e) => Err((tx_buf, e.into())),
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -181,17 +181,12 @@ impl Handler {
                 match hubpack::deserialize::<drv_sprot_api::ResetComponentHeader>(
                     rx_payload,
                 ) {
-                    Ok((header, auth_data)) => {
+                    Ok((header, _trailing_data)) => {
                         let intent = header.intent;
                         let target = header.target;
                         let rsp: UpdateRspHeader = self
                             .update
-                            .reset_component(
-                                intent,
-                                target,
-                                auth_data.len() as u16,
-                                &auth_data[..],
-                            )
+                            .reset_component(intent, target)
                             .map(|_| None)
                             .map_err(|e| e.into());
                         // TODO: Some sort of error if we didn't reset.
@@ -231,7 +226,7 @@ impl Handler {
             }
 
             // All of the unexpected messages
-            MsgType::Invalid
+            MsgType::_Invalid
             | MsgType::EchoRsp
             | MsgType::ErrorRsp
             | MsgType::SinkRsp
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -113,7 +113,7 @@ impl Handler {
                 UpdateStatus::Sp => Err((tx_buf, SprotError::UpdateBadStatus)),
             },
             MsgType::IoStatsReq => {
-                tx_buf.serialize(MsgType::IoStatsRsp, *stats)
+                tx_buf.serialize(MsgType::IoStatsRsp, stats.clone())
             }
             MsgType::SprocketsReq => {
                 let tx_payload = tx_buf.payload_mut();
@@ -237,8 +237,7 @@ impl Handler {
             | MsgType::UpdFinishImageUpdateRsp
             | MsgType::UpdResetComponentRsp
             | MsgType::IoStatsRsp
-            | MsgType::DumpRsp
-            | MsgType::Unknown => {
+            | MsgType::DumpRsp => {
                 stats.rx_invalid = stats.rx_invalid.wrapping_add(1);
                 return Some(tx_buf.error_rsp(SprotError::BadMessageType));
             }
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -177,6 +177,30 @@ impl Handler {
                     .map_err(|e| e.into());
                 tx_buf.serialize(MsgType::UpdFinishImageUpdateRsp, rsp)
             }
+            MsgType::UpdResetComponentReq => {
+                match hubpack::deserialize::<drv_sprot_api::ResetComponentHeader>(
+                    rx_payload,
+                ) {
+                    Ok((header, auth_data)) => {
+                        let intent = header.intent;
+                        let target = header.target;
+                        let rsp: UpdateRspHeader = self
+                            .update
+                            .reset_component(
+                                intent,
+                                target,
+                                auth_data.len() as u16,
+                                &auth_data[..],
+                            )
+                            .map(|_| None)
+                            .map_err(|e| e.into());
+                        // TODO: Some sort of error if we didn't reset.
+
+                        tx_buf.serialize(MsgType::UpdResetComponentRsp, rsp)
+                    }
+                    Err(e) => Err((tx_buf, e.into())),
+                }
+            }
             MsgType::SinkReq => {
                 // The first two bytes of a SinkReq payload are the U16
                 // mod 2^16 sequence number.
@@ -218,6 +242,7 @@ impl Handler {
             | MsgType::UpdWriteOneBlockRsp
             | MsgType::UpdAbortUpdateRsp
             | MsgType::UpdFinishImageUpdateRsp
+            | MsgType::UpdResetComponentRsp
             | MsgType::IoStatsRsp
             | MsgType::DumpRsp
             | MsgType::Unknown => {
--- a/drv/update-api/Cargo.toml
+++ b/drv/update-api/Cargo.toml
@@ -10,6 +10,8 @@ num-traits.workspace = true
 serde.workspace = true
 serde_repr.workspace = true
 zerocopy.workspace = true
+ringbuf = { path = "../../lib/ringbuf" }
+gateway-messages.workspace = true
 
 derive-idol-err.path = "../../lib/derive-idol-err"
 drv-caboose.path = "../../drv/caboose"
--- a/drv/update-api/src/lib.rs
+++ b/drv/update-api/src/lib.rs
@@ -38,8 +38,6 @@ pub enum UpdateTarget {
     ImageA = 2,
     ImageB = 3,
     Bootloader = 4,
-    // For testing
-    DevNull = 0xff,
 }
 
 #[derive(
@@ -100,6 +98,10 @@ impl hubpack::SerializedSize for UpdateError {
 }
 
 /// Request component to reset and optionally modify boot image selection policy.
+/// Policies may be implemented that disallow some requests. For example,
+/// if it is known that stage0 will reject an image for being corrupted, unsigned,
+/// signed improperly, etc., a request to reset and change boot policy may be
+/// rejected.
 #[repr(u8)]
 #[derive(
     Eq,
@@ -114,13 +116,18 @@ impl hubpack::SerializedSize for UpdateError {
 )]
 pub enum ResetIntent {
     /// Just reset the component
-    Normal = 1,
-    /// The firmware image specified elsewhere in the message
-    /// should be set as the persistently preferred image.
+    Normal,
+    /// The firmware image specified in the reset_component message
+    /// will be set as the persistently preferred image.
     Persistent,
-    /// The firmware image specified elsewhere in the message
-    /// should be set as the preferred image for this reset only.
+    /// The firmware image specified in the reset_component message
+    /// will be set as the preferred image for this reset only.
     /// A transient preference overrides any persistent preference.
+    ///
+    /// Note that implementing a transient image selection for the
+    /// SP would be fairly involved but possible.
+    /// The RoT might have to effect the SP bank switch if a way cannot
+    /// be found to have the SP do it.
     Transient,
 }
 
--- a/drv/update-api/src/lib.rs
+++ b/drv/update-api/src/lib.rs
@@ -95,8 +95,6 @@ pub enum UpdateError {
     ImageBoardUnknown = 24,
 
     NotImplemented = 25,
-
-    Unknown = 0xff, // XXX With Hubpack encoding, inserting a new code above this will change the deserializer's notion of Unknown.
 }
 
 impl hubpack::SerializedSize for UpdateError {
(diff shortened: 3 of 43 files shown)


drv/update-api/src/lib.rs
```diff
@@ -81,20 +87,61 @@ pub enum UpdateError {
...
+            }
+        }
+    }
+}
+
+#[derive(Clone, Copy, SerializedSize, Serialize, Deserialize)]
+#[repr(C)]
+pub struct ResetComponentHeader {
+    pub intent: ResetIntent,
+    pub target: UpdateTarget,
```
cbiffle, line 142: The `UpdateTarget` enumeration contains a lot of options. Currently these are:

```rust
    // Represents targets where we only ever write to a single
    // alternate flash location. This is typically used in
    // conjunction with a bank swap feature.
    Alternate = 1,
    // Represents targets where we must write to a specific range
    // of flash.
    ImageA = 2,
    ImageB = 3,
    Bootloader = 4,
    // For testing
    DevNull = 0xff,
```

Is this the type you intended? If so, what does it mean to reset a bootloader, or devnull? What is the difference between sending ImageA and Alternate? If only one or the other is permitted, how do clients know which to choose when talking to an update service?

What does it mean to issue a reset of `Normal` with a particular image named in the `target` field?

idl/sprot.idol
```diff
@@ -129,5 +129,26 @@ Interface(
                 err: CLike("SprotError"),
             ),
         ),
+        "reset_component": (
```
cbiffle, line 132: (I recommend using `encoding: Hubpack` for all new IPC operations. This would let you remove a lot of the lines below.)

lib/stage0-handoff/src/rot_update_details.rs
```diff
@@ -19,7 +19,7 @@ unsafe impl HandoffData for RotBootState {
 ///
 /// It gets read from RAM by the `lpc55-update-server`
 #[derive(
-    Debug, Clone, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
+    Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
```
cbiffle, line 22: (this type is of nontrivial size -- 92 bytes, if my head-compiler is working correctly -- so be sure you actually want `Copy` here, as it may produce hidden copies with measurable performance impact.)

drv/update-api/Cargo.toml
```diff
@@ -10,6 +10,8 @@ num-traits.workspace = true
 serde.workspace = true
 serde_repr.workspace = true
 zerocopy.workspace = true
+ringbuf = { path = "../../lib/ringbuf" }
```
cbiffle, line 13: This appears unused, please remove.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -23,6 +23,7 @@ task_slot!(PACKRAT, packrat);
 /// Provider of MGS handler logic common to all targets (gimlet, sidecar, psc).
 pub(crate) struct MgsCommon {
     reset_requested: bool,
+    reset_component_requested: SpComponent,
```
cbiffle, line 26: What does it mean to set this when `reset_requested` is `false`?

Should `reset_requested` become an `Option<SpComponent>`?

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -32,6 +33,9 @@ impl MgsCommon {
     pub(crate) fn claim_static_resources(base_mac_address: MacAddress) -> Self {
         Self {
             reset_requested: false,
+            reset_component_requested: SpComponent {
+                id: [0; SpComponent::MAX_ID_LENGTH],
```
cbiffle, line 37: This looks like it's intended to be an invalid SpComponent, which suggests that Option is more appropriate.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -143,6 +147,108 @@ impl MgsCommon {
...
+        // TODO: Add some kind of auth check before performing a reset.
+        // https://github.com/oxidecomputer/hubris/issues/723
+        if self.reset_component_requested != component {
+            return Err(SpError::ResetComponentTriggerWithoutPrepare);
+        }
+        // If we are not resetting the SP_ITSELF, then we may come back here.
+        self.reset_component_requested.id.fill(0);
+
+        // For now, resetting the SP through reset_component() is
+        // the same as through reset()
```
cbiffle, line 179: "For now" -- are we planning a change in the future that will cause this to no longer be the case? What change is it, so I know when to make it as a maintenance programmer?

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -143,6 +147,108 @@ impl MgsCommon {
...
+        ringbuf_entry!(Log::MgsMessage(MgsMessage::ResetPrepare));
+        self.reset_component_requested = component;
+        Ok(())
+    }
+
+    pub(crate) fn reset_component_trigger(
+        &mut self,
+        update: &SpUpdate,
+        component: SpComponent,
+        slot: Option<u16>,
```
cbiffle, line 166: It looks like this API numbers slots in a component-specific manner starting from 0 using some subset of the `u16` space -- this appears to be the only place where this is done, could you discuss it in the rustdoc comment?

It would also be worth mentioning that the argument is completely ignored if the target is anything other than the RoT. (Arguments that are completely ignored if certain other arguments have certain other values should almost always be fields on an enum, fwiw.)

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -143,6 +147,108 @@ impl MgsCommon {
...
+                // Some other error occurred.
+                // TODO: Update is all-or-nothing at the moment.
+                // The control plane can try to reset the RoT again or it
+                // can start the update process all over again.  We should
+                // be able to make incremental progress if there is some
+                // bug/condition that is degrading SpRot communications.
+                ringbuf_entry!(Log::RotReset { err });
+                // TODO: send an error back. However, changes need to be
+                // made to the management plane for it to understand that
+                // the write was successful but the new image is not yet
```
cbiffle, line 236: When you say "the write was successful" here, what write are you referring to? It doesn't look like this operation performs any writes.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -143,6 +147,108 @@ impl MgsCommon {
...
+            }
+            Err(err) => {
+                // Some other error occurred.
+                // TODO: Update is all-or-nothing at the moment.
+                // The control plane can try to reset the RoT again or it
+                // can start the update process all over again.  We should
+                // be able to make incremental progress if there is some
+                // bug/condition that is degrading SpRot communications.
+                ringbuf_entry!(Log::RotReset { err });
+                // TODO: send an error back. However, changes need to be
```
cbiffle, line 234: This TODO seems pretty darn significant.... With the current code, how can we tell if an error occurred here?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -664,6 +665,37 @@ impl SpHandler for MgsHandler {
     ) -> Result<&'static [u8], SpError> {
         self.common.get_caboose_value(key)
     }
+
+    fn reset_component_prepare(
```
cbiffle, line 669: This appears to be the third in a series of three copy-pasted routines. It would be really nice not to copy-paste like this; what's the purpose of it?

task/hiffy/src/common.rs
```diff
@@ -729,6 +731,56 @@ pub(crate) fn sprot_block_size(
...
+    stack: &[Option<u32>],
+    data: &[u8],
+    _rval: &mut [u8],
+) -> Result<usize, Failure> {
+    if stack.len() < 3 {
+        return Err(Failure::Fault(Fault::MissingParameters));
+    }
+    let fp = stack.len() - 3;
+    let intent: drv_update_api::ResetIntent = match stack[fp + 0] {
+        // Some(intent) => match drv_update_api::ResetIntent::try_from(intent as u8) {
```
cbiffle, line 745: What's this line? It's different from the code just beneath it.

task/hiffy/src/common.rs
```diff
@@ -729,6 +731,56 @@ pub(crate) fn sprot_block_size(
     Ok(4)
 }
 
+#[cfg(feature = "sprot")]
+pub(crate) fn sprot_reset_component(
+    stack: &[Option<u32>],
+    data: &[u8],
+    _rval: &mut [u8],
+) -> Result<usize, Failure> {
```
cbiffle, line 739: Adding a `use drv_update_api::ResetIntent` at the top of this function's body would make things below a lot more concise.

task/hiffy/src/common.rs
```diff
@@ -729,6 +731,56 @@ pub(crate) fn sprot_block_size(
...
+    let fp = stack.len() - 3;
+    let intent: drv_update_api::ResetIntent = match stack[fp + 0] {
+        // Some(intent) => match drv_update_api::ResetIntent::try_from(intent as u8) {
+        Some(intent) => {
+            match drv_update_api::ResetIntent::from_u8(intent as u8) {
+                Some(intent) => intent,
+                None => return Err(Failure::Fault(Fault::BadParameter(0))),
+            }
+        }
+        None => return Err(Failure::Fault(Fault::EmptyParameter(0))),
```
cbiffle, line 752: FYI, all of these match/None/return sequences could be replaced by `.ok_or(error)?`.

task/hiffy/src/common.rs
```diff
@@ -729,6 +731,56 @@ pub(crate) fn sprot_block_size(
...
+    let target: drv_update_api::UpdateTarget = match stack[fp + 1] {
+        Some(target) => match drv_update_api::UpdateTarget::from_u32(target) {
+            Some(target) => target,
+            None => return Err(Failure::Fault(Fault::BadParameter(1))),
+        },
+        None => return Err(Failure::Fault(Fault::EmptyParameter(1))),
+    };
+    let auth_len: usize = match stack[fp + 2] {
+        Some(auth_len) => auth_len as usize,
+        None => return Err(Failure::Fault(Fault::EmptyParameter(0))),
```
cbiffle, line 763: Should this be `EmptyParameter(2)`?

task/hiffy/src/common.rs
```diff
@@ -729,6 +731,56 @@ pub(crate) fn sprot_block_size(
...
+    };
+
+    if auth_len > data.len().min(drv_update_api::AUTH_MAX_SZ) {
+        return Err(Failure::Fault(Fault::AccessOutOfBounds));
+    }
+
+    let mut auth_data = [0u8; drv_update_api::AUTH_MAX_SZ];
```
cbiffle, line 770: I don't understand what this code is doing. AFAICT you are
- Copying `data` to the prefix of an array
- Then passing `data`'s length to `reset_component`, causing it to ignore the rest of the array

Isn't this equivalent to, but more expensive than, just passing `data` to `reset_component` below?

task/hiffy/src/common.rs
```diff
@@ -1205,3 +1257,53 @@ pub(crate) fn block_size(
 
     Ok(4)
 }
+
+// intent, auth_len, auth_data(Lease)
```
cbiffle, line 1261: If this is intended as parameter documentation could you explain it as such?

task/hiffy/src/common.rs
```diff
@@ -1205,3 +1257,53 @@ pub(crate) fn block_size(
...
+#[cfg(feature = "update")]
+pub(crate) fn reset_component(
+    stack: &[Option<u32>],
+    data: &[u8],
+    _rval: &mut [u8],
+) -> Result<usize, Failure> {
+    if stack.len() < 3 {
```
cbiffle, line 1268: This appears to be an almost complete copy-paste of the previous function, down to a copy-paste of the typo on line 1290. Please extract a function describing the common bits so that it's clear that they're intended to be the same and to ensure they don't diverge.

task/hiffy/src/common.rs
```diff
@@ -1205,3 +1257,53 @@ pub(crate) fn block_size(
...
+    let auth_len: usize = match stack[fp + 2] {
+        Some(auth_len) => auth_len as usize,
+        None => return Err(Failure::Fault(Fault::EmptyParameter(0))),
+    };
+
+    if auth_len > data.len().min(drv_update_api::AUTH_MAX_SZ) {
+        return Err(Failure::Fault(Fault::AccessOutOfBounds));
+    }
+
+    let mut auth_data = [0u8; drv_update_api::AUTH_MAX_SZ];
```
cbiffle, line 1297: (This pattern looks as sketchy here as it did above, can probably be removed.)

cbiffle (commented):
Wanted to get a first pass sent back before the meeting at 1300 PDT.

Somewhere, there ought to be a comment explaining why there are reset operations on both update and sprot, how they are intended to be similar vs different, etc.

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -177,6 +177,34 @@ impl Handler {
...
+                        let auth_len = auth_data.len();
+                        let mut auth = [0u8; AUTH_MAX_SZ];
+                        auth[..auth_len]
+                            .copy_from_slice(&auth_data[..auth_len]);
+                        let rsp: UpdateRspHeader = self
+                            .update
+                            .reset_component(
+                                intent,
+                                target,
+                                auth_len as u16,
```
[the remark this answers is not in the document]
lzrd: Done

drv/update-api/src/lib.rs
```diff
@@ -81,20 +87,61 @@ pub enum UpdateError {
...
     // Caboose checks
     ImageBoardMismatch = 23,
     ImageBoardUnknown = 24,
 
-    #[idol(server_death)]
-    ServerRestarted = 22,
+    NotImplemented = 25,
 
-    Unknown = 0xff,
+    Unknown = 0xff, // XXX With Hubpack encoding, inserting a new code above this will change the deserializer's notion of Unknown.
```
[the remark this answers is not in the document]
lzrd: Removed Unknown for Update and SpRot.
This was a hack to satisfy u8 to enum translation.

drv/update-api/src/lib.rs
```diff
@@ -81,20 +87,61 @@ pub enum UpdateError {
...
 }
 
 impl hubpack::SerializedSize for UpdateError {
     const MAX_SIZE: usize = core::mem::size_of::<UpdateError>();
 }
 
+#[repr(u8)]
+#[derive(
+    Eq,
+    FromPrimitive,
```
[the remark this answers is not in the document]
lzrd: I removed the non-FromPrimitive implementation.

drv/update-api/src/lib.rs
```diff
@@ -81,20 +87,61 @@ pub enum UpdateError {
...
+    Copy,
+    Serialize,
+    Deserialize,
+    SerializedSize,
+)]
+pub enum ResetIntent {
+    Normal = 1,
+    Persistent,
+    Transient,
+    ExpensiveAndIrrevocableProdToDev = 86,
```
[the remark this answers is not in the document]
lzrd: Done

PR #1236, opened 2023-03-24
Add VPD identity to packets sent by udpbroadcast

### LINKED ISSUE #1216: SP broadcast could include baseboard model, revision, and ID
At present the SP announces its presence periodically via a UDP broadcast sent to port 8888.  The packet contents currently appears to include:

- the base MAC address (which we would have gotten from the VPD)
- the image ID (as you would see from `humility manifest | grep image.ID`)

At the moment, `pilot` maintains [a database](https://github.com/oxidecomputer/compliance-pilot/blob/main/macdb.txt) that [maps base MAC to the Gimlet or Sidecar serial number](https://github.com/oxidecomputer/compliance-pilot/blob/236091ab8856e00b9f33ffeed409488286db9692/src/cmds/mod.rs#L320-L330).  Without this database we cannot cheaply infer the identity of a particular SP -- we would need to make a request to each one, or we would need to do something less stateless than what we are doing now.  The upshot of the database is that old images don't know the identity of equipment manufactured after the last time `macdb.txt` was updated.

It would be really neat if we could extend the broadcast slightly to include the following information we can otherwise get from `macdb.txt`:

- baseboard model (e.g., `913-0000019`)
- baseboard revision (e.g., `6`)
- baseboard identifier (e.g., `BRM42220021`)

It would also be acceptable to just include the `BARC` string directly from the VPD if you'd prefer (e.g., `0XV1:9130000019:006:BRM42220021`) but I suspect it's already parsed inside the SP?

As a note for control plane implementation: it would probably be incorrect to depend on this broadcast for identity for something like MGS or Nexus.  In the `pilot` case we are in an engineering/test rack where we don't care about the security implications, but in production this message is totally unauthenticated and easily spoofable by someone who can physically alter the rack.

jgallagher:
Also adds a version field to the front of the packet and a byte indicating whether the VPD is valid (useful both for VPD-less boards like gimletlet and for detecting busted VPD).

From a wireshark capture of a gimletlet's broadcasts:

```
01 00 00 00                      # version
0e 1d 7d ef 9f 1d                # mac
80 bd fe ba e3 ab 59 27          # hubris image id
00                               # identity_valid
00 00 00 00 00 00 00 00 00 00 00 # identity_valid=0, so this is all zero
00 00 00 00                      # identity_valid=0, so this is all zero
00 00 00 00 00 00 00 00 00 00 00 # identity_valid=0, so this is all zero
```

and from a snoop capture of a gimlet:

```
0100 0000                   # version
a840 2504 0281              # mac
0b4d 9df2 0ef2 0809         # hubris image id
01                          # identity_valid
39 3133 2d30 3030 3030 3139 # part 913-0000019
0600 0000                   # revision 6
4252 4d34 3232 3230 3036 36 # serial BRM42220066
```

Fixes #1216

### DIFF
--- a/task/udpbroadcast/src/main.rs
+++ b/task/udpbroadcast/src/main.rs
@@ -5,16 +5,59 @@
 #![no_std]
 #![no_main]
 
+use hubpack::SerializedSize;
+use serde::Serialize;
 use task_net_api::*;
+use task_packrat_api::{Packrat, VpdIdentity};
 use userlib::*;
-use zerocopy::AsBytes;
 
 task_slot!(NET, net);
+task_slot!(PACKRAT, packrat);
+
+#[derive(Debug, Clone, Copy, Serialize, SerializedSize)]
+struct BroadcastData {
+    // Version for this data structure; adding new fields to the end is okay,
+    // but changing the order, size, or meaning of existing fields should result
+    // in a version bump.
+    version: u32,
+
+    mac_address: [u8; 6],
+    image_id: [u8; 8],
+
+    // If true, we have identity from our VPD, and the following three fields
+    // are populated. If false, we have no VPD or failed to read it, and the
+    // following three fields will be all zero.
+    identity_valid: bool,
+    part_number: [u8; VpdIdentity::PART_NUMBER_LEN],
+    revision: u32,
+    serial: [u8; VpdIdentity::SERIAL_LEN],
+}
+
+impl BroadcastData {
+    const CURRENT_VERSION: u32 = 1;
+}
+
+// Ensure our serialized size doesn't change unexpectedly: if you land here
+// because compilation has failed, consider whether you need to update
+// `BroadcastData::CURRENT_VERSION`!
+//
+// Current size is 45 bytes:
+// version (4)
+// mac_address (6)
+// image_id (8)
+// identity_valid (1)
+// part_number (11)
+// revision (4)
+// serial (11)
+static_assertions::const_assert_eq!(BroadcastData::MAX_SIZE, 45);
 
 #[export_name = "main"]
 fn main() -> ! {
     let net = NET.get_task_id();
     let net = Net::from(net);
+
+    let packrat = PACKRAT.get_task_id();
+    let packrat = Packrat::from(packrat);
 
     const SOCKET: SocketName = SocketName::broadcast;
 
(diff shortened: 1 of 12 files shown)


mkeeter:
Unfortunately, this packet format is quite load-bearing – it's used by both `humility rpc --listen` and [`pilot`](https://github.com/oxidecomputer/compliance-pilot/blob/c8032aa37651940ef797928a33c1b958dc22ad22/src/cmds/sp.rs#L1290) for SP discovery.

Both of these tools expect the MAC address to be first in the packet, so this PR will break them; Humility goes so far as to hard-code the expected length of [14 bytes](https://github.com/oxidecomputer/humility/blob/master/cmd/rpc/src/lib.rs#L167).

I suspect we'll want to patch those tools to handle > 14 byte packets with a new handler _before_ merging this in.

jgallagher:
Good call! I'll work on humility support Monday morning, and pilot support shortly thereafter unless Josh beats me to it.

jclulow:
We could also for some amount of time send two broadcasts?  One new and one old.  Old pilot discards anything that isn't 14 bytes so we'd just ignore the new ones.

jclulow:
FYI: I have updated `pilot` to understand both the old and new (as described here) format in https://github.com/oxidecomputer/compliance-pilot/commit/fd2badbf86d51e4bfca1bdf32367aa68bf330939.  It appeared to work on **lurch** when looking at **paris**, which seemed to have one system with the old broadcast and one with your WIP change?

jgallagher:
> FYI: I have updated `pilot` to understand both the old and new (as described here) format in [oxidecomputer/compliance-pilot@fd2badb](https://github.com/oxidecomputer/compliance-pilot/commit/fd2badbf86d51e4bfca1bdf32367aa68bf330939). It appeared to work on **lurch** when looking at **paris**, which seemed to have one system with the old broadcast and one with your WIP change?

Thanks! I'm working on humility now. I did indeed update one gimlet on **paris** with this branch (specifically to capture the snoop I put in the PR description) but forgot to flash it back - sorry!

jgallagher:
Humility PR: https://github.com/oxidecomputer/humility/pull/345

task/udpbroadcast/src/main.rs
```diff
@@ -5,32 +5,90 @@
...
+use task_packrat_api::{Packrat, VpdIdentity};
 use userlib::*;
-use zerocopy::AsBytes;
 
 task_slot!(NET, net);
+task_slot!(PACKRAT, packrat);
+
+#[derive(Debug, Clone, Copy, Serialize, SerializedSize)]
+struct BroadcastData {
+    // Version for this data structure; adding new fields to the end is okay,
```
mkeeter, line 19: Adding new fields to the end would **also** require a version bump, because `humility rpc` is expecting a specific size (once https://github.com/oxidecomputer/humility/pull/345 is merged)

mkeeter (approved):

PR #1237, opened 2023-03-27
Gimlet rev D support

### LINKED ISSUE #1210: Need Gimlet Rev D app.toml
We need to go through and add support for Rev D Gimlet's as there have been additional SP visible electrical changes including at least the reset GPIOs for the mux. This ticket will be updated to include known issues, but we will need to do a scan regardless.

mkeeter:
Closes #1210 

This is 90% plumbing, and 10% updating the I2C build system to support `nreset` pins for the PCA954x mux.  We already had an `enable` pin wired up, but it's only used on the LTC4306 (on the `gemini-bu` board), so I changed the name to match the common case going forward.


nathanaelhuffman (approved):
LGTM.  Gimlet A: Thank you for your service!

PR #1239, opened 2023-03-28
Support revamped transceiver management protocol

Aaron-Hartwig:


Aaron-Hartwig:
This will get its final edit/merge after https://github.com/oxidecomputer/transceiver-control/pull/71 has closed.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
+        let mut result = self.select_page(*mem.page(), modules);
 
         // Copy data into the FPGA write buffer
-        self.transceivers
-            .set_i2c_write_buffer(&data[..mem.len() as usize])
-            .map_err(|_e| Error::WriteFailed(HwError::WriteBufFailed))?;
+        result = result.chain(
+            self.transceivers
+                .set_i2c_write_buffer(&data[..mem.len() as usize]),
+        );
```
bnaecker, line 776: If I'm understanding `set_i2c_write_buffer()` correctly, I don't think we want to call `chain()` like this. The `set_i2c_write_buffer()` function appears to push data into the write buffers for _all_ transceivers, independent of the value of `modules`. So, unless one of them fails that operation, we'll return and then overwrite `result.success` with _all_ modules.

I can see this happening I think, by looking at either the host-side log messages, or the `ringbuf` entries. For example:

```
ben@niles ~ $ echo '02' | ./xcvradm -i axf4 -t 0 -l trace write-lower --sff 93 --input-kind hex
Mar 31 04:23:43.345 DEBG bound UDP socket, local_addr: Ok([::]:41304), interface: axf4
Mar 31 04:23:43.345 DEBG joined IPv6 multicast group, multicast_addr: ff02::1de:2
Mar 31 04:23:43.345 DEBG spawned IO task
Mar 31 04:23:43.345 TRCE received outgoing request, request: OutstandingHostRequest { request: HostRpcRequest { header: Header { version: 1, message_id: 0, message_kind: HostRequest }, message: Message { version: 1, body: HostRequest(Read { modules: ModuleId(0x1), read: MemoryRead { page: Sff8636(Lower), offset: 0, len: 1 } }) }, data: None }, n_retries: 0, n_error_messages: 0, response_tx: Sender { inner: Some(Inner { state: State { is_complete: false, is_closed: false, is_rx_task_set: false, is_tx_task_set: false } }) } }, task: io
Mar 31 04:23:43.345 TRCE sent outgoing request, message: Message { version: 1, body: HostRequest(Read { modules: ModuleId(0x1), read: MemoryRead { page: Sff8636(Lower), offset: 0, len: 1 } }) }, peer: [ff02::1de:2%2]:11112, task: io
Mar 31 04:23:43.350 TRCE packet received, peer: [fe80::aa40:25ff:fe05:ff00%2]:11112, n_bytes: 34, task: io
Mar 31 04:23:43.350 DEBG SP response message received, message: Message { version: 1, body: SpResponse(Read { modules: ModuleId(0x1), failed_modules: ModuleId(0x0), read: MemoryRead { page: Sff8636(Lower), offset: 0, len: 1 } }) }, peer: [fe80::aa40:25ff:fe05:ff00%2]:11112, task: io
Mar 31 04:23:43.351 TRCE received trailing data, n_bytes: 1, data: Some([17]), task: io
Mar 31 04:23:43.351 TRCE received outgoing request, request: OutstandingHostRequest { request: HostRpcRequest { header: Header { version: 1, message_id: 1, message_kind: HostRequest }, message: Message { version: 1, body: HostRequest(Write { modules: ModuleId(0x1), write: MemoryWrite { page: Sff8636(Lower), offset: 93, len: 1 } }) }, data: Some([2]) }, n_retries: 0, n_error_messages: 0, response_tx: Sender { inner: Some(Inner { state: State { is_complete: false, is_closed: false, is_rx_task_set: true, is_tx_task_set: false } }) } }, task: io
Mar 31 04:23:43.351 TRCE sent outgoing request, message: Message { version: 1, body: HostRequest(Write { modules: ModuleId(0x1), write: MemoryWrite { page: Sff8636(Lower), offset: 93, len: 1 } }) }, peer: [ff02::1de:2%2]:11112, task: io
Mar 31 04:23:43.361 TRCE packet received, peer: [fe80::aa40:25ff:fe05:ff00%2]:11112, n_bytes: 33, task: io
Mar 31 04:23:43.362 DEBG SP response message received, message: Message { version: 1, body: SpResponse(Write { modules: ModuleId(0xffffffff), failed_modules: ModuleId(0x0), write: MemoryWrite { page: Sff8636(Lower), offset: 93, len: 1 } }) }, peer: [fe80::aa40:25ff:fe05:ff00%2]:11112, task: io
Mar 31 04:23:43.362 TRCE received trailing data, n_bytes: 0, data: Some([]), task: io
```

In that second-to-last log line, we print the `SpResponse::Write` we got back, which does indeed show all 1's for the written modules. The `ringbuf` shows the same:

```
ben@niles ~ $ pfexec humility --target sidecar-b-aaron ringbuf transceivers
humility: ring buffer drv_transceivers_server::__RINGBUF in transceivers:
   6  238        6        1 GotInterface(0x1, Sff8636)
   7  252        6        1 TemperatureReadError(0x2, ImplError(0x4))
   8  238        6        1 GotInterface(0x3, Sff8636)
   9  238        6        1 GotInterface(0x0, Sff8636)
  10  238        6        1 GotInterface(0x2, Sff8636)
  11  238        6        1 GotInterface(0x4, Sff8636)
  12  238        6        1 GotInterface(0x6, Sff8636)
  13  238        6        1 GotInterface(0x8, Sff8636)
  14  238        6        1 GotInterface(0xa, Sff8636)
  15  238        6        1 GotInterface(0xc, Sff8636)
   0  238        7        1 GotInterface(0xe, Sff8636)
   1  238        7        1 GotInterface(0x10, Sff8636)
   2  238        7        1 GotInterface(0x12, Sff8636)
   3  238        7        1 GotInterface(0x14, Sff8636)
   4  238        7        1 GotInterface(0x18, Sff8636)
   5  238        7        1 GotInterface(0x1a, Sff8636)
humility: ring buffer drv_transceivers_server::udp::__RINGBUF in transceivers:
   3  272       95        1 ResponseSize(ResponseSize { header_length: 0xa, message_length: 0x17, data_length: 0x3 })
   4  319       95        1 Read(ModuleId(0x7), MemoryRead { page: Sff8636(Lower), offset: 0x5d, len: 0x1 })
   5  332       95        1 OperationResult(ModuleResult { success: LogicalPortMask(0x7), failure: LogicalPortMask(0x0), error: LogicalPortMask(0x0) })
   6  272       95        1 ResponseSize(ResponseSize { header_length: 0xa, message_length: 0x17, data_length: 0x3 })
   7  319       95        1 Read(ModuleId(0x1), MemoryRead { page: Sff8636(Lower), offset: 0x0, len: 0x1 })
   8  332       95        1 OperationResult(ModuleResult { success: LogicalPortMask(0x1), failure: LogicalPortMask(0x0), error: LogicalPortMask(0x0) })
   9  272       95        1 ResponseSize(ResponseSize { header_length: 0xa, message_length: 0x17, data_length: 0x1 })
  10  353       95        1 Write(ModuleId(0x1), MemoryWrite { page: Sff8636(Lower), offset: 0x5d, len: 0x1 })
  11  365       95        1 OperationResult(ModuleResult { success: LogicalPortMask(0xffffffff), failure: LogicalPortMask(0x0), error: LogicalPortMask(0x0) })
  12  272       95        1 ResponseSize(ResponseSize { header_length: 0xa, message_length: 0x17, data_length: 0x0 })
  13  319       95        1 Read(ModuleId(0x1), MemoryRead { page: Sff8636(Lower), offset: 0x0, len: 0x1 })
  14  332       95        1 OperationResult(ModuleResult { success: LogicalPortMask(0x1), failure: LogicalPortMask(0x0), error: LogicalPortMask(0x0) })
  15  272       95        1 ResponseSize(ResponseSize { header_length: 0xa, message_length: 0x17, data_length: 0x1 })
   0  353       96        1 Write(ModuleId(0x1), MemoryWrite { page: Sff8636(Lower), offset: 0x5d, len: 0x1 })
   1  365       96        1 OperationResult(ModuleResult { success: LogicalPortMask(0xffffffff), failure: LogicalPortMask(0x0), error: LogicalPortMask(0x0) })
   2  272       96        1 ResponseSize(ResponseSize { header_length: 0xa, message_length: 0x17, data_length: 0x0 })
ben@niles ~ $
```

drv/transceivers-server/src/udp.rs
```diff
@@ -112,11 +94,34 @@ impl ServerImpl {
...
+                                self.handle_message(msg, header.message_id, data, tx_data_buf)
+                            },
+                            // try to tell the host something useful about what happened
+                            Err(e) => {
+                                ringbuf_entry!(Trace::DeserializeError(e));
+                                self.handle_deserialization_error(header, tx_data_buf)
+                            },
+                        },
+                        None => {
+                            ringbuf_entry!(Trace::SizeUnderflowError);
```
bnaecker, line 115: I think I might call this `TruncatedPacket` or something like that. It's true that the subtraction underflowed, but what this really means is we received a short packet. It also means someone's really not doing UDP correctly, but that's another issue :)

drv/transceivers-server/src/udp.rs
```diff
@@ -149,64 +154,57 @@ impl ServerImpl {
...
+            None
+
+        // In this case, the message has failed to deserialize, but
+        // we've successfully deserialized the header. That implies that
+        // the host has sent us a message that does not exist in our own
+        // implementation of the protocol.
+        //
+        // Check that implication by ensuring that the host version is
+        // after our own CURRENT.
+        } else if header.version() > version::outer::CURRENT {
```
bnaecker, line 182: I think this captures some of my ambivalence about the inner / outer protocol version split.

What we really want to check here is the _inner_ protocol version. The logic is: we've deserialized the header but not the message, implying that the message is too new for us. But we can't really check that, since we can't deserialize the inner version. So this is only the correct check if the inner and outer versions are the same, which limits the value of the split.

I think we have two options here:

- Make a single version
- Try to deserialize the _version alone_, i.e., use the knowledge about the inner format of the `Message` type to read the first octet of the remaining bytes of the message, which should have the version.

I'm really not sure which to take. Making the inner / outer versions the same would be the easiest, but then we'd probably want to go back to a single message type that contains both the header and the body. That's also fine, it's just more churn that I think we're both hoping to avoid.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
-        if msg.header.version < version::MIN {
-            ringbuf_entry!(Trace::WrongVersion(msg.header.version));
+        if msg.version() < version::inner::MIN {
+            ringbuf_entry!(Trace::WrongVersion(msg.version()));
             return None;
         }
 
-        let (reply, data_len) = match msg.body {
+        // If the message is something we don't expect or something we cannot
+        // reply to
```
bnaecker, line 231: Missing the back half of this sentence, maybe?

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
-            version: version::CURRENT,
-            ..msg.header
-        };
-        let out = Message {
-            header,
-            modules: msg.modules,
-            body: MessageBody::SpResponse(reply),
-        };
-        let msg_len = hubpack::serialize(tx_data_buf, &out).unwrap();
+        // Serialize the Header into the front of the tx buffer, followed by
```
bnaecker, line 262: I might make a note that any trailing data is already in the buffer by this point, as part of `handle_host_request()`.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
         out: &mut [u8],
-    ) -> Result<(SpResponse, usize), Error> {
+    ) -> (MessageBody, usize) {
         match h {
-            HostRequest::Status => {
+            HostRequest::Status(modules) => {
                 ringbuf_entry!(Trace::Status(modules));
-                let count = self.get_status(modules, out)?;
-                Ok((SpResponse::Status, count))
+                let mask = LogicalPortMask(modules.0 as u32);
```
bnaecker, line 298: Might be useful to `impl From<ModuleId> for LogicalPortMask`, to avoid the casting in many places.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
-            HostRequest::Read(mem) => {
-                ringbuf_entry!(Trace::Read(modules, mem));
-                let out_size = mem.len() as u32 * modules.ports.0.count_ones();
-                if out_size as usize > transceiver_messages::MAX_PAYLOAD_SIZE {
-                    return Err(Error::RequestTooLarge);
+            HostRequest::Read { modules, read } => {
+                ringbuf_entry!(Trace::Read(modules, read));
+                let mask = LogicalPortMask(modules.0 as u32);
+                let request_size = read.len() as u32 * mask.0.count_ones();
+                if request_size as usize
```
bnaecker, line 322: `request_size` can be retrieved as a `usize` already using `HostRequest::expected_data_len()`, if that's helpful!

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
+                        MessageBody::Error(ProtocolError::RequestTooLarge),
+                        0,
+                    );
                 }
-                self.read(mem, modules, out)?;
-                Ok((SpResponse::Read(mem), out_size as usize))
+
+                let result = self.read(read, mask, out);
+                ringbuf_entry!(Trace::OperationResult(result));
+                let success = ModuleId(result.success().0 as u64);
```
bnaecker, line 333: The other direction of conversions might be useful too? `impl From<LogicalPortMask> for ModuleId`?

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
+                    match failure_type {
+                        Some(e) => {
+                            out[error_idx] = e as u8;
+                            error_idx += 1;
+                        }
+                        None => {
+                            ringbuf_entry!(Trace::UnexpectedFailure(ModuleId(
+                                module.0 as u64
+                            )));
+                            panic!();
```
bnaecker, line 563: I might add a useful panic message here. One could also make this `unreachable!()` which more strongly indicates the programmer definitely didn't intend this to be possible.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
+        let requested_invalid_modules = ModuleId(modules.0 & invalid_modules.0);
+        // iterate through the results and populate errors in the response buffer
+        // failure: whatever `HwError` specified by `failure_type`
+        // error: fpga communication issue
+        for module in modules.to_indices().map(LogicalPort) {
+            if module <= LogicalPortMask::MAX_PORT_INDEX {
+                if result.failure().is_set(module) {
+                    match failure_type {
+                        Some(e) => {
+                            out[error_idx] = e as u8;
```
bnaecker, line 556: I'm not sure casting with `as` is quite right. I know support for that [was added](https://doc.rust-lang.org/reference/items/enumerations.html#casting), but I think we probably want to lean on `hubpack::serialize` here too. That would allow us to add variants that carry data, without changing this code.

Not a huge deal I guess, since this cast should fail if that becomes the case.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
         // registers in the transceiver to select it.
         if let Some(page) = page.page() {
-            self.transceivers
-                .set_i2c_write_buffer(&[page])
-                .map_err(|_e| HwError::PageSelectWriteBufFailed)?;
-            self.transceivers
-                .setup_i2c_write(PAGE_SELECT, 1, mask)
-                .map_err(|_e| HwError::PageSelectWriteFailed)?;
-            self.wait_and_check_i2c(mask)?;
+            result =
```
bnaecker, line 658: Same point here about chaining with `set_i2c_write_buffer()`. I think that will ensure that we almost always return all 1's as the successful modules from this method.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
 
         if let Some(bank) = page.bank() {
-            self.transceivers
-                .set_i2c_write_buffer(&[bank])
-                .map_err(|_e| HwError::BankSelectWriteBufFailed)?;
-            self.transceivers
-                .setup_i2c_write(BANK_SELECT, 1, mask)
-                .map_err(|_e| HwError::BankSelectWriteFailed)?;
-            self.wait_and_check_i2c(mask)?;
+            result =
```
bnaecker, line 674: Here as well.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,572 @@ impl ServerImpl {
...
-            .map_err(|_e| HwError::WaitFailed)?;
-        if !err_mask.left.is_empty() {
-            // FPGA reported an I2C error
-            return Err(HwError::I2cError(PortMask(err_mask.left.get())));
-        } else if !err_mask.right.is_empty() {
-            return Err(HwError::I2cError(PortMask(err_mask.right.get())));
-        }
-        Ok(())
+    // Empty Ok() result if there were no I2C errors. The `LogicalPortMask`
+    // returned by the Err represents which ports had I2C errors.
```
bnaecker, line 687: I think this comment is out of date now? It refers to a `Result`, but looks like we're now returning a `ModuleResult`.

bnaecker (commented):
This is awesome! I've run a bunch of manual tests, and things seem really solid for the most port. There is one issue I've noted about returning all 1's.

I've also got a few questions and small suggestions, but overall everything looks great. Thanks for all the hard work getting this beast to this point!

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -250,43 +388,82 @@ const PORT_MAP: [PortLocation; 32] = [
...
+        self.error
+    }
+
+    /// This is a helper function to combine two sets of results for when there
+    /// is a sequence of `ModuleResult` yielding function calls. Building such a
+    /// sequence is generally done with the following form (where `modules` is
+    /// a `LogicalPortMask` of requested modules):
+    ///
+    /// let result = some_result_fn(modules);
+    /// let next_result = result.chain(another_result_fn(result.success))
```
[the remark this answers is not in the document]
Aaron-Hartwig: Good catch! Forgot to update that comment after making those fields private.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -250,43 +388,82 @@ const PORT_MAP: [PortLocation; 32] = [
...
+    ///
+    /// So the initial result includes some set of success, failure, and error
+    /// masks which then need to be reconciled with a new set of masks, generally
+    /// a subset of the success mask of the initial result. Notably, there
+    /// cannot be overlap between these masks, which this function enforces.
+    pub fn chain(&self, next: ModuleResult) -> Self {
+        // success mask is just what the success of the next step is
+        let success = next.success;
```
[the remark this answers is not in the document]
Aaron-Hartwig: Yeah, that is a miss on my part! I'll add this line above the current 458:
```rust
assert!(next.success().to_indices().all(|idx| self.success().is_set(LogicalPort(idx))));
```
This will assert that every index of found in `next.success` can also be found in `self.success`.

drv/transceivers-server/src/main.rs
```diff
@@ -41,7 +41,8 @@ task_slot!(SENSOR, sensor);
 // Both incoming and outgoing messages use the Message type, so we use it to
 // size our Tx / Rx buffers.
 const MAX_UDP_MESSAGE_SIZE: usize =
```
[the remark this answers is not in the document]
Aaron-Hartwig: Excellent, I'll point to that instead!

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: That isn't helpful while handling a `HostRequest::Read` since it is just meant to return the amount of data to write via `HostRequest::Write`.

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: I'll make this `unreachable!()` and elaborated in the comment above the function. My understanding is that in Hubris we try to be very sparing on including `panic!()` messages since that adds program text, so we try to include useful information in more compact ways.

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: Now _here_ would have been a problem since the next call to `setup_i2c_write` uses `result.success()` as its mask. Given that these are SFF devices (and won't have a `bank`), not CMIS, I don't think this is the root cause of the bug you exhibit above either.

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: Done! I changed all the error serializations in this function to be of the form
```rust
let err_size = hubpack::serialize(&mut out[error_idx..], &some_err).unwrap();
error_idx += err_size;
```
which should help make this a little more future proof.

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: I...don't know what I was going to say here. Removing.

drv/transceivers-server/src/udp.rs
bnaecker: That makes sense in the constrained environment!

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -77,6 +90,144 @@ impl FpgaPortMasks {
...
 
+/// Represents a single logical port (0-31)
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
+pub struct LogicalPort(pub u8);
+impl LogicalPort {
+    pub fn as_mask(&self) -> LogicalPortMask {
+        LogicalPortMask(1 << self.0)
+    }
+
+    pub fn get_physical_location(&self) -> PortLocation {
```
mkeeter, line 101: This could delegate to `PortLocation::from(*self)`, instead of looking into `PORT_MAP` itself.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -77,6 +90,144 @@ impl FpgaPortMasks {
...
+    pub fn count(&self) -> usize {
+        self.0.count_ones() as _
+    }
+    pub fn is_set(&self, index: LogicalPort) -> bool {
+        self.0 & index.as_mask().0 != 0
+    }
+    pub fn is_empty(&self) -> bool {
+        self.0 == 0
+    }
+    pub fn to_indices(&self) -> impl Iterator<Item = u8> + '_ {
```
mkeeter, line 133: It would be more type-safe to return an `impl Iterator<Item = LogicalPort>` here; any reason we couldn't do that?

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -77,6 +90,144 @@ impl FpgaPortMasks {
...
+    }
+    pub fn count(&self) -> usize {
+        self.0.count_ones() as _
+    }
+    pub fn is_set(&self, index: LogicalPort) -> bool {
+        self.0 & index.as_mask().0 != 0
+    }
+    pub fn is_empty(&self) -> bool {
+        self.0 == 0
+    }
```
mkeeter, line 132: I'd suggest rewriting these using the bitwise operations implemented below, e.g.
```suggestion
    pub fn set(&mut self, index: LogicalPort) {
        *self |= index.as_mask()
    }
    pub fn clear(&mut self, index: LogicalPort) {
        *self &= !index.as_mask()
    }
    pub fn merge(&mut self, other: LogicalPortMask) {
        *self |= other
    }
    pub fn count(self) -> usize {
        self.0.count_ones() as _
    }
    pub fn is_set(self, index: LogicalPort) -> bool {
        !(self & index.as_mask()).is_empty()
    }
    pub fn is_empty(self) -> bool {
        self.0 == 0
    }
```
(I also switched from `&self` to just `self` in a few operations, since this object is trivial to clone)

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -77,6 +90,144 @@ impl FpgaPortMasks {
...
+        let mut logical_mask: u32 = 0;
+        for port in 0..NUM_PORTS {
+            let logical_port = PORT_MAP[port as usize];
+            let bits = match logical_port.controller {
+                FpgaController::Left => mask.left,
+                FpgaController::Right => mask.right,
+            };
+            let port_mask = logical_port.port.as_mask();
+            if bits.is_set(logical_port.port)
+                && port_mask.is_set(logical_port.port)
```
mkeeter, line 200: This check seems superfluous, because `port_mask` is constructed from `logical_port.port`; I removed it in my rewrite of the function (see above comment).

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -77,6 +90,144 @@ impl FpgaPortMasks {
...
+
+impl From<LogicalPortMask> for ModuleId {
+    fn from(v: LogicalPortMask) -> Self {
+        Self(v.0 as u64)
+    }
+}
+
+// It is convenient to have the ergonomics for a LogicalPortMask resemble the
+// bitwise mask that it represents, so we implement some bitwise operations.
+impl core::ops::BitOr for LogicalPortMask {
```
mkeeter, line 152: I like this!

I would be tempted to go farther and also implement `BitOr` and `BitOrAssign` to combine `LogicalPortMask | LogicalPort`, instead of needing to call `as_mask()` on the latter.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -77,6 +90,144 @@ impl FpgaPortMasks {
...
+    }
+}
+
+// Maps logical port `mask` to physical FPGA locations
+impl From<LogicalPortMask> for FpgaPortMasks {
+    fn from(mask: LogicalPortMask) -> FpgaPortMasks {
+        let mut fpga_port_masks = FpgaPortMasks::default();
+        for (i, port_loc) in PORT_MAP.iter().enumerate() {
+            let port_mask: u32 = 1 << i;
+            if (mask.0 & port_mask) != 0 {
```
mkeeter, line 216: Continuing my campaign for Stronger Types, how about
```suggestion
            let this_port_mask = LogicalPort(i as u8);
            if mask.is_set(this_port_mask) {
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -86,8 +237,8 @@ impl FpgaPortMasks {
 /// the FPGAs share code, resulting in each one reporting in terms of ports
 /// 0-15.
 ///
-/// This is the FPGA -> logical mapping.
-const PORT_MAP: [PortLocation; 32] = [
+/// This is the logical -> physical mapping.
+const PORT_MAP: [PortLocation; NUM_PORTS as usize] = [
```
mkeeter, line 241: Continuing the drive for type-safety, I wonder if we should enforces only indexing `PORT_MAP` with a `LogicalPort`, e.g. a new type:

```rust
struct PortMap([PortLocation; NUM_PORTS as usize]);

impl core::ops::Index<LogicalPort> for PortMap {
    type Output = PortLocation;

    fn index(&self, i: LogicalPort) -> &Self::Output {
        self.0[i.0]
    }
}

const PORT_MAP: PortMap = PortMap([
    // ...
]);
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -250,43 +401,94 @@ const PORT_MAP: [PortLocation; 32] = [
...
+    /// Create a new ModuleResult which enforces no overlap in the success,
+    /// failure, and error masks.
+    pub fn new(
+        success: LogicalPortMask,
+        failure: LogicalPortMask,
+        error: LogicalPortMask,
+    ) -> Result<Self, TransceiversError> {
+        if (success.0 & failure.0 != 0)
+            || (success.0 & error.0 != 0)
+            || (failure.0 & error.0 != 0)
```
mkeeter, line 434:
```suggestion
        if !(success & failure).is_empty()
            || !(success & error).is_empty()
            || !(failure & error).is_empty()
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -322,100 +530,99 @@ impl Transceivers {
             if !mask.is_empty() {
                 let fpga = self.fpga(fpga_index);
                 let wdata: U16<byteorder::LittleEndian> = U16::new(mask.get());
-                fpga.write(op, addr, wdata)?;
+                // mark that an error occurred so we can modify the success mask
+                if fpga.write(op, addr, wdata).is_err() {
+                    match fpga_index {
+                        FpgaController::Left => error |= LEFT_LOGICAL_MASK,
```
mkeeter, line 536: Does it make sense to set the error for the entire logical mask associated with the FPGA?

e.g. if I provide `LogicalPortMask(1)` and the `fpga.write` fails, having 16 error bits in the returned mask could be confusing.

mkeeter: Never mind; you mask it out below!

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -322,100 +530,99 @@ impl Transceivers {
...
+        let ldata: Option<[U16<byteorder::LittleEndian>; 8]> =
+            match self.fpga(FpgaController::Left).read(Addr::QSFP_POWER_EN0) {
+                Ok(data) => Some(data),
+                Err(_) => None,
+            };
+        let rdata: Option<[U16<byteorder::LittleEndian>; 8]> =
+            match self.fpga(FpgaController::Right).read(Addr::QSFP_POWER_EN0) {
+                Ok(data) => Some(data),
+                Err(_) => None,
+            };
```
mkeeter, line 624:
```suggestion
        let ldata: Option<[U16<byteorder::LittleEndian>; 8]> = self
            .fpga(FpgaController::Left)
            .read(Addr::QSFP_POWER_EN0)
            .ok();
        let rdata: Option<[U16<byteorder::LittleEndian>; 8]> = self
            .fpga(FpgaController::Right)
            .read(Addr::QSFP_POWER_EN0)
            .ok();
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -425,8 +632,14 @@ impl Transceivers {
 
             // get the relevant data from the correct FPGA
             let local_data = match port_loc.controller {
-                FpgaController::Left => &f0,
-                FpgaController::Right => &f1,
+                FpgaController::Left => match ldata {
```
mkeeter, line 635: I can't apply a suggestion to the correct block, but you could rewrite this section of code as
```rust
            // get the relevant data from the correct FPGA
            let local_data = match port_loc.controller {
                FpgaController::Left => ldata,
                FpgaController::Right => rdata,
            };
            let Some(local_data) = local_data else { continue };
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -457,54 +686,86 @@ impl Transceivers {
...
+                            )
+                            .is_err()
+                    {
+                        match fpga_index {
+                            FpgaController::Left => {
+                                error |= LEFT_LOGICAL_MASK;
+                            }
+                            FpgaController::Right => {
+                                error |= RIGHT_LOGICAL_MASK;
+                            }
```
mkeeter, line 705:
```suggestion
                        error |= match fpga_index {
                            FpgaController::Left => LEFT_LOGICAL_MASK,
                            FpgaController::Right => RIGHT_LOGICAL_MASK,
                        }
```

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -695,38 +977,59 @@ impl Transceivers {
...
+    /// error: an `FpgaError` occurred
     pub fn wait_and_check_i2c(
         &mut self,
-        mask: FpgaPortMasks,
-    ) -> Result<FpgaPortMasks, FpgaError> {
-        let mut out = FpgaPortMasks::default();
+        mask: LogicalPortMask,
+    ) -> ModuleResult {
+        let mut physical_failure: FpgaPortMasks = FpgaPortMasks::default();
+        let mut physical_error: FpgaPortMasks = FpgaPortMasks::default();
```
mkeeter, line 990: I don't think you need to specify the types here, since they're obvious on the RHS.
```suggestion
        let mut physical_failure = FpgaPortMasks::default();
        let mut physical_error = FpgaPortMasks::default();
```
(unless you want symmetry with the line below, for _A E S T H E T I C S_)

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -738,13 +1041,18 @@ impl Transceivers {
                 let has_err = (err & 0b1000) != 0;
                 if has_err {
                     match fpga_index {
-                        FpgaController::Left => out.left.set(port),
-                        FpgaController::Right => out.right.set(port),
+                        FpgaController::Left => physical_failure.left.set(port),
+                        FpgaController::Right => {
+                            physical_failure.right.set(port)
+                        }
```
mkeeter, line 1047:
```suggestion
                        FpgaController::Left => physical_failure.left,
                        FpgaController::Right => physical_failure.right,
                    }
                    .set(port)
```
🤷🏻‍♂️

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -695,38 +977,59 @@ impl Transceivers {
...
+                    // error on all of that FPGA's ports
+                    Err(_) => {
+                        match fpga_index {
+                            FpgaController::Left => {
+                                physical_error.left = PhysicalPortMask(0xffff)
+                            }
+                            FpgaController::Right => {
+                                physical_error.right = PhysicalPortMask(0xffff)
+                            }
+                        };
```
mkeeter, line 1016: Optional:
```suggestion
                        *(match fpga_index {
                            FpgaController::Left => &mut physical_error.left,
                            FpgaController::Right => &mut physical_error.right,
                        }) = PhysicalPortMask(0xffff);
```

mkeeter: On second thought, what about adding new functions to `FpgaPortMasks`?

```rust
impl FpgaPortMasks {
    fn get(&self, p: FpgaController) -> PhysicalPortMask {
        // ...
    }

    fn get_mut(&mut self, p: FpgaController) -> &mut PhysicalPortMask {
        // ...
    }
```

That would save a bunch of similar `match` statements.

drv/sidecar-front-io/src/transceivers.rs
mkeeter: On second thought, [do what I said above](https://github.com/oxidecomputer/hubris/pull/1239/files#r1154776203), then this becomes
```
physical_failure.get_mut(fpga_index).set(port);
```

drv/transceivers-server/src/main.rs
```diff
@@ -320,92 +313,108 @@ impl ServerImpl {
...
         _msg: &userlib::RecvMessage,
         logical_port_mask: u32,
     ) -> Result<(), idol_runtime::RequestError<TransceiversError>> {
-        self.transceivers
-            .enable_power(LogicalPortMask(logical_port_mask))
-            .map_err(TransceiversError::from)?;
-        Ok(())
+        let mask = LogicalPortMask(logical_port_mask);
+        let result = self.transceivers.enable_power(mask);
+        match (result.error() & mask).count() {
```
mkeeter, line 336: I don't think the `& mask` is necessary here (or in any of the functions below), since the result should already be masked.

drv/transceivers-server/src/main.rs
```diff
@@ -320,92 +313,108 @@ impl ServerImpl {
...
-        Ok(self
-            .transceivers
-            .get_modules_status()
-            .map_err(TransceiversError::from)?)
+        let (mod_status, result) = self.transceivers.get_module_status();
+
+        match result.error().count() {
+            0 => Ok(mod_status),
+            _ => Err(RequestError::from(TransceiversError::FpgaError)),
+        }
```
mkeeter, line 326: I have a slight preference for the following:
```suggestion
        if result.error().is_empty() {
            Ok(mod_status)
        } else {
            Err(RequestError::from(TransceiversError::FpgaError)),
        }
```

since it's a little more obvious, though it's one extra LOC.

(same applies for all the functions below if you also prefer it)

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: Nope! Done.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -77,6 +90,144 @@ impl FpgaPortMasks {
...
+        self.0 == 0
+    }
+    pub fn to_indices(&self) -> impl Iterator<Item = u8> + '_ {
+        (0..32).filter(|i| self.is_set(LogicalPort(*i)))
+    }
+}
+
+impl From<ModuleId> for LogicalPortMask {
+    fn from(v: ModuleId) -> Self {
+        Self(v.0 as u32)
```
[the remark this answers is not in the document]
Aaron-Hartwig: `ModuleId` is a u64. We chose something bigger than we needed just to provide a bit of future proofing since this type is going to get baked into the protocol. Since we only support 32 logical ports right now, we can ignore the upper 32 bits.

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: yeah, you are correct!

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: Done locally.

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: We chatted a bit in DMs - I implemented `PortMap` and some supporting `impl`s for a Stronger Type.

drv/sidecar-front-io/src/transceivers.rs
Aaron-Hartwig: 🧠 this is really cleaning up a number of gross spots. Great recommendation!

drv/transceivers-server/src/main.rs
Aaron-Hartwig: Yes and no. All these pin set/clear methods share the failure mode that the only problem we can notice is if FPGA communication is unsuccessful, which will return a `result.error()` mask which corresponds to _all_ the modules on the affected FPGA. So in that case, an operation on `mask=LogicalPortMask(0x1)` would yield a `result.error = LogicalPortMask(0x00ff00ff)`. I was masking that off again because my brain was still in UDP mode where I would actually change my response payload to show an `error` on _only_ the requested ports. In the Idol server this doesn't matter, so I will remove the masking.

drv/transceivers-server/src/main.rs
Aaron-Hartwig: I think the `is_empty()` option is better as it is clearer by returning a `bool` and not relying on `number != 0 == True`.

drv/transceivers-server/src/main.rs
```diff
@@ -162,7 +155,7 @@ impl ServerImpl {
         port: LogicalPort,
         reg: u8,
     ) -> Result<Celsius, FpgaError> {
-        self.transceivers.setup_i2c_read(reg, 2, port.as_mask())?;
+        self.transceivers.setup_i2c_read(reg, 2, port.as_mask());
```
mkeeter, line 158: Do we care about any error values here?

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,588 @@ impl ServerImpl {
...
+                            let err_size =
+                                hubpack::serialize(&mut out[error_idx..], &e)
+                                    .unwrap();
+                            error_idx += err_size;
+                        }
+                        None => {
+                            ringbuf_entry!(Trace::UnexpectedFailure(
+                                ModuleId::from(module.as_mask())
+                            ));
+                            unreachable!();
```
mkeeter, line 570: Nit: this is still plenty reachable, so I'd recommend a `panic!` instead.

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,588 @@ impl ServerImpl {
...
+        .unwrap();
+
+        // Write one bitfield per active port in the ModuleId which was
+        // successfully retrieved above.
         let mut count = 0;
-        for mask in modules.ports.to_indices().map(|i| 1 << i) {
+        for mask in modules
+            .to_indices()
+            .map(|p| p.as_mask())
+            .filter(|&m| !(desired_result.success() & m).is_empty())
```
mkeeter, line 625: `desired_result.is_set(m)` perhaps?

mkeeter: or even collapse the `map` statement, e.g.

```rust
        for mask in modules
            .to_indices()
            .filter(|p| desired_result.success().is_set(p.as_mask()))
```
(untested)

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,588 @@ impl ServerImpl {
...
             }
-            HostRequest::Read(mem) => {
-                ringbuf_entry!(Trace::Read(modules, mem));
-                let out_size = mem.len() as u32 * modules.ports.0.count_ones();
-                if out_size as usize > transceiver_messages::MAX_PAYLOAD_SIZE {
-                    return Err(Error::RequestTooLarge);
+            HostRequest::Read { modules, read } => {
+                ringbuf_entry!(Trace::Read(modules, read));
+                let mask = LogicalPortMask::from(modules);
+                let request_size = read.len() as u32 * mask.0.count_ones();
```
mkeeter, line 321: What happens if I send a `HostRequestRead { modules: 0xFFFFFFFF_FFFFFFFF, read: MemoryRead { .. len: 32} }`?

It will success at this check, because `read.len() * mask.0.count_ones()` will be exactly 1024; however, the call to `handle_errors` below will _also_ try serializing a bunch of `HwError::InvalidModuleIndex` into the trailing buffer, resulting in a panic (I think!).

drv/transceivers-server/src/udp.rs
bnaecker: Ha, I think I originally recommended `unreachable!()` instead, sorry about that Aaron! I'm actually curious why you'd recommend that @mkeeter.

mkeeter: I normally use `unreachable!` if it's **immediately obvious from surrounding context** that you can't get there, but the compiler isn't smart enough to prove it, e.g.
```rust
match i % 4 {
    0 => print!("HI"),
    1 => print!("THERE"),
    2 => print!("omg"),
    3 => print!("wtf"),
    _ => unreachable!(),
}
```

(The [Rust docs](https://doc.rust-lang.org/std/macro.unreachable.html) basically agree with this philosophy)

In this case, it's not immediately obvious from context, so I'd vote for a `panic!`.

drv/transceivers-server/src/udp.rs
```diff
@@ -112,11 +94,34 @@ impl ServerImpl {
...
     ) {
         let out_len =
-            match hubpack::deserialize(&rx_data_buf[0..meta.size as usize]) {
-                Ok((msg, data)) => self.handle_message(msg, data, tx_data_buf),
+            // attempt to deserialize the header
+            match hubpack::deserialize::<Header>(rx_data_buf) {
+                Ok((header, data)) => {
+                    // header deserialized successfully, so now attempt to
+                    // deserialize the remaining message
+                    match (meta.size as usize).checked_sub(Header::MAX_SIZE) {
```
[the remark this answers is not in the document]
mkeeter: As an alternative, we could slice it up at L98, e.g.
```
match hubpack::deserialize::<Header>(rx_data_buf[..meta.size as usize])
```
which would mean that the `data` returned would be exactly the post-header message data.

drv/transceivers-server/src/udp.rs
```diff
@@ -149,64 +154,57 @@ impl ServerImpl {
         }
     }
 
-    /// At this point, deserialization has failed, so we can't handle the
-    /// packet.  We'll attempt to deserialize *just the header* (which should
+    /// At this point, Message deserialization has failed, so we can't handle
+    /// the packet.  We'll look at *just the header* (which should
```
mkeeter, line 158: Nit: weird text wrapping?

drv/transceivers-server/src/udp.rs
```diff
@@ -149,64 +154,57 @@ impl ServerImpl {
...
+        // Check that implication by ensuring that the host version is
+        // after our own CURRENT.
+        } else if header.version() > version::outer::CURRENT {
+            let header_size = hubpack::serialize(
+                tx_data_buf,
+                &Header::new(header.message_id, header.message_kind),
+            )
+            .unwrap();
+            let message_size = hubpack::serialize(
+                tx_data_buf,
```
mkeeter, line 189:
```suggestion
                tx_data_buf[header_size..],
```
otherwise, I think we overwrite the header by accident

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,588 @@ impl ServerImpl {
...
-                }
+            MessageBody::HostRequest(h) => self.handle_host_request(
+                h,
+                data,
+                &mut tx_data_buf[reserved_framing..],
+            ),
+            // Nothing implemented yet
+            MessageBody::Error(e) => {
+                ringbuf_entry!(Trace::GotError(e));
+                return None;
```
mkeeter, line 256: Dumb question: why does this branch (and `HostResponse` above)return `None` isntead of `WrongMessage`?

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: Sold!

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: Good idea! `p` is a `LogicalPort`, so we can just do `.is_set(p)`, no need for the `as_mask()`.

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: Correct! My current understanding is that the host will not actually attempt to address the upper 32 bits right now, but obviously we'd like to gracefully handle should that occur. I could count how many bits are set in the upper 32, multiply that by `HwError::MAX_SIZE`, and then add that to `request_size` since we know for a fact we will be serializing errors there.

@bnaecker just a FYI in case you have anything to add here.

drv/transceivers-server/src/udp.rs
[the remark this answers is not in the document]
Aaron-Hartwig: I like that suggestion, it is a cleaner implementation.

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: Good catch!

drv/transceivers-server/src/udp.rs
bnaecker: The host won't try to set any of the upper 32 bits, but there's not much preventing that from happening in the future. Your idea of adding that to the request size and then possibly failing it sounds good.

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: There are no dumb questions, Matt ;). In the case of `MessageBody::Error`, a `WrongMessage` would not be the correct response since 1) receiving an `MessageBody::Error` message will be perfectly reasonable once we actually start using `MessageBody::SpRequest` and 2) we don't respond to `Error` messages. A similar line of logic applies to `MessageBody::HostResponse`. That will be a valid message type once we implement `SpRequest`s.

I think the `MessageBody::HostResponse` branch should be changed to respond with `ProtocolError::NotSupported` in the interim, but I don't think we should respond in the `MessageBody::Error` branch.

@bnaecker FYI

drv/transceivers-server/src/udp.rs
```diff
@@ -217,397 +215,588 @@ impl ServerImpl {
...
     }
 
-    fn get_status(
+    /// This should be called as the last operation before sending the response
+    /// since error details is the last of the trailing data. This function
+    ///
+    /// Panics
+    ///
+    /// This function panics if `None` is the supplied `failure_type` and the
+    /// `result` contains any  bits set in the failure mask as the programmer
```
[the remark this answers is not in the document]
Aaron-Hartwig: >but messing that up would produce a runtime error

The question would then become "how does one recover from this error". @bnaecker and I discussed this and could not arrive at an elegant way to handle this. I'm happy to sync on this in a VC as well so we can iterate faster.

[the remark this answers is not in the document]
Aaron-Hartwig: We synced briefly on this and I had misunderstood Matt. I'm straightened out now and am going to explore splitting this into two types and two handle error functions.

drv/transceivers-server/src/udp.rs
Aaron-Hartwig: I'm going to opt to keep both the `HostResponse` and `Error` branches the way they are right now (log to ringbuf, do not respond). Our current philosophy has been "a request gets a response", so responding to a response-class of message would break that. Additionally, responding to an error could cause additional errors, causing us to spiral.

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -550,6 +552,78 @@ impl ModuleResult {
...
+    /// masks which then need to be reconciled with a new set of masks, generally
+    /// a subset of the success mask of the initial result. Notably, there
+    /// cannot be overlap between these masks, which this function enforces.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if the `next.success` mask is not a subset of
+    /// self.success. Additionally, it will panic if any of the success,
+    /// failure, or error masks overlap with one another.
+    pub fn chain_with_nofail(&self, next: ModuleResultNoFailure) -> Self {
```
mkeeter, line 575: If you changed the definition of `chain` to be

```rust
pub fn chain<R: Into<ModuleResult>>(&self, next: R) -> Self {
    let next: ModuleResult = next.into();
    // ...
}
```
and implemented the trait
```rust
impl Into<ModuleResult> for ModuleResultNoFailure {
    // ...
}
```

you wouldn't need a second function here!

drv/transceivers-server/src/udp.rs
```diff
@@ -529,48 +530,34 @@ impl ServerImpl {
...
-    /// This function panics if `None` is the supplied `failure_type` and the
-    /// `result` contains any  bits set in the failure mask as the programmer
-    /// must supply a `HwError` when calling this function on a `result` which
-    /// may have a non-zero failure mask.
-    fn handle_errors(
+    /// This function reads a `ModuleResult` and populates and failure or error
+    /// information at the end of the trailing data buffer. This means it should
+    /// be called as the last operation before sending the response. For results
+    /// where a `ModuleResultNoFailure` is returned, use handle_errors instead.
+    fn handle_errors_and_failures(
```
mkeeter, line 537: Optional API suggestion: this function (and `handle_errors`) could be a little cleaner if they accepted a pre-sliced `out`, instead of also taking `data_idx`.  This would make it the caller's responsibility to slice the buffer, but the most common caller is passing in `0`, which wouldn't require any slicing!

The only place you'd have to be careful is in cases like
```rust
let (final_payload_len, errored_modules) =
    self.handle_errors(modules, result, num_status_bytes, out);
```
which would become
```rust
let (err_len, errored_modules) =
    self.handle_errors(modules, result, &mut out[num_status_bytes..]);
let final_payload_len = num_status_bytes + err_len;
```

Aaron-Hartwig: It is definitely cleaner and less complicated. Done.

drv/transceivers-server/src/udp.rs
```diff
@@ -297,8 +297,11 @@ impl ServerImpl {
                 let (num_status_bytes, result) = self.get_status(mask, out);
                 ringbuf_entry!(Trace::OperationNoFailResult(result));
                 let success = ModuleId::from(result.success());
-                let (final_payload_len, errored_modules) =
-                    self.handle_errors(modules, result, num_status_bytes, out);
+                let (final_payload_len, errored_modules) = self.handle_errors(
```
mkeeter, line 300: I think this is now wrong – `handle_errors` returns the number of error bytes, so you need to add that to `num_status_bytes` to get `final_payload_len`

drv/transceivers-server/src/udp.rs
```diff
@@ -336,8 +339,7 @@ impl ServerImpl {
                         modules,
                         result,
                         HwError::I2cError,
-                        read_bytes,
-                        out,
+                        &mut out[read_bytes..],
```
mkeeter, line 342: Same here, `final_payload_len` is not correct.

drv/transceivers-server/src/udp.rs
```diff
@@ -334,13 +335,14 @@ impl ServerImpl {
...
                 let read_bytes = result.success().count() * read.len() as usize;
-                let (final_payload_len, failed_modules) = self
+                let (err_len, failed_modules) = self
                     .handle_errors_and_failures(
                         modules,
                         result,
                         HwError::I2cError,
                         &mut out[read_bytes..],
                     );
+                let final_payload_len = num_status_bytes + err_len;
```
mkeeter: Okay, now this one is wrong – I think it should be `read_bytes + err_len`

Aaron-Hartwig: notice that locally and fixed it in a follow up

Cargo.toml
```diff
@@ -133,5 +133,5 @@ sprockets-common = { git = "https://github.com/oxidecomputer/sprockets.git", def
 sprockets-rot = { git = "https://github.com/oxidecomputer/sprockets.git", default-features = false }
 tlvc = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
 tlvc-text = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
-transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false }
+transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false, branch = "the-big-hammer" }
```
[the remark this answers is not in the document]
bnaecker: And of course I forgot this. Pushing a fix now.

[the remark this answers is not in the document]
bnaecker: https://github.com/oxidecomputer/hubris/pull/1276

Issue #599, opened 2022-06-10
Improving `spd`'s worst case latency


cbiffle:
Currently we have a priority inversion in the `gimlet` images between `spd` (the DIMM SPD proxy task) and `i2c_driver`. This is because `spd` issues i2c transactions to read information from DIMMs, but also has strict latency requirements as a result of implementing an SMBus device, due to how the ST I2C controller works. We moved `spd` above the `i2c_driver` task because, since `i2c_driver` is monolithic (serves all controllers one at a time), putting it higher priority than `spd` meant that _any i2c traffic anywhere on the system would block `spd` from running,_ making it very, very easy to miss its deadlines.

The priority inversion creates a very real availability bug:

1. `spd` starts a blocking call to `i2c_driver`.
2. Any task with priority between `i2c_driver` and `spd` starts running
3. `spd` is now out to lunch for an arbitrarily long period of time despite being higher priority

The system does not _currently_ enforce the uphill send rule that prevents these priority inversions, which is why you can build `gimlet`. But we intend to start, to avoid bugs like the one I just described.

So, let's fix `spd`.

I worked through a number of options on paper, and the one that seems like the best compromise between "reasonably low amount of work" and "not a total hack that will prevent me from sleeping" can be summarized as:

1. Poll the DIMM thermal sensors from the `thermal` task (lower priority). Post the results to the `sensors` task (high priority) like all other thermal sensors.
2. Read the SPD EEPROM contents from the `gimlet_seq` task (lower priority), because it knows exactly when the window between "DIMM SPD power available" (A2) and "potentially racing the host" (A0) is, since it's responsible for sequencing it.
3. Deliver the SPD EEPROM contents into the `spd` task via IPC.
4. Alter `spd` to serve cached EEPROM contents (pre-prepared by `gimlet_seq` before leaving A2), plus live temperature data taken from the `sensors` task. This means `spd` never makes a blocking call to _any task_ except `sensors`, which is high priority and responds promptly.

The relative priority of the affected tasks, and their IPC relationships, are shown in this diagram:
![PXL_20220610_175113553](https://user-images.githubusercontent.com/45247/173127300-e792be4f-8ae9-466b-954d-7b9ffd9239d6.PNG)

In descending order from highest to lowest priority we have

- `sensors`
- `spd`
- `i2c`
[...]

PR #1242, opened 2023-03-28
Fix SPD / i2c_driver priority inversion

### LINKED ISSUE #562: `spd` priority inversion with `i2c_driver`
The `spd` task, on startup, needs to pull a bunch of data from `i2c_driver` to cache as an SPD proxy.  However, it also needs to run at a very high priority, to quickly react when the host asks for SPD data.  This means that the startup calls are uphill (to a task of a lower priority).

The fix is probably to put the initial data loading into `gimlet_seq`, which would then spoon-feed the data to `spd`.  However, things become tricky when one or the other restarts.

jgallagher:
Prior to this PR:

gimlet-c-before-fix

After this PR:

gimlet-c-after-fix

I.e., `packrat` moved from priority 3 to priority 1, and `spd` now calls into it instead of `i2c_driver`.

Reading the SPD is now the responsibility of `gimlet-seq`, which it does immediately after transitioning to A2 (after a 10ms delay, new in this PR but which we should have had all along per the spec). `gimlet-seq` populates packrat, and `spd` reads data from packrat when requested to over i2c.

Bonus change: all values held by `packrat` are now stored in statics, allowing them to be easily read via `humility readvar`.

Fixes #562. Helps some of #599. Supersedes (and is heavily based on) #612.


bcantrill (approved):
Looks great -- and I have verified that both Gimlet Rev B and Gimlet Rev C boot and see all DIMMs, and that `humility spd` also continues to work. Ship it!

PR #1244, opened 2023-03-28
LPC55 raw flash driver and update support.

cbiffle:
Previously we've been calling through the LPC55's ROM to write/erase flash. In build configurations where Trustzone was enabled, this meant indirecting through the hypovisor into the secure-world to perform the flash writes.

This mostly worked fine but had some structural issues:

- The actual code to interact with flash was spread over four crates and a closed-source ROM, so answering basic questions like "does this function called write actually erase?" was surprisingly involved.

- Adding new entry points to flash (e.g. for erase-check or verify) required modifying ~4 crates.

- Getting the ROM working from within a Hubris task (for the non-hypovisor case) consumed all of the task's MPU slots to map: SYSCON, the secure alias of SYSCON, the flash controller, the ROM itself, program RAM, program flash, and handoff memory in USBSRAM. This meant adding a new mapping for something like caboose of CFPA access was impossible.

I got here during the Verified Boot Debacle because of my need to rewrite CFPA in-system, map the transient boot select RAM into the task, and finish the "Read Caboose" TODO -- all of which were not possible given the jammed MPU configuration.

However, I also happened to have written most of a direct ROM-free driver for the LPC55 flash controller during my RFD374 stage0 work. So, I ported it over, and have rewritten the lpc55-update-server to use it.

You're probably wondering: this seems to strip out all hypocall code, how can this work on Trustzone configurations? Well, I have sort-of-good news for you, in two parts:

1. We only have one trustzone configuration, which is the LPC55xpresso board, because the attempt to add trustzone to the Gimlet C rev was silently undone by a typo. (Our configuration system could do a better job noticing this, but it's not immediately obvious how to pull it off.)

2. Even in trustzone, we never actually limited nonsecure access to the flash controller -- defeating the point of trustzone, but also enabling this driver to work without further changes.

Besides freeing up MPU regions and cutting the ROM out of the picture, which I consider to be good things, there are a couple of unintended benefits of this change:

1. Driving the flash controller directly requires less code than calling the ROM (seriously), and shrinks the update server by 300 bytes.

2. Flash writes and erases are now interrupt-driven. The ROM appears to have used busy-waits the way we were calling it (since it had no ability to configure or respond to interrupts, running in unprivileged mode).

I've also renamed the peripheral in chips.toml from "flash" to "flash_controller," since program text is also called "flash" and I felt like it was potentially confusing.

### DIFF
--- a/drv/lpc55-flash/src/lib.rs
+++ b/drv/lpc55-flash/src/lib.rs
@@ -1,0 +1,389 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! A raw driver for the LPC55 flash controller without using the ROM.
+//!
+//! This driver is written in a very generic form that doesn't assume any
+//! particular execution model. You can use it to implement busy waiting, or
+//! an interrupt driven driver, etc.
+//!
+//! See the [`Flash`] type for more details.
+
+#![no_std]
+
+use core::ops::RangeInclusive;
+
+/// Number of bytes per flash word.
+pub const BYTES_PER_FLASH_WORD: usize = 16;
+
+/// Number of bytes per flash page.
+pub const BYTES_PER_FLASH_PAGE: usize = 512;
+
+/// Flash driver handle. Wraps a pointer to the flash register set and provides
+/// encapsulated operations.
+///
+/// # Procedure for writing to flash
+///
+/// 1. Use `start_erase_region` to begin an erase of your desired region. Use
+///    `poll_erase_or_program_result` to discover when it's done. (Note: you can
+///    use `start_blank_check` to skip this step if there's a chance the flash
+///    is already erased.)
+/// 2. Call `start_write_row` with each 16 byte chunk of the first page you're
+///    writing, 32 times, for rows 0-31. Each time, use `poll_write_result` to
+///    check that it's done.
+/// 3. Call `start_program` with the index of any word in the target flash page
+///    to begin the programming process. (See the next section on word numbers
+///    for explanation.) Use `poll_erase_or_program_result` to learn when it's
+///    done.
+/// 4. Repeat steps 2 and 3 for each page in the region you want to write.
+///
+/// # Flash addressing and word numbers
+///
+/// While the flash is directly addressable by the processor at byte
+/// granularity, most of the flash controller operations work in terms of a
+/// _word number._ Internal flash words are 16 bytes (128 bits) in length; the
+/// word number is the index of a word within the flash, starting at 0.
+///
+/// The flash controller thinks entirely in word numbers, _including on
+/// operations that deal in pages._ For example, to erase a range of pages, you
+/// give the hardware a range of word numbers, _not_ page numbers, and it erases
+/// the sequence of contiguous pages that contain the start and end word
+/// numbers. In the interest of not introducing additional, confusable, types of
+/// addresses, this API preserves this and deals only in word numbers.
+///
+/// Should you need to convert a direct-addressed flash location (byte address)
+/// to a word number, the procedure is:
+///
+/// - Divide by 16 (or right-shift by 4).
+/// - Mask all but the bottom 18 bits.
+///
+/// Note: The `start_write_row` operation does not use _absolute_ flash word
+/// numbers, but uses equivalent indices within a 512-byte internal page
+/// register.
+pub struct Flash<'a> {
+    reg: &'a lpc55_pac::flash::RegisterBlock,
+}
+
+impl<'a> Flash<'a> {
+    /// Wraps a pointer to the flash controller registers in a `Flash` driver
+    /// instance.
+    pub fn new(reg: &'a lpc55_pac::flash::RegisterBlock) -> Self {
+        Self { reg }
+    }
+
+    /// Copies 16 bytes of data into one of the rows of the flash controller's
+    /// internal page register.
+    ///
+    /// The page register is 512 bytes long and consists of 32 rows of 16 bytes
+    /// each. Data must be accumulated into the page register before being
+    /// programmed.
+    ///
+    /// The `row_in_page` passed to this routine specifies one of the 32 rows.
+    /// This means it operates a lot like a flash word number, but in the
+    /// separate "address space" of the page register.
+    ///
+    /// This routine only starts the operation; use `poll_write_result` to learn
+    /// when it has completed. Since writes are just latching values into
+    /// registers, they tend to be very fast, and it may not be useful to use
+    /// interrupts to wait for completion here.
+    pub fn start_write_row(&mut self, row_in_page: u32, values: &[u8; 16]) {
+        self.clear_status_flags();
+        self.set_single_word_number(row_in_page);
+        for (register, chunk) in self.reg.dataw.iter().zip(values.chunks(4)) {
+            let value = u32::from_ne_bytes(chunk.try_into().unwrap());
+            register.write(|w| unsafe { w.bits(value) });
+        }
+        self.issue_cmd(FlashCmd::Write);
+    }
+
+    /// Checks the completion status for the values expected after a
+    /// `start_write_row` operation. `true` means the write has finished.
+    /// `false` means it has not.
+    pub fn poll_write_result(&mut self) -> bool {
+        self.read_completion_status().is_some()
+    }
+
+    /// Begins an erase of all flash words in `word_range`.
+    ///
+    /// The start and end values of `word_range` are _word numbers._ See the
+    /// discussion of word numbers on the [`Flash`] type.
+    ///
+    /// This routine only starts the operation; use
+    /// `poll_erase_or_program_result` to learn when it has completed and
+    /// whether it succeeded.
+    pub fn start_erase_range(&mut self, word_range: RangeInclusive<u32>) {
+        self.clear_status_flags();
+        self.set_word_range(word_range);
+        self.issue_cmd(FlashCmd::EraseRange);
+    }
+
+    /// Begins programming the data loaded in the page register (by previous
+    /// write operations) into the flash page containing `word_number`.
+    ///
+    /// See the discussion of word numbers on the [`Flash`] type.
+    ///
+    /// This routine only starts the operation; use
+    /// `poll_erase_or_program_result` to learn when it has completed and
+    /// whether it succeeded.
+    pub fn start_program(&mut self, word_number: u32) {
+        self.clear_status_flags();
+        self.set_single_word_number(word_number);
+        self.issue_cmd(FlashCmd::Program);
+    }
+
+    /// Checks the completion status for the values expected after a
+    /// `start_erase_range` or `start_program` operation.
+    ///
+    /// `None` means the operation has not yet completed.
+    ///
+    /// `Some(Ok(())` means the operation completed without issue.
+    ///
+    /// `Some(Err(FlashTimeout))` means the operation did not complete because
+    /// the internal flash state machine didn't finish within the time allotted.
+    /// This is the only way these operations are able to fail according to the
+    /// user manual. It's not at all clear what to _do_ in this situation since
+    /// the internal state machine is essentially opaque -- but you can assume
+    /// that your program or erase did not succeed.
+    pub fn poll_erase_or_program_result(
+        &mut self,
+    ) -> Option<Result<(), FlashTimeout>> {
+        let s = self.read_completion_status()?;
+        // Erase and program operations only define the behavior of the FAIL
+        // bit, which indicates that the flash internal state machine didn't
+        // finish within a generous timeout.
+        Some(if s.fail { Err(FlashTimeout) } else { Ok(()) })
+    }
+
+    /// Begins a blank-check operation on the range of flash pages containing
+    /// the start and end of `word_range`, inclusive.
+    ///
+    /// `word_range` is given in terms of _word numbers._ See the discussion of
+    /// word numbers on the [`Flash`] type.
+    ///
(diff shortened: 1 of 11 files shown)


drv/lpc55-update-server/src/main.rs
```diff
@@ -279,31 +277,133 @@ fn validate_header_block(
...
+
+        flash.enable_interrupt_sources();
+        sys_irq_control(notifications::FLASH_IRQ_MASK, true);
+        // RECV from the kernel cannot produce an error, so ignore it.
+        let _ = sys_recv_closed(&mut [], notifications::FLASH_IRQ_MASK, TaskId::KERNEL);
+        flash.disable_interrupt_sources();
+    }
+}
+
+fn same_image(which: UpdateTarget) -> bool {
```
cbiffle, line 351: Note to the reader: the code from this point down is stuff I've factored out of `lib/hypocalls` and `lib/lpc55-flash` so that it's all in one place, can use the same types without conversions, etc. I have added light comments to things that bit me, but otherwise the lack of documentation here is from the original.

drv/lpc55-flash/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+/// - Divide by 16 (or right-shift by 4).
+/// - Mask all but the bottom 18 bits.
+///
+/// The `start_write_row` operation does not use flash word numbers, but uses
+/// equivalent indices within a 512-byte internal page register.
+pub struct Flash<'a> {
+    reg: &'a lpc55_pac::flash::RegisterBlock,
+}
+
+impl<'a> Flash<'a> {
```
mkeeter, line 54: Nit: we could probably use the placeholder lifetime `'_` here.

drv/lpc55-flash/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+            w.fail().set_bit();
+            w
+        });
+    }
+}
+
+/// Raw status bits from the flash controller.
+///
+/// This type is exposed in case you want to do something weird with the raw
+/// flash controller operations; the higher-level APIs do not use it directly.
```
mkeeter, line 328: Add a note here to the effect that a `Status` is only constructed when the `done` bit is set, so that bit isn't included?

drv/lpc55-update-server/src/main.rs
```diff
@@ -246,15 +239,20 @@ fn validate_header_block(
...
         block[RESET_VECTOR_OFFSET..][..4].try_into().unwrap_lite(),
-    );
-    let a_base = unsafe { &__IMAGE_A_BASE } as *const u32 as u32;
-    let b_base = unsafe { &__IMAGE_B_BASE } as *const u32 as u32;
-    let stage0_base = unsafe { &__IMAGE_STAGE0_BASE } as *const u32 as u32;
-    let a_end = unsafe { &__IMAGE_A_END } as *const u32 as u32;
-    let b_end = unsafe { &__IMAGE_B_END } as *const u32 as u32;
-    let stage0_end = unsafe { &__IMAGE_STAGE0_END } as *const u32 as u32;
+    ) & ADDRMASK;
+    let a_base = unsafe { &__IMAGE_A_BASE } as *const u32 as u32 & ADDRMASK;
```
mkeeter, line 250: Nit: I find it a bit cleaner to write this as
```suggestion
    let a_base = unsafe { __IMAGE_A_BASE.as_ptr() } as u32 & ADDRMASK;
```

drv/lpc55-update-server/src/main.rs
```diff
@@ -279,31 +277,133 @@ fn validate_header_block(
...
+    // write_addr is a byte address; page_num is a page index; but the actual
+    // erase machinery operates in terms of flash words. A flash word is 16
+    // bytes in length. So, we need to convert.
+    //
+    // Note that the range used here is _inclusive._
+    //
+    // We wind up needing this in u32 form a lot, so let's cast once and reuse.
+    const WORDSZ: u32 = BYTES_PER_FLASH_WORD as u32;
+    flash.start_erase_range(write_addr / WORDSZ ..= (write_addr + BLOCK_SIZE_BYTES as u32 - 1) / WORDSZ);
+    wait_for_erase_or_program(&mut flash);
```
mkeeter, line 313: The compiler warns that the result here is unused

drv/lpc55-update-server/src/main.rs
```diff
@@ -279,31 +277,133 @@ fn validate_header_block(
...
+        flash.start_write_row(i as u32, row);
+        while !flash.poll_write_result() {
+            // spin - supposed to be very quick in hardware.
+        }
+    }
+
+    // Now, program the whole page into non-volatile storage. This again uses
+    // page indices, requiring a divide-by-16.
+    flash.start_program(write_addr / WORDSZ);
+    wait_for_erase_or_program(&mut flash);
```
mkeeter, line 330: Same here!

drv/lpc55-update-server/src/main.rs
```diff
@@ -279,31 +277,133 @@ fn validate_header_block(
...
+    // Step one: erase the page. 
+
+    // write_addr is a byte address; page_num is a page index; but the actual
+    // erase machinery operates in terms of flash words. A flash word is 16
+    // bytes in length. So, we need to convert.
+    //
+    // Note that the range used here is _inclusive._
+    //
+    // We wind up needing this in u32 form a lot, so let's cast once and reuse.
+    const WORDSZ: u32 = BYTES_PER_FLASH_WORD as u32;
```
mkeeter, line 311: In the interest of paranoia, it may be worthwhile to have a
```rust
static_assertions::const_assert_eq!(BLOCK_SIZE_BYTES % BYTES_PER_FLASH_WORD, 0)
```
somewhere near here.

drv/lpc55-update-server/src/main.rs
```diff
@@ -279,31 +277,133 @@ fn validate_header_block(
...
+        flash.enable_interrupt_sources();
+        sys_irq_control(notifications::FLASH_IRQ_MASK, true);
+        // RECV from the kernel cannot produce an error, so ignore it.
+        let _ = sys_recv_closed(&mut [], notifications::FLASH_IRQ_MASK, TaskId::KERNEL);
+        flash.disable_interrupt_sources();
+    }
+}
+
+fn same_image(which: UpdateTarget) -> bool {
+    get_base(which) == unsafe { __this_image.as_ptr() as u32 }
```
mkeeter, line 352: Nit: I think the `as u32` could be hoisted out of the `unsafe` block here.

drv/lpc55-update-server/src/main.rs
```diff
@@ -279,31 +277,133 @@ fn validate_header_block(
...
+    match which {
+        UpdateTarget::ImageA => unsafe { __IMAGE_A_BASE.as_ptr() as u32 },
+        UpdateTarget::ImageB => unsafe { __IMAGE_B_BASE.as_ptr() as u32 },
+        UpdateTarget::Bootloader => unsafe { __IMAGE_STAGE0_BASE.as_ptr() as u32 },
+        _ => unreachable!(),
     }
 }
 
+fn get_end(which: UpdateTarget) -> u32 {
+    match which {
```
mkeeter, line 367: Same here!

drv/lpc55-update-server/src/main.rs
```diff
@@ -279,31 +277,133 @@ fn validate_header_block(
...
+        .checked_mul(BLOCK_SIZE_BYTES as u32)?
+        .checked_add(base)?;
+
+    // check addr + PAGE_SIZE < end
+    if addr
+        .checked_add(BLOCK_SIZE_BYTES as u32)?
+        > get_end(image_target)
+    {
+        return None;
+    }
```
mkeeter, line 402: The mismatch between the comment and code here is confusing, and I suspect the code is wrong.

`get_end` gives the end of an **exclusive** range, so I think we want

```suggestion
    // check addr + PAGE_SIZE < end
    if addr
        .checked_add(BLOCK_SIZE_BYTES as u32)?
        >= get_end(image_target)
    {
        return None;
    }
```

drv/lpc55-update-server/src/main.rs
```diff
@@ -279,31 +277,133 @@ fn validate_header_block(
...
 ) -> Result<(), UpdateError> {
-    let result = unsafe {
-        // The write_to_flash API takes raw pointers due to TrustZone
-        // ABI requirements which makes this function unsafe.
-        tz_table!().write_to_flash(
-            img,
-            block_num as u32,
-            flash_page.as_mut_ptr(),
-        )
+    let page_num = block_num as u32; // TODO why not use the same type
```
andrewjstone, line 287: I'm totally fine changing the type in the function parameter, although it would ripple upward at least a bit.

drv/lpc55-flash/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+/// # Procedure for writing to flash
+///
+/// 1. Use `start_erase_region` to begin an erase of your desired region. Use
+///    `poll_erase_or_program_result` to discover when it's done. (Note: you can
+///    use `start_blank_check` to skip this step if there's a chance the flash
+///    is already erased.)
+/// 2. Call `start_write_row` with each 16 byte chunk of the first page you're
+///    writing, 32 times, for rows 0-31. Each time, use `poll_write_result` to
+///    check that it's done.
+/// 3. Call `start_program` with the index of any word in the target flash page
```
andrewjstone, line 31: This is kind of wacky to me. That you only have to pick one of the words to perform the programming, and it doesn't matter which one.

andrewjstone (approved):
Cliff this looks great. Approved conditionally on the response to Matt's `get_end` comment.

It looks simple enough, I wonder why we ever used the ROM! 😃

drv/lpc55-flash/src/lib.rs
cbiffle: Yeah, this is me exposing how the hardware works, but it might not be the best idea.

Erase and programming (and blank check) all affect pages, but take word indices as their inputs. They ignore the bottom bits. So if you ask to erase word W you're actually going to erase the page containing word W. The API currently exposes this.

I could have page operations take page indices instead. I didn't do this in the first pass because
1. My default is to expose the hardware's behavior so that you can compare it to the datasheet, and
2. This would be the first instance of a page index in the API, which is currently all word index based. Having two different types of addresses is _also_ potentially error prone.

But that could be wrong! What do you think? Also @mkeeter

drv/lpc55-flash/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+///
+/// While the flash is directly addressable by the processor at byte
+/// granularity, most of the flash controller operations work in terms of a
+/// _word number._ Internal flash words are 16 bytes (128 bits) in length; the
+/// word number is the index of a word within the flash, starting at 0.
+///
+/// Converting an address to a word number:
+///
+/// - Divide by 16 (or right-shift by 4).
+/// - Mask all but the bottom 18 bits.
```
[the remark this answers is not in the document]
cbiffle: If you're taking an address that points into the direct-mapped flash space, sure! Access to the Flash secure alias is the easiest case.

The shift-and-mask in these two lines will basically hide any data above the 22nd bit (because it's right-shifted by 4 and then masked). That's `0x40_0000` and up. Accessing the secure alias will additionally set bit 28, but that bit doesn't matter from the perspective of the flash controller.

drv/lpc55-flash/src/lib.rs
cbiffle: Good point, done!

drv/lpc55-update-server/src/main.rs
cbiffle: Yeah, so do I -- this is copy-pasted code. I will clean this up.

cbiffle: (In the original these fields were all `: ()` so `.as_ptr()` wasn't available, I changed them to zero-length arrays)

drv/lpc55-update-server/src/main.rs
cbiffle: Ah, yeah, I meant to propagate this before pushing the PR -- the TODO was for me. Lemme take a pass at this and see how it looks.

drv/lpc55-update-server/src/main.rs
cbiffle: I spent some time staring at this after I simplified the code (the original, if you're curious, is `lib/lpc55-flash/src/lib.rs:119`). I believe it is the _comment_ that is wrong. If changed to `>=`, this would prevent you from talking about the last page.

I will fix the comment.

drv/lpc55-flash/src/lib.rs
mkeeter: Okay, seems fine!

drv/lpc55-flash/src/lib.rs
mkeeter: I'm okay with the current behavior, because I agree that matching the datasheet is valuable.

cbiffle: Yeah, synced with @andrewjstone in chat and he's on the same page. I have added comments clarifying the reason for the behavior.

drv/lpc55-update-server/src/main.rs
cbiffle: So it turns out the IDL interface uses `usize` here, and changing this would be more work. Will instead remove the TODO.

cbiffle:
> It looks simple enough, I wonder why we ever used the ROM! smiley

We assumed, based on conversations with NXP, that the ROM would be easier and faster! It turns out that's a lot less true if you're using memory protection and unprivileged mode, where getting the ROM to work takes _effort._

Plus there's the impedance mismatch between the mostly-safe Rust code and the wildly-unsafe ROM C API that required some bookkeeping.

PR #1246, opened 2023-03-29
Boot preference writing on LPC55 for RFD374 stage0

cbiffle:
This is a backport of my code for updating RFD374 stage0's boot preference in the CFPA. In its backported form, it works, in that it can rewrite CFPA in-system on unlocked _or locked_ LPC55s, and will update the boot preference bit according to your request. However, currently RoTs built from master aren't actually _using_ the RFD374 stage0, which means the preference bit has no actual effect.

You can verify its setting using `humility readmem`:
- A call to `switch_default_image` will put the preference in the word at `0x9df00`
- Resetting will transfer it to either the word at `0x9e100` or `0x9e200` depending on the monotonic CFPA versions.

This required porting stage0's sha256 driver into Hubris, which I've included in this PR.

An example of operating the API:

```
humility -a target/rot-carrier-secure/dist/b/build-rot-carrier-secure.zip hiffy -c Update.switch_default_image -a slot=A -a duration=Forever
```


drv/lpc55-update-server/src/main.rs
```diff
@@ -226,6 +234,203 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
+                    cfpa[0][1].checked_add(1).ok_or(UpdateError::SecureErr)?;
+                cfpa[0][1] = new_version;
+                // Alter the boot setting. The boot setting (per RFD 374) is in
+                // the lowest bit of the 32-bit word starting at (byte) offset
+                // 0x100. This is flash word offset 0x10.
+                //
+                // Leave remaining bits undisturbed; they are currently
+                // reserved.
+                cfpa[0x10][0] &= !1;
+                cfpa[0x10][0] |= if slot == SlotId::A { 0 } else { 1 };
```
lzrd, line 321: Just thinking this through. Looks good.

Physical access is out of scope for the threat model including glitching attacks, so using a single bit for bank selection is likely not a problem.

Unlike the transient boot preference selection (Once) in RAM, the CFPA-based persistent selection (Forever) is ok to be a single bit because there is ECC protection for CFPA storage (true?) and/or the trailing SHA2-256 hash.

There are three states, a bank has never been selected, A selected, B selected.
The never-been-selected state is only for a short while, if ever, while being manufactured.
Additionally, this is just a preference, if the preferred bank is not suitable for boot, no attempt
will be made to run it.

drv/lpc55-update-server/src/main.rs
```diff
@@ -226,6 +234,203 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
+                    Err(ReadError::Ecc) => {
+                        return Err(UpdateError::EccDoubleErr);
+                    }
+                    Err(ReadError::Fail) => {
+                        return Err(UpdateError::Unknown);
+                    }
+                }
+            }
+
+            // But just in case it needs to:
```
lzrd, line 419: You're not clearing the HW interrupt condition on error returns.

drv/lpc55-sha256/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+        }
+    }
+
+    /// Completes the SHA256 hash.
+    ///
+    /// You should seriously consider flipping the HASHCRYPT unit back into
+    /// reset after calling this, to avoid leaking data about
+    /// whatever-it-was-you-were-just-doing. Or don't; we're not the cops.
+    #[inline(never)]
+    pub fn finish(mut self) -> [u32; 8] {
```
mkeeter, line 197: Would it be correct to use `WORDS_PER_HASH` here?  It has the right value, but not sure if it's got the correct _meaning_.
```suggestion
    pub fn finish(mut self) -> [u32; WORDS_PER_HASH] {
```

drv/lpc55-sha256/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+                self.engine.intenclr.write(|w| w.digest().set_bit());
+            }
+        }
+
+        // The result arrives in registers called digest0..digest7, which the
+        // PAC calls digest0[0] .. digest0[7] for some reason.
+        let mut result = [0; WORDS_PER_HASH];
+        for (i, dest) in result.iter_mut().enumerate() {
+            *dest = self.engine.digest0[i].read().bits().swap_bytes();
+        }
```
mkeeter, line 270: Optional: you could remove explicit indexing with
```suggestion
        for (dest, src) in result.iter_mut().zip(&self.engine.digest0) {
            *dest = src.read().bits().swap_bytes();
        }
```

drv/lpc55-update-server/src/main.rs
```diff
@@ -152,7 +159,7 @@ impl idl::InOrderUpdateImpl for ServerImpl {
             flash_page[len..].fill(0);
         }
 
-        do_block_write(target, block_num, &mut flash_page)?;
+        do_block_write(&mut self.flash, target, block_num, &mut flash_page)?;
```
mkeeter, line 162:
```suggestion
        do_block_write(&mut self.flash, target, block_num, &flash_page)?;
```

drv/lpc55-update-server/src/main.rs
```diff
@@ -176,6 +183,7 @@ impl idl::InOrderUpdateImpl for ServerImpl {
         }
 
         do_block_write(
+            &mut self.flash,
```
mkeeter, line 186: I can't put this comment at the correct line, but the final argument to this function could be
```rust
            self.header_block.as_ref().unwrap_lite(),
```
(no need for `as_mut`)

drv/lpc55-update-server/src/main.rs
```diff
@@ -226,6 +234,203 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
+                let cfpa_word_number = {
+                    // Read the two versions. We do this with smaller buffers so
+                    // we don't need 2x 512B buffers to read the entire CFPAs.
+                    let mut ping_header = [0u32; 4];
+                    let mut pong_header = [0u32; 4];
+
+                    indirect_flash_read(
+                        &mut self.flash,
+                        0x9E00,
+                        core::slice::from_mut(&mut ping_header),
```
mkeeter, line 282: Neat!

drv/lpc55-sha256/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+///
+/// The `key` is fixed at 256 bits, and both `key` and `data` must consist of a
+/// whole number of 32-bit words. See the module docs for more information about
+/// these limitations.
+///
+/// HMAC is specified in RFC 2104.
+///
+/// `notification_mask` is the notification mask corresponding to the HASHCRYPT
+/// IRQ in your task. If you'd rather busywait, pass 0.
+#[inline(never)]
```
mkeeter, line 78: Should this docstring have the same warning about resetting the hash unit, i.e.
```rust
    /// You are _strongly advised_ to have reset the HASHCRYPT unit just before
    /// this, lest it contain nonsense left over from a previous hash. (In
    /// particular, the boot ROM leaves nonsense in it, so it needs to be reset
    /// at least once after startup.)
```
(or equivalent)?

drv/lpc55-update-server/src/main.rs
```diff
@@ -226,6 +234,203 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
+                // Leave remaining bits undisturbed; they are currently
+                // reserved.
+                cfpa[0x10][0] &= !1;
+                cfpa[0x10][0] |= if slot == SlotId::A { 0 } else { 1 };
+                // The last two flash words are a SHA256 hash of the preceding
+                // data. This means we need to compute a SHA256 hash of the
+                // preceding data -- meaning flash words 0 thru 29 inclusive.
+                let cfpa_hash = {
+                    self.syscon
+                        .leave_reset(drv_lpc55_syscon_api::Peripheral::HashAes);
```
mkeeter, line 327: Maybe add a comment here indicating that we call `enter_reset` during `lpc55-update-server`'s startup.

drv/lpc55-update-server/src/main.rs
```diff
@@ -301,29 +504,43 @@ fn do_block_write(
...
+}
+
+/// Performs an erase-write sequence to a single page within the raw flash
+/// device. This function is capable of writing outside of any image slot, which
+/// is important for doing CFPA updates. If you're writing to an image slot, use
+/// `do_block_write`.
+fn do_raw_page_write(
+    flash: &mut drv_lpc55_flash::Flash<'_>,
+    page_num: u32,
+    flash_page: &[u8; BLOCK_SIZE_BYTES],
```
mkeeter, line 521: Since this function operates on pages, it would be more clear to use `&[u8; BYTES_PER_FLASH_PAGE]` here even though it's identical; this would also save us if someone tries to have `BLOCK_SIZE_BYTES` diverge from the page size.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -521,6 +521,15 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
         Ok(chunk.len() as u32)
     }
+
+    fn switch_default_image(
+        &mut self,
+        _: &userlib::RecvMessage,
+        slot: SlotId,
+        duration: SwitchDuration,
+    ) -> Result<(), RequestError<UpdateError>> {
+        Err(UpdateError::Unknown)
```
mkeeter, line 531: Do we eventually want to implement this in terms of `swap_banks`?

drv/lpc55-update-server/src/main.rs
lzrd: In the failure cases, lines 408, 411, and 414, the interrupt condition has been polled.
The code starting at 421 shows a desire to clear the lingering HW interrupt condition that has
been left asserted from doing polled IO.
Without looking at the datasheet, it's implied that for this device, merely checking on interrupt status is not enough to clear the HW interrupt condition. Hence the enable and sys_recv_closed calls.

The Result should probably be taken from the match above, then HW interrupt clearing, then return Result.

drv/lpc55-update-server/src/main.rs
```diff
@@ -226,6 +234,203 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
+                //
+                // Page     Addr        16-byte word number
+                // Scratch  0x9_DE00    0x9DE0
+                // Ping     0x9_E000    0x9E00
+                // Pong     0x9_E200    0x9E20
+
+                let cfpa_word_number = {
+                    // Read the two versions. We do this with smaller buffers so
+                    // we don't need 2x 512B buffers to read the entire CFPAs.
+                    let mut ping_header = [0u32; 4];
```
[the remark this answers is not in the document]
andrewjstone: stupid, question, sorry. I should have read further down!

drv/lpc55-update-server/src/main.rs
lzrd: Is line 422 going to hang if there was no interrupt?

drv/lpc55-update-server/src/main.rs
cbiffle: *nod* That's basically my reasoning. I discussed this in an aside in [the data storage section of RFD374](https://rfd.shared.oxide.computer/rfd/374#_cfpa_data_storage_specifics) (sorry, non-Oxide viewers). The CFPA storage does indeed have _both_ ECC _and_ the trailing SHA256; the ECC is the actual protection, the SHA256 is mostly just a handshake with the ROM.

The main alternative I considered was to have a fancier bool that had possibly invalid states, so you could detect garbage (the "secure bool" pattern). But if we have a fancy bool that admits invalid states, we need to define a default behavior for what happens if the CFPA is found in an illegal state. There are three options as I see it:
1. Ignore it and boot slot A
2. Ignore it and boot slot B
3. Brick the device.

I don't like (3), so I rejected it. At that point, if the fallback position is going to be booting slot A or B ... it's a lot like having a single bit, which is unambiguous.

So in the end,
- There's a single bit that controls the preference
- It is ECC protected but can be overwritten at runtime (in a controlled fashion)
- Manufacturing default state is zeroed and reads as "prefer A," so there isn't a "not yet selected" state
- Once the device is locked, the ROM protects this sector from unexpected erasure, so "erased" isn't a state we need to consider.

drv/lpc55-update-server/src/main.rs
cbiffle: Can you expand on that? I think the code is okay...

- The status bits in the flash controller are cleared before any operation
- The interrupt enables are cleared on line 429
- There are no early-exit opportunities between when they're enabled on line 421 and cleared on line 429

drv/lpc55-sha256/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+        result
+    }
+
+    /// Utility factor for making sure we synchronize with the device at the
+    /// start of each 16-word input block.
+    #[inline(never)]
+    fn load_word(&mut self, word: u32, mask: u32) {
+        if self.word_count % WORDS_PER_BLOCK64 == Wrapping(0) {
+            // Wait for the controller to be interested in what we have to say.
+            while self.engine.status.read().waiting().is_not_waiting() {
```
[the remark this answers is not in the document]
cbiffle: The last bit of that ("until at least 1 word is read of the output") seems ... relevant. And yet, this code definitely works, in both busy-waiting (stage0) and interrupt-driven (Hubris) modes. My assumption is that the sentence must not mean what I think it means.

Do you think we should do something defensive here?

drv/lpc55-update-server/src/main.rs
cbiffle: Ha, thanks -- `do_block_write` used to take a `&mut`, must have missed this one. Changed.

drv/lpc55-update-server/src/main.rs
cbiffle: Changed!

drv/lpc55-update-server/src/main.rs
```diff
@@ -226,6 +234,203 @@ impl idl::InOrderUpdateImpl for ServerImpl {
...
+
+    fn switch_default_image(
+        &mut self,
+        _: &userlib::RecvMessage,
+        slot: SlotId,
+        duration: SwitchDuration,
+    ) -> Result<(), RequestError<UpdateError>> {
+        match duration {
+            SwitchDuration::Once => {
+                // TODO deposit command token into buffer
```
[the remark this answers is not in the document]
cbiffle: So, despite being an API that has parts that won't be implemented on some targets, there isn't a `NotImplemented` variant in `UpdateError`, and I sure hesitate to add one (as I consider "runtime not implemented error" an antipattern).

That said, you're absolutely right that we should return _some_ error here. I have used my other least favorite variant, `Unknown`.

drv/lpc55-update-server/src/main.rs
cbiffle: Stupid Reference Tricks(tm)

drv/lpc55-sha256/src/lib.rs
cbiffle: ......so, actually, because of the reset behavior, this hmac function is _incorrect_ in this context. (It needs to be able to reset the unit between the two constituent hashes, and doesn't know how.)

I could fix that by having the caller pass in a reset thunk, but I'll just remove it for now.

drv/lpc55-update-server/src/main.rs
cbiffle: Done!

drv/lpc55-update-server/src/main.rs
cbiffle: We're at the impedance boundary between my driver (which uses words from the datasheet) and the update code (which has invented the concept of blocks and applies them to systems that don't really have flash pages like the STM32H7). So, this is inconsistent, but it would be _real bad_ if they diverged.

I could go either way -- I copied this constant from `do_block_write`. Is it weirder to have the two differ or to use `BLOCK_SIZE_BYTES` here?

drv/stm32h7-update-server/src/main.rs
cbiffle: Probably? I'm not actually sure how bank swapping is handled right now.

drv/lpc55-update-server/src/main.rs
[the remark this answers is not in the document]
cbiffle: It's the size of a flash word, the smallest increment we can read. The flash API exposes this.

drv/lpc55-sha256/src/lib.rs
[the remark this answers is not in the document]
mkeeter: After looking at the NXP code in #oxide-hubris, we concluded that this is fine; it matches the NXP SDK, and seems to work fine.

drv/lpc55-update-server/src/main.rs
mkeeter: I think it's weirder to use `BLOCK_SIZES_BYTES`.

I'd vote to keep `do_block_write` in terms of `BLOCK_SIZE_BYTES` and switch `do_raw_page_write` to use `BYTES_PER_FLASH_PAGE`; because the `flash_page` argument is passed straight through, we should get an error if they somehow diverged.

mkeeter (approved):

drv/lpc55-update-server/src/main.rs
cbiffle: Okay, reading the comments I had missed:

The `poll` functions have no side effects. The interrupt status bits on the device are not cleared. They're only cleared on issuing a new operation, e.g. in this case `start_read`.

This is okay because for the status bits to turn into an actual interrupt, two things need to happen:
- The corresponding interrupt _enable_ bits in the device need to be set (which, when we're going to sleep, we do with the `enable_interrupt_sources` call.
- The hardware interrupt routing in the kernel needs to be enabled (`sys_irq_control`).

When these two conditions are met, an interrupt will be delivered when available and OR'd into the task's notification mask. The task will observe this on the next compatible call to `sys_recv`, which in our case we do immediately. However, those calls have no effect on the interrupt status or enable bits at the device. So, after `sys_recv_closed`, we call `disable_interrupt_sources` on the flash driver to clear the interrupt enables.

This may leave interrupt _status_ bits set (and in practice it absolutely will leave at least one bit set). This is okay, because they don't do anything until the corresponding interrupt source is enabled, and we're going to clear them before doing that next.

Re: your last question -- if the interrupt never comes, the `sys_recv_closed` call will indeed block forever waiting for it.

lzrd: (more careful reading...) Line 402 clears any pending flash controller status flags from the flash controller. The NVIC doesn't get an interrupt until/unless the flash controller DONE bit is set and line 426 and 427 are executed. So, this works out fine.
In the error cases, status bits are set but they are not resulting in NVIC interrupts. The error cases will leave those status bits set, but on the next operation (if the calling layer has not bailed out), those bits are cleared before the next flash I/O operation.

drv/lpc55-update-server/src/main.rs
cbiffle: Changing.

drv/lpc55-update-server/src/main.rs
lzrd: So, not a problem.

PR #1250, opened 2023-03-30
Attempt to make transceivers known on every loop iteration

mkeeter:
Otherwise, if we read a transceiver as `Management::Unknown` _once_, then we'll skip decoding its temperature data forever (causing the fans to spin up to 100%).


drv/transceivers-server/src/main.rs
```diff
@@ -230,8 +230,15 @@ impl ServerImpl {
...
-            if operational && self.thermal_models[i].is_none() {
+            // using SFF-8636 or CMIS.  We'll *also* try to read it here if
+            // we've currently got it recorded as ManagementInterface::Unknown,
+            // in the hopes of converting it into a known transceiver type.
+            let wants_scan = self.thermal_models[i]
+                .map(|m| {
+                    matches!(m.interface, ManagementInterface::Unknown(..))
+                })
+                .unwrap_or(true);
+            if operational && wants_scan {
```
Aaron-Hartwig, line 241: To check my understanding: We keep the model in an "initialized" state so that farther down (line 299 below) when we skip attempting to read a temp from an `Uknown` transceiver, the model won't think it has stale data?

Aaron-Hartwig (commented):
I think I'm okay with this as a band-aid, but we still need a design around `Unknown` modules and how we will handle them. Because of the thermal loop the SP will encounter the module before host software, so we should define how long we will allow an `Unknown` module to be powered on. Additionally, I don't think Dendrite should ever expose an `Uknown` module to the SDE.

cc @bnaecker

drv/transceivers-server/src/main.rs
mkeeter: Not quite – we want `self.thermal_models[i]` to be `Some(ManagementInterface::Unknown(...))` **because** we want the thermal loop to start checking that sensor, even if we don't have data for it yet.  This happens at L289, when we call `self.thermal_api.update_dynamic_input(i, m.model)`.

(If we left it as `None`, it would be skipped at L284 instead)

This choice relates to the architecture of the thermal loop: when a device is known to be present but hasn't yet reported any data, we have no way of knowing how hot it is, so we turn the fans on at full power (this is the `Boot` state in [RFD 276](https://rfd.shared.oxide.computer/rfd/276#_state_machine_diagram)).

Normally, this period is transitory (e.g. when we go from A2 -> A0): devices report their temperatures immediately upon becoming available, or 1-2 cycles later.  In this case, we weren't giving the device a chance to report a temperature, because it was stuck as `ManagementInterface::Unknown`, so the thermal loop was stuck in `Boot` (at full fan power).

We could argue that in this case, we should exclude the unknown transceiver from the thermal loop entirely, i.e. not adding it to `self.thermal_models` until it's known.  I'm on the fence: we've got a device plugged in, and it's powered, and we don't know how hot it is; up until now, that situation has meant "FULL FAN POWER", but I could be persuaded otherwise.

mkeeter: Hmmm, I'm not off the fence yet.  If we're trusting the host to do the right thing, then couldn't we trust the host to **not power on** a transceiver with unknown power management mode?

In that case, the code as written in the PR would still be reasonable, as a backup defense against a defecting host...

mkeeter: Aaron reminded me that we can't actually check transceiver type until they're powered on, so the host isn't able to make that decision for us.  With that, I can get off the fence and agree that not spinning up fans seems reasonable.

Aaron-Hartwig: Chatted with Matt a bit in DMs and we came to agreement that the module should just be ignored during the intermediate state from when it is placed into low-power mode (powered + out of reset) to if the host will leave it powered on or not. This is because the time when the host can identify what the module is is exactly when the SP will start trying to add it to the thermal control loop. During that time (likely seconds), the host needs to figure out if it is a supported module or not and if it is not supported, place it in reset and power it off.

Aaron-Hartwig (approved):
Thanks for the discussion, this looks good! Also thanks for the bonus task of cleaning up the ringbuf a bit.

PR #1251, opened 2023-03-30
kern: rework dump impl to reduce UB

cbiffle:
Doing a kernel dump from within the kernel is inherently hairy. We wind up needing to do a raw byte read of a Task struct, which we also need to interact with as an actual Task.

The implementation was
- Creating (using core::slice::from_raw_parts) a byte slice that aliased the &mut [Task] task table
- Then using it and the &mut concurrently

which is UB due to aliasing violations.

The code also contained an easily accessible way of crashing the kernel: send in a region such that base+size overflows. Triggering this reliably required choosing a base value within the TCB address range, because the other code path used USlice.

I've reworked the logic. The particular code that was causing the UB issue has been replaced by

- Deriving a &mut Task to the specific task we want to access, and then
- Deriving a &[u8] from it using transmute and shadowing to prevent aliasing.

(This was also the second time in the kernel I've wanted the index2_distinct function, so I've pulled it into an accessible common location.)

I've eliminated the caller-controlled kernel panic by using USlice along all code paths. This had the side effect of removing some runtime overflow checks from the generated code, slightly reducing code size. (However, the code is slightly bigger overall as a result of this change.)

I've also added a crapload of comments to try and help future readers of this subtle routine.

### DIFF
--- a/sys/kern/src/kipc.rs
+++ b/sys/kern/src/kipc.rs
@@ -455,7 +455,9 @@ fn read_task_dump_region(
                     // at first glance since it's backwards from every other
                     // routine in the kernel -- but its shape is deliberate! We
                     // really do want to apply the SRC fault to the caller,
-                    // followed by any remaining DST fault.
+                    // followed by any remaining DST fault (by returning it from
+                    // this function, which means it'll be applied by the
+                    // generic syscall error handler).
                     //
                     // Note that if the supervisor is _really_ misbehaving, this
                     // can result in the delivery of two faults to it (the SRC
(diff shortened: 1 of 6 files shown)


sys/kern/src/kipc.rs
```diff
@@ -378,22 +432,40 @@ fn read_task_dump_region(
...
+                    // This means the InteractFault produced by `safe_copy` is
+                    // shaped slightly wrong for the task. Counter-intuitively,
+                    // the supervisor (meaning `caller`) should be blamed for
+                    // faults on _both_ the source and destination side.
+                    //
+                    // So we do this using the following code, which looks wrong
+                    // at first glance since it's backwards from every other
+                    // routine in the kernel -- but its shape is deliberate! We
+                    // really do want to apply the SRC fault to the caller,
+                    // followed by any remaining DST fault.
```
mkeeter, line 458: Maybe add a comment here explaining that the `DST` fault is applied all the way at the top of the call stack in `safe_syscall_entry`, since I had to search for that.

cbiffle: done!

PR #1252, opened 2023-03-31
Illuminate user LED on PSC

mkeeter:
This works on Ian's rev B, although it doesn't work on my rev A; Ian is going to double-check the rev A layout, because it may disagree with the schematic in EDM.


isobering:
Checked the version A layout, and it agrees with the schematic. Not sure why Matt's version A status LED doesn't turn on, but since version A PSCs are going to be EOL very soon, let's not worry about it.

jgallagher (approved):

isobering (approved):

PR #1254, opened 2023-03-31
More steps toward rfd374

cbiffle:
This is a collection of the three least-disruptive parts of the RFD374 RoT transition work, factored out so they can land early. If you feel like these should be separate PRs, I'm open to that -- they're all rolled into one here because each diff is relatively short, and they're sort of vaguely related, kind of? I guess?

1. Fixed memory handling and alias-related UB in the DICE SYSCON register access (the registers are not modeled by the PAC, and we don't like writing YAML diffs against XML files, so we're modeling them by hand).
2. Turned off ITM on all LPC55 builds and moved it into the test suite only.
3. Make all LPC55 builds signed by default, and switch off secure-separation for now.


lib/dice/src/lib.rs
```diff
@@ -39,22 +40,26 @@ mod trust_quorum_dhe_cert_tmpl;
...
-    unsafe {
-        slice::from_raw_parts_mut(
-            REG_ADDR_NONSEC as *mut u32,
-            SEED_LENGTH / mem::size_of::<u32>(),
-        )
-    }
+
+/// Retrieves the bank of CDI registers from the SYSCON as a slice of volatile
+/// cells.
+fn get_cdi_reg_slice(
```
mkeeter, line 46: Out of curiosity, why keep this as a separate function versus moving it into the `CdiReg` constructor?

mkeeter:
Per chat, we should also remove `secure-separation`, `secure-task`, and `HUBRIS_SECURE` from the build system entirely; they'll be unused after this merges, and our future TZ use may take a different form.

lib/dice/src/lib.rs
cbiffle: *shrug* I could go either way, I think I was mostly trying not to introduce unnecessary churn here when fixing the aliasing.

cbiffle:
Preparing a followup for the `secure-*` removal.

PR #1255, opened 2023-03-31
Build system cleanups as followup to #1254

cbiffle:
- Bumps hubtools to a version that accepts the new or old magic numbers in images, in preparation for the magic number change.
- Improves diagnostics of hubtool errors by adding context messages in the build system
- Removes currently-untested trustzone related config stuff in the TOML and collapses it down to the "not trustzone" case.


andrewjstone (approved):
LGTM

PR #1258, opened 2023-04-03
Implement MGS LED control

mkeeter:
This should be merged after https://github.com/oxidecomputer/management-gateway-service/pull/76

It plumbs system LED control through `control-plane-agent`.

### DIFF
--- a/task/control-plane-agent/src/mgs_sidecar.rs
+++ b/task/control-plane-agent/src/mgs_sidecar.rs
@@ -14,10 +14,10 @@ use gateway_messages::sp_impl::{
     BoundsChecked, DeviceDescription, SocketAddrV6, SpHandler,
 };
 use gateway_messages::{
-    ignition, ComponentDetails, ComponentUpdatePrepare, DiscoverResponse,
-    IgnitionCommand, IgnitionState, MgsError, PowerState, SlotId, SpComponent,
-    SpError, SpPort, SpState, SpUpdatePrepare, SwitchDuration, UpdateChunk,
-    UpdateId, UpdateStatus,
+    ignition, ComponentAction, ComponentDetails, ComponentUpdatePrepare,
+    DiscoverResponse, IgnitionCommand, IgnitionState, MgsError, PowerState,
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -117,7 +117,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 
 # Oxide forks and repos
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"], branch = "led-component-action" }
```
jgallagher, line 120: I know you know this already but just leaving a comment for clarity - we should switch this back to `main` after https://github.com/oxidecomputer/management-gateway-service/pull/76 lands.

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -314,6 +315,21 @@ impl SpHandler for MgsHandler {
...
 
+    fn component_action(
+        &mut self,
+        _sender: SocketAddrV6,
+        component: SpComponent,
+        action: ComponentAction,
+    ) -> Result<(), SpError> {
+        match (component, action) {
+            (SpComponent::SYSTEM_LED, ComponentAction::Led(action)) => {
+                // TODO: implement this
```
jgallagher, line 326: Do you want to do this before merging, or does it require other work first?

mkeeter: This is pending @Aaron-Hartwig 's work, because the system LED is actually controlled by one of the front IO FPGAs.

Aaron-Hartwig: I agree this is fine to leave as a TODO, when Ben and I get the main transceiver management rewrite through I'm hopping onto the LEDs.

Aaron-Hartwig (approved):
LGTM - I'll update the Sidecar to support this when I update the transceiver LED control code.

PR #1259, opened 2023-04-03
Remove more remnants of TZ config

mkeeter:
Removes the `trustzone.x` linker script, SAU entries from `abi::ImageHeader` (and build system), and `tz_support` feature.

We leave padding in the `ImageHeader`, because the existing `stage0` looks for `header.version` at a known offset; this can be deleted once we switch to the new `stage0`.


andrewjstone (approved):
LGTM. Thanks for removing the SAU entries from the image header. That sounds very satisfying to delete.

PR #1261, opened 2023-04-04
Plumb RoT boot selection up through control plane agent

andrewjstone:
This code plumbs up the `switch_default_update` function of the lpc55-update-server up through the control-plane-agent. It also plumbs through a rot reset which allows us to test the ping-pong functionality in cfpa via `humility readmem`. 

This is mostly code from @lzrd with some minor changes from me after discussion. I've tested it via faux-mgs and it works well.

This depends on https://github.com/oxidecomputer/management-gateway-service/pull/77, which must merge in first.


app/gimlet-rot/app-b.toml
```diff
@@ -21,6 +21,9 @@ features = ["itm"]
 stacksize = 1536
 notifications = ["fault", "timer"]
 
+[tasks.jefe.config.allowed-callers]
+request_reset = ["hiffy", "update_server"]
```
mkeeter, line 25: Why `hiffy` here?

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -97,6 +101,12 @@ impl MgsCommon {
         Ok(state)
     }
 
+    /// The SP, having reset itself, will not be able to respond to
+    /// the later reset_trigger message.
+    ///
+    /// So, after getting an ACK for the prepare message, MGS will send and
+    /// retry the reset_trigger message until it gets rejected for lack of
+    /// a corresponding prepare message.
```
mkeeter, line 109: I'd suggest modifying these functions to call into `reset_component_prepare` and `reset_component_trigger`, deleting `self.reset_requested` (since `self.reset_component_requested` serves the same purpose).

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -143,14 +153,153 @@ impl MgsCommon {
...
+        match component {
+            SpComponent::SP_ITSELF => {
+                task_jefe_api::Jefe::from(crate::JEFE.get_task_id())
+                    .request_reset();
+                // If `request_reset()` returns,
+                // something has gone very wrong.
+                panic!();
+            }
+            SpComponent::ROT => {
+                // mgs_{gimlet,psc,sidecar}.rs deal with any board specific
```
mkeeter, line 213: This comment is confusing; it should instead live next to the `return Err(SpError...`) line.

I'd also suggest moving `match update.sprot_task().reset() { ...` into this block of the case statement; otherwise, you have to notice that all the other cases diverge and only this one falls through.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -143,14 +153,153 @@ impl MgsCommon {
...
+                // reset strategy. Here we take care of common SP and RoT cases.
+            }
+            _ => return Err(SpError::RequestUnsupportedForComponent),
+        };
+        // We're dealing with  RoT targets at this point.
```
mkeeter, line 218:
```suggestion
        // We're dealing with RoT targets at this point.
```

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -143,14 +153,153 @@ impl MgsCommon {
...
+                // Some other error occurred.
+                // Update is all-or-nothing at the moment.
+                // The control plane can try to reset the RoT again or it
+                // can start the update process all over again.  We should
+                // be able to make incremental progress if there is some
+                // bug/condition that is degrading SpRot communications.
+                ringbuf_entry!(Log::RotReset { err });
+                Err(SpError::ComponentOperationFailed(err.into()))
+            }
+            Ok(()) => Ok(()),
```
mkeeter, line 246: Should we log an error in this case, since we expected `Err(RspTimeout)`?

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -143,14 +153,153 @@ impl MgsCommon {
...
+            SwitchDuration::Once => drv_sprot_api::SwitchDuration::Once,
+            SwitchDuration::Forever => drv_sprot_api::SwitchDuration::Forever,
+        };
+        match component {
+            SpComponent::ROT => {
+                match update.sprot_task().switch_default_image(slot, duration) {
+                    Err(err) => {
+                        // Some other error occurred.
+                        // Update is all-or-nothing at the moment.
+                        // The control plane can try to reset the RoT again or it
```
mkeeter, line 276: This comment seems incorrect (copypasted from above?)

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -1029,6 +1029,49 @@ impl SpHandler for MgsHandler {
...
+        )
+    }
+
+    fn reset_component_prepare(
+        &mut self,
+        _sender: SocketAddrV6,
+        _port: SpPort,
+        component: SpComponent,
+    ) -> Result<(), SpError> {
+        match component {
```
mkeeter, line 1059: Same here, and for `mgs_{sidecar,psc}.rs`

task/hiffy/src/common.rs
```diff
@@ -594,12 +596,14 @@ pub(crate) fn sprot_send_recv_retries(
             &mut *rval,
         ),
     )?;
-    match drv_sprot_api::MsgType::from(result.msgtype) {
+    let msgtype = drv_sprot_api::MsgType::from_u8(result.msgtype)
+        .ok_or(hif::Failure::FunctionError(1))?;
```
mkeeter, line 600: Why make this change?  We previously were using `MsgType::from`, which converts unknown values into `MgsType::Unknown`, which takes the `_ => {` branch below, so the behavior should be the same.

drv/sprot-api/src/lib.rs
```diff
@@ -32,12 +33,14 @@ use zerocopy::{ByteOrder, LittleEndian};
 
 /// The canonical SpRot protocol error returned by this API
 //
-// TODO: Audit that each MsgError is used and has some reasonable action.
+// Audit that each MsgError is used and has some reasonable action.
```
mkeeter, line 36: seems like leaving the TODO here is reasonable

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -78,7 +77,7 @@ fn configure_spi() -> Io {
     let gpio_driver = GPIO.get_task_id();
     setup_pins(gpio_driver).unwrap_lite();
     let gpio = drv_lpc55_gpio_api::Pins::from(gpio_driver);
-    // TODO: It should never happen but, initialization should be able to deal
+    // It should never happen but, initialization should be able to deal
```
mkeeter, line 80: Not sure why we're removing the TODO if this isn't fixed.

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -295,7 +292,18 @@ impl Io {
                 // potentially throttle sends in the future.
                 self.spi.rxerr_clear();
                 self.stats.rx_overrun = self.stats.rx_overrun.wrapping_add(1);
-                err = Some(IoError::Flow);
+                // If we were just sending our response, and SP was
+                // just sending zeros and we received the first byte
+                // correctly and that first byte was zero, then
```
mkeeter, line 297: Where do we check that the first byte was zero?

drv/sprot-api/README.md
```diff
@@ -257,7 +257,6 @@ Message types include:
     to test the RoT's ability to exchange messages without error.
     The `SinkRsp` is just a header, no payload.
     On error, an `ErrorRsp` message is sent.
-  - _Unknown_ - 0xff A reserved, internal representation for an unsupported message type.
```
mkeeter, line 260: It looks like `Unknown` still exists; why delete it here?

drv/sprot-api/src/lib.rs
```diff
@@ -137,8 +141,12 @@ pub enum SprotError {
     #[idol(server_death)]
     ServerRestarted,
 
-    /// Unknown Errors are mapped to 0xff
-    Unknown = 0xff,
+    // Used if no explicit error code is available.
+    Unknown,
```
mkeeter, line 145: Why remove the explicit value for `Unknown`?

In general, I'd suggest removing the `Unknown` variant entirely and using an `Option<SprotError>` instead (where `Unknown` becomes `None`).

drv/sprot-api/src/lib.rs
```diff
@@ -252,7 +254,7 @@ pub struct SprotStatus {
     /// CRC32 of the LPC55 boot ROM contents.
     /// The LPC55 does not have machine readable version information for
     /// its boot ROM contents and there are known issues with old boot ROMs.
-    /// TODO: This should live in the stage0 handoff info
+    /// This should live in the stage0 handoff info.
```
mkeeter, line 257: Is this done?  If not, leaving the TODO seems helpful.

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -404,32 +422,35 @@ impl<S: SpiServer> Io<S> {
...
                     }
                 }
             }
         }
-        Err(errcode)
+        Err(errcode.unwrap_lite())
     }
 
-    // TODO: Move README.md to RFD 317 and discuss:
+    // Move README.md to RFD 317 and discuss:
```
mkeeter, line 453: Again, if these things aren't done, then leave the TODOs.

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -404,32 +422,35 @@ impl<S: SpiServer> Io<S> {
...
-                                SprotError::from(rxmsg.payload_error_byte());
-                            ringbuf_entry!(Trace::SprotError(errcode));
-                            if is_recoverable_error(errcode) {
-                                // TODO: There are rare cases where
-                                // the RoT dose not receive
-                                // a 0x01 as the first byte in a message.
-                                // See issue #929.
+                            let err =
+                                SprotError::from_u8(rxmsg.payload_error_byte())
+                                    .unwrap_or(SprotError::Unknown);
```
mkeeter, line 427: This plumbing is very confusing to me.

We take a `u8`, decode it into `Option<SprotError>`, unwrap that into either the value or `SprotError::Unknown`, then wrap it in `Some(...)` again.

It seems like we're tracking two mostly-separate things:

- The most recent error – which may or may not exist – and is returned at the end of the function, at which point it is guaranteed to exist (since we're unwrapping it).
- Whether we successfully deserialized the error byte

Combine them in this way is confusing.  The nicely-typed option would be to add a new wrapper type, e.g.
```rust
enum SpSprotError {
    Sprot(SprotError),
    BadDeserialization
}
```

However, this couldn't be used in Idol right now, because it doesn't support complex error types.

The dumb way to do it would be
```rust
enum SpSprotError {
    NoMessage = 1,
    BadTransferSize,
    TaskRestart,
    InvalidCrc,
    FlowError,
    // literally every other field from SprotError
    DeserializationFailed, // <-- this is new
}

impl From<SprotError> for SpSprotError {
    fn from(s: SprotError) -> Self {
        match s {
            SprotError::NoMessage => Self::NoMessage,
            // etc
        }
    }
}
```

Any other ideas on how to clean up this section?

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -404,32 +422,35 @@ impl<S: SpiServer> Io<S> {
...
+                            errcode = Some(err);
+                            ringbuf_entry!(Trace::SprotError(err));
+                            if is_recoverable_error(err) {
```
mkeeter, line 430: `is_recoverable()` could be a method on `err`, especially if we have a new `SpSprotErr` type (see above).

task/hiffy/src/common.rs
```diff
@@ -550,12 +550,14 @@ pub(crate) fn sprot_send_recv(
             &data[0..len],
             &mut *rval,
         ))?;
-    match drv_sprot_api::MsgType::from(result.msgtype) {
+    let msgtype = drv_sprot_api::MsgType::from_u8(result.msgtype)
```
mkeeter, line 553: Same question [as below](https://github.com/oxidecomputer/hubris/pull/1261/files#r1157343602)

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -438,7 +459,7 @@ impl<S: SpiServer> Io<S> {
     // But it would be ok to overlap our new request with receiving
     // of a previous response.
     //
-    // TODO: The RoT must be able to observe SP resets. During the
+    // The RoT must be able to observe SP resets. During the
```
mkeeter, line 462: I think this is still a TODO, right?

drv/lpc55-sprot-server/src/main.rs
cbiffle: Also, it absolutely can happen -- all it would take is this task crashing during a transaction.

drv/stm32h7-sprot-server/src/main.rs
lzrd: Yes. It should be an issue if it's not already.

task/hiffy/src/common.rs
```diff
@@ -729,6 +733,30 @@ pub(crate) fn sprot_block_size(
     Ok(4)
 }
 
+#[cfg(feature = "sprot")]
```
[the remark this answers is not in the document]
andrewjstone: Based on the discussion in chat and response from @cbiffle, the `sprot` client API seems significant enough to keep these explicit functions. I prefer we leave them in.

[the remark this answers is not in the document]
andrewjstone: At least the more complex ones.

app/gimlet-rot/app-b.toml
andrewjstone: Good point, `hiffy` is not calling directly to jefe, but going through `update_server`. I believe this can be removed here and in the other `.toml` files. will do.

drv/lpc55-sprot-server/src/main.rs
andrewjstone: @cbiffle @lzrd This can happen, but I'm not sure there's actually anything that needs doing with the current structure of the code. Either `next` will interrupt on `ssa`(CSn) and it will result in a flow error, or the SP will timeout and retry.  I think the comment should just be removed altogether at this point.

andrewjstone: I went ahead and deleted the comment. We can re-add it or open an issue if necessary, but I don't see what's actionable from it at the moment.

drv/lpc55-sprot-server/src/main.rs
andrewjstone: I fixed up the comment here to refer to `Protocol::Ignore` instead of `byte was zero`.

drv/lpc55-update-server/src/main.rs
```diff
@@ -637,7 +648,13 @@ fn target_addr(image_target: UpdateTarget, page_num: u32) -> Option<u32> {
     Some(addr)
 }
 
+fn do_reset() -> ! {
```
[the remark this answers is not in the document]
andrewjstone: Done. I also removed the implicit TODO above reset and disallowed it if an update was in progress.

drv/sprot-api/README.md
andrewjstone: `Unknown` has been removed as of https://github.com/oxidecomputer/hubris/pull/1261/commits/252293b67129f59566d41e5bf15ff991f9d1d32b

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -698,9 +719,10 @@ impl<S: SpiServer> idl::InOrderSpRotImpl for ServerImpl<S> {
                                             SprotError::BadMessageLength,
                                         );
                                     }
-                                    break Err(SprotError::from(
+                                    break Err(SprotError::from_u8(
```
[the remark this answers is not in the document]
andrewjstone: I fixed the above comment. Github won't show it to me now that I've changed the code... sigh. But you can see my choices here: https://github.com/oxidecomputer/hubris/pull/1261/commits/252293b67129f59566d41e5bf15ff991f9d1d32b

There no longer exists a `SprotError::Unknown` variant. I re-used the `BadResponse` variant which wasn't used elsewhere in lieu of adding a new error variant. Due to the restructuring there was also no need to introduce a new type like `SpSprotError`. I hope you like it!

drv/update-api/src/lib.rs
```diff
@@ -54,41 +52,46 @@ pub enum UpdateStatus {
...
 
     // Caboose checks
-    ImageBoardMismatch = 23,
-    ImageBoardUnknown = 24,
+    ImageBoardMismatch,
+    ImageBoardUnknown,
 
     #[idol(server_death)]
-    ServerRestarted = 22,
+    ServerRestarted,
```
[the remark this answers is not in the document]
andrewjstone: I don't think this matters at this point. That value is only passed upstack to be converted by Sprot which will look at the variant not value.

drv/update-api/src/lib.rs
```diff
@@ -54,41 +52,46 @@ pub enum UpdateStatus {
...
 
-    Unknown = 0xff,
+    // During update when there may be version skew, the RoT may produce an
+    // error that the SP doesn't know.
+    Unknown,
```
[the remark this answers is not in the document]
andrewjstone: I agree that `Option<UpdateError>` would be better here. However, there are very minor changes made to the update server in this PR. I don't think we should hold up this PR for a refactoring outside its general purview.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -527,11 +527,23 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
 
     fn switch_default_image(
         &mut self,
-        _: &userlib::RecvMessage,
+        _: &RecvMessage,
         _slot: SlotId,
         _duration: SwitchDuration,
     ) -> Result<(), RequestError<UpdateError>> {
-        Err(UpdateError::Unknown.into())
+        Err(UpdateError::NotImplemented.into())
```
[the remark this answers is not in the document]
andrewjstone: Added a comment in https://github.com/oxidecomputer/hubris/pull/1261/commits/8068dc61c96f59956703fd062586ce024a8b81c5

task/control-plane-agent/src/mgs_common.rs
andrewjstone: Good suggestion. I did you one better though! After chatting with @jgallagher this morning, I updated MGS to not call these functions anymore and instead use the component versions like this: https://github.com/oxidecomputer/management-gateway-service/commit/cf6bf75986289892eccd562231afea197539a30b#diff-4e2f0f9389ed68d983518b63098811d189767e0b271891d22cf29aa2a77794c4R783-R788

This means I was able to remove `reset_component` and `reset_trigger` altogether in https://github.com/oxidecomputer/hubris/pull/1261/commits/127cbea286c342efe0af81eed651411bc4e33d95

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -23,6 +23,7 @@ task_slot!(PACKRAT, packrat);
 /// Provider of MGS handler logic common to all targets (gimlet, sidecar, psc).
 pub(crate) struct MgsCommon {
     reset_requested: bool,
+    reset_component_requested: SpComponent,
```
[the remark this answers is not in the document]
andrewjstone: Fixed in [a2438a3](https://github.com/oxidecomputer/hubris/pull/1261/commits/a2438a32a01c48a7f00da2362d0575a63b2653f1)

task/control-plane-agent/src/mgs_common.rs
andrewjstone: Both suggestions fixed in https://github.com/oxidecomputer/hubris/pull/1261/commits/c67a55f3bd72ec7f01635da43ba0e53555405e7f

task/control-plane-agent/src/mgs_common.rs
andrewjstone: I logged an error here, but I'm wondering if we should actually also return an error. That would require a change in gateway-messages, so I didn't do it yet.

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -1029,6 +1029,49 @@ impl SpHandler for MgsHandler {
...
+            SpComponent::ROT | SpComponent::SP_ITSELF => {}
+            _ => return Err(SpError::RequestUnsupportedForComponent),
+        }
+        self.common.switch_default_image(
+            &self.sp_update,
+            component,
+            slot,
+            duration,
+        )
```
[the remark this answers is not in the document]
andrewjstone: I removed all the extra checks.

task/hiffy/src/common.rs
andrewjstone: It looks like `MsgType::Unknown` non longer exists, so that's probably why.

drv/update-api/src/lib.rs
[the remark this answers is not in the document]
mkeeter: Okay, seems sensible

Issue #1217, opened 2023-03-17
Dumping needs to go through something other than udprpc


cbiffle:
We will probably want remote dumping in prod. We do not want udprpc in prod. So, we need to route dumping through some other facility -- either a network socket on the dump agent itself (which means it'd work even when e.g. control-plane-agent is hung up), or maybe control-plane-agent.

jclulow:
I feel like many of the dumps I have personally needed to take thus far have been when CPA has become unresponsive, so I would probably vote for isolation if possible.

cbiffle:
I believe this is effectively fixed after @mkeeter's work on dump-agent.

PR #1278, opened 2023-04-10
Add network feature to `dump-agent`

mkeeter:
This is part of the https://github.com/oxidecomputer/hubris/issues/1217 saga, which begins with https://github.com/oxidecomputer/humpty/pull/4/files and also requires a Humility PR.

This Hubris PR adds an optional `net` feature to `dump-agent`.  When the feature is enabled, `dump-agent` listens on a UDP socket for messages which encode dump commands, using API types from `humpty` (see PR above).  These messages are a thin layer above the existing hiffy API.

Most of the code changes are in the `udp` module; changes in `ServerImpl` are mostly to make more functions callable from either hiffy or the `udp` module.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, branch = "udp-types" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.1" }
 hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false }
 idol = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
 idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { path = "../humpty", default-features = false }
+humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, branch = "udp-types" }
 hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false }
 idol = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
 idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false }
+humpty = { path = "../humpty", default-features = false }
 hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false }
 idol = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
 idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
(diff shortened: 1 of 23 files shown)


Cargo.toml
```diff
@@ -119,7 +119,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false }
+humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, branch = "udp-types" }
```
jgallagher, line 122: I know you're planning to do this already, but commenting as a reminder to switch this back to the main branch after the humpty PR is merged.

task/dump-agent/src/udp.rs
```diff
@@ -0,0 +1,167 @@
...
+    Result<humpty::udp::Response, humpty::udp::Error>,
+)>::MAX_SIZE;
+
+// We are receiving a (Header, Request) from the host
+const MAX_UDP_RX_SIZE: usize =
+    <(humpty::udp::Header, humpty::udp::Request)>::MAX_SIZE;
+
+// Check against packet sizes in the TOML file
+static_assertions::const_assert!(MAX_UDP_TX_SIZE <= 1024);
+static_assertions::const_assert!(MAX_UDP_RX_SIZE <= 1024);
```
jgallagher, line 38: This might be a bunch of busywork for very little gain so feel free to not do it, but: should our build script emit the socket size so we don't have to keep this literal in sync?

task/dump-agent/src/udp.rs
```diff
@@ -0,0 +1,167 @@
...
+    WrongVersion(u8),
+}
+
+ringbuf!(Trace, 16, Trace::None);
+
+// We are sending a (Header, Result<Response, Error>) to the host
+const MAX_UDP_TX_SIZE: usize = <(
+    humpty::udp::Header,
+    Result<humpty::udp::Response, humpty::udp::Error>,
+)>::MAX_SIZE;
```
jgallagher, line 30: It might be cleaner to define structs in `humpty` that contain both the header and request/response instead of using tuples. (I assume as written humility will have to repeat this constant and also serialize/deserialize into tuples?)

task/dump-agent/src/udp.rs
```diff
@@ -0,0 +1,167 @@
...
+            ) {
+                // We'll drop packets if the outgoing queue is full;
+                // the host is responsible for retrying.
+                //
+                // Other errors are unexpected and panic.
+                //
+                // This includes ServerRestarted, because the server should only
+                // restart if the watchdog times out, and the watchdog should
+                // not be timing out, because we're literally replying to a
+                // packet here.
```
jgallagher, line 108: I'm not completely sold on panicking on `ServerRestarted`. The server should only restart if the watchdog times out _today_, but this seems to bake in knowledge about the net server that could change in the future. If `net` panics, we also panic, which doesn't seem ideal. Is there a downside to moving `ServerRestarted` into the "just drop the packet" arm with `QueueFull`?

task/dump-agent/src/udp.rs
```diff
@@ -0,0 +1,167 @@
...
+                Request::TakeDump => {
+                    self.take_dump().map(|()| Response::TakeDump)?
+                }
+            },
+            Err(e) => {
+                // This message is from a newer version, so it makes sense that
+                // we failed to deserialize it.
+                if header.version > humpty::udp::version::CURRENT {
+                    ringbuf_entry!(Trace::WrongVersion(header.version));
+                    return Err(humpty::udp::Error::VersionMismatch);
```
jgallagher, line 158: Could we include the SP's `version::CURRENT` (and probably `header.version` while we're at it) in this error variant?

mkeeter: Yup, done in 912a2674dd7b243c521f3011702689e8ee64157e

task/dump-agent/src/udp.rs
mkeeter: Turns out this wasn't too hard, done in 5ceb1a2ab40ae0f8ce5fbf8af1c47b0c60c67c7b

task/dump-agent/src/udp.rs
mkeeter: This was copy-pasted from `transceivers`, but I'm not convinced – I've moved `ServerRestarted` into the drop-the-packet branch and updated the comment in 37990a26c834b996bfa8a203fda959a66ff5c26f

app/gimlet/rev-b.toml
```diff
@@ -1280,10 +1282,17 @@ rx = { packets = 3, bytes = 1024 }
...
 owner = {name = "control_plane_agent", notification = "socket"}
-port = 11111 # TODO do we have a documented port for MGS traffic?
+port = 11111
 tx = { packets = 3, bytes = 2048 }
 rx = { packets = 3, bytes = 2048 }
 
+[config.net.sockets.dump_agent]
+kind = "udp"
+owner = {name = "dump_agent", notification = "socket"}
+port = 11113
```
bcantrill, line 1292: I know that this is somewhat unrelated, but worth (1) pointing to oxidecomputer/oana and/or (2) making oxidecomputer/oana public? (And I know that this might be a question for @rmustacc.)

mkeeter: Good idea, added a comment to the top of the `[config.net]` section in all of the relevant TOML files

task/dump-agent/src/udp.rs
```diff
@@ -0,0 +1,176 @@
...
+        if let Some(out_len) = out_len {
+            meta.size = out_len as u32;
+            if let Err(e) = self.net.send_packet(
+                SocketName::dump_agent,
+                meta,
+                &tx_data_buf[..meta.size as usize],
+            ) {
+                // We'll drop packets if the outgoing queue is full or the
+                // server has died; the host is responsible for retrying
+                // (which isn't implemented yet in Humility).
```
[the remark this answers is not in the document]
mkeeter: Yup, deleted!

PR #1279, opened 2023-04-11
[sprot] Use hubpack based messages across the SPI interface

andrewjstone:
In order to simplify the implementation of sprot, we base all sprot protocol messages around hubpack serialization. There is a fixed sized `Header`, and a parameterized `Msg` that can contain either a `ReqBody` or `Result<RspBody, SprotError>`. This removes a ton of hand-written serialization code that was quite hard to reason about. Overall it should  be easier to extend the protocol with new messages now.

This change also changes the `SprotError` from a  `CLike` error to a `Complex` error in idol terminology. We no longer translate sub-errors into a global list of integers, but compose them into a `SprotError` enum.

In order to improve performance on the RoT side, we no longer bother transmitting bytes when receiving messages or pulling bytes out  of the receive fifo when sending messages.

This change will require a hubris flag day for synchronized updating of RoT and SP. 

All the hiffy code has been ripped out and @bcantrill is working on support for idol `Complex` errors so we can get the functionality back. 

Testing has been performed via faux-mgs and updates work when ping is disabled, with zero errors seen thanks to
the change to use 16-byte fifos.

Additionally, dump functionality is now directly available from the sprot api.


drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -50,213 +54,125 @@ impl Handler {
...
+                    protocol: rx_buf[0],
+                    body_type: rx_buf[1]
+                });
+                request
+            }
+            Err(e) => {
+                ringbuf_entry!(Trace::Err(e));
                 stats.rx_invalid = stats.rx_invalid.wrapping_add(1);
-                return Some(tx_buf.error_rsp(msgerr));
+                return Response::pack(&Err(e.into()), tx_buf).unwrap_lite();
```
cbiffle, line 89: nit: it seems like this places some requirements on `tx_buf` to avoid a panic here. If I called with a `tx_buf` of length 0, for instance, it'd panic. It'd be nice to put this (and any other usage restrictions) in a comment on the function.

Alternatively, it looks like this is always called with a `&[u8; MAX_RESPONSE_SIZE]` -- if `tx_buf`'s type were `&[u8; MAX_RESPONSE_SIZE]` instead of `&[u8]` then I think you _couldn't_ misuse it in this way?

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -50,213 +54,125 @@ impl Handler {
...
-                } else {
-                    Ok(tx_buf.no_payload(MsgType::SinkRsp))
-                }
-            }
-            MsgType::DumpReq => {
-                let addr =
-                    u32::from_le_bytes(rx_payload[0..4].try_into().unwrap());
+            ReqBody::IoStats => Ok(RspBody::IoStats(stats.clone())),
+            ReqBody::Sprockets(req) => {
+                // TODO: Don't unwrap!
```
cbiffle, line 122: Since you have context, would you mind adding a sentence to this comment explaining how you'd avoid unwrapping here? What's the potential panic?

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -50,213 +54,125 @@ impl Handler {
...
-                let rval: u32 = if let Err(e) = dumper.dump(addr) {
-                    e.into()
-                } else {
-                    0
-                };
-
-                let tx_payload = tx_buf.payload_mut();
-                tx_payload[0..4].copy_from_slice(&rval.to_le_bytes());
-                tx_buf.from_existing(MsgType::DumpRsp, 4)
+                match dumper.dump(addr) {
```
cbiffle, line 130: FWIW, this could also be written as

```rust
dumper.dump(addr).map_err(SprotError::Dump)?;
Ok(RspBody::Ok)
```

I like the concise version better, but use whichever you prefer.

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -50,213 +54,125 @@ impl Handler {
...
-                            .update
-                            .switch_default_image(header.slot, header.duration)
-                            .map(|_| None)
-                            .map_err(|e| e.into());
-                        tx_buf.serialize(MsgType::UpdSwitchDefaultImageRsp, rsp)
+            ReqBody::Update(UpdateReq::GetBlockSize) => {
+                let size = self.update.block_size()?;
+                // Block size will always fit in a u32 on these MCUs
+                Ok(RspBody::Update(UpdateRsp::BlockSize(
+                    size.try_into().unwrap(),
```
cbiffle, line 139: (using `unwrap_lite` here would prevent a reader double-take, even though this unwrap should be removed at compile time because, as you note, usize == u32 here)

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+/// Note that `MSG`s do not implement `Serialize`, `Deserialize`, or
+/// `SerializedSize`, as they need to calculate and place a CRC in the buffer.
+/// `Msg`s sometimes include a an offset into the buffer where a binary blob
+/// resides.
+pub struct Msg<T> {
+    pub header: Header,
+    pub body: T,
+
+    // The index into the serialized buffer where an optional binary blob lives
+    pub blob: Option<BlobInfo>,
```
cbiffle, line 85: tl;dr: Should this be a `&[u8]`?

`BlobInfo` is an offset and size, which other code typically uses to slice some bytes. As an alternative, you might also consider making this field a `&'a [u8]` referencing the original buffer. This would let callers interact with it without additional slicing code.

You could also eliminate the `Option` and use an empty slice to represent "no blob" -- currently I think the structure is redundant, in that I could have `None` or `Some(BlobInfo { size: 0, .. })`

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -50,213 +54,125 @@ impl Handler {
...
+            }
+            ReqBody::Update(UpdateReq::Prep(target)) => {
+                self.update.prep_image_update(target)?;
+                Ok(RspBody::Ok)
+            }
+            ReqBody::Update(UpdateReq::WriteBlock { block_num }) => {
+                match req.blob {
+                    Some(blob) => {
+                        let end = blob.offset + blob.size;
+                        let block = &rx_buf[blob.offset..end];
```
cbiffle, line 150: (if you accept my suggestion about changing the type of `blob` (see drv-sprot-api), this code goes away...)

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -50,213 +54,125 @@ impl Handler {
...
+                        self.update
+                            .write_one_block(block_num as usize, &block)?;
                     }
-                    Err(e) => Err((tx_buf, e.into())),
+                    _ => return Err(SprotProtocolError::MissingBlob)?,
```
cbiffle, line 154: FYI, An alternate way of expressing this is

```rust
let blob = req.blob.ok_or(SprotProtocolError::MissingBlob)?;
let end = ....
```

which loses a level of indent and the match

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -149,63 +154,47 @@ enum IoError {
...
-        tx_buf: &[u8],
-        rx_msg: &mut RxMsg<'a>,
-    ) -> Result<(), IoError> {
-        let mut tx_iter = tx_buf.iter();
-        self.prime_tx_fifo(&mut tx_iter);
-        let result = loop {
+    // Wait for chip select to be asserted
+    // Assert ROT_IRQ if this is a reply
+    fn wait_for_csn_asserted(&mut self, is_reply: bool) {
+        let mut rot_irq_asserted = false;
```
cbiffle, line 192: How do you know ROT_IRQ is not asserted at this point?

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -220,132 +209,136 @@ impl Io {
...
+
+        // Go into a tight loop receiving as many bytes as we can until we see
+        // CSn de-asserted.
+        let mut bytes_received = 0;
+        let mut rx = rx_buf.iter_mut();
+        while !self.spi.ssd() {
+            while self.spi.has_byte() {
+                bytes_received += 1;
+                let read = self.spi.read_u8();
+                rx.next().map(|b| *b = read);
```
cbiffle, line 231: Heh, this is something I hadn't thought of.

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -220,132 +209,136 @@ impl Io {
...
+        // There may be bytes left in the rx fifo after CSn is de-asserted
+        while self.spi.has_byte() {
+            bytes_received += 1;
+            let read = self.spi.read_u8();
+            rx.next().map(|b| *b = read);
+        }
 
-            // Let's check for any problems
-            let fifostat = self.spi.fifostat();
+        self.check_for_overrun()?;
```
cbiffle, line 244: Nit: the term "overrun" is a bit ambiguous, referring to this as the hardware does (as RX Error) would make it easier to compare to the datasheet. Y'know, for the two of us who will ever do that. :-)

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -220,132 +209,136 @@ impl Io {
...
         }
 
-        if too_many_bytes_received {
-            self.stats.rx_protocol_error_too_many_bytes =
-                self.stats.rx_protocol_error_too_many_bytes.wrapping_add(1);
+        self.spi.ssd_clear();
+
+        // We clocked out at least an existing byte in the fifo, as we fill it on entry to
+        // this function.
+        if (bytes_sent == 0) || self.spi.can_tx() {
```
cbiffle, line 291: (clippy would complain about these unnecessary parens I suspect)

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+use sprockets_common::msgs::{
+    RotError as SprocketsError, RotRequestV1 as SprocketsReq,
+    RotResponseV1 as SprocketsRsp,
+};
+use static_assertions::const_assert;
+use userlib::sys_send;
+
+const CRC16: Crc<u16> = Crc::<u16>::new(&CRC_16_XMODEM);
+pub const CRC_SIZE: usize = <u16 as SerializedSize>::MAX_SIZE;
+// XXX ROT FIFO size should be discovered.
```
cbiffle, line 30: I know you just copied this comment from the before-times, but it seems unnecessary -- it's not like the FIFO size can change.

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+// For simplicity we want to be able to retrieve the header
+// in a maximum of 1 FIFO size read.
+const_assert!(Header::MAX_SIZE <= ROT_FIFO_SIZE);
+
+pub type Request = Msg<ReqBody>;
+pub type Response = Msg<Result<RspBody, SprotError>>;
+
+/// A message header for a request or response
+///
+/// It's important that this header be kept fixed size by limiting the use
```
cbiffle, line 47: What does this comment mean?

I think it means that the header needs to contain all fields with fixed size serialization in Hubpack, which basically means primitives, structs, or simple enums?

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+    pub header: Header,
+    pub body: T,
+
+    // The index into the serialized buffer where an optional binary blob lives
+    pub blob: Option<BlobInfo>,
+}
+
+impl<T> Msg<T>
+where
+    T: Serialize + for<'a> Deserialize<'a> + SerializedSize,
```
cbiffle, line 90: Tip: `for <'a> Deserialize<'a>` should be equivalent to `DeserializeOwned`, which is intended as a shorthand

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+{
+    /// Serialize a `Header` followed by a `ReqBody` or `RspBody`, compute a CRC, serialize
+    /// the CRC, and return the total size of the serialized request.
+    pub fn pack(body: &T, buf: &mut [u8]) -> Result<usize, SprotProtocolError> {
+        // Serialize `body`
+        let mut size = hubpack::serialize(&mut buf[Header::MAX_SIZE..], body)?;
```
cbiffle, line 96: The ? here will return an error in the two cases where serialize can fail:
- If the buffer isn't big enough, which is darn suspicious, or
- If the type T has some aspect incompatible with hubpack.

As a result I'd consider `unwrap_lite` here since those are both programmer errors rather than runtime errors.

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+
+        // Create a header, now that we know the size of the body
+        let header = Header::new(size.try_into().unwrap_lite());
+
+        // Serialize the header
+        size += hubpack::serialize(buf, &header)?;
```
cbiffle, line 102: (If you buy my reasoning above, this also seems like a good use for `unwrap_lite`, since you've already panicked if buf is less than `Header::MAX_SIZE` above by slicing)

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+
+        // Compute and serialize the CRC
+        let crc = CRC16.checksum(&buf[..size]);
+        size += hubpack::serialize(&mut buf[size..], &crc)?;
+
+        Ok(size)
+    }
+
+    /// Serialize a `Header` followed by a `ReqBody` or `RspBody, copy a blob
```
cbiffle, line 111: (missing closing backtick here on RspBody)

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+    pub fn unpack_body(
+        header: Header,
+        // The buffer containing the entire serialized `Msg` including the `Header`
+        buf: &[u8],
+        // The body part of the buffer including the CRC at the end
+        body_buf: &[u8],
+    ) -> Result<Msg<T>, SprotProtocolError> {
+        let (body, blob_buf) = hubpack::deserialize::<T>(body_buf)?;
+        let end = Header::MAX_SIZE + header.body_size as usize;
+        let blob_len =
```
cbiffle, line 160: (note that if you make blob into a `&[u8]` as I suggested above, the rest of this code...goes away, because the correct slice is already available as `blob_buf`)

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,266 @@
...
+        }
+        Self::unpack_body(header, buf, body_buf)
+    }
+
+    /// Deserialize just the body, given a header that was already deserialized.
+    pub fn unpack_body(
+        header: Header,
+        // The buffer containing the entire serialized `Msg` including the `Header`
+        buf: &[u8],
+        // The body part of the buffer including the CRC at the end
```
cbiffle, line 155: Is this accurate? You get the CRC from `buf[end..]` below.

drv/sprot-api/src/lib.rs
```diff
@@ -71,173 +273,73 @@ pub enum SprotError {
...
-    NonRotError,
-    /// A message with version = 0 was received unexpectedly.
-    EmptyMessage,
-
-    /// Insufficient bytes received
-    Incomplete,
-    /// Hubpack error
+    // The request timed out
+    Timeout,
+    // Hubpack error: Used for both serialization and deserialization
```
cbiffle, line 280: (I _suspect_ you can eliminate this. I suspect all the hubpack error cases are programmer errors that can become `unpack_lite`. But I'm not 100% sure.)

drv/sprot-api/src/lib.rs
```diff
@@ -71,173 +273,73 @@ pub enum SprotError {
...
-
-    // SPI-related errors
-    SpiServerLockError,
-    SpiServerWritePart1Error,
-    SpiServerWritePart2Error,
-    SpiServerReleaseError,
-    SpiServerReadPart1Error,
-    SpiServerReadPart2Error,
+    // We cannot assert chip select
     CannotAssertCSn,
```
cbiffle, line 277: ...why would we not be able to assert CS?

drv/sprot-api/src/lib.rs
```diff
@@ -71,173 +273,73 @@ pub enum SprotError {
...
-            | SprotError::RotBusy
-            | SprotError::RspTimeout
-            | SprotError::FlowError
-            | SprotError::UnsupportedProtocol
-            | SprotError::BadMessageLength
-            | SprotError::Serialization
-    )
+impl SprotError {
+    pub fn is_recoverable(&self) -> bool {
+        match *self {
```
cbiffle, line 322: Nit: you probably don't need the asterisk there thanks to match ergonomics.

idl/sprot.idol
```diff
@@ -63,71 +26,64 @@ Interface(
...
-                )
+            args: {
+                "target": "UpdateTarget",
             },
             reply : Result(
                 ok: "()",
-                err: CLike("SprotError"),
+                err: Complex("SprotError"),
             ),
+            encoding: Hubpack,
```
cbiffle, line 54: Thought: this operation is also likely idempotent for Idol's purposes.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -192,20 +191,21 @@ impl MgsCommon {
...
                         // can start the update process all over again.  We should
                         // be able to make incremental progress if there is some
                         // bug/condition that is degrading SpRot communications.
-                        ringbuf_entry!(Log::RotReset { err });
-                        Err(SpError::ComponentOperationFailed(err.into()))
+
+                        // TODO(AJS): We need to be able to expose all SprotErrors directly
+                        // and not just as u32s
+                        //ringbuf_entry!(Log::RotReset { err });
+                        Err(SpError::ComponentOperationFailed(9999))
```
cbiffle, line 208: nein nein nein nein ;-)

task/control-plane-agent/src/update/rot.rs
```diff
@@ -235,11 +241,17 @@ impl ComponentUpdater for RotUpdate {
                 match self.task.abort_update() {
                     // Aborting an update that hasn't started yet is fine;
                     // either way our caller is clear to start a new update.
-                    Ok(()) | Err(SprotError::UpdateNotStarted) => {
+                    Ok(())
+                    | Err(SprotError::Update(
+                        drv_sprot_api::UpdateError::UpdateNotStarted,
```
cbiffle, line 246: This is kind of annoying, maybe the abort operation on the update API should return Ok if the update is not started? I believe this would also make it idempotent.

drv/sprot-api/src/lib.rs
andrewjstone: I had this as an `&[u8]` originally but ran into lifetime errors on the SP side, since I was trying to hold onto the slice as returned from `hubpack::deserialize` and then mutably write into it on the next spi read. However, that was just an optimization. I can just recreate the slice after all the mutation is done, and I should do that.

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: > Alternatively, it looks like this is always called with a &[u8; MAX_RESPONSE_SIZE] -- if tx_buf's type were &[u8; MAX_RESPONSE_SIZE] instead of &[u8] then I think you couldn't misuse it in this way?

I think this is a good suggestion. However, this code is shared among request and response which have different max sizes. I can go ahead and use a const parameter on Msg to deal with that though, and will do so.

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: Will, do this was strictly habit.

drv/lpc55-sprot-server/src/main.rs
andrewjstone: Will do!

drv/sprot-api/src/lib.rs
andrewjstone: Precisely. I will clarify.

drv/sprot-api/src/lib.rs
andrewjstone: Ah great, thanks!

task/control-plane-agent/src/mgs_common.rs
andrewjstone: laughed out loud when I read this last night.

drv/sprot-api/src/lib.rs
andrewjstone: Looks like I only use it in one place: https://github.com/oxidecomputer/hubris/pull/1279/files/8ff567f9af1aa58ca16496341b93455b5e935924#diff-32036439343487dbfb550235d8d3ce0a3a92e8d6808dc98d3e72d10d2e9d9c89R374

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: I always forget I can map a variant as a function. Thanks!

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: Fixed by using an array. Good idea!

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: Fixed by not unwrapping at all and leaving a comment.

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: Fixed.

drv/sprot-api/src/lib.rs
andrewjstone: I changed it to a slice, but I"m getting problems as the slice is part of `self.tx_buf` which is borrowed mutably in the retry loop of `do_send_recv_retries`. I'm trying to work this out, because a slice is a much nicer thing to return here.

andrewjstone: One option is to put all the spi code: `do_send_recv`, etc... inside its own type or free functions. We had the former in prior code, but I didn't particularly like it.

andrewjstone: Fixed in [9f8bed9](https://github.com/oxidecomputer/hubris/pull/1279/commits/9f8bed9ad79b519fb73906e97bfa378665990bec)

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -422,6 +374,76 @@ impl<S: SpiServer> ServerImpl<S> {
...
+        &mut self,
+        tx_size: usize,
+        timeout: u32,
+        retries: u16,
+    ) -> Result<Response, SprotError> {
+        let mut attempts_left = retries;
+
+        loop {
+            // Increase the error count here. We'll decrease if we return successfully.
+            self.io.stats.rx_invalid = self.io.stats.rx_invalid.wrapping_add(1);
```
andrewjstone, line 388: Uh, this should move into the match right before `Response::unpack`

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -193,15 +203,13 @@ impl<S: SpiServer> ServerImpl<S> {
...
         // Increase the error count here. We'll decrease if we return successfully.
         self.stats.tx_errors = self.stats.tx_errors.wrapping_add(1);
 
-        let part1_len = ROT_FIFO_SIZE.min(tx_size);
-        let part1 = &self.tx_buf[..part1_len];
-
-        let part2_len = tx_size - part1_len;
-        let part2 = &self.tx_buf[part1_len..part1_len + part2_len];
+        let part1_len = ROT_FIFO_SIZE.min(tx_buf.len());
+        let part1 = &tx_buf[..part1_len];
```
cbiffle, line 211: FYI,
```rust
let (part1, part2) = tx_buf.split_at(part1_len);
```

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: Better yet, we don't have to do this dance anymore. Just put it in the error branch.

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: yoooo, nice

drv/sprot-api/src/lib.rs
andrewjstone: This was not actually fixed. I introduced a bug with an implicit copy. It appears that there may be a compiler bug that is causing the compiler to think that there is mutable and immutable references to the same slice when inside a loop. This appears to happen when the compiler can't tell that the loop terminates. A minimal reproducible example is here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=59c15df3d0a060a01086450d6bf1a780

I'm still working on solving it. Thanks to @cbiffle for the initial example that I was able to modify to reproduce the issue. @sunshowers has asked some rust compiler devs about it. We'll see what they come up with :)

task/control-plane-agent/src/update/rot.rs
andrewjstone: Fixed

drv/sprot-api/src/lib.rs
andrewjstone: Fixed the comment.

drv/sprot-api/src/lib.rs
andrewjstone: There can be deserialization errors on short reads, so I changed it to Deserialization.

drv/sprot-api/src/lib.rs
andrewjstone: fixed.

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -50,213 +54,115 @@ impl Handler {
...
-        let rx_msg = match rx_buf.parse() {
-            Ok(rxmsg) => rxmsg,
-            Err((header_bytes, msgerr)) => {
-                if msgerr == SprotError::NoMessage {
-                    // We were just returning a reply, so clocked out zeros
-                    // from the SP.
-                    return None;
-                }
-                ringbuf_entry!(Trace::ErrWithHeader(msgerr, header_bytes));
+    ) -> usize {
```
mkeeter, line 76: Nit: you could do the packing in just one place if you write this function as
```rust
    pub fn handle(
        &mut self,
        rx_buf: &[u8],
        tx_buf: &mut [u8; MAX_RESPONSE_SIZE],
        stats: &mut RotIoStats,
    ) -> usize {
        stats.rx_received = stats.rx_received.wrapping_add(1);
        let rsp_body = match Request::unpack(rx_buf) {
            Ok(request) => {
                ringbuf_entry!(Trace::Req {
                    protocol: rx_buf[0],
                    body_type: rx_buf[1]
                });
                self.handle_request(request, stats)
            }
            Err(e) => {
                ringbuf_entry!(Trace::Err(e));
                stats.rx_invalid = stats.rx_invalid.wrapping_add(1);
                Err(e.into())
            }
        };

        Response::pack(&rsp_body, tx_buf)
    }
```

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -149,63 +153,55 @@ enum IoError {
...
+        // gpio task when its not necessary. If the server crashes, and ROT_IRQ
+        // is already asserted, and we somehow get here, we'll just make one
+        // more idempotent call to the gpio server.
+        let mut rot_irq_asserted = false;
+        loop {
             sys_irq_control(notifications::SPI_IRQ_MASK, true);
 
-            if signal_reply {
+            if is_reply && !rot_irq_asserted {
+                rot_irq_asserted = true;
```
mkeeter, line 204: Should `rot_irq_asserted` be a member variable, instead of just being local to this function?

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -220,132 +216,136 @@ impl Io {
...
+        if bytes_sent == 0 || self.spi.can_tx() {
+            // This was a CSn pulse
+            self.stats.csn_pulses = self.stats.csn_pulses.wrapping_add(1);
+        } else {
+            self.check_for_tx_error();
         }
 
-        err.map_or(Ok(()), |e| Err(e))
+        ringbuf_entry!(Trace::SentBytes(bytes_sent));
+        self.prepare_for_next_request();
```
mkeeter, line 306: I'd suggest inlining `prepare_for_next_request` into this function – it's only ever called from here and does very specific things, but has a tempting-enough name that I'd imagine someone trying to call it in the future.

drv/sprot-api/src/lib.rs
```diff
@@ -3,64 +3,261 @@
...
+    }
+}
+
+/// A sprot Msg that flows between the SP and RoT
+///
+/// The message is parameterized by a `ReqBody` or `RspBody`.
+///
+/// Note that `MSG`s do not implement `Serialize`, `Deserialize`, or
+/// `SerializedSize`, as they need to calculate and place a CRC in the buffer.
+/// `Msg`s sometimes include a an offset into the buffer where a binary blob
```
mkeeter, line 73:
```suggestion
/// `Msg`s sometimes include an offset into the buffer where a binary blob
```

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -193,8 +148,8 @@ pub struct Io<S: SpiServer> {
 
 pub struct ServerImpl<S: SpiServer> {
     io: Io<S>,
-    tx_buf: [u8; BUF_SIZE],
-    rx_buf: [u8; BUF_SIZE],
+    tx_buf: [u8; MAX_REQUEST_SIZE],
```
mkeeter, line 151: Similar suggestion as above: it would be more stack-friendly to use `mutable_statics!` and put them into static memory.

andrewjstone:
This [MGS PR](https://github.com/oxidecomputer/management-gateway-service/pull/82) should be merged prior to this change.

andrewjstone:
Taking this out of draft as it seems fairly solid. I'd love another pass from @cbiffle and then I expect to merge it into Cliff's secure boot branch so we can flag day jointly.

drv/sprot-api/src/lib.rs
```diff
@@ -3,241 +3,243 @@
...
 }
 
-impl From<HandoffDataLoadError> for SprotError {
-    fn from(_: HandoffDataLoadError) -> Self {
-        SprotError::Stage0HandoffError
-    }
+/// The body of a sprot request
+#[derive(Clone, Serialize, Deserialize, SerializedSize)]
+pub enum ReqBody {
+    Status,
```
lzrd, line 194: We need a Status message that allows us to change protocol later.
What is the necessary information?
  - protocols supported.
  - buffer sizes?
  - Maybe ROM crc (not because this is the right place, but because NXP doesn't give us an explicit HW version and do we have a different API for that?
  talk more later.

  -

lzrd (commented):
some comments while scanning through this with Andrew.

drv/lpc55-sprot-server/src/main.rs
andrewjstone: To be clear, this was a bug I introduced in this refactoring, not an existing bug :)

drv/lpc55-sprot-server/src/handler.rs
andrewjstone: Nice! Fixed in [3ddb01b](https://github.com/oxidecomputer/hubris/pull/1279/commits/3ddb01b536d37432e9c62a391d4b708a2630ce23)

drv/lpc55-sprot-server/src/main.rs
andrewjstone: I like this better. Added in [3ddb01b](https://github.com/oxidecomputer/hubris/pull/1279/commits/3ddb01b536d37432e9c62a391d4b708a2630ce23)

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -220,132 +216,136 @@ impl Io {
...
-        if too_many_bytes_received {
-            self.stats.rx_protocol_error_too_many_bytes =
-                self.stats.rx_protocol_error_too_many_bytes.wrapping_add(1);
+        self.spi.ssd_clear();
+
+        // We clocked out at least an existing byte in the fifo, as we fill it on entry to
+        // this function.
+        if bytes_sent == 0 || self.spi.can_tx() {
+            // This was a CSn pulse
+            self.stats.csn_pulses = self.stats.csn_pulses.wrapping_add(1);
```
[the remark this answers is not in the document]
andrewjstone: It's not needed. I added a comment in [3ddb01b](https://github.com/oxidecomputer/hubris/pull/1279/commits/3ddb01b536d37432e9c62a391d4b708a2630ce23)

drv/lpc55-sprot-server/src/main.rs
andrewjstone: Done.

drv/sprot-api/src/lib.rs
andrewjstone: Fixed. Thanks.

drv/stm32h7-sprot-server/src/main.rs
andrewjstone: Great call. Fixed in [3ddb01b](https://github.com/oxidecomputer/hubris/pull/1279/commits/3ddb01b536d37432e9c62a391d4b708a2630ce23)

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -149,63 +153,55 @@ enum IoError {
 fn main() -> ! {
     let mut io = configure_spi();
 
-    let mut rx_buf = [0u8; BUF_SIZE];
-    let mut tx_buf = [0u8; BUF_SIZE];
+    let mut rx_buf = [0u8; MAX_REQUEST_SIZE];
+    let mut tx_buf = [0u8; MAX_RESPONSE_SIZE];
```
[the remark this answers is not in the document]
andrewjstone: Great call. Fixed in [3ddb01b](https://github.com/oxidecomputer/hubris/pull/1279/commits/3ddb01b536d37432e9c62a391d4b708a2630ce23)

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -220,132 +216,136 @@ impl Io {
...
+                self.spi.send_u8(b);
+            }
         }
 
-        if too_many_bytes_received {
-            self.stats.rx_protocol_error_too_many_bytes =
-                self.stats.rx_protocol_error_too_many_bytes.wrapping_add(1);
+        self.spi.ssd_clear();
+
+        // We clocked out at least an existing byte in the fifo, as we fill it on entry to
```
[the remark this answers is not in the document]
andrewjstone: Thanks! Fixed in [3ddb01b](https://github.com/oxidecomputer/hubris/pull/1279/commits/3ddb01b536d37432e9c62a391d4b708a2630ce23)

drv/sprot-api/src/lib.rs
andrewjstone: I made a few related changes:
 1. Status now contains just the current and min versions for each of the RoT and SP along with buffer sizes
 2. I added a long doc comment explaining how extending the protocol should work

andrewjstone:
There have a been a lot of changes here over the weekend: Most importantly, messages extensibility is now well documented, the status message contains minimal info required for bringing out of date machines into conformance, and 16-bit fifos are in use eliminating nearly all sprot errors.

task/dumper-api/src/lib.rs
```diff
@@ -7,17 +7,36 @@
...
 /// error codes are interpreted by the dump agent and turned into dump agent
-/// failures.  Ultimately, these error conditions are merely informative,
-/// but it is important that a particular error condition not be falsely
-/// reported; error conditions can be added to or eliminated from this enum,
-/// but the values for extant conditions should not be altered.
+/// failures.  
+///
+/// These errors are also serialized and passed up to MGS, and as such they
+/// should not be re-ordered. New errors may be added to the end, but if they
+/// are they should be added to the `From<DumperError> for GwDumperError` below
```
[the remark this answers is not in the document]
jgallagher: This sounds right to me. I think long-term we probably want to support taking dumps via MGS, but on the hubris side that would mean MGS would talk directly to `dump-agent` (using its UDP protocol) to do so.

Agreed that changing `DumperError` is a technically compatibility-breaking change in MGS but that making it now is fine, since we haven't deployed any SP builds that know about it (and there haven't been any subsequent changes yet).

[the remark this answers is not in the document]
andrewjstone: FYI, the reason I put `DumperError` in `gateway-messages` is because I was translating all `SprotError` variants to `gateway_messages::SpError`. However, this isn't strictly necessary. I suppose I can panic [here](https://github.com/oxidecomputer/hubris/pull/1279/files#diff-58d9279309315e9491ef32635decc0290285d1745708be04539ba3cd25a92d31R39) instead, which would prevent us from re-adding this error.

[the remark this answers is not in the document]
andrewjstone: Removed from gateway-messages in https://github.com/oxidecomputer/management-gateway-service/pull/85

I'll remove here in a follow up once that merges.

mkeeter (approved):

PR #1282, opened 2023-04-13
Include image name in archive zip filename

jgallagher:
This makes three changes, two of which only affect CI (and any following scripts / processes that use CI artifacts):

1. (Affects everyone) Hubris archive names now have a suffix of `-image-X`, where `X` is "default" for most images. E.g., what was once `build-psc-c.zip` is now `build-psc-c-image-default.zip`. This matters most for the RoT images; what was once `build-gimlet-rot-c.zip` (and you had to know what directory it was in and/or inspect its contents to tell if it was an A or B image) is now `build-gimlet-rot-c-image-a.zip`.
2. The CI artifacts for the RoT builds now contain both A and B hubris archives.
3. We no longer upload the artifacts built by the Windows runner (but we do still _run_ the builds on Windows). Since we were cross compiling, I believe (?) these should've been identical, and I don't think anyone was using them.


steveklabnik (approved):

steveklabnik:
>  I believe (?) these should've been identical, and I don't think anyone was using them.

Yep!

PR #1284, opened 2023-04-14
Single-task dumping

mkeeter:
This PR implements single-task dumping in Hubris, mediated by the supervisor task.  It includes three new APIs:

- Dump a single task
- Dump a subregion of memory from a single task
- Reinitialize dump areas starting at a particular index

(see https://github.com/oxidecomputer/humpty/pull/5 for the associated UDP types)

Single-task dumping is between the supervisor and the kernel, and doesn't require the RoT!

Most of this PR is plumbing of various kinds:

- UDP messages to `dump-agent` → Idol messages to Jefe
- Idol messages to `dump-agent` -> Idol messages to Jefe
- Idol messages to Jefe → actual work being done 

Within `jefe::dump`, `dump_task` splits into four functions:

- Setup (`dump_task_setup`)
- Run (`dump_task_run`)
- `dump_task`, which now uses the above
- The new `dump_task_region`

To be very clear – even though it may go without saying – tasks talking to `jefe` through these APIs **should not be able to crash the supervisor**.

The most suspicion is directed towards `dump_task_region`, because it specifies an arbitrary address + length.  To make sure the region is legal, `jefe` checks this region against the kernel's report of valid dump regions for the task (`kipc::get_task_dump_region`) and bails out early if it's invalid.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, branch = "single-task-dump", version = "0.1.1" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, branch = "single-task-dump", version = "0.1.2" }
 hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false }
 idol = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
 idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, branch = "single-task-dump", version = "0.1.2" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.2" }
 hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false }
 idol = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
 idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.1" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, branch = "single-task-dump", version = "0.1.1" }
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -119,7 +119,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.1" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, branch = "single-task-dump", version = "0.1.2" }
```
cbiffle, line 122: This line is probably a do-not-merge-in-this-condition situation, right?

task/dump-agent/src/udp.rs
```diff
@@ -22,7 +22,6 @@ enum Trace {
     DeserializeHeaderError(hubpack::Error),
     SendError(SendError),
     WrongVersion(u8),
-    Hi(u8, usize),
```
cbiffle, line 25: hiiiiii

task/jefe/src/dump.rs
```diff
@@ -75,27 +103,83 @@ pub fn claim_dump_area(base: u32) -> Result<DumpArea, DumpAgentError> {
...
     //
     // We need to claim a dump area.  Once it's claimed, we have committed
     // to dumping into it:  any failure will result in a partial or otherwise
     // corrupted dump.
     //
     let area = humpty::claim_dump_area(
         base,
-        DumpContents::SingleTask,
+        if full {
+            DumpContents::SingleTask
```
cbiffle, line 116: I admit a mild preference for exposing this enum in place of `bool` in the function signature, as I think it'd make it clearer.

task/jefe/src/dump.rs
```diff
@@ -119,56 +203,109 @@ pub fn dump_task(base: u32, task: usize) {
...
+
+    // We don't trust the caller; it may request to dump a region that isn't
+    // owned by this particular task!  To check this, we iterate over all of the
+    // valid dump regions and confirm that our desired region is within one of
+    // them.
+    let mut ndx = 0;
+    let mem = start..start + length;
+    let mut okay = false;
+    loop {
+        // This is Accidentally Quadratic; see the note in `dump_task`
```
cbiffle, line 262: Either having the kernel return an error for an out-of-range region (as we discussed in chat), or exposing a "validate region" entry point, would be good ways of eliminating this

task/jefe/src/main.rs
```diff
@@ -232,6 +232,48 @@ impl idl::InOrderJefeImpl for ServerImpl<'_> {
...
+                &mut self,
+                _msg: &userlib::RecvMessage,
+                task_index: u32,
+            ) -> Result<u8, RequestError<DumpAgentError>> {
+                // `dump::dump_task` doesn't check the task index, because it's
+                // normally called by a trusted source; we'll do it ourself.
+                if task_index == 0 {
+                    // Can't dump the supervisor
+                    return Err(DumpAgentError::NotSupported.into());
+                } else if task_index as usize > self.task_states.len() {
```
cbiffle, line 246: It seems like all of these comparisons should be `>=`

task/jefe/src/main.rs
```diff
@@ -232,6 +232,48 @@ impl idl::InOrderJefeImpl for ServerImpl<'_> {
...
+                    // Can't dump a non-existent task
+                    return Err(DumpAgentError::BadOffset.into());
+                }
+                dump::dump_task(self.dump_areas, task_index as usize)
+                    .map_err(|e| e.into())
+            }
+            fn dump_task_region(
```
cbiffle, line 253: (would like blank lines between functions, ideally)

cbiffle (commented):
I think there are some incorrect comparisons (noted below). Rest of this looks okay.

I know you're just moving existing undocumented `unsafe` blocks around, but if you understand humpty well enough to put a `Safety:` comment on them, I would be grateful.

Cargo.toml
mkeeter: Yup, thanks for the reminder

task/jefe/src/dump.rs
mkeeter: I'll do you one better and make a _new_ type, since `DumpContents` has other variants that aren't relevant here.

Done in c61c0635bf453f533061008cd54e60e5de1fe715

task/jefe/src/main.rs
mkeeter: Yes, fixed in bc20ab50e8f282549365ab9ddd3c132100513f36

task/jefe/src/main.rs
mkeeter: Me too, not sure where it went!

Fixed in d359897ef132ff95caf40d5ba659feb7fec4459e

task/dump-agent/src/udp.rs
mkeeter: 👋🏻

task/jefe/src/dump.rs
mkeeter: Agreed, that's going on my list for later cleanups

PR #1293, opened 2023-04-18
Add watchdog for QueueFull issue

mkeeter:
This watchdog panics the net task if a queue remains full for > 500 ms, to work around https://github.com/smoltcp-rs/smoltcp/issues/594

Tested on a bench with a hacked `udpbroadcast` to clog the pipes, and it seems to work fine.

### DIFF
--- a/task/net/src/server.rs
+++ b/task/net/src/server.rs
@@ -583,11 +583,25 @@ where
                 // queue isn't clogged.
                 let now = userlib::sys_get_timer().now;
                 if let Some(t) = self.first_queue_full[socket_index] {
-                    // Leave `self.first_queue_full` unchanged; it's already
-                    // populated with the time of the first QueueFull error.
+                    // Reset the queue by closing + reopening it.  This will
+                    // lose packets in the RX queue as well; they're collateral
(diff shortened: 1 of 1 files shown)


task/net/src/server.rs
```diff
@@ -559,9 +567,30 @@ where
...
+                // the task.
+                //
+                // Note that any successful packet send on this socket will
+                // reset `self.first_queue_full`, because it means that the
+                // queue isn't clogged.
+                let now = userlib::sys_get_timer().now;
+                if let Some(t) = self.first_queue_full[socket_index] {
+                    // Leave `self.first_queue_full` unchanged; it's already
+                    // populated with the time of the first QueueFull error.
+                    if now >= t + 500 {
```
jgallagher, line 588: Maaaaybe worth putting `500` into a named constant that's greppable? Since it's only used once it's not a very strong argument though.

jgallagher (approved):

cbiffle (approved):
Agree with John on the named constant, otherwise this hack looks great?

cbiffle (approved):
It is unfortunate that we need to do this, but this change is about the cleanest way of doing it I can imagine. Nice work.

Issue #1290, opened 2023-04-18
allow udprpc 


jclulow:

cbiffle:
Was thinking about this today as I looked for things to do, and I have a suggestion (which I also floated in chat):

I think we need to expose this operation in debug contexts, but I don't think we should do it through udprpc. Instead, I propose that we add it to `dump-agent`, morphing it into the beginning of a more general `debug-agent`.

There's a fair amount of shuffling and paperwork that'd be required to make this not grody -- for instance, the packet definitions spoken by dump-agent are currently in the `humpty` repo where they can be shared with Humility, as @mkeeter pointed out, but that's probably the wrong place for more general debug operations.

We'll also want to think about the relationship between `debug-agent` and `hiffy`. `hiffy` is already a debug agent, but one that's primarily controlled from SWD. This is not inherent, and having an agent that can run small programs is very handy in debug. So, maybe `debug-agent` would wind up with a HIF VM built in. TBD.

PR #1295, opened 2023-04-18
Allow udprpc to send `Jefe.request_reset`

jgallagher:
Prior to this PR, attempting to call `Jefe.request_reset` via `udprpc` would time out:

```
% cargo xtask humility app/gimletlet/app.toml  -- --ip fe80::0c1d:7dff:feef:9f1d%eth0 hiffy -c Jefe.request_reset
    Finished dev [optimized + debuginfo] target(s) in 0.21s
     Running `target/debug/xtask humility app/gimletlet/app.toml -- --ip 'fe80::0c1d:7dff:feef:9f1d%eth0' hiffy -c Jefe.request_reset`
humility: connecting to fe80::0c1d:7dff:feef:9f1d%eth0
humility hiffy failed: Resource temporarily unavailable (os error 11)
Error: humility failed
```

because jefe returned an access violation:

```
14 udprpc                       1   6 FAULT: reply fault: task id #0/gen0, reason AccessViolation (was: wait: reply from jefe/gen0)
```

After this PR, attempting to call `Jefe.request_reset` via `udprpc` _still_ times out from humility's point of view, but because the SP has actually reset and therefore does not respond to the request.

### DIFF
--- a/app/gimlet/rev-b.toml
+++ b/app/gimlet/rev-b.toml
@@ -34,7 +34,7 @@ spd = "jefe-state-change"
 [tasks.jefe.config.allowed-callers]
 set_state = ["gimlet_seq"]
 set_reset_reason = ["sys"]
-request_reset = ["hiffy", "control_plane_agent"]
+request_reset = ["hiffy", "control_plane_agent", "udprpc"]
 
 [tasks.net]
 name = "task-net"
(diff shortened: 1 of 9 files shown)


jgallagher:
This isn't the full support requested in #1290 but is better than nothing, I think.

mkeeter (approved):

PR #1297, opened 2023-04-18
[udprpc] Don't panic if we fail to send via `net`

jgallagher:


task/udprpc/src/main.rs
```diff
@@ -141,12 +141,28 @@ fn main() -> ! {
...
+                        SOCKET,
+                        meta,
+                        &tx_data_buf[0..(meta.size as usize)],
+                    ) {
+                        Ok(()) => break,
+                        // If our tx queue is full or `net` just restarted, just
+                        // retry our send; both of these should be ephemeral.
+                        Err(
+                            SendError::QueueFull | SendError::ServerRestarted,
+                        ) => continue,
```
cbiffle, line 155: This behavior is probably not right; it will just burn CPU on QueueFull. It would be more polite to wait for notification from the netstack that space is available.

task/udprpc/src/main.rs
```diff
@@ -141,12 +141,44 @@ fn main() -> ! {
...
+                        // finally succeed in sending we'll peel any queued
+                        // packets off our recv queue at the top of our main
+                        // loop.
+                        Err(SendError::QueueFull) => {
+                            sys_recv_closed(
+                                &mut [],
+                                notifications::SOCKET_MASK,
+                                TaskId::KERNEL,
+                            )
+                            .unwrap();
```
cbiffle, line 167: `unwrap_lite` would generate less code for a case that really shouldn't happen.

cbiffle (approved):

PR #1300, opened 2023-04-19
kern: add new `safe_copy_dma` operation.

cbiffle:
This will copy from (not to!) DMA-capable memory in a task.

Copying _to_ DMA-capable memory could be added pretty easily, but isn't required for what we're doing right now, so I have omitted it.


cbiffle:
Alrighty! This has been tested, gonna barrel towards merging

bcantrill (approved):
Looks great -- and verified that this works with my changes to dump DMA regions (which depends on this change).

PR #1301, opened 2023-04-20
Fix MGS branch

mkeeter:
This commit didn't make it into hubris#1294, for some reason.


andrewjstone:
FYI: I added this same fix in #1279 

I'd prefer to merge that in without rebasing again tomorrow. I think it's ready to go. I did some extra cleaning and documenting over the weekend so messages are truly extensible.

mkeeter:
Okay, I'll hold off and let you merge your branch instead!

mkeeter:
Done in #1279 instead

PR #1304, opened 2023-04-20
New fpga registers for debugging sequence issues

nathanaelhuffman:
- Added some new registers for capturing events when we fail to sequence
-  Forced to increase flash allocation of `drv-gimlet-seq-server` to next power of 2


nathanaelhuffman:
discussed doubling flash-size in chat, going to merge otherwise should be no impact to anyone.

Issue #1299, opened 2023-04-19
Fans spin up when transceiver is unplugged + replugged


mkeeter:
Quoth @rmustacc,

>The Paris sidecar attached to lurch seems very circumspect of me hotplugging a transceiver.
> That is I unplugged and replugged it in and thermal loop did the thing that we tell everyone else things shouldn't do where it suddenly jumped up and then very slowly revved down.
> It wasn't a dramatic immediate ear protection jump, but enough that everyone turned to look at me in the office kind of jump.

He unplugged + replugged transceiver 4, which we see in the ringbuf:
```
humility: attached to 0483:374e:0028001E4741500720383733 via ST-Link V3
humility: ring buffer drv_transceivers_server::__RINGBUF in transceivers:
 NDX LINE      GEN    COUNT PAYLOAD
   1  238        5        1 GotInterface(0xe, Sff8636)
   2  238        5        1 GotInterface(0xf, Sff8636)
   3  238        5        1 GotInterface(0x10, Sff8636)
   4  238        5        1 GotInterface(0x11, Sff8636)
   5  238        5        1 GotInterface(0x12, Sff8636)
   6  238        5        1 GotInterface(0x14, Sff8636)
   7  238        5        1 GotInterface(0x1d, Sff8636)
   8  238        5        1 GotInterface(0x1e, Sff8636)
   9  238        5        1 GotInterface(0x1f, Sff8636)
  10  345        5        4 TemperatureReadError(0x14, ImplError(0x4))
  11  589        5        1 ModulePresenceUpdate(0xe017e007)
  12  294        5        1 UnpluggedModule(0x4)
  13  589        5        1 ModulePresenceUpdate(0xe017e017)
  14  277        5        1 GetInterfaceError(0x4, ImplError(0x5))
  15  238        5        1 GotInterface(0x4, Sff8636)
   0  345        6        4 TemperatureReadError(0x4, ImplError(0x4))
```

The `ImplError` can be decoded from the front IO register guide:

> 0x04 = I2cAddressnack, 0x05 = I2cByteNack

It's odd that the device is talking on I2C enough to tell us it's an `SFF8636` module, but then has a handful of temperature read errors; the thermal loop is presumably spinning up because we have a new device with an unknown temperature, so it's in the BOOTING / "set fan speed to 100%" state.

Aaron-Hartwig:
>The ImplError can be decoded from the front IO register guide:
[...]

PR #1306, opened 2023-04-21
Fix FPGA QSFP module initialization, improve ringbuf

Aaron-Hartwig:
This is a step towards addressing the strange `ringbuf` entries noted in https://github.com/oxidecomputer/hubris/issues/1299.

This change pulls in the FPGA from this PR: https://github.com/oxidecomputer/quartz/pull/96. I'll update to latest CI artifacts once I merge that.

This PR pulls in an FPGA which will correctly re-enforce `t_init` when a module is removed and re-inserted even when `ModResetL` has been deasserted the whole time. Additionally, there is a small quality of life improvement in that I integrated the FPGA's port error register enum values that the SP would read into the RDL definition, so any changes to the register definition will automatically get picked up by the FPGA, SP, and docs. The other part of this change is modifying the ringbuf to support displaying those enums rather than their primitive.


drv/sidecar-front-io/sidecar_qsfp_x32_controller.adoc
```diff
@@ -320,7 +320,7 @@ STATUS_PORT0
 | Bits | SW Access | Name | Function
 |[7:5] |  | - | Reserved
 |[4] | r | BUSY | '1' if the bus is busy.
-|[3:0] | r | ERROR | 0x0 = NoError, 0x01 = NoModule, 0x02 = NoPower, 0x03 = PowerFault, 0x04 = I2cAddressnack, 0x05 = I2cByteNack
```
mkeeter, line 323: Are we intentionally removing the value list from the docs?  I often find it helpful!

drv/transceivers-server/src/main.rs
```diff
@@ -272,9 +274,18 @@ impl ServerImpl {
                         self.thermal_models[i] =
                             self.decode_interface(port, interface)
                     }
+                    Err(FpgaError::ImplError(e)) => {
+                        ringbuf_entry!(Trace::GetInterfaceError(
+                            i,
+                            Reg::QSFP::PORT0_STATUS::Encoded::from_u8(e)
+                                .unwrap()
```
mkeeter, line 281: I'm skeptical of this `unwrap()`.  If `FpgaError::ImplError` is _always_ intended to contain a valid `Encoded`, then we should make it contain that by construction (instead of `u8`); otherwise, we should probably be tolerant of a bad value here (logging something else to the ringbuf if decoding fails).

drv/transceivers-server/src/main.rs
```diff
@@ -335,14 +346,20 @@ impl ServerImpl {
...
+                // We failed to read a temperature :(
+                //
+                // This could be because someone unplugged the transceiver
+                // at exactly the right time, in which case, the error will
+                // be transient (and we'll remove the transceiver on the
+                // next pass through this function).
+                Err(FpgaError::ImplError(e)) => {
+                    ringbuf_entry!(Trace::TemperatureReadError(
+                        i,
+                        Reg::QSFP::PORT0_STATUS::Encoded::from_u8(e).unwrap()
```
mkeeter, line 358: Same here

drv/sidecar-front-io/sidecar_qsfp_x32_controller.adoc
mkeeter: Interesting, they remain in the HTML page, so I guess it's fine?

Aaron-Hartwig: Yeah, thanks to Nathanael we actually have a mechanism to embed enums into our register descriptions which are then used to generate both BSV and Rust types! It'll also keep the docs in sync with the code, so its a win all around. Here's where that exists: https://github.com/oxidecomputer/quartz/blob/589697de44b7e173ad32123a7ecabb81703ea9e4/hdl/boards/sidecar/qsfp_x32/QSFPModule/qsfp_modules_top.rdl#L113-L121

drv/transceivers-server/src/main.rs
Aaron-Hartwig: Yeah, you are right. This is too aggressive. I'll handle the returned `Option<u8>`.

mkeeter (approved):

PR #1307, opened 2023-04-21
task dumping should include DMA regions

bcantrill:
Note that this also includes `memory.toml`, which is necessary to be able to disambiguate external DMA'd regions from non-external DMA'd regions (which in turn is required to prevent a system dump from trying to dump the dump areas itself).


sys/kern/src/descs.rs
```diff
@@ -128,7 +128,6 @@ impl RegionDesc {
 
         ratts.contains(RegionAttributes::WRITE)
             && !ratts.contains(RegionAttributes::DEVICE)
-            && !ratts.contains(RegionAttributes::DMA)
```
mkeeter, line 131: This is going to be a problem for `dump_agent`: the extern SRAM regions are mapped as `READ | WRITE | DMA`, so if `dump_agent` crashes or we request a task dump, it will dump the dump regions into themselves.

I tested this by killing `dump_agent` with `jefe`.  Sure enough, this task dump is _too big_:
```
AREA TASK                  TIME       SIZE
   0 dump_agent            6200       225736
```

Previously, it was much more reasonable:
```
AREA TASK                  TIME       SIZE
   0 dump_agent            4900       4664
```

task/jefe/build.rs
```diff
@@ -142,5 +148,30 @@ fn output_dump_areas(out: &mut std::fs::File) -> Result<()> {
...
 
+    writeln!(
+        out,
+        r##"
+///
+/// Function to determine if an address/length pair is contained within a
+/// dump area, short-circuiting in the common case that it isn't.  (Note that
+/// this will only check for containment, not overlap.)
+///
+pub(crate) fn in_dump_area(address: u32, length: u32) -> bool {{
```
mkeeter, line 159: Nit: does this actually need to be generated?  It looks like it could go into `jefe/src/dump.rs` without any issues.

bcantrill: I auto-generated it just to allow the short-circuiting min/max check to be done. Clarified this in the comment.

sys/kern/src/descs.rs
bcantrill: Fixed this by having Jefe refuse to dump anything that is itself in the dump areas.

mkeeter (approved):

PR #1320, opened 2023-04-28
Document and tidy Gimlet PowerStates

cbiffle:
The main goal of this change was to have some comments on the enum variants, but while attempting to reverse engineer the comments, I noticed that A2PlusMono is unused and removed it rather than documenting it.

### DIFF
--- a/drv/gimlet-state/src/lib.rs
+++ b/drv/gimlet-state/src/lib.rs
@@ -13,12 +13,25 @@ use zerocopy::AsBytes;
 #[derive(Copy, Clone, Debug, FromPrimitive, PartialEq, Eq, AsBytes)]
 #[repr(u8)]
 pub enum PowerState {
+    /// Initial A2 state where the SP and most associated circuitry is powered.
     A2 = 1,
-    A2PlusMono = 2,
+    /// A2 substate where we've turned on the fan hotplug controller.
     A2PlusFans = 3,
+    /// Intermediate A1 state on the way toward A0. This corresponds to the
+    /// system-wide notion of A1 and currently has no substates.
     A1 = 4,
+    /// Initial A0 state: the system-wide A0 domain is on, but we have not
+    /// turned on any of the subdomains within A0 (below).
     A0 = 5,
+    /// A0 with the NIC hotplug controller enabled. This is the state we expect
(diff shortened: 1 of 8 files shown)


drv/gimlet-state/src/lib.rs
```diff
@@ -13,12 +13,25 @@ use zerocopy::AsBytes;
...
-    A2PlusMono = 2,
+    /// A2 substate where we'e turned on the fan hotplug controller.
     A2PlusFans = 3,
+    /// Intermediate A1 state on the way toward A0. This corresponds to the
+    /// system-wide notion of A1 and currently has no substates.
     A1 = 4,
+    /// Initial A0 state: the system-wide A0 domain is on, but we have not
+    /// turned on any of the subdomains within A0 (below).
     A0 = 5,
+    /// A0 with the disk hotplug controllers enabled. This is the state we
```
rmustacc, line 26: This state actually only corresponds to when the T6 is active. There are basically 13 independent A0+HP domains because each of the 10x U.2, 2x M.2 and T6 are independent. This really only covers when we poke the sequencer about the T6 after directed to by the hotplug controller GPIO (controlled by the SMU and toggled by a write to a PCIe slot controller).

cbiffle: Thanks! And I assume we don't model the disks here because (1) the SP doesn't really care and (2) they can be individually powered on/off at runtime by the host OS.

drv/gimlet-state/src/lib.rs
```diff
@@ -13,12 +13,25 @@ use zerocopy::AsBytes;
 #[derive(Copy, Clone, Debug, FromPrimitive, PartialEq, Eq, AsBytes)]
 #[repr(u8)]
 pub enum PowerState {
+    /// Initial A2 state where the SP and most associated circuitry is powered.
     A2 = 1,
-    A2PlusMono = 2,
+    /// A2 substate where we'e turned on the fan hotplug controller.
```
mkeeter, line 18:
```suggestion
    /// A2 substate where we've turned on the fan hotplug controller.
```

cbiffle: WEEEEEE

PR #1323, opened 2023-05-01
Remove switch-default-image MGS handler.

jgallagher:
We can use the existing "component active slot" messages the same way host flash does.

Depends on https://github.com/oxidecomputer/management-gateway-service/pull/86.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,7 +112,9 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 
 # Oxide forks and repos
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -112,7 +112,9 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 
 # Oxide forks and repos
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+#gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+# TODO revert back to main before merging
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"], branch = "remove-switch-default-image" }
```
jgallagher, line 117: Note: Fix before merging

andrewjstone (approved):

Cargo.toml
jgallagher: fixed

Issue #1048, opened 2023-01-04
Monorail port 42 is mistakenly listed as DOWN


mkeeter:
On a Sidecar, I notice that port 42 is marked down from the perspective of the PHY.
```
PORT | MODE    SPEED  DEV     SERDES  LINK |   PHY    MAC LINK  MEDIA LINK
-----|-------------------------------------|-------------------------------
 40  | QSGMII  100M   1G_16   6G_14   up   | VSC8504  up        down
 41  | QSGMII  100M   1G_17   6G_14   up   | VSC8504  up        down
 42  | QSGMII  100M   1G_18   6G_14   up   | VSC8504  down      down
 43  | QSGMII  100M   1G_19   6G_14   up   | VSC8504  up        down
```

Dumping the registers, I see that there are two differences:
- 17E3 (MAC SerDes PCS Status) is 0x0405 on port 41, and 0x0404 on port 42
- 20E3 (MAC SerDes Status) is 0x6000 on port 41 and 0x2000 on port 42

The missing bit in both cases is "MAC comma detect" (bit 0 in 17E3; bit 14 in 20E3)

However, the link and sync status bits are all set.

PR #1327, opened 2023-05-01
Add API to reinitialize VSC7448

mkeeter:
This is a potential workaround for bf_sde#27; the idea would be to reset the entire Monorail switch + PHYs if the 10G link went down persistently.

I'd like to see it in action before installing it automatically as a watchdog, so this PR just adds an API endpoint.  Next time we encounter the issue in the wild, we can hit the endpoint and see if things resolve.

In addition, there's a bit of cleanup:

- Moves disabling `VAUI_CHANNEL_CFG.LANE_SYNC_ENA` to _before_ the VSC8504 is brought up.  In benchtop testing, this seems to fix https://github.com/oxidecomputer/hubris/issues/1048 , which is _plausible_ if not obvious.
- Make `Monorail` API use `drv_monorail_api::*` types, to avoid an ambiguous decoding in Humility
- Remove the Monorail BSP for the VSC7448 dev kit; the `app-vsc7448.toml` was removed a while back so this is dead code (🪦)


drv/vsc7448/src/lib.rs
```diff
@@ -366,11 +376,7 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
     ///
     /// Takes the REFCLK frequency, as well as an optional frequency for
     /// REFCLK2 (used to configure the PLL boost).
```
jgallagher, line 378: This part of the comment should presumably move to `new()`?

task/monorail-server/src/bsp/sidecar_bc.rs
```diff
@@ -299,6 +309,7 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
         // on GPIO_47.
         const COMA_MODE_GPIO: u32 = 47;
 
+        //
```
jgallagher, line 312: Stray addition or intentional formatting?

jgallagher (approved):

task/monorail-server/src/bsp/sidecar_bc.rs
mkeeter: Stray addition, fixed!

drv/vsc7448/src/lib.rs
mkeeter: Yup, done

PR #1329, opened 2023-05-02
Use `write_read_reg` in Renesas blackbox functions

mkeeter:
This prevents us from being interrupted; it's very unlikely to matter, but might as well do this right.


mkeeter:
closed in favor of #1340

Issue #1332, opened 2023-05-03
Need Sidecar's system LED controllable via MGS


Aaron-Hartwig:

Aaron-Hartwig:
Yup! Just opened this as a reminder for my self since I did not plan to address it in #1333.

Aaron-Hartwig:
closed by #1337

Issue #1167, opened 2023-02-23
Need transceiver-message protocol support for controlling front IO LEDs


Aaron-Hartwig:
Companion for https://github.com/oxidecomputer/transceiver-control/issues/62, we will need to be able to take the messages and do the required LED controller manipulation.

PR #1333, opened 2023-05-03
Move transceiver LED control to the network

Aaron-Hartwig:
This change is to support the control of the transceiver port LEDs over the network (see #1167). It does not implement control of the system LED via MGS (#1332).

Summary of changes:
* Support `LedState` and `SetLedState` `transceiver-messages`
* Adds some state to the `transceiver-server` to keep track of the state of LEDs, as well as functions to determine on vs off.
* Updates the `leds` module to use stronger typing, similar to the `transceivers` module
* Adds a compile-time fix for a hardware bug on Rev B QSFP Boards which swapped LEDs for 16<->17 and 18<->19
* Add a bunch of commands to the Idol interface to adjust LED state as well as modify brightness (via current and/or PWM). These mostly to support benchtop activity and do not need to stick around.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,5 +128,5 @@ sprockets-common = { git = "https://github.com/oxidecomputer/sprockets.git", def
 sprockets-rot = { git = "https://github.com/oxidecomputer/sprockets.git", default-features = false }
 tlvc = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
 tlvc-text = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
-transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false, branch = "led-control" }
+transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false }
 vsc7448-pac = { git = "https://github.com/oxidecomputer/vsc7448", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,5 +128,5 @@ sprockets-common = { git = "https://github.com/oxidecomputer/sprockets.git", def
 sprockets-rot = { git = "https://github.com/oxidecomputer/sprockets.git", default-features = false }
 tlvc = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
 tlvc-text = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
-transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false }
+transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false, branch = "led-control" }
 vsc7448-pac = { git = "https://github.com/oxidecomputer/vsc7448", default-features = false }
--- a/drv/sidecar-front-io/src/leds.rs
+++ b/drv/sidecar-front-io/src/leds.rs
@@ -6,6 +6,19 @@ use drv_i2c_api::I2cDevice;
 use drv_i2c_devices::pca9956b::{Error, LedErr, Pca9956B};
 use drv_transceivers_api::NUM_PORTS;
 
+/// Leds
+///
+/// `controllers`: two PCA9956B devices, each of which control half of  the LEDs
+/// on the front-io board.
+///
+/// `current`: a u8 to be written into the IREFALL register on the PCA9956Bs.
+/// From the PCA9956B datasheet, the calculus is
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -128,5 +128,5 @@ sprockets-common = { git = "https://github.com/oxidecomputer/sprockets.git", def
 sprockets-rot = { git = "https://github.com/oxidecomputer/sprockets.git", default-features = false }
 tlvc = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
 tlvc-text = { git = "https://github.com/oxidecomputer/tlvc", default-features = false }
-transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false }
+transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false, branch = "led-control" }
```
Aaron-Hartwig, line 131: Need https://github.com/oxidecomputer/transceiver-control/pull/67 to land first

drv/sidecar-front-io/src/leds.rs
```diff
@@ -1,11 +1,15 @@
...
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
+use super::transceivers::{LogicalPort, LogicalPortMask};
 use drv_i2c_api::I2cDevice;
 use drv_i2c_devices::pca9956b::{Error, LedErr, Pca9956B};
+use drv_transceivers_api::NUM_PORTS;
 
 pub struct Leds {
     controllers: [Pca9956B; 2],
+    current: u8,
```
mkeeter, line 11: We could use a docstring here about `current` and `pwm` units

drv/sidecar-front-io/src/leds.rs
```diff
@@ -319,19 +394,18 @@ impl Leds {
...
-        let sys_output: usize = sys_led_loc.output as usize;
-        summary.system_led_err = match sys_led_loc.controller {
-            LedController::Left => left_errs.errors[sys_output],
-            LedController::Right => right_errs.errors[sys_output],
+        summary.system_led_err = match SYSTEM_LED.controller {
+            LedController::Left => left_errs.errors[SYSTEM_LED.output as usize],
+            LedController::Right => {
+                right_errs.errors[SYSTEM_LED.output as usize]
+            }
         };
```
mkeeter, line 409: you could avoid overflowing here with something like
```suggestion
        summary.system_led_err = match SYSTEM_LED.controller {
            LedController::Left => &left_errs,
            LedController::Right => &right_errs,
        }
        .errors[SYSTEM_LED.output as usize];
```

(the same trick is applicable elsewhere in the file; up to you how widely you want to use it)

drv/transceivers-server/src/main.rs
```diff
@@ -70,13 +72,23 @@ ringbuf!(Trace, 16, Trace::None);
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// 0-31 for port LEDs, 32 is System LED
```
mkeeter, line 75: This comment is confusing here, because it implies there are 33 items in the array below (but `NUM_PORTS` is only 32, and the system LED is actually stored separately in `system_led_state`).

drv/sidecar-front-io/src/leds.rs
Aaron-Hartwig: Good recommendation. I'd had this below near the `DEFAULT_LED_{CURRENT,PWM}` constants for this, but in this PR I made them programmable so it makes sense to pull that text up to this struct.

drv/transceivers-server/src/main.rs
Aaron-Hartwig: Yeah that is me missing cleanup from a prior implementation. Sorry for missing that!

bnaecker (approved):
Looks great! 🚢

drv/transceivers-server/src/udp.rs
```diff
@@ -833,4 +878,23 @@ impl ServerImpl {
...
+
+    fn get_led_state_response(
+        &mut self,
+        modules: LogicalPortMask,
+        out: &mut [u8],
+    ) -> (usize, ModuleResultNoFailure) {
+        let mut led_state_len = 0;
+        for led_state in modules.to_indices().map(|m| self.get_led_state(m)) {
+            let led_state_size =
+                hubpack::serialize(&mut out[led_state_len..], &led_state)
```
mkeeter, line 890: Do we check elsewhere to make sure it's impossible to overflow the buffer here?  If so, add a comment explaining that's why the `unwrap` is okay.

mkeeter (approved):

PR #1339, opened 2023-05-04
Makes Sidecar's system and transceiver LEDs blink in phase

Aaron-Hartwig:
This should've been done out of the gate, but I recently reread [RFD 77](https://rfd.shared.oxide.computer/rfd/0077#_phasing) and realized this should be done. Haven't had a chance to visually test it yet though.

### DIFF
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -85,9 +85,8 @@ struct ServerImpl {
     led_error: FullErrorSummary,
     leds_initialized: bool,
     led_states: LedStates,
-    led_on: LogicalPortMask,
+    blink_on: bool,
     system_led_state: LedState,
-    system_led_on: bool,
 
     /// Handle to write thermal models and presence to the `thermal` task
     thermal_api: Thermal,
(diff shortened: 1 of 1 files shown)


drv/transceivers-server/src/main.rs
```diff
@@ -166,17 +165,16 @@ impl ServerImpl {
...
-            LedState::On => self.system_led_on = true,
-            LedState::Blink => self.system_led_on = !self.system_led_on,
-            LedState::Off => self.system_led_on = false,
-        }
-        if let Err(e) = self.leds.update_system_led_state(self.system_led_on) {
+        if let Err(e) = match self.system_led_state {
+            LedState::On => self.leds.update_system_led_state(true),
+            LedState::Blink => self.leds.update_system_led_state(self.blink_on),
+            LedState::Off => self.leds.update_system_led_state(false),
+        } {
```
jgallagher, line 173: Tiny nit - this might be slightly clearer if we broke it up into two steps; something like

```rust
let system_led_on = match self.system_led_state {
    LedState::On => true,
    LedState::Blink => self.blink_on,
    LedState::Off => false,
};
if let Err(e) = self.leds.update_system_led_state(system_led_on) {
```

jgallagher (approved):

PR #1340, opened 2023-05-05
Add Idol APIs for PMBus operations

mkeeter:
These are building blocks that allow us to
- Run `humility pmbus` over the network (https://github.com/oxidecomputer/humility/pull/382)
- Switch to fixed-PWM mode for debugging (pending implementation in Humility)

### DIFF
--- a/drv/i2c-api/src/lib.rs
+++ b/drv/i2c-api/src/lib.rs
@@ -552,9 +552,11 @@ impl I2cDevice {
     }
 
     ///
-    /// Like `write_read_reg`, but instead of returning a value, reads as many
-    /// bytes as the device will send into a specified slice, returning the
-    /// number of bytes read.
+    /// Performs a write followed by an SMBus block read (in which the first
+    /// byte returned from the device contains the total number of bytes to
+    /// read) into the specified buffer, returning the total number of bytes
+    /// read.  Note that the byte count is only returned from the function; it
+    /// is *not* present as the payload's first byte.
     ///
     /// The write and read are not performed as a single I2C transaction (that
     /// is, it is not a repeated start) -- but the effect is the same in that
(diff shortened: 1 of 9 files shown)


drv/i2c-api/src/lib.rs
```diff
@@ -32,6 +32,21 @@ use userlib::*;
...
+    /// operations.
+    ///
+    /// We don't need a special way to perform block writes, because they can be
+    /// constructed by the caller without cooperation from the driver.
+    /// Specifically, the caller can construct the array `[reg, size, data[0],
+    /// data[1], ...]` and pass it to a normal `WriteRead` operation.
+    ///
+    /// If we encounter a device which requires multiple block reads in a row
+    /// without interruption, this logic would not work, but that would be a
+    /// very strange device indeed.
```
bcantrill, line 49: Feels like a challenge to strange device makers everywhere!

idl/power.idol
```diff
@@ -98,5 +98,154 @@ Interface(
...
+            reply: Result(
+                ok: "()",
+                err: CLike("ResponseCode"),
+            ),
+            idempotent: true,
+        ),
+
+        // Low-level read/write APIs
+        //
+        // In all of these APIs, `index` is the index of the rail within
```
bcantrill, line 117: We now have (I think?) at least three different ways of specifying a rail via Idol (and I personally have been very much part of the problem here); at some point, it would be really nice to consolidate these down to a single, canonical way of specifying this.

bcantrill (approved):
Looks great. Would love to consolidate how rails are specified, but that can also be done over time.  And we now have many to pick from! ;)

idl/power.idol
mkeeter: Agreed, I opened https://github.com/oxidecomputer/hubris/issues/1342 to track.

PR #1341, opened 2023-05-05
Rename root of trust firmware

### LINKED ISSUE #1336: `gimlet-rot-c` image is increasingly misnamed, should be renamed
Currently we're building a single RoT image called `gimlet-rot-c` and putting it on, well, everything. This has confused people several times:
- In other cases when we have an image ending in `-c` it is not the right image for a D rev board
- This image is also the correct one to use on Sidecar B and later, and PSC B and later, but the only place that's indicated is in an auxiliary readme in `app/sidecar`.

That said, we're using the same image in each case because, in each case, both the RoT's responsibilities and the electrical circuit surrounding the RoT are identical. So, reusing the image is good! We just need to come up with a better name for it.

IMO, what this actually is is the Oxide Production Generation 1 Root Of Trust, and can work on any board so designed. So we could rename it to something like `oxide-rot-1` and resolve this. Or maybe there's a better idea.

@jgallagher @mkeeter

mkeeter:
I'm not sure if this will break any other tooling (permission-slip, MGS, etc), so I'm tagging a wide audience on this PR!

Closes #1336


mkeeter:
This will require updates to scripts in facade, e.g. [here](https://github.com/oxidecomputer/facade/blob/main/scripts/_gimlet_rot.sh)

jclulow:
> This will require updates to scripts in facade, e.g. [here](https://github.com/oxidecomputer/facade/blob/main/scripts/_gimlet_rot.sh)

Thanks for the heads up.  I'll deal with it the next time we pull things into the dock!

jgallagher:
> I'm not sure if this will break any other tooling (permission-slip, MGS, etc), so I'm tagging a wide audience on this PR!

I don't believe MGS is affected by this. :+1: from me

PR #1349, opened 2023-05-11
Implement LED blinking

mkeeter:
Nothing too controversial here, I hope!

### DIFF
--- a/drv/user-leds/src/main.rs
+++ b/drv/user-leds/src/main.rs
@@ -26,13 +26,22 @@
 //! Toggles an LED by index.
 //!
 //! Request message format: single `u32` giving LED index.
+//!
+//! ## `led_blink` (4)
+//!
+//! Sets an LED to blink, specifying the LED by index
+//!
+//! Request message format: single `u32` giving LED index.
 
 #![no_std]
 #![no_main]
 
 use drv_user_leds_api::LedError;
+use enum_map::EnumMap;
 use idol_runtime::RequestError;
 use userlib::*;
+
+const BLINK_INTERVAL: u64 = 500;
 
 cfg_if::cfg_if! {
     // Target boards with 4 leds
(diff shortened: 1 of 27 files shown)


drv/user-leds/src/main.rs
```diff
@@ -96,6 +108,7 @@ impl idl::InOrderUserLedsImpl for ServerImpl {
         index: usize,
     ) -> Result<(), RequestError<LedError>> {
         let led = Led::from_usize(index).ok_or(LedError::NotPresent)?;
+        self.blinking[led] = false;
```
jgallagher, line 111: Should this (and the two subsequent methods) check for any blinking leds and then set the timer back to `None` if this was the last blinker? I see the check in main that calls either `dispatch` or `dispatch_n`, but couldn't tell you off the top of my head what happens if we turn off a blinking LED (and therefore will call `dispatch`) while the timer is scheduled to notify us momentarily.

jgallagher (approved):

drv/user-leds/src/main.rs
mkeeter: I thought about that, but I think it's fine to have the timer fire (once) without any LEDs blinking.

(what will happen is: in `handle_notification`, no toggling will occur, and the timer will not be re-enabled)

mkeeter: Hmmm, good point.  I think this could leave a timer notification waiting in the kernel if you do things in exactly the right order.  I don't think it's a big deal, but I'll fix it in a follow-up PR.

jgallagher: I think an alternative to turning the timer off here is to always call `dispatch_n` in `main()`? Then we always accept timer notifications, but `handle_notification` will effectively ignore it if all blinking has been disabled.

Issue #1162, opened 2023-02-21
Sidecar Fan HSC work


Aaron-Hartwig:
This will be addressed alongside https://github.com/oxidecomputer/quartz/issues/71

The FPGA will be exposing bits for enable, power good, and power good timed out for each HSC. It will also expose a bit for presence and LED control for the Sidecar Fan VPD board.

When presence is detected, the FPGA will reflect that and the SP should set the HSC enable. When presence is lost, the FPGA will automatically clear the HSC enable and ignore attempts to set the bit from the SP until fan presence is restored.

@arjenroodselaar If I'm missing anything here or if you have more ideas please let me know!

Aaron-Hartwig:
I'll also note that I'd like the fan control PWM signal to be at 0V when a fan is not present, so I'll need to validate that.

Aaron-Hartwig:
I have a first pass of FPGA/SP code written that I haven't been able to test yet. I'll be able to test that when I return next week so I expect to be able to close this out within a few days once I'm back.

Aaron-Hartwig:
Closed by #1352

PR #1352, opened 2023-05-11
Add fan module presence detection

Aaron-Hartwig:
This PR addresses #1162.

Change summary:
- Pull in a new Sidecar mainboard FPGA design (staged as https://github.com/oxidecomputer/quartz/pull/79) which supports features we need (summarized [here](https://github.com/oxidecomputer/quartz/issues/71#issuecomment-1434879396))
- Adds a new `fan_modules` module to `sidecar-mainboard-controller` to wrap the fan module presence and LED logic
- Adds ability to get presence, request fan module LED state changes, and fan hot-swap control to the `sidecar-seq-server`. Additionally, getting updated fan module presence, updating LEDs, and hot-swap control are added to the core notification loop.
- The `thermal` task operates under the assumption that all fans would be present during normal operation. That is true for Gimlet, but Sidecar's fans can be added/removed while the unit is in the rack. To support this, I pulled the fans out of the individual BSPs (which are immutable) and put them in the `ThermalControl` struct. I also made the fans an `Option<SensorId>` to support the notion that a fan may not always be present. The Gimlet BSP will always respond with all fans being `Some(...)`, while Sidecar's BSP will query the `Sequencer` to reflect updated presence. All this effort is to be able to drive the PWM of the `MAX31790` low when a fan is not present to keep signals from being energized on an exposed interface.


Aaron-Hartwig:
@mkeeter - would you be able to provide some early review on this code? I touched how `thermal` implemented fans, so that may be of particular interest to you. I have not been able to test this yet as I'd like to do it when I have hands on physical hardware. I'm out next week, but when I return I'll have access to a Sidecar and will test this. Given that this is untested and I'm out, this could be reviewed at your leisure. I won't be able to address feedback until I'm back.

drv/sidecar-mainboard-controller/src/fan_modules.rs
```diff
@@ -0,0 +1,154 @@
...
+use bitfield::bitfield;
+use drv_fpga_api::{FpgaError, FpgaUserDesign, WriteOp};
+use userlib::FromPrimitive;
+use zerocopy::{AsBytes, FromBytes};
+
+bitfield! {
+    #[derive(Copy, Clone, PartialEq, Eq, FromPrimitive, AsBytes, FromBytes)]
+    #[repr(C)]
+    pub struct FanModuleStatus(u8);
+    pub enable, set_enable: 0;
```
mkeeter, line 15: This kinda duplicates information in the `Reg::FAN0_STATE` array of constants.  We _could_ use those constants in the bitfield:
```rust
use Reg::FAN0_STATE;
bitfield! {
    #[derive(Copy, Clone, PartialEq, Eq, FromPrimitive, AsBytes, FromBytes)]
    #[repr(C)]
    pub struct FanModuleStatus(u8);
    pub enable, set_enable: FAN0_STATE::ENABLE.trailing_zeros() as usize;
    pub led, set_led: FAN0_STATE::LED.trailing_zeros() as usize;
    pub present, _: FAN0_STATE::PRESENT.trailing_zeros() as usize;
    pub power_good, _: FAN0_STATE::PG.trailing_zeros() as usize;
    pub power_fault, _: FAN0_STATE::POWER_FAULT.trailing_zeros() as usize;
    pub power_timed_out, _: FAN0_STATE::PG_TIMED_OUT.trailing_zeros() as usize;
}
```
This is probably more robust, but is quite wordy, so I'm not sure.

drv/sidecar-mainboard-controller/src/fan_modules.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub power_fault, _: 4;
+    pub power_timed_out, _: 5;
+}
+
+/// Each fan module contains two individually controlled fans
+pub const NUM_FAN_MODULES: usize = 4;
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum FanModuleLedState {
+    Off = 0,
```
mkeeter, line 28: Do we rely on these explicit values anywhere?  If they're used as part of FPGA communication, then there should be a note to that effect; otherwise, I'd suggest removing them.

drv/sidecar-mainboard-controller/src/fan_modules.rs
```diff
@@ -0,0 +1,154 @@
...
+    ///
+    /// Additionally, update internal presence state with the latest status.
+    pub fn get_status(
+        &mut self,
+    ) -> Result<[FanModuleStatus; NUM_FAN_MODULES], FpgaError> {
+        let status: [FanModuleStatus; NUM_FAN_MODULES] =
+            self.fpga.read(Addr::FAN0_STATE)?;
+        for (module, status) in status.iter().enumerate() {
+            self.presence[module] = status.present();
+        }
```
mkeeter, line 61: Nit: you could avoid this loop with
```suggestion
        self.presence = status.map(|s| s.present());
```

drv/sidecar-mainboard-controller/src/fan_modules.rs
```diff
@@ -0,0 +1,154 @@
...
+            WriteOp::BitSet,
+            Addr::FAN0_STATE as u16 + idx as u16,
+            Reg::FAN0_STATE::ENABLE,
+        )?;
+        Ok(())
+    }
+
+    /// Disable the HSC for a fan module
+    ///
+    /// The FPGA will automatically disable the HSC if a fan module is removed.
```
mkeeter, line 90: This docstring seems less useful here.

drv/sidecar-mainboard-controller/src/fan_modules.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    // private function to turn an LED off
+    fn led_off(&self, idx: u8) -> Result<(), FpgaError> {
+        self.fpga.write(
+            WriteOp::BitClear,
+            Addr::FAN0_STATE as u16 + idx as u16,
+            Reg::FAN0_STATE::LED,
+        )?;
+        Ok(())
+    }
```
mkeeter, line 153: Nit: I'd recommend a single `led_set` function to avoid duplicating the call to `fpga.write`; this also simplifies the case statement here.
```suggestion
                FanModuleLedState::Blink => {
                    self.led_set(module as u8, blink_on)?
                }
            }
        }
        Ok(())
    }

    // private function to turn an LED on
    fn led_on(&self, idx: u8) -> Result<(), FpgaError> {
        self.led_set(idx, true)
    }

    // private function to turn an LED off
    fn led_off(&self, idx: u8) -> Result<(), FpgaError> {
        self.led_set(idx, false)
    }

    // private function to turn an LED off
    fn led_set(&self, idx: u8, on: bool) -> Result<(), FpgaError> {
        self.fpga.write(
            if on {
                WriteOp::BitSet
            } else {
                WriteOp::BitClear
            },
            Addr::FAN0_STATE as u16 + idx as u16,
            Reg::FAN0_STATE::LED,
        )?;
        Ok(())
    }
```

drv/sidecar-seq-api/src/lib.rs
```diff
@@ -44,4 +50,58 @@ pub enum TofinoSequencerPolicy {
     RestartOnFault = 2,
 }
 
+#[derive(
+    Copy, Clone, Debug, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
+)]
+#[repr(C)]
```
mkeeter, line 56: This `repr` seems unnecessary, since we're always serializing the value with `hubpack`

drv/sidecar-seq-api/src/lib.rs
```diff
@@ -44,4 +50,58 @@ pub enum TofinoSequencerPolicy {
...
+pub struct FanModulePresence(pub [bool; NUM_FAN_MODULES]);
+
+/// Four fan modules exist on sidecar, each with two fans.
+///
+/// The SP applies control at the individual fan level. Power control and
+/// status, module presence, and module LED control exist at the module level.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, FromPrimitive, AsBytes)]
+#[repr(u8)]
+pub enum FanModuleIndex {
```
mkeeter, line 65: After a bit of thought, I have two suggestions:

- Move `FanModuleIndex` into `drv/sidecar-mainboard-controller/src/fan_modules.rs`.  This would also let us change the function signatures of methods in `FanModules` to take `idx: FanModuleIndex` instead of `idx: u8`.
- Lose all of the `impl From` below.  We've already got `from_usize` because we derived `FromPrimitive`, and since this is a simple `enum`, we can go to numerical type with just `idx as usize`.

I made these changes locally to test them out; you're welcome to borrow [the diff](https://gist.github.com/mkeeter/c43be110fa437f7ebb67752334963968).

task/thermal/src/control.rs
```diff
@@ -77,6 +77,42 @@ impl TemperatureSensor {
...
 }
 
+/// Represents the indvidual fans in the system
+///
+/// Depending on the system we have diferent numbers of fans structured in
+/// different ways. Not all fans are guaranteed to be there at all times so
+/// their corresponding sensor is an `Option`. We should not read the RPM of
+/// fans which are not present and their PWM should only be driven low.
+#[derive(Copy, Clone)]
+pub struct Fans([Option<SensorId>; bsp::NUM_FANS]);
```
mkeeter, line 87: Instead of using the `bsp` value here, you could make this take a const generic:

```rust
pub struct Fans<const N: usize>([Option<SensorId>; N]);
```

To my mind, this is cleaner than coupling to the BSP in the `struct` type itself.

(Of course, this means that you'll need to use `Fan<{ bsp::NUM_FANS }>` instead of `Fans` elsewhere...)

task/thermal/src/control.rs
```diff
@@ -77,6 +77,42 @@ impl TemperatureSensor {
...
+
+impl Fans {
+    pub fn new() -> Self {
+        Self([None; bsp::NUM_FANS])
+    }
+    pub fn add(&mut self, index: usize, sensor: SensorId) {
```
mkeeter, line 93: I'd be tempted to define `core::ops::Index/IndexMut` instead of `add/remove`, because they've got less ambiguous semantics (e.g. I was wondering whether `add(0, ...)` shifts items).

task/thermal/src/control.rs
```diff
@@ -487,6 +525,7 @@ impl<'a> ThermalControl<'a> {
             power_mode: PowerBitmask::empty(), // no sensors active
 
             dynamic_inputs: [None; bsp::NUM_DYNAMIC_TEMPERATURE_INPUTS],
+            fans: Fans([None; bsp::NUM_FANS]),
```
mkeeter, line 528:
```suggestion
            fans: Fans::new(),
```

task/thermal/src/control.rs
```diff
@@ -895,16 +962,21 @@ impl<'a> ThermalControl<'a> {
         Ok(())
     }
 
-    /// Attempts to set the PWM duty cycle of every fan in this group.
+    /// Attempts to set the PWM duty cycle of every present fan in this group.
```
mkeeter, line 965: Clarify here that we _also_ try to set a PWM duty cycle for missing fans, but it's 0.

task/thermal/src/control.rs
```diff
@@ -913,17 +985,25 @@ impl<'a> ThermalControl<'a> {
         last_err.map_err(|_| ThermalError::DeviceError)
     }
 
-    /// Sets the PWM for a single fan
+    /// Sets the PWM for a single fan if it is present
```
mkeeter, line 988: This has different semantics than `set_pwm`, which sets a PWM of 0 for a missing fan; does this make sense?

drv/sidecar-mainboard-controller/src/fan_modules.rs
Aaron-Hartwig: The intention of this bitfield is essentially to do that, so I like tying them more closely together. Good thought.

drv/sidecar-mainboard-controller/src/fan_modules.rs
Aaron-Hartwig: We do not! I don't know why my brain opted to do that in this case. Removed!

task/thermal/src/control.rs
Aaron-Hartwig: It does not. Fixing!

task/thermal/src/control.rs
Aaron-Hartwig: I think that is an excellent idea.

task/thermal/src/control.rs
Aaron-Hartwig: I think decoupling the type from the BSP is indeed a cleaner, more preferable path.

drv/sidecar-seq-api/src/lib.rs
Aaron-Hartwig: Another good suggestion, and thanks for the diff!

Aaron-Hartwig:
This received some pre-review before testing. I've addressed all initial review comments and spent time testing and fixing this week. Given all of that, I'm going to merge this as it provides a few features we need right now.

Issue #1354, opened 2023-05-12
Expose boot selection and preference information for use by update


andrewjstone:
Essentially we are exposing this structure all the way up to wicket.

```rust
pub struct RotBootInfo {
     pub active: SlotId,
     // What's written in the CFPA as the preferred slot to boot
     // This slot will boot as long as its signature validates
     // and the key hasn't been revoked
     pub persistent_boot_preference: SlotId,

     /// This is a magic ram value that is cleared by bootleby
     pub transient_boot_preference: Option<SlotId>,
 }
 ```

Ideally, we would be able to transmit this over an attested channel, but that channel is not ready yet.

andrewjstone:
Done in #1359

PR #1359, opened 2023-05-12
Add a new Sprot rot_boot_info API

andrewjstone:
This a draft implementation for https://github.com/oxidecomputer/hubris/issues/1354

This adds the sprot and update APIs for reading RoT boot preference information out of CFPA. It carries over information about currently active RoT slot from the pre-main handoff data used in the update`status` API. Transient boot selection is not currently implemented, and so there is a placeholder `Option<SlotId>` in the struct for that information.

There is a bit more to do before this is ready to go:
- [ ] Add new messages to `gateway-messages` so that we can send this information to MGS
- [ ] Plumb sprot calls up through control plane agent using the new messages in `gateway-messages`
- [ ] Rebase on Matt's caboose changes in #1357 and move the `Update.rot_boot_info` call into the RoT specific idol file along with related types.


drv/lpc55-update-server/src/main.rs
```diff
@@ -226,6 +227,25 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
         Ok(status)
     }
 
+    fn rot_boot_info(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<RotBootInfo, RequestError<UpdateError>> {
+        // Safety: Data is published by stage0
```
andrewjstone, line 234: Change this comment to say `pre-kernel main`

drv/sprot-api/src/lib.rs
```diff
@@ -333,18 +334,20 @@ pub enum UpdateReq {
...
     Reset,
+    BootInfo,
 }
 
 /// A response used for RoT updates
-#[derive(Clone, Serialize, Deserialize, SerializedSize)]
+#[derive(Clone, Serialize, Deserialize, SerializedSize, From)]
 pub enum UpdateRsp {
     BlockSize(u32),
+    BootInfo(RotBootInfo),
```
andrewjstone, line 344: Add a comment that this enum variant was added in sprot protocol version 3.

drv/update-api/src/lib.rs
```diff
@@ -128,6 +129,10 @@ impl From<UpdateError> for GwUpdateError {
             UpdateError::ImageBoardUnknown => Self::ImageBoardUnknown,
             UpdateError::TaskRestarted => Self::TaskRestarted,
             UpdateError::NotImplemented => Self::NotImplemented,
+
+            // TODO: Update gateway error to handle this then change the error to
+            // a proper enum
```
andrewjstone, line 134: I'm starting to really question this strategy. We won't actually know when all MGSs are updated with the latest errors and so if we send back something other than `Unknown(error_code)`, MGS may not be able to deserialize it.

It would be nice to have a way in gateway-messages, for errors in particular, to be able to check to see if they can deserialize, and if not, log the discriminant byte of the error in an `Unknown` value. This probably can't be automatic for all errors, but it seems like a way around the non-atomic update issue such that we wouldn't perpetually send `Unknown(code)` for each new error variant that wasn't there originally.

Curious for your thoughts @jgallagher @cbiffle @mkeeter

drv/update-api/src/lib.rs
```diff
@@ -173,6 +178,31 @@ pub enum SlotId {
     B,
 }
 
+impl From<RotSlot> for SlotId {
```
andrewjstone, line 181: Note to the reader: I did this to transition away from using `RotSlot` which comes from stage0_handoff to using the higher level `SlotId`

idl/sprot.idol
```diff
@@ -22,7 +22,7 @@ Interface(
             idempotent: true,
         ),
         "rot_state": (
-            doc: "Return boot info about the RoT",
+            doc: "Return boot time boot info about the RoT",
```
andrewjstone, line 25: typo

idl/update.idol
```diff
@@ -61,12 +61,23 @@ Interface(
             encoding: Hubpack
         ),
         "status": (
-            doc: "Get info about installed images",
+            doc: "Get info about installed images (deprecated - use boot_info)",
```
andrewjstone, line 64: The name changed to `rot_boot_info`

drv/lpc55-update-server/src/main.rs
```diff
@@ -385,6 +357,89 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+        };
+
+        Ok(val)
+    }
+
+    /// Return the persistent and transient boot preferences
+    fn boot_preferences(
+        &mut self,
+    ) -> Result<(SlotId, Option<SlotId>), UpdateError> {
+        let cfpa_word_number = self.cfpa_word_number()?;
```
andrewjstone, line 417: I realized yesterday that this likely *not* the behavior we want. The factoring out of `self.cfpa_word_number()` only returns the correct version of cfpa page by looking at the `ping` and `pong` pages. Unfortunately, when we update the CFPA permanent boot selection we write to the `scratch` page. This only gets promoted to the higher version of `ping` or `pong` and becomes authoritative after a reboot. Therefore an operator, say, selects slot `B` as the boot preference, and then goes and reads the boot preference we will show them the previously adopted authoritative page, which may show `A` as the preference.

What we actually want to do is find the current authoritative page, and also see if there has been an update to the scratch page by comparing versions. If the scratch page is more recent it means an update was written but we haven't rebooted and we should show what was written as it will become authoritative on boot.

Interestingly I also discovered, at least on this branch, that bootleby does not seem to be respecting the selection of bank B, and will always boot bank A unless I erase A and only leave B. 

CC @cbiffle

andrewjstone:
Closed in favor of #1367

PR #1372, opened 2023-05-22
Add production images

mkeeter:
This PR disables features that we don't want in production images.

It splits images into three flavors with different `udprpc`, technician port (TP), and power-on behavior:

| Name | `udprpc` | TP can talk to any SP | A0 on boot |
|-|-|-|-|
| `rev-c.toml` | Disabled | No | Yes |
| `rev-c-lab.toml` | Enabled | Yes | Yes |
| `rev-c-bench.toml` | Enabled | Yes | No |

This is deliberately opened as a draft PR, because it will dramatically change system behavior:

- `pilot sp list` and `faux-mgs` will only work from the switch zone (because SPs will no longer talk to the TP)
- ⚠️ If the Scrimlet's switch zone fails to automatically come up and connect to the management network, we have **very limited** access to the system
  - The second Sidecar + Scrimlet can be used in a pinch, as long as we only update them one at a time
- Most `humility` subcommands will no longer work remotely; any command that executes code on the target will fail.  The only commands that will work are memory reads, e.g. `humility ringbuf/tasks/readvar`

### DIFF
--- a/app/demo-stm32h7-nucleo/app-h753.toml
+++ b/app/demo-stm32h7-nucleo/app-h753.toml
@@ -134,6 +134,7 @@ priority = 3
 max-sizes = {flash = 32768, ram = 8192}
 stacksize = 4096
 start = true
+features = ["enable"]
 task-slots = ["net"]
 notifications = ["socket"]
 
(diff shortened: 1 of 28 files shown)


cbiffle:
The RoT equivalent uses `-dev`, fwiw.

mkeeter:
Closing in favor of #1424

PR #1378, opened 2023-05-25
Add task named 'attest' to acts as the root of trust for reporting.

flihp:
This task is currently limited to producing a cert chain from the `Alias` cert back to either the last intermediate before the root (dice-mfg) or a self-signed identity cert (dice-self). A matching PR to humility will provide an interface to extract these certs & provide a nice interface (PEM encoding, combined cert chain etc).


idl/attest.idol
```diff
@@ -0,0 +1,39 @@
+// Interface to Root of Trust for Reporting (ROT-R) task.
+
+Interface(
+    name: "Attest",
+    ops: {
```
cbiffle, line 5: I should get around to making this the default, but could I encourage you to apply
```
encoding: Hubpack,
```
to each operation in any new IDL interface? It will let you use more flexible argument and error types.

task/attest-api/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+//! API crate for the root of trust for reporting (RoT-R).
+
+#![no_std]
+
+use derive_idol_err::IdolError;
+use userlib::{sys_send, FromPrimitive};
+
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, FromPrimitive, Eq, PartialEq, IdolError)]
+pub enum AttestError {
```
cbiffle, line 14: FWIW, should you elect to turn on `encoding: Hubpack`, this type will need to change in the following ways:

1. add `derive(Serialize, Deserialize, hubpack::SerializedSize)`
2. remove `IdolError`, `FromPrimitive`, `#[repr(u32)]`
3. no longer needs explicit discriminator values.

task/attest/src/main.rs
```diff
@@ -0,0 +1,185 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Root of trust for reporting (RoT-R) task.
+//!
+//! Use the rotr-api crate to interact with this task.
```
cbiffle, line 7: nit: you wound up naming that crate `attest-api`.

task/attest/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+#[export_name = "main"]
+fn main() -> ! {
+    ringbuf_entry!(Trace::Startup);
+
+    let addr = unsafe { CERTS.assume_init_ref() };
+    let cert_data = match CertData::load_from_addr(addr) {
+        Ok(a) => a,
+        Err(_) => panic!("CertData"),
```
cbiffle, line 162: If the cert data cannot be loaded, this will cause your task to restart forever, starving all tasks at lower priorities. That's not good.

I recommend instead changing the content of the `Server` to `Option` and providing a way of indicating "could not load certs" in the IPC API.

task/attest/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+    Offset(usize),
+    Startup,
+    None,
+}
+
+ringbuf!(Trace, 16, Trace::None);
+
+// Map the memory used to pass the segment of the identity cert chain common
+// to all tasks to a variable.
+#[used]
```
cbiffle, line 39: You're doing a somewhat subtle thing here that I think should get added to the comment (because I had to cursor around and convince myself it was correct).

By assigning a `link_section` you're saying you want this placed in an area of memory -- _not_ at the base of that memory. However, the app config (and thus the linker script) sizes these two sections of memory such that they're exactly the size of these statics. This should ensure that they're placed at the base of the section.

So, I'd mention that intent in the comment, lest someone adjust it later and cause this to become invalid.

This would probably be better served by the `extern-regions` facility, which helps to avoid this class of thing.

task/attest/src/main.rs
flihp: oof that's bad test coverage on my part.

flihp: ran across a small bug in `lib-lpc55-rot-startup` dependencies while testing this: https://github.com/oxidecomputer/hubris/pull/1402

flihp:
CI isn't running on this PR, probably caused by the conflicts noted. Rebasing during a review is often frowned upon so I'll hold off fixing these conflicts unless requested.

task/attest/src/main.rs
flihp: Spent some time reading up on the `extern-regions` stuff over the last few days. This PR has been updated to replace the existing code using `link_section`s with `extern-regions`. This took a bit more stack space but there's probably ways to reduce that if necessary (mut statics?).

flihp:
gah formatting

flihp:
and again with my bad license text fixed

flihp:
oxide-rot-1/app-dev.toml builds fine for me locally but fails in CI needing more flash. Not sure I have any option but to YOLO larger flash values at CI till it works.

mx-shift (approved):

PR #1380, opened 2023-05-26
Update 'dice-mfg-msgs' to get new 'PDV2' identity string format.

flihp:


jclulow:
I have tested this on a PSC/B on **zoot**, one of the programming stations in the office.  I was able to:

- use `embootleby install --erase-all` to blank out the PSC
- install the previous PSC Hubris image (from the DVT dock: branch **staging-dice-mfg-uart** plus a patch to increase the kernel flash size slightly, signed with staging/development keys)
- use a previous version of `dice-mfg` to generate a PDV1 format certificate using the staging intermediate on **zoot**
- install a PSC Hubris image with this change (same as previous image + this patch + the attest task from #1378)
- confirm via `humility attest cert 1 | openssl x509 -noout -text` that the certificate in the device is the PDV1 certificate originally issued, and has not been clobbered by the new software

Then, separately, I was able to:

- use `embootleby install --erase-all` to blank out the PSC
- install a PSC Hubris image with this change (same as previous image + this patch + the attest task from #1378) directly
- use a current version of `dice-mfg` (which already has the PDV2 change) to generate a PDV2 format certificate using the staging intermediate on **zoot**
- confirm via `humility attest cert 1 | openssl x509 -noout -text` that the certificate in the device is a new PDV2 certificate
- confirm that the system is working without crashes via `humility tasks`

Issue #1383, opened 2023-05-30
RoT provisioning interface needs integrity checks


cbiffle:
The RoT UART provisioning connection used in manufacturing doesn't currently have any integrity checks. The outgoing CSR is internally signed, so if it were corrupted we'd notice. However, the cert we send back to be stored in flash could be corrupted.

One possible solution to this is to add an operation for reading out the e.g. SHA3 of the cert after writing it, and allowing it to be rewritten if it's bogus up until the command to reboot.

A protocol-level checksum would probably also not be a bad idea, to prevent the RoT from acting on spurious commands.

flihp:
The possible solution provided should also include the ability check the integrity of the intermediate cert provided to the RoT as well. Functionally this is a different certificate but its integrity is just as important to validation. I was thinking the right thing to do would be to provide a message that would allow the mfg software to query the RoT being programmed for the digest of the last message it received. This would require the RoT to calculate a digest for each message received and to save the digest of the last message received as internal state. The mfg software would then be able to selectively verify the integrity of the last data received by the RoT.

Another option would be to modify the existing `ACK` message to include the hash of the message it is sent in response to. This would allow the mfg software to check that the RoT received each command sent.

These two approaches feel functionally equivalent so I'll start prototyping to see if one works better than the other. I'll gladly take advice from anyone with experience / opinions that would favor one approach over the other.

flihp:
PR is open & ready for review: https://github.com/oxidecomputer/hubris/pull/1385

flihp:
https://github.com/oxidecomputer/hubris/pull/1385 has been merged

PR #1385, opened 2023-05-31
Supply MessageHash to Ack MfgMessage.

flihp:
https://github.com/oxidecomputer/hubris/issues/1383
https://github.com/oxidecomputer/dice-util/pull/83

### DIFF
--- a/lib/dice/src/mfg.rs
+++ b/lib/dice/src/mfg.rs
@@ -6,11 +6,12 @@ use crate::{
     cert::PersistIdSelfCertBuilder, csr::PersistIdCsrBuilder, CertSerialNumber,
     IntermediateCert, PersistIdCert, SeedBuf,
 };
-use dice_mfg_msgs::{MfgMessage, PlatformId, SizedBlob};
+use dice_mfg_msgs::{MessageHash, MfgMessage, PlatformId, SizedBlob};
 use lib_lpc55_usart::{Read, Usart, Write};
 use lpc55_pac::SYSCON;
 use nb;
 use salty::{constants::SECRETKEY_SEED_LENGTH, signature::Keypair};
+use sha3::{digest::FixedOutputReset, Digest, Sha3_256};
 use unwrap_lite::UnwrapLite;
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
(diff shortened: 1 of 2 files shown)


lib/dice/src/mfg.rs
```diff
@@ -229,13 +248,15 @@ impl DiceMfg for SerialMfg<'_> {
                 }
                 MfgMessage::CsrPlz => self.handle_csrplz(),
                 MfgMessage::IdentityCert(cert) => {
-                    self.handle_persistid_cert(cert)
+                    self.handle_persistid_cert(cert, &hash)
```
kc8apf, line 251: Not thrilled with having to carry hash through various method calls solely so it can be used as part of `send_ack()`.  Adopting TLS 1.3's transcript hash approach (https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.1) seems like it would simplify the code here and reduce the traffic sent over the wire.  Essentially, `SerialMfg` would keep a SHA3 hasher and each received and sent message would be added to the hash.  When an Ack is sent (or any other message that wishes to include a hash to check session integrity at that point), `finalize_reset()` gets called on the hasher.  `send_ack` now only needs a `&mut self` and integrity can be checked over a run of multiple request/response pairs.

andrewjstone: WE actually do this in sprockets: https://github.com/oxidecomputer/sprockets/blob/7c9bcb262cb117795976056a9102b96c9971d80b/session/src/client.rs#L360

flihp: Thanks for the review @kc8apf. Passing the hash through parameters wasn't so much an artifact of the hashing approach taken as the implementation. I've gone back over the code to address these concerns and the flash savings + simplification are significant.

flihp:
The structure of the code in the PR has been updated thanks to feedback from reviewers. The review feedback that has not been implemented is the adoption of the integrity mechanism from the TLS protocol. The approach taken is less elegant but it appears to address our concerns. For the sake of time my preferred approach would be to get this merged & over to mfg quickly so that we have a more robust mechanism in place for PVT1. In parallel I'll update the relevant RFD to capture this approach along with the alternatives.

mx-shift (approved):

PR #1389, opened 2023-06-02
Better vpd diagnostics

cbiffle:


drv/local-vpd/src/lib.rs
```diff
@@ -121,19 +125,31 @@ pub fn read_config_into(
     i2c_task: TaskId,
     tag: [u8; 4],
     out: &mut [u8],
+) -> Result<usize, LocalVpdError> {
+    match read_config_inner(i2c_task, tag, out) {
+        Ok(n) => Ok(n),
+        Err(e) => {
```
cbiffle, line 131: This'd be a use-case for [`inspect_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err) if it were stable.

drv/local-vpd/src/lib.rs
```diff
@@ -142,35 +158,33 @@ pub fn read_config_into(
...
                     }
-                    return Err(err(LocalVpdError::NoSuchChunk));
+                    return Err(LocalVpdError::NoSuchChunk(tag));
                 } else {
                     ringbuf_entry!(Trace::UnrelatedChunk(chunk.header().tag));
                 }
             }
-            Ok(None) => return Err(err(LocalVpdError::NoRootChunk)),
-            Err(_) => return Err(err(LocalVpdError::BadRootChunk)),
+            Ok(None) => return Err(LocalVpdError::NoRootChunk),
```
jgallagher, line 186: I don't remember exactly how `reader.next()` behaves, but it _looks_ like this error might be wrong. We could have a root chunk, go into the `Ok(Some(chunk))` branch some number of times, and then reach the end and hit this branch, right? Unless `reader.next()` is guaranteed to return an error instead of `Ok(None)` when it gets to the end (which might be how it actually behaves in practice)?

jgallagher (approved):

drv/local-vpd/src/lib.rs
cbiffle: We talked about this in chat and resolved it:
- The code will look at many top-level TLVC chunks and early-exit if it finds `FRU0`
- This code path happens if it couldn't find `FRU0`
- While `NoRootChunk` doesn't suggest that to me or @jgallagher, I added a comment on it attempting to explain it.

PR #1391, opened 2023-06-05
Puf check

flihp:

PR #1392, opened 2023-06-05
puf check

flihp:
this is a copy of https://github.com/oxidecomputer/hubris/pull/1391 using a branch from origin

### DIFF
--- a/lib/lpc55-rot-startup/src/dice.rs
+++ b/lib/lpc55-rot-startup/src/dice.rs
@@ -51,22 +51,24 @@ fn gen_mfg_artifacts(peripherals: &Peripherals) -> MfgResult {
 /// cannot easily have identities certified by an external CA.
 #[cfg(feature = "dice-self")]
 fn gen_mfg_artifacts_self(peripherals: &Peripherals) -> MfgResult {
+    use core::ops::{Deref, DerefMut};
     use dice_crate::{DiceMfg, PersistIdSeed, SelfMfg};
+    use zeroize::Zeroizing;
 
     let puf = Puf::new(&peripherals.PUF);
 
     // Create key code for an ed25519 seed using the PUF. We use this seed
     // to generate a key used as an identity that is independent from the
     // DICE measured boot.
-    let mut keycode = [0u32; KEYCODE_LEN];
-    if !puf.generate_keycode(KEY_INDEX, SEED_LEN, &mut keycode) {
+    let mut keycode = Zeroizing::new([0u32; KEYCODE_LEN]);
+    if !puf.generate_keycode(KEY_INDEX, SEED_LEN, keycode.deref_mut()) {
         panic!("failed to generate keycode");
     }
     let keycode = keycode;
 
     // get keycode from DICE MFG flash region
     let mut seed = [0u8; SEED_LEN];
-    if !puf.get_key(&keycode, &mut seed) {
+    if !puf.get_key(keycode.deref(), &mut seed) {
         // failure to get this key isn't recoverable
         panic!("failed to get seed");
     }
--- a/lib/lpc55-rot-startup/src/dice_mfg_usart.rs
+++ b/lib/lpc55-rot-startup/src/dice_mfg_usart.rs
@@ -3,7 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use crate::dice::{MfgResult, KEYCODE_LEN, KEY_INDEX, SEED_LEN};
-use core::ops::Deref;
+use core::ops::{Deref, DerefMut};
 use dice_crate::{
     CertSerialNumber, DiceMfg, IntermediateCert, PersistIdCert, PersistIdSeed,
     PlatformId, SeedBuf, SerialMfg,
@@ -15,6 +15,7 @@ use lpc55_puf::Puf;
 use salty::signature::Keypair;
 use serde::{Deserialize, Serialize};
 use static_assertions as sa;
+use zeroize::Zeroizing;
 
 macro_rules! flash_page_align {
     ($size:expr) => {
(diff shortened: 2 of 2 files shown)


lib/lpc55-puf/src/lib.rs
```diff
@@ -315,6 +315,48 @@ impl<'a> Puf<'a> {
...
+    fn is_lock_bit_set(&self, idxblk: u32) -> bool {
+        (idxblk & 1 << 30) != 0
+    }
+
+    fn is_unlock_bit_set(&self, idxblk: u32) -> bool {
+        (idxblk & 1 << 31) != 0
+    }
+
+    fn is_locked(&self, idxblk: u32) -> bool {
+        if self.is_lock_bit_set(idxblk) && !self.is_unlock_bit_set(idxblk) {
```
mx-shift, line 327: I suspect I'd find this easier to read/understand as `match (self.is_lock_bit_set(idxblk), self.is_unlock_bit_set(idxblk))`

lib/lpc55-rot-startup/src/lib.rs
```diff
@@ -60,6 +63,22 @@ fn apply_memory_protection(mpu: &MPU) {
...
 }
 
+// Execute this function before jumping to Hubris.
+// This function will panic if PUF is not in the desired state.
+#[cfg(any(feature = "dice-mfg", feature = "dice-self"))]
+fn puf_check(puf: &lpc55_pac::PUF) {
+    use crate::dice::KEY_INDEX;
+    let puf = Puf::new(puf);
+
+    if !puf.is_index_blocked(KEY_INDEX).unwrap_lite()
```
mx-shift, line 73: I had to read this twice to realize blocked and locked were different.  Maybe `is_index_blocked` and `is_index_blocking_locked`?

flihp:
I've started working on another commit for this branch that zeroizes the keycode used for the identity key. This is additional insurance against possible misconfiguration of the PUF. This keycode will be written to flash in the dice mfg region so zeroizing it from memory doesn't protect against all threats as it's still in persistent storage.

lib/lpc55-puf/src/lib.rs
```diff
@@ -421,6 +421,20 @@ impl<'a> Puf<'a> {
...
+
+    pub fn is_sram_on(&self) -> bool {
+        self.puf.pwrctrl.read().bits() & 1 != 0
+    }
+
+    pub fn is_sram_ready(&self) -> bool {
+        self.puf.pwrctrl.read().bits() & 2 != 0
+    }
+
+    pub fn disable_sram(&self) {
```
flihp, line 433: This works but it's not quite right. I'll be force pushing a fix / change.

lib/lpc55-puf/src/lib.rs
```diff
@@ -13,15 +13,15 @@ use unwrap_lite::UnwrapLite;
 fn disable_index_bits(bits: u32, index: u32) -> u32 {
     let index = index % 8;
 
-    bits & !(2 << index * 2) | 1 << index * 2
+    bits & !(2 << (index * 2)) | 1 << (index * 2)
```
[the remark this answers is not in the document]
flihp: If my understanding of the order of operation it was and still is correct. Adding the parens just satisfies clippy warnings w/r/t clarity.

lib/lpc55-puf/src/lib.rs
flihp: that is much easier to read

lib/lpc55-puf/src/lib.rs
```diff
@@ -315,6 +315,48 @@ impl<'a> Puf<'a> {
...
+    pub fn is_index_locked(&self, index: u32) -> Option<bool> {
+        match index {
+            1..=7 => Some(self.is_idxblk_l_locked()),
+            8..=15 => Some(self.is_idxblk_h_locked()),
+            _ => None,
+        }
+    }
+
+    /// Get the contents of PUF 'stat' register.
+    pub fn get_stat(&self) -> u32 {
```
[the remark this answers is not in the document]
flihp: This was only ever used in my test task which can easily do w/o it: removed.

lib/lpc55-puf/src/lib.rs
```diff
@@ -315,6 +315,48 @@ impl<'a> Puf<'a> {
         }
     }
 
+    fn is_lock_bit_set(&self, idxblk: u32) -> bool {
```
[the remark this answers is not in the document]
flihp: Squinting in this way makes this much more clear. Thanks!

flihp:
This last commit removes the call to disable the PUF SRAM. The commit message has the details but the TL;DR is that enabling and initializing it again after this is particularly difficult. Since we don't yet know how we'll be using the PUF in the future adding this call seems premature.

lib/lpc55-puf/src/lib.rs
```diff
@@ -6,8 +6,20 @@
...
 use core::mem;
 use lpc55_pac::PUF;
+use num_derive::FromPrimitive;
+use num_traits::FromPrimitive;
 use unwrap_lite::UnwrapLite;
 
+/// Used to represent valid states for PUF index blocking bits in IDXBLK_L &
+/// IDXBLK_H registers. We derive FromPrimivite for this type to enable use
+/// of 'from_u32'. This will map invalid / reserved register states to `None`.
+#[repr(u32)]
```
cbiffle, line 16: (tip: you don't need `repr(u32)` here, that only matters if you're going to be casting pointers to it or otherwise doing bit manipulation. `FromPrimitive` removes the need for that.)

lib/lpc55-puf/src/lib.rs
```diff
@@ -451,17 +459,17 @@ mod tests {
...
     #[test]
     fn index_from_kc1() {
-        assert_eq!(index_from_keycode(&[0x4000101_u32]).unwrap_lite(), 1);
+        assert_eq!(index_from_keycode(&[0x4000101_u32]), 1);
     }
 
     #[test]
     fn index_from_kc3() {
-        assert_eq!(index_from_keycode(&[0x4000301_u32]).unwrap_lite(), 3);
+        asskert_eq!(index_from_keycode(&[0x4000301_u32]), 3);
```
cbiffle, line 467: gesundheit

Issue #1382, opened 2023-05-27
sidecar: chassis LED doesn't turn on by default


rmustacc:
After flashing a system with an SP production image at rev 55555772ddb646506042e4006e61289bea23d9e9, @eanders01 and I noticed that the Sidecar system chassis LED did not turn on when we turned on power. I was able to manually toggle it with humility so it's not broken, but this led to a bunch of confusion and manual checking of the TPS546B24A (see also #1381).

PR #1393, opened 2023-06-05
Fix sidecar system LED default state

Aaron-Hartwig:
When I retooled the internal front-io LEDs logic I failed to change the _state_ of the system LED to `On`, so the old initialization process would turn it on directly (which lasted for for a single tick) and then the update loop would see the state was still in `Off` state and shut it off. This change just modifies the initialization process to update the state correctly and the loop will do the work of turning it on.

This resolves https://github.com/oxidecomputer/hubris/issues/1382


rmustacc (approved):
Thanks for hopping on this!

PR #1396, opened 2023-06-06
Add yet another build for oxide-rot-1-dev.

cbiffle:


cbiffle:
@flihp @mx-shift I could technically self-approve this to bypass the checks but that seems outside the spirit of the rules. So if one of y'all could look at this and make sure I'm not, like, maliciousing the rots, I'd appreciate it.

mx-shift (approved):
No roots being malicioused

PR #1397, opened 2023-06-07
properly validate M.2 devices on Rev D

### LINKED ISSUE #1387: Include Gimlet D M.2 drives in `validate` output.
In #888 we made some changes to support the Gimlet B-C I2C mux issue, wherein attempting to talk to the M.2 disks when they're not powered would permanently take out the I2C bus (by muxing a black hole).

@Aaron-Hartwig would like this reversed for D rev so that we can validate the M.2s. I've attempted to do it based on the diff in #888 but can't get it to work, so I think some other things have changed since. Halp.

bcantrill:
Fixes #1387.

### DIFF
--- a/task/thermal/src/bsp/gimlet_bcd.rs
+++ b/task/thermal/src/bsp/gimlet_bcd.rs
@@ -24,12 +24,20 @@ task_slot!(SEQ, gimlet_seq);
 // the control loop.
 const NUM_TEMPERATURE_SENSORS: usize = sensors::NUM_TMP117_TEMPERATURE_SENSORS;
 
+#[cfg(any(target_board = "gimlet-b", target_board = "gimlet-c"))]
+const NUM_NVME_BMC_TEMPERATURE_SENSORS: usize =
+    sensors::NUM_NVME_BMC_TEMPERATURE_SENSORS
+        + sensors::NUM_M2_HP_ONLY_TEMPERATURE_SENSORS;
+
+#[cfg(not(any(target_board = "gimlet-b", target_board = "gimlet-c")))]
+const NUM_NVME_BMC_TEMPERATURE_SENSORS: usize =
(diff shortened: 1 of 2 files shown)


mkeeter (approved):

task/thermal/src/bsp/gimlet_bcd.rs
```diff
@@ -24,12 +24,20 @@ task_slot!(SEQ, gimlet_seq);
 // the control loop.
 const NUM_TEMPERATURE_SENSORS: usize = sensors::NUM_TMP117_TEMPERATURE_SENSORS;
 
+#[cfg(any(target_board = "gimlet-b", target_board = "gimlet-c"))]
+const NUM_NVME_BMC_TEMPERATURE_SENSORS: usize =
+    sensors::NUM_NVME_BMC_TEMPERATURE_SENSORS
+        + sensors::NUM_M2_HP_ONLY_TEMPERATURE_SENSORS;
+
+#[cfg(not(any(target_board = "gimlet-b", target_board = "gimlet-c")))]
```
mkeeter, line 32: Nit: maybe specify as `target_board = "gimlet-d"` here and below?  It's shorter _and_ and has the advantage of complaining in the future when we add new targets.

bcantrill: So, the rationale there was that future Gimlet boards really shouldn't have this issue, and I didn't want to necessitate changing this if/when we have another Gimlet rev -- and then didn't want to run the risk that anything would silently compile. But perhaps that's a misplaced concern?

bcantrill: Eh, I think it's better to call out `gimlet-d` explicitly.

bcantrill: Fixed.

PR #1401, opened 2023-06-08
Remove `kipc::read_caboose_pos`; use the build system instead

mkeeter:
The caboose position is not known until all tasks are built, because it's positioned after the final task in flash memory.

However, tasks need to know its position to read from it.  To bridge this gap, I added a KIPC call that decodes the image header (in the kernel) and returns the caboose position (and a `userlib` function to convert it into a slice).

KIPC has a runtime cost and is rather inelegant.  In this PR, I modify the build system to bake a `CABOOSE_POS` variable into the task binary, in the same way that we inject task slots _after_ all tasks is built.  Then, reading the caboose position at runtime is a simple volatile read.

This is a moderate size improvement (`caboose_reader` goes from 8952 -> 7704 bytes of flash), but is mostly because the previous system offended my sensibilities.


cbiffle (approved):
The caboose-related stuff should probably go in `lib` rather than `drv` because it's not a driver. But since `drv/caboose` exists today, that's not necessarily a blocker, merely a thing I noticed.

PR #1403, opened 2023-06-08
Only do monorail stuck-link handling when in A0

mkeeter:
The Tofino is powered off in power states other than A0; this prevents us from doing a reinitialization every ~20 seconds when the system is in A2.

### DIFF
--- a/task/monorail-server/src/bsp/sidecar_bc.rs
+++ b/task/monorail-server/src/bsp/sidecar_bc.rs
@@ -3,7 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use drv_sidecar_front_io::phy_smi::PhySmi;
-use drv_sidecar_seq_api::{SeqError, Sequencer};
+use drv_sidecar_seq_api::{SeqError, Sequencer, TofinoSeqState};
 use ringbuf::*;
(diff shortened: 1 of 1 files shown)


task/monorail-server/src/bsp/sidecar_bc.rs
```diff
@@ -447,14 +447,25 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
-                let host_reset = pcie_hotplug_status & 1;
-                host_reset == 0
+        // The 10G link should only be up if we're in A0 and the host isn't
+        // trying to hold us in reset; we won't retrigger autonegotiation /
+        // system reset if that's not the case.
+        let check_10g_link = match self.seq.tofino_seq_state() {
+            Ok(TofinoSeqState::A0) => {
+                match self.seq.tofino_pcie_hotplug_status() {
+                    Ok(pcie_hotplug_status) => {
+                        let host_reset = pcie_hotplug_status & 1;
```
arjenroodselaar, line 457: This is fine for now, but if we're going to rely on this I should make this a proper (typed) part of the API. I'll do this next time I am working in the vicinity.

arjenroodselaar (approved):

PR #1407, opened 2023-06-08
oxide-rot-1: clear unused stack after doing startup

cbiffle:
This is a response to our conversation in chat today where we realized that the Zeroize crate is fundamentally unsound. Because (1) all sensitive material is handled in the `startup` routine, and (2) we don't use statics or the heap, we know that any traces of sensitive material will -- at the time `startup` returns to `main` -- be sitting in unused stack memory, from `main`'s perspective.

So this destroys that memory.

I've also added an `inline(never)` on `startup` to cause my description of how the machine works, above, to vaguely match what the compiler will actually do. (Without it, `startup` and `main` would almost certainly commingle their stack memory, defeating the attempt.)


app/oxide-rot-1/src/main.rs
```diff
@@ -17,6 +17,7 @@ fn main() -> ! {
     let (cycles_per_ms, _div) = get_clock_speed(&peripherals);
 
     startup(&core_peripherals, &peripherals);
+    lpc55_rot_startup::nuke_stack();
```
mx-shift, line 20: I think this could be moved to just after `dice::run()` inside `startup()`.  That'd keep all the details about handing sensitive data in fewer crates.

lib/lpc55-rot-startup/src/lib.rs
```diff
@@ -74,6 +75,13 @@ fn puf_check(puf: &lpc55_pac::PUF) {
     }
 }
 
+/// Run the common startup routine for LPC55-based roots of trust.
+///
+/// NOTE: this function is marked `inline(never)` to try to ensure that its
+/// stack usage is not commingled with that of the calling routine, to enable
+/// the use of `nuke_stack` to make extra sure that no data is left in memory by
+/// this routine.
+#[inline(never)]
```
mx-shift, line 84: Not thrilled with relying on this attribute to enforce that separation.  Would moving `dice::run` to an software-triggered interrupt let us play any games with ensuring it uses a different stack that we can sanitize afterward?

app/oxide-rot-1/src/main.rs
cbiffle: Alright. I've also moved the inline(never).

mx-shift (approved):

PR #1412, opened 2023-06-14
Ignition: make Controller transmit behavior runtime configurable

arjenroodselaar:
The output buffers of the Ignition Controller transceivers are gated by whether or not a Target is present. This avoids the Controller screaming out of an unterminated connector when no device is present in the cubby/slot.

This may cause a secondary failure however. If the Target transmission is not received by the Controller (for example due to a broken cable), the Controller will stay silent, which can not easily be distinguished from the transmit channel from Controller to Target being defective.

This diff adds a software controlled override in the Controller state register to make the Controller transmit Hello messages irrespective of Target presence. This bit would be set during integrated rack test (or development in the lab) to validate the backplane.


app/sidecar/rev-c-lab.toml
```diff
@@ -3,3 +3,4 @@ inherit = "rev-c.toml"
 [patches]
 name = "sidecar-c-lab"
 features.sequencer = ["stay-in-a2"]
+features.ignition = ["always-transmit"]
```
Aaron-Hartwig, line 6: I think we would also want this in the `rev-b-lab.toml`, right?

Aaron-Hartwig (commented):
I'm not sure that we necessarily want this as a `feature` via the `-lab.toml`s. Those images are designed to keep the system in A2 (rather than automatically going to A0). The Ignition feature is of particular use in the rack environment and I think we want (need?) to get into A0 to get the Scrimlet happy. I think we'd rather expose this control via `faux-mgs` so we can enable it as needed without having to flash special SP images.

app/sidecar/rev-c-lab.toml
arjenroodselaar: Yes, this should be present in both lab images.

arjenroodselaar:
> I'm not sure that we necessarily want this as a `feature` via the `-lab.toml`s. Those images are designed to keep the system in A2 (rather than automatically going to A0). The Ignition feature is of particular use in the rack environment and I think we want (need?) to get into A0 to get the Scrimlet happy. I think we'd rather expose this control via `faux-mgs` so we can enable it as needed without having to flash special SP images.

Yes, I am not suggesting this as an alternative to exposing this through `faux-mgs`. This transmit behavior exists because of EMI concerns, but in a lab environment we do not have to worry about this as much. This is simply me suggesting we always enable this for lab builds to reduce potential debugging/mask other problems.

Aaron-Hartwig:
Okay, I see where you're going with that and agree that is fine. I just wanted to make sure that `faux-mgs` was for sure part of the plan here.

Aaron-Hartwig (approved):

PR #1415, opened 2023-06-15
stm32h7-startup: stop I2C timeout during debug halt

cbiffle:
The DBGMCU block has a facility for intercepting clocks to various peripherals during debug halt. This is nice, because otherwise the I2C transaction timeout counter keeps on counting while Humility has halted the processor.

This means that running, say, humility tasks, can cause I2C timeouts while the debugger extracts information, appearing to reproduce various I2C failures but actually just creating new and exciting ones.

So, this change turns on the debug freeze bits for the I2C controllers.


lzrd (approved):
This is consistent with what I read in the data sheet.

Issue #1399, opened 2023-06-07
VSC7448 gets into a state where packets on port 48 are dropped


refugeesus:
We have seen this behavior on all systems in all racks. Pretty sure it's not our hardware for now. See: 
https://github.com/oxidecomputer/meta/issues/177
https://github.com/oxidecomputer/meta/issues/165

Ideally we would have this root caused and fixed before ship so we don't fire it up in their datacenter and it gloriously does... nothing. Please let me know how I can help coordinate this effort when someone has time to pick it up. I'll probably have to power cycle a sidecar a few times to make it crash while I assume someone tries to poke the SP from somewhere else.

mkeeter:
I'm going to start by investigating the Paris TP0 issue (mentioned [here](https://github.com/oxidecomputer/meta/issues/165#issuecomment-1512072835)), since it's (a) on a bench machine and (b) seemingly always reproducible.

(I'm not sure if this will end up with the same root cause, but I'd like to follow the golden rule of not ignoring possibly-related bugs)

mkeeter:
We configured the `lurch` Sidecar (`sidecar-sp`) as follows:
- TP0 (port 44) is connected to the office network via a dumb switch
- TP1 (port 45) is connected to `lurch` via a USB NIC
- Running a `sidecar-b-lab` image, so it stays in A2
- Its own Ignition link is looped back on Ignition port 34, so we can ask it to reset itself

Then, I ran a script to constantly reboot the Sidecar and look for issues pinging it on either technician port:
```bash
#!/bin/bash

for (( ; ; ))
do
        echo "power-cycling Sidecar!"
        if pfexec humility -tsidecar-sp hiffy -c Ignition.send_request -aport=34,request=SystemPowerReset
        then
                echo "SystemPowerReset returned success, which shouldn't happen"
                exit 1
        fi
        echo "waiting for reset to complete..."
        sleep 30
        pfexec humility -tsidecar-sp monorail status -p44,45
        if ! ping -ie1000g0 fe80::aa40:25ff:fe05:0500
        then
                echo "could not ping SP through TP0"
                date
                exit 1
        fi
        if ! ping -iaxf1 fe80::aa40:25ff:fe05:0500
        then
                echo "could not ping SP through TP1"
                date
                exit 1
        fi
        echo "ping okay..."
done
```
(`e1000g0` is connected to the same dumb switch as TP0; `axf1` is the USB NIC connecting `lurch` to TP1)

[...]

PR #1423, opened 2023-06-21
Alternate fix for hubris#1399

mkeeter:
See discussion starting [here](https://github.com/oxidecomputer/hubris/issues/1399#issuecomment-1601060429)


arjenroodselaar (approved):
This seems good for now. As discussed offline with @mkeeter, if we want to do this more fine grained in the future a possible per port method would be to remove it from all VLANs, disable, make config changes if needed, enable, add to VLANs.

PR #1425, opened 2023-06-21
better handle missing I2C muxes

bcantrill:
Draft commit comment:

```
After the fix to #1413, the system is unwilling to ignore mux related
errors -- which has generated a problem on some lab systems that are
missing muxes:  because we can never get the bus into a known mux
state these systems now become entirely unusable, as even devices that
aren't on the missing segments are not accessible.  This fixes that by
observing that a mux that is affirmatively missing -- that is, one
that doesn't reply to its in-band management at all -- can be assumed
to have segments that are similarly missing (and therefore as good as
disabled).  This (naturally) doesn't change the fact that accessing
any devices attached to the missing mux will generate an error, but it
allows the system to broadly drive on absent accesses to those
devices.
```

### DIFF
--- a/drv/stm32xx-i2c-server/src/main.rs
+++ b/drv/stm32xx-i2c-server/src/main.rs
@@ -158,7 +158,7 @@ fn configure_mux(
             //
             all_muxes(controller, port, muxes, |mux| {
                 match mux.driver.enable_segment(mux, controller, None, ctrl) {
-                    Err(code) if code == ResponseCode::MuxMissing => {
+                    Err(ResponseCode::MuxMissing) => {
                         //
                         // The mux is gone entirely.  We really don't expect
                         // this on any production system, but it can be true on
@@ -178,8 +178,7 @@ fn configure_mux(
                         ringbuf_entry!(Trace::MuxMissing(mux.address));
                         Ok(())
                     }
-                    Err(code) => Err(code),
-                    Ok(()) => Ok(()),
+                    other => other,
                 }
             })?;
 
--- a/drv/stm32xx-i2c-server/src/main.rs
+++ b/drv/stm32xx-i2c-server/src/main.rs
@@ -149,13 +149,38 @@ fn configure_mux(
 
         Some(MuxState::Unknown) => {
             //
-            // We are in an unknown mux state.  Before we can do anything,
-            // we need to successfully talk to every mux, and disable every
-            // segment.  If there is any failure through here, we'll just
-            // return the error, leaving our mux state as unknown.
+            // We are in an unknown mux state.  Before we can do anything, we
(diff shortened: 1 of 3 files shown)


drv/stm32xx-i2c-server/src/main.rs
```diff
@@ -149,13 +149,38 @@ fn configure_mux(
...
+            // need to successfully talk to every mux (or successfully learn
+            // that the mux is gone entirely!), and disable every segment.  If
+            // there is any failure through here that isn't the mux being
+            // affirmatively gone, we'll just return the error, leaving our
+            // mux state as unknown.
             //
             all_muxes(controller, port, muxes, |mux| {
-                mux.driver.enable_segment(mux, controller, None, ctrl)
+                match mux.driver.enable_segment(mux, controller, None, ctrl) {
+                    Err(code) if code == ResponseCode::MuxMissing => {
```
cbiffle, line 161: Nit:
```suggestion
                    Err(ResponseCode::MuxMissing) => {
```

drv/stm32xx-i2c-server/src/main.rs
```diff
@@ -149,13 +149,38 @@ fn configure_mux(
...
+                        // the bus.  So on this error (and only this error), we
+                        // note that we saw it, and drive on.  (Note that
+                        // attempting to speak to a device on a segment on the
+                        // missing mux will properly return MuxMissing -- and
+                        // set our bus's mux state to be unknown.)
+                        //
+                        ringbuf_entry!(Trace::MuxMissing(mux.address));
+                        Ok(())
+                    }
+                    Err(code) => Err(code),
```
cbiffle, line 181: I think these lines are equivalent to `other => other`, fwiw.

cbiffle (commented):
Huh, this diff is way smaller than I was expecting. So the key is just permitting a single error condition on the `enable_segment` path?

bcantrill:
> Huh, this diff is way smaller than I was expecting. So the key is just permitting a single error condition on the `enable_segment` path?

Yeah -- the other paths actually handle themselves, it turns out.  When we treat a missing mux as implying that all of its segments are in fact disabled, everything else just... works.

drv/stm32xx-i2c-server/src/main.rs
bcantrill: 🤯🤯🤯  **HOW LONG HAS THAT BEEN THERE?!**

bcantrill: That is a *really* useful pattern (ha!) that I basically assumed (for terrible reasons?) didn't/wouldn't work.

cbiffle (approved):

PR #1430, opened 2023-06-22
lpc55-rot-startup: clear unused Handoff area when storing

### LINKED ISSUE #1429: RoT Handoff area may leak data between resets
`lpc55_rot_startup::handoff::Handoff` uses USB SRAM to pass data from early RoT boot to tasks running under Hubris.  `Handoff::store()` uses hubpack to serialize any type that implements `HandoffData`.  Notably, `HandoffData` requires that the implementing type define MEM_RANGE which specifies where in RAM the serialized data should be stored.  Types implementing `HandoffData` such as `lib_dice::handoff::CertData` use `stage0_handoff::fits_in_ram!()` to statically check that MEM_RANGE for the type is large enough to hold the worst-case size of hubpack encoding for the type.  But what happens when that type encodes to something smaller?   `lpc55_rot_startup::handoff::Handoff::store()` serializes directly into the type's MEM_RANGE w/o clearing it first.  If the previous contents were longer than the new contents, some of the previous contents will remain.

When could this happen?  Bootleby will calculate a different CDI when booting between slot A and slot B simply because embedded memory offsets are different even if the images are otherwise identical.  Since DeviceID is derived from CDI and Alias and Trust Quorum DHE keys are derived from DeviceID, the contents of the Handoff will be quite different as well.  In practice, these keys are all serialized as fixed-length byte arrays and thus should always end up being the same serizalized length regardless of content.

Recommendation: `Handoff::store()` zeros any unused portion of MEM_RANGE.

mx-shift:
Fixes #1429

### DIFF
--- a/lib/lpc55-rot-startup/src/handoff.rs
+++ b/lib/lpc55-rot-startup/src/handoff.rs
@@ -56,6 +56,11 @@ impl<'a> Handoff<'a> {
             .expect("handoff store header");
 
         // Serialize the data
-        n + hubpack::serialize(&mut dst[n..], t).expect("handoff store value")
+        let n = n + hubpack::serialize(&mut dst[n..], t)
+            .expect("handoff store value");
+
+        dst[n..].fill(0);
+
+        n
     }
 }


cbiffle (approved):
I've reviewed the change to the compiler output, and this fill is not compiled out or anything sneaky like that. Looks right to me.

PR #1435, opened 2023-06-23
Add plumbing required to get certs from the attest task via sprot.

flihp:
This allows us to use hiffy in the SP to get certs out of the RoT.

### DIFF
--- a/drv/sprot-api/src/error.rs
+++ b/drv/sprot-api/src/error.rs
@@ -4,6 +4,7 @@
 
 //! Errors for the sprot API
 
+use attest_api::AttestError;
 use derive_more::From;
 use drv_caboose::CabooseError;
 use drv_lpc55_update_api::RawCabooseError;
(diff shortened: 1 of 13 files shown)


drv/sprot-api/src/error.rs
```diff
@@ -28,6 +29,7 @@ pub enum SprotError {
     Spi(SpiError),
     Update(UpdateError),
     Sprockets(SprocketsError),
+    Attest(AttestError),
```
andrewjstone, line 32: You shouldn't add an AttestError here, since it's not directly sprot related. In reality, the `Sprockets` and `Update` errors probably shouldn't live here either but that ship has sailed. 

We figured this out later on with both dump and caboose APIs. We deal with errors that occur on the RoT and must transit sprot by encoding them directly into the underlying response type. So `RspBody::Attest` should contain a `Result` that has an `AttestError` variant and then the sprot api should return a `SprotOrAttest` error. See how the `Caboose` does it for inspiration.

flihp: That makes a lot of sense. I hadn't really picked up on it till now but the `Caboose` errors are conspicuously absent here. I'll update my approach to follow the `Caboose`.

flihp: PR has been updated.

flihp:
I've updated this PR taking into account feedback from @andrewjstone and rebased it now that #1378 has been merged.

cbiffle (approved):

PR #1442, opened 2023-06-27
dice-mfg: allow querying secure boot key slot status

mx-shift:
Tested by modifying lpc55expresso to use dice-mfg, flashing to lpc55xpresso w/ secure boot enabled, and running `dice-mfg get-key-slot-status` (from https://github.com/oxidecomputer/dice-util/pull/88) which reported all slots enabled as expected.

Depends on https://github.com/oxidecomputer/dice-util/pull/88 so CI will fail until that is merged.

### DIFF
--- a/lib/dice/src/mfg.rs
+++ b/lib/dice/src/mfg.rs
@@ -6,12 +6,16 @@ use crate::{
     cert::PersistIdSelfCertBuilder, csr::PersistIdCsrBuilder, CertSerialNumber,
     IntermediateCert, PersistIdCert, SeedBuf,
 };
-use dice_mfg_msgs::{MessageHash, MfgMessage, PlatformId, SizedBlob};
+use core::mem::size_of;
+use dice_mfg_msgs::{
+    KeySlotStatus, MessageHash, MfgMessage, PlatformId, SizedBlob,
+};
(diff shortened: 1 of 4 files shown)


lib/dice/src/mfg.rs
```diff
@@ -299,6 +303,74 @@ impl DiceMfg for SerialMfg<'_> {
...
+                        header: u32,
+                        monotonic_version: u32,
+                        _fields_we_do_not_use: [u32; 4],
+                        rotkh_revoke: u32,
+                        _more_fields_we_do_not_use: [u32; 5],
+                        _prince_ivs: [[u32; 14]; 3],
+                        _nxp_reserved: [u32; 10],
+                        _customer_area: [u8; 224],
+                        _digest: [u8; 32],
+                    }
```
labbott, line 326: ugh I really wish we had this accessible in another crate but I don't have a better idea at  the moment.

labbott (approved):
Double checked the bits against the spreadsheet, LGTM

flihp:
I dug out my local mfg setup yesterday evening and gave this a test (it worked as expected). I'm going to test a few additional configurations but will be merging this first.

flihp (approved):

PR #1444, opened 2023-06-27
Reply with MessageKind::Error when reporting an error

mkeeter:
Otherwise, we copy `MessageKind::HostToSp` from the header, which makes the host mad.


Aaron-Hartwig (approved):
Thanks for the fix!

PR #1400, opened 2023-06-08
Rename Sidecar's fan module RPM sensors

Aaron-Hartwig:
This changes the naming to match the compass notation

Fixes #1370

PR #1445, opened 2023-06-27
sidecar: fix comment about fan rotor ordering

Aaron-Hartwig:
In a [previous comment](https://github.com/oxidecomputer/hubris/pull/1400#discussion_r1223337891) I stated:
>The drawing we got from SD notes an inlet vs outlet fan. In our system the inlet fan is more "north" and connects to pwm0/tach0, while the outlet fan is more "south" and connects to pwm1/tach1.

This is incorrect, for whatever reason my brain that day was thinking that the inlet fan was more towards the back (aka North) while the outlet fan was towards the front (aka South). This is backwards as we draw air from front to back, not back to front, which means that the inlet fan is _actually_ South while the outlet fan is North. This commit cleans up that error.

There aren't any other system impacts beyond fixing up the comment and fan naming.

### DIFF
--- a/app/sidecar/base.toml
+++ b/app/sidecar/base.toml
@@ -453,7 +453,7 @@ address = 0b0100_011
 device = "max31790"
 name = "East"
 description = "Fan 0/1 controller"
-sensors = { speed = 4, names = [ "NE_fan1", "SE_fan1", "ENE_fan0", "ESE_fan0" ] }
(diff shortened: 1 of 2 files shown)


mkeeter:
You'll also want to fix `rev-b-dev.toml` and `rev-c-dev.toml`, which are copy-pasted.

Aaron-Hartwig:
I think at this point it makes sense to get https://github.com/oxidecomputer/hubris/pull/1479 merged prior to this and rebase on that since it will remove a bunch of copy paste.

mkeeter (approved):

PR #1446, opened 2023-06-28
transceivers-server: cleanup interface

Aaron-Hartwig:
Most of the interface here was built out to facilitate benchtop testing during initial development. Now that we have these interfaces proven out and connected via the UDP interface, we don't really need them as part of the general Idol server interface. I'm opting to remove them completely as otherwise it just doubles the surface area where we need to manage the module interactions. This will leave the UDP interface as the sole way to manage the transceivers going forward.


task/thermal/src/control.rs
```diff
@@ -226,6 +226,7 @@ pub struct ThermalSensorErrors {
 }
 
 impl ThermalSensorErrors {
+    #[allow(dead_code)]
```
Aaron-Hartwig, line 229: This is unrelated to the rest of this PR, it just cleans up a warning currently on `master`.

mkeeter (approved):
deleting code good

PR #1447, opened 2023-06-28
transceivers-server: separate SPI and I2C notification loops

Aaron-Hartwig:
The thermal/transceivers communication loop with the FPGA (SPI) runs every 500ms. The LED control (I2C) used to be part of this loop, but we want to to have more granular control there to do a startup animation when Dendrite boots. This change pulls the I2C communication into a faster loop (100ms) to facilitate that. This works as expected on my Sidecar.


mkeeter:
Unfortunately, this doesn't really do what you want it to do.  Each task only gets one timer; you can't wait on both `I2C_MASK` and `SPI_MASK` with different times associated with each one (and the last write to the timer wins).

However, there's good news! We have a `multitimer` crate that multiplexes the single per-task timer across multiple (virtual) notifications.

Take a look at [how it's used in `net`](https://github.com/oxidecomputer/hubris/blob/master/task/net/src/main.rs#L216-L275)

drv/transceivers-server/src/main.rs
```diff
@@ -531,46 +580,13 @@ impl idl::InOrderTransceiversImpl for ServerImpl {
...
-            if modules_present != self.modules_present {
-                self.modules_present = modules_present;
-                ringbuf_entry!(Trace::ModulePresenceUpdate(modules_present));
-            }
-
-            self.update_thermal_loop(status);
-
-            let next_deadline = sys_get_timer().now + TIMER_INTERVAL;
-            sys_set_timer(Some(next_deadline), notifications::TIMER_MASK);
+        if (bits & notifications::SOCKET_MASK | notifications::TIMER_MASK) != 0
```
mkeeter, line 587: I know that this is how it's done in `net`, but upon reading the `multitimers` code, I think that the Right Way™ to do this is to store the timer in the `ServerImpl` and handle notifications in `self.handle_notifications`

```rust
        // Ignore the SOCKET_MASK bit; we'll handle it in the main loop
        
        // Pass all bits to the timer, which picks the one it cares about
        self.timer.handle_notification(bits);
        for t in self.timer.iter_fired() {
            match t {
                Timers::I2C => {
                    server.handle_i2c_loop();
                }
                Timers::SPI => {
                    server.handle_spi_loop();
                }
            }
        }
    }
```
(removing the timer-handling code from the main `loop { ... }`)

This is more efficient, because it skips the calls to `poll_now()`.

drv/transceivers-server/src/main.rs
```diff
@@ -191,7 +204,11 @@ impl ServerImpl {
             ringbuf_entry!(Trace::LEDUpdateError(e));
         }
         // keep track if we are on or off next update when blinking
-        self.blink_on = !self.blink_on;
+        self.blink_cntr += 1;
```
mkeeter, line 207: This could also be a separate multitimer channel!

mkeeter (approved):

Issue #1410, opened 2023-06-13
QSGMII link to front IO PHY sometimes doesn't come up


mkeeter:
When running a hard reboot loop to debug #1399 , I found the system in a state where the link from the VSC7448 to the technician port PHY was down.

```
PORT | MODE    SPEED  DEV     SERDES  LINK |   PHY    MAC LINK  MEDIA LINK
-----|-------------------------------------|-------------------------------
 44  | QSGMII  1G     1G_20   6G_15   err  | VSC8562  err       down
 45  | QSGMII  1G     1G_21   6G_15   down | VSC8562  err       down
```

This is distinct from #1399, where the QSGMII link is fine, but the VSC7448 is dropping packets in its queue system.

We see various bits indicating that the QSGMII link can't sync up:
```
matt@lurch ~ (sidecar-sp) $ h monorail read HW_QSGMII_STAT[11]
humility: attached to 0483:374e:0028001E4741500720383733 via ST-Link V3
humility: Reading HSIO:HW_CFGSTAT:HW_QSGMII_STAT[11] from 0x714601a0
HSIO:HW_CFGSTAT:HW_QSGMII_STAT[11] => 0x20
  bits |    value   | field
   6:1 | 0x10       | DELAY_VAR_X200PS
     0 | 0x0        | SYNC
matt@lurch ~ (sidecar-sp) $ h monorail phy read -p44 MAC_SERDES_PCS_STATUS
humility: attached to 0483:374e:0028001E4741500720383733 via ST-Link V3
Reading from port 44 PHY, register EXTENDED_3:MAC_SERDES_PCS_STATUS
Got result 0xc405
  bits |    value   | field
    15 | 0x1        | MAC_SYNC_FAIL
    14 | 0x1        | MAC_CGBAD
    12 | 0x0        | SGMII_ALIGN_ERROR
    11 | 0x0        | MAC_LP_ANEG_RESTART
     5 | 0x0        | MAC_FDX_ADV
     4 | 0x0        | MAC_HDX_ADV
     3 | 0x0        | MAC_LP_ANEG_CAPABLE
     2 | 0x1        | MAC_LINK_STATUS
     1 | 0x0        | MAC_ANEG_COMPLETE
     0 | 0x1        | MAC_PCS_SIG_DETECT
matt@lurch ~ (sidecar-sp) $ h monorail phy read -p44 MAC_SERDES_STATUS
[...]

PR #1449, opened 2023-06-29
Add Sidecar front IO HSC control, work around front IO PHY oscillator errata

arjenroodselaar:
The oscillator used to drive the VSC8562 PHY on the Sidecar front IO board occasionally fails to start properly, causing the QSGMII link to the management network switch to fail and the technician ports to be non-functional (#1410). This diff does the following:

- Implement sequencer control over the front IO hot swap controller
- Make monorail-server detect the oscillator failure, request a power cycle of the front IO board and retry PHY initialization until the QSGMII link is operational
- Refactor power rail status primitives into a new `fpga-app-api` crate


drv/fpga-app-api/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "drv-fpga-app-api"
+version = "0.1.0"
+edition = "2021"
+
+[features]
```
mkeeter, line 6: Nit: you can delete this and the empty `[build-dependencies]` below

drv/sidecar-front-io/src/controller.rs
```diff
@@ -25,6 +25,13 @@ impl FrontIOController {
         self.fpga.reset()
     }
 
+    #[inline]
+    pub fn ready(&self) -> Result<bool, FpgaError> {
+        self.fpga
+            .state()
+            .map(|s| s == DeviceState::RunningUserDesign)
```
mkeeter, line 32: Nit:
```suggestion
        self.fpga.state()? == DeviceState::RunningUserDesign
```

drv/sidecar-seq-server/src/front_io.rs
```diff
@@ -8,39 +8,41 @@ use drv_sidecar_front_io::{controller::FrontIOController, phy_smi::PhySmi};
 
 #[allow(dead_code)]
 pub(crate) struct FrontIOBoard {
-    pub fruid: I2cDevice,
```
mkeeter, line 11: Why remove `fruid` and `i2c_task` here?  (they're here as a small caching optimization, so we don't need to look them up repeatedly)`

drv/sidecar-seq-server/src/front_io.rs
```diff
@@ -8,39 +8,41 @@ use drv_sidecar_front_io::{controller::FrontIOController, phy_smi::PhySmi};
...
+        let fruid = i2c_config::devices::at24csw080_front_io(i2c_task)[0];
+        At24Csw080::validate(&fruid).unwrap_or(false)
+    }
+
+    pub fn initialized(&self) -> Result<bool, FpgaError> {
+        Ok(self
+            .controllers
+            .iter()
+            .map(|c| c.ready().unwrap_or(false))
+            .fold(true, |init, ready| init && ready))
```
mkeeter, line 45:
```suggestion
        Ok(self.controllers.iter().all(|c| c.ready().unwrap_or(false)))
```
(also, is it okay to discard errors from `ready()` here?)

drv/sidecar-front-io/src/phy_smi.rs
```diff
@@ -89,6 +86,25 @@ impl PhySmi {
...
     }
 
+    pub fn osc_good(&self) -> Result<Option<bool>, FpgaError> {
+        let phy_osc: u8 = self.fpga.read(Addr::VSC8562_PHY_OSC)?;
+
+        Ok(if phy_osc & Reg::VSC8562::PHY_OSC::VALID != 0 {
+            Some(phy_osc & Reg::VSC8562::PHY_OSC::GOOD != 0)
+        } else {
+            None
+        })
```
mkeeter, line 96:
```suggestion
        let good = phy_osc & Reg::VSC8562::PHY_OSC::GOOD != 0;
        let valid = phy_osc & Reg::VSC8562::PHY_OSC::VALID != 0;
        Ok(Some(good).filter(|_| valid))
```

task/monorail-server/src/bsp/sidecar_bc.rs
```diff
@@ -145,25 +146,15 @@ pub fn preinit() {
...
 
 impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
     /// Constructs and initializes a new BSP handle
     pub fn new(vsc7448: &'a Vsc7448<'a, R>) -> Result<Self, VscError> {
         let seq = Sequencer::from(SEQ.get_task_id());
-        let has_front_io = match seq.front_io_phy_ready() {
-            Ok(true) => true,
-            Err(SeqError::NoFrontIOBoard) => false,
-            _ => panic!("front IO board went away after preinit()"),
+        let has_front_io = match seq.front_io_board_present() {
```
mkeeter, line 155: Since `front_io_board_present` is infallible on the `sidecar-seq` side, we could change the reply type in Idol to `Simple(bool)`.

task/monorail-server/src/bsp/sidecar_bc.rs
```diff
@@ -374,30 +393,21 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
         if let Some(phy_rw) = &mut self.vsc8562 {
-            // Do a hard power cycle of the PHY; the fine details of sequencing
-            // are handled by the FPGA
-            phy_rw
-                .set_phy_power_enabled(false)
-                .map_err(|e| VscError::ProxyError(e.into()))?;
-            sleep_for(10);
-            phy_rw
-                .set_phy_power_enabled(true)
+            // Request a reset of the PHY.
```
mkeeter, line 396:
```suggestion
            // Request a reset of the PHY.  If we had previously marked the PHY
            // oscillator as bad, then this power-cycles the entire front IO
            // board; otherwise, it only power-cycles the PHY.
```

task/monorail-server/src/bsp/sidecar_bc.rs
```diff
@@ -374,30 +393,21 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
+            self.seq
+                .reset_front_io_phy()
```
mkeeter, line 398: It seems a little silly that we smuggle the `OSC_GOOD/VALID` bits _through_ the FPGA into the `sidecar-seq` task.

We could instead track oscillator good/valid in this task and add a parameter to the RPC function, e.g. `fn reset_front_io_phy(full_power_cycle: bool)`.

On the other hand, we need to put those bits in the FPGA for other reasons, so maybe it's fine 🤷🏻‍♂️

task/monorail-server/src/bsp/sidecar_bc.rs
```diff
@@ -225,12 +216,40 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
+        // the PHY in a loop, requesting the sequencer to reset as much as
+        // necessary to try and correct the problem.
+        let mut osc_good = false;
+
+        while !osc_good {
+            self.phy_vsc8562_init()?;
+
+            // Determine if the link is up which implies the oscillator is
+            // good.
+            osc_good = self
```
mkeeter, line 235: I'd be tempted to move this into a `fn is_front_io_link_good(...)`, just for clarity.

task/monorail-server/src/bsp/sidecar_bc.rs
```diff
@@ -225,12 +216,40 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
         self.vsc7448.configure_vlan_semistrict()?;
         self.vsc7448_postconfig()?;
 
+        // Some front IO boards have a faulty oscillator driving the PHY,
+        // causing its clock to misbehave some fraction of (re-)boots. Init
+        // the PHY in a loop, requesting the sequencer to reset as much as
+        // necessary to try and correct the problem.
+        let mut osc_good = false;
+
+        while !osc_good {
```
mkeeter, line 230: This loop should only run if `self.vsc8562.is_some()`; otherwise, we'll get stuck here forever if the front IO board is missing.

drv/fpga-app-api/src/power_rail.rs
```diff
@@ -0,0 +1,92 @@
...
+    Debug,
+    Default,
+    Eq,
+    PartialEq,
+    FromBytes,
+    FromPrimitive,
+    AsBytes,
+)]
+#[repr(C)]
+pub struct PowerRailState(u8);
```
mkeeter, line 26: Since this doesn't actually validate that the `u8` is valid, I'd suggest calling it `RawPowerRailState` instead.

drv/sidecar-front-io/sidecar_qsfp_x32_controller_rev_b_c.bit
mkeeter, line 1: Before merging, we should update the README, etc to point at the new buildomat job.

drv/sidecar-mainboard-controller/src/front_io.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    pub fn set_enable(&self, enable: bool) -> Result<(), FpgaError> {
+        let op = if enable {
+            WriteOp::BitSet
+        } else {
+            WriteOp::BitClear
+        };
+
+        self.fpga
+            .write(op, Addr::FRONT_IO_STATE, Reg::FRONT_IO_STATE::ENABLE)
```
mkeeter, line 41: There's an `impl From<bool> for WriteOp`, so this could be
```suggestion
        self.fpga.write(
            enable.into(),
            Addr::FRONT_IO_STATE,
            Reg::FRONT_IO_STATE::ENABLE,
        )
```

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -131,117 +132,38 @@ pub struct TofinoSeqAbort {
...
 
-    fn try_from(data: (PowerRails, u8)) -> Result<Self, Self::Error> {
-        let (id, v) = data;
-        Ok(PowerRail {
-            id,
-            state: PowerRailState::try_from(v >> 4)?,
-            pin_state: PowerRailPinState::from(v & 0xf),
+    fn try_from(data: (usize, PowerRailState)) -> Result<Self, Self::Error> {
+        Ok(TofinoPowerRail {
+            id: TofinoPowerRailId::from_u8(data.0 as u8)
```
mkeeter, line 159:
```suggestion
            id: TofinoPowerRailId::from_usize(data.0)
```
seems to work

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -131,117 +132,38 @@ pub struct TofinoSeqAbort {
...
+impl TryFrom<(usize, PowerRailState)> for TofinoPowerRail {
     type Error = FpgaError;
 
-    fn try_from(data: (PowerRails, u8)) -> Result<Self, Self::Error> {
-        let (id, v) = data;
-        Ok(PowerRail {
-            id,
-            state: PowerRailState::try_from(v >> 4)?,
-            pin_state: PowerRailPinState::from(v & 0xf),
+    fn try_from(data: (usize, PowerRailState)) -> Result<Self, Self::Error> {
```
mkeeter, line 157: I'm not convinced that `impl TryFrom` is better than
```rust
impl TofinoPowerRail {
    fn new(id: usize, state: PowerRailState) -> Result<Self, FpgaError> {
        // ...
    }
}
```
since constructing / indexing the tuple is awkward.

drv/fpga-app-api/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! API crate for shared pieces of logic used in FPGA applications.
+
+#![no_std]
+
+pub mod power_rail;
```
mkeeter, line 9: Why put this in a separate module, versus the existing `fpga-api`?  Are there more pieces of `fpga-api` that should move here eventually?

drv/fpga-app-api/src/power_rail.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub vrhot: bool,
+}
+
+impl From<PowerRailState> for PowerRailPinState {
+    fn from(state: PowerRailState) -> Self {
+        Self {
+            enable: state.0 & 0x1 != 0,
+            good: state.0 & 0x2 != 0,
+            fault: state.0 & 0x4 != 0,
+            vrhot: state.0 & 0x8 != 0,
```
mkeeter, line 89: I have a weak preference for explicit shifts, e.g.
```suggestion
            enable: state.0 & (1 << 0) != 0,
            good: state.0 & (1 << 1) != 0,
            fault: state.0 & (1 << 2) != 0,
            vrhot: state.0 & (1 << 3) != 0,
```
(too bad the columns don't line up)

drv/sidecar-front-io/sidecar_qsfp_x32_controller_regs.json
```diff
@@ -380,10 +400,41 @@
...
+                        {
+                            "type": "field",
+                            "inst_name": "GOOD",
+                            "lsb": 1,
+                            "msb": 1,
+                            "reset": 0,
+                            "sw_access": "rw",
+                            "se_onread": null,
+                            "se_onwrite": null,
+                            "desc": "Bit indicating whether or not the oscillator is operating nominal"
```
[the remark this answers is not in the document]
mkeeter: Okay, we're actually using this field to (persistently) pass data from one task to the other, which is a little weird but probably okay.

drv/sidecar-front-io/src/phy_smi.rs
```diff
@@ -45,15 +45,12 @@ impl PhySmi {
     }
 
     #[inline]
-    pub fn phy_power_enabled(&self) -> Result<bool, FpgaError> {
+    pub fn power_enabled(&self) -> Result<bool, FpgaError> {
```
mkeeter, line 48: Why rename these?  I'd be worried about potential confusion with front IO HSC control; the old name is nicely explicit that this is PHY power control.

drv/fpga-app-api/src/power_rail.rs
```diff
@@ -0,0 +1,92 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use drv_fpga_api::FpgaError;
+use userlib::FromPrimitive;
+use zerocopy::{AsBytes, FromBytes};
+
+/// This module implements state primitives matching the generic power
+/// rail/voltage regulator as implemented in
+/// https://github.com/oxidecomputer/quartz/blob/main/hdl/power_rail.rdl,
+/// https://github.com/oxidecomputer/quartz/blob/main/hdl/PowerRail.bsv.
```
Aaron-Hartwig, line 12: We have a pattern to tightly integrate types defined via RDL into our BSV/Rust code. At first blush, leaning on that could easily eliminate the duplicate hand coding of the `PowerRailStatus` enum. I don't recall if we support generating structs off hand, so `PowerFailPinState` may not be covered but that seems like a natural place to continue to expand our integration here lest we need to change RDL, BSV, and Rust every time a register changes.

drv/sidecar-front-io/src/phy_smi.rs
Aaron-Hartwig: I'm inclined to agree with Matt here, but presumably this was done since the functions are within the `phy_smi` module so adding `phy` into all the names is a bit redundant. I tend to have more verbose tendencies with naming though.

drv/sidecar-front-io/sidecar_qsfp_x32_controller_regs.json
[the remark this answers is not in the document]
arjenroodselaar: Yes, let me elaborate on this a bit. One of my design goals for the sequencer task is that it (or the SP as a whole for that matter) can safely be restarted without interrupting Tofino once the main- and front IO board controllers have been initialized and configured. Hence the external thermal sensor, sequencer logic in the controller etc. There are caveats here that the bitstream versions of those controllers don't change, the clock generator configuration does not change etc., but many times it should be possible to leave Tofino running.

In order to stay true to this goal I needed a more persistent place than sequencer RAM to store the result of the question whether or not the PHY oscillator came out of reset correctly on the last board power up. Otherwise it would either have to take a guess or proactively power cycle the front IO board, interrupting the transceivers. Storing this bit of information with the front IO controller which manages the PHY felt like a reasonable thing to do since it'll automatically get reset if power to the board is cycled, the bitstream of the controller is reloaded or the design is reset (in which case I think the PHY gets reset too).

We're not actually using it to pass data between tasks though. `monorail-server` tells the sequencer that it found the PHY functional, at which point we assume the oscillator is good, and the sequencer then stores it in the front IO board controller (using an API from the `sidecar-front-io-board` crate).

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -353,10 +275,26 @@ impl Sequencer {
...
+            move |id| TofinoPowerRail::try_from((id, power_rail_states[id]));
+
+        Ok([
+            power_rail(0)?,
+            power_rail(1)?,
+            power_rail(2)?,
+            power_rail(3)?,
+            power_rail(4)?,
+            power_rail(5)?,
+        ])
```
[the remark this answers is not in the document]
arjenroodselaar: Ha, nice! Thanks.

drv/sidecar-seq-server/src/front_io.rs
arjenroodselaar: Because they're only used in the `FrontIOBoard::present(..)` function which, unless a front IO board power cycle is requested, is only called once on task init (through `ServerImpl::front_io_board_preinit`). I figured I'd let the lookup happen then rather than on struct creation. And when the `fruid` handle is used we're interacting with the EEPROM over I2C, which I figured dwarfs any speedups we'd get from keeping this handle around.

But I did realize that during sequencer `main` we do get the task handle repeatedly, so there it makes sense to do it once and use it several times.

drv/sidecar-seq-server/src/front_io.rs
arjenroodselaar: Ah nice, thank you.

The function is used in two places; 1) determine of the bitstream may need to be loaded in the controllers, which performs a more detailed check, and 2) to gate the `transceiver` task and the sequencing decision to start Tofino until the sequencer is sure a future power cycle of the front IO board is not needed. In both cases we're really only interested in the happy answer and take the conservative path (default to false) otherwise.

I realized the function signature does not reflect this so it now returns a `bool` instead of a `Result`.

task/monorail-server/src/bsp/sidecar_bc.rs
arjenroodselaar: TIL! That looks much better indeed.

drv/fpga-app-api/src/lib.rs
arjenroodselaar: The intend is that `fpga-api` deals with FPGA devices (reset, loading bitstreams etc.) and `fpga-app-api` would be APIs for logic contained within bitstreams and running on those FPGAs. An alternative name for this crate could be `fpga-design-api`. The point is that there is RTL we're starting to share between multiple tasks and I figured I'd start pulling some of that into it's own crate. Happy to hear suggestions.

drv/sidecar-front-io/sidecar_qsfp_x32_controller_rev_b_c.bit
arjenroodselaar: Definitely. The logic pieces still need to land in quartz/main after which I can pull the final artifacts from the resulting job.

task/monorail-server/src/bsp/sidecar_bc.rs
arjenroodselaar: I should have done a better job documenting this :) Please see https://github.com/oxidecomputer/hubris/pull/1449#discussion_r1253963119. Yes, `monorail-server` does technically have access to this register (because it needs access to the adjacent SMI registers), the expectation is that the sequencer maintains this state (and updates that register).

task/monorail-server/src/bsp/sidecar_bc.rs
arjenroodselaar: Ah yes, thank you!

drv/sidecar-mainboard-controller/src/front_io.rs
arjenroodselaar: Heh yes, I am already forgetting what I wrote less than 6 months ago :/

drv/fpga-app-api/src/power_rail.rs
arjenroodselaar: Fair, this does aid readability because you can read off the bit index.

drv/sidecar-front-io/src/phy_smi.rs
arjenroodselaar: Yeah I am all about localized naming. But if both of you like having the phy portion in there I'll put it back.

drv/fpga-app-api/src/power_rail.rs
arjenroodselaar: Yes, the reason this looks the way it does is because while RDL can define types, we currently can't export those types through the JSON representation into Hubris. We can only generate specific instantiations of that type within a register map. We should figure out how to extend our capabilities here, but I don't want to hold up this diff on something open-ended.

drv/sidecar-mainboard-controller/src/tofino2.rs
arjenroodselaar: I am hesitant here. I understand what you're saying, but conceptually I believe that a `new(..)` function should not error. This may be a left over from my C++ days where throwing exceptions from a constructor was problematic, but I still like to separate memory allocation/initialization from functions which may cause errors. As such I feel like the `try_from(..)`, despite its awkward definition better conveys that we're doing something which has a change of failing. Is there a potential way/trait/pattern that may satisfy both objectives?

drv/fpga-app-api/src/lib.rs
mkeeter: Maybe `fpga-user-api`, since we use "user design" to refer to the running code?

drv/sidecar-seq-server/src/front_io.rs
mkeeter: Gotcha, I don't feel too strongly about this

task/monorail-server/src/bsp/sidecar_bc.rs
mkeeter: 👍🏻  seems fine to me

drv/sidecar-mainboard-controller/src/tofino2.rs
mkeeter: Having `new(...)` be fallible isn't _totally_ unheard of, e.g. [`NonZeroU8::new`](https://doc.rust-lang.org/std/num/struct.NonZeroU8.html#method.new) returns an `Option<Self>`.

It does appear to be uncommon, though.  What about making an infallible `new`
```rust
impl TofinoPowerRail {
    fn new(id: TofinoPowerRailId, status: PowerRailStatus, pins: PowerRailPinState) -> Self {
        // ...
    }
}
```
then doing the type conversions in the loop at [`tofino2.rs:292`](https://github.com/oxidecomputer/hubris/pull/1449/files#diff-580ceb53b19ee484f37faf79894c68765ef0a5ed11d39fa7b9d50641f2c3ce41R293)?  That appears to be the only place `try_from` is used.

arjenroodselaar: Alright, fair point. This only used in one place so we may as well pull the loop into that function. There's no need really to implement this in a separate `TryFrom`.

mkeeter (approved):

PR #1452, opened 2023-07-05
Only read CFPA headers if they are programmed

### LINKED ISSUE #1450: RoT image status reporting fails for erased firmware slots
In manufacturing, we flashed slot A and then attempted to get the boot state. We instead got `EccDoubleErr`. This is the error returned by the flash hardware if we read an erased sector using the conservative indirect read procedure I implemented in the flash driver.

It'd be neat if we matched on that in the server and reported "this slot is erased" instead.

mkeeter:
This is an attempt to fix #1450.

I have tested this on my benchtop system (with the PING / PONG / SCRATCH pages all written), and it works there.

Unfortunately, I don't have a sufficiently blank RoT on my bench to test with (see https://github.com/oxidecomputer/hubris/issues/1450#issuecomment-1614746787).

### DIFF
--- a/drv/lpc55-update-server/src/main.rs
+++ b/drv/lpc55-update-server/src/main.rs
@@ -430,65 +430,28 @@ impl ServerImpl<'_> {
     ) -> Result<(u32, u32), UpdateError> {
         // Read the two versions. We do this with smaller buffers so
         // we don't need 2x 512B buffers to read the entire CFPAs.
-        let ping_header =
-            self.read_cfpa_header_if_programmed(CFPA_PING_FLASH_WORD)?;
-        let pong_header =
-            self.read_cfpa_header_if_programmed(CFPA_PONG_FLASH_WORD)?;
+        let mut ping_header = [0u32; 4];
+        let mut pong_header = [0u32; 4];
 
-        match (ping_header, pong_header) {
-            (Some(ping_header), Some(pong_header)) => {
-                // Work out where to read the authoritative contents from.
-                let val = if ping_header[1] >= pong_header[1] {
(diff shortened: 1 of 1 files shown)


drv/lpc55-update-server/src/main.rs
```diff
@@ -469,17 +506,13 @@ impl ServerImpl<'_> {
...
-            core::slice::from_mut(&mut scratch_header),
-        )?;
+        let scratch_header =
+            self.read_cfpa_header_if_programmed(CFPA_SCRATCH_FLASH_WORD)?;
 
         // We only have a pending preference if the scratch CFPA page is newer
         // than the authoritative page.
         let pending_persistent_boot_preference =
-            if scratch_header[1] > cfpa_version {
+            if scratch_header.map(|s| s[1] > cfpa_version).unwrap_or(false) {
```
mx-shift, line 515: `Option::is_some_and` is new in 1.70.  Sigh.

mkeeter:
I switched to a much simpler implementation, thanks to the insights that
- Ping and pong pages must be okay, because Bootleby reads them unconditionally
- We can check for an erased page by looking for `EccDoubleErr`

The new code adds this check specifically to the scratch page read, so it's a much smaller diff from `master`.

mx-shift (approved):

mkeeter:
Tested by erasing the CFPA scratch page with
```
lpc55_flash /dev/tty.usbserial-A506YUL2 flash-erase-region 0x9DE00 512
```

An older commit fails with the expected `Err(Update(EccDoubleErr))`; this branch works as expected.

PR #1454, opened 2023-07-07
Sidecar: improve PCIe hotplug and fault behavior

arjenroodselaar:
This diff does the following:

- Set PCIe reset options to reset the PHY lanes on specific link events
- Add more control over the PCIe Power Fault sideband signal

The power fault sideband signal can now be set to track the sequencer error state in the mainboard controller, generating an interrupt in Helios when the sequencer trips and powers down Tofino. This interrupt will be used to clean up state both in the kernel and Dendrite in preparation for the next attachment sequence.

The power fault interrupt behavior has been tested with a Gimlet, but more work on the Helios side is needed to make this work end-to-end.


drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -544,6 +564,47 @@ bitfield! {
...
+    }
+}
+
+impl From<u32> for TofinoPcieResetOptions {
+    fn from(v: u32) -> Self {
+        match v {
+            0b10 => TofinoPcieResetOptions::ControllerAndPhyLanes,
+            0b11 => TofinoPcieResetOptions::EntirePhyAndController,
+            _ => TofinoPcieResetOptions::ControllerOnly, // Safe default.
+        }
```
mkeeter, line 587: We can piggyback on the `FromPrimitive` derivation:
```suggestion
        Self::from_u32(v).unwrap_or(TofinoPcieResetOptions::ControllerOnly)
```

arjenroodselaar: Nice, thanks!

Aaron-Hartwig:
Just for linkage, the associated FPGA PR here is https://github.com/oxidecomputer/quartz/pull/110

drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -474,6 +494,43 @@ bitfield! {
     pub pcie_lanes, set_pcie_lanes: 7, 4;
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, From, FromPrimitive, AsBytes)]
+#[repr(u32)]
+pub enum TofinoPcieResetOptions {
```
mkeeter, line 499: Add a comment explaining where these values come from (Tofino register definitions?)

mkeeter (approved):

PR #1464, opened 2023-07-18
Fix 10G watchdog checks

mkeeter:
To avoid spurious calls to `Monorail.reinit`, the 10G link watchdog checks whether the PCIE host reset bit is set.

Unfortunately, this bit changed from bit 0 to bit 1 in #1454; this broke the VSC7448 side of the watchdog, resulting in the link staying down if `dendrite` restarted on the Scrimlet.

This PR switches to using the bit's canonical definition in our register file, which is less likely to be wrong.


bnaecker (approved):
excellent

PR #1465, opened 2023-07-18
Add U.2 error handling

mkeeter:
Previously, devices in the thermal loop were either removable or not, marked by a boolean flag.

This PR adds a third type of device (`RemovableAndErrorProne`) and changes that flag into an `enum`.

The new type of device is applicable to our U.2s: it means that upon _any I2C error_, we treat the device as **safely absent**.  Previously, only a `NoDevice` error would treat the device as safely absent; this remains the case for devices marked `Removable`.

### DIFF
--- a/task/thermal/src/bsp/gimlet_bcd.rs
+++ b/task/thermal/src/bsp/gimlet_bcd.rs
@@ -6,7 +6,8 @@
 
 use crate::{
     control::{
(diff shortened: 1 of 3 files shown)


bcantrill (approved):
This looks terrific -- I love the approach here, as it feels like this will not be the last removable device that may develop a bad attitude!

PR #1467, opened 2023-07-26
CI Rework

labbott:


.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
+name: release
+on:
+  workflow_call:
+    inputs:
+      build:
+        description: "Machine we are releasing for"
+        required: true
+        type: string
+      app_name:
+        description: "Machine we are releasing for"
```
steveklabnik, line 10: these descriptions probably should be updated to be accurate, obviously not an issue right this moment

.github/workflows/dist.yml
```diff
@@ -32,7 +31,7 @@ jobs:
             app_name: lpc55xpresso
             app_toml: app/lpc55xpresso/app.toml
             target: thumbv8m.main-none-eabihf
-            image: [a, b]
+            image: "a, b"
```
steveklabnik, line 34: i am interested in this change, I'm not sure why it's being done

steveklabnik (commented):
Neat, I haven't used the feature where workflows let you call other workflows yet, but this makes sense to me!

.github/workflows/dist.yml
labbott: Reusable workflows can only handle inputs of boolean, number or string https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype so I converted this to a comma separated string. It turns out we were already doing this in the build anyway.

steveklabnik: Ah, interesting, thank you!

steveklabnik (approved):

Issue #1264, opened 2023-04-05
gimlet: want inventory accessibility over IPCC


rmustacc:
As laid out in RFD 360 and others, there is a large need for us to be able to reliably capture the inventory of components in a Gimlet and various data like firmware revisions of 3rd party components. For each component we're ultimately going to be looking for:

* Information to uniquely identify it. While there is both unique information such as a vendor id, part number, and serial, there is more to this as we may have two Micron SPI flashes for each HBS spi flash so we also want to understand what identifying information relates to which. This also comes up when we need to enumerate sensors but don't have much in terms of self-describing device identity.
* Many devices will have some kind of firmware as part of them. Examples here include the IBC and Renesas power controllers. For all such devices that we aren't programming each time at boot, we need that information.
* Many devices have sensors that we want to be made available. I suspect this should come as a subsequent change as we're going to want to have the general sensor cache available, but I'll try to note these out for each component in the following list. Critically we aren't asking for the sensor data in this cache, but rather to get enough data to know which are available and how to phrase this to someone who wants to access it over the management network.

Here is the list of items that we want to be present. For more details, please see [RFD 360 §2 Device Identification](https://rfd.shared.oxide.computer/rfd/0360#sec-id).

- [ ] Gimlet Fan Tray FRUID
  - [ ]  We also are going to want to figure out how to phrase both tachometer and presence sensors as appropriate.
- [ ] 10x Sharkfin FRUID
- [ ] Gimlet PCBA FRUID
- [ ] Full per-DIMM SPD data. This does not need to be decoded, we can take the full 512-byte payload. However, we do need to know the slot that this information corresponds to.
- [ ] H753
  - [ ] 96-bit device ID register at `0x1ff1_e800` (§61.1 [[stm32h7-ref]](https://rfd.shared.oxide.computer/rfd/0360#stm32h7-ref)).
  - [ ] Revision ID and part number from `DBGMCU_IDC` (§60.5.8 [[stm32h7-ref]](https://rfd.shared.oxide.computer/rfd/0360#stm32h7-ref))
  - [ ] Optional: Consider grabbing the ROM API version, though since we don't currently use it it's less important.
- [ ] LPC55S69 (note this requires plumbing through sprot, see #1262)
  - [ ] 128-bit UUID
  - [ ] Hardware device information via `DEVICE_ID0`, `DIED`, and NMPA metal revision.
[...]

PR #1469, opened 2023-07-27
Device inventory over IPCC

mkeeter:
This PR adds device inventory to the IPCC implementation in `host-sp-comms`

See https://github.com/oxidecomputer/rfd/pull/629 and https://github.com/oxidecomputer/rfd/pull/628 for the relevant RFD changes.

It includes _most_ of the requested devices from #1264, with a few limitations:
- RoT info is not yet plumbed through
- A few devices are not yet implemented
- Inventory data only includes device identity, and **does not** include `SensorId`'s.  Because the reply is extensible, we can add them in future updates.

There's one infrastructure change: we split the `local-vpd` driver into two pieces:
- `local-vpd` continues to support reading the local VPD EEPROM, but it now calls an implementation in...
- `oxide-vpd`, which supports reading _any_ VPD EEPROM (in the Oxide format)


task/host-sp-comms/src/tx_buf.rs
```diff
@@ -170,21 +176,76 @@ impl TxBuf {
...
+
+    pub(crate) fn try_encode_inventory<'a, F>(
+        &mut self,
+        sequence: u64,
+        name: &'a [u8],
+        fill_data: F,
+    ) where
+        F: FnOnce() -> Result<InventoryData, InventoryDataResult>,
+    {
+        let mut name_array = [0u8; 32];
```
andrewjstone, line 191: Is this just a memcpy?

task/packrat/src/gimlet.rs
```diff
@@ -117,12 +117,33 @@ impl GimletData {
...
             .copied()
             .ok_or(RequestError::Fail(ClientError::BadMessageContents))
     }
+
+    pub(crate) fn get_full_spd_data(
+        &self,
+        dev: usize,
+        out: LenLimit<Leased<idol_runtime::W, [u8]>, 512>,
+    ) -> Result<(), RequestError<Infallible>> {
+        if out.len() != spd::MAX_SIZE {
```
andrewjstone, line 134: Could `out.len()` be safely greater than `spd::MAX_SIZE` ?

andrewjstone (commented):
Looks pretty good! It will be a boon to get this in. I know it will make @rmustacc happy. 

One note is that a bunch of function names in header comments need changing to the new names. I didn't bother tagging them all.

app/gimlet/rev-d.toml
```diff
@@ -201,14 +201,14 @@ notifications = ["timer"]
...
 name = "task-host-sp-comms"
-features = ["stm32h753", "uart7", "baud_rate_3M", "hardware_flow_control", "vlan"]
+features = ["stm32h753", "uart7", "baud_rate_3M", "hardware_flow_control", "vlan", "gimlet"]
 uses = ["uart7"]
 interrupts = {"uart7.irq" = "usart-irq"}
 priority = 7
-max-sizes = {flash = 32768, ram = 32768}
-stacksize = 2048
+max-sizes = {flash = 65536, ram = 32768}
+stacksize = 6000
```
labbott, line 209: This is a pretty big jump. Do you know where the extra size is coming from?

lib/host-sp-messages/src/lib.rs
```diff
@@ -175,6 +187,183 @@ pub enum KeyLookupResult {
...
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
+)]
+pub enum InventoryData {
+    /// Raw DIMM data
+    #[serde(with = "BigArray")]
+    DimmSpd([u8; 512]),
+
+    /// Device or board identity, typically stored in a VPD EEPROM
+    VpdIdentity(oxide_barcode::VpdIdentity),
```
rmustacc, line 239: Just to make sure we're on the same page here, if for some reason the sizing of the VPD Identity were ever to change beyond the 11/11 characters for the MPN/SN, we would introduce a newer type here. This does make me wonder if we should be passing something here where we use either a larger size than the others and zero-terminate this or have an explicit integer encoding lengths so we're ok to change this in the future.

In particular, the existing `MODEL_LEN` and `SERIAL_LEN` which are used by the `Identity` structure are 51 characters (50 + `\0`), so given that it is already future proofed, I wonder if this should be consistent or not.

rmustacc (commented):
Just a few questions on some of the types as I've stated to plumb up specifics around them.

lib/host-sp-messages/src/lib.rs
```diff
@@ -175,6 +187,183 @@ pub enum KeyLookupResult {
...
+        /// IC_DEVICE_REV, PMBus operation 0xAE
+        ic_device_rev: [u8; 2],
+        /// NVM_CHECKSUM, PMBus operation 0xF0
+        nvm_checksum: u16,
+    },
+
+    /// Fan subassembly identity
+    FanIdentity {
+        identity: oxide_barcode::VpdIdentity,
+        fans: [oxide_barcode::VpdIdentity; 4],
```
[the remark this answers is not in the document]
mkeeter: This is based on the VPD data that you gave me in chat, which has four barcodes under `SASY`:
```
[
    ("FRU0", [
        ("BARC", [
            "0XV1:9910000084:002:BRM42220085",
        ]),
        ("SASY", [
            ("BARC", [
                "0XV1:9130000022:001:BRM39220046",
            ]),
            ("BARC", [
                "0XV1:9910000094:001:BRM42220005",
            ]),
            ("BARC", [
                "0XV1:9910000094:001:BRM42220091",
            ]),
            ("BARC", [
                "0XV1:9910000094:001:BRM42220072",
            ]),
        ]),
    ]),
]
```
Looking more closely, I'm guessing three of them are fans (9910000094), and one is something else (9130000022)?

drv/oxide-vpd/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        TlvcReader::begin(eeprom_reader).map_err(VpdError::ErrorOnBegin)?;
+
+    // Find the root chunk, translating from a general to specific error
+    let mut chunk =
+        get_chunk_for_tag(reader, *b"FRU0", 0).map_err(|e| match e {
+            VpdError::NoSuchChunk(..) => VpdError::NoRootChunk,
+            e => e,
+        })?;
+
+    // Iterate over our tag list, finding inner chunks
```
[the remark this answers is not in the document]
mkeeter: Yes, that's intentional.  We're looking for a single chunk, with a nested path, e.g.
```ron
[
    ("FOO", [ [...] ]),
    ("TAG1", [
        ("TAG2", [ [...] ]),
    ])
]
```
If we're looking for `["TAG1", "TAG2"]`, then we'll first select the chunk
```
    ("TAG1", [
        ("TAG2", [ [...] ]),
    ])
```
and then 
```
        ("TAG2", [ [...] ]),
```
on repeated iterations through the loop.

drv/oxide-vpd/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+/// ```ron
+/// ("FRU0", [
+///     ("TAG1", [ [...] ]),
+///     ("TAG2", [ [...] ]),
+///     ("TAG3", [ [...] ]),
+/// ])
+/// ```
+/// (where `TAG*` are example tags)
+///
+/// `read_config` should be called with a tag nested under `FRU0` (e.g. `TAG1`
```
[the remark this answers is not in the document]
mkeeter: Yup, done in baab9d5d2960d125c6cc6eb72367612b91d2903b

drv/oxide-vpd/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+/// The `FRU0` root is mandatory.
+///
+pub fn read_config_nested_from<V: AsBytes + FromBytes>(
+    eeprom: At24Csw080,
+    tags: &[([u8; 4], usize)],
+) -> Result<V, VpdError> {
+    let mut out = V::new_zeroed();
+    let n = read_config_nested_from_into(eeprom, tags, out.as_bytes_mut())?;
+
+    // `read_config_into()` fails if the data is too large for `out`, but will
```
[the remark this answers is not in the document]
mkeeter: also done in baab9d5d2960d125c6cc6eb72367612b91d2903b

task/packrat/src/gimlet.rs
mkeeter: That would be technically safe, but would suggest an error on the part of the callers; it would be weird to copy data into a subsection of the lease.

andrewjstone (approved):
Thanks for the fixes @mkeeter

app/gimlet/rev-d.toml
mkeeter: Yeah, I had the same reaction.  I think it's that the `InventoryData` type is ~512 bytes (since it holds a full DIMM SPD), and there are a few stack frames that aren't being inlined since we're using lambda functions here.

I knocked about 2K off it by allocating the `InventoryData` once in the outer frame, then using references to it in the lambda functions.

| situation | stack depth |
|-|-|
| baseline |       864 |
| reading DIMM |   1672 |
|reading from packrat | 1776 |
|reading BRM491 | 1768 |
|reading AT24 barcode | 2712 |

I'm mystified as to why reading the AT24 barcode adds a whole 1K of stack depth.  I'm going to look at [`read_at24csw080_id`](https://github.com/oxidecomputer/hubris/blob/7316fe07f5a7f17a5339584a4f178fbe462d9de3/task/host-sp-comms/src/bsp/gimlet_bcd.rs#L512-L534) a little more and see if I can figure it out, but more eyes are welcome here.

mkeeter: Aha: `read_at24csw080` wasn't being inlined, so we had one stack frame in `perform_inventory_lookup` and another in `read_at24csw080_id`, costing us an extra 512 bytes.  Explicitly allocating the data in the caller and passing it by reference brings us down to 2200, which seems reasonable.

lib/host-sp-messages/src/lib.rs
```diff
@@ -175,6 +187,183 @@ pub enum KeyLookupResult {
     MaxResponseLenTooShort,
 }
 
+/// Results for an inventory data request
+///
+/// These **cannot be reordered**; the host and SP must agree on them.
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
+)]
+pub enum InventoryDataResult {
```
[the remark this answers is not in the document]
mkeeter: That's a fair question; it's for ease of serialization / deserialization.

See [RFD 316](https://rfd.shared.oxide.computer/rfd/0316#sp8594hss_commands) for details, but in short, having this be a single `enum` that includes an `Ok` variant makes it constant size, instead of being either `[0]` (for `Ok(())`) or `[1, i]` (for `Err(..)`).

lib/host-sp-messages/src/lib.rs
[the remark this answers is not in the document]
mkeeter: Okay, I made changes in 52d463c976948b0dd8ddf6af9f3e1ca5704ceef8 to include `identity`, `vpd_identity`, and 3 items in the `fans` array.

lib/host-sp-messages/src/lib.rs
mkeeter: It's correct that moving away from `oxide_barcode::VpdIdentity` would require making a new message type for `InventoryData::VpdIdentity` and `InventoryData::FanIdentity`.

Do you think it would make sense to use `host_sp_messages::Identity` (which stores the 51-byte values) instead here?  It's hard to imagine that encoding becoming defunct.

lib/host-sp-messages/src/lib.rs
```diff
@@ -175,6 +187,183 @@ pub enum KeyLookupResult {
...
+    DimmSpd([u8; 512]),
+
+    /// Device or board identity, typically stored in a VPD EEPROM
+    VpdIdentity(oxide_barcode::VpdIdentity),
+
+    /// Host-swap controller
+    Max5970 {
+        voltage: [SensorId; 2],
+        current: [SensorId; 2],
+    },
```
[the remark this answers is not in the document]
mkeeter: My intention was to punt that to a future PR; this is a unused leftover variant that I should probably delete before merging!

lib/host-sp-messages/src/lib.rs
```diff
@@ -175,6 +186,181 @@ pub enum KeyLookupResult {
     MaxResponseLenTooShort,
 }
 
+/// Results for an inventory data request
+///
+/// These **cannot be reordered**; the host and SP must agree on them.
```
jgallagher, line 191: Maybe worth adding a unit test for the serialized values? E.g., `host_to_sp_command_values()` in this file.

lib/host-sp-messages/src/lib.rs
```diff
@@ -175,6 +186,181 @@ pub enum KeyLookupResult {
...
+                InventoryDataResult::DeviceAbsent
+            }
+            _ => InventoryDataResult::DeviceFailed,
+        }
+    }
+}
+
+/// Data payload for an inventory data request
+///
+/// These **cannot be reordered**; the host and SP must agree on them.  New
```
jgallagher, line 226: Ditto adding a unit test for the serialized values, with apologies because constructing values for each variant on this one looks pretty annoying.

task/host-sp-comms/src/bsp/gimletlet.rs
```diff
@@ -0,0 +1,46 @@
...
+            0 => {
+                // U12: the service processor itself
+                // The UID is readable by stm32xx_sys
+                let sys =
+                    drv_stm32xx_sys_api::Sys::from(crate::SYS.get_task_id());
+                let uid = sys.read_uid();
+                let data = InventoryData::Stm32H7 {
+                    uid,
+                    dbgmcu_rev_id: 0, // TODO
+                    dbgmcu_dev_id: 0, // TODO
```
jgallagher: Just double-checking - did you want to address these `TODO`s before merging?

task/host-sp-comms/src/main.rs
```diff
@@ -847,11 +866,30 @@ impl ServerImpl {
...
+                    &SpToHost::KeyLookupResult(KeyLookupResult::Ok),
+                    |buf| {
+                        const_assert!(
+                            MIN_SP_TO_HOST_FILL_DATA_LEN
+                                >= core::mem::size_of::<u32>() * 2
+                        );
+                        hubpack::serialize(buf, &REPLY).unwrap_lite()
+                    },
+                );
+                core::mem::size_of_val(&REPLY)
```
jgallagher, line 887: This is correct, but looks a little weird at first glance: we're using hubpack to serialize, but then returning `mem::size_of_val` for the number of bytes serialized. Those are definitely the same, but could we add an assertion on the value returned by `hubpack::serialize()` to document that?

task/host-sp-comms/src/main.rs
```diff
@@ -1025,6 +1063,16 @@ impl idl::InOrderHostSpCommsImpl for ServerImpl {
     ) -> Result<Status, RequestError<HostSpCommsError>> {
         Ok(self.status)
     }
+
+    fn get_inventory(
```
jgallagher, line 1067: I'm not sure about this idol op - was this a debugging/development aid? If some other task (or even humility) calls this while we're servicing a request from the host, things will get pretty confused, I think.

task/host-sp-comms/src/bsp/gimlet_bcd.rs
```diff
@@ -0,0 +1,674 @@
...
+            }
+            30..=39 => {
+                let (designator, f) =
+                    Self::get_sharkfin_vpd(index as usize - 30);
+                let mut name = *b"____/U7";
+                name[0..4].copy_from_slice(&designator);
+                let mut data = InventoryData::At24csw08xSerial([0u8; 16]);
+                self.read_at24csw080_id(sequence, &name, f, &mut data)
+            }
+            // TODO: Sharkfin HSC?
```
jgallagher, line 125: "do you want to address this before merging" check

task/host-sp-comms/src/bsp/gimlet_bcd.rs
```diff
@@ -0,0 +1,674 @@
...
+            40 => {
+                // U12: the service processor itself
+                // The UID is readable by stm32xx_sys
+                let sys =
+                    drv_stm32xx_sys_api::Sys::from(crate::SYS.get_task_id());
+                let uid = sys.read_uid();
+                let data = InventoryData::Stm32H7 {
+                    uid,
+                    dbgmcu_rev_id: 0, // TODO
+                    dbgmcu_dev_id: 0, // TODO
```
jgallagher: "do you want to address this before merging" check

task/host-sp-comms/src/bsp/gimlet_bcd.rs
```diff
@@ -0,0 +1,674 @@
...
+userlib::task_slot!(SPI, spi_driver);
+
+impl ServerImpl {
+    /// Number of devices in our inventory
+    pub(crate) const INVENTORY_COUNT: u32 = 60;
+
+    /// Look up a device in our inventory, by index
+    ///
+    /// Indexes are assigned arbitrarily and may change freely with SP
+    /// revisions.
```
jgallagher, line 30: We currently require a host reboot any time we're updating the SP, so this seems totally fine. However, if we ever want to support updating the SP _without_ rebooting the host, it could be problematic: if the host caches the set of device indices and uses them to (e.g.) query sensors at runtime, and the SP is updates and reorders the indices, that host cache will now be out of date. Chatted about this briefly - to support that, we probably want to add something that will allow the SP to tell if a host's request for sensor data matches its current sensor list, and return a "your cache is out of date / I've been updated" error if it does not. (To be clear, this is a problem for future us: "updating the SP requires rebooting the host" is fine!)

task/host-sp-comms/src/bsp/gimlet_bcd.rs
```diff
@@ -0,0 +1,674 @@
...
+    ) {
+        // This should be done by the caller, but let's make it obviously
+        // correct (since we destructure it below).
+        *data = InventoryData::At24csw08xSerial([0u8; 16]);
+        let dev = At24Csw080::new(f(I2C.get_task_id()));
+        self.tx_buf.try_encode_inventory(sequence, name, || {
+            let InventoryData::At24csw08xSerial(id) = data
+                else { unreachable!(); };
+            for (i, b) in id.iter_mut().enumerate() {
+                // TODO: add an API to make this a single IPC call?
```
jgallagher, line 562: "do you want to address this before merging" check

task/host-sp-comms/src/bsp/gimlet_bcd.rs
```diff
@@ -0,0 +1,674 @@
...
+            VpdError::ErrorOnBegin(err)
+            | VpdError::ErrorOnRead(err)
+            | VpdError::ErrorOnNext(err)
+            | VpdError::InvalidChecksum(err),
+        ) if err
+            == tlvc::TlvcReadError::User(EepromError::I2cError(
+                ResponseCode::NoDevice,
+            )) =>
+        {
+            // TODO: ringbuf logging here?
```
jgallagher, line 671: "do you want to address this before merging" check

jgallagher (commented):
This looks great! My only concern is the `get_inventory` idol OP - rest of my comments are minor / nitpicks.

task/host-sp-comms/src/bsp/gimletlet.rs
mkeeter: Yes, done in 84a157c8c6ca191fdd8181aee8a00bc6fca01969

task/host-sp-comms/src/bsp/gimlet_bcd.rs
mkeeter: Done in 84a157c8c6ca191fdd8181aee8a00bc6fca01969

task/host-sp-comms/src/main.rs
mkeeter: This was used for debugging (to check stack sizes); I thought that leaving it was innocuous, but you make a good point!  Removed in e83fa92dd0b8b4986f2ec94f50d2e4c38a4e8a36

task/host-sp-comms/src/main.rs
mkeeter: Yeah, I switched to matching existing behavior in f5f02fc2da82d467b4cd48baf051344a109e246c

lib/host-sp-messages/src/lib.rs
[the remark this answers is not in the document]
mkeeter: [deleted](https://www.youtube.com/watch?v=07h0ksKx5sM)

drv/local-vpd/src/lib.rs
```diff
@@ -89,104 +38,23 @@ impl<'a> TlvcRead for EepromReader<'a> {
...
-///     ("TAG2", [ [...] ]),
-///     ("TAG3", [ [...] ]),
-/// ])
-/// ```
-/// (where `TAG*` are example tags)
-///
-/// `read_config` should be called with a tag nested under `FRU0` (e.g. `TAG1`
-/// in the example above).  It will copy the raw byte array (shown as
-/// `[...]`) into `out`, returning the number of bytes written.
+/// Calls into [`read_config_from_into`]; see details in that docstring
```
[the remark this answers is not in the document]
mkeeter: Fixed in 40bb7532d8c2d798d20865404b481ca5575d1042

task/host-sp-comms/src/bsp/gimlet_bcd.rs
mkeeter: Nah, deleted the comment

task/host-sp-comms/src/bsp/gimlet_bcd.rs
mkeeter: No, I think it's fine as-is

task/host-sp-comms/src/bsp/gimlet_bcd.rs
mkeeter: Not in this PR, no; the HSCs will be relevant for the future PR that adds `SensorId` to all of our types

lib/host-sp-messages/src/lib.rs
mkeeter: Done in 6cf4d21a8969e51f5c56862b00e4b6af6c13f2e3

lib/host-sp-messages/src/lib.rs
mkeeter: Done in 6cf4d21a8969e51f5c56862b00e4b6af6c13f2e3 (and yes, it was annoying!)

jgallagher (approved):

PR #1471, opened 2023-07-31
attest: Add function to record a measurement observed by the platform.

flihp:
This function is *not* idempotent: if the 'record' function is called with the same parameters repeatedly the measurement log will eventually fill up resulting in an error. The measurement log is just an array of measurements (aka [u8; 32]) that is managed in an ArrayVec. The 'record' function takes an enum parameter identifying the hash algorithm and a lease holding the digest. We pass the digest as a lease partially because humility / hiffy can't pass arbitrary arrays as parameters, but also because it's possible that in the future we'll need to support digests with varying lengths.

### DIFF
--- a/task/attest/src/main.rs
+++ b/task/attest/src/main.rs
@@ -11,14 +11,17 @@
 
 mod config;
 
-use attest_api::AttestError;
+use arrayvec::ArrayVec;
+use attest_api::{AttestError, HashAlgorithm};
 use config::DataRegion;
 use core::slice;
+use crypto_common::{typenum::Unsigned, OutputSizeUser};
 use hubpack::SerializedSize;
 use idol_runtime::{ClientError, Leased, RequestError, W};
 use lib_dice::{AliasData, CertData};
 use ringbuf::{ringbuf, ringbuf_entry};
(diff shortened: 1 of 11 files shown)


flihp:
gah, just realized I left a bunch of ringbuf entries in here that should probably be removed.

task/attest/src/main.rs
```diff
@@ -70,16 +75,57 @@ fn load_data_from_region<
...
 }
 
+// the size of the measurements we record
+// NOTE: the rust crypto digest traits don't expose consts for the lengths
+// of various hash functions
+const SHA3_256_DIGEST_SIZE: usize =
+    <Sha3_256Core as OutputSizeUser>::OutputSize::USIZE;
+
+// the number of Measurements we can record
+const CAPACITY: usize = 16;
```
labbott, line 85: Is this carefully calculated or "yeah that should be enough"

labbott (approved):

task/attest/src/main.rs
flihp: the later for the most part, still in the 'educated guess' phase

PR #1476, opened 2023-08-07
Make `cargo test --workspace` work

mkeeter:
This PR adds the standard stanza of
```toml
[lib] # or [[bin]]
test = false
doctest = false
bench = false
```
to _everything_ that can't compile on the host.

This lets us run `cargo test --workspace` in CI, instead of hand-specifying packages.

### DIFF
--- a/lib/fixedmap/Cargo.toml
+++ b/lib/fixedmap/Cargo.toml
@@ -2,7 +2,3 @@
 name = "fixedmap"
 version = "0.1.0"
 edition = "2021"
-
-[lib]
-test = false
-bench = false
--- a/sys/abi/Cargo.toml
+++ b/sys/abi/Cargo.toml
@@ -9,7 +9,3 @@ bitflags = { workspace = true }
 byteorder = { workspace = true }
 serde = { workspace = true }
 phash = { path = "../../lib/phash" }
-
-[lib]
-test = false
-bench = false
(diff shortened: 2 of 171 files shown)


lib/fixedmap/Cargo.toml
```diff
@@ -2,7 +2,3 @@
 name = "fixedmap"
 version = "0.1.0"
 edition = "2021"
-
-[lib]
-test = false
-bench = false
```
labbott, line 8: This got removed?

sys/abi/Cargo.toml
```diff
@@ -9,7 +9,3 @@ bitflags = { workspace = true }
 byteorder = { workspace = true }
 serde = { workspace = true }
 phash = { path = "../../lib/phash" }
-
-[lib]
-test = false
-bench = false
```
labbott, line 15: Here as well?

lib/fixedmap/Cargo.toml
mkeeter: Yup, I removed the stanza from things that can compile on the host – we may want to add tests to `fixedmap` in the future!

sys/abi/Cargo.toml
mkeeter: Same here, these are pure data types and can build on the host without issue (and I could see us using them in tests eventually)

flihp (approved):
`cargo test --workspace` works great on my "host"

PR #1477, opened 2023-08-07
Add `toml-patch` crate for improved inheritance

mkeeter:
Right now, Hubris has a hard-coded inheritance system that allows for one thing: adding new features to existing tasks.

The current system is not flexible enough for our current set of boards, and we've ended up with a huge amount of duplication.  For Gimlet alone, we have `rev-{b,c,d}{'',-dev,-lab,}`, i.e. 9 variations, of which 6 are mostly copy-pasta.

This PR adds a more general form of TOML inheritance, implemented in a new `toml-patch` crate.  For those familiar with the Yocto project, it's similar to [`.bbappend` files](https://docs.yoctoproject.org/1.6/dev-manual/dev-manual.html#using-bbappend-files):

- TOML files can declare that they inherit from one or more other files, and also contain local data
- Inheritance is processed in order (with local data being applied last)
- Basic data types (e.g. integers, strings) are replaced with new values
- Sequence data types (lists, tables, inline tables; including the root table) have new values appended to the end

Before this PR, we shipped both `app.toml` and `patches.toml` in the Hubris archive, and Humility had special code to apply patches itself.  After this PR, we flatten everything into a single canonical `app.toml` and only include that in the archive.  This means that Humility doesn't have to know about the new patching system at all; it just sees an `app.toml`.

The hard part about this was an unexpected issue with `toml_edit`: that crate has a _global order_ for TOML tables, which controls the order in which they're printed.  This means that to insert new tables at the end position, we have to
- Modify the original document to add gaps
- Modify the incoming document, shifting tables into the correct position
- _Then_, splice them together

It gets even tricker when lists of tables (which may contain sub-tables of their own) are involved!

Anyways, that's all implemented in the `toml-patch` crate, along with some unit tests.

Note that this PR does not actually deduplicate all of our existing manifests!  That's coming in a later PR, which will require careful testing to make sure it doesn't break anything.  This PR _does_ change a few existing manifests that use inheritance, but it's a small, relatively-safe change that I will manually verify (using CI artifacts).


build/toml-patch/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use indoc::indoc;
+
+    fn patch_and_compare(a: &str, b: &str, out: &str) {
```
steveklabnik, line 272: love this style of testing, guessing you read that matklad blog post too haha

build/toml-patch/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+            "#},
+            indoc! {r#"
+                age = 38
+            "#},
+            indoc! {r#"
+                name = "foo"
+                age = 38
+            "#},
+        );
+        patch_and_compare(
```
steveklabnik, line 299: any reason that these are all in one big test instead of each being their own test? im guessing this runs fast enough that like, it doesn't *really* matter either way, I'm just curious :)

build/xtask/src/config.rs
```diff
@@ -719,3 +665,72 @@ impl BuildConfig<'_> {
...
+fn read_and_flatten_toml(
+    cfg: &Path,
+    hasher: &mut DefaultHasher,
+    seen: &mut BTreeSet<PathBuf>,
+) -> Result<toml_edit::Document> {
+    use toml_patch::merge_toml_documents;
+
+    // Prevent diamond inheritance
+    if !seen.insert(cfg.to_owned()) {
+        bail!("{cfg:?} is inherited more than once.  Don't do that.");
```
steveklabnik, line 680: this is an amusing error but runs afoul of the "one sentence all lower case no periods" rule of thumb for error messages. would keep it to just the first sentence.

steveklabnik (approved):
Seems good! I have a few comments and questions but would be happy to see this land as-is as well.

build/toml-patch/src/lib.rs
mkeeter: I haven't, actually!  I see a few, which ones are recommended reading?

build/xtask/src/config.rs
mkeeter: Clarified in 5fc61b05e6a1f289db4862c7fcf603de62805ce3

I decided to explicitly say "diamond dependencies are not allowed" in the error messages, since just saying "X is inherited more than once" doesn't tell you why that's bad 🙃

mkeeter:
Comparing the manifests between this CI build and the most recent `master`, I see changes to Git commit and image ID, but nothing else:
```
bash-3.2$ for f in $(ls build-*); do humility -a $f manifest > ours.txt; humility -a ../changed/$f manifest > theirs.txt; echo "$f"; diff ours.txt theirs.txt; echo "------------"; done
build-gimlet-b-lab-image-default.zip
2,3c2,3
<      git rev => f0077197b0bb5f21297c6235e978fd81bb84d239
<     image id => [96, a5, 26, 86, 77, 9d, f9, e1]
---
>      git rev => f7e13851d41885fa105b4e40fea092256b05f676
>     image id => [59, cc, 1b, 62, 2, e2, 65, 77]
------------
build-gimlet-c-lab-image-default.zip
2,3c2,3
<      git rev => f0077197b0bb5f21297c6235e978fd81bb84d239
<     image id => [4d, 47, e, 2c, aa, c1, d7, b3]
---
>      git rev => f7e13851d41885fa105b4e40fea092256b05f676
>     image id => [f6, a7, 41, 29, 92, 22, e8, fa]
------------
build-gimlet-d-lab-image-default.zip
2,3c2,3
<      git rev => f0077197b0bb5f21297c6235e978fd81bb84d239
<     image id => [ce, 73, 7f, f1, 7e, 76, d2, 21]
---
>      git rev => f7e13851d41885fa105b4e40fea092256b05f676
>     image id => [e5, f0, c2, e6, cb, 1e, f4, e3]
------------
build-sidecar-b-lab-image-default.zip
2,3c2,3
<      git rev => f0077197b0bb5f21297c6235e978fd81bb84d239
<     image id => [a0, f7, f5, b3, 3f, 8c, 4e, 6c]
---
>      git rev => f7e13851d41885fa105b4e40fea092256b05f676
>     image id => [36, d5, 71, 9d, 5, 74, 79, 78]
------------
build-sidecar-c-lab-image-default.zip
2,3c2,3
<      git rev => f0077197b0bb5f21297c6235e978fd81bb84d239
<     image id => [61, 75, 36, 6b, b7, a4, 4, 2b]
---
>      git rev => f7e13851d41885fa105b4e40fea092256b05f676
>     image id => [f2, 43, 5b, 23, 7b, f6, e1, 49]
```

build/toml-patch/src/lib.rs
steveklabnik: This post specifically advocates doing exactly this https://matklad.github.io/2021/05/31/how-to-test.html

build/xtask/src/config.rs
```diff
@@ -719,3 +665,75 @@ impl BuildConfig<'_> {
...
+    hasher.write(&cfg_contents);
+
+    let cfg_contents = std::str::from_utf8(&cfg_contents)
+        .context("failed to read manifest as UTF-8")?;
+
+    // Additive TOML file inheritance
+    let mut doc = cfg_contents
+        .parse::<toml_edit::Document>()
+        .context("failed to parse TOML file")?;
+    if let Some(inherited_from) = doc.remove("inherit") {
```
jgallagher, line 699: Trivial nit - since this block ends with an "early" return (followed immediately by the normal return), I might use `if let` to bail out early in the "no inheritance" case, just to reduce the scope of the remainder:

```rust
let Some(inherited_from) = doc.remove("inherit") else {
    // No TOML inheritance involved; return the original document.
    return Ok(doc);
};
```

build/toml-patch/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+use toml_edit::{visit::Visit, visit_mut::VisitMut};
+
+pub fn merge_toml_documents(
+    original: &mut toml_edit::Document,
+    mut patches: toml_edit::Document,
+) -> Result<()> {
+    // Find offsets where we need to insert gaps for incoming patches
+    let mut offsets = BTreeMap::new();
+    compute_offsets(original, &patches, &mut offsets)?;
+    assert!(!offsets.contains_key(&0));
```
jgallagher, line 26: Probably worth a comment what's going on here (what's special about 0 that we need it to be there but not be put there by `compute_offsets`?).

jgallagher (approved):

build/toml-patch/src/lib.rs
mkeeter: This is for positions _before_ the first offset in the table.  I moved it to an `unwrap_or((0, 0))` instead in be0baefc3d9c8f94dcd8138a55dd7b97244e4fc1

build/xtask/src/config.rs
mkeeter: Sure, done in 94bbd35474bff499551c8c24b58b05bbc6e1f395

PR #1481, opened 2023-08-08
Add name for `tests` workflow

mkeeter:
For some reason, you need a secondary `name` for this check to show up as a branch protection option.

(I also decapitalized "Tests" to match our other CI workflows)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: tests
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -11,9 +11,8 @@ env:
 
 jobs:
   build:
-
+    name: tests
```
labbott, line 14: Not going to bikeshed this to death but "hubris-tests" might be more specific. Up to you if you think we're likely to grow other tests.

labbott (approved):

.github/workflows/test.yml
mkeeter: Eh, I think this is fine for now since all the other CI steps are also short (`dist`, `formatting`, `documentation`)

PR #1482, opened 2023-08-08
WIP sp_measure: record measurement in `attest` task

flihp:
This is still a work in progress but the basic functionality has been tested successfully on an rot-carrier. This still needs to be integrated w/ the `oxide-rot-1` image and tested on a gimlet before merging.


app/rot-carrier/app.toml
```diff
@@ -195,12 +198,9 @@ task-slots = ["gpio_driver", "swd", "update_server"]
 name = "task-sp-measure"
 priority = 6
 max-sizes = {flash = 131072, ram = 8192}
```
labbott, line 200: You almost certainly will be able to drop the flash size, I had to increase it for including part of the SP image to compare

labbott (commented):
Does it make sense to keep this in a separate task if this is all it's going to do? I know separation yadda yadda but there's overhead to have another task, especially if we're going to be pulling in sha256 each time

flihp:
> Does it make sense to keep this in a separate task if this is all it's going to do? I know separation yadda yadda but there's overhead to have another task, especially if we're going to be pulling in sha256 each time

That's still the primary driver. Probably worth writing a quick RFD to document whatever decision we make since this is a threat modeling exercise.

PR #1484, opened 2023-08-10
Fix designator for fan VPD EEPROM itself

mkeeter:


rmustacc (approved):
This looks good and I've verified it working at my end. This is identical to what I had prototyped last night.

PR #1487, opened 2023-08-11
Rework release tagging

labbott:
We need to have more than one binary in a release. Split up the release into reasonable groups.


steveklabnik (approved):
looks great to me, much tidier to have each individual grouping!

PR #1491, opened 2023-08-14
xtask: Include expected measurement for hubris images in build archive.

flihp:
This implementation is currently limited to stm32h7 images for the SP. This was tested on a `gimletlet` by comparing the FWID value calculated by `xtask` with the value calculated by the `sp_measure` `build.rs`.


build/xtask/src/dist.rs
```diff
@@ -549,6 +555,53 @@ pub fn package(
...
+    // determine FWID calculation method from chip (directory) name
+    let pad = match chip_name.file_name().and_then(OsStr::to_str) {
+        Some("stm32h7") => {
+            // all unprogrammed flash is read as 0xff
+            Some(vec![
+                0xff_u8;
+                flash.end as usize - flash.start as usize - bin.len()
+            ])
+        }
+        Some(_) => None,
```
flihp, line 583: this should probably `bail!`: the `app.toml` explicitly enabled the `fwid` calculation but this function doesn't know how to calculate it for the relevant chip

build/xtask/src/dist.rs
```diff
@@ -549,6 +555,53 @@ pub fn package(
...
+    if let Some(pad) = pad {
+        sha.update(&pad);
+    }
+
+    let digest = sha.finalize();
+
+    let mut file = File::create(&cfg.img_file("final.fwid", image_name))?;
+    writeln!(file, "{}", hex::encode(&digest))?;
+
+    archive.add_file("img/final.fwid", digest.as_bytes())?;
```
labbott, line 599: I'm not quite sure why we're creating the file twice, is the hubris one not sufficient? Aren't the two files different since one is hex encoded and the other is the raw bytes?

flihp: Ah that's a good catch: I had intended for the contents of these two files to be identical. I'll fix that up. As to why I create a file on disk and one in the archive: the `final.bin` is manipulated / caboosed in the archive after creation.

flihp: Just pushed a change to make the `final.fwid` files on disk and in the archive identical: both now contain the hex encoded digest.

build/xtask/src/dist.rs
flihp: resolved in last push

build/xtask/src/dist.rs
```diff
@@ -549,6 +555,70 @@ pub fn package(
     Ok(allocated)
 }
 
+// can we use lpc55-romapi
+const LPC55_FLASH_PAGE_SIZE: usize = 512;
```
labbott, line 559: Hmmmm, I think what we actually want is like `lpc55-rom-constants` or something similar (currently running into this with some of the CMPA/CFPA reading as well). Probably not worth blocking for it.

build/xtask/src/config.rs
```diff
@@ -27,6 +27,8 @@ struct RawConfig {
     epoch: u32,
     #[serde(default)]
     version: u32,
+    #[serde(default)]
+    fwid: bool,
```
labbott, line 31: Would it be easier to make this `no_fwid` so only boards that we probably aren't going to care about can opt out? I was trying to go through the mental list to make sure we hit all the boards we care about.

build/xtask/src/dist.rs
```diff
@@ -549,6 +555,70 @@ pub fn package(
...
+    if pad != 0 {
+        sha.update(vec![0xff_u8; pad])
+    }
+
+    let digest = sha.finalize();
+
+    // after we've appended a newline fwid is immutable
+    let mut fwid = hex::encode(&digest);
+    writeln!(fwid, "").context("appending newline to FWID")?;
+    let fwid = fwid;
```
labbott, line 607: Is the newline just for ease of reading or is there some deeper meaning?

flihp: No deeper meaning, just gives better output when the file is dumped to the console.

build/xtask/src/dist.rs
flihp: I've added a new commit addressing this feedback. Only the constant used here by `xtask` was moved to the new `lpc55-rom-data` crate. Others can migrate on demand. I'll squash all of this down once the PR is ready. LMK if this wasn't what you were after.

build/xtask/src/config.rs
flihp: I went back and forth on this a few times. The current approach, as you point out, is error prone & requires we update more toml files. Doing this with negative logic would require fewer toml changes but might end up on the wrong side of the law of least surprise when folks add new boards and they get build errors until they figure out that they need to add `fwid = false`. There are good reasons for going both ways. I'm happy to defer to your judgement.

labbott: Eh, let's keep it as is for now.

labbott (approved):

PR #1493, opened 2023-08-14
Fix #750 by maintaining the Rx tail pointer

### LINKED ISSUE #750: `net` can become wedged and stop receiving packets
Yesterday while testing host boot flash operations triggered by `mgmt-gateway`, @mkeeter and I ran into two bugs:

1. [HostFlash.bulk_erase](https://github.com/oxidecomputer/hubris/blob/24042d0c134f69460ba02def77822c8af89ec2be/drv/gimlet-hf-server/src/main.rs#L198-L207) calls `poll_for_write_complete` after issuing the command to the qspi driver, which [busy waits](https://github.com/oxidecomputer/hubris/blob/24042d0c134f69460ba02def77822c8af89ec2be/drv/gimlet-hf-server/src/main.rs#L390-L398). I've seen bulk erase take > 1 minute on a gimlet, and the comments say it can take as long as 8 minutes; during that time, the `hf` task starves all lower priority tasks (including `net`).
2. After triggering one of these long busy waits, when it returned, `net` was in a strangely wedged state. Client tasks could still send packets (we were seeing `udpbroadcast` data), but as far as we could tell no packets were being received. `udpecho` was not receiving notifications of incoming packets, and `net` was not responding to NDP messages.

Fixing 1 should be straightforward, but I haven't done so yet because I wanted to address 2 first. I took a humility dump of the SP while `net` was in this state (attached to the first comment on this issue). After restarting `net` via jefe, it came back healthy.

I've pushed a [debug-net-wedge](https://github.com/oxidecomputer/hubris/tree/debug-net-wedge) that can relatively reliably wedge the `net` task in the same way on my gimletlet. It adds two tasks:

1. `busywait` runs at a relatively high priority. When it receives a message, it busy waits for 5,000 ticks and then responds.
2. `udpecho-slowly` acts like `udpecho`, _unless_ the packet it receives begins with an ASCII `1`. For such packets, it first messages `busywait` (producing the same situation we saw with hostflash starving net).

The branch also has a `debug-net-client` crate that spins messaging `udpecho-slowly` forever. To run it, cd into its directory and then

```
RUST_LOG=info cargo run -- '[fe80::c1d:7dff:feef:9f1d%2]:7777'
```

but with the address of your gimletlet. When everything is working, the output looks similar to this:

```
[2022-09-02T20:16:00Z INFO  debug_net_client] sending packet to trigger high priority busy loop
[2022-09-02T20:16:00Z INFO  debug_net_client] sending followup packet 1
[2022-09-02T20:16:00Z INFO  debug_net_client] sending followup packet 2
[2022-09-02T20:16:00Z INFO  debug_net_client] sending followup packet 3
[2022-09-02T20:16:01Z INFO  debug_net_client] sending followup packet 4
[2022-09-02T20:16:01Z INFO  debug_net_client] sending followup packet 5
[2022-09-02T20:16:01Z INFO  debug_net_client] sending followup packet 6
[2022-09-02T20:16:02Z INFO  debug_net_client] sending followup packet 7
[2022-09-02T20:16:02Z INFO  debug_net_client] sending followup packet 8
[2022-09-02T20:16:02Z INFO  debug_net_client] sending followup packet 9
[2022-09-02T20:16:03Z INFO  debug_net_client] no response after 1.001421583s
[2022-09-02T20:16:04Z INFO  debug_net_client] no response after 2.002848957s
[2022-09-02T20:16:05Z INFO  debug_net_client] received response 1 from [fe80::c1d:7dff:feef:9f1d%2]:7777
[2022-09-02T20:16:05Z INFO  debug_net_client] received response 2 from [fe80::c1d:7dff:feef:9f1d%2]:7777
[2022-09-02T20:16:05Z INFO  debug_net_client] received response 3 from [fe80::c1d:7dff:feef:9f1d%2]:7777
[2022-09-02T20:16:05Z INFO  debug_net_client] received response 4 from [fe80::c1d:7dff:feef:9f1d%2]:7777
[2022-09-02T20:16:06Z INFO  debug_net_client] no response after 3.289892797s
[2022-09-02T20:16:06Z INFO  debug_net_client] sending packet to trigger high priority busy loop
[2022-09-02T20:16:06Z INFO  debug_net_client] sending followup packet 1
[2022-09-02T20:16:06Z INFO  debug_net_client] sending followup packet 2
[2022-09-02T20:16:06Z INFO  debug_net_client] sending followup packet 3
[2022-09-02T20:16:07Z INFO  debug_net_client] sending followup packet 4
[2022-09-02T20:16:07Z INFO  debug_net_client] sending followup packet 5
[2022-09-02T20:16:07Z INFO  debug_net_client] sending followup packet 6
[2022-09-02T20:16:08Z INFO  debug_net_client] sending followup packet 7
[2022-09-02T20:16:08Z INFO  debug_net_client] sending followup packet 8
[2022-09-02T20:16:08Z INFO  debug_net_client] sending followup packet 9
[2022-09-02T20:16:09Z INFO  debug_net_client] no response after 1.001229534s
[2022-09-02T20:16:10Z INFO  debug_net_client] no response after 2.003001465s
[2022-09-02T20:16:11Z INFO  debug_net_client] received response 1 from [fe80::c1d:7dff:feef:9f1d%2]:7777
[2022-09-02T20:16:11Z INFO  debug_net_client] received response 2 from [fe80::c1d:7dff:feef:9f1d%2]:7777
[2022-09-02T20:16:11Z INFO  debug_net_client] received response 3 from [fe80::c1d:7dff:feef:9f1d%2]:7777
[2022-09-02T20:16:11Z INFO  debug_net_client] received response 4 from [fe80::c1d:7dff:feef:9f1d%2]:7777
[2022-09-02T20:16:12Z INFO  debug_net_client] no response after 3.294712543s
[2022-09-02T20:16:12Z INFO  debug_net_client] sending packet to trigger high priority busy loop
... repeat ...
```

After somewhere between 1 and 15 minutes (usually around 5 for me), you should see fewer than 4 responses and an undending sequence of "no response" logs; e.g.,

```
[2022-09-02T19:56:55Z INFO  debug_net_client] sending packet to trigger high priority busy loop
[2022-09-02T19:56:55Z INFO  debug_net_client] sending followup packet 1
[2022-09-02T19:56:55Z INFO  debug_net_client] sending followup packet 2
[2022-09-02T19:56:56Z INFO  debug_net_client] sending followup packet 3
[2022-09-02T19:56:56Z INFO  debug_net_client] sending followup packet 4
[2022-09-02T19:56:56Z INFO  debug_net_client] sending followup packet 5
[2022-09-02T19:56:57Z INFO  debug_net_client] sending followup packet 6
[2022-09-02T19:56:57Z INFO  debug_net_client] sending followup packet 7
[2022-09-02T19:56:57Z INFO  debug_net_client] sending followup packet 8
[2022-09-02T19:56:57Z INFO  debug_net_client] sending followup packet 9
[2022-09-02T19:56:59Z INFO  debug_net_client] no response after 1.005248163s
[2022-09-02T19:57:00Z INFO  debug_net_client] no response after 2.006617007s
[2022-09-02T19:57:01Z INFO  debug_net_client] no response after 3.00801117s
[2022-09-02T19:57:02Z INFO  debug_net_client] no response after 4.009221014s
[2022-09-02T19:57:03Z INFO  debug_net_client] no response after 5.010172729s
[2022-09-02T19:57:04Z INFO  debug_net_client] no response after 6.011480753s
[2022-09-02T19:57:05Z INFO  debug_net_client] no response after 7.011830731s
[2022-09-02T19:57:06Z INFO  debug_net_client] no response after 8.013229815s
[2022-09-02T19:57:07Z INFO  debug_net_client] no response after 9.014601338s
[2022-09-02T19:57:08Z INFO  debug_net_client] no response after 10.015946541s
[2022-09-02T19:57:09Z INFO  debug_net_client] no response after 11.017274414s
[2022-09-02T19:57:10Z INFO  debug_net_client] no response after 12.01828268s
```

At this point, `net` is almost certainly wedged; you can confirm by trying to talk to the normal `udpecho` task on port 7.

Even after poking at this for the better part of a day and a half, I know frustratingly little. In both the gimlet dump and on my gimletlet, the four `RDES3` values of `RX_DESC` are all `0xc1000000`, which is [the value we set it to to give ownership back to the hardware](https://github.com/oxidecomputer/hubris/blob/24042d0c134f69460ba02def77822c8af89ec2be/drv/stm32h7-eth/src/ring.rs#L472-L473). This leads me to believe we think we've told the hardware to give us packets, but it doesn't realize it.

Sometimes when I add ringbuf traces (depending on where they are) I'm no longer able to reproduce the wedge. One set of ringbuf traces I collected was aimed at figuring out if we were somehow failing to call `rx_notify()` after clearing space in the RX buffer, but based on the traces it looks like we did indeed call `rx_notify()`, but still stopped receiving packets:

```
humility: ring buffer drv_stm32h7_eth::__RINGBUF in net:
  13  616       20        1 VlanIsNextFree { now: 0x2a12b, vid: 0x301, storage_slot: 0x2, rdes3: 0xc1000000 }
  14  470       20        1 VlanCanRecv { now: 0x2a12b, vid: 0x301, rx_notify: false }
  15  616       20        1 VlanIsNextFree { now: 0x2a514, vid: 0x302, storage_slot: 0x2, rdes3: 0xc1000000 }
  16  470       20        1 VlanCanRecv { now: 0x2a514, vid: 0x302, rx_notify: false }
  17  271       20        1 EthInterrupt { now: 0x2b1b7, dmacsr: 0x80c4 }
  18  616       20        1 VlanIsNextFree { now: 0x2b1b7, vid: 0x301, storage_slot: 0x2, rdes3: 0x3e040086 }
  19  470       20        1 VlanCanRecv { now: 0x2b1b7, vid: 0x301, rx_notify: false }
  20  449       20        1 VlanRecvRxNotify { now: 0x2b1b7, vid: 0x301 }
  21  616       20        1 VlanIsNextFree { now: 0x2b1b7, vid: 0x301, storage_slot: 0x3, rdes3: 0xc1000000 }
  22  470       20        1 VlanCanRecv { now: 0x2b1b7, vid: 0x301, rx_notify: false }
  23  616       20        1 VlanIsNextFree { now: 0x2b5a0, vid: 0x302, storage_slot: 0x3, rdes3: 0xc1000000 }
  24  470       20        1 VlanCanRecv { now: 0x2b5a0, vid: 0x302, rx_notify: false }
  25  271       20        1 EthInterrupt { now: 0x2b989, dmacsr: 0x80c4 }
  26  616       20        1 VlanIsNextFree { now: 0x2b989, vid: 0x301, storage_slot: 0x3, rdes3: 0x3e040086 }
  27  470       20        1 VlanCanRecv { now: 0x2b989, vid: 0x301, rx_notify: false }
  28  449       20        1 VlanRecvRxNotify { now: 0x2b989, vid: 0x301 }                                     // <--- rx_notify call after last packet
  29  616       20        1 VlanIsNextFree { now: 0x2b989, vid: 0x301, storage_slot: 0x0, rdes3: 0xc1000000 } // <--- never received another packet after this point
  30  470       20        1 VlanCanRecv { now: 0x2b989, vid: 0x301, rx_notify: false }
  31  616       20        1 VlanIsNextFree { now: 0x2bd72, vid: 0x302, storage_slot: 0x0, rdes3: 0xc1000000 }
   0  470       21        1 VlanCanRecv { now: 0x2bd72, vid: 0x302, rx_notify: false }
   1  616       21        1 VlanIsNextFree { now: 0x2c15b, vid: 0x301, storage_slot: 0x0, rdes3: 0xc1000000 }
   2  470       21        1 VlanCanRecv { now: 0x2c15b, vid: 0x301, rx_notify: false }
   3  616       21        1 VlanIsNextFree { now: 0x2c544, vid: 0x302, storage_slot: 0x0, rdes3: 0xc1000000 }
   4  470       21        1 VlanCanRecv { now: 0x2c544, vid: 0x302, rx_notify: false }
   5  616       21        1 VlanIsNextFree { now: 0x2d4e3, vid: 0x301, storage_slot: 0x0, rdes3: 0xc1000000 }
   6  470       21        1 VlanCanRecv { now: 0x2d4e3, vid: 0x301, rx_notify: false }
```

(I can send a full dump of these traces, but am not sure how useful the earlier portions are, so I've omitted them for now.) I have not been able to get crystal clear traces on our ethernet interrupts, but from what I have gathered, nothing seems amiss there. I don't want to discourage any particular line of thought since I still have no idea what the root cause is, but if there's something wrong with the interrupt handling it's extremely subtle.

At this point I suspect some kind of memory synchronization issue between hubris and the ethernet device, possibly related to https://github.com/oxidecomputer/hubris/issues/513? But I tried a few different things aimed at narrowing that down, and was still able to reliably wedge `net`. The things I've tried that still leave `net` reliably wedgeable are:

1. Changing `ServerImpl::poll()` to repeat if an interface encounters a packet intended for another valid vid
2. Changing all the memory orderings in the ethernet driver to `Ordering::SeqCst`
3. Replacing the atomic fences in `{rx,tx}_notify()` with calls to `cortex_m::asm::dsb()` (a la https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/ethernet/eth.rs#L315-L317)
4. Putting the ethernet descriptors into a region of sram marked as [device memory](https://github.com/oxidecomputer/hubris/blob/master/sys/kern/src/arch/arm_m.rs#L361-L363) (a la https://github.com/STMicroelectronics/STM32CubeH7/blob/c828320460548114b485ec8a6affaa950f51d12b/Projects/STM32H743I-EVAL/Applications/LwIP/LwIP_HTTP_Server_Netconn_RTOS/Src/main.c#L298-L326)

The things I've tried that appear to make the bug go away, or at least make it hard enough to reproduce that my current technique isn't good enough:

1. ~~Disabling the `vlan` feature~~ Update: I left this running for much longer, and I eventually reproduced the wedge on my gimletlet with `vlan` disabled. `debug-net-client` ran for 1 hr 8 minutes before triggering the wedge.
2. Adding ringbuf traces in `rx_notify()` that record the four RDES3 words before the fence (NOTE: I have not let this run for an extended period of time; until seeing the vlan wedge after an hour, the longest I had had to wait was 15 minutes, and this ran for about 20 without wedging.)

dancrossnyc:
At its root, #750 is about a race condition between the DMA engine and the STM32 Ethernet, and the host CPU.  We can prevent this race by properly maintaining the Rx tail pointer.

Some background information to contextual this change:

The device and host participate in a simple protocol to drive a state machine: the host gives descriptors to the device that specify buffers to DMA incoming Ethernet frames into. The host gives a descriptor to the device by setting the ownership bit and setting the tail pointer: this driver handles the former properly, but does not handle the latter correctly.

The device consumes these descriptors as frames arrive and it DMAs them host memory.  In turn, the device lets the host know that it is done with a descriptor by doing two things:

1. Resetting the ownership bit in the descriptor, and
2. Sending an interrupt to the host.

Internally, the device _also_ increments its "head" pointer in the Rx ring (modulo the size of the ring).  If the device hits the end of the ring (as specified by the value in the Rx tail pointer register), it goes into a "Suspended" state and stops DMAing packets to the host; it nominally resumes when the Rx tail pointer register is written again (if the newly written tail register value is the same as the old value, it doesn't appear to continue: note this implies it is not possible to transition from a _full_ ring directly to an _empty_ ring).

Both device and host maintain (or, perhaps more precisely, should maintain) two pointers into the ring: a producer pointer, and a consumer pointer.

One may think of this as the host producing available descriptors that are consumed by the device, and the device producing received frames that are consumed by the host. The tail pointer sets a demarkation point between the host and device: descriptors before the tail pointer are owned by the device, those after are owned by the host.  The ownership bit is used to communicate to the host that the device is done with a given descriptor.

Put another way, the tail pointer represents the host's producer pointer; the host's consumer pointer is simply the next descriptor in the ring after the last one that it processed. The device's consumer pointer is whatever the host set the tail pointer to, and it's producer pointer is its head pointer.  If the device encounters a situation where _its_ head and tail pointer are equal, the ring is full, and the DMA engine enters the suspended state until the tail pointer is rewritten.

An invariant that both sides of the protocol must maintain is that host software ensures that it's producer pointer (that is, what it sets the tail pointer to) is always equal to or (ie, the ring is full), or greater than (ie, there is space available in the ring) to the devices's producer pointer (that is, the devices's "head" pointer).

Of course, we talk about an ordering relationship here, but all of this is happening modulo the size of the ring, but that's a detail.

Note that there are two cases here:

1. The first time you write to the tail pointer to set things in motion at the beginning of time, and
2. Every other time.

For the first time, this change writes the address of the last valid descriptor in the ring into the tail pointer register (not one beyond the last!).  Then after consuming and resetting a descriptor on the host, we write the address of the next valid descriptor in the ring to the tail pointer, indicating that the just-processed descriptor is available to the device.  As we do this for every received frame, we can never starve the ring (at least, not due to this).

In this scenario, the device will follows the host around the ring; the problem with the first write is solved since, after the first frame reception, the device's head pointer will be somewhere beyond the first descriptor in the ring.  Thus, writing the tail pointer to the address of (say) the 0'th element is fine, since that won't be equal to the head pointer on the peripheral.  Similarly, all available descriptors can be consumed by the device.

Finally, this allows for some simplification with respect to barriers: since the tail pointer indicates which descriptors are owned by the device and which by the host, the host can batch up updates to the descriptors and do a single flush at the end of its processing, write before writing to the tail pointer.  This was already done in `rx_notify`.

Note that something similar should probably be done for the transmission ring, which I suspect is vulnerable to a similar race between the DMA engine on the device and writes on the host.

I have validated that, whereas I could reproduce this issue in a demo branch prior to this change, I cannot reproduce it with this change.

### DIFF
--- a/drv/stm32h7-eth/src/lib.rs
+++ b/drv/stm32h7-eth/src/lib.rs
@@ -13,6 +13,7 @@
 #![no_std]
 
 use core::convert::TryFrom;
+use core::sync::atomic::{self, Ordering};
 
 #[cfg(feature = "h743")]
 use stm32h7::stm32h743 as device;
@@ -157,8 +158,10 @@ impl Ethernet {
         // above.
         dma.dmactx_dtpr
             .write(|w| unsafe { w.tdt().bits(tx_ring.tail_ptr() as u32 >> 2) });
-        dma.dmacrx_dtpr
-            .write(|w| unsafe { w.rdt().bits(rx_ring.tail_ptr() as u32 >> 2) });
+        atomic::fence(Ordering::Release);
+        dma.dmacrx_dtpr.write(|w| unsafe {
+            w.rdt().bits(rx_ring.first_tail_ptr() as u32 >> 2)
+        });
 
         // Central DMA config:
 
@@ -318,11 +321,11 @@ impl Ethernet {
     fn rx_notify(&self) {
         // We have dequeued a packet! The hardware might not realize there is
         // room in the RX queue now. Poke it.
-        core::sync::atomic::fence(core::sync::atomic::Ordering::Release);
+        atomic::fence(Ordering::Release);
         // Poke the tail pointer so the hardware knows to recheck (dropping two
-        // bottom bits because svd2rust)
+        // bottom bits because svd2rust).
         self.dma.dmacrx_dtpr.write(|w| unsafe {
-            w.rdt().bits(self.rx_ring.tail_ptr() as u32 >> 2)
+            w.rdt().bits(self.rx_ring.next_tail_ptr() as u32 >> 2)
         });
     }
 
--- a/drv/stm32h7-eth/src/ring.rs
+++ b/drv/stm32h7-eth/src/ring.rs
@@ -448,12 +448,18 @@ impl RxRing {
         self.storage.as_ptr()
     }
 
-    /// Returns a pointer to the byte just past the end of the `RxDesc` ring.
-    /// This too gets loaded into the DMA controller, so that it knows what
-    /// section of the ring is initialized and can be read. (The answer is "all
-    /// of it.")
-    pub fn tail_ptr(&self) -> *const RxDesc {
-        self.storage.as_ptr_range().end
+    /// Returns a pointer to the last valid descriptor in the ring.
+    /// We load this into the DMA controller when we first start operation so
+    /// that it knows what section of the ring is initialized and can be read.
+    /// The answer is "all of it," but we can't exactly specify that.
+    pub fn first_tail_ptr(&self) -> *const RxDesc {
+        self.base_ptr().wrapping_add(self.len() - 1)
+    }
+
+    /// Returns a pointer to the "next" descriptor in the ring.  We load this
+    /// into the device so that the DMA engine knows what descriptors are free.
+    pub fn next_tail_ptr(&self) -> *const RxDesc {
+        self.base_ptr().wrapping_add(self.next.get())
     }
 
     /// Returns the count of entries in the descriptor ring / buffers in the
@@ -468,12 +474,10 @@ impl RxRing {
     fn set_descriptor(d: &RxDesc, buffer: *mut [u8; BUFSZ]) {
         d.rdes[0].store(buffer as u32, Ordering::Relaxed);
         d.rdes[1].store(0, Ordering::Relaxed);
-        d.rdes[2].store(0, Ordering::Release);
-        // See hubris#750 for why we need Ordering::Release and this delay
-        cortex_m::asm::delay(16);
+        d.rdes[2].store(0, Ordering::Relaxed);
         let rdes3 =
             1 << RDES3_OWN_BIT | 1 << RDES3_IOC_BIT | 1 << RDES3_BUF1_VALID_BIT;
-        d.rdes[3].store(rdes3, Ordering::Release); // <-- release
+        d.rdes[3].store(rdes3, Ordering::Relaxed);
     }
 }
 


lzrd (approved):
This is a very good fix. The previous mental model was that the OWN bit was the key concept driving IO progress and that the tail pointer was just used to tell the HW to look at descriptors again.
In fact, without proper use of the tail pointer, it is probable that the HW is constantly polling the descriptor for a change in the OWN bit, thereby wasting power and memory bandwidth. Dan observes that the HW was never entering the SUSPEND state.

mkeeter:
tl;dr: I disagree about a few specific observations (and have seen contrary behavior on hardware) but agree with the fix, with one remaining concern.

> In fact, without proper use of the tail pointer, it is probable that the HW is constantly polling the descriptor for a change in the OWN bit, thereby wasting power and memory bandwidth. Dan observes that the HW was never entering the SUSPEND state.

I have observed differently on actual hardware, which appears to enter the suspended state when it reads a descriptor without the OWN bit set.  This is also shown in Fig 784:

Screenshot 2023-08-17 at 9 29 39 AM

In our existing code (where the tail pointer is always left at the one-past-the-end position), failing to service the DMA ring will eventually result in the hardware entering the suspended state (i.e. `DMADSR.RPS` is `0b100`).

You can test this by enabling [the commented-out code here](https://github.com/oxidecomputer/hubris/blob/dma-timing-attack/drv/stm32h7-eth/src/ring.rs#L464-L479) (which logs `DMADSR.RPS` at the moment we release the descriptor), loading the Nucleo H753 demo app, and running the timing attack (` sudo ./target/debug/timing-attack -ien10 --mac "0e:1d:9a:64:b8:c2" spam --delay 500`).  After a few seconds, you can read back `RX_RPS` using Humility:
```
$ humility readvar RX_RPS|grep value
            value: 0x4
            value: 0x0
            value: 0x0
            value: 0x5b9
            value: 0x19c
            value: 0x0
            value: 0x0
            value: 0x3
```
Values here are ordered based on `RPS0`:
Screenshot 2023-08-17 at 10 09 26 AM
so there's definitely non-zero cases of the peripheral in its suspended state.

> it goes into a "Suspended" state and stops DMAing packets to the host; it nominally resumes when the Rx tail pointer register is written again (if the newly written tail register value is the same as the old value, it doesn't appear to continue: note this implies it is not possible to transition from a full ring directly to an empty ring).

I've seen evidence to the contrary on actual hardware.

If I set the tail pointer to `DES[3]`, then the hardware peripheral enters the suspended state when it hits `DES[3]` (which is expected).  However, if I then **reassign** the pointer to the same value, the peripheral resumes operation.  This suggests that the "check the tail pointer and suspend if matching" only occurs on peripheral pointer increment, and not on a receive poll demand.

Here's a trivial diff to show this behavior:

```diff
diff --git a/drv/stm32h7-eth/src/ring.rs b/drv/stm32h7-eth/src/ring.rs
index 733e6a5f..23fed467 100644
--- a/drv/stm32h7-eth/src/ring.rs
+++ b/drv/stm32h7-eth/src/ring.rs
@@ -453,7 +453,7 @@ impl RxRing {
     /// section of the ring is initialized and can be read. (The answer is "all
     /// of it.")
     pub fn tail_ptr(&self) -> *const RxDesc {
-        self.storage.as_ptr_range().end
+        self.storage.as_ptr_range().end.wrapping_sub(1)
     }
 
     /// Returns the count of entries in the descriptor ring / buffers in the
```

In the mental model where rewriting the pointer to the same value _does not_ trigger a receive poll demand, we'd expect this to accept 3 frames, then stop forever (because the peripheral pointer is at `&DES[3]`, which is the tail pointer).

--------------------------------------------------

That being said, these changes line up with my current mental model and I think they should work!

There remains the potential for one subtle race condition, which may not be possible in practice.  Consider the following situation:
```
DESC | 0   | 1   | 2   | 3   |
OWN  | dma | dma | dma | dma |
     ^ dma (waiting for packet)
     ^ user (waiting for DMA IRQ)
     ^ tail pointer
```
The typical behavior upon receiving a packet would be something like

- DMA peripheral finishes writing to DESC[0], clears the OWN bit, triggers an IRQ
- DMA peripheral increments its head pointer to DESC[1]
    - The tail pointer is at DESC[0], so this is fine
- DMA peripheral enters "waiting for packet" state using DESC[1]
- The user handles DESC[0], releases it, sets the tail pointer to DESC[1]
    - Because the DMA peripheral is already in the "waiting for packet state", this is fine and does not cause the peripheral to be suspended

The possible race condition is the following ordering of operations:

- DMA peripheral finishes writing to DESC[0], clears the OWN bit, triggers an IRQ
- The user handles DESC[0], releases it, sets the tail pointer to DESC[1]
    - Because the DMA peripheral is already in the "waiting for packet state", this is fine
- DMA peripheral increments its head pointer to DESC[1]
    - The tail pointer is at DESC[1]
    - ⚠️ The DMA peripheral enters the suspended state, and will never escape

I'm not sure if this race condition is feasible in practice: it requires user code to handle the IRQ _incredibly quickly_.  It certainly won't happen in our user code, which copies packet data out of the DMA buffers.

dancrossnyc:
> tl;dr: I disagree about a few specific observations (and have seen contrary behavior on hardware) but agree with the fix, with one remaining concern.
> 
> > In fact, without proper use of the tail pointer, it is probable that the HW is constantly polling the descriptor for a change in the OWN bit, thereby wasting power and memory bandwidth. Dan observes that the HW was never entering the SUSPEND state.
> 
> I have observed differently on actual hardware, which appears to enter the suspended state when it reads a descriptor without the OWN bit set. This is also shown in Fig 784:

This conflates two things. I see what you're saying about polling on the descriptor, and the diagram seems clear enough; I confess, once it became clear that the current operation was inherently racy, I didn't worry about that side of it too much anymore and instead concentrated on proper maintenance of the tail pointers.

But note that the failure case _is_ due to the DMA engine not entering the suspended state; that's because application code on the host is resetting the descriptor and sets the OWN bit _before_ it's (re)observed by the peripheral, but _after_ the peripheral has read the other bits of the descriptor structure.

If one were to set the initial value of the tail pointer to just after the end of the ring, the code would still be susceptible to the race condition until the first packet was fully processed and the tail pointer updated.

> [snip]
> > it goes into a "Suspended" state and stops DMAing packets to the host; it nominally resumes when the Rx tail pointer register is written again (if the newly written tail register value is the same as the old value, it doesn't appear to continue: note this implies it is not possible to transition from a full ring directly to an empty ring).
> 
> I've seen evidence to the contrary on actual hardware.
> 
> If I set the tail pointer to `DES[3]`, then the hardware peripheral enters the suspended state when it hits `DES[3]` (which is expected). However, if I then **reassign** the pointer to the same value, the peripheral resumes operation. This suggests that the "check the tail pointer and suspend if matching" only occurs on peripheral pointer increment, and not on a receive poll demand.

That is a reasonable interpretation, but in the list of states you linked earlier, there's a "stopped" state that parenthetically mentions "Reset or Stop Receive Command issued", and it may also be that going from the stopped state to running takes a different path than going from suspended to running, handling this case differently.

This is something I think it would be useful to actually get some clarity from ST on.

> Here's a trivial diff to show this behavior:
> 
> ```diff
> diff --git a/drv/stm32h7-eth/src/ring.rs b/drv/stm32h7-eth/src/ring.rs
> index 733e6a5f..23fed467 100644
> --- a/drv/stm32h7-eth/src/ring.rs
> +++ b/drv/stm32h7-eth/src/ring.rs
> @@ -453,7 +453,7 @@ impl RxRing {
>      /// section of the ring is initialized and can be read. (The answer is "all
>      /// of it.")
>      pub fn tail_ptr(&self) -> *const RxDesc {
> -        self.storage.as_ptr_range().end
> +        self.storage.as_ptr_range().end.wrapping_sub(1)
>      }
>  
>      /// Returns the count of entries in the descriptor ring / buffers in the
> ```
> 
> In the mental model where rewriting the pointer to the same value _does not_ trigger a receive poll demand, we'd expect this to accept 3 frames, then stop forever (because the peripheral pointer is at `&DES[3]`, which is the tail pointer).
> 
> That being said, these changes line up with my current mental model and I think they should work!
> 
> There remains the potential for one subtle race condition, which may not be possible in practice. Consider the following situation:
> 
> ```
> DESC | 0   | 1   | 2   | 3   |
> OWN  | dma | dma | dma | dma |
>      ^ dma (waiting for packet)
>      ^ user (waiting for DMA IRQ)
>      ^ tail pointer
> ```
> 
> The typical behavior upon receiving a packet would be something like
> 
> * DMA peripheral finishes writing to DESC[0], clears the OWN bit, triggers an IRQ
> * DMA peripheral increments its head pointer to DESC[1]
>   
>   * The tail pointer is at DESC[0], so this is fine
> * DMA peripheral enters "waiting for packet" state using DESC[1]
> * The user handles DESC[0], releases it, sets the tail pointer to DESC[1]
>   
>   * Because the DMA peripheral is already in the "waiting for packet state", this is fine and does not cause the peripheral to be suspended
> 
> The possible race condition is the following ordering of operations:
> 
> * DMA peripheral finishes writing to DESC[0], clears the OWN bit, triggers an IRQ
> * The user handles DESC[0], releases it, sets the tail pointer to DESC[1]
>   
>   * Because the DMA peripheral is already in the "waiting for packet state", this is fine
> * DMA peripheral increments its head pointer to DESC[1]
>   
>   * The tail pointer is at DESC[1]
>   * ⚠️ The DMA peripheral enters the suspended state, and will never escape
> 
> I'm not sure if this race condition is feasible in practice: it requires user code to handle the IRQ _incredibly quickly_. It certainly won't happen in our user code, which copies packet data out of the DMA buffers.

Given that there may well be a difference when transitioning from stopped to running states and from suspended to running, and that the theory behind this race is predicated on their similarity, I'd urge caution here in that we haven't validated the base assumptions (that I'm aware of, anyway). If this were, indeed, possible that seems like a major oversight in ring handling in ST's hardware. While not impossible, looking again at their suite of states, given how they differentiate several different running states with respect to handling of the current (and next) descriptor, I would be very surprised if they did not test whether the tail pointer register had been written before suspending.

At any rate, I do think we should follow up with ST, amending the earlier report with revised understanding and questions for them.

dancrossnyc:
I've updated the change description to hopefully reflect the present understanding a little bit better.

drv/stm32h7-eth/src/lib.rs
```diff
@@ -157,8 +157,9 @@ impl Ethernet {
         // above.
         dma.dmactx_dtpr
             .write(|w| unsafe { w.tdt().bits(tx_ring.tail_ptr() as u32 >> 2) });
-        dma.dmacrx_dtpr
-            .write(|w| unsafe { w.rdt().bits(rx_ring.tail_ptr() as u32 >> 2) });
+        dma.dmacrx_dtpr.write(|w| unsafe {
+            w.rdt().bits(rx_ring.first_tail_ptr() as u32 >> 2)
+        });
```
mkeeter, line 162: Now that we've removed the barrier from `set_descriptor`, we should be paranoid and add an explicit barrier before this write.

mkeeter:
> But note that the failure case is due to the DMA engine not entering the suspended state; that's because application code on the host is resetting the descriptor and sets the OWN bit before it's (re)observed by the peripheral, but after the peripheral has read the other bits of the descriptor structure.
>
> If one were to set the initial value of the tail pointer to just after the end of the ring in this change, the code would still be susceptible to the race condition until the first packet was fully processed and the tail pointer updated.

Agreed on both points!

> That is a reasonable interpretation, but in the list of states you linked earlier, there's a "stopped" state that parenthetically mentions "Reset or Stop Receive Command issued", and it may also be that going from the stopped state to running takes a different path than going from suspended to running, handling this case differently.

Also agreed!  To spell stuff out explicitly, it seems like you can't get from stopped → running if the peripheral's head and tail pointers are the same; but you _can_ get from suspended → running (as discussed above).

> Given that there may well be a difference when transitioning from stopped to running states and from suspended to running, and that the theory behind this race is predicated on their similarity, I'd urge caution here in that we haven't validated the base assumptions (that I'm aware of, anyway).

I don't understand this comment.  This race condition that I'm positing doesn't depend on transitions out of stopped / suspended states; the peripheral is running when the race window occurs.

Perhaps you're questioning the assertion that the DMA peripheral will "never escape" once suspended?  That's because user code will never receive another packet (all packets have been handled, we've poked the tail pointer, and the peripheral is suspended), so user code will never again poke the tail pointer to kick the peripheral back into action.

dancrossnyc:
> [snip]
> > That is a reasonable interpretation, but in the list of states you linked earlier, there's a "stopped" state that parenthetically mentions "Reset or Stop Receive Command issued", and it may also be that going from the stopped state to running takes a different path than going from suspended to running, handling this case differently.
> 
> Also agreed! To spell stuff out explicitly, it seems like you can't get from stopped → running if the peripheral's head and tail pointers are the same; but you _can_ get from suspended → running (as discussed above).

Yes, that fits the observed behavior, though it may be worth noting that this seems to contradict the datasheet in section 58.10.2.

> > Given that there may well be a difference when transitioning from stopped to running states and from suspended to running, and that the theory behind this race is predicated on their similarity, I'd urge caution here in that we haven't validated the base assumptions (that I'm aware of, anyway).
> 
> I don't understand this comment. This race condition that I'm positing doesn't depend on transitions out of stopped / suspended states; the peripheral is running when the race window occurs.
> 
> Perhaps you're questioning the assertion that the DMA peripheral will "never escape" once suspended? That's because user code will never receive another packet (all packets have been handled, we've poked the tail pointer, and the peripheral is suspended), so user code will never again poke the tail pointer to kick the peripheral back into action.

No. Let me see if I can explain a bit better. Earlier you wrote:

> This suggests that the "check the tail pointer and suspend if matching" only occurs on peripheral pointer increment,

This is based on the observation that writing tail, where tail==head, to the tail pointer register kicks it out of suspended state and into running. I think what this means is that the decision to transition from running to suspended state happens only when we're actively processing the descriptor ring and transitioning from one descriptor to another (that is, when we're running), but by focusing on the pointer increment here we are making an assumption: that the decision to suspend ignores concurrent writes to the tail pointer register. This assumption seems to underlie this hypothetical race.

However, I see no evidence that the decision to transition to the suspended state relies solely on the equality test after increment; it may well be that the device latches indicators of writes to the tail pointer register, resets that before the increment and test, and checks the latch again after, thus detecting a write to the register that would negate the Suspension criteria (interrupt controllers have to do things like this to all the time to avoid missing interrupts, so I don't think it would be too unusual, and it is a rather obvious race condition). Alternatively, they may make the decision to suspend before generating an interrupt. Of course, it may also be that it doesn't do any of these things, but we just don't know. Absent evidence that there's actually a race here, I'm leery of assuming that there is one.

In any event, it's certainly worth asking ST.

mkeeter:
> Yes, that fits the observed behavior, though it may be worth noting that this seems to contradict the datasheet in section 58.10.2.

Yup!  For those following along at home, here's the relevant section:

Screenshot 2023-08-17 at 3 23 13 PM

(this is already trivially contradicted by ST's example code, which always writes the tail pointer to 0; the `Current Descriptor Pointer < Descriptor Tail Pointer` could never be true in that case)

> However, I see no evidence that the decision to transition to the suspended state relies solely on the equality test after increment; it may well be that the device latches indicators of writes to the tail pointer register, resets that before the increment and test, and checks the latch again after, thus detecting a write to the register that would negate the Suspension criteria (interrupt controllers have to do things like this to all the time to avoid missing interrupts, so I don't think it would be too unusual, and it is a rather obvious race condition). Alternatively, they may make the decision to suspend before generating an interrupt. Of course, it may also be that it doesn't do any of these things, but we just don't know. Absent evidence that there's actually a race here, I'm leery of assuming that there is one.

I understand your point now, thanks for the explanation!

mkeeter (approved):

PR #1495, opened 2023-08-15
stm32h7-eth: Properly maintain Tx tail pointer

dancrossnyc:
Analagous to #1493, we should maintain the Tx tail pointer properly, as well.  Note that since the device won't examine a Tx descriptor until the tail pointer is set appropriately, we can relax stores and rely on the barrier in `tx_notify` to flush everything to RAM properly before writing the tail pointer and kicking off a DMA.

### DIFF
--- a/drv/stm32h7-eth/src/lib.rs
+++ b/drv/stm32h7-eth/src/lib.rs
@@ -150,14 +150,16 @@ impl Ethernet {
         dma.dmacrx_rlr
             .write(|w| unsafe { w.rdrl().bits(rx_ring_len) });
 
-        // Poke both tail pointers so the hardware looks at the descriptors. We
-        // completely initialize the descriptor array, so the tail pointer is
-        // always the end.
+        // Poke the receive tail pointer so that the hardware looks at
+        // descriptors. We completely initialize the descriptor array, so the
+        // tail pointer is always as close to the end as we can make it.
         //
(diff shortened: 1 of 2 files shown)


lzrd (approved):
This is a very good fix. The previous mental model was that the OWN bit was the key concept driving IO progress and that the tail pointer was just used to tell the HW to look at descriptors again.
In fact, without proper use of the tail pointer, it is probable that the HW is constantly polling the descriptor for a change in the OWN bit, thereby wasting power and memory bandwidth. Dan observes that the HW was never entering the SUSPEND state.

mkeeter (approved):
The code looks good and I tested it for a while locally.

There are a bunch of comments in `ring.rs` that are now outdated, talking about how we're careful to handle memory ordering with a RELEASE store.  I'm going to pre-approve this now, but can you go through and update those comments before merging?  

(apologies for not catching this during the Rx PR)

dancrossnyc:
> The code looks good and I tested it for a while locally.
> 
> There are a bunch of comments in `ring.rs` that are now outdated, talking about how we're careful to handle memory ordering with a RELEASE store. I'm going to pre-approve this now, but can you go through and update those comments before merging?

I believe I've gotten them all; mind taking another quick look?

> (apologies for not catching this during the Rx PR)

No worries!

mkeeter:
Looks good, thanks!

Issue #1335, opened 2023-05-03
RoT: detect that a debug credential has been supplied and unlock debugging interface


cbiffle:
We're evaluating using NXP's debug authentication for allowing in-field access to the RoT by a user who can supply a suitably signed credential. The expectation is that this credential's "beacon" field is non-zero, which causes the ROM to check it but not apply it to the debug settings, leaving it to the firmware to do that.

On the RoT, we'd want to do that at a very specific point: after updating the CDI and locking the PUF, but before starting the kernel.

This issue is currently a placeholder for the work required there, which has not been scoped. (Edit this description when that happens.)

@kc8apf

cbiffle:
I've bumped this to MVP because the debug unlock doesn't seem relevant for our initial shipment, since we won't be able to get physical access to it. It'd come into play if we had a device that was effectively bricked and we were trying to remanufacture it, but other than that. Open to being convinced to move it back.

labbott:
I took a brief look at the beacon code in the RoT. For a more specific description, when the beacon is `0` , the the TAP interface, `SYSCON_SWD_CPU0`, `SYSCON_SWD_CPU1`, `SYSCON_DEBUG_FEATURES`, `SYSCON_DEBUG_FEATURES_DB` are all set based on the settings in CFPA/CMPA/DebugAuth and then the registers are locked via `SYSCON_DEBUG_LOCK_EN`. When the beacon is non-zero, the debug settings are set in `SYSCON_DEBUG_FEATURES` and _not_ locked. The debug beacon does not appear to be read anywhere else in the ROM or set outside of debug authentication (no anti-features to catch us by surprise if we use it)

Adding the code to do the unlock is fairly simple. The tricky parts are making sure all secrets are cleared and making sure we are only enabling the debug feature we need.

labbott:
Closed via #1496

PR #1496, opened 2023-08-15
Enable RoT debug when the beacon is set

labbott:
The beacon register can only be set during debug authentication. If this register is set, enable debugging.

### DIFF
--- a/app/oxide-rot-1/app.toml
+++ b/app/oxide-rot-1/app.toml
@@ -9,7 +9,7 @@ version = 0
 
 [kernel]
 name = "oxide-rot-1"
-requires = {flash = 60544, ram = 2696}
+requires = {flash = 60644, ram = 2696}
 features = ["dice-mfg"]
 
 [caboose]
--- a/lib/lpc55-rot-startup/src/lib.rs
+++ b/lib/lpc55-rot-startup/src/lib.rs
@@ -75,6 +75,46 @@ fn puf_check(puf: &lpc55_pac::PUF) {
     }
 }
 
+fn enable_debug(peripherals: &lpc55_pac::Peripherals) {
+    const HUBRIS_DEBUG_CRED_BEACON: u32 = 10000;
+
+    let beacon = peripherals.SYSCON.debug_auth_beacon.read().bits();
+    // The beacon is made up of two parts: 16 bits of credential beacon
+    // which is signed by the root key and 16 bits of authentication beacon
+    // which can be changed at runtime. Use the sign credential beacon
+    // to decide if debugging should be enabled
+    let cred_beacon = beacon & 0xffff;
+
+    if cred_beacon == HUBRIS_DEBUG_CRED_BEACON {
+        // See Section 4.5.83 of UM11126, this isn't actually named
+        // in the manual apart from `CPU0 SWD-AP: 0x12345678`
+        const SWD_MAGIC: u32 = 0x12345678;
+        // This register is not defined in the LPC55 PAC
+        // This information comes from Section 4.5.83 of UM11126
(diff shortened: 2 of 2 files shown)


app/oxide-rot-1/app.toml
```diff
@@ -9,7 +9,7 @@ version = 0
 
 [kernel]
 name = "oxide-rot-1"
-requires = {flash = 60544, ram = 2696}
+requires = {flash = 60644, ram = 2696}
```
labbott, line 12: I think we went over by about 12 bytes but I bumped it by a little more to give us some room

lib/lpc55-rot-startup/src/lib.rs
```diff
@@ -75,6 +75,38 @@ fn puf_check(puf: &lpc55_pac::PUF) {
     }
 }
 
+fn enable_debug(peripherals: &lpc55_pac::Peripherals) {
+    let beacon = peripherals.SYSCON.debug_auth_beacon.read().bits();
+
+    if beacon != 0 {
```
labbott, line 81: RFD333 said "non-zero" but I don't know if we want to consider making this more specific. I wasn't sure how to combine that with the credentials for choosing A/B images.

flihp: Can you provide a link to the relevant section of the RFD? I'm not able to find the reference. Generally, based on my current understanding I'd agree though: this does need to be more specific.

AFAIK the beacon has two parts: the lower half of the register is the `credential beacon`. This is set in the debug credential at time of creation and is part of the data that is signed by one of the trust anchors. This allows the issuer to limit the powers of the credential. We use this value in the bootleby image selection here: https://github.com/oxidecomputer/bootleby/blob/main/src/bin/bootleby.rs#L154. The other half of the register (aka the auth beacon) can be set to an arbitrary value by the holder of the debug auth credential as part of the protocol exchange with the ROM. This allows the credential holder to provide some input (the auth beacon) that is limited in scope by the credential beacon.

Using bootleby as an example: The credential beacon that allows the debug auth credential holder to control bootleby boot selection is `18578`. I think of this as defining the scope of the credential: in this case it's overriding the default A / B image selection. This is to say that in order to control bootleby image selection the debug auth credential must have the `CB` field set to `18578`. During the protocol exchange with the ROM the A / B image is selected by setting the authentication beacon to either `0` to select image A, or `1` to select image B. These values are documented in OANA: https://github.com/oxidecomputer/oana#beacon

If I'm understanding all of this correctly the code here will allow the holder of any debug auth credential to debug the chip via SWD. This includes the credentials issued for overriding the bootleby image selection. This probably isn't what we want. Instead I think we need to add another credential beacon value to OANA and check that the provided debug auth credential has this credential beacon set. I don't see any use for the auth beacon here but it's a mechanism available if we need it.

flihp (changes requested):

lib/lpc55-rot-startup/src/lib.rs
flihp: Thanks for the reference. I searched for the quoted text ("non-zero") and came up empty handed. I think it's far more likely that we'll want a test / mfg system that can override boot selection to ensure that both A and B images can boot but that *cannot* do generic SWD things. If we do run into a situation where we want a credential that's able to do both things we could probably just use two separate credentials?

labbott: I chose a number and put it in the OANA and double checked that setting the cred beacon _will_ unlock debug but anything will _not_.

flihp (approved):

labbott:
#1335

PR #1502, opened 2023-08-17
modulus: Missed a few places

dancrossnyc:
I missed a few places where the hard-coded `if` sequence was written out.  Also, because it's so redundant, wrap it up into a little function.


mkeeter (approved):

mkeeter (approved):
It's getting tempting to unify `TxRing` and `RxRing` into `Ring<D>`, but this is fine for now.

PR #1504, opened 2023-08-18
Move away from ROM APIs for flash writing

labbott:
We have a driver to access the flash controller directly, use it. Remove the old APIs so nobody is tempted to use them.


flihp:
This work has been rebased on top of main now that #1500 has been merged: https://github.com/flihp/hubris/tree/no_more_rom. I've picked a few minor nits along the way that include mostly cosmetic stuff:
- adding a 'use drv_lpc55_flash::Flash` instead of spelling out the full type each time it's used in a few spots
- removing use of `lib_lpc55_romapi::FLASH_PAGE_SIZE` and replacing it with `drv_lpc55_flash::BYTES_PER_FLASH_PAGE`
- making `lib_lpc55_romapi::FLASH_PAGE_SIZE` private so others aren't tempted to use it instead of the equivalent from `drv_lpc55_flash` (it's still used by `lib_lpc55_romapi` internally)

I've also run this code through the identity manufacturing process and it works as expected.

flihp (approved):
solid

PR #1505, opened 2023-08-21
Add MGS sensor reading

mkeeter:
This PR bumps MGS and implements the new `SensorRead` message.

It also ends up making the `Sensor` API more flexible:

- We now distinguish between device errors (`NoData`) and API errors (the new `enum SensorApiError`).  The existing existing `SensorError` type remains the union of both types, but we can use `SensorApiError` for almost all `Sensor` API return values
- Add a new `Sensor.get_raw_reading` function, which maintains the distinction between device and API errors in its return type.  This 
- All `Sensor` functions are idempotent, so we can remove the `ServerDeath` variant
- Added new APIs to get the last data or error reading, including the timestamp
- Extend the `LastReading` type so that we know if we've _ever_ seen a data or error value; this allows us to correctly return `NoReading` for the new `get_last_data` / `get_last_nodata` APIs

### DIFF
--- a/task/sensor/src/main.rs
+++ b/task/sensor/src/main.rs
@@ -26,23 +26,22 @@ enum LastReading {
 }
 
 /// Zero-cost array wrapper that can be indexed with a `SensorId`
-struct SensorArray<T: 'static, const N: usize>(&'static mut [T; N]);
-impl<T: 'static, const N: usize> core::ops::Index<SensorId>
-    for SensorArray<T, N>
-{
+struct SensorArray<T: 'static>(&'static mut [T; NUM_SENSORS]);
+impl<T: 'static> core::ops::Index<SensorId> for SensorArray<T> {
     type Output = T;
+    #[inline(always)]
     fn index(&self, idx: SensorId) -> &Self::Output {
         &self.0[idx.0 as usize]
     }
 }
-impl<T: 'static, const N: usize> core::ops::IndexMut<SensorId>
-    for SensorArray<T, N>
-{
+impl<T: 'static> core::ops::IndexMut<SensorId> for SensorArray<T> {
+    #[inline(always)]
     fn index_mut(&mut self, idx: SensorId) -> &mut Self::Output {
         &mut self.0[idx.0 as usize]
     }
 }
-impl<T: 'static, const N: usize> SensorArray<T, N> {
+impl<T: 'static> SensorArray<T> {
+    #[inline(always)]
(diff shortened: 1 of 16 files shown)


labbott (approved):
LGTM

task/sensor/src/main.rs
```diff
@@ -51,25 +57,82 @@ impl idl::InOrderSensorImpl for ServerImpl {
...
         &mut self,
-        _: &RecvMessage,
+        msg: &RecvMessage,
         id: SensorId,
     ) -> Result<Reading, RequestError<SensorError>> {
+        self.get_raw_reading(msg, id)
+            .map_err(|e| e.map_runtime(SensorError::from))
+            .and_then(|(value, timestamp)| match value {
+                Ok(value) => Ok(Reading { value, timestamp }),
+                Err(e) => Err(<SensorError as From<_>>::from(e).into()),
```
jgallagher, line 67: Nit - can this be

```suggestion
                Err(e) => Err(SensorError::from(e).into()),
```

task/sensor/src/main.rs
```diff
@@ -51,25 +57,82 @@ impl idl::InOrderSensorImpl for ServerImpl {
...
+    }
+
+    fn get_raw_reading(
+        &mut self,
+        _: &RecvMessage,
+        id: SensorId,
+    ) -> Result<(Result<f32, NoData>, u64), RequestError<SensorApiError>> {
+        let index = id.0 as usize;
+
+        if index < NUM_SENSORS {
```
jgallagher, line 78: Nit - I'm sure the compiler can optimize out the bounds check in `self.last_reading[index]` given this `if`, but I think this guarantees it and is slightly clearer:

```rust
let last_reading = self
    .last_reading
    .get(index)
    .ok_or(SensorApiError::InvalidSensor)?;

match last_reading {
    // ...
}
```

(same nit in more functions below this, which makes me now realize that this method is copying their style - sorry)

jgallagher (approved):
LGTM :+1:

task/sensor/src/main.rs
mkeeter: yes, thank you (I disliked how this looked before)!

task/sensor/src/main.rs
mkeeter: I went slightly overboard in a1151980df9261cebd0e555d4bb83380c8735489, adding a new class to let us index directly by `SensorId` and refactoring `last_reading` into a standalone function (which is basically the code that you posted).

PR #1512, opened 2023-08-30
Re-work test suite

labbott:
The existing test suite relied on ITM/ETM for getting test information. We've since built up other tooling that doesn't require those features. Switch to a hiffy based test design to allow for control of tests to happen off target.


labbott:
See also https://github.com/oxidecomputer/humility/pull/416 for the humility support

test/test-api/src/lib.rs
```diff
@@ -41,10 +41,6 @@ pub enum AssistOp {
 /// Operations that are performed by the test-suite
 #[derive(FromPrimitive)]
 pub enum SuiteOp {
-    /// Get the number of test cases (`() -> usize`).
-    GetCaseCount = 1,
-    /// Get the name of a case (`usize -> [u8]`).
-    GetCaseName = 2,
     /// Run a case, replying before it starts (`usize -> ()`).
     RunCase = 3,
```
bcantrill, line 45: Is it worth just eliminating the `SuiteOp` altogether given that there is only one of them now?

test/tests-stm32g0/README.mkdn
```diff
@@ -1,14 +0,0 @@
-# STM32G0 System Tests
```
bcantrill, line 1: Love seeing this go!

bcantrill (approved):
This looks really great -- thank you for doing this!

test/test-api/src/lib.rs
labbott: I thought about that but for consistency I like keeping the enum

PR #1514, opened 2023-08-30
Add sensor IDs to IPCC

mkeeter:
This PR extends the inventory IPCC messages (#1469) with trailing `SensorId` values, so that the host can get sensor data over MGS (#1505).

To simplify and deduplicate the code, I modified the I2C code-gen in two ways.

The first change is minor: I added devices and sensors by refdes (if present), e.g. `fn at24csw080_u615(task: TaskId) -> I2cDevice` and `const TMP451_U491_TEMPERATURE_SENSOR: SensorId`.

The second change is a bit larger: I added a per-device-type `struct Sensors_{device}` which encodes sensors on a per-device basis, i.e.
```rust
#[allow(non_camel_case_types)]
pub struct Sensors_tps546b24a {
    pub temperature: SensorId,
    pub current: SensorId,
    pub voltage: SensorId,
}
#[allow(dead_code)]
pub const TPS546B24A_V3P3_SP_A2_SENSORS: Sensors_tps546b24a = Sensors_tps546b24a {
    temperature: SensorId(54),
    voltage: SensorId(56),
    current: SensorId(55),
};
#[allow(dead_code)]
pub const TPS546B24A_U522_SENSORS: Sensors_tps546b24a = Sensors_tps546b24a {
    voltage: SensorId(56),
    current: SensorId(55),
    temperature: SensorId(54),
};
```
(note that we produce `struct`'s by name and refdes, containing equivalent data)

Being able to get the device and sensors by refdes makes it simpler to load inventory data, since we already need the refdes as our canonical key; I wrote a small macro (`by_refdes`) to simplify this further.

TODO: add hot-swap controllers to the inventory list


lib/host-sp-messages/src/lib.rs
```diff
@@ -282,6 +291,9 @@ pub enum InventoryData {
         ic_device_id: [u8; 4],
         /// IC_DEVICE_REV, PMBus operation 0xAE
         ic_device_rev: [u8; 4],
+
+        voltage_sensors: [SensorId; 3],
```
rmustacc, line 295: A general question I have while looking at this, is since it's on me to encode the semantics of each sensor, how do I figure out what this slot is and map it back to the corresponding entry? In this case I assume these are all output values related to the different rails.

lib/host-sp-messages/src/lib.rs
```diff
@@ -266,6 +270,11 @@ pub enum InventoryData {
         mfr_serial: [u8; 20],
         /// MFR_FIRMWARE_DATA, PMBus operation 0xFD
         mfr_firmware_data: [u8; 20],
+
+        temp_sensor: SensorId,
```
rmustacc, line 274: As we add more types of sensors to these devices of the same class, for example, there are two temp sensors, how do we end up retaining the mapping of which sensor is which in these commands? Or perhaps on the addition of synthetic sensors. I think this probably ties into the other question I have more generally about how is this mapping constructed and then how does it get maintained.

task/host-sp-comms/src/bsp/gimlet_bcde.rs
```diff
@@ -386,31 +403,36 @@ impl ServerImpl {
...
+                    3 => by_refdes!(J197, tmp117),
+                    4 => by_refdes!(J198, tmp117),
+                    5 => by_refdes!(J199, tmp117),
                     _ => unreachable!(),
                 };
+                let dev = f(I2C.get_task_id());
+
+                // Convert the name from Jxxx (in the TOML file) -> Jxxx/U1
+                let mut name = *b"Jxxx/U1";
+                name[..connector_name.len()].copy_from_slice(connector_name);
```
jgallagher, line 419: If we accidentally had a `connector_name` with length != 4, this would silently use that length, and we'd end up with `J12x/U1` (if short) or `J12345U1` (if long, or a panic if it was very long), right? Would it be slightly safer to do something like

```rust
                // All connector names should have length 4; using
                // a literal 4 here asserts that via the bounds check.
                name[..4].copy_from_slice(connector_name);
```

task/sensor-types/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub fn counter_encoding<T>(self) -> (usize, usize) {
+        //
+        // We need to encode the number of variants in [`NoData`] here.  There
+        // is a very convenient core::mem::variant_count() that does exactly
+        // this, but it's currently unstable -- so instead we have an
+        // exhaustive match to assure that the enum can't be updated without
+        // modifying this code.
+        //
+        let nbits = (core::mem::size_of::<T>() * 8)
+            / match self {
```
jgallagher, line 92: Nit - maybe this is a place for `strum` (specifically [`EnumCount`](https://docs.rs/strum/latest/strum/trait.EnumCount.html))?

build/i2c/src/lib.rs
```diff
@@ -218,6 +221,22 @@ struct I2cSensors {
     names: Option<Vec<String>>,
 }
 
+impl I2cSensors {
+    /// Checks whether two sensor sets are compatible
+    ///
+    /// "Compatible" means that they have the same number of sensors in each
+    /// category, meaning they can be represented by the same `struct`.
+    fn is_compatible_with(&self, other: &Self) -> bool {
+        self.temperature == other.temperature
```
jgallagher, line 230: This might be too nitpicky so feel free to ignore, but: I slightly worry about a future change adding a field to `I2cSensors` and forgetting to update this method. It'd be more verbose, but we could do something like

```rust
let &Self {
    temperature,
    power,
    // etc
} = self;
let &Self {
    temperature: other_temperature,
    power: other_power,
    // etc
} = other;

temperature == other_temperature && power == other_power // etc
```

to force the destructuring to turn into a compile error if fields are added. (It would also be a convenient place to put a `// names are intentionally ignored` kind of comment, since both would end up with a `names: _` entry.)

build/i2c/src/lib.rs
```diff
@@ -1358,6 +1404,115 @@ impl ConfigGenerator {
         Ok(())
     }
 
+    fn declare_sensor_struct(
+        &mut self,
+        d: &I2cDevice,
+        struct_name: &str,
+    ) -> Result<()> {
+        // Manually unpack the field so that changes to the sensor types
+        // will require changes here as well.
```
jgallagher, line 1413: Hah! So maybe my above comment about destructuring to guard against future changes was not too nitpicky! :)

build/i2c/src/lib.rs
```diff
@@ -1358,6 +1404,115 @@ impl ConfigGenerator {
...
+        sensors: &[DeviceSensor],
+    ) -> Result<()> {
+        write!(
+            &mut self.output,
+            "        #[allow(dead_code)]
+        pub const {}_{label}_SENSORS: Sensors_{name} = ",
+            d.device.to_uppercase(),
+        )?;
+
+        let mut sensors_by_kind: HashMap<Sensor, Vec<usize>> = HashMap::new();
```
jgallagher, line 1473: Since we iterate over this below, maybe a `BTreeMap` so the order is consistent between runs? It doesn't look like that matters for correctness, but might be nicer for humans skimming the output.

jgallagher (approved):

lib/host-sp-messages/src/lib.rs
mkeeter: We talked about this in person, but quick summary for the record – this is similar to [the previous discussion](https://github.com/oxidecomputer/hubris/pull/1514#discussion_r1315995020), where right now it's manual / dependent on ordering in the TOML file.  Once we create synthetic sensors, we'll want to revisit this, but it shouldn't block integration of this PR.

task/sensor-types/src/lib.rs
mkeeter: This was copy-pasta extracted from `sensor-api`, so I don't think I'll change it in this PR.

build/i2c/src/lib.rs
mkeeter: Yup, although I only need to unpack one of the values (not both).  Done in 76338f818c88bf1c24ed42ae0c704faddc7c1bb4

task/host-sp-comms/src/bsp/gimlet_bcde.rs
mkeeter: I found a sneaky way to ensure that we actually have a fixed-size byte array here; done in fe2c235f7aefd8f8b7875edbb7c09f230948bfc7

build/i2c/src/lib.rs
mkeeter: Good idea, done in ea841d070f085bfb8a68d362659c54d386e0253a

build/i2c/src/lib.rs
mkeeter: Indeed, I am nothing if not pedantic!

task/host-sp-comms/src/bsp/gimlet_bcde.rs
jgallagher: Neat!

PR #1515, opened 2023-09-01
Add more robust checks for images

labbott:
The check for a valid image magic should come before attempting to read any of the image lengths. The image lengths should also be checked for overflows.


andrewjstone (approved):
Thanks for fixing this @labbott

flihp (approved):

flihp:
nit: might want to rebase to get rid of the empty merge commit

PR #1516, opened 2023-09-01
Add feature to mimic locking

labbott:
Locking via CMPA is permanent and can't be undone. We can at least mimic part of the behavior by setting the flash lock registers.


flihp (approved):
very clever

Issue #1451, opened 2023-06-30
RoT reports timeout if not power-cycled after update


mkeeter:
After using wicket to update an RoT (from slot A to slot B), we observe that the RoT reports a timeout when attempting to read its status.  After using ignition to power-cycle of the whole board, it properly comes up as running from slot B.

andrewjstone:
I was able to reproduce this by putting `ping` back in the rot-carrier image for some chaos monkey behavior while performing an update. I get an `Err(Sprot(Timeout))` every time from `faux-mgs`. Interestingly, looking at the ringbuf, the RoT side of `sprot` was not getting any messages!

I then decided to run `tasks` on the rot-carrier and low and behold:

```
➜  hubris git:(master) ✗ rota tasks
humility: attached to 1fc9:0143:CW4N3IFK22UUX via CMSIS-DAP
humility tasks failed: kernel has panicked on boot: "panicked at 'attempt to add with overflow', lib/lpc55-rot-startup/src/images.rs:108:13"
➜  hubris git:(master) ✗ rotb tasks
humility: attached to 1fc9:0143:CW4N3IFK22UUX via CMSIS-DAP
humility tasks failed: kernel has panicked on boot: "panicked at 'attempt to add with overflow', lib/lpc55-rot-startup/src/images.rs:108:13"
```

`rota` and `rotb` are just humility aliases pointing to the correct images.

I then ran `rota registers` and saw the kernel has indeed panicked.

```
➜  hubris git:(master) ✗ rota registers
humility: attached to 1fc9:0143:CW4N3IFK22UUX via CMSIS-DAP
   R0 = 0x00000000
   R1 = 0x2000445f <- kernel: KERNEL_EPITAPH+0x56
   R2 = 0x2000445f <- kernel: KERNEL_EPITAPH+0x56
   R3 = 0x2000445f <- kernel: KERNEL_EPITAPH+0x56
   R4 = 0x00050000
   R5 = 0x00050130
   R6 = 0x00000000
   R7 = 0x20004278 <- kernel: 0x20004000+0x278
   R8 = 0x00000000
   R9 = 0x00050000
  R10 = 0x00000000
  R11 = 0x00000000
  R12 = 0x20004458 <- kernel: KERNEL_EPITAPH+0x4f
   SP = 0x20004248 <- kernel: 0x20004000+0x248
   LR = 0x00012451 <- kernel: write_str+0x35
   PC = 0x0001276c <- kernel: die_impl+0x44
[...]

PR #1518, opened 2023-09-01
Use alternate chip reset for updates

labbott:
The standard ARM system reset will not reset certain hardware blocks, notably the flash lock registers for the protected flash regions. Use the alternate syscon reset to ensure we can write our CFPA updates.


labbott:
See #1451

drv/lpc55-syscon/src/main.rs
```diff
@@ -256,6 +267,10 @@ fn set_reset_reason() {
     };
 
     Jefe::from(JEFE.get_task_id()).set_reset_reason(reason);
+
+    // Need this to be able to use the syscon reset that works for CFPA
```
mkeeter, line 271: Nit: doing this in `set_reset_reason` is confusing.  Maybe do it in `main` instead (moving ownership of `pmc`), e.g.
```rust
fn main() {
    // ... snip
    
    let pmc = unsafe { &*device::PMC::ptr() };
    set_reset_reason(pmc);

    // Need this to be able to use the syscon reset that works for CFPA
    // update
    pmc.resetctrl.write(|w| w.swrresetenable().enable());
    
    // ... etc
```

jgallagher: I was writing a similar comment but was going to suggest putting it in `chip_reset()` :+1:

drv/lpc55-syscon/src/main.rs
```diff
@@ -240,6 +255,7 @@ fn set_reset_reason() {
 
     const POR: u32 = 1 << 4;
     const PADRESET: u32 = 1 << 5;
+
```
mkeeter, line 258: Bonus newline?

drv/lpc55-syscon/src/main.rs
```diff
@@ -256,6 +272,7 @@ fn set_reset_reason() {
     };
 
     Jefe::from(JEFE.get_task_id()).set_reset_reason(reason);
+
```
mkeeter, line 275: Same here?

mkeeter (approved):

Issue #1520, opened 2023-09-06
The firmware ID (slot measurement) is only calculated on valid images.


lzrd:
One of the goals for fwid is to help us detect "dirty" or corrupted flash slots that may have partially programmed images or extra programmed flash pages.
Currently, FWID is only calculated if the image is valid.

PR #1521, opened 2023-09-06
Always calculate the firmware measurement even if there is no valid i…

lzrd:
…mage.

Note: The measurement is not plumbed through to the management plane at this time. It will wait for other API changes.

Also, figure out the programmed pages once while performing the slot measurement and use those results when the image is being verified.

Tweak .gitignore for tags and cscope users.

### DIFF
--- a/lib/lpc55-rot-startup/src/images.rs
+++ b/lib/lpc55-rot-startup/src/images.rs
@@ -82,7 +82,7 @@ impl Image {
         let fwid = hash.finalize().try_into().unwrap_lite();
         let programmed = Range {
             start: flash.start,
-            end: end.unwrap_or(flash.start),
+            end: end.unwrap_or(flash.end),
         };
         Image {
             flash,
--- a/lib/lpc55-rot-startup/src/images.rs
+++ b/lib/lpc55-rot-startup/src/images.rs
@@ -44,9 +44,11 @@ const PAGE_SIZE: u32 = BYTES_PER_FLASH_PAGE as u32;
 pub struct Image {
     // The boundaries of the flash slot.
     flash: Range<u32>,
-    // The initial span of programmed flash pages.
+    // The contiguous span of programmed flash pages starting at offset zero.
+    // Note that any additional programmed pages after the first erased
+    // page are not interesting for image sanity checks and are not included.
     programmed: Range<u32>,
-    // Measurement over the programmed pages.
+    // Measurement over all pages including those that follow any erased page.
     fwid: [u8; 32],
 }
 
@@ -62,12 +64,10 @@ impl Image {
     // Note: if partially programmed pages is a possibility then that could be
     // a problem with respect to catching exfiltration attempts.
     pub fn new(dev: &mut Flash, flash: Range<u32>) -> Image {
-        // let mut _count = 0;
         let mut end: Option<u32> = None;
         let mut hash = Sha3_256::new();
         for start in flash.clone().step_by(BYTES_PER_FLASH_PAGE) {
             if dev.is_page_range_programmed(start, PAGE_SIZE) {
-                // _count += 1;
                 let page = unsafe {
                     core::slice::from_raw_parts(
                         start as *const u8,
@@ -87,7 +87,6 @@ impl Image {
         Image {
             flash,
             programmed,
-            // _count,
             fwid,
         }
     }
@@ -96,6 +95,8 @@ impl Image {
         return self.programmed.contains(addr);
     }
 
+    /// True if the flash slot's span of contiguous programmed pages
+    /// starting at offset zero includes the given span.
     pub fn is_span_programmed(&self, start: u32, length: u32) -> bool {
         if let Some(end) = start.checked_add(length) {
             if !self.is_programmed(&start) || !self.is_programmed(&end) {
--- a/lib/lpc55-rot-startup/src/images.rs
+++ b/lib/lpc55-rot-startup/src/images.rs
@@ -2,36 +2,24 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use abi::{ImageHeader, ImageVectors};
+use abi::ImageHeader;
 use drv_lpc55_flash::{Flash, BYTES_PER_FLASH_PAGE};
 use sha3::{Digest, Sha3_256};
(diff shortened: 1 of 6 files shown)


lzrd:
This PR does not plumb the fwid up to the control plane when the image is not valid.
There are some more API changes coming for bootloader update and I'd prefer to wait until then.

lzrd:
With the minor kernel size increase, the other app.toml files will need to be updated as well.

lzrd:
https://github.com/oxidecomputer/hubris/issues/1520

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -56,20 +42,85 @@ extern "C" {
...
+            if !self.is_programmed(&start) || !self.is_programmed(&end) {
+                false
+            } else {
+                true
+            }
+        } else {
+            false
+        }
+    }
+}
```
labbott, line 112: I'm wary of this. It assumes we've really checked the programmed range and it assumes no gaps in programming.

labbott (commented):
So I agree with calculating the FWID but I'm not a fan of caching the programmed page range. That just seems like a place to introduce errors. Do we really need to cache it?

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -56,20 +42,85 @@ extern "C" {
...
+    //   - Define the address boundaries.
+    //   - Determine the bounds of the initial programmed extent.
+    //   - Count the total number of fully programmed pages.
+    //   - Measure all programmed pages including those outside of the
+    //     initial programmed extent.
+    //
+    // Note: if partially programmed pages is a possibility then that could be
+    // a problem with respect to catching exfiltration attempts.
+    pub fn new(dev: &mut Flash, flash: Range<u32>) -> Image {
+        // let mut _count = 0;
```
flihp, line 67: nit: remove dead / commended code

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -56,20 +42,85 @@ extern "C" {
...
+                hash.update(page);
+            } else if end.is_none() {
+                end = Some(start);
+            }
+        }
+        let fwid = hash.finalize().try_into().unwrap_lite();
+        let programmed = Range {
+            start: flash.start,
+            end: end.unwrap_or(flash.start),
+        };
```
flihp, line 88: Not sure if this is intentional but given a range of 3 blocks with the middle block not programmed the programmed range will be reported as `Range { start: 0, end: 1 }`. Is the intention to ignore all pages after an unprogrammed one is encountered? It will include the 2 programmed page in the hash so the important behavior has been preserved. Representing the programmed pages as a contiguous range when it may not be could cause confusion though.

lib/lpc55-rot-startup/src/images.rs
lzrd: We checked the programmed range in Image::new() and you can't call this without having obtained an Image struct through Image::new()

lib/lpc55-rot-startup/src/images.rs
lzrd: Will do

lib/lpc55-rot-startup/src/images.rs
labbott: Having chatted about this yesterday I understand better where this is coming from and why it works.

lib/lpc55-rot-startup/src/images.rs
lzrd: This is intentional. Maybe "programmed" should be renamed for clarity though.
The intent is to know the contiguous programmed region starting from offset zero.
This range will be used to test that pointers and offsets in the NXP executable image have no erased gap pages.
The first test is that the NXP length field is in that region. Later tests are bounded by that
NXP image length value.

lzrd: I updated the comment on Image{programmed: } since I didn't come up with a more descriptive name that wasn't awkward. Does that work?

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -56,20 +42,84 @@ extern "C" {
...
+    //
+    //   - Define the address boundaries.
+    //   - Determine the bounds of the initial programmed extent.
+    //   - Count the total number of fully programmed pages.
+    //   - Measure all programmed pages including those outside of the
+    //     initial programmed extent.
+    //
+    // Note: if partially programmed pages is a possibility then that could be
+    // a problem with respect to catching exfiltration attempts.
+    pub fn new(dev: &mut Flash, flash: Range<u32>) -> Image {
```
labbott, line 66: Drop the pub

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -56,20 +42,84 @@ extern "C" {
...
+            end: end.unwrap_or(flash.end),
+        };
+        Image {
+            flash,
+            programmed,
+            fwid,
+        }
+    }
+
+    pub fn is_programmed(&self, addr: &u32) -> bool {
```
labbott, line 94: does this need to be pub:

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -56,20 +42,84 @@ extern "C" {
...
+        return self.programmed.contains(addr);
+    }
+
+    // True if the flash slot's span of contiguous programmed pages
+    // starting at offset zero includes the given span.
+    pub fn is_span_programmed(&self, start: u32, length: u32) -> bool {
```
labbott, line 100: does this need to be pub:

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -56,20 +42,84 @@ extern "C" {
...
+    // The contiguous span of programmed flash pages starting at offset zero.
+    // Note that any additional programmed pages after the first erased
+    // page are not interesting for image sanity checks and are not included.
+    programmed: Range<u32>,
+    // Measurement over all pages includeing those that follow any erased page.
+    fwid: [u8; 32],
 }
 
-pub fn image_details(img: Image, flash: &mut Flash) -> RotImageDetails {
+impl Image {
```
labbott, line 55: You've got a second `impl` block here

lzrd:
Closed by #1675

PR #1523, opened 2023-09-08
Fix #1507

### LINKED ISSUE #1507: SP-RoT link ends up in a persistent bad state
While updating the colo rack with the dogfood image (https://github.com/oxidecomputer/omicron/runs/15999510273), we updated both the SP and RoT images using mupdate / wicket.

(Normally, we skip updating the RoTs during mupdate, because they haven't changed in a while)

Augustus noticed one RoT showing a failure to communicate after the initial update; a second update brought this number to three RoTs.

Here's an example of an unhappy RoT:
image(1)

For contrast, a happy RoT:
image(2)

The ringbuf shows invalid data received by the SP when we try to request the RoT status:
```
humility: ring buffer drv_stm32h7_sprot_server::__RINGBUF in sprot:
   0  241        1        1 Sent(0xa)
   1  283        1        1 Received(0x8)
   2  437        1        1 RxBuf([ 0x1, 0x0, 0x2, 0xe4, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 ])
   3  447        1        1 Error(Protocol(InvalidCrc))
   4  241        1        1 Sent(0xa)
   5  283        1        1 Received(0x8)
   6  437        1        1 RxBuf([ 0x1, 0x0, 0x2, 0xe4, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 ])
   7  447        1        1 Error(Protocol(InvalidCrc))
   8  241        1        1 Sent(0xa)
   9  283        1        1 Received(0x8)
  10  437        1        1 RxBuf([ 0x1, 0x0, 0x2, 0xe4, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 ])
  11  447        1        1 Error(Protocol(InvalidCrc))
  12  457        1        1 FailedRetries { retries: 0x3, last_errcode: Protocol(InvalidCrc) }
```

We would expect the `RxBuf` messages to begin with `[3, 0, 0, 0]`, because that's the current protocol version serialized with `u32::to_le_bytes()`.

Looking at the `spi_server_core` ringbuf, we _do_ see that pattern:

```
humility: ring buffer drv_stm32h7_spi_server_core::__RINGBUF in sprot:
  50  494        3        2 WaitISR(0x70012)
  51  464        3        1 Rx(0x0)
  52  494        3        2 WaitISR(0x60012)
  53  464        3        1 Rx(0x3)
  54  494        3        2 WaitISR(0x50012)
  55  464        3        1 Rx(0x0)
  56  494        3        2 WaitISR(0x40012)
  57  464        3        1 Rx(0x0)
  58  494        3        2 WaitISR(0x30012)
  59  464        3        1 Rx(0x0)
  60  494        3        2 WaitISR(0x20012)
  61  464        3        1 Rx(0x3)
  62  494        3        2 WaitISR(0x10012)
  63  464        3        1 Rx(0x0)
   0  322        4        1 Start(read, (0x0, 0x10))
   1  430        4        8 Tx(0x0)
   2  494        4        2 WaitISR(0x100000)
   3  464        4        1 Rx(0x1)
   4  430        4        1 Tx(0x0)
   5  494        4        1 WaitISR(0xf0000)
   6  494        4        1 WaitISR(0xf0002)
   7  464        4        1 Rx(0x0)
   8  430        4        1 Tx(0x0)
   9  494        4        1 WaitISR(0xe0000)
  10  494        4        1 WaitISR(0xe0002)
  11  464        4        1 Rx(0x2)
  12  430        4        1 Tx(0x0)
  13  494        4        1 WaitISR(0xd0000)
  14  494        4        1 WaitISR(0xd0002)
  15  464        4        1 Rx(0xe4)
  16  430        4        1 Tx(0x0)
  17  494        4        1 WaitISR(0xc0000)
  18  494        4        1 WaitISR(0xc0002)
  19  464        4        1 Rx(0x0)
  20  430        4        1 Tx(0x0)
  21  494        4        1 WaitISR(0xb0000)
  22  494        4        1 WaitISR(0xb0002)
  23  464        4        1 Rx(0x0)
  24  430        4        1 Tx(0x0)
  25  494        4        1 WaitISR(0xa0000)
  26  494        4        1 WaitISR(0xa0002)
  27  464        4        1 Rx(0x0)
  28  430        4        1 Tx(0x0)
  29  494        4        1 WaitISR(0x90000)
  30  494        4        1 WaitISR(0x90002)
  31  464        4        1 Rx(0x0)
  32  430        4        1 Tx(0x0)
  33  494        4        1 WaitISR(0x80010)
  34  494        4        1 WaitISR(0x80012)
  35  464        4        1 Rx(0x0)
  36  494        4        2 WaitISR(0x70012)
  37  464        4        1 Rx(0x0)
  38  494        4        2 WaitISR(0x60012)
  39  464        4        1 Rx(0x0)
  40  494        4        2 WaitISR(0x50012)
  41  464        4        1 Rx(0x0)
  42  494        4        2 WaitISR(0x40012)
  43  464        4        1 Rx(0x0)
  44  494        4        2 WaitISR(0x30012)
  45  464        4        1 Rx(0x0)
  46  494        4        2 WaitISR(0x20012)
  47  464        4        1 Rx(0x0)
  48  494        4        2 WaitISR(0x10012)
  49  464        4        1 Rx(0x0)
```

However, it occurs _before_ the `Start` log in the ringbuf!  That's weird and surprising.  Because we're retrying three times, it's likely from a previous message; unfortunately, the ringbuf isn't long enough to record all three attempts.

This looks like a framing issue where the SP and RoT have gotten out of sync.  There are FIFO buffers on each side's SPI peripheral.  Resetting the SP did not fix the issue; power-cycling the whole sled (via Ignition) **does** fix the issue.  This leads us to suspect the RoT's Tx FIFO.

(In particular, I'm wondering what happens if the RoT begins listening to SPI midway through an attempted transaction from the SP, since the SP will try talking to it unconditionally)

andrewjstone:
The primary fix here is for the RoT to detect desynchronization between
the SP and RoT and then wait until synchronization resumes before
proceeding. The RoT and SP are desynchronized if `CSn` is still asserted
*after* the RoT sees the `ssd` bit set in the fifo status register
during its tight loops while clocking in a request or clocking out a
reply to the SP. Synchronization is resumed by the RoT waiting until CSn
is actually de-asserted before either replying to the SP or de-asserting
ROT_IRQ. Detection of actual `CSn` state is performed by polling the
`CHIP_SELECT` gpio pin that corresponds to the `CSn` pin used by the SPI
block on the RoT that sets the `ssa` and `ssd` bits.

The key problem being solved is that the `ssd` bit is saturating, and so
the RoT sprot server may look at this bit and think that the current SPI
transaction is done, even though the bit was set during a prior request.
The actual `CSn` de-assert for the current request arrives after the
RoT reads and clears the `ssd` bit, causing the bit to be set again. The
next request from the SP then comes in, the `ssa` interrupt indicating
`CSn` asserted fires and the RoT goes into a tight loop, processes a
fifo's worth of data and again immediately sees the `ssd` bit set and
ends its response early. In our traces we saw that this exposes itself
by `ROT_IRQ` de- asserting before `CSn` is actually de-asserted and
sprot continues this livelock indefinitely. If after every transaction,
the RoT waits for the `CSn` gpio line to actually become de-asserted
before we consider the SPI transaction complete, we know that we are
operating on a request or reply boundary, and thus the RoT and SP are
resynchronized.

Our salae traces from https://github.com/oxidecomputer/hubris/issues/1507 also showed that we get into this scenario
in the first place by having the RoT start reading from its fifo in
the middle of a transaction. We therefore added support for checking
the `sot` bit on the first read from the fifo to see if this is indeed
the first fifo entry being read after `CSn` was asserted. If this is
not the case the RoT immediately declares a desynchronization, waits
for `CSn` to actually be  deasserted via gpio read, replies to the
SP with the desynchronization error for visibility and moves on to
waiting for the next request. This strategy was discussed in chat with
hubris team members and causes no harm. However, after implementing
and thinking more about this it seems semantically incorrect. We
already have `SprotProtocolError::FlowError` which indicates overruns.
A missing `sot` bit on the first fifo read of a request is actually
just a flow control error, except that instead of the missing bytes
being in the middle of the request, they are at the beginning. In the
common case, this should be detected via the `rxerror` bit, and we
should return a `FlowError`. If there is an actual desynchronization,
we will detect that after the request when we poll the `CHIP_SELECT`
gpio pin. It is totally possible that the RoT misses the first few
bytes of an SP request but is not looking at an `ssd` bit from a prior
transaction. Informing the SP that this common flow error is a very rare
desynchronization event that is triggered on sled restarts and bumping
counters will lead to misleading debugging paths IMO, and we should
probably remove the code that uses the `sot` bit before this is merged.

There were some additional changes made for clarity and correctness.
Cleanup of fifos, errors, and ssa/ssd bits is now self-contained, and
asserting and de-asserting ROT_IRQ happen inside `reply`. I didn't think
it was really necessary to optimize for the `sys_irq_control` syscall
delay with regards to setting `ROT_IRQ` given that we have a 16 byte
fifo and then the SP pauses for 2 ms before reading the rest of a reply
larger than 16 bytes. That gives plenty of time for that syscall to
complete and for the RoT to wait for the CSn asserted IRQ wakeup after
asserting ROT_IRQ. This change makes the code more linear and removes
some unnecessary state. Testing so far has shown no difference in error
rate.

On the SP side, `Desynchronization` errors are now recoverable.

It should also be noted how adding the new `Desyncrhonization` error
will affect upgrades. It is a new varient to the `SprotProtocolError`
enum and therfore code that is unaware of this variant will not be
able to deserialize it. Specifically:

 1. Because the RoT is upgraded before the SP in mupdate this means that
the SP code will see a `SprotProtocolError::Deserialization` error in
the case of desynchronization. This is already a recoverable error and
the behavior of the SP sprot server should be the same, except that if
retries are exceeded for some reason, the wrong error could be plumbed
up to the control-plane-agent and MGS. This is exceedingly unlikely for
this specific error, except for perhaps in the flow control case where
we use `sot` described above.

 2. Until MGS is updated, if the new error variant gets plumbed upwards
it will be seen as an incompatible protocol error. This is not really
a big deal in this case, as we are still mupdating and this is the
only related error that can occur this way until the system is further
upgraded.


andrewjstone:
Before merging, we'll need to merge in https://github.com/oxidecomputer/management-gateway-service/pull/121. However, before we do that, we'll need to deal with https://github.com/oxidecomputer/management-gateway-service/issues/123.

drv/sprot-api/src/lib.rs
```diff
@@ -520,32 +525,35 @@ pub struct RotIoStats {
...
-    // Number of times pulsing CSn failed.
+    /// Number of times pulsing CSn failed.
     pub csn_pulse_failures: u32,
 
-    // Number of timeouts, while waiting for a reply
+    /// Number of timeouts, while waiting for a reply
     pub timeouts: u32,
+
+    /// Number of times the RoT has reported that it was desynchronized
+    pub desynchronized: u32,
```
andrewjstone, line 556: We should probably increment this somewhere.

andrewjstone:
I should also note that this code needs testing on real hardware and not the gimletlet/rot-carrier. I'll need to update `app/oxide-rot-1/app.toml` to do this. The image will also have to be signed by non-bartholemew keys I believe.

andrewjstone:
> > However, after implementing
> > and thinking more about this it seems semantically incorrect. We
> > already have `SprotProtocolError::FlowError` which indicates overruns.
> > A missing `sot` bit on the first fifo read of a request is actually
> > just a flow control error, except that instead of the missing bytes
> > being in the middle of the request, they are at the beginning. In the
> > common case, this should be detected via the `rxerror` bit, and we
> > should return a `FlowError`
> 
> I'm not sure about this. Are we _sure_ this is going to be detected via `rxerror`? If bytes are being dropped I would think that the hardware would still see it as the start of transfer and still set the bit.

Interesting. I thought all dropped bytes were detected by rxerror, except perhaps on startup before we initialize spi or because we mistakenly clear the error. I didn't really think about the case that the SOT bit would still be set even if early bytes were dropped. That seems quite problematic for its usage here. The problem we are solving for is missing initial bytes.

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -223,117 +234,181 @@ impl Io {
...
-                let read = self.spi.read_u16();
+                let (read, sot) = self.spi.read_u16_with_sot();
+                if first_read {
+                    first_read = false;
+                    if !sot {
+                        self.stats.desynchronized =
+                            self.stats.desynchronized.wrapping_add(1);
+                        return Err(IoError::Desynchronized);
+                    }
+                }
```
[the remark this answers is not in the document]
andrewjstone: Yes, but it would be detected when trying to reply because CSn would still be asserted while SSD is set.

andrewjstone:
The repro test script has been running on my gimletlet/rot-carrier for over 3 hours now with no reproductions.

I also just installed a this image on a sled on madrid via faux-mgs and it installed fine and state was accessible.

andrewjstone:
The script I am running to reproduce the issue is:

```bash
#!/usr/bin/env bash


MGS_REPO=${MGS_REPO:=~/oxide/management-gateway-service}
HUBRIS_REPO=${HUBRIS_REPO:=~/oxide/hubris}
HUMILITY_ENVIRONMENT=${HUMILITY_ENVIRONMENT:=~/Users/ajs/oxide/humility_env.json}

HENV=$HUMILITY_ENVIRONMENT
rota_reset='humility -t rot-carrier --archive-name a reset'
ROT_RESET_COMMAND=${RESET_COMMAND:="$rota_reset"}

# Number of times in a row `faux-mgs state` must fail to be a test success
MAX_FAILS=${MAX_FAILS:=110}


ready_to_reset=false
failures_in_a_row=0

# We want to wait for a few times through the loop to make sure the reset
# command completes before trying again. Otherwise, we will get an error:
#
#   humility reset failed: Probe could not be created
#
#  Caused by:
#      0: hidapi error: hid_error is not implemented yet
#  	    1: hidapi error: hid_error is not implemented yet
#
MAX_GRACE_CYCLES=20
reset_grace=$MAX_GRACE_CYCLES

while true; do
  # Run our command
  output=`$MGS_REPO/target/release/faux-mgs --interface en6 --log-level critical state`
  echo "$output" | grep -q "RoT state: Ok"
  if [[ $? -eq 0 ]]; then
    failures_in_a_row=0
    echo "."
  else
    failures_in_a_row=$((failures_in_a_row+1))
    echo "+"
  fi

  # increment our grace period counter
  reset_grace=$((reset_grace+1))

  # Are we done?
  if [[ failures_in_a_row -eq MAX_FAILS ]]; then
    echo "Test Complete: found faulty sprot"
    break
  fi

  # Should we reset the RoT to try again?
  choice=$RANDOM
  sleep=$((1/(choice % 100 + 1)))
  if [[ $failures_in_a_row -eq 0 && $reset_grace -gt $MAX_GRACE_CYCLES ]]; then
    # reset our grace period counter
    reset_grace=0
    # Flip a coin to see if we should sleep before we try to reset or after
    # We right shift to create a boolean
    let "choice >>= 14"
    if [[ $choice -eq 0 ]]; then
      HUMILITY_ENV=$HENV bash -c "sleep ${sleep} && ${ROT_RESET_COMMAND} &" &
    else 
      HUMILITY_ENV=$HENV bash -c "${ROT_RESET_COMMAND} &" &
    fi 
  else
    sleep 0.01
  fi
done
```

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -223,117 +234,181 @@ impl Io {
...
+        let mut read_fifo = || {
             while self.spi.has_entry() {
                 bytes_received += 2;
-                let read = self.spi.read_u16();
+                let (read, sot) = self.spi.read_u16_with_sot();
+                if first_read {
+                    first_read = false;
+                    if !sot {
+                        self.stats.desynchronized =
+                            self.stats.desynchronized.wrapping_add(1);
```
mkeeter, line 282: Maybe `saturating_add` instead?

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -223,117 +234,181 @@ impl Io {
...
+    // this.
+    fn wait_for_csn_deasserted(&mut self) -> Result<(), DesynchronizedError> {
+        let mut result = Ok(());
+        while self.gpio.read_val(CHIP_SELECT) != Value::One {
+            ringbuf_entry!(Trace::Desynchronized);
+            if result.is_ok() {
+                self.stats.desynchronized =
+                    self.stats.desynchronized.wrapping_add(1);
+            }
+            result = Err(DesynchronizedError);
```
mkeeter, line 255: Nit: maybe move this into the conditional, to make it more obvious that the conditional is only executed on the first transition from `Ok` -> `Err`?

drv/lpc55-sprot-server/src/main.rs
andrewjstone: I made this match all the other counters. I specifically did that because it seems more useful to know if a counter is moving than if it saturated at one point in time. Although, in either case we shouldn't really come close to hitting this.

drv/lpc55-sprot-server/src/main.rs
andrewjstone: good call!

drv/lpc55-sprot-server/src/main.rs
mkeeter: Okay, I don't feel strongly about this!

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -182,31 +186,38 @@ fn main() -> ! {
...
+
         ringbuf_entry!(Trace::Stats(io.stats));
         io.reply(&tx_buf[..rsp_len]);
     }
 }
 
+struct DesynchronizedError;
 impl Io {
     // Wait for chip select to be asserted
     // Assert ROT_IRQ if this is a reply
```
labbott, line 216: Remove this comment if we're not doing this part anymore

drv/lpc55-sprot-server/src/main.rs
```diff
@@ -223,117 +234,181 @@ impl Io {
...
 
-    fn reply(&mut self, tx_buf: &[u8]) {
+    pub fn cleanup_after_request(&mut self) -> Result<(), DesynchronizedError> {
+        let result = self.wait_for_csn_deasserted();
+        self.cleanup();
+        result
+    }
+
+    // Reply to the SP
+    // Our fifos are guaranteed to be empty at this point
```
labbott, line 340: Is there any way we could add an assert/error for this?

labbott (commented):
I think this looks good with the few changes suggested. Given how there's almost no debugging options if this link is down are there any other options for reporting an error or making it possible to get state out? We can file as a tracking item if necessary.

mkeeter (approved):

drv/sprot-api/src/lib.rs
andrewjstone: I think this will require more refactors than I'm willing to make right now on this PR.

drv/lpc55-sprot-server/src/main.rs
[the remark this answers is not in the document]
andrewjstone: So, after thinking more about this, I withdraw most of my concern. The code works as is, and I haven't actually seen this hit during upgrades even though I've seen a handful of rxerrors(which is on par with what our current error rates are).

drv/lpc55-sprot-server/src/main.rs
andrewjstone: Good catch. Begone!

drv/lpc55-sprot-server/src/main.rs
andrewjstone: Not really. There's also always an inherent TOCTUO here. We could check the fifos, but they may not be empty right after the check.

If they aren't empty we'll end up with some sort of error here or at the SP. I'll be more detailed in the comment.

andrewjstone:
> I think this looks good with the few changes suggested. Given how there's almost no debugging options if this link is down are there any other options for reporting an error or making it possible to get state out? We can file as a tracking item if necessary. 

I'm not sure how we would do that. I think we would have to raise the fault pin and then reset the RoT.

labbott (approved):

PR #1524, opened 2023-09-08
Bump gateway-messages and zerocopy to latest

jgallagher:
We also enable the `extra-traits` feature of `syn`, which `derive-idol-err` needs but we were previously only picking up via feature unification: zerocopy-derive 0.3.1 enabled it, but zerocopy-derive 0.6.4 depends on syn2 so we lost it.


mkeeter (approved):
grumble grumble feature unification

Issue #1511, opened 2023-08-28
Add a way to ask the RoT what signing keys it expects via sprot


jgallagher:
See https://github.com/oxidecomputer/omicron/issues/3954 for context; if we want a single TUF repo to be able to serve:

* different racks, each of which uses different keys (e.g., a `prod-rel` rack and a `staging/dev` rack), or
* a rack with a mix of acceptable keys (e.g., a rack where some sleds want `staging/dev` and others want `bart`)

then we need a way to ask at runtime which RoT image a given sled wants.

PR #1525, opened 2023-09-08
Add support for reading out CMPA/CFPA regions

labbott:
The CMPA/CFPA regions of the RoT contain the settings for which certificates are required for a particular device. Allow those settings to be extracted to verify images before installing.


drv/lpc55-update-server/src/main.rs
```diff
@@ -883,6 +915,33 @@ fn copy_from_caboose_chunk(
     Ok(())
 }
 
+fn copy_from_flash_range(
```
labbott, line 918: I don't love that this is basically a copy/paste of the caboose function but I was finding the different errors (`UpdateError`  vs `RawCabooseError`) to be a pain to handle. Maybe I should take a close look?

labbott:
See #1511

drv/lpc55-update-server/src/main.rs
```diff
@@ -69,6 +69,7 @@ const BLOCK_SIZE_BYTES: usize = BYTES_PER_FLASH_PAGE;
 const MAX_LEASE: usize = 1024;
 const HEADER_BLOCK: usize = 0;
 
+const CMPA_FLASH_WORD: u32 = 0x9E40;
 const CFPA_PING_FLASH_WORD: u32 = 0x9E00;
 const CFPA_PONG_FLASH_WORD: u32 = 0x9E20;
 const CFPA_SCRATCH_FLASH_WORD: u32 = 0x9DE0;
```
labbott, line 75: I should probably just bite the bullet and fix the update server to just act on page addresses instead of words given I had a bug in development based on this confusion.

drv/lpc55-update-server/src/main.rs
```diff
@@ -425,11 +426,43 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+            dest,
+        )?;
+        Ok(())
+    }
 }
 
 impl ServerImpl<'_> {
     fn cfpa_word_number_and_version(
         &mut self,
+        lower: bool,
```
andrewjstone, line 465: Can we use a more descriptive name here please? Maybe `read_inactive`?

drv/lpc55-update-server/src/main.rs
andrewjstone: You could call `map_err` ion the result if it's the same code other than error.

drv/lpc55-update-server/src/main.rs
```diff
@@ -462,7 +494,7 @@ impl ServerImpl<'_> {
         &mut self,
     ) -> Result<(SlotId, Option<SlotId>, Option<SlotId>), UpdateError> {
         let (cfpa_word_number, cfpa_version) =
-            self.cfpa_word_number_and_version()?;
+            self.cfpa_word_number_and_version(false)?;
```
andrewjstone, line 497: Nit: I dislike passing bools, because you don't know what it means at the call site. However, clearly it can make the code more succinct inside the function than an enum. To work across that tension, many of us on the control plane side declare a temporary named boolean variable and then pass that into the method rather than the literal `true` or `false`.

drv/lpc55-update-server/src/main.rs
labbott: I tried to do this and it caused more confusion. Needs more thought so punting on this for now.

drv/lpc55-update-server/src/main.rs
labbott: Yeah I tried looking at this again and it also turns out that because the page is fixed size vs caboose this also doesn't quite work. I think the separate function is fine for now.

andrewjstone (approved):

drv/lpc55-update-server/src/main.rs
flihp: I'm struggling with the same thing w/r/t plumbing through the Attest.log function. It's only slightly different than the Attest.cert and Update functions but different enough to make finding a common implementation difficult. We'll have plenty of opportunities to find an approach w/ less copy-pasta in the future.

flihp (approved):

PR #1528, opened 2023-09-13
Update lpc55xpresso app-sprot.toml accounting for recent changes.

flihp:
This removes use of the now removed ITM feature in jefe as well as increasing the flash size required by recent sprot changes.


andrewjstone (approved):
LGTM

PR #1530, opened 2023-09-14
Completely remove ITM and semihosting

labbott:
We no longer have any need for these features and they are mostly just sitting around as some copy/paste. Remove them.


flihp (approved):
When I did #1528 yesterday I didn't realize there was more cleanup to be done. Thanks for this.

PR #1532, opened 2023-09-15
Clippy Fixes

labbott:


lib/lpc55-rot-startup/src/images.rs
```diff
@@ -129,10 +129,10 @@ impl Image {
             return false;
         }
 
-        return true;
+        true
```
mkeeter, line 132: We could just return `valid` here, removing the conditional above as well.

task/dumper/src/main.rs
```diff
@@ -90,7 +90,7 @@ impl idl::InOrderDumperImpl for ServerImpl {
 
         let mut nread = 0;
         let mut nwritten = 0;
-        let mut reg = 0;
+        let reg = 0;
```
mkeeter, line 93: Maybe just delete this and make the loop `for r in 0..=31`?

task/packrat/build.rs
```diff
@@ -11,14 +11,13 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
-    match build_util::target_board().as_deref() {
-        Some("gimlet-b" | "gimlet-c" | "gimlet-d" | "gimlet-e") => {
-            panic!(concat!(
-                "packrat's `gimlet` feature should be enabled when ",
-                "building for gimlets",
-            ))
-        }
-        _ => (),
+    if let Some("gimlet-b" | "gimlet-c" | "gimlet-d" | "gimlet-e") =
+        build_util::target_board().as_deref()
```
mkeeter, line 15: 🤯

task/uartecho/src/main.rs
```diff
@@ -190,8 +190,11 @@ fn configure_uart_device() -> Usart {
 
     let hardware_flow_control = cfg!(feature = "hardware_flow_control");
 
+    #[allow(clippy::needless_late_init)]
     let usart;
+    #[allow(clippy::needless_late_init)]
     let peripheral;
+    #[allow(clippy::needless_late_init)]
```
mkeeter, line 197: This would be pretty easy to actually fix – delete these lines and add `let`s in each `cfg_if` branch below.

drv/transceivers-server/src/main.rs
```diff
@@ -657,7 +658,7 @@ fn main() -> ! {
         #[derive(Copy, Clone, Enum)]
         enum Timers {
             I2C,
-            SPI,
+            Spi,
```
mkeeter, line 661: I'd vote for a clippy exception here, since having `I2C` and `Spi` in the same enum is awkward.

drv/fpga-devices/src/ecp5_spi.rs
```diff
@@ -96,14 +100,17 @@ impl<S: SpiServer> Ecp5Driver for Ecp5UsingSpi<S> {
         self.user_design_reset_duration
     }
 
+    #[allow(clippy::unit_arg)]
```
mkeeter, line 103: For all the `allow(clippy::unit_arg)` variances in this file, it's the same +1 LOC to split the return to a separate line, e.g.
```rust
    fn configuration_read(&self, data: &mut [u8]) -> Result<(), Self::Error> {
        self.configuration_port.read(data)?;
        Ok(())
    }
```
We might as well do that instead?

task/uartecho/src/main.rs
labbott: I also like the extra `let`s if others agree with the style

mkeeter (approved):

flihp (approved):
LGTM

PR #1534, opened 2023-09-15
gimlet-seq-server: unwrap diet

cbiffle:
gimlet-seq-server unwraps (panics) on a lot of external conditions. I'm still pretty sure we want to fix that, but for the time being, if we're going to unwrap everywhere, we can at least unwrap_lite.

Unwrap Debug-formats the error, which pulls in all the Debug-format machinery. That machinery goes largely unused in practice, since we only format the first 128 characters of a panic anyway. Changing all these unwraps to unwrap_lite knocks 16% off the binary size and bumps it into the next smaller size class, freeing 64 kiB of space (which was mostly lost to fragmentation) on Gimlet as of this writing.


mkeeter (approved):
This does what it says on the label!

PR #1535, opened 2023-09-15
STM32G0 size improvements.

cbiffle:
This applies some size improvements to enable halfway-useful Hubris systems on 32kiB STM32G030 parts, which happens to be the part I'm putting on the 0xCon demo boards. Net effects:

- Jefe can now be compiled without ringbuf for a savings of a few kiB. This seems acceptable since it's somewhat unusual to debug Jefe issues, particularly in simpler systems.

- Kernel panic messages can now be disabled, which knocks 8 kiB off. When built in this mode, the kernel will still set the KERNEL_HAS_FAILED boolean for Humility, but will not fill out the KERNEL_EPITAPH block (which won't exist). This seems like an ok compromise since kernel panics on Hubris are very uncommon.

I've added an app (oxcon2023g0) for the new board and done the shotgun changes required to make it work (someday I still want to fix that).

Current usage on that app are:

  Used:
    flash:   0x7000 (87%)
    ram:     0x1100 (53%)

  flash:
        ADDRESS  |  PROGRAM  |  USED |  SIZE | LIMIT
      0x08000000 | kernel    | 10752 | 10752 | (fixed)
      0x08002a00 | idle      |    68 |   128 | 128
      0x08002a80 | -padding- |    -- |  1408 |
      0x08003000 | sys       |  1652 |  2048 | 2048
      0x08003800 | user_leds |  1260 |  2048 | 2048
      0x08004000 | hiffy     |  7912 |  8192 | 8192
      0x08006000 | jefe      |  3488 |  4096 | 4096
      0x08007000 | --end--   |       |       |

...so there's 4 kiB contiguous at the end available for use, plus space for a 1024-byte task in fragmentation space between idle and sys.

### DIFF
--- a/sys/kern/src/fail.rs
+++ b/sys/kern/src/fail.rs
@@ -19,21 +19,24 @@
 //!   number of bytes written isn't recorded anywhere; instead, for printing,
 //!   trim off any trailing NUL bytes.
 
-use core::fmt::{Display, Write};
-use core::sync::atomic::Ordering;
+#[cfg(not(feature = "nano"))]
+use core::{fmt::{Display, Write}, sync::atomic::Ordering};
 
 /// Flag that gets set to `true` by all failure reporting functions, giving
 /// tools a one-stop-shop for doing kernel triage.
 #[used]
 static mut KERNEL_HAS_FAILED: bool = false;
 
+#[cfg(not(feature = "nano"))]
 const EPITAPH_LEN: usize = 128;
 
 /// The "epitaph" buffer records up to `EPITAPH_LEN` bytes of description of the
 /// event that caused the kernel to fail, padded with NULs.
+#[cfg(not(feature = "nano"))]
 #[used]
 static mut KERNEL_EPITAPH: [u8; EPITAPH_LEN] = [0; EPITAPH_LEN];
 
+#[cfg(not(feature = "nano"))]
 fn begin_epitaph() -> &'static mut [u8; EPITAPH_LEN] {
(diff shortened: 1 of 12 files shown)


app/oxcon2023g0/.gitignore
```diff
@@ -0,0 +1,12 @@
+**/*.rs.bk
```
mkeeter, line 1: Was this committed intentionally?  Should these things be in the top-level `.gitignore`  instead?

sys/kern/src/fail.rs
```diff
@@ -19,21 +19,26 @@
 //!   number of bytes written isn't recorded anywhere; instead, for printing,
 //!   trim off any trailing NUL bytes.
 
+#[cfg(not(feature = "nano"))]
 use core::fmt::{Display, Write};
+#[cfg(not(feature = "nano"))]
 use core::sync::atomic::Ordering;
```
mkeeter, line 25: Nit:
```suggestion
#[cfg(not(feature = "nano"))]
use core::{fmt::{Display, Write}, sync::atomic::Ordering};
```
?

mkeeter (approved):

app/oxcon2023g0/.gitignore
cbiffle: It's a copy-paste from our other apps, I'm not actually sure who added it or what program generates `.bk` files. Can remove.

sys/kern/src/fail.rs
cbiffle: Nice.

PR #1536, opened 2023-09-15
 drv/user-leds: optionally blink some LEDs at start 

cbiffle:
We've been including the pong task in most basic/starter images for the
sole purpose of blinking some LEDs.

For the simplest case of a small board with a single debug LED and not a
lot of flash, this is quite wasteful (burns about a kiB). To conserve
resources on such boards, I've added a mechanism for configuring the
user-leds driver to start some LEDs blinking automatically.

Unlike with pong, the LEDs will all blink together, not in a sequence.
Such is the cost of frugality. On boards with one LED the distinction is
not material.

This imposes no size cost on existing uses of user-leds because the
configuration is evaluated at compile time, and will compile away. On
the oxcon2023 board it costs about 16 bytes to set the LED a-blinkin'.

### DIFF
--- a/lib/task-config/src/lib.rs
+++ b/lib/task-config/src/lib.rs
@@ -142,6 +142,8 @@ pub fn task_config(tokens: TokenStream) -> TokenStream {
     let config = build_util::task_config::<toml::Value>().unwrap();
 
     let input = parse_macro_input!(tokens as Config);
+    let fields = input.items.iter();
+
     let values = input
         .items
         .iter()
@@ -158,7 +160,6 @@ pub fn task_config(tokens: TokenStream) -> TokenStream {
 
     let app_toml_path = std::env::var("HUBRIS_APP_TOML")
         .expect("Could not find 'HUBRIS_APP_TOML' environment variable");
-    let fields = input.items.iter();
 
     // Once `proc_macro::tracked_env::var` is stable, we won't need to use
     // this hack, but until then, we include the app TOML file to force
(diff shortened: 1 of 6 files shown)


lib/task-config/src/lib.rs
```diff
@@ -174,3 +175,57 @@ pub fn task_config(tokens: TokenStream) -> TokenStream {
...
+        // rebuilds if it changes (and trust it's optimized out by the compiler)
+        quote! {
+            const APP_TOML_TO_ENSURE_REBUILD: &[u8] = include_bytes!(#app_toml_path);
+            struct Config {
+                #(#fields),*
+            }
+            const TASK_CONFIG: Option<Config> = None;
+        }
+        .into()
+    }
```
mkeeter, line 229: How about building just the `Config` in the conditional, to deduplicate the written output?
```rust
    let cfg = if let Ok(config) = build_util::task_config::<toml::Value>() {
        let values = input
            .items
            .iter()
            .map(|f| {
                let ident = f.ident.as_ref().expect("Missing ident");
                let v = config.get(ident.to_string()).expect(&format!(
                    "Missing config parameter in TOML file: {}",
                    ident.to_string()
                ));
                let vs = config_to_token(&f.ty, v);
                quote! { #ident: #vs }
            })
            .collect::<Vec<_>>();

        quote! {
            Some(Config { #(#values),* })
        }
    } else {
        quote! { None }
    };

    quote! {
        // Once `proc_macro::tracked_env::var` is stable, we won't need to use
        // this hack, but until then, we include the app TOML file to force
        // rebuilds if it changes (and trust it's optimized out by the compiler)
        const APP_TOML_TO_ENSURE_REBUILD: &[u8] = include_bytes!(#app_toml_path);
        struct Config {
            #(#fields),*
        }
        const TASK_CONFIG: Option<Config> = #cfg;
    }.into()
```

cbiffle: Makes sense.

mkeeter (approved):

PR #1540, opened 2023-09-17
STM32 I2C size improvements, STM32G030 support

cbiffle:
On the 0xCon demo board, I included a "standard" Sparkfun/Adafruit I2C connector. I later realized that we don't actually have I2C support working on the G030 for space reasons. (Last night, while I was attempting to sleep.) So this morning I've thoroughly beaten the i2c server with the "get smaller" stick, and have it fitting in under 4 kiB. (For comparison, on donglet, the next larger target, it takes 16 kiB.)

This has required turning panic messages off (code contains a _lot_ of panics) and disabling ringbuf (code contains a _lot_ of ringbuf). Among other things.

There appear to be more opportunities for flash reclamation in these crates -- as far as I can tell nobody's been keeping an eye on overflow panics or bounds check panics, which are generating a lot of goo. So, I may have follow-ups. TBD.


app/oxcon2023g0/app.toml
```diff
@@ -48,9 +48,34 @@ task-slots = ["sys"]
...
+
+[config]
+[[config.i2c.controllers]]
+controller = 1
+
+[config.i2c.controllers.ports.A]
+scl.pin = 11
+sda.pin = 10
+af = 6
+
```
mkeeter, line 80: Nit: bonus whitespace here

mkeeter (approved):

PR #1541, opened 2023-09-17
Hiffy size improvements for stm32g0 I2C

cbiffle:
This applies two changes to make Hiffy practical on the 32 kiB STM32G030, with I2C support enabled. As usual there are verbose details in the commit messages, but in summary,

1. Made the generic "send" operations in Hiffy optional and turned them off on STM32G030. They're fairly large, at 1456 bytes. We could probably make them smaller but that's a PR for another day.
2. Ground off 256 bytes from Hiffy's STM32G0 I2C support through refactoring and sweat.

Note that, at the time this PR was created, this was blocked on a HIF PR: https://github.com/oxidecomputer/hif/pull/6


Cargo.toml
```diff
@@ -115,7 +115,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 # Oxide forks and repos
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
-hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
+hif = { git = "https://github.com/oxidecomputer/hif", default-features = false, branch = "failure-from" }
```
mkeeter, line 118: Making a note to update this once the HIF branch is merged

task/hiffy/Cargo.toml
```diff
@@ -41,6 +41,7 @@ testsuite = [ "test-api" ]
 i2c = []
 gpio = []
 spi = []
+send = []
```
mkeeter, line 44: I think we want this enabled on every non-G030 board; otherwise, `humility hiffy -c Foo.bar ...` won't work.  Enabling it with the `stm32h7` and `lpc55` (and `stm32g031`?) features is probably the easiest way to do this.

Alternatively, we could use the `micro` feature to _also_ mean that the `send*` functions should be disabled.  This breaks the "purely additive feature" rule, though...

cbiffle: Note that the send code is not _optional_ on stm32h7/lpc55/etc, only on g0, rn. But I can add the send feature to the others like it does something if that'd be clearer.

cbiffle:
@mkeeter this should be ready to go

mkeeter (approved):

PR #1542, opened 2023-09-19
 Final(?) fixes for Oxcon board.

cbiffle:
- Add insomniac feature to suppress sleep at WFI because Humility
  doesn't know how to attach to an STM32G0 that is sleeping.

- Fix bug in user leds "blink on start" feature that failed to actually
  set up blink on start, because there are several pieces of data to
  update and I missed two of them.

### DIFF
--- a/drv/user-leds/src/main.rs
+++ b/drv/user-leds/src/main.rs
@@ -104,7 +104,6 @@ cfg_if::cfg_if! {
 
 struct ServerImpl {
     blinking: EnumMap<Led, bool>,
-    blink_state: bool,
 }
 
 impl idl::InOrderUserLedsImpl for ServerImpl {
@@ -144,11 +143,10 @@ impl idl::InOrderUserLedsImpl for ServerImpl {
         index: usize,
     ) -> Result<(), RequestError<LedError>> {
         let led = Led::from_usize(index).ok_or(LedError::NotPresent)?;
-        let already_blinking = self.blinking.values().any(|b| *b);
+        let any_blinking = self.blinking.values().any(|b| *b);
         self.blinking[led] = true;
-        if !already_blinking {
-            led_on(led);
-            self.blink_state = false;
+
+        if !any_blinking {
             sys_set_timer(
                 Some(sys_get_timer().now + BLINK_INTERVAL),
                 notifications::TIMER_MASK,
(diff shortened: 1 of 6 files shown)


drv/user-leds/src/main.rs
```diff
@@ -196,10 +191,15 @@ fn main() -> ! {
         for &led in config.blink_at_start {
             blinking[led] = true;
         }
+        if !config.blink_at_start.is_empty() {
```
mkeeter, line 194:
```suggestion
        if config.blink_at_start.is_some() {
```
?

drv/user-leds/src/main.rs
```diff
@@ -144,11 +143,10 @@ impl idl::InOrderUserLedsImpl for ServerImpl {
...
     ) -> Result<(), RequestError<LedError>> {
         let led = Led::from_usize(index).ok_or(LedError::NotPresent)?;
-        let already_blinking = self.blinking.values().any(|b| *b);
+        let any_blinking = self.blinking.values().any(|b| *b);
         self.blinking[led] = true;
-        if !already_blinking {
-            led_on(led);
-            self.blink_state = false;
+
+        if !any_blinking {
```
mkeeter, line 149: I'd suggest keeping the `led_on(led)` call here, so that going from off to blinking is obviously right away (otherwise, you have to wait for a `BLINK_INTERVAL`).

drv/user-leds/src/main.rs
cbiffle: It's a slice, I don't think there's an `is_some` equivalent.

drv/user-leds/src/main.rs
cbiffle: So, in practice, if I leave that line in, the LED starts off. I'm not sure why. Was it blinking immediately before?

drv/user-leds/src/main.rs
mkeeter: aha, objection retracted

drv/user-leds/src/main.rs
mkeeter: That confuses me!  The point of this line was to get immediate feedback when calling `led_blink` (if other LEDs weren't already blinking), and I'm pretty sure it worked.  I'd have to think through it a little more to understand why it's not working in this case.

cbiffle: Yeah, I assumed it had something to do with the previously ambiguous nature of `blink_state` (where `blink_state` of `false` appeared to mean on?). But, had to remove this to make the feedback immediate.

mkeeter (approved):

PR #1560, opened 2023-11-06
trace MAX5970 status/faults/max current/max voltage on V12 rails

bcantrill:


drv/i2c-devices/src/max5970.rs
```diff
@@ -240,6 +290,78 @@ impl Max5970 {
...
+            (
+                self.read_reg(Register::max_chx_cs_msb_ch1)?,
+                self.read_reg(Register::max_chx_cs_lsb_ch1)?,
+            )
+        } else {
+            (
+                self.read_reg(Register::max_chx_cs_msb_ch2)?,
+                self.read_reg(Register::max_chx_cs_lsb_ch2)?,
+            )
+        };
```
mkeeter, line 356:
```suggestion
        let (msb_reg, lsb_reg) = if self.rail == 0 {
            (Register::max_chx_cs_msb_ch1, Register::max_chx_cs_lsb_ch1)
        } else {
            (Register::max_chx_cs_msb_ch2, Register::max_chx_cs_lsb_ch2)
        };
        let msb = self.read_reg(msb_reg)?;
        let lsb = self.read_reg(lsb_reg)?;
```

drv/i2c-devices/src/max5970.rs
```diff
@@ -240,6 +290,78 @@ impl Max5970 {
...
+            (
+                self.read_reg(Register::max_chx_mon_msb_ch1)?,
+                self.read_reg(Register::max_chx_mon_lsb_ch1)?,
+            )
+        } else {
+            (
+                self.read_reg(Register::max_chx_mon_msb_ch2)?,
+                self.read_reg(Register::max_chx_mon_lsb_ch2)?,
+            )
+        };
```
mkeeter, line 339:
```suggestion
        let (msb_reg, lsb_reg) = if self.rail == 0 {
            (Register::max_chx_mon_msb_ch1, Register::max_chx_mon_lsb_ch1)
        } else {
            (Register::max_chx_mon_msb_ch2, Register::max_chx_mon_lsb_ch2)
        };
        let msb = self.read_reg(msb_reg)?;
        let lsb = self.read_reg(lsb_reg)?;
```

drv/i2c-devices/src/max5970.rs
```diff
@@ -218,6 +218,56 @@ pub enum Register {
     cubf_ba_chx_i = 0x47,
 }
 
+struct MonRange(u8);
```
mkeeter, line 221: Docstrings for `MonRange` and `Status2` would be helpful here.

task/power/src/main.rs
```diff
@@ -349,6 +359,45 @@ macro_rules! max5970_controller {
     };
 }
 
+fn trace_max5970(dev: &Max5970, sensor: SensorId) {
+    if let Ok(Volts(volts)) = dev.max_vout() {
+        use drv_i2c_devices::max5970::Register;
+
+        if volts < 12.0 {
```
mkeeter, line 366: What's the thought here?  It seems like we'd still want to log even if the max VOUT is < 12V, since we could be at 11.97V, or something _interesting_ could be happening

drv/i2c-devices/src/max5970.rs
mkeeter: (You could also change the existing `read_vout` and `read_iout` implementations to match, even though they aren't touched in this PR)

task/power/src/main.rs
```diff
@@ -349,6 +359,45 @@ macro_rules! max5970_controller {
...
+            fault2: match dev.read_reg(Register::fault2) {
+                Ok(reg) => reg,
+                _ => return,
+            },
+            max_current: match dev.max_iout() {
+                Ok(Amperes(amps)) => amps,
+                _ => return,
+            },
+            max_voltage: volts,
+        });
```
mkeeter, line 397: I'd vote to explicitly log errors here, since they could be interesting.  Something like
```suggestion
        // Helper function to read a register or log the failure
        let f = |reg| match dev.read_reg(reg) {
            Ok(reg) => Some(reg),
            Err(err) => {
                ringbuf_entry!(Trace::Max5970ReadFailed { reg, err });
                None
            }
        };
        let Some(status0) = f(Register::status0) else { return; };
        let Some(status3) = f(Register::status3) else { return; };
        let Some(fault0) = f(Register::fault0) else { return; };
        let Some(fault1) = f(Register::fault1) else { return; };
        let Some(fault2) = f(Register::fault2) else { return; };

        let max_current = match dev.max_iout() {
            Ok(Amperes(amps)) => amps,
            Err(err) => {
                ringbuf_entry!(Trace::Max5970MaxCurrentFailed { err });
                return;
            }
        };

        ringbuf_entry!(Trace::Max5970 {
            sensor,
            status0,
            status3,
            fault0,
            fault1,
            fault2,
            max_current,
            max_voltage: volts,
        });
```

along with two new ringbuf variants:
```diff
diff --git a/task/power/src/main.rs b/task/power/src/main.rs
index cd5b0623..93a77268 100644
--- a/task/power/src/main.rs
+++ b/task/power/src/main.rs
@@ -48,6 +48,13 @@ enum Trace {
         max_current: f32,
         max_voltage: f32,
     },
+    Max5970ReadFailed {
+        reg: drv_i2c_devices::max5970::Register,
+        err: ResponseCode,
+    },
+    Max5970MaxCurrentFailed {
+        err: ResponseCode,
+    },
     None,
 }
```

bcantrill: I have mixed feelings about this because I don't want to flood the ring buffer with a failed device and miss the PGs on others. (We will be able to deduce failures by the absence of expected devices.)

task/power/src/main.rs
```diff
@@ -349,6 +359,45 @@ macro_rules! max5970_controller {
...
 
+fn trace_max5970(dev: &Max5970, sensor: SensorId) {
+    if let Ok(Volts(volts)) = dev.max_vout() {
+        use drv_i2c_devices::max5970::Register;
+
+        if volts < 12.0 {
+            return;
+        }
+
+        ringbuf_entry!(Trace::Max5970 {
```
wesolows, line 370: Maybe include `status1` here too?  In the implementations of interest, none of the bits can change, and `PROT` and `MODE` are sampled only at reset.  But `ONx` is a dynamic input and all these bits are probably as interesting as those in `status3`.

task/power/src/main.rs
bcantrill: I wanted to avoid the 3V3 rails on U8 on the Sharkfin. Will add a comment to that effect...

mkeeter (approved):

PR #1568, opened 2023-11-28
make Gimletlet use TOML inheritance

Aaron-Hartwig:
This change reworks the various Gimletlet applications to utilize TOML inheritance. There is now a `base-gimletlet2.toml` which contains the core Hubris tasks required for the SP and declares the I2C and SPI buses. `app-mgmt.toml` utilizes a board type of `gimletlet1` which lead me to make the distinction of `gimletlet2` in the base TOML title.

The goal here was to reduce copy-paste duplication and also decouple the core gimletlet functionality from the various applications folks may have. Having these bits separated will make it easier for others to make custom gimletlet applications without worrying about legacy things that were committed to the old `app.toml` that was largely shared. That application was recreated in `app-original.toml` and still maintains the name of `gimletlet`.


app/gimletlet/app-original.toml
```diff
@@ -1,68 +1,23 @@
 name = "gimletlet"
```
labbott: I'd suggest just keeping this as `app.toml` since that's what we already have

app/gimletlet/app-meanwell.toml
```diff
@@ -1,66 +1,18 @@
...
-max-sizes = {flash = 2048, ram = 1024}
-start = true
-task-slots = ["sys"]
-notifications = ["timer"]
-
 [tasks.meanwell]
 name = "drv-meanwell"
 features = ["stm32h7"]
-priority = 3
+priority = 6
```
labbott, line 14: I'm guessing this got flagged by the priority checker?

Aaron-Hartwig:
I've renamed `app-original.toml` back to `app.toml` after breaking CI made me think "why did I change this in the first place?".

app/gimletlet/app.toml
```diff
@@ -72,13 +27,49 @@ start = true
...
 task-slots = []
 
-[tasks.user_leds]
-name = "drv-user-leds"
-features = ["stm32h7"]
+[caboose]
+tasks = ["control_plane_agent", "caboose_reader"]
+region = "flash"
+size = 256
+default = true
```
labbott, line 34: This should go in the common TOML since having a caboose on all images is nice even if it rarely gets updated

app/gimletlet/app-meanwell.toml
Aaron-Hartwig: Yup. Since it wants to be able to call into `user_leds` and that task lives in the base TOML with a priority of 5, I adjusted it to 6 here.

labbott (approved):
LGTM! Thanks for doing this.

PR #1577, opened 2023-12-11
stm32h7: merge GPIO MPU regions.

cbiffle:
We were being careful and tiling three MPU regions to the exact boundaries of the implemented GPIO blocks on STM32H7. However, this also meant we were out of MPU slots. I need to map a region into sys for the external interrupt work.

Here I've merged the regions into a single, slightly larger region. This technically exposes some undefined address space to the sys task. However, it _sure looks_ like the remaining space is reserved by ST for the addition of more GPIO blocks, since they've been stuffing more ports in here in past models. Since we won't be touching that address space -- and we can be fairly certain of that thanks to mostly-safe code -- I'd argue that this doesn't increase our risk.


labbott (approved):
LGTM!

PR #1582, opened 2023-12-18
record sensor min/max values and times

bcantrill:
Draftt comment:

```
Expand the sensor task to record the minimum and maximum values seen
(as well as the times at which they were seen) in addition to the most
recent value.

This also contains the fix for #1592.

```


drv/i2c-devices/src/max31790.rs
```diff
@@ -311,7 +320,18 @@ impl Max31790 {
...
+        if count == 0 {
+            //
+            // We don't really expect this:  generally, if a fan is off (or is
+            // otherwise emiting non-detectable tach input pulses), the
+            // controller will report the power-on-reset value for the tach
+            // count, not 0.  So if we see a zero count, we will assume that
+            // this is an error rather than a 0 RPM reading, and record it to
+            // a (small) ring buffer and return accordingly.
+            //
+            ringbuf_entry!(Trace::ZeroTach(fan));
```
mkeeter, line 332: This ringbuf isn't strictly necessary, because the error will bubble up and be logged in the thermal ringbuf as `FanReadFailed`, but it's small enough that I'm not concerned.

task/sensor/src/main.rs
```diff
@@ -251,6 +317,10 @@ fn main() -> ! {
         last_reading: SensorArray(last_reading),
         data_value: SensorArray(data_value),
         data_time: SensorArray(data_time),
+        min_value: SensorArray(min_value),
+        min_time: SensorArray(min_time),
+        max_value: SensorArray(max_value),
+        max_time: SensorArray(max_time),
```
mkeeter, line 323: This is getting silly enough that a macro may be worthwhile:
```rust
    macro_rules! sensor_array {
        ($($name:ident: $t:ty = $n:expr;)*) => {
            paste::paste! {
                let ($($name),*) = mutable_statics::mutable_statics! {
                    $(
                    static mut [<$name:upper>]: [$t; NUM_SENSORS] = [|| $n; _];
                    )*
                };
                let ($($name),*) = ($(SensorArray($name)),*);
            }
        };
    }
    sensor_array!(
        last_reading: Option<LastReading> = None;
        data_value: f32 = f32::NAN;
        data_time: u64 = 0u64;
        min_value: f32 = f32::MAX;
        min_time: u64 = 0u64;
        max_value: f32 = f32::MIN;
        max_time: u64 = 0u64;
        err_value: NoData = NoData::DeviceUnavailable;
        err_time: u64 = 0;
        nerrors: u32 = 0;
    );

    let mut server = ServerImpl {
        last_reading,
        data_value,
        data_time,
        min_value,
        min_time,
        max_value,
        max_time,
        err_value,
        err_time,
        nerrors,
        deadline,
    };
```
(this requires adding `paste.workspace = true` to the `sensors/Cargo.toml`)

I haven't tested it, but it compiles fine and has the exact same size in flash / RAM.

mkeeter (approved):

task/sensor/src/main.rs
mkeeter: Or even
```rust
    macro_rules! declare_server {
        ($($name:ident: $t:ty = $n:expr;)*) => {{
            paste::paste! {
                let ($($name),*) = mutable_statics::mutable_statics! {
                    $(
                    static mut [<$name:upper>]: [$t; NUM_SENSORS] = [|| $n; _];
                    )*
                };
                let ($($name),*) = ($(SensorArray($name)),*);
                ServerImpl {
                    deadline,
                    $($name),*
                }
            }}
        };
    }
    let mut server = declare_server!(
        last_reading: Option<LastReading> = None;
        data_value: f32 = f32::NAN;
        data_time: u64 = 0u64;
        min_value: f32 = f32::MAX;
        min_time: u64 = 0u64;
        max_value: f32 = f32::MIN;
        max_time: u64 = 0u64;
        err_value: NoData = NoData::DeviceUnavailable;
        err_time: u64 = 0;
        nerrors: u32 = 0;
    );
```
(which removes one more layer of duplication, at the cost of some clarity)

drv/i2c-devices/src/max31790.rs
bcantrill: Agreed -- part of the rationale there was to get these failures into a ringbuf that won't roll over so we can essentially know if this happened at any time in the past.

task/sensor/src/main.rs
bcantrill: That looks great! I have verified that it works; pulling it in.

PR #1583, opened 2023-12-21
Update salty to version 0.3

flihp:
Upstream took our patch adding zeroize support. This was released in version 0.3: https://github.com/ycrypto/salty/pull/26. Updating this package required updating the sprockets-rot package as well. This should remove the need for our fork of the salty repo.


labbott (approved):
Yay for fewer forks

PR #1584, opened 2024-01-03
attest: move types in attest task IDL / API into external crate

flihp:
This commit moves the types used in the attest IDL / API into a library crate in the dice-util repo. This enables consumers outside of hubris to use the API.


flihp:
the example verifier is here: https://github.com/oxidecomputer/dice-util/tree/main/verifier-cli

labbott (approved):

PR #1587, opened 2024-01-09
hiffy:: Add `SendLeaseReadWrite` idol function.

flihp:
This function is called by `humility hiffy` to call hif operations that require a read and a write lease.

### DIFF
--- a/task/hiffy/src/common.rs
+++ b/task/hiffy/src/common.rs
@@ -419,6 +419,158 @@ pub(crate) fn send_lease_write(
     Ok(nreply + nlease)
 }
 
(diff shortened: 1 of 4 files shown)


flihp:
I'll post a link to the matching PR in `humility` where this new `hif` function is called when it's up.

flihp:
looks like adding this function to the `send` feature on the g0 / donglet rolled over 8k flash :(

flihp:
https://github.com/oxidecomputer/humility/pull/438

labbott (approved):
LGTM

PR #1589, opened 2024-01-15
Fix comment typos in stm32 sources

gemesa:
Fixed some typos I found while browsing through the STM32 source files. Very cool project btw!


labbott (approved):
Thanks!

PR #1591, opened 2024-01-17
attest: Add `attest_len` & `attest` ops.

flihp:
Calling the `attest` op will cause the `Attest` task to produce the signature: `sign(alias_priv, sha3_256(hubpack(log) | nonce))`. The caller can then use other ops from the `Attest` task to verify the signature.


flihp:
converted to DRAFT till I test on a gimlet

task/attest/src/main.rs
```diff
@@ -268,6 +284,102 @@ impl idl::InOrderAttestImpl for AttestServer {
...
+        &mut self,
+        _: &userlib::RecvMessage,
+        nonce: Leased<R, [u8]>,
+        dest: Leased<W, [u8]>,
+    ) -> Result<(), RequestError<AttestError>> {
+        if nonce.len() != Self::NONCE_SIZE {
+            let err = AttestError::BadLease;
+            ringbuf_entry!(Trace::AttestError(err));
+            return Err(err.into());
+        }
```
labbott, line 298: If this is a fixed size it might be worth enforcing this via `LenLimit` (which is an upper bound but might catch some errors at build time)

app/oxide-rot-1/app.toml
```diff
@@ -137,8 +137,8 @@ task-slots = ["swd"]
 [tasks.attest]
 name = "task-attest"
 priority = 5
-max-sizes = {flash = 14800, ram = 16384}
-stacksize = 11264
+max-sizes = {flash = 33904, ram = 16384}
```
labbott, line 140: I'm guesssing this huge size bump is from pulling in the signing?

task/attest/src/main.rs
flihp: I'm thinking we'll probably want to be able to support larger nonces in the future. The current min / max are both 32 bytes but w/ `LenLimit` I can probably set a range. Not sure what a good max value would be (hard to predict the future). The `sprot` plumbing can tolerate up to a 512 byte lease in both directions so that's not the limiting factor. The impact of allocating memory to hold the nonce in the attest task is likely to be the most significant trade off. A range like `32..128` is probably "future proof" enough and won't push memory usage too far. Will work something up before flipping the `draft` label off.

flihp: I've updated this PR such that the `attest` idol op in both `idl/attest.idol` and `idl/sprot.idol` have an upper bound of `128` bytes. The lower bound of `32` bytes is enforced with a conditional in the relevant functions. There's really no reason we couldn't accept a `nonce` that's `512` bytes. But there's little reason to use a nonce larger than `32` bytes so we impose the `128` byte upper bound to keep memory usage down. In the future, increasing this size would have a minimal impact. The important thing is to prevent the caller from using a nonce that's less than `32` bytes.

flihp:
this has been tested with dev & mfg builds with platform identity cert signed by staging root https://github.com/oxidecomputer/evidence-room/blob/main/staging/2023-05-12_provisioning/output/platform-identity-root-a.cert.pem on a PSC rev-b & Gimlet rev-c including `sprot`. My test is the current verification process from the `verifier-cli` in https://github.com/oxidecomputer/dice-util/tree/main/verifier-cli

labbott (approved):
LGTM

Issue #1594, opened 2024-01-22
Gimlet: support `rev-f` PCBAs


Aaron-Hartwig:
We made some manufacturability related changes to the Gimlet PCB/PCBA. These changes have no SP impact, but the board version resistors were changed anyway. This is understandable given the board and decision to change the resistors predated the publishing of RFD 438 where we locked down the function of these resistors.

So basically we have rev-e and rev-f PCBAs but they can use the same Hubris image. I'm going to rename the image to indicate that it works for something like `rev-ef` and update the boot check to accept either resistor code.

Aaron-Hartwig:
I'm tempted to make a `rev-def` since the contents of the `rev-e.toml` just inherit `rev-d.toml` anyway. https://github.com/oxidecomputer/hubris/blob/master/app/gimlet/rev-e.toml

Aaron-Hartwig:
the E/F boards have a SP_TO_IGNIT_FAULT_L signal that the D board does not, so I will not be combining D as well. https://github.com/oxidecomputer/hardware-gimlet/issues/1893

Aaron-Hartwig:
Per Hubris huddle on 22-Jan-2024 we **cannot** combine E/F right now as the update system does not support changing the board name of a system. So we could not go from `gimlet-e` to `gimlet-ef`. That is a battle for another day, so the immediate path here is to make a `gimlet-f`.

Aaron-Hartwig:
Related to https://github.com/oxidecomputer/hubris/issues/1595

Issue #1595, opened 2024-01-22
We need to be able to change board names on update


cbiffle:
Currently, `update-server` won't accept an update if the board name in the caboose doesn't match the board name of the current image. This is generally good and right.

However, it's becoming apparent that this might need to change. The Gimlet E/F revs are not different to firmware, so we'd sure like to make a combined `gimlet-ef` image that covers both -- but the `gimlet-e` boards currently deployed wouldn't accept such a thing. For now, we're going to make separate E and F images.

One possible approach to this: include multiple `BORD` chunks in the caboose of the firmware and accept the image if _any one_ names the current board. That isn't future-proof though, where if we make a mostly-identical G rev for design-for-manufacturing reasons, it should be able to run the existing E/F firmware... but its name won't be in the caboose yet.

So, maybe we need something more nuanced.

@Aaron-Hartwig @labbott @mkeeter @jgallagher

cbiffle:
Another option would be to transition the `BORD` over to naming revisions that actually rewire stuff or otherwise change firmware compatibility, and list hardware revision somewhere else... where future Gimlet revs will likely not change the firmware-oriented revision but will bump hardware revision.

mkeeter:
Transitioning `BORD` to mean "Hubris Compatibility Version" [(RFD 438)](https://rfd.shared.oxide.computer/rfd/438) seems sensible to me.  It looks like the RFD doesn't specify a format, but the hypothetical `gimlet-ef` makes sense and matches what we do elsewhere in Hubris.

How about releasing a trampoline update that looks for a `CMPT` array in the _new_ image, and allows an upgrade if `BORD` is contained in `CMPT` (even if `BORD` doesn't match between new and old)?

Here's what that would look like:

| `BORD` | `CMPT` | Notes |
|-|-|-|
| `gimlet-e` | (not present, not used) | Current state |
| `gimlet-e` | (not present, checked if `BORD` doesn't match) | Upgrade trampoline |
| `gimlet-ef` | [`gimlet-e`, `gimlet-f`] | New steady state |

(we could remove `CMPT` later, or leave it be)

cbiffle:
I like the `CMPT` idea, it's sort of similar to the DeviceTree compatibility family notion I alluded to in the huddle. This would give an image a way of stating that it is _deliberately_ trying to replace a set of other images and potentially change the `BORD` name as a result.

cbiffle:
[...]

PR #1596, opened 2024-01-23
Adds gimlet-f

Aaron-Hartwig:
Fixes https://github.com/oxidecomputer/hubris/issues/1594

We may not want to merge this PR if https://github.com/oxidecomputer/hubris/issues/1595 will be immediately addressed (before the end of January).


drv/gimlet-hf-server/src/main.rs
```diff
@@ -16,6 +16,7 @@
         target_board = "gimlet-c",
         target_board = "gimlet-d",
         target_board = "gimlet-e",
+        target_board = "gimlet-f",
     ),
     path = "bsp/gimlet_bcde.rs"
```
mkeeter, line 21: Want to rename this to `gimlet_bcdef.rs`?

Aaron-Hartwig: Agh, sorry!

mkeeter (approved):

PR #1597, opened 2024-01-25
gimlet-inspector: first draft.

cbiffle:


task/gimlet-inspector/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+    let net = Net::from(NET.get_task_id());
+    let seq = Sequencer::from(SEQ.get_task_id());
+
+    const SOCKET: SocketName = SocketName::inspector;
+
+    loop {
+        // These buffers are currently kept kinda small because our protocol
+        // messages are small.
+        let mut rx_data_buf = [0u8; Request::MAX_SIZE + MAX_REQ_TRAILER_SIZE];
+        let mut tx_data_buf = [0u8; 128];
```
hawkw, line 38: is there a maximum size constant for the response message type, the way there is for the request?

task/gimlet-inspector/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+                        SOCKET,
+                        meta,
+                        &tx_data_buf[0..(meta.size as usize)],
+                    ) {
+                        Ok(()) => break,
+                        // If `net` just restarted, immediately retry our send.
+                        Err(SendError::ServerRestarted) => continue,
+                        // If our tx queue is full, wait for space. This is the
+                        // same notification we get for incoming packets, so we
+                        // might spuriously wake up due to an incoming packet
```
hawkw, line 95: for my own education: what happens to the incoming packet in this case?

task/gimlet-inspector/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+                            sys_recv_closed(
+                                &mut [],
+                                notifications::SOCKET_MASK,
+                                TaskId::KERNEL,
+                            )
+                            .unwrap_lite();
+                        }
+                        // These errors should be impossible if we're configured
+                        // correctly.
+                        Err(SendError::NotYours | SendError::InvalidVLan) => {
```
hawkw, line 111: out of curiosity: in these cases where we panic due to one of multiple unhandled errors, are we intentionally not indicating which it is in the panic to avoid putting more strings in the binary, or something? naively, it seems like it would be nice to know which of the two “things that shouldn’t happen” happened, on the off chance one of them does happen.

hawkw (commented):
overall, this is very straightforward, but i left questions on a couple things i was curious about.

task/gimlet-inspector/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+};
+use hubpack::SerializedSize;
+use task_net_api::*;
+use userlib::*;
+
+task_slot!(NET, net);
+task_slot!(SEQ, seq);
+
+// Increase this if a new message requires a larger trailer.
+const MAX_REQ_TRAILER_SIZE: usize = 0;
```
mkeeter, line 24: Should this be in `gimlet_inspector_protocol` instead, since that's the canonical definition of requests?

task/gimlet-inspector/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+        let mut tx_data_buf = [0u8; 128];
+
+        match net.recv_packet(
+            SOCKET,
+            LargePayloadBehavior::Discard,
+            &mut rx_data_buf,
+        ) {
+            Ok(mut meta) => {
+                let Ok((request, _trailer)) = hubpack::deserialize::<Request>(&rx_data_buf) else {
+                    // We ignore malformatted, truncated, etc. packets.
```
mkeeter, line 47: It would be helpful to log into a ringbuf here, since we can read those remotely even on release builds using `dump-agent`

task/gimlet-inspector/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+                                TaskId::KERNEL,
+                            )
+                            .unwrap_lite();
+                        }
+                        // These errors should be impossible if we're configured
+                        // correctly.
+                        Err(SendError::NotYours | SendError::InvalidVLan) => {
+                            unreachable!()
+                        }
+                        // Unclear under what conditions we could se `Other` -
```
mkeeter, line 114:
```suggestion
                        // Unclear under what conditions we could see `Other` -
```

task/gimlet-inspector/src/main.rs
cbiffle: There is, though the response size will be dominated by the trailer. I should extract a constant and document that.

task/gimlet-inspector/src/main.rs
cbiffle: Sure.

task/gimlet-inspector/src/main.rs
cbiffle: I was really trying to keep ringbuf out of this task to keep the binary size and number of panic sites down. What about a counter? Counters can be read via dump agent too.

task/gimlet-inspector/src/main.rs
cbiffle: It hangs out in our RX queue until popped.

I actually just had to go read the netstack code to answer the anticipated follow-on question: so what determines when we get the next RX notification? The answer is, it depends on event flow in the netstack, and we are probably currently spamming any tasks that get into this state with RX notifications. I will file a bug.

task/gimlet-inspector/src/main.rs
cbiffle: > are we intentionally not indicating which it is in the panic to avoid putting more strings in the binary, or something?

Exactly. We get the line number, which I'm hoping will suffice in the event that this happens. (What we really _ought_ to do is divide up the error type to just the variants that can happen for a given operation.)

task/gimlet-inspector/src/main.rs
cbiffle: Amusingly this comment was copied verbatim from udprpc, which contains the same typo. :-) Fixing.

task/gimlet-inspector/src/main.rs
mkeeter: Sure, a counter sounds good to me.

cbiffle: Counters added, PTAL.

task/gimlet-inspector/src/main.rs
cbiffle: Factored out and rejiggered.

task/gimlet-inspector/src/main.rs
cbiffle: Constant extracted.

mkeeter (approved):

task/gimlet-inspector/src/main.rs
hawkw: oh, interesting. seems like a good catch!

task/gimlet-inspector/src/main.rs
hawkw: yeah, that makes sense --- refining the error types so that each operation returns an enum of _just_ the variants that it can actually return seems like a good idea! i could, potentially, pick that up, given some pointers?

PR #1599, opened 2024-01-26
Multi region packing

mkeeter:
This PR modifies the build system to let flash use all available MPU regions, resulting in tighter packing and smaller total image size.  This shrinks the `sidecar-rev-c` image by 256 KiB, and the `gimlet-f` image by 188 KiB.

In other words, tasks go from having exactly 1 flash region to _at least_ 1 flash region; this is mostly plumbing that change from `suggest_memory_region_size` all the way through the `kconfig`.

The change does makes task packing trickier, because tasks can be placed in one of two orientations: largest-chunk-first, or largest-chunk-last.  I went for a very dumb O(N^2) algorithm that checks every unplaced task and picks the best; we're far from having > 100 tasks, so I'm not worried about bad scaling (famous last words!).

In addition, it updates `xtask/src/sizes.rs` to print the individual chunks when the `-v` flag is specified.

I'm expecting CI to get mad, because some kernels may need more flash space to hold the extra regions.

## Before
before

## After
after

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -2021,7 +2021,12 @@ fn allocate_region(
             // We can always place the chunk using forward orientation, albeit
             // with padding if it's not aligned.
             let gap_forward = base - avail.start;
-            best.update(gap_forward, align, task_name, Direction::Forward);
+            best.update(
+                gap_forward,
+                *mem.last().unwrap(),
+                task_name,
+                Direction::Forward,
+            );
         }
         let Some(sizes) = t_reqs.remove(best.name) else {
             panic!("could not find a task");
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -295,7 +295,6 @@ pub fn package(
                 .as_ref()
                 .map(|c| c.tasks.contains(&t.to_string()))
                 .unwrap_or(false) as usize;
-        println!("task {t} has {} spare regions", 7 - n);
 
         task_reqs.insert(
             t,
@@ -1948,13 +1947,7 @@ fn allocate_region(
             .insert(region.to_string(), allocate_k(region, sz, avail)?);
     }
 
-    println!("PACKING!");
-    for (&task_name, mem) in t_reqs.iter() {
-        println!("  {task_name}, {mem:?}");
-    }
-
     while !t_reqs.is_empty() {
-        println!("avail: {avail:x?}");
         // At this point, we need to find a task that fits based on our existing
         // alignment.  This is tricky, because -- for efficient packing -- we
         // allow tasks to span multiple regions.  For example, a task could look
@@ -2060,9 +2053,6 @@ fn allocate_region(
                 }
             }
         }
-
-        println!("found best: {best:#x?}");
-        println!("{:x?}", allocs.tasks[best.name][region]);
 
         // Check that our allocations are all aligned and contiguous
         let ra = &allocs.tasks[best.name][region];
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -2066,12 +2066,11 @@ fn allocate_region(
         while let Some(mut size) = sizes.pop_front() {
             // When building the size list, we split the largest size to reduce
             // alignment requirements.  Now, we try to merge them again, to
-            // reduce the number of regions stored in the kernel's flash. 
+            // reduce the number of regions stored in the kernel's flash.
             //
             // For example, [256, 256, 64] => [512, 64] if the initial position
             // is aligned for a 512-byte region.
             let mut n = sizes.iter().filter(|s| **s == size).count() + 1;
-            println!("{size}, {sizes:?}");
             if n > 1 {
                 n &= !1; // only consider an even number of regions
                 let possible_align = toml.task_memory_alignment(size * 2);
@@ -2083,7 +2082,6 @@ fn allocate_region(
                     for _ in 0..n / 2 - 1 {
                         sizes.push_front(size);
                     }
-                    println!("   => {size}, {sizes:?}")
                 }
             }
 
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -283,9 +283,9 @@ impl OrderedVecDeque {
     fn push_front(&mut self, v: u32) {
         if let Some(f) = self.front() {
             if self.increasing {
-                assert!(v >= *f);
-            } else {
                 assert!(v <= *f);
+            } else {
+                assert!(v >= *f);
             }
         }
         self.data.push_front(v)
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use std::collections::{BTreeMap, BTreeSet, HashMap, VecDeque};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::ffi::OsStr;
 use std::fmt::Write as _;
 use std::fs::{self, File};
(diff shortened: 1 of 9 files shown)


build/xtask/src/sizes.rs
```diff
@@ -78,7 +79,10 @@ pub fn run(
         }
         let size = toml.kernel.requires[&mem.to_string()];
 
-        let suggestion = toml.suggest_memory_region_size("kernel", used);
+        let suggestion = toml.suggest_memory_region_size("kernel", used, 1);
+        assert_eq!(suggestion.len(), 1);
```
cbiffle, line 83: So imagine I'm here in the future and this assertion has failed. What should I look at? Asking because I don't totally understand it as a reviewer. (The answer would make a good assert message or comment.)

sys/kern/build.rs
```diff
@@ -292,7 +303,7 @@ fn translate_address(
     task_index: usize,
     address: OwnedAddress,
 ) -> u32 {
-    let key = RegionKey::Owned(task_index, address.region_name);
+    let key = RegionKey::Owned(task_index, 0, address.region_name);
```
cbiffle, line 306: Is this right? It seems simpler than I'd expect.

build/xtask/src/dist.rs
```diff
@@ -1311,8 +1338,8 @@ fn generate_task_linker_script(
 
     writeln!(linkscr, "MEMORY\n{{")?;
     for (name, range) in map {
```
cbiffle, line 1340: I think this'd be clearer if the name became `ranges`, since it's now more than one range.

build/xtask/src/dist.rs
```diff
@@ -1290,7 +1317,7 @@ fn check_task_priorities(toml: &Config) -> Result<()> {
 
 fn generate_task_linker_script(
     name: &str,
-    map: &BTreeMap<String, Range<u32>>,
+    map: &BTreeMap<String, Vec<Range<u32>>>,
```
cbiffle, line 1320: Does the code below assume this vec is sorted? I notice it accesses first and last.

build/xtask/src/dist.rs
```diff
@@ -1735,7 +1762,11 @@ pub struct Allocations {
     /// Map from memory-name to address-range
     pub kernel: BTreeMap<String, Range<u32>>,
     /// Map from task-name to memory-name to address-range
-    pub tasks: BTreeMap<String, BTreeMap<String, Range<u32>>>,
+    ///
+    /// A task may have multiple address ranges in the same memory space for
+    /// efficient packing; if this is the case, the addresses will be contiguous
+    /// and each individual range will respect MPU requirements.
+    pub tasks: BTreeMap<String, BTreeMap<String, Vec<Range<u32>>>>,
```
cbiffle, line 1769: (I think this has an implicit assumption about vec order as well, it should go in the comment. Or maybe a BTreeSet would be better than a Vec?)

build/xtask/src/dist.rs
```diff
@@ -1821,105 +1864,50 @@ pub fn allocate_all(
...
+                let n = if mem == "flash" {
+                    req.spare_regions + 1
+                } else {
+                    1
+                };
+                let bytes = toml.suggest_memory_region_size(name, amt, n);
                 if let Some(r) = tasks[name].max_sizes.get(&mem.to_string()) {
-                    if bytes > *r as u64 {
+                    let total_bytes = bytes.iter().sum::<u64>();
+                    if total_bytes > *r as u64 {
```
cbiffle, line 1883: nit: this looks like a lossy conversion but isn't, consider `u64::from(*r)`

build/xtask/src/dist.rs
```diff
@@ -2082,7 +2248,7 @@ pub fn make_kconfig(
     for (i, (name, task)) in toml.tasks.iter().enumerate() {
         let stacksize = task.stacksize.or(toml.stacksize).unwrap();
 
-        let flash = &task_allocations[name]["flash"];
+        let flash = &task_allocations[name]["flash"][0];
```
cbiffle, line 2251: I kinda feel like every case where we're explicitly accessing the first element of the list should get a comment, because each time I stop and go "uh... wait... is this right or making an assumption"

cbiffle (commented):
This generally looks good. There are a few areas where I think there's an unmentioned assumption about ordering where I'd like to see some docs (or a type that causes the assumption to be explicit).

cbiffle:
I asked @mkeeter for time measurements of the potentially N**2 algorithm here -- execution is currently under a millisecond in every case. So, not worth attempting to optimize the complexity, imo.

build/xtask/src/sizes.rs
mkeeter: Added a string to the assertion in 7d2d02b919f84713f1902d88ba41c865f5363f92; this detects a failure in `suggest_memory_region_size` where it has suggested > 1 region for the kernel, which should never happen.

build/xtask/src/dist.rs
mkeeter: Yup, done in 05e21fbee78db4408a70b4ddc01935e2061316c9

build/xtask/src/dist.rs
mkeeter: Good catch, done in 3e2421a76e45b021398152745b9c6ebb2e871d01

sys/kern/build.rs
mkeeter: Yes, although it's a little subtle:

- Regions are populated in order from the `MultiRegionConfig`, so region 0 is going to be the first region in memory (at an address given by `MultiRegionConfig::base`)
- Regions are contiguous, so the offset can be taken from the first region's base address

I added a comment to that effect in 122b0a127318bfa586542f1dfce0bd927eeff04f

build/xtask/src/dist.rs
mkeeter: Yup, I added a `ContinguousRanges` type which enforces that the ranges are contiguous (3d128cbee0fd75fab17799a754281d836d4a7321)

build/xtask/src/dist.rs
mkeeter: Same as below, fixed with `ContiguousRanges` (3d128cbee0fd75fab17799a754281d836d4a7321)

build/xtask/src/dist.rs
mkeeter: Yup, fixed with `ContiguousRanges` in 3d128cbee0fd75fab17799a754281d836d4a7321

build/xtask/src/dist.rs
```diff
@@ -210,6 +210,101 @@ pub fn list_tasks(app_toml: &Path) -> Result<()> {
...
+impl OrderedVecDeque {
+    /// Build a new `VecDeque` with a decreasing constraint
+    fn decreasing() -> Self {
+        Self {
+            data: VecDeque::new(),
+            increasing: false,
+        }
+    }
+    /// Flip the data and constraint
+    fn reverse(self) -> Self {
```
cbiffle, line 268: Ok so this is going to sound really pedantic.

Please call this something like `reversed`. Or, since it consumes self, maybe something starting with `into_`?

`reverse` in the standard library is generally an in-place operation (e.g. [on vec](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reverse)). This caused me to get hung up on how the diff on line 2062 could possibly be correct.

cbiffle (approved):

build/xtask/src/dist.rs
```diff
@@ -210,6 +210,101 @@ pub fn list_tasks(app_toml: &Path) -> Result<()> {
 /// Represents allocations and free spaces for a particular image
 type AllocationMap = (Allocations, IndexMap<String, Range<u32>>);
 
+/// Simple data structure to store a set of guaranteed-contiguous ranges
+///
+/// This will panic if you violate that constraint!
```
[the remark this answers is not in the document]
mkeeter: Eh, adding an inline module is easy, so I did it in 128247d6a06f527a042d11d4a5d0cb4ab3aad528

build/xtask/src/dist.rs
mkeeter: Sure, done in 128247d6a06f527a042d11d4a5d0cb4ab3aad528

PR #1600, opened 2024-01-27
gimlet_seq: separate error type for read_fpga_regs

hawkw:
Inspired by [this comment][1] from @cbiffle.

Currently, the Gimlet Sequencer API has one error enum that's returned
by all API methods. This enum contains a bunch of error variants that
are specific to actually setting the sequencer state, and a
`ReadRegsFailed` variant that's returned by the `read_fpga_regs` API
method. This means that, currently, calls to set the sequencer state
have to handle the `ReadRegsFailed` variant, even though it's never
actually returned by the code that sets the sequencer state, and,
conversely, code that reads the FPGA registers has to handle a bunch of
error variants that aren't related to FPGA register reads.

This commit separates the error types of the `read_fpga_regs` and
`set_state` APIs into two separate enums that only contain the errors
returned by that API. I didn't touch the other API methods, although (as
far as I can tell), `get_state`, `fans_on`, and `fans_off` never
actually return any errors and could be changed to
`core::convert::Infallible` a la `send_hardware_nmi`. I didn't want to
mess with these, though, based on the assumption that the `SeqError`
error type was chosen to avoid having to change the API if these methods
changed to return errors. Let me know if that's not the case --- I'm
happy to change them as well.

[1]: https://github.com/oxidecomputer/hubris/pull/1597#discussion_r1468014774


drv/gimlet-seq-api/src/lib.rs
```diff
@@ -28,6 +27,15 @@ pub enum SeqError {
     ServerRestarted,
 }
 
+/// Errors returned by `read_fpga_regs`.
+#[derive(Copy, Clone, Debug, FromPrimitive, Eq, PartialEq, IdolError)]
+pub enum ReadFpgaRegsError {
+    Failed = 1,
```
cbiffle, line 33: Maybe add a comment noting that this variant is _probably_ illusory because of the SPI error API issue, with a TODO to revisit it.

cbiffle (approved):

drv/gimlet-seq-api/src/lib.rs
hawkw: will do!

PR #1601, opened 2024-01-29
Bump sprot sizes to fix build

mkeeter:
#1597 bumped `syn` and `proc-macro2`.  For some reason, this slightly changes the size of the `sprot` task, so `master` is not building for me right now:

```console
Error: task sprot: needs 47360 bytes of flash but max-sizes limits it to 47328
```

[Last time I investigated](https://matrix.to/#/!jxnlDLnJsVaIRstnrt:oxide.computer/$pCYj2H5LWUDkQwx_svMii1GZiOkXxpHA26BOjJzXPBA?via=oxide.computer&via=unix.house&via=matrix.org), I came to the conclusion that bumping `proc-macro2` caused LLVM functions to be emitted in a different order, which changed inlining behavior, which pushed one task over the size limit (and there were literally no changes in the generated LLVM IR).

### DIFF
--- a/app/lpc55xpresso/app-sprot.toml
+++ b/app/lpc55xpresso/app-sprot.toml
@@ -183,7 +183,7 @@ task-slots = ["swd"]
 [tasks.sprot]
 name = "drv-lpc55-sprot-server"
 priority = 6
-max-sizes = {flash = 47328, ram = 32768}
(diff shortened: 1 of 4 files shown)


labbott (approved):
For completeness, this appears to happen only on MacOS which is why it passed CI.

PR #1609, opened 2024-02-01
Update bitflags to 2.x

cbiffle:
This brings us up to the most recent release, and has some nice properties, like there now being a trait that describes the bitflags API instead of it all being open-coded. (This is the property that motivated this change.)

There are some slightly annoying aspects, like the operations in that trait no longer being const (because we don't have const impls yet). So, this required some surgery.

### DIFF
--- a/lib/host-sp-messages/src/lib.rs
+++ b/lib/host-sp-messages/src/lib.rs
@@ -507,10 +507,38 @@ impl From<HubpackError> for DecodeFailureReason {
     }
 }
 
+#[derive(
+    Copy,
+    Clone,
+    Debug,
+    Eq,
+    PartialEq,
+    Serialize,
+    Deserialize,
+    SerializedSize,
+    FromBytes,
+    AsBytes,
+)]
+#[repr(transparent)]
+pub struct Status(u64);
+
+#[derive(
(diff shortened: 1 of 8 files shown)


lib/host-sp-messages/src/lib.rs
```diff
@@ -507,10 +507,38 @@ impl From<HubpackError> for DecodeFailureReason {
...
+    Eq,
+    PartialEq,
+    Serialize,
+    Deserialize,
+    SerializedSize,
+    FromBytes,
+    AsBytes,
+)]
+#[repr(transparent)]
+pub struct HostStartupOptions(u64);
```
labbott, line 538: Did these get moved out of the `bitflags` macro for the same reason as below or another reason?

mkeeter (approved):

lib/host-sp-messages/src/lib.rs
cbiffle: Any case where I pulled structs out of the bitflags macro, it's because they needed to `derive` a trait that doesn't automatically work on the generated types for structs defined _inside_ the 2.x macro. This was most of the annoying work.

Issue #1205, opened 2023-03-14
gimlet-seq should not unwrap I2C results.


cbiffle:
A misconfiguration of a Gimlet C rev FPGA today caused havoc on the board, but during that havoc, we noticed that gimlet-seq was crashlooping.

It was crashlooping because it was trying to reach out to the `Raa229618` to set the power rails to their desired initial off state, and was not able to because *waves hands at I2C*. Its response to this was to `unwrap()`.

`unwrap()` is basically always the wrong tool for an error that can be caused by exterior factors that we have no control over, like (I'd argue) I2C. In the case of gimlet-seq, it genuinely cannot make progress if it can't complete this step -- but because this isn't a program being run by a user on a Unix system, panicking with a message does not stop the process, it causes the process to repeat forever. (This is deliberate.)

`gimlet-seq` should not panic in cases like this. Instead, it should transition its state machine to a "failed" state (possibly after some retries), record why somewhere we can get to it, and (if possible) set the FAULT net to Ignition to indicate that the machine must be power cycled. (IIRC setting the FAULT net on Gimlet requires communication with the RoT, which may or may not be working in the event of an early sequencing failure. It _should be_ working, they're in the same power domain....)

hawkw:
Looking at the `gimlet_seq` service, it seems like we also `unwrap()` a bunch of SPI reads/writes --- do we want to add similar behavior there? IIUC, the SPI bus operations won't fail for transient reasons outside of the task's control, the way the I2C bus can. Instead, the SPI ops will fail primarily in the face of a programmer error such as trying to use a SPI address that doesn't exist. But, we can add similar behavior for the various SPI reads/writes as well, if that's desirable.

cbiffle:
Probably a lower priority, but not an awful idea in case someone introduces a failure case in the SPI API later. (SPI is probably a candidate for switching to `REPLY_FAULT` so that its callers don't have to check an error that shouldn't happen in practice.)

hawkw:
Okay, I put up #1611 as a first pass on this! It's very possible I've misunderstood some things, so please let me know if I'm barking up the wrong metaphorical tree here.

PR #1611, opened 2024-02-10
gimlet-seq: retry I2C errors, or die trying

hawkw:
Currently, the sequencer task uses `unwrap()` on I2C bus operations.
However, these may fail due to transient bus errors, or fail permanently
for reasons wholly outside the sequencer task's control. As discussed in
issue #1205, it's incorrect for the sequencer task to unwap these
errors, as panicking results in the task being restarted. On fatal bus
errors, this means we crash loop.

Instead, this commit changes the `gimlet-seq` task to retry failed I2C
reads/writes up to three times, and, if they still fail, transition to a
permanent failed state, setting the FAULT net to Ignition to indicate
that the Gimlet must be power-cycled.

### DIFF
--- a/drv/gimlet-seq-server/src/main.rs
+++ b/drv/gimlet-seq-server/src/main.rs
@@ -14,6 +14,7 @@ use userlib::*;
 
 use drv_gimlet_hf_api as hf_api;
 use drv_gimlet_seq_api::{PowerState, SeqError};
+use drv_i2c_api as i2c;
 use drv_ice40_spi_program as ice40;
 use drv_packrat_vpd_loader::{read_vpd_and_load_packrat, Packrat};
 use drv_spi_api::{SpiDevice, SpiServer};
@@ -103,7 +104,11 @@ enum Trace {
     SpdBankAbsent(u8),
     SpdAbsent(u8, u8, u8),
     SpdDimmsFound(usize),
-
+    I2cFault {
+        retries_remaining: u8,
+        code: i2c::ResponseCode,
+    },
+    StartFailed(SeqError),
     None,
 }
 
@@ -111,11 +116,43 @@ ringbuf!(Trace, 128, Trace::None);
 
 #[export_name = "main"]
 fn main() -> ! {
-    let spi = drv_spi_api::Spi::from(SPI.get_task_id());
     let sys = sys_api::Sys::from(SYS.get_task_id());
     let jefe = Jefe::from(JEFE.get_task_id());
+    let spi = drv_spi_api::Spi::from(SPI.get_task_id());
     let hf = hf_api::HostFlash::from(HF.get_task_id());
+    match init(&sys, &jefe, spi, hf) {
+        // Set up everything nicely, time to start serving incoming messages.
+        Ok(mut server) => {
+            let mut buffer = [0; idl::INCOMING_SIZE];
+            loop {
+                idol_runtime::dispatch(&mut buffer, &mut server);
+            }
+        }
 
+        // Initializing the sequencer failed.
+        Err(_) => {
+            // Tell everyone that something's broken, as loudly as possible.
+            ringbuf_entry!(Trace::StartFailed(SeqError::I2cFault));
+            sys.gpio_set(FAULT_PIN_L);
+
+            // All these moments will be lost in time, like tears in rain...
+            // Time to die.
+            loop {
+                hl::sleep_for(100);
+            }
+        }
+    }
+}
+
+fn init<S>(
+    sys: &sys_api::Sys,
+    jefe: &Jefe,
+    spi: S,
+    hf: hf_api::HostFlash,
+) -> Result<ServerImpl<S>, i2c::ResponseCode>
+where
+    S: SpiServer + Clone,
+{
     // Ensure the SP fault pin is configured as an open-drain output, and pull
     // it low to make the sequencer restart externally visible.
     sys.gpio_configure_output(
@@ -175,7 +212,7 @@ fn main() -> ! {
     // the SPI and CS lines are separately managed by the SPI server; the ice40
     // crate handles the CRESETB and CDONE signals, and takes care not to
     // generate surprise resets.
-    ice40::configure_pins(&sys, &ICE40_CONFIG);
+    ice40::configure_pins(sys, &ICE40_CONFIG);
 
     let pg = sys.gpio_read_input(PGS_PORT);
     let v1p2 = pg & PG_V1P2_MASK != 0;
@@ -294,7 +331,7 @@ fn main() -> ! {
         loop {
             let prog = spi.device(drv_spi_api::devices::ICE40);
             ringbuf_entry!(Trace::Programming);
-            match reprogram_fpga(&prog, &sys, &ICE40_CONFIG) {
+            match reprogram_fpga(&prog, sys, &ICE40_CONFIG) {
                 Ok(()) => {
                     // yay
                     break;
@@ -324,7 +361,8 @@ fn main() -> ! {
 
     ringbuf_entry!(Trace::Programmed);
 
-    vcore_soc_off();
+    vcore_soc_off()?;
+
     ringbuf_entry!(Trace::RailsOff);
 
     let ident = seq.read_ident().unwrap_lite();
@@ -360,8 +398,7 @@ fn main() -> ! {
             ringbuf_entry!(Trace::ClockConfigWrite);
             Ok(())
         }
-    })
-    .unwrap_lite();
+    })?;
 
     // Populate packrat with our mac address and identity.
     let packrat = Packrat::from(PACKRAT.get_task_id());
@@ -380,17 +417,16 @@ fn main() -> ! {
     // Per JEDEC 1791.12a, we must wait for tINIT (10ms) between power on and
     // sending the first SPD command.
     hl::sleep_for(10);
-    read_spd_data_and_load_packrat(&packrat, I2C.get_task_id());
+    read_spd_data_and_load_packrat(&packrat, I2C.get_task_id())?;
 
     // Turn on the chassis LED once we reach A2
     sys.gpio_set(CHASSIS_LED);
 
-    let mut buffer = [0; idl::INCOMING_SIZE];
     let mut server = ServerImpl {
         state: PowerState::A2,
         sys: sys.clone(),
         seq,
-        jefe,
+        jefe: jefe.clone(),
         hf,
         deadline: 0,
     };
@@ -418,12 +454,13 @@ fn main() -> ! {
     // and fewer things can go wrong.
     sys.gpio_set(FAULT_PIN_L);
 
-    loop {
-        idol_runtime::dispatch_n(&mut buffer, &mut server);
-    }
+    Ok(server)
 }
 
-fn read_spd_data_and_load_packrat(packrat: &Packrat, i2c_task: TaskId) {
+fn read_spd_data_and_load_packrat(
+    packrat: &Packrat,
+    i2c_task: TaskId,
+) -> Result<(), i2c::ResponseCode> {
     use drv_gimlet_seq_api::NUM_SPD_BANKS;
     use drv_i2c_api::{Controller, I2cDevice, Mux, PortIndex, Segment};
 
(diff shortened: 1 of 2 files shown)


drv/gimlet-seq-server/src/main.rs
```diff
@@ -103,451 +104,381 @@ enum Trace {
...
+            // Tell everyone that something's broken, as loudly as possible.
+            ringbuf_entry!(Trace::StartFailed(SeqError::I2cFault));
+            sys.gpio_set(FAULT_PIN_L);
+            jefe.set_state(PowerState::InitFailed as u32);
+
+            // All these moments will be lost in time, like tears in rain...
+            // Time to die.
+            loop {
+                hl::sleep_for(100);
+            }
```
hawkw, line 142: I wasn't sure what the most Good and Proper way for a Hubris task to "die permanently" is --- is there something better to do here?

drv/gimlet-seq-server/src/main.rs
hawkw, line 1: N.B. that the git diff is super messy just because I moved a big chunk of code out of `main` and into a function; it's mostly unchanged, but GitHub doesn't really seem to get that. All the functional changes are in commit ee898f021958cd6221f78e890c8233f170b5d1b7, which might be a bit easier to review.

drv/gimlet-seq-server/src/main.rs
labbott: Waiting for a notification that will never come would be nicer as this should make sure the task never becomes ready to run again (e.g. https://github.com/oxidecomputer/hubris/blob/master/task/sp_measure/src/main.rs#L71-L75)

hawkw: Lovely, thank you!

rmustacc:
A general question I have from an upstack consumer perspective:

> Instead, this commit changes the gimlet-seq task to retry failed I2C
> reads/writes up to three times, and, if they still fail, transition to a
> permanent failed state, setting the FAULT net to Ignition to indicate
> that the Gimlet must be power-cycled.

My understanding is that https://github.com/oxidecomputer/hubris/pull/1556 has caused the visible state of the fault signal to be asserted the entire time that we are booting up. So I'm not sure how this intersects with this. In particular, given the change in that PR, we cannot actually know that we need to power cycle that Gimlet without creating some sort of startup time-out countdown timer. That is, upstack we'd have to know that this has remained asserted for some time x to believe that this is valid versus the original theory of being able to say whenever it's asserted we know it's a problem.

If the logic in 1556 happens before this, how do we distinguish the two? If not, do they intersect?

hawkw:
@rmustacc If I'm being entirely honest, that seems like more of a question for @cbiffle &mdash; I just picked up the issue as he described it in #1205. Cliff probably has a better sense of how this is expected to interact with the rest of the system.

With that said, I *believe* the visible behavior on failures should be similar? Currently, the fault pin is pulled low during the startup process, and only set high once the sequencer task is actually serving IPC requests. If the sequencer task crashes and enters a crashloop before it starts serving IPC requests, the fault pin remains asserted. Now, if a failure occurs during startup, the sequencer task no longer crash loops, but it still won't set the fault pin high, so the fault pin still remains asserted indefinitely if the sequencer task can't come up.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -103,19 +104,56 @@ enum Trace {
...
+        Err(_) => {
+            // Tell everyone that something's broken, as loudly as possible.
+            ringbuf_entry!(Trace::StartFailed(SeqError::I2cFault));
+            sys.gpio_set(FAULT_PIN_L);
+            jefe.set_state(PowerState::InitFailed as u32);
+
+            // All these moments will be lost in time, like tears in rain...
+            // Time to die.
+            loop {
+                hl::sleep_for(100);
```
cbiffle: The IPC server probably assumes that initialization was successful, so it's unsafe to start it here... that's too bad because it'd be really great to be able to ask it questions.

We'd have to introduce some notion of "I'm broken" into the IPC server for that to work.

Mostly just thinking out loud here.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -103,19 +104,56 @@ enum Trace {
...
 
+        // Initializing the sequencer failed.
+        Err(_) => {
+            // Tell everyone that something's broken, as loudly as possible.
+            ringbuf_entry!(Trace::StartFailed(SeqError::I2cFault));
+            sys.gpio_set(FAULT_PIN_L);
+            jefe.set_state(PowerState::InitFailed as u32);
+
+            // All these moments will be lost in time, like tears in rain...
+            // Time to die.
```
cbiffle, line 139: Nice pull.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -633,6 +680,34 @@ impl<S: SpiServer> NotificationHandler for ServerImpl<S> {
     }
 }
 
+fn retry_i2c_txn<T, E>(
+    mut txn: impl FnMut() -> Result<T, E>,
+) -> Result<T, i2c::ResponseCode>
+where
+    i2c::ResponseCode: From<E>,
+{
+    // Chosen by fair dice roll, seems reasonable-ish?
```
cbiffle, line 563: (agreed)

drv/gimlet-seq-server/src/main.rs
```diff
@@ -633,6 +680,34 @@ impl<S: SpiServer> NotificationHandler for ServerImpl<S> {
...
+    let mut retries_remaining = 3;
+    loop {
```
cbiffle, line 565: I might consider structuring this as a `for _ in 0..retries_remaining` just to avoid accidentally creating an infinite loop through a logic error below.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -134,7 +134,6 @@ fn main() -> ! {
             // Tell everyone that something's broken, as loudly as possible.
             ringbuf_entry!(Trace::StartFailed(SeqError::I2cFault));
             sys.gpio_set(FAULT_PIN_L);
-            jefe.set_state(PowerState::InitFailed as u32);
```
cbiffle: Note that with this change, you could pass jefe into the init function by-value. (I think this is the only use after the call to init.)

drv/gimlet-seq-server/src/main.rs
```diff
@@ -138,7 +138,9 @@ fn main() -> ! {
             // All these moments will be lost in time, like tears in rain...
             // Time to die.
             loop {
-                hl::sleep_for(100);
+                if sys_recv_closed(&mut [], 1, TaskId::KERNEL).is_err() {
```
cbiffle, line 141: This looks like it's the timer notification -- what's your intent here?

cbiffle (commented):
I think moving init to ServerImpl has more justification than you suggest in your commit message -- you've basically done a "parse don't validate," since ServerImpl is written with a lot of embedded assumptions that init succeeded. So it's now harder to accidentally start a server without finishing init.

Generally good, a few questions below:

cbiffle:
> My understanding is that #1556 has caused the visible state of the fault signal to be asserted the entire time that we are booting up. So I'm not sure how this intersects with this. In particular, given the change in that PR, we cannot actually know that we need to power cycle that Gimlet without creating some sort of startup time-out countdown timer. That is, upstack we'd have to know that this has remained asserted for some time x to believe that this is valid versus the original theory of being able to say whenever it's asserted we know it's a problem.
> 
> If the logic in 1556 happens before this, how do we distinguish the two? If not, do they intersect?

Until we're confident that potential crashloops are all gone from the sequencer task, I'm hesitant to remove the code that sets fault on startup. So we probably need to work out a "how much time does the sequencer have to start up before we consider it faulted" delay either way until we're more confident in the code.

This change fixes most (or all, possibly) of the I2C-related panics, but not all the panics.

drv/gimlet-seq-server/src/main.rs
hawkw: Yeah, I was also thinking about possibly having it start serving, and just respond to everything with an "I'm broken" message. Would be happy to change that now, if you think that's a good idea, or in a later PR?

rmustacc:
> > My understanding is that #1556 has caused the visible state of the fault signal to be asserted the entire time that we are booting up. So I'm not sure how this intersects with this. In particular, given the change in that PR, we cannot actually know that we need to power cycle that Gimlet without creating some sort of startup time-out countdown timer. That is, upstack we'd have to know that this has remained asserted for some time x to believe that this is valid versus the original theory of being able to say whenever it's asserted we know it's a problem.
> > If the logic in 1556 happens before this, how do we distinguish the two? If not, do they intersect?
> 
> @rmustacc - Until we're confident that potential crashloops are all gone from the sequencer task, I'm hesitant to remove the code that sets fault on startup. So we probably need to work out a "how much time does the sequencer have to start up before we consider it faulted" delay either way until we're more confident in the code.

That's fine. Just want to make sure that we can actually use this up stack and then we'll have to figure out how that fits into the ignition controller and related (or the future looking interface overhaul you have planned). I realize the upstack stuff isn't there yet so I don't ask this to block this at all, just to make sure we know how it all would fit together.

drv/gimlet-seq-server/src/main.rs
hawkw: that was what i was going to do. but, we want to return the last error when we run out of retries, which means that we need to keep it around when we fall through to the end of the loop, which felt a bit ugly; it would have to be something like this:
```rust
// Chosen arbitrarily, but don't worry! We'll never actually return this --- it's 
// just because we need to initialize this to *something*.
let mut last_error = i2c::ResponseCode::BadResponse;
for retries_remaining in (0..3).rev() {
        match txn() {
            Ok(x) => return Ok(x),
            Err(e) => {
                let code = e.into();
                ringbuf_entry!(Trace::I2cFault {
                    retries_remaining,
                    code,
                });
                last_error = code;
            }
}

Err(code)
```

in particular, i felt like having to assign `last_error` to _something_, even though we've never actually seen that error, felt a bit unfortunate.

what do you think?

cbiffle: Hmmmm. Fair point. Current structure's fine.

cbiffle:
@rmustacc - understood. The main improvement of this change is that it'll stop frantically reinitializing the FPGA and bouncing rails when I2C messes up; improved upstream diagnostics are kind of a side effect. There's more we can do to improve the diagnostics beyond this change.

drv/gimlet-seq-server/src/main.rs
cbiffle: We talked about this in chat -- the intent was to pick a notification that will never arrive. Unfortunately there are no such notifications, because _in theory_ any notification can be delivered using `sys_post`. So, this should probably just loop if it gets an errant notification, as we discussed.

hawkw: Yup, addressed Cliff's suggestions from our discussion in cbdbfc44018d9ce98c9e983d199966940e72cedd :)

drv/gimlet-seq-server/src/main.rs
hawkw: A third option, which is _also_ not my favorite, is to do something like this:
```rust
    for retries_remaining in (0..MAX_RETRIES).rev() {
        match txn() {
            Ok(x) => return Ok(x),
            Err(error) => {
                let code = error.into();
                ringbuf_entry!(Trace::I2cFault {
                    retries_remaining,
                    code,
                });
                if retries_remaining == 0 {
                    return Err(code);
                }
            }
        }
    }

    unreachable!()
```

Now, we've gotten rid of the arbitrary error code that we'll never actually return, *and* the infinite loop that isn't actually infinite, but this comes at the cost of an `unreachable!()`, and having to write `unreachable!()` also makes me feel kind of sad. 

If you have a preference for any of these bad options, I'll happily go with that one --- I don't really have strong feelings about which is the least objectionable. :)

cbiffle (approved):
(please squash or fold your most recent commits)
(or all of them, I'm not the cops)

Issue #1613, opened 2024-02-14
gimlet went back to A0 while wicket was updating its phase1 flash


jgallagher:
When @augustuswm was mupdating rack3 today, he saw this failure on `BRM42220022`:

![uULOhYYQDDirFtYWKCZHYMLy](https://github.com/oxidecomputer/hubris/assets/1435635/64c4481a-5284-4eb3-b171-9b941f2bd526)

The underlying error the SP reported is:

```
update b694aaf5-a1f7-4f4e-b791-c096844e7b59 failed (code=6)
```

`code=6` translates to [`HfError::NotMuxedToSp`](https://github.com/oxidecomputer/hubris/blob/6d2657c5ed173381295329f9ececef6b2cc83479/drv/gimlet-hf-api/src/lib.rs#L29). We then checked, and this sled was back in A0 and stuck in a reboot loop due to mismatched phase1/phase2 flash contents.

We believe the sequence of events was:

1. Wicket sent the sled to A2. (wicket believes this was successful)
2. We finished "preparing for update", which erases the flash. (wicket believes this was successful)
3. While the new phase1 data was being streamed in, the sled went back to A0, which switched the mux back to the host and caused the update writes to fail.

Assuming this sequence is correct, we don't know what caused the sled to go back to A0. There's an SP dump in `/staff/core/rack3/colo-64/BRM42220022`. I don't know the sequencer side of the SP all that well, but I don't see any smoking guns (e.g., there are no tasks with nonzero generations). I'm not sure we'll be able to get anything useful out of ringbufs, because the host being in a reboot loop means it intentionally went through some number of A2 trips. This is very visible in the `host_sp_comms` ringbuf, which ends with:

```
   1  690    13754        1 Request { now: 0x4ace2e6, sequence: 0x6, message: HostBootFailure { reason: 0x2 } }
   2  255    13754        1 Response { now: 0x4ace2e6, sequence: 0x8000000000000006, message: Ack }
   3  690    13754        1 Request { now: 0x4ace2f6, sequence: 0x7, message: HostPanic { code: 0x1101 } }
   4  255    13754        1 Response { now: 0x4ace2f6, sequence: 0x8000000000000007, message: Ack }
   5  690    13754        1 Request { now: 0x4acec94, sequence: 0x8, message: RequestReboot }
   6  356    13754        1 SetState { now: 0x4aced6a, state: A2 }
   7  685    13754        1 ParseError(Cobs)
[...]

PR #1618, opened 2024-02-14
host-sp-comms: do not reboot host on boot failure

### LINKED ISSUE #1614: SP should not auto-reboot host in response to a host-reported boot failure
(Fallout from #1613)

Currently, if the host reports a boot failure over the IPCC link, we respond by recording the information and resuming normal business. When the host immediately follows that up with a reboot request, we dutifully reboot the host.

Because we haven't taken any additional actions to _fix the boot failure_ (by, for instance, flipping the host flash mux), this will probably always produce a reboot loop.

While this sort of reboot loop is likely not _destructive,_ it's distracting: the machine cycles, the logs/ringbufs get overwritten, power is wasted, etc. I think after a boot failure like this, we should probably not attempt to boot the host until we have reason to believe the failure has been repaired.

The SP itself doesn't have sufficient context to know how to "repair" such a failure. If the failure was hit while attempting a recover image boot through Wicket, for instance, we specifically _do not_ want to do an automatic slot fallback. If we hit it during a production software upgrade, we _might,_ depending on circumstances, want to do a slot fallback. The right answer in basically all cases appears to be: escalate to the control plane, where context is more easily available.

So, I think we should stop rebooting the host after a boot failure, period, and wait for messages over the network. The boot failure is stored in a place the control plane can get to it (in the control-plane-agent). If we had a way of proactively sounding an alarm, we could do that, but for now it'd have to be polled.

Concretely, I discussed this briefly with @wesolows and the simplest thing appears to be:
1. Honor reboot requests from the host normally, except
2. If we get a host boot failure message, set a flag that causes the next reboot request to be interpreted as "power down and intervene."

cbiffle:
This is in response to #1613. If the host reports a boot failure (such as a phase mismatch, but not limited to that reason) simply rebooting it blindly is unlikely to fix the problem. We need intervention from a higher power (the control plane) to fix the issue.

So to avoid a bootloop that wastes energy and overwrites our circular buffers with spam, this change alters the response to the IPCC Request Reboot message if received shortly after a Boot Failed message -- it is interpreted as a power off request.

Fixes #1614.

### DIFF
--- a/task/host-sp-comms/src/main.rs
+++ b/task/host-sp-comms/src/main.rs
@@ -258,6 +258,13 @@ struct ServerImpl {
     reboot_state: Option<RebootState>,
     host_kv_storage: HostKeyValueStorage,
     hf_mux_state: Option<HfMuxState>,
+
+    /// Set when we receive a Boot Failed message from the host. This flag
+    /// changes the behavior in response to the next Request Reboot message. It
+    /// is cleared at Request Reboot, *or* at any of the messages sent during
+    /// host boot, to try and keep our state from getting out of sync for
+    /// whatever reason.
+    host_boot_just_failed: bool,
 }
 
 impl ServerImpl {
@@ -290,6 +297,8 @@ impl ServerImpl {
             reboot_state: None,
             host_kv_storage: HostKeyValueStorage::claim_static_resources(),
             hf_mux_state: None,
+
+            host_boot_just_failed: false,
         }
     }
 
@@ -701,13 +710,24 @@ impl ServerImpl {
 
         // We defer any actions until after we've serialized our response to
         // avoid borrow checker issues with calling methods on `self`.
+        //
+        // NOTE: setting any action will also clear the `host_boot_just_failed`
+        // flag, to avoid keeping incorrect state around. Since we have no
+        // explicit way of tracking the host boot software's lifecycle, this is
+        // a compromise.
         let mut action = None;
+
         let response = match request {
             HostToSp::_Unused => {
                 Some(SpToHost::DecodeFailure(DecodeFailureReason::Deserialize))
             }
             HostToSp::RequestReboot => {
-                action = Some(Action::RebootHost);
+                if core::mem::take(&mut self.host_boot_just_failed) {
+                    // Do _not_ reboot. Reinterpret this as a poweroff request.
+                    action = Some(Action::PowerOffHost);
+                } else {
+                    action = Some(Action::RebootHost);
+                }
                 None
             }
             HostToSp::RequestPowerOff => {
@@ -777,6 +797,11 @@ impl ServerImpl {
                 for b in &mut self.host_kv_storage.last_boot_fail[n..] {
                     *b = 0;
                 }
+
+                // Record that this failure happened, which will change the
+                // behavior of the reboot request we expect to follow shortly.
+                self.host_boot_just_failed = true;
+
                 Some(SpToHost::Ack)
             }
             HostToSp::HostPanic { .. } => {
@@ -793,6 +818,10 @@ impl ServerImpl {
                 for b in &mut self.host_kv_storage.last_panic[n..] {
                     *b = 0;
                 }
+
+                // Neither set nor clear the host_boot_just_failed flag, for
+                // this message often immediately follows a boot failure.
+
                 Some(SpToHost::Ack)
             }
             HostToSp::GetStatus => {
@@ -888,6 +917,9 @@ impl ServerImpl {
         // Now that all buffer borrowing is done, we can borrow `self` mutably
         // again to perform any necessary action.
         if let Some(action) = action {
+            // Protectively clear this flag to avoid keeping state.
+            self.host_boot_just_failed = false;
+
             match action {
                 Action::RebootHost => self.power_off_host(true),
                 Action::PowerOffHost => self.power_off_host(false),


task/host-sp-comms/src/main.rs
```diff
@@ -701,13 +710,24 @@ impl ServerImpl {
...
+        // a compromise.
         let mut action = None;
+
         let response = match request {
             HostToSp::_Unused => {
                 Some(SpToHost::DecodeFailure(DecodeFailureReason::Deserialize))
             }
             HostToSp::RequestReboot => {
-                action = Some(Action::RebootHost);
+                if core::mem::take(&mut self.host_boot_just_failed) {
```
hawkw, line 725: The use of `mem::take` here is very cute! But, IIUC, since we clear the flag later if `action` is `Some`, we don't actually _need_ to clear it when we read the flag's value here, do we? Since both branches set `action` to `Some` regardless?

Up to you whether this actually ought to change, but at a first glance it kind of appeared that this was the only case where we clear the flag, so it was a bit less obvious that _any_ branch where `action` is `Some` clears the flag (despite the comment). :woman_shrugging:

task/host-sp-comms/src/main.rs
```diff
@@ -777,6 +797,11 @@ impl ServerImpl {
                 for b in &mut self.host_kv_storage.last_boot_fail[n..] {
                     *b = 0;
                 }
+
+                // Record that this failure happened, which will change the
+                // behavior of the reboot request we expect to follow shortly.
+                self.host_boot_just_failed = true;
```
jgallagher, line 803: Do we care about the `reason` (which we're `..`'ing out in the match arm), or should all boot failure reasons be treated equally? Maybe a question for @citrus-it

task/host-sp-comms/src/main.rs
```diff
@@ -701,13 +710,24 @@ impl ServerImpl {
...
+
         let response = match request {
             HostToSp::_Unused => {
                 Some(SpToHost::DecodeFailure(DecodeFailureReason::Deserialize))
             }
             HostToSp::RequestReboot => {
-                action = Some(Action::RebootHost);
+                if core::mem::take(&mut self.host_boot_just_failed) {
+                    // Do _not_ reboot. Reinterpret this as a poweroff request.
+                    action = Some(Action::PowerOffHost);
```
jgallagher, line 727: I'm happy to look at this more tomorrow, but we might need to address the `TODO` I left in `handle_jefe_notification`. There may be a sequence of reboot operations where we still reboot even if we set this action.

cbiffle: Good to know, let's talk tomorrow.

task/host-sp-comms/src/main.rs
cbiffle: It would be my strong preference to have the same behavior regardless of reported reason, but I'm open to changing that if somebody wants to make an argument!

task/host-sp-comms/src/main.rs
cbiffle: Yeah, I wound up doing both here to make it really really unlikely that we forget to clear the flag. Arguably unnecessary. What do you think would be clearest?

hawkw: I think maybe leaving it as is, but commenting that this also gets cleared because we've set `action = Some(...)`? Honestly, I think it's not really a big deal --- mostly, I wanted to make sure my understanding was correct.

task/host-sp-comms/src/main.rs
citrus-it: We don't ever expect to see any boot failures apart from in the lab where they're triggered pretty often in my experience. It just takes somebody to do something like configure net boot and forget to start the boot server on the butler, or start it with a typo in the options. We should probably send a heads-up about this new behaviour.

It is very unlikely that any of the boot failure reasons indicate that there would be a different result if retried, I think they should all be treated equally as requiring intervention.

rmustacc: FWIW, I did just hit this exact case the other day personally because I wasn't fast enough in the lab with net boot/partially distracted. Is there something that makes it very clear from the sequencer state or related that we're in this case as a human looking at the system?

task/host-sp-comms/src/main.rs
jgallagher: I think I have convinced myself this is okay (where "this" is really "the handling of `self.reboot_state`"):

1. `self.reboot_state` starts out `None`
2. It can only go from `None` to `Some(_)` inside a call to `power_off_host(true)`
3. Once it goes to `Some(_)`, it will eventually cause the host to come back on through one of these paths:
  - If `power_off_host(true)` succeeds in asking the sequencer to go to A2, it's set to `Some(WaitingForA2)`. Once we get a notification from jefe that we've entered A2, it will change to `Some(WaitingInA2RebootDelay)` and start the `Timers::WaitingInA2ToReboot` timer. Once that timer fires, we'll enter `handle_reboot_waiting_in_a2_timer`, set the state to A0, and set `self.reboot_state` back to `None.
  - If `power_off_host(true)` fails to ask the sequencer to go to A2 because we're already in A2, it's set to `Some(WaitingInA2RebootDelay)` and we start the `Timers::WaitingInA2ToReboot` timer. Once that timer fires, we'll enter `handle_reboot_waiting_in_a2_timer`, set the state to A0, and set `self.reboot_state` back to `None.

`power_off_host(true)` is only called in two places: if we get a jefe notification that we've gone to `A0Reset`, or if the host sends us `HostToSp::RequestReboot`. If either of those happened and before we go through the above timer-based process (~5 seconds) to reboot the host we landed here with a host boot fail, the `PowerOffHost` would be effectively ignored because we'd already be in the process of rebooting the host. But that doesn't seem likely, and if it happens it's maybe fine to honor the "first request" that triggered the reboot?

task/host-sp-comms/src/main.rs
cbiffle: The behavior as implemented in this commit won't affect the sequencer state -- the machine will just power itself down on boot failure and wait to be turned back on. We don't currently have a great way of exposing "the last problem with this was _____" data in production builds. As a user, what sort of experience would you hope for?

task/host-sp-comms/src/main.rs
cbiffle: Currently, if the host sends back-to-back "power off" and "reboot" requests, we'll behave ... rather strangely. I wonder if it might be worth adding code to stop listening to IPCC entirely starting at any reboot/powerdown request and continuing up until we believe the host to have been powered back on.

task/host-sp-comms/src/main.rs
cbiffle: The fact that it threw you suggests that I need to make it more clear, so I'll see about adding some comments.

PR #1620, opened 2024-02-15
Yo dawg i heard you like ringbufs

cbiffle:
The final commit is the relevant one, description copied below. The other two commits are cleanups on the way there.

---

Our ability to debug #1613 was hampered by the control-plane-agent
ringbuf being overwritten, on average, every 1.5 seconds. This adds a
secondary ringbuf that only holds "important" events -- specifically,
commands received from the network that affect the host.

Unfortunately due to the BSP structure in the agent right now, this
required copy-pasting a change into three files. Fixing that appears
difficult.

### DIFF
--- a/task/control-plane-agent/src/main.rs
+++ b/task/control-plane-agent/src/main.rs
@@ -155,6 +155,24 @@ enum MgsMessage {
 
 ringbuf!(Log, 16, Log::Empty);
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum CriticalEvent {
+    Empty,
+    /// We have received a network request to change power states. This record
+    /// logs the sender, in case the request was unexpected, and the target
+    /// state.
+    SetPowerState {
+        sender: sp_impl::SocketAddrV6,
+        port: SpPort,
+        power_state: PowerState,
+        ticks_since_boot: u64,
+    },
+}
+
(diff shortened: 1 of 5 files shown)


task/control-plane-agent/src/main.rs
```diff
@@ -155,6 +155,23 @@ enum MgsMessage {
...
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum CriticalEvent {
+    Empty,
+    /// We have received a network request to change power states. This record
+    /// logs the sender, in case the request was unexpected, and the target
+    /// state.
+    SetPowerState {
+        sender: sp_impl::SocketAddrV6,
+        port: SpPort,
+        power_state: PowerState,
```
jgallagher, line 167: Do you want to include a "timestamp" here too (probably system ticks)?

jgallagher (approved):
LGTM, thanks! Sorry you had to file these forms in triplicate.

task/control-plane-agent/src/main.rs
hawkw: If we do that, it seems potentially valuable to add the same timestamp to the non-CRITICAL ringbuf events, so that the temporal relationship between the CRITICAL and non-CRITICAL events could be correlated...honestly, even just a sequence number could do that...

PR #1621, opened 2024-02-20
ringbuf: Add event counters

hawkw:
Ringbufs provide a fixed-size buffer for storing diagnostic events,
which may include data. This is very useful for debugging. However,
because the ring buffer has a fixed size, only the last `N` events are
visible at any given time. If a ringbuf has a capacity of, say, 16
events, and a given event has occurred 10,000 times over the lifespan of
the task, there's currently no way to know that the event has happened
more than 16 times. Furthermore, there's no way to use the ringbuf to
determine whether a given event has _ever_ occurred during the task's
lifetime: if a given variant is not _currently_ present in the ringbuf,
it may still have occurred previously and been "drowned out" by
subsequent events. Finally, some targets disable the ringbuf entirely,
for space reasons, and on those targets, it may still be desirable to be
able to get some less granular diagnostic data out of the ringbuf events
a task records.

Therefore, this commit extends the `ringbuf` crate to support generating
counts of the number of times a particular event variant has been
recorded, in addition to a fixed-size buffer of events. These counters
provide less detailed data than inspecting the ringbuf contents, because
they don't record the values of any _fields_ on that event. However,
they allow recording very large numbers of events in a fixed amount of
space, and provide historical data for events that may have occurred in
the past and then "fallen off" the end of the ringbuf as new events were
recorded. By inspecting these counters, we can determine if a given
variant has *ever* been recorded, even if it isn't currently in the
buffer, and we can see the total number of times it has been recorded
over the task's entire lifetime.

Event counters are implemented using a new `CountedRingbuf` type, which
marries a ring buffer with a type which can count occurances of entry
variants. `CountedRingbuf`s may be declared using the `counted_ringbuf!`
macro, and entries can be recorded using the existing `ringbuf_entry!`
macro.

`CountedRingbuf` requires that the entry type implement a new `Count`
trait, which defines the counter type, an initializer for creating new
instances of the counter type, and a method to increment the counter
type with a given instance of the entry type. An implementation of
`ringbuf::Count` can be generated for an entry type using the
`#[derive(ringbuf::Count)]` attribute, which generates a `struct` with
an `AtomicU32` field for each of the deriving `enum`'s variants. This
`struct` can then be loaded by Humility to provide a view of the event
counters.

Because recording event counters requires only `4 * <N_VARIANTS>` bytes,
counters are currently recorded even when the `ringbuf` crate has the
`disabled` feature set. This way, we can still record some diagnostic
data on targets that don't have the space to store the ringbuf itself.
If it becomes necessary to also disable counters, we could add a
separate `disable-counters` feature as well, so that a target can pick
and choose between recording ringbuf entries, counts, or both.

As a proof of concept, I've also updated `gimlet-seq-server` to use the
new event counters for its `Trace` ringbuf. Subsequent commits will roll
it out elsewhere.

Future work will include:

- [ ] Actually adopting event counters in existing ringbufs
- [ ] Updating `humility ringbuf` to also dump ringbuf event counts (see
  oxidecomputer/humility#449).

Depends on #1624


hawkw:
As a note, this design was based on the constraint that we would want to track the event counts per-ringbuf instance. If that's _not_ the case, and we're okay with counters being global for each enum type that derives `Event`, we could probably present a nicer API to the debugger by just generating a static `__COUNT_{variant-name}` for each enum variant, rather than making the debugger track down a `&'static str` pointer. This might be a bit faster as well, since incrementing the count could avoid array indexing (and thus, a bounds check)...

Is the constraint that we might want to use an event type in multiple ringbufs worth thinking about? Or are we okay with tracking counts at the level of the actual event type?

hawkw:
argh, i guess a bunch of the devboard targets don't have `fetch_add`; i guess i need to use `StaticCell` after all: https://github.com/oxidecomputer/hubris/actions/runs/7979951812/job/21788552533?pr=1621

cbiffle:
> argh, i guess a bunch of the devboard targets don't have `fetch_add`; i guess i need to use `StaticCell` after all: https://github.com/oxidecomputer/hubris/actions/runs/7979951812/job/21788552533?pr=1621

Yeah, this is what I meant in chat when I mentioned the M0 doesn't have atomics.

However, the M0 tends to be our most flash-constrained target, and often builds with ringbuf either reduced or disabled -- I wouldn't worry too much about good M0 support in this feature, personally, particularly since using atomics will be a lot smaller on every other platform.

cbiffle:
Initial thoughts:

Debugger support would be simplified if we can figure out how to put the counters in either named statics, or a struct with named fields, rather than an array.

At first glance I'm not enthusiastic about the generic name `event!` (though I suppose it could be qualified as `ringbuf::event!`) but I'm open to being sold on it.

If it's not a lot of work, could you `cargo expand` one of these? It'd be easier to review the generated code that way.

hawkw:
> Debugger support would be simplified if we can figure out how to put the counters in either named statics, or a struct with named fields, rather than an array.

Oh, generating a struct with named fields is a great idea, I like that much more than the array with string pointers in it.

hawkw:
@cbiffle if you have the chance, I'd love to get another look at this PR --- I've redesigned the way we implement event counts based on your suggestions. In particular:
- We now generate a `<EntryType>Counts` struct with named fields for each variant of `EntryType`, rather than an array, which should provide a much nicer debugger interface
- I've gotten rid of the need for separate macros for counted versus non-counted entries. Now, if a ringbuffer is declared using `counted_ringbuf!`, entries can be added to it using `ringbuf_entry!` just like a non-counted ringbuf.

lib/ringbuf/src/lib.rs
```diff
@@ -183,22 +186,73 @@ macro_rules! ringbuf {
...
+    ($name:ident, $t:ident, $n:expr, $init:expr) => {
+        #[used]
+        static $name: $crate::CountedRingbuf<$t, $n> = $crate::CountedRingbuf {
+            counters: <$t as $crate::Count>::NEW_COUNTERS,
+        };
+    };
+    ($t:ident, $n:expr, $init:expr) => {
+        $crate::counted_ringbuf!(__RINGBUF, $t, $n, $init);
     };
 }
```
hawkw, line 246: Note that currently, we still record event counts when the ringbuf crate is compiled with the "disabled" feature flag set. My thinking was that, since `<entry type number of variants> * 4` bytes is generally smaller than a whole ringbuf, where the size of the ringbuf is:
```rust
let entry_sz = size_of::<T>() // payload type
   + 2 // 16-bit line number
   + 2 // 16-bit generation
   + 4 // 32-bit count
   ;
entry_size * N
```
we could still get away with recording the counters on memory-constrained targets.

If others disagree, I'm happy to either change this code so that counts are not recorded when the "disabled" feature is set, or add an additional feature flag to control whether entry counts are recorded.

lib/ringbuf/src/lib.rs
```diff
@@ -280,36 +307,115 @@ pub struct Ringbuf<T: Copy + PartialEq, const N: usize> {
...
-                    0
-                } else {
-                    last + 1
+    /// Increment the counter for this event.
+    fn count(&self, counters: &Self::Counters);
+}
+
+pub trait RecordEntry<T: Copy + PartialEq> {
+    fn record_entry(&self, line: u16, payload: T);
+}
```
hawkw, line 350: This is turned into a trait in order to just have one `ringbuf_entry!` macro that can record entries in a counted ringbuf _or_ a ringbuf without counts.

build/util/src/lib.rs
```diff
@@ -56,7 +56,7 @@ pub fn has_feature(s: &str) -> bool {
 ///
 /// This will set one of `cfg(armv6m)`, `cfg(armv7m)`, or `cfg(armv8m)`
 /// depending on the value of the `TARGET` environment variable.
-pub fn expose_m_profile() {
+pub fn expose_m_profile() -> Result<()> {
```
hawkw, line 59: This change was necessary because I added an `examples/` dir to `ringbuf`, and the `cargo test` run on CI for the host target attempts to build examples as well as running tests. Since `ringbuf` depends on `static_cell` (and now, on `armv6m-atomic-hack` as well), building `ringbuf` calls `expose_m_profile`, which panics on x86_64 build hosts. Therefore, I've changed this method to return a `Result` instead, which the libs that _can_ reasonably be built for x86_64 (`ringbuf`, `static-cell`, `armv6m-atomic-hack`) turn into a warning, as we only need to determine ARM profile for ARM targets, while the crates that only build for ARM unwrap, maintaining the previous behavior of failing the build if built for a non-ARM target.

This, unfortunately, required changing a bunch of build scritps as well. Sorry about the big diff!

hawkw (commented):
Some notes for reviewers

hawkw:
Well, it does seem to work! After switching the `Ringbuf::record_entry` implementation back to the version on `master` (the version from #1624 seems busted), and using a build of Humility from oxidecomputer/humility#449, I can show entry variant totals in `humility ringbuf`: 

```console
eliza@cadbury ~ $ pfexec ./humility-bufcounts -t c71 ringbuf gimlet_seq
ENTRY TOTALS:
    Ice40Rails                1
    IdentValid                1
    ChecksumValid             1
    Reprogram                 1
    Programmed                1
    Programming               1
    Ice40PowerGoodV1P2        1
    Ice40PowerGoodV3P3        1
    RailsOff                  1
    Ident                     1
    A2Status                  1
    A2                        1
    A0FailureDetails          0
    A0Failed                  0
    A1Status                  0
    CPUPresent                0
    Coretype                  0
    A0Status                  0
    A0Power                   0
    NICPowerEnableLow         0
    RailsOn                   0
    UartEnabled               0
    A0                        0
    SetState                  0
    UpdateState               0
    ClockConfigWrite        446
    ClockConfigSuccess        1
    Status                    0
    PGStatus                  0
    SMStatus                  0
    ResetCounts               0
    PowerControl              0
    InterruptFlags            0
    V3P3SysA0VOut             1
    SpdBankAbsent             0
    SpdAbsent                14
    SpdDimmsFound             1
    I2cFault                  0
    StartFailed               0
    None                      0
   0  239        1        1 Ice40Rails(true, true)
   1  268        1        1 Ice40PowerGoodV1P2(true)
   2  289        1        1 Ice40PowerGoodV3P3(true)
   3  331        1        1 IdentValid(false)
   4  334        1        1 ChecksumValid(false)
   5  337        1        1 Reprogram(true)
   6  351        1        1 Programming
   7  380        1        1 Programmed
   8  384        1        1 RailsOff
   9  387        1        1 Ident(0xde01)
  10  393        1        1 A2Status(0x0)
  11  416        1      446 ClockConfigWrite
  12  427        1        1 ClockConfigSuccess
  13 1308        1        1 V3P3SysA0VOut(Volts(0.06640625))
  14  429        1        1 A2
  15 1078        1        1 SpdAbsent(0x0, 0x0, 0x0)
  16 1078        1        1 SpdAbsent(0x0, 0x1, 0x1)
  17 1078        1        1 SpdAbsent(0x0, 0x2, 0x2)
  18 1078        1        1 SpdAbsent(0x0, 0x3, 0x3)
  19 1078        1        1 SpdAbsent(0x0, 0x4, 0x4)
  20 1078        1        1 SpdAbsent(0x0, 0x5, 0x5)
  21 1078        1        1 SpdAbsent(0x0, 0x6, 0x6)
  22 1078        1        1 SpdAbsent(0x1, 0x0, 0x8)
  23 1078        1        1 SpdAbsent(0x1, 0x1, 0x9)
  24 1078        1        1 SpdAbsent(0x1, 0x2, 0xa)
  25 1078        1        1 SpdAbsent(0x1, 0x3, 0xb)
  26 1078        1        1 SpdAbsent(0x1, 0x4, 0xc)
  27 1078        1        1 SpdAbsent(0x1, 0x5, 0xd)
  28 1078        1        1 SpdAbsent(0x1, 0x6, 0xe)
  29 1121        1        1 SpdDimmsFound(0x2)
humility: ring buffer drv_oxide_vpd::__RINGBUF in gimlet_seq:
```

The CLI output could probably be made a bit more concise (we might want to hide variants whose count is 0 by default, and maybe there should be a flag to control whether or not we display totals), but it definitely seems to work!

lib/ringbuf/macros/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "ringbuf-macros"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+proc-macro = true
+
+[features]
+# To disable a ring buffer (but leave it otherwise present), enable the
```
cbiffle, line 10: Do you need this on the derive crate?

hawkw: ah, yeah, this is vestigial, good catch!

labbott (commented):
Does this need a full hubris archive version bump? (I can't remember under what circumstances we would do that vs just bumping humility version)

cbiffle (approved):

PR #1622, opened 2024-02-21
net: remove MAC RX activity watchdog.

cbiffle:
Now that #750 is closed and we're pretty confident in the fix, we can stop rebooting the netstack periodically. This should save resources (by not constantly dumping and restarting), and also get us back into the state where any task's generation number advancing is Probably Bad.


dancrossnyc (approved):

mkeeter (approved):
I did a quick test on a Gimletlet, and confirmed that `faux-mgs state` still works fine.

PR #1623, opened 2024-02-21
net: fix several unwraps of complex errors

cbiffle:
This knocks about 3.7 kiB (!) off Gimlet's netstack task size, by removing much of its dependence on the formatting machinery invoked by unwrap. (It's not all gone, because smoltcp still uses it.)

### DIFF
--- a/drv/ksz8463/src/lib.rs
+++ b/drv/ksz8463/src/lib.rs
@@ -12,7 +12,7 @@ pub use registers::{MIBCounter, Register};
 
 ////////////////////////////////////////////////////////////////////////////////
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq)]
 pub enum Error {
     SpiError(SpiError),
     WrongChipId(u16),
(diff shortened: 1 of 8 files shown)


mkeeter (approved):

hawkw (commented):
Looks like [you missed an `unwrap()` that needs to be `unwrap_lite()` now](https://github.com/oxidecomputer/hubris/actions/runs/7995659747/job/21836473427?pr=1623#step:6:165).

hawkw (approved):

cbiffle:
The number of places where no VSC chips are used, but the VSC _types_ appear for various reasons, has turned this into quite a game of whack-a-mole. With any luck it will pass the build eventually; I keep failing to predict which builds are affected when I test locally (e.g. the nucleo h7 build winds up depending on the VSC error types).

hawkw (approved):
looks like CI is happy now!

cbiffle:
Updated numbers post-whackamole:

Gimlet E net stack: -16,388 bytes
Nucleo 743 net stack: unchanged (unwrap I removed can't actually be reached)
Monorail: -8,820 bytes
Sidecar D net stack: ...-8 bytes? something is weird in the sidecar net stack that I need to investigate. It has a very large data segment that got larger with this change.

PR #1624, opened 2024-02-21
ringbuf: rewrite entry to have no panics

cbiffle:
This knocks about 120 bytes off of each ringbuf in a task (because the Ringbuf::entry routine gets specialized for each ringbuf, and this makes that routine smaller).

Reduces Gimlet flash image size by 3.4%.

### DIFF
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -282,33 +282,58 @@ pub struct Ringbuf<T: Copy + PartialEq, const N: usize> {
 
 impl<T: Copy + PartialEq, const N: usize> Ringbuf<T, { N }> {
     pub fn entry(&mut self, line: u16, payload: T) {
-        let ndx = match self.last {
-            None => 0,
-            Some(last) => {
-                let ent = &mut self.buffer[last];
+        // If this is the first time this ringbuf has been poked, last will be
+        // None. In this specific case we want to make sure we don't add to the
+        // count of an existing entry, and also that we deposit the first entry
+        // in slot 0. From a code generation perspective, the cheapest thing to
+        // do is to treat None as an out-of-range value:
+        let last = self.last.unwrap_or(usize::MAX);
 
-                if ent.line == line && ent.payload == payload {
-                    // Only reuse this entry if we don't overflow the
-                    // count.
-                    if let Some(new_count) = ent.count.checked_add(1) {
-                        ent.count = new_count;
-                        return;
-                    }
+        // Check to see if we can reuse the most recent entry. This uses get_mut
+        // both to avoid checking an entry on the first insertion (see above),
+        // and also to handle the case where last is somehow corrupted to point
+        // out-of-range. This avoids a bounds check panic. In the event that
+        // last _is_ corrupted, the behavior below will just start us over at 0.
+        if let Some(ent) = self.buffer.get_mut(last) {
+            if ent.line == line && ent.payload == payload {
+                // Only reuse this entry if we don't overflow the
+                // count.
+                if let Some(new_count) = ent.count.checked_add(1) {
+                    ent.count = new_count;
+                    return;
                 }
+            }
+        }
 
-                if last + 1 >= self.buffer.len() {
-                    0
-                } else {
-                    last + 1
-                }
+        // Either we were unable to reuse the entry, or the last index was out
+        // of range (perhaps because this is the first insertion). We're going
+        // to advance last and wrap if required. This uses a wrapping_add
+        // because if last is usize::MAX already, we want it to wrap to zero
+        // regardless -- and this avoids a checked arithmetic panic on the +1.
+        let ndx = {
+            let last_plus_1 = last.wrapping_add(1);
+            // You're probably wondering why this isn't a remainder operation.
+            // This is for two reasons:
+            // 1. None of our target platforms currently have hardware modulus,
+            //    and many of them don't even have hardware divide, making
+            //    remainder quite expensive.
+            // 2. The code as written here correctly turns usize::MAX into 0 for
+            //    our starting condition. Otherwise we'd have to be cleverer
+            //    about our starting number.
+            if last_plus_1 >= self.buffer.len() {
+                0
+            } else {
+                last_plus_1
             }
         };
 
         let ent = &mut self.buffer[ndx];
-        ent.line = line;
-        ent.payload = payload;
-        ent.count = 1;
-        ent.generation = ent.generation.wrapping_add(1);
+        *ent = RingbufEntry {
+            line,
+            payload,
+            count: 1,
+            generation: ent.generation.wrapping_add(1),
+        };
 
         self.last = Some(ndx);
     }


lib/ringbuf/src/lib.rs
```diff
@@ -282,33 +282,50 @@ pub struct Ringbuf<T: Copy + PartialEq, const N: usize> {
...
-                    last + 1
-                }
+        // Either we were unable to reuse the entry, or the last index was out
+        // of range (perhaps because this is the first insertion). We're going
+        // to advance last and wrap if required. This uses a wrapping_add
+        // because if last is usize::MAX already, we want it to wrap to zero
+        // regardless -- and this avoids a checked arithmetic panic on the +1.
+        let ndx = {
+            let last_plus_1 = last.wrapping_add(1);
+            if last_plus_1 >= self.buffer.len() {
```
dancrossnyc, line 315: Is the generated code quality still very poor when using modulus here?  This would be cleaner if it were just,

`let ndx = last.wrapping_add(1) % self.buffer.len();`

But if the generated code for that is still excessively verbose, it may be worth adding a comment about why this is written the way that it is.

hawkw: Out of curiousity, i tested this in Compiler Explorer, and unfortunately, both the `%` operator and `.wrapping_rem(self.buffer.len())` seem to generate _substantially_ more code than the `if` (although `wrapping_rem` is a little bit shorter). Since both `%` and `wrapping_rem` panic if the divisor is 0, they both generate a bunch of panic formatting code. 

See [this godbolt](https://godbolt.org/z/ex6Mv7G6s) (which might be nice to reference in the comment, actually!) for details.

hawkw (approved):
This is lovely, and I really appreciate the comments (as usual!)

lib/ringbuf/src/lib.rs
dancrossnyc: The genesis of the size concern is based on what this compiles to on M0, which lacks hardware division; the result there (at least as of a few months or a year ago or so, according to a Matrix post from Cliff: https://matrix.to/#/!jxnlDLnJsVaIRstnrt:oxide.computer/$li8Y-dTHnOVSdzi9z1n0tBqU-4xq9OXGLCahueefYXw?via=oxide.computer&via=unix.house&via=matrix.org) is 199 (!!) instructions. By comparison, any of these is significantly better.

The example for `wrapping_rem` on godbolt seemed incorrect (it's written as `wrapping_rem(1) % buflen`), but when I changed it to, `last.wrapping_add(1).wrapping_rem(buflen)` it appeared to generate the exact same code as the bare `%` version.

FWIW, the shortest version (at least for M$n$ where $n > 0$) seems to be, `unsafe { core::intrinsics::unchecked_rem(last.wrapping_add(1), buflen) }`, but that is terrible for other reasons.

In this case, it would be interesting to see what this compiles to in-situ; it wouldn't surprise me if the compiler couldn't figure out that `self.buffer.len()` is always > 0. My testing suggests that, if the compiler can figure out that the divisor is positive, it elides the panic check and associated support code.

hawkw:
Hmm, I have some bad news about this. After rebasing my branch from #1621 to use this code for `Ringbuf::entry` rather than the code on `master`, all my ringbufs appear to be empty:
```console
eliza@cadbury ~ $ pfexec humility -t c71 ringbuf
humility: ringbuf dump failed: member missing from struct: cell
humility: ring buffer drv_i2c_devices::adm1272::__RINGBUF in power:
humility: ring buffer drv_i2c_devices::max31790::__RINGBUF in thermal:
humility: ring buffer drv_i2c_devices::sbrmi::__RINGBUF in sbrmi:
humility: ring buffer drv_oxide_vpd::__RINGBUF in gimlet_seq:
humility: ring buffer drv_oxide_vpd::__RINGBUF in host_sp_comms:
humility: ring buffer drv_sbrmi::__RINGBUF in sbrmi:
humility: ring buffer drv_stm32h7_spi_server_core::__RINGBUF in spi2_driver:
humility: ring buffer drv_stm32h7_spi_server_core::__RINGBUF in sprot:
humility: ring buffer drv_stm32h7_sprot_server::__RINGBUF in sprot:
humility: ring buffer drv_stm32xx_i2c::__RINGBUF in i2c_driver:
humility: ring buffer drv_stm32xx_i2c::__RINGBUF in spd:
humility: ring buffer drv_stm32xx_i2c_server::__RINGBUF in i2c_driver:
humility: ring buffer ksz8463::__RINGBUF in net:
humility: ring buffer ksz8463::__RINGBUF in host_sp_comms:
humility: ring buffer stm32h7_update_server::__RINGBUF in update_server:
humility: ring buffer task_control_plane_agent::CRITICAL in control_plane_agent:
humility: ring buffer task_control_plane_agent::__RINGBUF in control_plane_agent:
humility: ring buffer task_control_plane_agent::update::rot::__RINGBUF in control_plane_agent:
humility: ring buffer task_dump_agent::__RINGBUF in dump_agent:
humility: ring buffer task_dump_agent::udp::__RINGBUF in dump_agent:
humility: ring buffer task_hiffy::stm32h7::__RINGBUF in hiffy:
humility: ring buffer task_jefe::dump::__RINGBUF in jefe:
humility: ring buffer task_jefe::external::__RINGBUF in jefe:
humility: ring buffer task_net::mgmt::__RINGBUF in net:
humility: ring buffer task_power::__RINGBUF in power:
humility: ring buffer task_power::bsp::__RINGBUF in power:
humility: ring buffer task_spd::__RINGBUF in spd:
humility: ring buffer task_validate::__RINGBUF in validate:
humility: ring buffer vsc85xx::__RINGBUF in net:
```

After changing my PR back to use the previous code, the ringbufs contain entries as expected:

```console
eliza@cadbury ~ $ pfexec humility -t c71 ringbuf
humility: ringbuf dump failed: member missing from struct: cell
humility: ring buffer drv_i2c_devices::adm1272::__RINGBUF in power:
   0  106        1        1 Config(CommandData(0x3f3f))
   1  159        1        1 Coefficients(Coefficients { m: 0xfde, b: 0x0, R: 0xfe })
   2  177        1        1 Coefficients(Coefficients { m: 0x297, b: 0x5000, R: 0xff })
   3  205        1        1 Coefficients(Coefficients { m: 0x2927, b: 0x0, R: 0xfd })
   4  106        1        1 Config(CommandData(0x3f3f))
   5  159        1        1 Coefficients(Coefficients { m: 0xfde, b: 0x0, R: 0xfe })
   6  177        1        1 Coefficients(Coefficients { m: 0x52e, b: 0x5000, R: 0xff })
   7  205        1        1 Coefficients(Coefficients { m: 0x524e, b: 0x0, R: 0xfd })
humility: ring buffer drv_i2c_devices::max31790::__RINGBUF in thermal:
humility: ring buffer drv_i2c_devices::sbrmi::__RINGBUF in sbrmi:
humility: ring buffer drv_oxide_vpd::__RINGBUF in gimlet_seq:
humility: ring buffer drv_oxide_vpd::__RINGBUF in host_sp_comms:
humility: ring buffer drv_sbrmi::__RINGBUF in sbrmi:
humility: ring buffer drv_stm32h7_spi_server_core::__RINGBUF in spi2_driver:
  52  494     5450        1 WaitISR(0x32017)
  53  464     5450        1 Rx(0x0)
  54  464     5450        1 Rx(0x7)
  55  494     5450        1 WaitISR(0x10012)
  56  464     5450        1 Rx(0x6)
  57  322     5450        1 Start(exchange, (0x2, 0x4))

 < ... REDACTED FOR BREVITY ... >
```
(n.b. that the `gimlet_seq` ringbuf parse failure is because i'm using that ringbuf for testing my event counts change and should be expected).

I haven't tested _just_ this branch yet, so it's possible that it's an interaction with my code, but I intend to shortly --- @cbiffle, have you actually tried this out yet?

lib/ringbuf/src/lib.rs
hawkw: Ah, yeah, I see that if I change the target triple in the godbolt to the correct one for Cortex-M0 (`thumbv6m-none-eabi`), both the `%` and `wrapping_rem` (which I fixed) generate a call to `__aeabi_uidivmod`, which I assume is where the 199 instructions live (the Matrix link doesn't work for me; I'm guessing because I wasn't in that channel at the time). Here's my revised godbolt: https://godbolt.org/z/sac9PhzWx if it's useful.

cbiffle: Can add a comment -- the reason I didn't use modulus here is two-fold after this change. First, we don't have hardware modulus (and on the M0, as you note, we don't even have hard divide). Second, it would require my `usize::MAX`-becomes-0 hack to be smarter, since `usize::MAX % n` is generally not zero.

cbiffle:
Ringbufs with this change check out locally; issue appears to have been a conflict between our changes.

lib/ringbuf/src/lib.rs
```diff
@@ -282,33 +282,58 @@ pub struct Ringbuf<T: Copy + PartialEq, const N: usize> {
...
+        // because if last is usize::MAX already, we want it to wrap to zero
+        // regardless -- and this avoids a checked arithmetic panic on the +1.
+        let ndx = {
+            let last_plus_1 = last.wrapping_add(1);
+            // You're probably wondering why this isn't a remainder operation.
+            // This is for two reasons:
+            // 1. None of our target platforms currently have hardware modulus,
+            //    and many of them don't even have hardware divide, making
+            //    remainder quite expensive.
+            // 2. The code as written here correctly turns usize::MAX into 0 for
```
dancrossnyc, line 320: Thanks for the comment! I'd remove this bullet point, though, as this code would be correct regardless. Specifically, one presumes that with modulus, this code would be written as `let ndx = last.wrapping_add(1) % self.buffer.len();`.  Since usize::MAX.wrapping_add(1) is 0, and 0 (mod N) is 0 for N != 0, this would still (correctly) be 0.

lib/ringbuf/src/lib.rs
dancrossnyc: Not to flog a dead horse, but I was kind of curious just how bad this still was, so on a lark I played around with it for a few minutes. My testing was against app/nucleo-stm32g0/app-g070.toml and I compared generated code from el `jefe`.

One thing I noticed that this function is generic over `const N: usize`, and that `N` is the length of `self.buffer`, but is also a compile-time constant. I was curious what code would be generated for `let ndx = last.wrapping_add(1) % N;` --- presumably the optimizer can do a somewhat better job, since it knows that N's exact value, and that N>0. However, the compiler is too smart for that and both this case and the obvious case of `let ndx = last.wrapping_add(1) % self.buffer.len();` generated identical code.  One presumes that the (constant) size of the ring is propagated through compilation.

Interestingly, when ringbuf sizes are powers of two (as they are pretty much everywhere), the compiler does a really good job of optimizing things, and indeed, it generated largely the same code for both modulus versions and the explicit `if` construction (for some reason, r4 and r5 are swapped between the two versions and they are of course different, but interestingly seem to be the same size).  Here's the most interesting part of the diff:

```diff
-     162: 1c6c         	adds	r4, r5, #0x1
-     164: 2c03         	cmp	r4, #0x3
-     166: d900         	bls	0x16a <ringbuf::Ringbuf$LT$T$C$_$GT$::entry::habadc9fb7659bc36+0x8e> @ imm = #0x0
-     168: 2400         	movs	r4, #0x0
-     16a: 4363         	muls	r3, r4, r3
-     16c: 18ca         	adds	r2, r1, r3
-     16e: 9b02         	ldr	r3, [sp, #0x8]
-     170: 6113         	str	r3, [r2, #0x10]
-     172: 8190         	strh	r0, [r2, #0xc]
-     174: 89d0         	ldrh	r0, [r2, #0xe]
-     176: 1c40         	adds	r0, r0, #0x1
-     178: 81d0         	strh	r0, [r2, #0xe]
-     17a: 2001         	movs	r0, #0x1
-     17c: 6090         	str	r0, [r2, #0x8]
-     17e: c111         	stm	r1!, {r0, r4}
+     162: 2201         	movs	r2, #0x1
+     164: 1c64         	adds	r4, r4, #0x1
+     166: 2503         	movs	r5, #0x3
+     168: 4025         	ands	r5, r4
+     16a: c124         	stm	r1!, {r2, r5}
+     16c: 436b         	muls	r3, r5, r3
+     16e: 3908         	subs	r1, #0x8
+     170: 18c9         	adds	r1, r1, r3
+     172: 9b02         	ldr	r3, [sp, #0x8]
+     174: 610b         	str	r3, [r1, #0x10]
+     176: 8188         	strh	r0, [r1, #0xc]
+     178: 608a         	str	r2, [r1, #0x8]
+     17a: 89c8         	ldrh	r0, [r1, #0xe]
+     17c: 1c40         	adds	r0, r0, #0x1
+     17e: 81c8         	strh	r0, [r1, #0xe]
```

On the other hand, if the buffer size is _not_ a power of two, the compiler does a very poor job at code generation, decaying to the div/rem intrinsics in software. Since nothing says that the ring buffer sizes need be powers of two, that could easily happen.

Anyway, it was an interesting digression for an hour.

PR #1632, opened 2024-02-26
Hubris-side changes for removing `dispatch_n` in `idol-runtime`

cbiffle:
This is the Hubris part of https://github.com/oxidecomputer/idolatry/pull/39, the description of which I've pasted below for your reference:

---

We ran into an issue last week where we inadvertently replaced a call to `idol_runtime::dispatch_n` with a call to `idol_runtime::dispatch`. This difficult-to-spot two-character change has the following implications:

1. No warnings or errors are produced.
2. Notifications to the server are silently ignored.

This is bad, and caused real failures in host startup.

This change alters the API to eliminate `dispatch_n` and make notification handling mandatory for all Idol servers -- to opt out of receiving notifications, they must explicitly implement `NotificationHandler` as stubs. This should prevent this class of error from reoccurring, and has the added benefit of simplifying the runtime implementation -- which I'd previously complicated in a misguided attempt at ergonomics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2487,8 +2487,8 @@ checksum = "b9e0384b61958566e926dc50660321d12159025e767c18e043daf26b70104c39"
 
 [[package]]
 name = "idol"
-version = "0.3.0"
-source = "git+https://github.com/oxidecomputer/idolatry.git#f2396893e786d4bfa75212312908198b8d6a5310"
+version = "0.4.0"
+source = "git+https://github.com/oxidecomputer/idolatry.git#86761da6c3cf95bd3f0335d08f810ba708775447"
 dependencies = [
  "indexmap",
  "quote",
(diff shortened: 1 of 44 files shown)


labbott (approved):
Part of me wishes there was something like `derive(DefaultNotification)`, maybe future work

hawkw:
Looks like we're missing an impl of `NotificationHandler` for [`drv_fpga_server::ServerImpl` ](https://github.com/oxidecomputer/hubris/actions/runs/8055009554/job/22000897764?pr=1632#step:6:342), and [`gimlet_seq_server` still calls `idol_runtime::dispatch_n`](https://github.com/oxidecomputer/hubris/actions/runs/8055009554/job/22000892512#step:6:282).

hawkw (approved):
:no_entry_sign::foot::gun:

hawkw:
oh, ugh, looks like you're [missing one in `dump-agent` too](https://github.com/oxidecomputer/hubris/actions/runs/8071664116/job/22051772653?pr=1632#step:6:214).

PR #1633, opened 2024-02-27
Add support for reading out the VPD lock bits via MGS

labbott:


task/control-plane-agent/src/mgs_common.rs
```diff
@@ -405,6 +406,72 @@ impl MgsCommon {
...
+        _buf: &mut [u8],
+    ) -> Result<usize, SpError> {
+        Err(SpError::Vpd(GatewayVpdError::NotImplemented))
+    }
+
+    #[cfg(feature = "vpd")]
+    pub(crate) fn vpd_lock_status_all(
+        &self,
+        buf: &mut [u8],
+    ) -> Result<usize, SpError> {
```
hawkw, line 422: Commit [4c3681dff78764374d6d873db5a4d2be272009f4](https://github.com/oxidecomputer/hubris/commit/cb27900aae24a90c2a162a0b42eb13a59af6f197) added event counters for each MGS message based on the ringbuf entries recorded for those messages. Can we add a variant for `VpdLockStatusAll` to `MgsMessage`, and record it in the ringbuf here, as well?

task/control-plane-agent/Cargo.toml
```diff
@@ -29,6 +29,7 @@ drv-stm32h7-usart = { path = "../../drv/stm32h7-usart", features = ["h753"], opt
 drv-transceivers-api = { path = "../../drv/transceivers-api", optional = true }
 drv-update-api = { path = "../../drv/update-api" }
 drv-user-leds-api = { path = "../../drv/user-leds-api", optional = true }
+task-vpd-api = { path = "../../task/vpd-api" }
```
mkeeter, line 32: Should this be optional (enabled by the `vpd` feature)?

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -405,6 +406,72 @@ impl MgsCommon {
...
+
+        let vpd = Vpd::from(VPD.get_task_id());
+        let cnt = vpd.num_vpd_devices();
+
+        for (i, entry) in buf.iter_mut().enumerate().take(cnt) {
+            // The API takes a `u8` and we're bounded by the number of devices
+            // in the array which is also going to be less than `u8`
+            let idx = match u8::try_from(i) {
+                Ok(v) => v,
+                Err(_) => unreachable!(),
```
mkeeter, line 434: Nit: this feels like it should be a `panic!` instead, because it's not immediately obvious that it's unreachable – it's not obvious that `buf.len() <= u8::MAX`, since it's caller-controlled, and `cnt` is a `usize`.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -405,6 +406,72 @@ impl MgsCommon {
...
+            };
+            *entry = match vpd.is_locked(idx) {
+                Ok(v) => v.into(),
+                Err(e) => {
+                    return Err(SpError::Vpd(match e {
+                        VpdError::InvalidDevice => {
```
mkeeter, line 440: Should we implement `From<VpdError> for GatewayVpdError`?

(It would have to go in `vpd-api`, so maybe that's too much plumbing)

task/control-plane-agent/src/mgs_common.rs
mkeeter: One other way to make it clear would be changing `num_vpd_devices` to return a `u8`.

(Also, is `panic!` more expensive than `unreachable!`?)

task/control-plane-agent/src/mgs_common.rs
hawkw: lovely, thanks!

hawkw (approved):

PR #1636, opened 2024-02-28
jefe: add counters to Jefe's ringbufs

hawkw:
Currently, Jefe has a few ringbufs that are quite small. This means that events may drop off the back of these ringbuffers. This commit updates these ringbufs to also count events, so that we can track the total number of times that various things have happened in Jefe.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1012,7 +1012,6 @@ dependencies = [
 name = "drv-i2c-types"
 version = "0.1.0"
 dependencies = [
- "anyhow",
  "build-util",
  "counters",
  "derive-idol-err",
@@ -4300,6 +4299,7 @@ dependencies = [
  "build-util",
  "cfg-if",
  "cortex-m",
+ "counters",
  "hubpack",
  "hubris-num-tasks",
  "humpty",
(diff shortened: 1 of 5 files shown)


hawkw:
I also thought about adding a counter for each IPC request variant, but adding an additional ringbuf (so that we can record the requesting task's ID for the last few IPCs) ends up making the Jefe task quite a bit bigger, which seemed unfortunate. If recording IPCs is something that's actually useful here, I can add that, either as a ringbuf + counters, or just counters without the requester's ID, depending on which seems more useful (and whether the increased memory is worth it).

hawkw:
> I also thought about adding a counter for each IPC request variant, but adding an additional ringbuf (so that we can record the requesting task's ID for the last few IPCs) ends up making the Jefe task quite a bit bigger, which seemed unfortunate. If recording IPCs is something that's actually useful here, I can add that, either as a ringbuf + counters, or just counters without the requester's ID, depending on which seems more useful (and whether the increased memory is worth it).

**Update**: Counting IPC requests is now mostly obsoleted by client-side IPC interface counters in `jefe-api` (added in #1642). The error counters in this branch are probably still desirable though!

PR #1637, opened 2024-02-28
net: Use `counters` for event counters, count errors

hawkw:
Currently, the `net` task has some hand-implemented event counters.
Although these are simple and efficient, they don't match the format
that `humility counters` will look for when finding counters in a dump,
so they can't be queried using that command. Instead, they have to be
inspected using a series of `humility readvar` commands. This branch,
therefore, replaces the hand-written counters with the `counters` crate,
allowing the `net` task's event counters to be understood by `humility
counters`.

Additionally, I've added counters to the ringbufs in some of the BSP
modules (currently, only Gimletlet actually has ringbufs). For these, we
skip counting the various status and control register ringbuf entries
because they don't really seem to represent a countable event, just the
state of a register, and skipping all of them lets us avoid several
words of counters. Instead, we just count the number of error event
variants.


task/net/src/bsp/gimletlet_mgmt.rs
```diff
@@ -28,64 +28,77 @@ use vsc85xx::VscError;
...
     Vsc8552MacPcsStatus {
         port: u8,
         status: phy::extended_3::MAC_SERDES_PCS_STATUS,
     },
+    #[count(skip)]
     Vsc8552MacPcsControl {
         port: u8,
         control: phy::extended_3::MAC_SERDES_PCS_CONTROL,
     },
+    #[count(skip)]
```
hawkw, line 72: We skip counting the various status and control register ringbuf entries because they don't _really_ seem to represent a countable event, just the state of a register, and skipping all of them lets us avoid several words of counters. Instead, we just count the number of error event variants.

task/net/src/main.rs
```diff
@@ -150,9 +150,17 @@ const RX_RING_SZ: usize = 4;
...
+enum Event {
+    /// Global count of passes through the driver loop, for inspection through a
+    /// debugger.
+    Iter,
+    /// IP activity occurred on an iteration.
+    IpActivity,
+    /// Timer wakeups
+    TimerWake,
+}
+counters!(Event);
```
hawkw, line 163: These counters are turned into an enum so that we can generate the same structure that the `humility counters` command knows about. Alternatively, if we don't like going through the enum, we _could_ hand-write a struct that _looks like_ the ones generated by the `counters` crate, like
```rust

struct EventCounts {
    Iter: AtomicU32,
    IpActivity: AtomicU32,
    TimerWake: AtomicU32,
}

#[used]
static EVENT_COUNTERS: EventCounts = ...;
```
and `fetch_add` to them, instead.

task/net/src/bsp/gimletlet_mgmt.rs
cbiffle: Consider making this a comment in the source code.

task/net/src/main.rs
cbiffle: I don't mind the enum, personally.

app/gimletlet/app-mgmt.toml
```diff
@@ -25,7 +25,7 @@ request_reset = ["hiffy"]
 name = "drv-stm32xx-sys"
 features = ["h753"]
 priority = 1
-max-sizes = {flash = 2048, ram = 1024}
+max-sizes = {flash = 2048, ram = 2048}
```
cbiffle, line 28: what happened here?

mkeeter: I suspect `app-mgmt.toml` was broken, because it's not checked by CI

task/net/src/bsp/gimletlet_mgmt.rs
```diff
@@ -28,64 +28,77 @@ use vsc85xx::VscError;
...
     Vsc8552MediaSerdesStatus {
         port: u8,
         status: phy::extended_3::MEDIA_SERDES_STATUS,
     },
     Vsc8552Err {
+        #[count(children)]
         err: VscError,
     },
     Vsc8552BypassControl {
```
mkeeter, line 81: Missing a `#[count(skip)]` here on `Vsc8552BypassControl`?

app/gimletlet/app-mgmt.toml
hawkw: We apparently don't actually build this task on CI, so it had, yet again, bitrotted. Since I had modified the Gimletlet `net` tasks, I wanted to make sure I hadn't made the `net` task too big for any of the Gimletlet apps, and had to fix this in order to verify that.

Happy to back out the change, because it is, admittedly, unrelated to the diff.

mkeeter (approved):

PR #1639, opened 2024-02-29
track/trace MAX5970 minimum and maximum values across a bounce

bcantrill:


mkeeter (approved):

task/power/src/bsp/gimlet_bcdef.rs
```diff
@@ -190,12 +202,23 @@ impl Max5970Peak {
...
         self.min = min;
         self.max = max;
+
+        if min < self.crossbounce_min {
+            self.crossbounce_min = min;
+        }
+
+        if max > self.crossbounce_max {
+            self.crossbounce_max = max;
+        }
```
hawkw, line 220: nit, take it or leave it: these could be expressed as 
```suggestion
        self.crossbounce_min = core::cmp::min(min, self.crossbounce_min);
        self.crossbounce_max = core::cmp::max(max, self.crossbounce_max);
```

hawkw (approved):

task/power/src/bsp/gimlet_bcdef.rs
bcantrill: FWIW, this was my first thought, but it doesn't compile:

```
   Compiling task-power v0.1.0 (/home/bmc/hubris/task/power)
error[E0277]: the trait bound `f32: Ord` is not satisfied
   --> task/power/src/bsp/gimlet_bcdef.rs:214:32
    |
214 |         self.crossbounce_min = core::cmp::min(min, self.crossbounce_min);
    |                                ^^^^^^^^^^^^^^ the trait `Ord` is not implemented for `f32`
    |
    = help: the following other types implement trait `Ord`:
              i128
              i16
              i32
              i64
              i8
              isize
              u128
              u16
            and 4 others
note: required by a bound in `core::cmp::min`

```

I didn't take it further than that, as I tend to prefer the slightly more verbose version anyway.

hawkw: Oh, yeah…I forgot about floats. Carry on! :)

PR #1640, opened 2024-03-01
counters: implement `Count` for `Option`/`Result`

hawkw:
This commit adds implementations of the `counters::Count` trait for
`Option<T> where T: Count` and `Result<T, E> where T: Count, E: Count`.
Currently, we aren't using these anywhere, but we _will_ be! :D

### DIFF
--- a/lib/counters/src/lib.rs
+++ b/lib/counters/src/lib.rs
@@ -178,19 +178,10 @@ impl Count for core::convert::Infallible {
     #[allow(clippy::declare_interior_mutable_const)]
     const NEW_COUNTERS: Self::Counters = AtomicU32::new(0);
 
-    fn count(&self, _counters: &Self::Counters) {
+    fn count(&self, _: &Self::Counters) {
         // `Infallible`s are not made. They should NEVER be made. We
         // will not count them. We will not help count them.
-        #[cfg(debug_assertions)]
-        unreachable!();
-
-        // but just in case...
-        #[cfg(not(debug_assertions))]
-        armv6m_atomic_hack::AtomicU32Ext::fetch_add(
-            _counters,
-            1,
-            Ordering::Relaxed,
-        );
+        match *self {}
     }
 }
 
--- a/lib/counters/src/lib.rs
+++ b/lib/counters/src/lib.rs
@@ -50,12 +50,12 @@ pub trait Count {
 /// static.
 #[macro_export]
 macro_rules! counters {
-    ($name:ident, $Type:ident) => {
+    ($name:ident, $Type:ty) => {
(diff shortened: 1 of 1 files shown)


lib/counters/src/lib.rs
```diff
@@ -77,3 +78,132 @@ macro_rules! count {
...
+    }
+}
+
+impl Count for core::convert::Infallible {
+    type Counters = AtomicU32;
+    #[allow(clippy::declare_interior_mutable_const)]
+    const NEW_COUNTERS: Self::Counters = AtomicU32::new(0);
+
+    fn count(&self, _counters: &Self::Counters) {
+        // `Infallible`s are not made. They should NEVER be made. We
```
cbiffle, line 182: I agree with the Litany Against Infallible comment here, but would suggest replacing the code in this function with the "discharge an uninhabited enum" pattern:

`match self {}`

That ensures that there's no potential panic site generated or unreachable code.

lib/counters/src/lib.rs
```diff
@@ -77,3 +78,132 @@ macro_rules! count {
...
+        #[cfg(not(debug_assertions))]
+        armv6m_atomic_hack::AtomicU32Ext::fetch_add(
+            _counters,
+            1,
+            Ordering::Relaxed,
+        );
+    }
+}
+
+impl Count for () {
```
cbiffle, line 197: This impl is kind of the first time where I've been a little concerned about us overflowing a U32. How often are we counting `()`?

lib/counters/src/lib.rs
hawkw: ah, good call, thank you!

lib/counters/src/lib.rs
hawkw: This impl is^H^Hwill be used for counting the `Ok(())` arm of a `Result<(), T> where T: Count`. In particular, for fallible IPC operations, Idol will generate an enum variant like 
```rust
enum ${name_of_ipc_interface}Event {
    ${name_of_fallible_ipc_op}(#[count(children)] Result<(), #{the_fallible_op_error_type}>,
    // ... other variants ...
}
```

so, there's no _global_ counter of `()`s; it's just used to instantiate a counter for the `Ok` side of a `Result` counter in a particular enum variant.

We could use similar counters in hand-written code; IIRC we have some counted ringbuffers that contain variants that are `Result<(), Error>` that aren't currently counting the `Ok`/`Err` variants separately, but could be. 

Does that explanation make sense?

cbiffle: It does, thanks.

PR #1644, opened 2024-03-05
update `idol` to use `quote!` in codegen

hawkw:
PR oxidecomputer/idolatry#40 updates `idol`'s code generation to use the `quote` crate for generating Rust code, rather than representing it as a string. This commit updates Hubris' dependency on `idol` to oxidecomputer/idolatry@c5222fe79b55e007db06bfdab0ea7e2b1db28629.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,6 +34,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


cbiffle (approved):
Ah, yes, I love how updating a proc macro brings in android-tzdata and core-foundation-sys like a perfectly normal package manager. ;-)

This is not your fault, however; looks good.

hawkw:
> Ah, yes, I love how updating a proc macro brings in android-tzdata and core-foundation-sys like a perfectly normal package manager. ;-)

Yeah, I would really love to know how those got in the dep tree!

PR #1646, opened 2024-03-06
userlib: add a critical-section impl for task code.

cbiffle:
The embedded ecosystem is starting to catch up with the fact that not all of us are running a single address space in privileged mode, and crates like `portable-atomics` may now be in a state where we could rely on them.

Implementing things like atomic polyfills on platforms without atomics requires some way to control concurrency, and crates seem to be standardizing on the `critical-section` crate -- which unlike the old critical section in `cortex-m`, can be customized for our purposes.

So, this commit adds a backing implementation to `userlib` to allow the `critical-section` crate to be used in tasks. I'm slightly alarmed to note that our dependency graph already includes a very old version of `critical-section`, but I'm _pretty sure_ that should be safe since we've never provided an impl for it. It is presumably not being linked in.

I need to do more work to evaluate the code generation of crates like `portable-atomics`, but this part, at least, I think we can probably land.

### DIFF
--- a/sys/userlib/src/critical_section.rs
+++ b/sys/userlib/src/critical_section.rs
@@ -1,0 +1,33 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! An implementation to support the `critical-section` crate within a Hubris
+//! user task.
+//!
+//! Hubris is careful never to introduce non-local or asynchronous control flow
+//! into a program, and doesn't have threads. This means that, within the
+//! context of a task, we don't actually need to generate any code to implement
+//! a critical section --- they happen naturally.
+//!
+//! You might want to opt out of this implementation if you're doing something
+//! very weird with shared memory. By default, it is hard to do that in Hubris,
+//! but it is possible, and we'll assume you know what you've signed yourself up
+//! for should you choose to do it.
+
+use critical_section::RawRestoreState;
+
+struct HubrisCriticalSection;
+critical_section::set_impl!(HubrisCriticalSection);
+
+unsafe impl critical_section::Impl for HubrisCriticalSection {
+    #[inline(always)]
+    unsafe fn acquire() -> RawRestoreState {
+        // No action required.
+    }
+
+    #[inline(always)]
+    unsafe fn release(_token: RawRestoreState) {
+        // Again, no action required.
+    }
+}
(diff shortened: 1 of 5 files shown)


sys/userlib/src/critical_section.rs
```diff
@@ -0,0 +1,29 @@
+//! An implementation to support the `critical-section` crate within a Hubris
+//! user task.
+//!
+//! Hubris is careful never to introduce non-local or asynchronous control flow
+//! into a program, and doesn't have threads. This means that, within the
+//! context of a task, we don't actually need to generate any code to implement
+//! a critical section -- they happen naturally.
```
hawkw, line 11: _very_ tiny nit: in markdown, an em dash is `---`; i believe a `--` just renders as two en dashes.
```suggestion
//! a critical section --- they happen naturally.
```

sys/userlib/src/critical_section.rs
```diff
@@ -0,0 +1,29 @@
...
+//!
+//! You might want to opt out of this implementation if you're doing something
+//! very weird with shared memory. By default, it is hard to do that in Hubris,
+//! but it is possible, and we'll assume you know what you've signed yourself up
+//! for should you choose to do it.
```
hawkw, line 16: should this documentation explain _how_ you would opt out of this `critical_section` implementation if you're doing something weird? or is the idea that not telling users how to do that will discourage them from doing it? :)

hawkw (approved):
looks good to me --- i'm excited for the subsequent PR to get rid of `armv6m-atomic-hack`!

sys/userlib/src/critical_section.rs
cbiffle: Fixed!

sys/userlib/src/critical_section.rs
cbiffle: I am undecided tbh; it's a matter of disabling the `critical-section` feature on `userlib`, but I think the likelihood that we'll ever have to do it is pretty low.

hawkw:
it occurs to me that a _very_ nice side benefit of Hubris' use of separate compilation for tasks is that we _can_ allow an individual task to opt in/out of global `userlib` behaviors using Cargo feature flags --- if we had instead implemented tasks as a bunch of crates that were all dependencies of a single top-level build, feature unification would mean that _one_ task opting out of the default `critical_section` impl wouldn't work, as other tasks would still be enabling the feature flag. it might be worth discussing this in the benefits of the separate compilation model in docs/talks!

sys/userlib/src/critical_section.rs
hawkw: yeah, i think it's probably obvious enough that you don't need to explain it, especially if we build the docs with the Rustdoc feature that shows required feature flags...

cbiffle:
> it might be worth discussing this in the benefits of the separate compilation model in docs/talks!

@hawkw fwiw, this is also why we tend to appear in discussions of changing Cargo's feature unification behavior -- this is great behavior for our purposes, but is really annoying for some other people!

Explaining the use case publicly would probably help, you're right.

PR #1647, opened 2024-03-06
counters: allow deriving `Count` for generic types

hawkw:
Currently, placing the `#[derive(Count)]` attribute on a generic `enum`
doesn't compile. This branch changes the `#[derive(Count)]` code
generation to support generic types. If a type parameter is used as a
field with the `#[count(children)]` attribute, the generated counters
struct will also be generic over that type parameter, and contain the
counters type for that type's `Count` implementation, with a `where`
clause requiring that type to also implement `Count`. If the type
parameter is *not* used in a field that has the `#[count(children)]`
attribute, the generated counters struct needn't be generic over that
type parameter,[^error] so we exclude it from the counter's struct list
of type params. This required a bit of a complex change to the code
generation to implement.

I've added a new example of deriving `Count` for generic types,
including one that's generic over type parameters that _don't_ get used
in a `#[count(children)]` field. This example compiles, verifying that
we now generate code that makes sense when deriving `Count` for generic
enums.

[^error]: and, in fact, if the counters struct is generic over a type
    parameter that it doesn't actually use, it won't compile, as the
    type parameter is unused.

### DIFF
--- a/lib/counters/derive/src/lib.rs
+++ b/lib/counters/derive/src/lib.rs
@@ -6,6 +6,7 @@ extern crate proc_macro;
 use proc_macro::TokenStream;
 use proc_macro2::{Ident, Span};
 use quote::{quote, ToTokens};
+use std::collections::HashSet;
 use syn::{
     parse::{Parse, ParseStream},
     parse_macro_input, DeriveInput,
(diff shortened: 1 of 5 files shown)


lib/counters/derive/src/lib.rs
```diff
@@ -51,60 +52,89 @@ fn gen_count_impl(input: DeriveInput) -> Result<impl ToTokens, syn::Error> {
...
         let vis = &input.vis;
-
         let counts_ty = counts_ty(enum_name);
+        let where_clauses = where_clause_types
+            .iter()
+            .map(|ty| {
+                quote! { #ty: counters::Count }
+            })
+            .collect::<Vec<_>>();
+        // I'm not sure why we have to do this, but quote gets mad if it's not a vec...
```
cbiffle, line 122: FWIW I've also run into this.

cbiffle (approved):

lib/counters/derive/src/lib.rs
hawkw: Yeah, as far as I understand it, it just needs to be `IntoIterator`, but it seems like the `HashSet`'s `IntoIterator` implementation doesn't meet some required bound in `quote` somewhere...

PR #1649, opened 2024-03-07
gimlet-seq: record sequencer T6 state on power off

### LINKED ISSUE #1648: gimlet: capture sequencer T6 state on power off
There are several occasions when the T6 fails to come up successfully and therefore is powered off. When this happens it'd be helpful if the sequencer task captured and at the minimum right now tossed in the ringbuf as data for the power off transition prior to issuing the command to the sequencer. I think the registers we want are:

* `IFR`
* `NIC_CTRL`
* `NIC_STATUS`
* `OUT_STATUS_NIC1`
* `OUT_STATUS_NIC2`
* `AMD_STATUS`

I believe these'll help us understand where this is going on and how it fails.

hawkw:
As described in #1648, when the NIC fails to come up successfully and is
powered off, it's desirable to capture sequencer register state
associated with the NIC and stick them in the ringbuf before actually
commanding the sequencer to power off the NIC, in order to debug NIC
failures.

This branch changes the `gimlet-seq-server` task to record the values of
the `NIC_CTRL`, `NIC_STATUS`, `OUT_STATUS_NIC1` and `OUT_STATUS_NIC2`
registers in the ringbuf when `NIC_PWREN_L` is asserted high while in
the `A0PlusHP` state. Issue #1648 also requested that the `IFR` and
`AMD_STATUS` registers be recorded in the ringbuf when this occurs, but
those registers should already be present in the `Trace::Status` entry
that's recorded every time the sequencer task handles a notification,
immediately prior to checking if the task must transition the NIC's
state. So, I chose not to add additional ringbuf events for those
registers to avoid pushing other stuff off the end of the ringbuf.

Fixes #1648

### DIFF
--- a/drv/gimlet-seq-server/src/main.rs
+++ b/drv/gimlet-seq-server/src/main.rs
@@ -72,7 +72,6 @@ enum Trace {
     A0Status(u8),
     A0Power(u8),
     NICPowerEnableLow(bool),
-    NICDetails(Addr, u8),
     RailsOn,
     UartEnabled,
     A0(u16),
@@ -94,6 +93,12 @@ enum Trace {
     SMStatus {
         a1: u8,
         a0: u8,
+    },
+    NICStatus {
+        nic_ctrl: u8,
+        nic_status: u8,
+        out_status_nic1: u8,
+        out_status_nic2: u8,
     },
     ResetCounts {
         rstn: u8,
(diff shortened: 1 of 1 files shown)


rmustacc:
Is it easy for you to provide an example of the ringbuf data? This should be generated natively by booting the gimlet and the t6init service.

hawkw:
> Is it easy for you to provide an example of the ringbuf data? This should be generated natively by booting the gimlet and the t6init service.

Sure, one moment!

hawkw:
If you think it would be useful, we can also record the NIC registers _any_ time `NIC_PWREN_L` changes, rather than just when going high in the `A0PlsuHP` state?

cbiffle (approved):

rmustacc (approved):
Thanks for this, @hawkw.

> If you think it would be useful, we can also record the NIC registers any time NIC_PWREN_L changes, rather than just when going high in the A0PlsuHP state?

I don't think (famous last words), it's really necessary. The main thing we need when we're powering it off is whether or not we thought it successfully got out of reset. Realistically longer term (and not now) if we try to "power on" the nic aka clear reset because all the power is already on, then we shouldn't even proceed if we don't have power good and handwave handwave, generate an alert and trigger the power fault signals. But that's stuff for another time.

PR #1651, opened 2024-03-07
counters: add support for counting `bool`s

hawkw:
Currently, we have a bunch of ringbuf enum variants with single `bool` fields, like `Trace::NICPowerEnableLow(bool)` in `gimlet-seq-server`. This commit adds a `counters::Count` implementation for `bool`, allowing us to place `#[count(children)]` on these variants and generate separate `true` and `false` counters for them.


mkeeter (approved):
Neat!

PR #1652, opened 2024-03-07
Make error types in the net API more precise

cbiffle:
Since its original implementation, the netstack has had some error variants that were... strange. Some of them clearly indicate programming errors (like `NotYours`, which happens if you attempt to send through somebody else's socket), and others clearly indicate...well, nothing, because there was one called `Other`.

I tracked down what was generating these in practice and converted most of the cases to reply-fault, since they represented programmer errors that broke preconditions. I've cleaned up and commented the handling of the remaining error conditions in all clients.

### DIFF
--- a/task/dump-agent/src/udp.rs
+++ b/task/dump-agent/src/udp.rs
@@ -124,16 +124,11 @@ impl ServerImpl {
                 meta,
                 &tx_data_buf[..meta.size as usize],
             ) {
-                // We'll drop packets if the outgoing queue is full or the
-                // server has died; the host is responsible for retrying.
-                //
-                // Other errors are unexpected and panic.
                 ringbuf_entry!(Trace::SendError(e));
                 match e {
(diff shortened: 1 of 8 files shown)


task/dump-agent/src/udp.rs
```diff
@@ -124,16 +123,11 @@ impl ServerImpl {
...
                 &tx_data_buf[..meta.size as usize],
             ) {
-                // We'll drop packets if the outgoing queue is full or the
-                // server has died; the host is responsible for retrying.
-                //
-                // Other errors are unexpected and panic.
                 ringbuf_entry!(Trace::SendError(e));
                 match e {
+                    // We'll drop packets if the outgoing queue is full or the
+                    // server has died; the host is responsible for retrying.
```
mkeeter, line 129: I'm not sure if we actually have a retry implementation in `humility`, but this change doesn't make things any worse.

PR #1656, opened 2024-03-13
spi: A reSPIte from excessive error codes

hawkw:
Currently, the SPI API has a single big error enum which it returns from
all IPCs. This isn't all that great: for example, a `NothingToRelease`
error can be returned from the `Spi::read` IPC, which...doesn't make a
ton of sense. Additionally, many of these error variants are returned
only in the event of a programmer error in a client task, rather than
expected runtime errors, and probably should be `reply-fault`s rather
than error codes.

Per @cbiffle on Matrix:

> Re: the SPI errors, we've got
>
> - `TaskRestarted` - sure makes sense, also probably not appropriate to
>   unwrap though
> - `NothingToRelease` - should be reply-fault, indicates that you're
>   acting like you've locked the device when you haven't.
> - `BadDevice` - should be reply-fault, indicates that you've made up a
>   device index, that's what config is for
> - `BadTransferSize` - this one's subtle. The max transfer is
>   platform-specific, and there's currently no way to discover it. We
>   could treat this as a precondition and reply-fault it, but code that
>   wants to be diligent and check preconditions currently has no way to
>   do so! We could also define a common max value in the API and insist
>   upon it on all platforms, at which point this would be reply-fault.

Therefore, I've refactored the `SpiError` type to remove the
`NothingToRelease` and `BadDevice` variants, and changed the `Spi` IPC
interface so that the `lock` and `release` APIs only return runtime
errors on server death --- the `SpiError` type is now only returned by
`Spi::read`, `Spi::write`, and `Spi::exchange`. In situations where
`BadDevice` or `NothingToRelease` errors would previously have been
returned, the SPI server now sends a reply-fault instead.

The SPI API is a bit different from other driver APIs, as there's a
provision for running the SPI driver locally when a task has exclusive
ownership over a specific SPI bus, as well as running it as its own
driver task. In this case, the cases which would return reply-fault when
the driver is its own task now panic when the SPI interface's driver is
local to another task.

This reduces the total archive size for `gimlet/rev-f.toml` by 1%
(according to `cargo xtask sizes`), which is, honestly, not that much.
But, it still seems worthwhile.

### DIFF
--- a/drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
+++ b/drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
@@ -39,8 +39,6 @@ impl From<Error> for u8 {
             Error::SpiError(e) => match e {
                 SpiError::BadTransferSize => 3,
                 SpiError::TaskRestarted => 4,
-                SpiError::NothingToRelease => 5,
-                SpiError::BadDevice => 6,
             },
             Error::I2cError(e) => 8 + (e as u8),
         }
@@ -254,7 +252,7 @@ impl<'a, S: SpiServer> Ecp5Driver for DeviceInstance<'a, S> {
             .config
             .configuration_port
             .lock(spi_api::CsState::Asserted)
-            .map_err(Self::Error::from)
+            .map_err(|_| Self::Error::from(SpiError::TaskRestarted))
     }
 
     fn configuration_release(&self) -> Result<(), Self::Error> {
@@ -262,7 +260,7 @@ impl<'a, S: SpiServer> Ecp5Driver for DeviceInstance<'a, S> {
             .config
             .configuration_port
             .release()
-            .map_err(Self::Error::from)
+            .map_err(|_| Self::Error::from(SpiError::TaskRestarted))
     }
 }
 
@@ -316,10 +314,14 @@ impl<'a, S: SpiServer> FpgaUserDesign for DeviceInstance<'a, S> {
             .config
             .user_design
             .lock(spi_api::CsState::Asserted)
-            .map_err(Into::into)
+            .map_err(|_| SpiError::TaskRestarted.into())
     }
 
     fn user_design_release(&self) -> Result<(), FpgaError> {
-        self.driver.config.user_design.release().map_err(Into::into)
+        self.driver
+            .config
+            .user_design
+            .release()
+            .map_err(|_| SpiError::TaskRestarted.into())
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2770,6 +2770,7 @@ name = "ksz8463"
 version = "0.1.0"
 dependencies = [
  "drv-spi-api",
+ "idol-runtime",
  "num-traits",
  "ringbuf",
  "userlib",
--- a/drv/fpga-devices/src/ecp5_spi.rs
+++ b/drv/fpga-devices/src/ecp5_spi.rs
@@ -41,8 +41,6 @@ impl From<Ecp5UsingSpiError> for u8 {
             Ecp5UsingSpiError::SpiError(e) => match e {
                 SpiError::BadTransferSize => 3,
                 SpiError::TaskRestarted => 4,
-                SpiError::NothingToRelease => 5,
-                SpiError::BadDevice => 6,
             },
         }
     }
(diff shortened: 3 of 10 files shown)


andrewjstone:
> This looks pretty great @hawkw. Unfortunately, we plumbed up the `GwSpiError` to [MGS](https://github.com/oxidecomputer/management-gateway-service/blob/aa9bdaa814811c450c4ee163cf5a1815d3d383b3/gateway-messages/src/sp_to_mgs.rs#L795-L815) and so changing the discriminant values may result in some weird errors.

Hmm, Maybe it's fine, because of the `match` in the `From`.

drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
```diff
@@ -254,15 +252,15 @@ impl<'a, S: SpiServer> Ecp5Driver for DeviceInstance<'a, S> {
             .config
             .configuration_port
             .lock(spi_api::CsState::Asserted)
-            .map_err(Self::Error::from)
+            .map_err(|_| Self::Error::from(SpiError::TaskRestarted))
```
mkeeter, line 255: This is confusing at first glance – it's hard to know that the argument is `ServerDeath`, so it looks like it's discarding information.

What about adding `impl From<idol_runtime::ServerDeath> for SpiError` in `spi-api`, then doing `map_err(SpiError::from)` here (and everywhere else the pattern appears)?

hawkw: I was iffy about having an `impl From<idol_runtime::ServerDeath> for SpiError`, because that means that _any_ time we `?` a `ServerDeath` error, it _could_ be turned into a `SpiError` --- if we have some code that calls multiple APIs which can return `ServerDeath` errors, I didn't want them to accidentally get turned into a `SpiError` due to type inference.

hawkw:
> This looks pretty great @hawkw. Unfortunately, we plumbed up the `GwSpiError` to [MGS](https://github.com/oxidecomputer/management-gateway-service/blob/aa9bdaa814811c450c4ee163cf5a1815d3d383b3/gateway-messages/src/sp_to_mgs.rs#L795-L815) and so changing the discriminant values may result in some weird errors.

This PR doesn't actually change the definition of `GwSpiError` at all (it's not defined in Hubris, so it _can't_...); it just changes the discriminant values of the  Hubris types that are converted into `GwSpiError`s. The actual `GwSpiError`s shouldn't have changed at all, so if I understand correctly, this is fine.

I can also change the discriminants for the Hubris error enums back, if you like, but (AFAICT) it shouldn't make a difference?

jgallagher:
> > This looks pretty great @hawkw. Unfortunately, we plumbed up the `GwSpiError` to [MGS](https://github.com/oxidecomputer/management-gateway-service/blob/aa9bdaa814811c450c4ee163cf5a1815d3d383b3/gateway-messages/src/sp_to_mgs.rs#L795-L815) and so changing the discriminant values may result in some weird errors.
> 
> Hmm, Maybe it's fine, because of the `match` in the `From`.

FWIW (maybe less than nothing), this is why we landed on "duplicate the types and implement `From`". We should be able to make changes on the hubris side and find out in the `From` impl whether they're compatible without having to worry about MGS getting confused.

andrewjstone:
> > This looks pretty great @hawkw. Unfortunately, we plumbed up the `GwSpiError` to [MGS](https://github.com/oxidecomputer/management-gateway-service/blob/aa9bdaa814811c450c4ee163cf5a1815d3d383b3/gateway-messages/src/sp_to_mgs.rs#L795-L815) and so changing the discriminant values may result in some weird errors.
> 
> This PR doesn't actually change the definition of `GwSpiError` at all (it's not defined in Hubris, so it _can't_...); it just changes the discriminant values of the Hubris types that are converted into `GwSpiError`s. The actual `GwSpiError`s shouldn't have changed at all, so if I understand correctly, this is fine.
> 
> I can also change the discriminants for the Hubris error enums back, if you like, but (AFAICT) it shouldn't make a difference?

Nope, it's all good! Leave it as is. As @jgallagher reminded me, we did this on purpose :D

drv/fpga-devices/src/ecp5_spi_mux_pca9538.rs
mkeeter: Interesting, I understand this downside but am having trouble imagining cases where that would be the wrong thing to do.

Rephrasing for my own understanding:

1) We've got a function that returns a `Result<T, SpiError>`
2) That function calls one or more Idol APIs which return `ServerDeath`
3) ...and someone puts a `?` on one of those calls
4) ...but we don't want to return `SpiError::TaskRestarted`; instead, we want a compile-time error saying that "you messed up"

In particular, needing both 3 and 4 to be true seems weird to me.

(I poked at doing `|_: idol_runtime::ServerDeath| SpiError::TaskRestarted`, but it's needlessly verbose and adds a dependency, so I think that's not great either)

Anyways, I don't feel too strongly about this!

hawkw: Oh, I was thinking more about a case where another error type implements `From<T> where T: Into<SpiError>` and also has another case that might be convertable from a `ServerDeath` error...but I'm not actually sure if such a case exists, and I think the existing blanket impls for `From` might _prevent_ us from ever being able to do that. I think I've convinced myself that it doesn't matter.

hawkw: okay, i've addressed this suggestion in c7c49f932a58a2fecb42b82cdaee23e157df311d!

cbiffle (approved):

PR #1657, opened 2024-03-14
prevent I2C from hanging on lost interrupt

bcantrill:
Draft commit message:

```
This addresses #1631 by adding a (long) timeout whenever we wait for
an interrupt from the I2C controller.  This timeout shouldn't ever
fire:  if it does, something seriously is amiss, so we record more or
less everything we can out of the I2C controller and explicitly panic.
```

### DIFF
--- a/drv/stm32xx-i2c/src/lib.rs
+++ b/drv/stm32xx-i2c/src/lib.rs
@@ -80,7 +80,7 @@ pub struct I2cTimeout(pub u64);
 
 pub enum I2cControlResult {
     Interrupted,
-    TimedOut
+    TimedOut,
 }
 
 ///
@@ -159,6 +159,7 @@ pub enum ReadLength {
     Variable,
 }
 
+#[allow(clippy::upper_case_acronyms)]
 #[derive(Copy, Clone, Eq, PartialEq)]
 enum Register {
     CR1,
@@ -203,7 +204,7 @@ enum Trace {
     None,
 }
 
-counted_ringbuf!(Trace, 48, Trace::None);
+ringbuf!(Trace, 48, Trace::None);
 
 impl I2cMux<'_> {
     /// A convenience routine to translate an error induced by in-band
@@ -496,7 +497,7 @@ impl I2cController<'_> {
     /// has gone wrong with a device (which should rather be indicated by
     /// returning an error and resetting the controller if/as needed), but with
     /// the controller itself.
-    /// 
+    ///
     fn panic(&self) -> ! {
         let i2c = self.registers;
         let tgr = &i2c.timingr;
@@ -542,7 +543,7 @@ impl I2cController<'_> {
                 ringbuf_entry!(Trace::LostInterrupt);
                 self.panic();
             }
-            I2cControlResult::Interrupted => Ok(())
+            I2cControlResult::Interrupted => Ok(()),
         }
     }
 
--- a/drv/stm32xx-i2c/src/lib.rs
+++ b/drv/stm32xx-i2c/src/lib.rs
@@ -74,6 +74,16 @@ pub struct I2cController<'a> {
 }
 
 ///
+/// A structure to denote an absolute number of ticks to wait.
+///
+pub struct I2cTimeout(pub u64);
+
+pub enum I2cControlResult {
+    Interrupted,
+    TimedOut
+}
+
+///
 /// A structure that defines interrupt control flow functions that will be
 /// used to pass control flow into the kernel to either enable or wait for
 /// interrupts.  Note that this is deliberately a struct and not a trait,
@@ -82,6 +92,7 @@ pub struct I2cController<'a> {
 pub struct I2cControl {
     pub enable: fn(u32),
     pub wfi: fn(u32),
+    pub wfi_or_timeout: fn(u32, I2cTimeout) -> I2cControlResult,
 }
 
 #[derive(Copy, Clone, Eq, PartialEq)]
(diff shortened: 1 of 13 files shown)


drv/stm32xx-i2c/src/lib.rs
```diff
@@ -466,6 +492,62 @@ impl I2cController<'_> {
...
+        let tor = &i2c.timeoutr;
+
+        ringbuf_entry!(Trace::Panic(Register::CR1, i2c.cr1.read().bits()));
+        ringbuf_entry!(Trace::Panic(Register::CR2, i2c.cr2.read().bits()));
+        ringbuf_entry!(Trace::Panic(Register::OAR1, i2c.oar1.read().bits()));
+        ringbuf_entry!(Trace::Panic(Register::OAR2, i2c.oar2.read().bits()));
+        ringbuf_entry!(Trace::Panic(Register::TIMINGR, tgr.read().bits()));
+        ringbuf_entry!(Trace::Panic(Register::TIMEOUTR, tor.read().bits()));
+        ringbuf_entry!(Trace::Panic(Register::ISR, i2c.isr.read().bits()));
+        ringbuf_entry!(Trace::Panic(Register::PECR, i2c.pecr.read().bits()));
```
hawkw, line 513: for my own education, what's the motivation for recording each register as a separate ringbuf entry rather than one entry with all of them? something like this:
```rust
enum Trace {
    // ...
    Panic {
         cr1: u32,
         cr2: u32,
         oar1: u32,
         oar2: u32,
         // ...
    },
    // ...
}
```
it seems like recording each register as its own entry occupies more slots in the ringbuf, potentially pushing stuff off the end that might be useful for diagnosing the panic?

i'm guessing that the motivation for this is that keeping the entry type itself small makes the ringbuf as a whole smaller, and that's worth the risk of potentially losing earlier data? is that correct?

drv/stm32xx-i2c/src/lib.rs
```diff
@@ -148,33 +159,48 @@ pub enum ReadLength {
     Variable,
 }
 
+#[allow(clippy::upper_case_acronyms)]
 #[derive(Copy, Clone, Eq, PartialEq)]
```
hawkw, line 163: we may want to derive `Count` for this as well, and then add `#[count(children)]` to the events that include a `Register`?
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, counters::Count)]
```

task/spd/src/main.rs
```diff
@@ -256,6 +256,9 @@ fn main() -> ! {
         wfi: |notification| {
             let _ = sys_recv_closed(&mut [], notification, TaskId::KERNEL);
         },
+        wfi_or_timeout: |_notification, _timeout| {
```
mkeeter, line 259: The divergence between I2C controller and target seems significant here, to the point where a new type may be warranted.  My modest proposal:
- Add a new `I2cTarget` vtable type which only includes `enable` and `wfi`
  - Use this type in `operate_as_target`
- Remove `wfi` from the `I2cControl`, so you have to use `wfi_or_timeout`

drv/stm32xx-i2c/src/lib.rs
```diff
@@ -148,33 +159,48 @@ pub enum ReadLength {
...
-    AddrISR(u32),
+    Wait(Register, u32),
+    Write(Register, u32),
+    WriteWait(Register, u32),
+    Read(Register, u32),
+    ReadWait(Register, u32),
+    KonamiOperation(I2cKonamiCode),
+    Konami(Register, u32),
+    Reset(Register, u32),
+    Addr(Register, u32),
```
hawkw, line 185: if we want separate counts for each register we read/write to and we add a derived `Count` impl for `Register`, as I suggested above, we could add:
```suggestion
    Wait(#[count(children)] Register, u32),
    Write(#[count(children)] Register, u32),
    WriteWait(#[count(children)] Register, u32),
    Read(#[count(children)] Register, u32),
    ReadWait(#[count(children)] Register, u32),
    KonamiOperation(I2cKonamiCode),
    Konami(#[count(children)] Register, u32),
    Reset(#[count(children)] Register, u32),
    Addr(#[count(children)] Register, u32),
```

drv/stm32xx-i2c/src/lib.rs
```diff
@@ -148,33 +159,48 @@ pub enum ReadLength {
...
     AddrMatch,
     AddrNack(u8),
+    RxReg(Register, u32),
     Rx(u8, u8),
     RxNack(u8, u8),
     Tx(u8, u8),
     TxBogus(u8),
     TxOverrun(u8),
-    TxISR(u32),
+    TxReg(Register, u32),
```
hawkw, line 194: and similarly, do we want to add `#[count(children)]` here as well?

drv/stm32xx-i2c/src/lib.rs
```diff
@@ -148,33 +159,48 @@ pub enum ReadLength {
...
+    CR2,
+    OAR1,
+    OAR2,
+    TIMINGR,
+    TIMEOUTR,
+    ISR,
+    PECR,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, counters::Count)]
```
hawkw, line 175: now that this derives `Count`, we also need to change this like: https://github.com/oxidecomputer/hubris/blob/014b2a7fdd96506cbffe494bfb01adb26bc90598/drv/stm32xx-i2c/src/lib.rs#L207 to `counted_ringbuf!`

drv/stm32xx-i2c/src/lib.rs
bcantrill: Yes, all about keeping the entry size small to maximize the number of entries. There are other cases (e.g., recording the MAX5970 state) where there isn't much variance in the size of the entry, so we record all of the state in a single entry.

drv/stm32xx-i2c/src/lib.rs
bcantrill: We actually can't do that because there is no way (currently) to disable counted ringbufs entirely -- which we need to be able to do to fit in the Donglet.

drv/stm32xx-i2c/src/lib.rs
bcantrill: I actually don't want to count all children there -- it's deliberate, even if/when we have `counted_ringbuf` here (which we can't right now per my other comment).

drv/stm32xx-i2c/src/lib.rs
bcantrill: Deliberate to not do this; see my other comments.

drv/stm32xx-i2c/src/lib.rs
bcantrill: Similarly, no. ;)

drv/stm32xx-i2c/src/lib.rs
hawkw: cool, that's what i figured, but i wanted to check. thanks!

drv/stm32xx-i2c/src/lib.rs
hawkw: Ah, yup. I can go add that!

task/spd/src/main.rs
bcantrill: Has a proposal ever been so modest and reasonable?!  Doing that now...

drv/stm32xx-i2c/src/lib.rs
hawkw: Ah, it looks like, currently, the `ringbuf` crate's `counters` dependency _is_ optional, but that's not quite what we want, as I've learnt --- it's preferable to be able to opt _out_ of instrumentation on small targets rather than opt in on all the other targets; thus, "`ringbuf/disabled"`. I'll add a similar feature flag for disabling counts as well.

task/spd/src/main.rs
bcantrill: Done!

mkeeter (approved):

drv/stm32xx-i2c/src/lib.rs
hawkw: Okay, #1658 will let you disable both the actual ringbuf and counters for ringbufs, so we can add a `counted_ringbuf!` here and still fit in the STM32G031's RAM.

hawkw: Merged #1658, so we can make this counted now!

PR #1660, opened 2024-03-15
i2c_driver should use a counted ringbuf

bcantrill:


hawkw (approved):
Looks good to me!

We might want to add `#[count(children)]` to [`Trace::RepeatedStart`](https://github.com/oxidecomputer/hubris/blob/ee9d3af7411e248b06677c5458f5fe7670ddcd37/drv/stm32xx-i2c/src/lib.rs#L204) so that `true` and `false` entries are counted separately?

bcantrill:
> Looks good to me!
> 
> We might want to add `#[count(children)]` to [`Trace::RepeatedStart`](https://github.com/oxidecomputer/hubris/blob/ee9d3af7411e248b06677c5458f5fe7670ddcd37/drv/stm32xx-i2c/src/lib.rs#L204) so that `true` and `false` entries are counted separately?

We can spare the 4 bytes! `cargo xtask sizes` before (for Gimlet):

```
i2c_driver           flash   16152   16384   16384
                     ram     4048    4096    4096
```

And after the `count(children)` for `Trace::RepeatedStart`:

```
i2c_driver           flash   16228   16384   16384
                     ram     4052    4096    4096
```

drv/stm32xx-i2c-server/Cargo.toml
```diff
@@ -26,10 +26,37 @@ build-util = { path = "../../build/util" }
...
+    "stm32g0/stm32g030",
+    "drv-stm32xx-i2c/g030",
+    "drv-stm32xx-sys-api/g030",
+    "build-i2c/g030",
+    "ringbuf-disabled",
+]
+
+ringbuf-disabled = ["ringbuf/disabled", "ringbuf/counters-disabled"]
 panic-messages = ["userlib/panic-messages"]
 no-ipc-counters = ["idol/no-counters"]
```
hawkw, line 61: It occurs to me that, since some of the target features disable ringbufs and counters, we should maybe just have the features for those targets also disable IPC counters, rather than explicitly setting the `no-ipc-counters` feature in the app.toml for those targets...but that's something to do separately.

hawkw (approved):
Looks good to me!

PR #1661, opened 2024-03-15
kern: add `IRQ_STATUS` syscall

### LINKED ISSUE #1659: Need a way for a task to _check_ its interrupt state
Currently, tasks can enable/disable interrupts they own. Tasks refer to interrupts by using their own notification bit space; hardware interrupts are mapped into that notification bit space. This is nice because it prevents a task from even _referring to_ another task's interrupts. For more on this, start at [the "Interrupts from a task's perspective" section of the reference guide](https://hubris.oxide.computer/reference/#_interrupts_from_a_tasks_perspective).

Over in #1631 we're having a case where the I2C subsystem appears to lose an event. It winds up just parked forever waiting for an interrupt that never comes. I posted some comments there suggesting ways to disambiguate the various possible causes, and @bcantrill has landed code in the I2C driver to track the in-task versions of that. The remaining possibility is that, due to a logic error or whatever, the driver has not correctly configured its interrupts in the kernel. It would be really nice to record this fact in the driver's crash dump (now generated when the driver runs into this situation), but crash dumps only record the task state, and this state isn't in the task.

After debating a few possible mechanisms in chat, the approach that I personally am most comfortable with is: **add a new syscall that enables a task to query its interrupt state.** The I2C driver (or others!) could then use this to insert it into e.g. a ringbuf for diagnostics.

This operation is nonprivileged, since it only allows a task to inquire about its own interrupts, and seems pretty general, so I think it's worth making it a syscall rather than a kipc or a specialized debugging hack.

The syscall should probably take a notification mask as its argument, just like the existing `IRQ_CONTROL` syscall does, and would probably use the same code as `IRQ_CONTROL` to map that to an interrupt.

To support debugging #1631 any such syscall needs to distinguish between the following possible states:
- Enabled or not enabled. This indicates whether the interrupt is permitted by hardware to trigger an ISR, which would result in the associated notification being posted to the owning task. On ARM M-profile this corresponds to a bit in the NVIC ISERx register (for x in 0..=15).
- Pending or not pending. This indicates whether the interrupt has been generated but not yet triggered an ISR. Pending-but-not-enabled is one of the cases we're curious about in #1631. On ARM M-profile this corresponds to a bit in the NVIC ISPRx register (for x in 0..=15)
- Posted or not posted. This is a Hubris-specific thing: has the interrupt fired but the task hasn't noticed yet? This can happen if the task hasn't bothered to enter RECV with an appropriate notification mask. This can be determined by checking the bitwise AND of the notification mask argument with the task's `notifications` field; if any bits are set in both, the interrupt has been posted.

These are orthogonal, so they could potentially be returned in e.g. three bits of a word. Thus, from a Rust perspective, one possible signature for this syscall is

```rust
fn get_irq_status(mask: u32) -> IrqStatus;

struct IrqStatus(u32);
impl IrqStatus {
    fn is_enabled(self) -> bool { self.0 & 1 != 0 }
    fn is_pending(self) -> bool { self.0 & 2 != 0 }
    fn is_posted(self) -> bool { self.0 & 4 != 0 }
}
```

...or something to that effect.

If someone is excited about doing this, probably the best place to start is by reading the commit from the last time we added a syscall in January 2022: f0eddc9f9668cb5d5c39917576eee24cf5e1a26a

hawkw:
As described in #1659, we currently lack a mechanism for a task to query
the state of the interrupts mapped to its notification set. Because only
the kernel can touch the NVIC, such a mechanism must be added to the
kernel. 

This branch adds a new system call, `IRQ_STATUS` (syscall 13).
`IRQ_STATUS` takes a notification mask as an argument, and returns a set
of bit flags representing the state of the interrupts mapped to that
notification set:

- bit 0 is set if the interrupt has been enabled by the task (by calling
  `IRQ_CONTROL`) and unset if the interrupt has not been enabled
- bit 1 is set if an interrupt is currently pending for any task in the
  notification set
- bit 2 is set if a notification has been posted to the task for any
  interrupt in the notification set

If multiple bits are set in the notification mask, and all of those
notification bits are set to interrupts, the returned `IrqState`
bitflags will be the boolean OR of the states of all the interrupts.

## Design Notes

We also considered hanging this off of the existing `IRQ_CONTROL` system
call, which currently takes an argument of either 0 or 1 to indicate
whether to enable or disable interrupts. We could, potentially, have
added the ability to pass 2 as well to query IRQ status. However, that
system call currently doesn't return anything, so adding a return value
would make it spill more registers, and @cbiffle liked that the argument
to it was essentially a bool. Thus, we decided to add a new system call.

I also considered a design where `IRQ_STATUS' returns three words, which
are bitmaps of which individual notifications in the notification set
are enabled, pending, and posted, respectively. This would let us return
more information when multiple bits are set in the notification mask.
However, it has a couple of downsides:
 - We would have to spill more registers in the `IRQ_STATUS` syscall
stub, a cost which is inflicted on callers who only want to look up the
status of a single IRQ
 - Currently, the generated lookup table of IRQs can't be queried for
   individual bits, only for the whole mask. We'd need to change how we
   generate that LUT to support this. This seemed kinda annoying.

## Testing

In order to test this, I've also added a new KIPC operation to trigger a
software IRQ, given a task ID and notification mask. This is a privilege
operation that can only be called by the supervisor, so I've added a
test-runner IPC to ask the test runner (which runs as supervisor) to
please do it. Incidentally, this has the nice side benefit of letting us
test that IRQs are even dispatched to tasks at all, which is probably
worth having!

Closes #1659

### DIFF
--- a/sys/userlib/src/lib.rs
+++ b/sys/userlib/src/lib.rs
@@ -1652,3 +1652,87 @@ unsafe extern "C" fn sys_reply_fault_stub(_tid: u32, _reason: u32) {
         }
     }
 }
+
+/// Returns the current status of any interrupts mapped to the provided
+/// notification mask.
+///
+/// # Arguments
+///
+/// - `mask`: a notification mask for interrupts mapped to the current task.
+///
+/// # Returns
+///
+/// An [`IrqStatus`] (see the `abi` crate) describing the status of the
+/// interrupts in the notification mask.
+///
+/// # Faults
+///
+/// This syscall faults the caller if the given notification bitmask is not
+/// mapped to an interrupt in this task.
+#[inline(always)]
+pub fn sys_irq_status(mask: u32) -> abi::IrqStatus {
+    let status = unsafe { sys_irq_status_stub(mask) };
+    abi::IrqStatus::from_bits_truncate(status)
+}
+
+/// Core implementation of the IRQ_STATUS syscall.
+///
+/// See the note on syscall stubs at the top of this module for rationale.
+#[naked]
+unsafe extern "C" fn sys_irq_status_stub(_mask: u32) -> u32 {
+    cfg_if::cfg_if! {
+        if #[cfg(armv6m)] {
+            arch::asm!("
+                @ Spill the registers we're about to use to pass stuff.
+                push {{r4, lr}}
+                mov r4, r11
+                push {{r4}}
+
+                @ Load the constant syscall number.
+                eors r4, r4
+                adds r4, #{sysnum}
+                mov r11, r4
+                @ Move register arguments into place.
+                mov r4, r0
+
+                @ To the kernel!
+                svc #0
+
+                @ Move result into place.
+                mov r0, r4
+
+                @ Restore the registers we used and return.
+                pop {{r4}}
+                mov r11, r4
+                pop {{r4, r5, pc}}
+                ",
+                sysnum = const Sysnum::IrqStatus as u32,
+                options(noreturn),
+            )
+        } else if #[cfg(any(armv7m, armv8m))] {
+            arch::asm!("
+                @ Spill the registers we're about to use to pass stuff.
+                push {{r4, r11, lr}}
+
+                @ Move register arguments into place.
+                mov r4, r0
+                @ Load the constant syscall number.
+                mov r11, {sysnum}
+
+                @ To the kernel!
+                svc #0
+
+                @ Move result into place.
+                mov r0, r4
+
+                @ Restore the registers we used and return.
+                pop {{r4, r11, pc}}
+                ",
+                sysnum = const Sysnum::IrqStatus as u32,
+                options(noreturn),
+            )
+        } else {
+            compile_error!("missing sys_irq_status stub for ARM profile")
+        }
+    }
+}
--- a/sys/userlib/src/lib.rs
+++ b/sys/userlib/src/lib.rs
@@ -1704,7 +1704,7 @@ unsafe extern "C" fn sys_irq_status_stub(_mask: u32) -> u32 {
                 @ Restore the registers we used and return.
                 pop {{r4}}
                 mov r11, r4
-                pop {{r4, r5, pc}}
+                pop {{r4, pc}}
                 ",
                 sysnum = const Sysnum::IrqStatus as u32,
                 options(noreturn),
--- a/test/test-suite/src/main.rs
+++ b/test/test-suite/src/main.rs
@@ -124,6 +124,8 @@ test_cases! {
     test_idol_ssmarshal,
     test_idol_ssmarshal_multiarg,
     test_idol_ssmarshal_multiarg_enum,
+    test_irq_notif,
+    test_irq_status,
     #[cfg(feature = "fru-id-eeprom")]
     at24csw080::test_at24csw080,
 }
@@ -989,15 +991,15 @@ fn test_borrow_info() {
 
             // Borrow 0 is expected to be 16 bytes long and R/W.
             let info0 = caller.borrow(0).info().unwrap();
-            assert_eq!(
-                info0.attributes,
-                LeaseAttributes::READ | LeaseAttributes::WRITE
-            );
+            // assert_eq!(
+            //     info0.attributes,
+            //     LeaseAttributes::READ | LeaseAttributes::WRITE
+            // );
             assert_eq!(info0.len, 16);
 
             // Borrow 1 is expected to be 5 bytes long and R/O.
             let info1 = caller.borrow(1).info().unwrap();
-            assert_eq!(info1.attributes, LeaseAttributes::READ);
+            // assert_eq!(info1.attributes, LeaseAttributes::READ);
             assert_eq!(info1.len, 5);
 
             caller.reply(0);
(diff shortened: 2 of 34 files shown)


hawkw:
@cbiffle I'm somewhat on the fence about this potential change --- curious about how you feel about it: 
```diff
diff --git a/sys/kern/src/arch/arm_m.rs b/sys/kern/src/arch/arm_m.rs
index 278f4721..21a4d154 100644
--- a/sys/kern/src/arch/arm_m.rs
+++ b/sys/kern/src/arch/arm_m.rs
@@ -1142,20 +1142,16 @@ pub unsafe extern "C" fn DefaultHandler() {
 pub fn disable_irq(n: u32) {
     // Disable the interrupt by poking the Interrupt Clear Enable Register.
     let nvic = unsafe { &*cortex_m::peripheral::NVIC::PTR };
-    let reg_num = (n / 32) as usize;
-    let bit_mask = 1 << (n % 32);
     unsafe {
-        nvic.icer[reg_num].write(bit_mask);
+        nvic.icer[irq_reg(n)].write(irq_bit(n));
     }
 }
 
 pub fn enable_irq(n: u32) {
     // Enable the interrupt by poking the Interrupt Set Enable Register.
     let nvic = unsafe { &*cortex_m::peripheral::NVIC::PTR };
-    let reg_num = (n / 32) as usize;
-    let bit_mask = 1 << (n % 32);
     unsafe {
-        nvic.iser[reg_num].write(bit_mask);
+        nvic.iser[irq_reg(n)].write(irq_bit(n));
     }
 }
 
@@ -1165,8 +1161,8 @@ pub fn irq_status(n: u32) -> abi::IrqStatus {
     let mut status = abi::IrqStatus::empty();
 
     let nvic = unsafe { &*cortex_m::peripheral::NVIC::PTR };
-    let reg_num = (n / 32) as usize;
-    let bit_mask = 1 << (n % 32);
+    let reg_num = irq_reg(n);
+    let bit_mask = irq_bit(n);
 
     // See if the interrupt is enabled by checking the bit in the Interrupt Set
     // Enable Register.
@@ -1183,12 +1179,18 @@ pub fn irq_status(n: u32) -> abi::IrqStatus {
 
 pub fn pend_software_irq(InterruptNum(n): InterruptNum) {
     let nvic = unsafe { &*cortex_m::peripheral::NVIC::PTR };
-    let reg_num = (n / 32) as usize;
-    let bit_mask = 1 << (n % 32);
 
     // Pend the IRQ by poking the corresponding bit in the Interrupt Set Pending
     // Register (ISPR).
-    unsafe { nvic.ispr[reg_num].write(bit_mask) };
+    unsafe { nvic.ispr[irq_reg(n)].write(irq_bit(n)) };
+}
+
+fn irq_reg(irq: u32) -> usize {
+    (irq / 32) as usize
+}
+
+fn irq_bit(irq: u32) -> u32 {
+    1 << (irq % 32)
 }
 
 #[repr(u8)]
```

sys/userlib/src/lib.rs
```diff
@@ -1652,3 +1652,87 @@ unsafe extern "C" fn sys_reply_fault_stub(_tid: u32, _reason: u32) {
...
+                @ To the kernel!
+                svc #0
+
+                @ Move result into place.
+                mov r0, r4
+
+                @ Restore the registers we used and return.
+                pop {{r4}}
+                mov r11, r4
+                pop {{r4, r5, pc}}
```
cbiffle, line 1707: This pop isn't balanced with the prologue -- remove `r5`.

test/test-suite/src/main.rs
```diff
@@ -989,15 +991,15 @@ fn test_borrow_info() {
 
             // Borrow 0 is expected to be 16 bytes long and R/W.
             let info0 = caller.borrow(0).info().unwrap();
-            assert_eq!(
-                info0.attributes,
-                LeaseAttributes::READ | LeaseAttributes::WRITE
-            );
+            // assert_eq!(
```
cbiffle, line 994: (this looks accidental, uncomment or delete the assert)

test/test-suite/src/main.rs
```diff
@@ -989,15 +991,15 @@ fn test_borrow_info() {
...
+            //     info0.attributes,
+            //     LeaseAttributes::READ | LeaseAttributes::WRITE
+            // );
             assert_eq!(info0.len, 16);
 
             // Borrow 1 is expected to be 5 bytes long and R/O.
             let info1 = caller.borrow(1).info().unwrap();
-            assert_eq!(info1.attributes, LeaseAttributes::READ);
+            // assert_eq!(info1.attributes, LeaseAttributes::READ);
```
cbiffle, line 1002: (same)

test/tests-psc/app.toml
```diff
@@ -61,16 +66,16 @@ start = true
 task-slots = ["sys"]
 
 [tasks.i2c_driver.interrupts]
-"i2c2.event" = 0b0000_0010
-"i2c2.error" = 0b0000_0010
-"i2c3.event" = 0b0000_0100
-"i2c3.error" = 0b0000_0100
-"i2c4.event" = 0b0000_1000
-"i2c4.error" = 0b0000_1000
+"i2c2.event" = "0b0000_0010"
```
cbiffle, line 69: ?

cbiffle (commented):
This is looking good! Caught a couple things below.

test/test-suite/src/main.rs
hawkw: As we discussed out of band, this was broken due to a missing PartialEq impl that I was gonna fix but hadn't yet. I've put it back.

test/tests-psc/app.toml
hawkw: this whole test app has bitrotted due to IRQs having names now, I changed these just to get it to compile so I could see if the kernel size was good.

cbiffle:
> @cbiffle I'm somewhat on the fence about this potential change --- curious about how you feel about it:

re: the interrupt number extractor functions --- if you're excited about it, by all means! But I don't think it's necessary.

cbiffle (approved):

PR #1662, opened 2024-03-15
drv/stm32h7-fmc-demo: ill-advised memory controller

cbiffle:
This is an exercise fixture for the PSRAM mode of the Flexible Memory Controller on the STM32H7, for evaluating FPGA interface options. It lets you do many ill-advised things via IPC. They all appear to work, no matter what the datasheet implies.

Good luck!

### DIFF
--- a/drv/stm32h7-fmc-demo-server/src/main.rs
+++ b/drv/stm32h7-fmc-demo-server/src/main.rs
@@ -1,0 +1,342 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! "Server" that brings up the FMC and lets you pooooke it.
+
+#![no_std]
+#![no_main]
+
+use core::convert::Infallible;
+use sys_api::{Alternate, OutputType, Peripheral, Port, Pull, Speed};
+use userlib::*;
+
+#[cfg(feature = "h743")]
+use stm32h7::stm32h743 as device;
+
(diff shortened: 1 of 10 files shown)


drv/stm32h7-fmc-demo-server/src/main.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    fn poke16(
+        &mut self,
+        _msg: &RecvMessage,
+        addr: u32,
+        value: u16,
+    ) -> Result<(), RequestError<Infallible>> {
+        let ptr = addr as *mut u16;
+        unsafe { ptr.write_volatile(value) }
```
hawkw, line 235: hahahahha, _whew_ :D

drv/stm32h7-fmc-demo-server/src/main.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+impl idl::InOrderFmcDemoImpl for ServerImpl {
+    fn peek16(
+        &mut self,
+        _msg: &RecvMessage,
+        addr: u32,
+    ) -> Result<u16, RequestError<Infallible>> {
+        let ptr = addr as *const u16;
```
hawkw, line 213: presumably pointer provenance would eventually prefer us to use [`ptr::from_exposed_addr`](https://doc.rust-lang.org/std/ptr/fn.from_exposed_addr.html) for ... all of these shenanigans, but that's not real yet...

hawkw (commented):
this is deeply disquieting :)

nathanaelhuffman (approved):
lgtm, thanks for getting this set up!

PR #1663, opened 2024-03-15
tests-stm32h7: make general to any h7 system.

cbiffle:
This uses the baseline clock configuration and does no I/O, like the original test suite did before I messed that up. This will now run on any h7 board and should eliminate the need for having per-board copies of the core-system tests. (Board-specific functionality tests, of course, need to be separate.)

### DIFF
--- a/test/tests-stm32h7/src/main.rs
+++ b/test/tests-stm32h7/src/main.rs
@@ -1,0 +1,47 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#![no_std]
+#![no_main]
+
+// We have to do this if we don't otherwise use it to ensure its vector table
+// gets linked in.
+extern crate stm32h7;
+
+use cortex_m_rt::entry;
+use drv_stm32h7_startup::{system_init, ClockConfig};
+
+#[entry]
+fn main() -> ! {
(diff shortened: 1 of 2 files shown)


test/tests-stm32h7/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+        // PLL is not used, fields below are irrelevant and contain
+        // placeholders.
+        divm: 0,
+        vcosel: device::rcc::pllcfgr::PLL1VCOSEL_A::WIDEVCO,
+        pllrange: device::rcc::pllcfgr::PLL1RGE_A::RANGE8,
+        divn: 0,
+        divp: device::rcc::pll1divr::DIVP1_A::DIV2,
+        divq: 0,
+        divr: 0,
+
```
hawkw, line 42: rustfmt's mad about this line
```suggestion
```

hawkw (approved):

sys/abi/src/lib.rs
```diff
@@ -182,7 +182,7 @@ pub struct ULease {
     pub length: u32,
 }
 
-#[derive(Copy, Clone, Debug, FromBytes)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, FromBytes)]
```
hawkw, line 185: I think the one merge conflict on this branch is just that I also added these derives in #1661, but in a different order. Whoopsie.

cbiffle: Yup! I've just removed that part of my change in favor of yours.

PR #1665, opened 2024-03-18
ci: add what's being built to job names

hawkw:
Currently, the GitHub Actions UI contains a giant list of thbings which are all called "dist-single". It would be nice if they had names that actually indicated which app was being built.

This commit does that.

### DIFF
--- a/.github/workflows/build-one.yml
+++ b/.github/workflows/build-one.yml
@@ -25,7 +25,7 @@ on:
 
 jobs:
   build-one:
-     name: dist-single
+     name: "cargo xtask dist ${{ inputs.app_name }}/${{ inputs.app_toml}}"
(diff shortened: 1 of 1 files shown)


hawkw:
Now it looks like this, which is maybe a little better?
image

jclulow:
> Now it looks like this, which is maybe a little better?
> image
> 

Could you remove the common prefix, so we could see more of the different per-job bit?

hawkw:
> Could you remove the common prefix, so we could see more of the different per-job bit?

Yeah, I think that's better, will do!

hawkw:
@labbott thanks for the review! I believe that, before this can merge, someone with admin access to the repo will need to change the required checks in the repo's branch protection settings, since the names of the jobs have changed.

PR #1666, opened 2024-03-19
I2C: trace IRQ status on panic

hawkw:
Using the `irq_status` system call I added in #1661, we can extend the
panic function added to `stm32xx-i2c` in #1657 to also include the state
of any IRQs currently mapped to its notification mask. This way, if an
I2C driver panics, we can also include the kernel's understanding of its
IRQ state in the dump.

### DIFF
--- a/app/demo-stm32h7-nucleo/app-h743.toml
+++ b/app/demo-stm32h7-nucleo/app-h743.toml
@@ -37,7 +37,6 @@ name = "drv-stm32xx-i2c-server"
 stacksize = 1048
 features = ["h743"]
 priority = 2
-max-sizes = {flash = 16384, ram = 4096}
 uses = ["i2c1", "i2c2", "i2c3", "i2c4"]
 start = true
 task-slots = ["sys"]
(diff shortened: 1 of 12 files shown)


drv/stm32xx-i2c/src/lib.rs
```diff
@@ -203,8 +203,16 @@ enum Trace {
     Stop,
     RepeatedStart(#[count(children)] bool),
     LostInterrupt,
+    #[count(skip)]
     Panic(Register, u32),
     #[count(skip)]
```
hawkw, line 208: I added `#[count(skip)]` to the ringbuf entries that are only recorded on panics, since we don't really care about counting them --- they'll either be counted exactly the number of times we record them in the panic code, or they'll never be recorded at all. So, I figured it was worth saving the 8 bytes we can shave off by not generating counters for these.

app/demo-stm32h7-nucleo/app-h743.toml
```diff
@@ -37,7 +37,6 @@ name = "drv-stm32xx-i2c-server"
 stacksize = 1048
 features = ["h743"]
 priority = 2
-max-sizes = {flash = 16384, ram = 4096}
```
hawkw, line 40: We hit the flash size limit for the I2C server task, and @cbiffle said it would be fine to just remove the limits so that we don't have to keep playing guess-a-number forever.

cbiffle (approved):

PR #1667, opened 2024-03-19
ci: add test apps to `build-boards.yml`

hawkw:
Currently, Hubris' CI doesn't build the tests at all. This means that it's quite easy for the various test apps to bitrot --- for example, the PSC test app doesn't build, because [it maps IRQs by bit, rather than by name][1]. Named IRQ support was added back in 2022 (e.g. #424), so this TOML clearly hasn't been touched in...kind of a while, which I think is a pretty convincing argument for building these tests on CI!

[1]: https://github.com/oxidecomputer/hubris/blob/3325c6343a3e51a4d2c63cb3eac46085ba61dac3/test/tests-psc/app.toml#L68-L74

### DIFF
--- a/.github/workflows/build-boards.yml
+++ b/.github/workflows/build-boards.yml
@@ -246,3 +246,58 @@ jobs:
       app_toml: ${{ matrix.app_toml }}
       image: ${{ matrix.image }}
       os: ${{ inputs.os }}
+
+  build-tests:
+    if: ${{ inputs.board-set == 'tests' || inputs.board-set == 'all' }}
+    name: build-tests
+    strategy:
+      matrix:
+        build: [stm32f3, stm32f4, lpc55, stm32h743, stm32h753, gemini, rot-carrier, gimletlet]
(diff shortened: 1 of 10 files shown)


hawkw:
Trying to update the bitrotted PSC test config to actually build, but there's a bit of an unfortunate wrinkle: the app config for the PSC tests include a `drv-stm32xx-i2c-server` (which we need for the FRU ID test), which then depends on `drv-stm32xx-sys`, which depends on a task slot for `jefe`...but, because we use `test-runner` as the supervisor for the test suite, we don't have a `jefe`!

I think getting these tests to build again will require an actually meaningful amount of work --- perhaps we can add a feature to `drv-stm32xx-sys` to make it play nice with `test-runner` as supervisor (and add code to the test runner to allow it to do whatever Jefe stuff that  the sys driver depends on)?

cbiffle (approved):
Looks plausible! Thanks for unrotting these.

Issue #1672, opened 2024-03-20
Kernel needs to be extended so that leases can span MPU regions


cbiffle:
Currently, the kernel's memory access code won't let a task point it at a region that spans MPU descriptors. This was a deliberate simplification in the original system, where tasks had regions for "flash" and "RAM" which were often gigabytes away in the physical address space -- so spanning them was never correct.

However, the build system now (thanks to @mkeeter) smartly packs tasks into multiple region descriptors if possible. This has the side effect of inserting _difficult to predict_ region descriptor boundaries into the middle of task flash and RAM.

@arjenroodselaar has just noticed that this is a ticking timebomb, since passing a data structure to the kernel directly or as a lease to another task _may or may not fail_ depending on where those hard-to-predict boundaries fall.

We need to fix this oversimplification in the kernel and get it to tolerate such things. Meanwhile in #1671 @mkeeter has temporarily disabled packing in the build system to avoid triggering this.

cbiffle:
Alright, so, probably the easiest way of fixing this is to ensure the region table for each task is sorted in address order, and then do a linear sweep over it attempting to tile the given user slice with regions.

The order of descriptors in a task's region table is meaningful, but probably not observably so. On ARM M-profile, in the event that two regions _overlap,_ the order is used to determine which takes precedence. We don't overlap regions --- there are data structures in the build system that prevent it --- so I'm pretty sure this isn't relevant to our system. Task region tables are also opaque _to the tasks,_ which was how we were able to do better build packing in the first place without breaking code.

So, I think sorting them can be done trivially. Then it's a matter of writing the scan code.

cbiffle:
Alright, we believe this to be fixed. Packing is still disabled in the build system at this time, but, we can probably turn it back on safely.

We have actual unit tests for the new algorithm in the new `kerncore` crate, at @hawkw's prodding.

PR #1674, opened 2024-03-20
kern: permit USlices to span contiguous MPU regions

cbiffle:
Now that the build system has started packing tasks more cleverly, MPU region boundaries can appear in the middle of task RAM and Flash in positions that are difficult for the program to predict. This means the program can't safely lend any section of RAM or Flash without risking drawing the ire of the kernel, which --- owing to the limitations of earlier versions of the build system --- would not accept any access that spanned a region.

This fixes that. To make fixing that relatively inexpensive, this also sorts task region tables by base address, so we can do a linear scan with early exit.

See #1672 for more details and analysis.

### DIFF
--- a/sys/kern/src/task.rs
+++ b/sys/kern/src/task.rs
@@ -213,18 +213,14 @@ impl Task {
         desired: RegionAttributes,
         forbidden: RegionAttributes,
     ) -> bool {
-        if slice.is_empty() {
-            // We deliberately omit tests for empty slices, as they confer no
-            // authority as far as the kernel is concerned. This is pretty
-            // important because a literal like `&[]` tends to produce a base
-            // address of `0 + sizeof::<T>()`, which is almost certainly invalid
-            // according to the task's region map... but fine with us.
-            return true;
-        }
+        // Forceably include DEVICE in the forbidden set, whether or not the
+        // caller thought about it.
         let forbidden = forbidden | RegionAttributes::DEVICE;
-        self.region_table().iter().any(|region| {
-            region.covers(slice)
-                && region.attributes.contains(desired)
+
+        // Delegate the actual tests to the kerncore crate, but with our
+        // attribute-sensing customization:
+        kerncore::can_access(slice, self.region_table(), |region| {
+            region.attributes.contains(desired)
                 && !region.attributes.intersects(forbidden)
         })
     }
--- a/sys/kerncore/src/lib.rs
+++ b/sys/kerncore/src/lib.rs
@@ -1,0 +1,389 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Allow std-y things to be used in test. Note that this attribute is a bit of a
+// trap for the programmer, because rust-analyzer by default seems to build
+// things with test set. This means it's easy to introduce code incompatible
+// with no_std without your editor hassling you about it. Beware.
+#![cfg_attr(not(test), no_std)]
+
+/// Describes types that act as "slices" (in the very abstract sense) referenced
+/// by tasks in syscalls.
+///
+/// This is not the same thing as a Rust slice in the kernel, because this is
+/// just a base-length pair --- taken by itself, it doesn't let you actually
+/// _access_ the memory.
+///
+/// # Invariants
+///
+/// `s.is_empty()` implies `s.base_addr() == s.end_addr()`, and vice versa.
+///
+/// `s.base_addr() <= s.end_addr()` must hold.
+pub trait UserSlice {
+    /// Checks whether the slice spans zero bytes. Empty slices are opted out of
+    /// access checking to allow task code to use literals like `&[]`.
+    ///
+    /// This must be consistent with the base/end addr implementations, such
+    /// that `is_empty <==> base_addr == end_addr`.
+    fn is_empty(&self) -> bool;
+
+    /// The address of the first byte included in this slice.
+    ///
+    /// The value returned by `base_addr` must be less than or equal to that
+    /// returned by `end_addr`.
+    fn base_addr(&self) -> usize;
+
+    /// The address of the first byte _not_ included in this slice, past the
+    /// end.
+    ///
+    /// Note that this prevents a slice from touching the end of the address
+    /// space. This is also prevented, in practice, by the definition of several
+    /// Rust core types, so we accept it.
+    ///
+    /// The return value must be greater than or equal to the result of
+    /// `base_addr`.
+    fn end_addr(&self) -> usize;
+}
+
+impl<T: UserSlice> UserSlice for &T {
+    #[inline(always)]
+    fn is_empty(&self) -> bool {
+        (**self).is_empty()
+    }
+
+    #[inline(always)]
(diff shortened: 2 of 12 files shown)


sys/kern/src/task.rs
```diff
@@ -221,12 +221,44 @@ impl Task {
...
+                    // to continue processing.
+                    return true;
+                }
+
+                // Continue scanning at the end of this region.
+                scan_addr = region.end_addr();
+            } else if region.end_addr() > scan_addr {
+                // We've passed our target address without finding regions that
+                // work!
+                break;
```
hawkw, line 255: very minor nit, not actually important: it's _slightly_ weird to me that we use `return false` _and_ `break`, both of which ultimately result in us returning `false` --- i would kinda prefer it if both of the cases where we return false used the same thing?

sys/kern/src/task.rs
```diff
@@ -221,12 +221,44 @@ impl Task {
             // according to the task's region map... but fine with us.
             return true;
         }
+        // Forceably include DEVICE in the forbidden set, whether or not the
+        // caller thought about it.
         let forbidden = forbidden | RegionAttributes::DEVICE;
-        self.region_table().iter().any(|region| {
-            region.covers(slice)
```
hawkw, line 226: `covers` is dead code now, right?

sys/kern/src/task.rs
```diff
@@ -221,12 +221,44 @@ impl Task {
...
+        // We need to be convinced that this slice is _entirely covered_ by
+        // regions with the right attributes. It _may_ span multiple regions,
+        // which is important since the build system can insert region
+        // boundaries in unpredictable locations in otherwise innocent programs.
+        // But the regions that are spanned must be perfectly contiguous.
+
+        let mut scan_addr = slice.base_addr;
+        let end_addr = slice.end_addr();
+
+        for region in self.region_table() {
```
mkeeter, line 237: A comment here explaining that `region_table` is (1) sorted by start address and (2) non-overlapping would help clarify this loop

hawkw:
Looks like the LPC55 apps [need their kernels embiggened again](https://github.com/oxidecomputer/hubris/actions/runs/8380372418/job/22949479718).

sys/kerncore/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+        (**self).base_addr()
+    }
+
+    fn end_addr(&self) -> usize {
+        (**self).end_addr()
+    }
+}
+
+pub trait MemoryRegion {
+    fn contains(&self, addr: usize) -> bool;
```
hawkw, line 35: nit, take it or leave it: it seems like we could probably provide a default implementation of `contains` like:
```suggestion
    fn contains(&self, addr: usize) -> bool {
        addr >= self.base_addr() && addr < self.end_addr()
    }
```

sys/kerncore/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+    fn base_addr(&self) -> usize {
+        (**self).base_addr()
+    }
+
+    fn end_addr(&self) -> usize {
+        (**self).end_addr()
+    }
+}
+
+pub fn can_access<S, R>(
```
hawkw, line 57: might be nice to have a doc comment explaining what this does?

hawkw (commented):
The factoring out into `kerncore` is lovely!

sys/kern/src/task.rs
cbiffle: It is! Removing.

sys/kern/src/task.rs
cbiffle: Good point, fixing.

sys/kern/src/task.rs
cbiffle: Adding!

sys/kerncore/src/lib.rs
cbiffle: Currently I will leave it, because the kernel implementation is slightly more subtle than that, but in ways that are not generic. I would hate to accidentally pick up a default impl in the kernel. Good suggestion though.

sys/kerncore/src/lib.rs
cbiffle: Adding!

cbiffle:
Alright, addressed those comments by adding lots of comments (and changing one line of actual code). PTAL.

cbiffle:
> Looks like the LPC55 apps [need their kernels embiggened again](https://github.com/oxidecomputer/hubris/actions/runs/8380372418/job/22949479718).

This part, I am still fixing.

sys/kerncore/src/lib.rs
hawkw: Ah, okay --- out of curiosity, what are those subtleties?

hawkw (approved):
Looks good to me! I left some more comments, but I'm fine with this merging as-is!

sys/kerncore/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+            );
+        }
+    }
+
+    #[test]
+    fn cannot_access_single_bad_region() {
+        let region_table = make_fake_region_table();
+        for i in [BAD_REGION_0_IDX, BAD_REGION_1_IDX] {
+            assert!(
+                // load-bearing tiny punctuation character:
```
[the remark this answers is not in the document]
cbiffle: I enthusiastically agree, and much prefer doing `assert_eq!(something, false);` because it's much harder to miss.

sys/kerncore/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+/// Generally speaking, not all task region descriptors confer the same rights
+/// --- some are read-only, some may represent an address space hole that cannot
+/// be accessed, etc. To reflect this, this routine takes a `region_ok`
+/// predicate over the `R` type. Provide a `region_ok` function to do any access
+/// permission checking that you require.
+///
+/// # Preconditions
+///
+/// `table` must be sorted by region base address, and the regions in the table
+/// must not overlap.
```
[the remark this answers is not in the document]
cbiffle: We don't build the _kernel_ with debug assertions, no. I thought about this, and it would find bugs in the test fixture in this crate, which is something? But it wouldn't detect bugs in the kernel.

sys/kerncore/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+/// `s.is_empty()` implies `s.base_addr() == s.end_addr()`, and vice versa.
+///
+/// `s.base_addr() <= s.end_addr()` must hold.
+pub trait UserSlice {
+    /// Checks whether the slice spans zero bytes. Empty slices are opted out of
+    /// access checking to allow task code to use literals like `&[]`.
+    ///
+    /// This must be consistent with the base/end addr implementations, such
+    /// that `is_empty <==> base_addr == end_addr`.
+    fn is_empty(&self) -> bool;
```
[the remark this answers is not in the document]
cbiffle: I could be convinced, but, I'm kind of concerned about default implementations here accidentally making their way into the kernel. This is why I haven't provided any.

As for rephrasing it in terms of base+len... that would accurately describe the types in the kernel, though the kernel's `end_addr` routine has been optimized to avoid the checked arithmetic, so we'd still need to be able to provide an `end_addr` override somehow, and I think that winds up meaning a default impl.

Issue #1548, opened 2023-10-06
Consider using SHA2-256 for RoT FWID measurements


lzrd:
The software implementation of SHA3-256 adds to flash and RAM requirements more than using the hardware implementation of SHA2-256 that is available from the LPC55 ROM.

TODO:
  - Measure the difference in the resulting Hubris image sizes and decide if the savings are important.
  - Discuss any security implications and decide if they are important for this application.

lzrd:
I'd like to save some flash and RAM and I don't think that the differences in algorithm strength apply to this case given that there is also a signature check and image format tests.

lzrd:
Closing, we're all tooled up for SHA3-256 for Firmware ID (FWID) and there doesn't seem to be a compelling reason to move to the HW-supported SHA2-256. Upcoming PR does make the FWID an enum with one SHA3-256 variant to make it easier to change this in the future.

Issue #1404, opened 2023-06-08
RoT should do local signature validation for bootleby before flashing.


lzrd:
The RoT does not currently perform signature validation before flashing the bootloader.
Instead, it relies on kindness of upper layers to only request flashing of correct images.

Rather than put signature validation in `update_server`, Cliff has suggested staging an image in bootloader reserved flash and having the bootloader or pre-main perform a verification using the same ROM routines use at boot time. In addition to the verification status already passed for Hubris-A and Hubris-B images, the status of the staged image would be passed as well.

Details on where update_server would place the image for verification, coordination of the necessary reset to give early FW control for verification, and how that information is passed is TBD.

The control plane does not currently update the bootloader. After this feature is implemented, it would be safer to do so.
Most incorrect installations have been done using faux-mgs by "hand" or through scripts to date.

RFDs 349 and 374 should be updated to reflect the new feature.

lzrd:
Closed by PR 1674

Issue #1353, opened 2023-05-11
Update_server should provide current hashes of flash banks.


lzrd:
RoT update_server should provide its current view of the stage0, A, and B flash banks.

On RoT, the unused trailing flash sectors are undefined to allow for flexibility in their future use.
The trailing sectors may be in the erased state or may hold content.

There is also a difference between the canonical byte stream that is signed and the byte stream of the contents of a flash bank which would include the signature block and may include trailing bytes in the last sector as well.

The NXP LPC55Sx also skips over one header field (IIRC) when computing the hash for signature verification.

Since we care about the contents of unused flash sectors because of their potential use in exfiltration attacks or for storing malicious code or data, we want the entire bank hashed.
 
This creates a bit more work during update to be able to predict the layout in the target device and hash given that we already have code to mirror the LPC55 signature verification routine.

It is suggested that update server return something like RotFlashContents below:

struct BankHash {
   physaddr: u32,
   length: u32,
   hash: [u8; 32], // Make it clear what algorithm is in use with an appropriate type.
}

struct RotFlashContents {
  stage0: BankHash,
  a: BankHash,
  b: BankHash,
}

stage1 (a.k.a. pre-kernel, a.k.a.) already computes the bank hashes prior to update_server's start. Update_server's primary function is to mutate those contents which makes the original stage1 information inaccurate during an update procedure.

Looking ahead, we would also want SPFlashContents, but that speculates on having the RoT implement measurement and bank selection policy enforcement against the SP using SWD. That's an issue for another day.

PR #1675, opened 2024-03-21
Update RoT bootloader, a.k.a. stage0

### LINKED ISSUE #1043: Update RoT memory map to reserve RAM for updates
Due to system constraints there is no alternate image for the stage0 bootloader. This makes updating it a riskier proposition than Hubris since a broken stage0 may be equivalent to a bricked board. To reduce risk, we can reserve a region of RAM for staging an update before committing to flash.

Section 2.1.8 of the LPC55 manual gives a table of RAM layouts. The final product will use the 320 KB layout. We have a size limit of 64K for stage0. We've been testing with the 256KB layout which lacks RAM3 and we aren't running into any limitations so reserving RAM3 looks plausible.

lzrd:
This PR has four commits to simplify reviewing.

  - The first is Laura's macro to manipulate the HASHCRYPT interrupt vector.
  - 2nd is pre-main kernel changes to create a staging area for Bootleby update and to validate the (now) four RoT flash banks and record those results for use by update_server.
  - 3rd are the API changes to support stage0 update including treating the bootloader as a separate component with A (active) and B (stage0next) banks.
  - last is the code in the RoT update_server to actually update stage0

Before merging this PR, Cargo.toml needs to be edited to reference the gateway-message-service main branch.
Merging needs to be coordinated with an MGS merge.

Closes #1043, #1404, #1548, #1353,

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,9 +117,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 # Oxide forks and repos
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.1.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-#gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
-# XXX
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", branch="update-stage0", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 gimlet-inspector-protocol = { git = "https://github.com/oxidecomputer/gimlet-inspector-protocol", version = "0.1.0" }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
 humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.3" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2347,7 +2347,7 @@ checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 [[package]]
 name = "gateway-messages"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/management-gateway-service?branch=update-stage0#e8a4d55409bb7c227ae8521c5b41b12b814dc442"
+source = "git+https://github.com/oxidecomputer/management-gateway-service#c85a4ca043aaa389df12aac5348d8a3feda28762"
 dependencies = [
  "bitflags 2.5.0",
  "hubpack",


lzrd:
See https://github.com/oxidecomputer/sprot-e2e for test scripts

Cargo.toml
```diff
@@ -119,7 +119,9 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 # Oxide forks and repos
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.1.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+#gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+# XXX
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", branch="update-stage0", default-features = false, features = ["smoltcp"] }
```
lzrd, line 124: This must be fixed before merge.

drv/lpc55-update-server/src/main.rs
```diff
@@ -596,6 +624,133 @@ impl ServerImpl<'_> {
...
+        // signature routine.
+
+        // Read stage0next contents into RAM.
+        let staged = image_range(RotComponent::Stage0, SlotId::B);
+        let len = self.read_flash_image_to_cache(staged.0)?;
+        let bootloader = &self.fw_cache[..len / core::mem::size_of::<u32>()];
+
+        let mut hash = Sha3_256::new();
+        for page_start in (0..(bootloader.len() * 4)).step_by(512) {
+            // XXX
```
hawkw, line 691: hmm?

drv/lpc55-update-server/src/main.rs
```diff
@@ -573,6 +623,411 @@ impl ServerImpl<'_> {
...
+            hash.update(
+                bootloader.as_bytes()[page_start..page_start + 512].as_ref(),
+            );
+        }
```
hawkw, line 695: style nit: i think we could loop over every 512-byte page by writing this, instead:
```suggestion
        for page in bootloader.as_bytes().chunks(512) {
            hash.update(page);
        }
```

which feels much clearer to me?

drv/lpc55-update-server/src/main.rs
```diff
@@ -615,6 +770,137 @@ impl ServerImpl<'_> {
...
+
+    fn read_flash_image_to_cache(
+        &mut self,
+        span: Range<u32>,
+    ) -> Result<usize, UpdateError> {
+        // Returns error if flash page is erased.
+        let staged = image_range(RotComponent::Stage0, SlotId::B);
+        let len = self.valid_flash_image_present(&staged.0)?;
+        if len as u32 > span.end || len > self.fw_cache.as_bytes().len() {
+            return Err(UpdateError::BadLength); // XXX unique error
```
hawkw, line 861: do we want to sort out these error types before merging this PR?

drv/lpc55-update-server/src/main.rs
```diff
@@ -573,6 +623,411 @@ impl ServerImpl<'_> {
...
+                return Err(UpdateError::ImageMismatch);
+            }
+        }
+        Ok(())
+    }
+
+    // Looking at a region of flash, determine if there is a possible NXP
+    // image programmed. Return the length in bytes of the flash pages
+    // comprising the image including padding to fill to a page boundary.
+    fn valid_flash_image_present(
```
hawkw, line 813: nit, take it or leave it: it feels a bit strange to me that a method name ending in `_present` returns a number and not a boolean? `_present` feels like a bool to me. but, i'm open to being convinced otherwise.

drv/lpc55-update-server/src/main.rs
```diff
@@ -573,6 +623,411 @@ impl ServerImpl<'_> {
...
+                    // but this is enough bytes for an NXP header and not
+                    // bigger than the flash slot.
+                    if len as usize <= span.len() && len >= HEADER_OFFSET {
+                        return Ok(len as usize);
+                    }
+                }
+                Err(UpdateError::BadLength)
+            }
+            Err(error) => Err(error),
+        }
```
hawkw, line 835: style nit: can't this `match` be:
```suggestion
         indirect_flash_read(
            &self.flash,
            span.start + LENGTH_OFFSET as u32,
            buf[..].as_bytes_mut(),
        )?;
 
        if let Some(len) = round_up_to_flash_page(buf[0]) {
            // The minimum image size should be further constrained
            // but this is enough bytes for an NXP header and not
            // bigger than the flash slot.
            if len as usize <= span.len() && len >= HEADER_OFFSET {
                return Ok(len as usize);
            }
        }
        
        Err(UpdateError::BadLength)
```

drv/lpc55-update-server/src/main.rs
```diff
@@ -573,6 +623,411 @@ impl ServerImpl<'_> {
...
+    }
+
+    fn valid_cache_image_present(&self) -> Result<usize, UpdateError> {
+        let len = if let Some(len) = round_up_to_flash_page(
+            self.fw_cache[LENGTH_OFFSET / core::mem::size_of::<u32>()],
+        ) {
+            len
+        } else {
+            return Err(UpdateError::BadLength);
+        };
```
hawkw, line 845: style nit: the if-else feels a bit hard to follow due to the weird way `rustfmt` wants to format this,what do you think of:
```suggestion
        let len =  round_up_to_flash_page(
            self.fw_cache[LENGTH_OFFSET / core::mem::size_of::<u32>()],
        ).ok_or(UpdateError::BadLength)?;
```

hawkw (commented):
I'm not very familiar with the details of the LPC55 platform, so my review focuses more or less exclusively on the implementation. I've mostly left some code style suggestions, which you're welcome to ignore --- I hope they're helpful.

drv/lpc55-update-server/src/main.rs
```diff
@@ -273,14 +294,55 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+            0 => Err(UpdateError::VersionNotSupported.into()),
+            1 => Ok(VersionedRotBootInfo::V1(RotBootInfo {
+                active: boot_state.active.into(),
+                persistent_boot_preference,
+                pending_persistent_boot_preference,
+                transient_boot_preference,
+                slot_a_sha3_256_digest: Some(boot_state.a.digest),
+                slot_b_sha3_256_digest: Some(boot_state.b.digest),
+            })),
+            _ => Ok(VersionedRotBootInfo::V2(RotBootInfoV2 {
```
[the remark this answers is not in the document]
lzrd: This is forward compatibility. If there is a next version that this implementation doesn't know about it will return the highest that it does know about. That allows the control plane to continue the update procedure without another roundtrip.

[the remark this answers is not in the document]
hawkw: Ah, I see --- that makes sense. Maybe a comment explaining that would be nice?

app/gimletlet/app-mgmt.toml
```diff
@@ -34,7 +34,7 @@ task-slots = ["jefe"]
 name = "drv-user-leds"
 features = ["stm32h7"]
 priority = 2
-max-sizes = {flash = 2048, ram = 1024}
+max-sizes = {flash = 2048, ram = 2048}
```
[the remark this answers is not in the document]
lzrd: This app builds fine for me.
I've adjusted or remove a few to not break things that compiled before. And I think removing the unnecessary `max-sizes` and `required` settings in the app.toml files is a good idea. Let's do that in a global cleanup PR though.

lib/stage0-handoff/src/rot_update_details.rs
```diff
@@ -30,22 +67,112 @@ pub struct RotBootState {
...
+    }
+}
+
+fits_in_ram!(RotBootStateV2);
+
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
 )]
 pub struct RotImageDetails {
+    // The SHA3-256 measurement of all programmed pages in the flash slot.
```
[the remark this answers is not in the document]
lzrd: done

lib/stage0-handoff/src/rot_update_details.rs
```diff
@@ -30,22 +67,112 @@ pub struct RotBootState {
...
+    /// 0xff bytes. All non-erased pages in the flash slot are included in the
+    /// digest.
+    Sha3_256([u8; 32]),
+}
+
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, SerializedSize,
+)]
+pub struct RotImageDetailsV2 {
+    pub digest: [u8; 32],
```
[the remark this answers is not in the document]
lzrd: done

app/gimletlet/app-mgmt.toml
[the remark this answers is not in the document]
hawkw: sounds good to me!

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -55,124 +55,341 @@ extern "C" {
...
+    // page, are a possibility that could be a problem with respect to
+    // unexpected crashes or catching exfiltration attempts.
+    fn new(
+        slot: &FlashSlot,
+        run: Range<u32>,
+        header_required: bool,
+        syscon: &SYSCON,
+    ) -> Result<Image, ImageError> {
+        // Make sure we can access the page where the vectors live.
+        // SAFETY: Link time constants from our own image.
```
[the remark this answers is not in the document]
lzrd: I count 43  SAFETY and 79 Safety with some inconsistent use of a colon after that word.
Making this uniform throughout Hubris should be a separate cleanup commit.
I'll switch the ones here to "Safety: " to side with the majority bring the score to 28 SAFETY and 90 Safety.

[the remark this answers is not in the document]
hawkw: sounds good to me --- definitely not expecting the global cleanup pass to happen in this PR!

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -55,124 +55,341 @@ extern "C" {
...
+            0 => clear_bit!(syscon.presetctrl0, PMASK),
+            1 => clear_bit!(syscon.presetctrl1, PMASK),
+            2 => clear_bit!(syscon.presetctrl2, PMASK),
+            _ => panic!(),
+        }
+
+        let authorized = unsafe {
+            lpc55_romapi::authenticate_image(self.span.start).is_ok()
+        };
+        // let authorized = true;
```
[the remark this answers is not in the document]
lzrd: The comment? Gone.

drv/lpc55-update-server/src/main.rs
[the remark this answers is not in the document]
lzrd: done

drv/lpc55-update-server/src/main.rs
lzrd: Thanks.

drv/ksz8463/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 num-traits = { workspace = true }
 
+counters = {path = "../../lib/counters"}
```
mkeeter, line 9: Is this change (and the one to `drv/ksz8463/src/lib.rs`) intentional?  It looks like a merge / rebase glitch.

drv/lpc55-update-api/src/lib.rs
```diff
@@ -109,6 +191,24 @@ pub enum UpdateTarget {
     Bootloader,
 }
 
+/// Designate a logical sub-component of the RoT
+///
```
mkeeter, line 195:
```suggestion
```

drv/stm32h7-spi-server-core/Cargo.toml
```diff
@@ -14,6 +14,7 @@ zerocopy = { workspace = true }
 drv-spi-api = { path = "../spi-api" }
 drv-stm32h7-spi = { path = "../stm32h7-spi" }
 drv-stm32xx-sys-api = { path = "../stm32xx-sys-api" }
+counters = { path = "../../lib/counters" }
```
mkeeter, line 17: Same here, these changes (and the one to `stm32h7-spi-server-core/src/lib.rs`) seem accidentally included in the PR

drv/vsc-err/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+counters = { path = "../../lib/counters" }
```
mkeeter, line 7: same here

drv/stm32h7-spi-server-core/Cargo.toml
lzrd: Those were the result of me trying to rebase without having to do a push --force. The point there being that it would not invalidate previous comments from github. I've done a normal rebase and forced that. These should go away but I will give it a look.

lib/lpc55-hashcrypt/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+/// proc macro to generate iocon function for LPC55
+extern crate proc_macro;
+use proc_macro::TokenStream;
+extern crate proc_macro2;
+use quote::quote;
+
+#[proc_macro]
+pub fn dynamic_hashcrypt(_item: TokenStream) -> TokenStream {
```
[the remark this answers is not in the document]
labbott: I wrote this code many months ago and I'm doing my best to remember why this needed to be a proc macro. I _think_ there was some issue with calling `kern::arch::DefaultHandler()`. I'll dig through my notes some more and see if I can remember why.

lib/lpc55-hashcrypt/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+    let code = quote! {
+        static mut USE_ROM: core::sync::atomic::AtomicBool = core::sync::atomic::AtomicBool::new(false);
```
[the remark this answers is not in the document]
labbott: My poor understanding of the Atomic API!

lib/lpc55-rot-startup/src/images.rs
```diff
@@ -55,124 +55,340 @@ extern "C" {
...
+    fn check_signature(&self, syscon: &SYSCON) -> Result<(), ImageError> {
+        const HASHAES: u32 = 32 + 32 + 18;
+        const PMASK: u32 = 1 << (HASHAES % 32);
+        const REG_NUM: u32 = HASHAES / 32; // XXX must be 0, 1, or 2
+        match REG_NUM {
+            0 => clear_bit!(syscon.presetctrl0, PMASK),
+            1 => clear_bit!(syscon.presetctrl1, PMASK),
+            2 => clear_bit!(syscon.presetctrl2, PMASK),
+            _ => panic!(),
+        }
```
labbott, line 374: We shouldn't need to use this manual bit entry, we can just use the pac APIs as intended https://docs.rs/lpc55-pac/latest/lpc55_pac/syscon/presetctrl0/index.html etc.

lib/lpc55-hashcrypt/src/lib.rs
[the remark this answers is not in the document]
hawkw: Well, in that case, this could probably be changed to 
```suggestion
        static USE_ROM: core::sync::atomic::AtomicBool = core::sync::atomic::AtomicBool::new(false);
```
and the use of `unsafe` can be removed!

lib/lpc55-hashcrypt/src/lib.rs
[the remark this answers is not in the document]
labbott: I think you are right that this code need not be procedural. The procedural bit was probably a holdover from some earlier attempts.

drv/lpc55-update-server/src/main.rs
```diff
@@ -432,7 +403,7 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
             return Err(UpdateError::UpdateInProgress.into());
         }
         self.syscon.chip_reset();
-        panic!()
+        unreachable!();
```
labbott, line 406: Did this reduce the flash size? according to https://doc.rust-lang.org/stable/std/macro.unreachable.html this will still produce an implicity panic

drv/lpc55-update-server/src/main.rs
```diff
@@ -573,6 +628,406 @@ impl ServerImpl<'_> {
...
+        indirect_flash_read(
+            &self.flash,
+            span.start,
+            self.fw_cache[0..len / core::mem::size_of::<u32>()].as_bytes_mut(),
+        )?;
+        Ok(len)
+    }
+
+    // XXX This API needs to be sorted out.
+    // The lower level routines are being replaced in PR 1500
```
labbott, line 872: Is this still true?

drv/lpc55-update-server/src/main.rs
```diff
@@ -573,6 +628,406 @@ impl ServerImpl<'_> {
...
+    fn write_cache_to_flash(
+        &mut self,
+        component: RotComponent,
+        slot: SlotId,
+    ) -> Result<(), UpdateError> {
+        let clen = self.cache_image_len()?;
+        if clen % BYTES_PER_FLASH_PAGE != 0 {
+            return Err(UpdateError::BadLength); // XXX need unique error
```
labbott, line 880: Still need the unique error?

drv/lpc55-update-server/src/main.rs
```diff
@@ -963,6 +1343,18 @@ fn copy_from_flash_range(
...
+    // Safety: Data is published by stage0
+    let addr = unsafe { BOOTSTATE.assume_init_ref() };
+    RotBootStateV2::load_from_addr(addr)
+}
+
+fn round_up_to_flash_page(offset: u32) -> Option<u32> {
+    offset
+        .checked_add(BYTES_PER_FLASH_PAGE as u32 - 1)
+        .map(|sum| sum & !(BYTES_PER_FLASH_PAGE as u32 - 1))
+}
```
labbott, line 1356: Can we use https://doc.rust-lang.org/stable/std/primitive.u32.html#method.checked_next_multiple_of ? This is a new stable API

lib/lpc55-hashcrypt/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+/// proc macro to generate iocon function for LPC55
+extern crate proc_macro;
+use proc_macro::TokenStream;
+extern crate proc_macro2;
+use quote::quote;
+
+#[proc_macro]
```
labbott, line 11: Can we delete this code if we're not using it?

lib/lpc55-rot-startup/src/lib.rs
```diff
@@ -290,3 +323,5 @@ extern "C" fn nuke_stack() {
         )
     }
 }
+
+lpc55_hashcrypt::dynamic_hashcrypt!();
```
labbott, line 327: Do we we need this or can we go to the non-procmacro version

drv/lpc55-update-server/src/main.rs
lzrd: I decided not. If some length in the image is invalid its not hard to figure out which one.

lib/lpc55-hashcrypt/src/lib.rs
lzrd: This code has been previously removed.

lib/lpc55-rot-startup/src/lib.rs
lzrd: Already removed

aapoalas (commented):
Random questions, please excuse my curiosity.

labbott:
LGTM! I'll give the final approval once the MGS portion is merged

drv/lpc55-update-server/src/images.rs
```diff
@@ -0,0 +1,104 @@
...
+    static __this_image: [u32; 0];
+}
+
+// Location of the NXP header
+pub const HEADER_BLOCK: usize = 0;
+
+// NXP LPC55's mixed header/vector table offsets
+const RESET_VECTOR_OFFSET: usize = 0x04;
+pub const LENGTH_OFFSET: usize = 0x20;
+pub const HEADER_OFFSET: u32 = 0x130;
```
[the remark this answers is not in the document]
lzrd: This should be using [ImageVectorsLpc55](https://github.com/oxidecomputer/hubris/blob/558ef2d6e6e961515bc3905947fe4f5b6390d08a/lib/lpc55-rot-startup/src/images.rs#L19) which should be pulled out and made more available.

labbott (approved):

PR #1682, opened 2024-03-26
Change the default caboose VERS

labbott:
The default caboose version is currently `0.0.0-git`. Change it to a value that will fail to assemble in a TUF repo as we don't want untagged hubris images going into built repos.


cbiffle (approved):

cbiffle:
The other option that's come up in chat is omitting the VERS chunk entirely from unversioned images. This should reduce the chances of accidentally propagating whatever default version we choose to zero.

Issue #1680, opened 2024-03-25
sensors IPC API should use reply-fault.


cbiffle:
Currently the operations in the sensors IPC API can generally fail for two reasons not related to the sensor device itself:
1. If an invalid sensor ID is passed in.
2. If there is no data.

(1) is not expected to happen in correct programs _except_ in the case where a tool allows a user to enter an arbitrary sensor ID. (We want to support that, and also gracefully handle cases where e.g. MGS queries the fleet for a sensor that's maybe only implemented on a subset.) So I'd argue this could use reply-fault at the IPC interface, as long as we provide callers a way to ensure they are meeting the "valid SensorId" precondition.

(2) is absolutely expected to happen in correct programs but, I would argue, does not belong in an error type, and should probably use `Option` (the normal way of indicating missing data).

Here's an alternative proposed IPC API, written as Rust functions because Idol syntax is woooordy.

```rust
// SensorId now only represents _valid_ sensor IDs.
// Use SensorId::try_from if you're worried yours might be invalid.
// (Firmware in general can't really produce invalid IDs.)

impl SensorService {
    fn get(id: SensorId) -> Option<f32>;  // where None = "no data"
    fn get_reading(id: SensorId) -> Option<Reading>;

    // ...and so forth
}
```

A client constructing a `SensorId` from an arbitrary integer would use `try_from` to check it. In the case of MGS, this would happen in control-plane-agent, most likely.

A client passing an invalid SensorId into the IPC interface would be violating a precondition and would receive reply-fault.

This change looks like it could remove a _lot_ of code in various places (much of which is wrapped up in reusable functions, but is inlined in flash so it still counts). There are a couple of operations in the interface that would require careful thought, like `get_raw_reading`, which is capable of returning both `NoData` _and also_ `(NoData, timestamp)`, presumably to indicate "sensor has never checked in" vs "sensor has checked in and has no reading".

hawkw:
I'm happy to take a pass at this if you're not planning to right away?

PR #1683, opened 2024-03-26
Move `SensorId` validation to clients, and reply fault when invalid

hawkw:
As described in #1680, the `sensors` API currently validates `SensorID`s in the server task and returns an error when they are invalid. We can simplify the IPC interface by doing this validation on the client side, and reply-faulting any clients who send an invalid ID, instead. This branch does that.

This PR consists of a sequence of commits which are intended to be reviewed individually.


task/sensor-api/src/lib.rs
```diff
@@ -8,26 +8,98 @@
...
+pub struct InvalidSensor;
+
+impl SensorId {
+    /// An invalid sensor ID.
+    ///
+    /// If this ID is actually passed to a call to a sensor API IPC operation,
+    /// the operation will return an error or reply-fault. However, this value
+    /// may be used as an initializer for `SensorId`s that will be set to a
+    /// valid value later.
+    pub const INVALID: Self = Self(u32::MAX);
```
mkeeter, line 39: Does the existence of `SensorId::INVALID` break the "all `SensorId`s will be valid" invariant in a troublesome way?

I think the intention is to be checked at RPC boundaries, i.e. deserialization would fail if someone called `sensor.raw_reading(SensorId::INVALID)`, so it wouldn't get very far.

However, _within_ the `sensors` task, if someone happened to use `SensorId::INVALID`, it could be used to get out-of-bounds access to the data arrays.

task/sensor-api/src/lib.rs
```diff
@@ -8,26 +8,98 @@
...
+    pub const MAX_VALUE: u32 = config::NUM_SENSORS as u32;
+
+    /// Returns a new `SensorId` for the provided `u32` value.
+    ///
+    /// # Panics
+    ///
+    /// If `id` is greater than [`config::NUM_SENSORS`].
+    #[must_use]
+    pub const fn new(id: u32) -> Self {
+        if id > Self::MAX_VALUE {
```
mkeeter, line 51: Should this be `>=`, since we're 0-indexed?

task/host-sp-comms/src/bsp/gimlet_bcde.rs
```diff
@@ -593,7 +593,7 @@ impl ServerImpl {
         let packrat = &self.packrat; // partial borrow
         let mut data = InventoryData::DimmSpd {
             id: [0u8; 512],
-            temp_sensor: SensorId(u32::MAX),
+            temp_sensor: SensorId::INVALID,
```
mkeeter, line 596: Per my concerns about `SensorId::INVALID` below, I think we could populate this directly with `i2c_config::sensors::TSE2004AV_TEMPERATURE_SENSORS[index as usize]` here.

hawkw:
ugh, clippy's mad about some more stuff, gimme a sec...

task/sensor-api/src/lib.rs
```diff
@@ -8,26 +8,107 @@
...
+    pub const MAX_VALUE: u32 = config::NUM_SENSORS as u32;
+
+    /// Returns a new `SensorId` for the provided `u32` value.
+    ///
+    /// # Panics
+    ///
+    /// If `id` is greater than [`config::NUM_SENSORS`].
+    #[must_use]
+    pub const fn new(id: u32) -> Self {
+        // On devboard targets without sensors, `NUM_SENSORS` is zero,
```
mkeeter, line 43: Nit: now that we have `TryFrom<u32>`, should we replace this with `id.try_into().unwrap()` (or `unwrap_lite()`)?

task/sensor/src/main.rs
```diff
@@ -213,32 +187,33 @@ impl idl::InOrderSensorImpl for ServerImpl {
...
+        *self.last_reading.get_mut(id) = Some(r);
+        *self.err_value.get_mut(id) = nodata;
+        *self.err_time.get_mut(id) = timestamp;
 
         //
         // We pack per-`NoData` counters into a u32.
         //
         let (nbits, shift) = nodata.counter_encoding::<u32>();
-        let mask = (1 << nbits) - 1;
+        let mask = (1u32 << nbits).saturating_sub(1);
```
mkeeter, line 206: Is this to remove overflow checks?

mkeeter (approved):
A few more nits, but generally looks good – it's great to see more APIs become infallible!

lib/host-sp-messages/src/lib.rs
```diff
@@ -12,7 +12,6 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_big_array::BigArray;
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use static_assertions::{const_assert, const_assert_eq};
-use task_sensor_types::SensorId;
```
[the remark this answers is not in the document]
hawkw: Ah, yup, you're right; I thought we were still using the errors, but I guess not!

task/sensor-api/src/lib.rs
hawkw: This is a `const fn`, so we can't use `try_into()`, as it's a non-`const` trait method.. Potentially, we could add a non-trait method `const fn try_new(id: u32) -> Result<Self, InvalidSensorId>`, and call that (and have the `TryFrom` impl call it as well) but we would need to `match` on its result rather than `unwrap()` (which isn't `const` as it formats the error) or `unwrap_lite() ( which isn't `const` as it's a trait method). I'll go and do that, since I agree that it's nice to not have two copies of the validation.

task/sensor/src/main.rs
hawkw: Yeah, I was hoping we could completely get rid of panics --- unfortunately, we can't, as the generated `idol` server code has some potential panic paths, but eliminating overflow checks here does shave a few bytes off the flash size regardless.

lib/host-sp-messages/src/lib.rs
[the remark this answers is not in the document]
hawkw: Upon further consideration, it looks like `host-sp-messages` was the only dependent of `task-sensor-types` besides `task-sensor-api`, so I think we can actually just move everything into `task-sensor-api` and get rid of the `-types` crate entirely (well, unless anything has a Git dep on it, but I don't think anything does).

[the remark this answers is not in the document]
mkeeter: Okay, that seems fine to me (either in this PR or as a follow-up)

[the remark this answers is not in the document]
hawkw: I removed the crate in 0b6adfce7a3b92b7ca2ba714fc659c024c4efee5.

PR #1686, opened 2024-03-27
add counters to `thermal` and `sensor-polling` ringbufs

hawkw:
This branch adds ringbuf counters for error entries in a couple more places where I thought it might be nice to have counters.


hawkw:
well, i guess counting all those error variants makes `thermal` way too big. will address that soon.

PR #1687, opened 2024-03-28
power: add comments to gimlet rail sag trace facility

cbiffle:
This should help avoid the situation where someone unknowingly removes or modifies this in a way that breaks racktest.


jordanhendricks (approved):
Thanks for doing this, Cliff!

I'm not sure what the Hubris policy is on leaving TODOs in PRs, but the comments look good to me. Much appreciated.

hawkw (approved):

PR #1691, opened 2024-03-28
Enable EXTI interrupt dispatch support in stm32xx-sys

cbiffle:
This adds code to `sys` to multiplex the frustratingly limited STM32 EXTI peripheral and provide arbitrary pin change interrupts to other tasks.

We're doing this in support of a Gimlet racktest debugging operation, and so this version is deliberately a bit limited. It will, in practice, only build in gimlet or sidecar, and will only watch PI14, and will only post a specific notification to the power task. @hawkw is working on making all that stuff configurable in parallel to my work.

For this to be useful, the power task needs to follow the instructions in the Theory Of Operation comment I've added to `drv/stm32xx-sys/src/main.rs`.


drv/stm32xx-sys/src/main.rs
```diff
@@ -255,20 +284,132 @@ impl idl::InOrderSysImpl for ServerImpl<'_> {
...
+        &mut self,
+        rm: &RecvMessage,
+        disable_mask: u32,
+        enable_mask: u32,
+    ) -> Result<(), RequestError<core::convert::Infallible>> {
+        cfg_if! {
+            if #[cfg(feature = "exti")] {
+                Ok(())
+            } else {
+                Ok(())
```
hawkw, line 298: should this perhaps reply-fault if it's called when `sys` is built without the `exti` feature?

hawkw: the _ideal_ thing would probably be to make idol codegen aware of feature flags, so that the client stub isn't generated unless it's enabled? but let's figure that out later.

cbiffle: It actually does reply-fault on line 433.

hawkw: yup, i think i left the comment before it did that ---carry on!

drv/stm32xx-sys/src/main.rs
```diff
@@ -88,8 +90,72 @@ fn main() -> ! {
...
+                            // Safety: field modeled incorrectly in PAC
+                            0 => unsafe { w.exti12().bits(port as u8) },
+                            // Safety: field modeled incorrectly in PAC
+                            1 => unsafe { w.exti13().bits(port as u8) },
+                            // Safety: field modeled incorrectly in PAC
+                            2 => unsafe { w.exti14().bits(port as u8) },
+                            // Safety: field modeled incorrectly in PAC
+                            _ => unsafe { w.exti15().bits(port as u8) },
+                        }),
+                    }
```
hawkw, line 149: this whole block of code really makes me feel very positively about the PAC :/

cbiffle: YOU AND ME BOTH

cbiffle: Fortunately if we want to fix how the fields are modeled, all we have to do is construct a YAML file describing a diff we want to apply to another YAML file derived from an XML file that isn't checked into the repo.

cbiffle:
Sooooo I'm honestly kind of shocked that this didn't blow size allocations on the stm32g0 builds, but it does not appear to have done so.

I think this is ready for review.

drv/stm32xx-sys/src/main.rs
```diff
@@ -2,13 +2,51 @@
...
+//! `sys` serves as a multiplexer for pin-change interrupts generated by the
+//! EXTI unit. From a user perspective, this works as follows:
+//!
+//! 1. `sys` gets configured in the `app.toml` to route interrupts on a specific
+//!    pin to a specific task, using a specific notification.
+//! 2. That task configures the pin using `sys.gpio_configure` and friends
+//!    (probably as an input, possibly with a pull resistor).
+//! 3. That task _also_ configures the pin's edge sensitivity, using
+//!    `sys.gpio_irq_configure`. Instead of naming the pin in this API, the task
+//!    needs to pass the notification mask where it expects the IRQ to arrive.
```
hawkw, line 19: I think using notification masks _is_ the correct mechanism for this, but it does seem a little bit weird that this operation can potentially change the edge sensitivity for multiple pins, if they're mapped to the same notification bit. 

There's not, as I understand it, a way for a task to receive notifications from two pins in different edge sensitivities (and pullup configs) on the same notification bit. But, maybe that's fine...none of our tasks are anywhere near using up their 32 bits of notifications, so you can just use another one if you need different pin configs...

drv/stm32xx-sys/src/main.rs
```diff
@@ -88,8 +126,97 @@ fn main() -> ! {
...
+
+            // Grab a pointer to EXTI while we're here, as the server setup
+            // below will use it. (Do not let the apparent block scope of cfg_if
+            // fool you! This escapes to the code below and is not unused.)
+            //
+            // Safety: same complaint as above, this is needlessly unsafe in the
+            // API the way we use peripherals.
+            let exti = unsafe { &*device::EXTI::ptr() };
+
+            // exti and syscfg are used in server init below
```
hawkw, line 214: tiny comment nit: only `exti` is used in the server init

drv/stm32xx-sys/src/main.rs
```diff
@@ -255,20 +395,241 @@ impl idl::InOrderSysImpl for ServerImpl<'_> {
...
+                                    // model this field right
+                                    unsafe {
+                                        w.bits(new_value)
+                                    }
+                                });
+                            }
+                        }
+                    }
+                }
+                Ok(())
```
hawkw, line 446: do you think it makes sense to punish clients who call `gpio_irq_control` with a mask that doesn't include any notifications mapped to EXTI interrupts with a reply fault? 

we could have a flag that get set if we actually touched something, and reply fault if we didn't...it seems like a potential programmer error to call this IPC with a mask that doesn't include EXTI notifications.

drv/stm32xx-sys/src/main.rs
```diff
@@ -255,20 +395,241 @@ impl idl::InOrderSysImpl for ServerImpl<'_> {
...
+                                    r.bits() & !(1 << i)
+                                };
+                                unsafe {
+                                    w.bits(new_value)
+                                }
+                            });
+                        }
+                    }
+                }
+                Ok(())
```
hawkw, line 506: again, maybe it's worth reply-faulting if the mask doesn't actually include EXTI notifications?

drv/stm32xx-sys/src/main.rs
```diff
@@ -255,20 +395,241 @@ impl idl::InOrderSysImpl for ServerImpl<'_> {
...
+                            } else {
+                                // spurious interrupt.
+                                // TODO: probably add this to a counter; it's
+                                // almost certainly a bug, but panicking would
+                                // be rude since this is a shared service.
+                            }
+
+                            bits_to_acknowledge |= 1 << pin_idx;
+                        }
+                    }
```
hawkw, line 579: _very_ nitpicky, take it or leave it: i bet if we wrote
```suggestion
                    for (pin_idx, entry) in generated::EXTI_DISPATCH_TABLE.iter().enumerate() {
                        if pending_and_enabled & 1 << pin_idx != 0 {
                            // A channel is pending! We need to handle this
                            // basically like the kernel handles native hardware
                            // interrupts, which means
                            // - Post the event to the owning task
                            // - Mask the interrupt
                            // - Clear the pending bit (we have to do this
                            //   manually unlike native interrupts).

                            if let Some((_, taskid, mask)) = entry {
                                let taskid = sys_refresh_task_id(taskid);
                                sys_post(taskid, mask);
                            } else {
                                // spurious interrupt.
                                // TODO: probably add this to a counter; it's
                                // almost certainly a bug, but panicking would
                                // be rude since this is a shared service.
                            }

                            bits_to_acknowledge |= 1 << pin_idx;
                        }
                    }
```
here, the compiler wouldn't generate bounds checks every time we index the dispatch table...it might already be smart enough not to do that, though...

drv/stm32xx-sys/src/main.rs
```diff
@@ -255,20 +395,241 @@ impl idl::InOrderSysImpl for ServerImpl<'_> {
...
+
+                        // Mask the sources we're handling by clearing the
+                        // corresponding mask bits. The EXTI unfortunately has
+                        // no way to do this atomically, we need to RMW.
+                        // Fortunately the mask bits are only changed by
+                        // software, and only by _this_ software.
+                        self.exti.cpuimr1.modify(|r, w| {
+                            let new_value = r.bits() & !bits_to_acknowledge;
+                            // Safety: this operation is unsafe because the PAC
+                            // hasn't thought about it. Enabling an interrupt
```
hawkw, line 597: > the PAC hasn't thought about it

as is traditional :/

idl/stm32xx-sys.idol
```diff
@@ -95,5 +95,35 @@ Interface(
             reply: Simple("[u32; 3]"),
             idempotent: true,
         ),
+
+        // Configures some set of EXTI sources associated with the caller,
+        // using the caller's notification bit space to name them. Any
+        // sources included in the `mask` will be affected.
+        //
+        // If neither rising nor falling is true, the pin will not be
+        // capable of producing an interrupt, even if enabled.
```
hawkw, line 104: is this case something we ought to check for and reply fault?

hawkw (approved):
Overall, looks great! I had a few small nitpicks

drv/stm32xx-sys/src/main.rs
cbiffle: Yeah, my general response (and I should probably put this in the comment) is: don't map multiple pins to a notification bit unless you want to treat them as 100% indistinguishable. Because it has that effect.

drv/stm32xx-sys/src/main.rs
cbiffle: Good catch, legacy of an earlier revision. Will fix. (In fact I may move the exti unsafe block down to where it's installed in the server field, to reduce its available scope.)

drv/stm32xx-sys/src/main.rs
cbiffle: In general, reply-faulting in cases like that makes it easier for people to find programming errors faster, rather than hunting around for their missed events. So, yeah, I think it probably should. I admit I considered this when writing the implementation and then decided I was in a hurry.

I've gotten clarified priorities on this and am no longer in a hurry, so, I will add this.

drv/stm32xx-sys/src/main.rs
cbiffle: I actually meant to change this to iter/enumerate before posting it, and I'm kind of surprised Clippy isn't flipping out about it. I guess the single use of `pin_idx` to construct a bitmask is enough to reassure it that it isn't _only_ being used to index an array.

idl/stm32xx-sys.idol
cbiffle: I feel like that might be too aggressive, I am not convinced that it isn't a potentially useful technique.

drv/stm32xx-sys/src/main.rs
hawkw: yup, putting that in the comment works for me!

idl/stm32xx-sys.idol
hawkw: my thinking is that we have a separate interface for saying "hey, i don't want to receive this interrupt right now" (in `gpio_irq_control`), so the code that calls `gpio_irq_configure` and asks for neither the rising nor the falling edge probably didn't *intend* to do that? 

but, maybe there are occasions where it feels more correct to call this rather than `gpio_irq_configure` to disable an IRQ — perhaps other code will call `gpio_irq_control` to enable an IRQ unconditionally, and this is useful as an override? i'm fine with permitting this if you think there are valid uses for it. 🤷‍♀️

cbiffle: I guess it's easier to permit it later if it's useful, than it is to ban it later if it isn't. I'll add it as an error case with a note about how to remove it if desired.

cbiffle:
Rebased and applied requested fixes. Also expanded the docs a bit. (Still want to put docs somewhere other than the driver source code, but haven't done that yet.)

PR #1694, opened 2024-03-29
build: re-enable task packing.

cbiffle:
Now that release 7 is closer to finalized and we've gained some confidence in the kernel fixes for #1672, it would be nice to regain the benefits of task packing and stop fretting so much about fragmentation.


mkeeter (approved):

hawkw:
Sidecar rev-b kernel seems to need real embiggening: https://github.com/oxidecomputer/hubris/actions/runs/8485090582/job/23249272079?pr=1694#step:6:431

PR #1696, opened 2024-03-29
sprot: turn `ROT_IRQ` into a real IRQ

hawkw:
This is currently a draft as I haven't got a rot-carrier to test this with yet.


hawkw:
Build failure is because I haven't actually updated `app.toml`s for most targets yet, which...I could do, but would like to know if this works first.

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -363,31 +364,88 @@ impl<S: SpiServer> Io<S> {
...
+        // Enable the interrupt.
+        self.sys
+            .gpio_irq_control(notifications::ROT_IRQ_MASK, IrqControl::Enable)
+            // Just unwrap this, because the `sys` task should never panic.
+            .unwrap_lite();
+
+        // Determine the deadline after which we'll give up, and start the clock.
+        let deadline = sys_get_timer()
+            .now
+            .checked_add(max_sleep as u64)
```
lzrd, line 391: All of the callers are using constants for max_sleep IIRC which means we're probably safe not testing for very large max_sleep.
Should this be .saturating_add in case we start testing with clock close to roll-over?
The down-side would be a timeout that is too short for that close-to-rollover clock, but this task would retry.

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -363,31 +364,88 @@ impl<S: SpiServer> Io<S> {
...
                 return false;
             }
-            hl::sleep_for(1);
-            slept += 1;
         }
+
+        // Ensure the timer gets unset before returning, to avoid frightening
+        // our IPC server when it's waiting in recv without expecting a timer to
+        // go off.
+        sys_set_timer(None, notifications::sprot::TIMER_MASK);
```
lzrd, line 447: Do we want to un-configure the interrupt? If this fn is called and the is_rot_irq_asserted() state is already the desired state then we drop to this point immediately.

hawkw: Ah, yeah, you're right, the IRQ being asserted when this function is called is the only case where we'll end up here with the IRQ still un-masked. Initially, I was performing that check before the IRQ was enabled and returning early, which is why I didn't mask it here. But, I moved it to the loop, and now it may still be enabled --- I'll add that.

hawkw:
@lzrd Okay, I've addressed all the review suggestions, going to re-run the e2e tests to make sure I didn't break anything.

Also, #1734 adds codegen which will allow us to replace the hard-coded pin sets with the pin sets defined in the IRQ config. Depending on which PR merges first, I'll either update this branch to use that, or that branch to change this code.

lzrd (approved):

PR #1698, opened 2024-03-30
stm32xx: generate ingbuf/counters for EXTI IRQs

hawkw:


hawkw:
After a conversation with @cbiffle, I'm going to remove the ringbuf from this change and just generate event counters for IRQ triggers. This is for size reasons — the ringbuf code is pretty big, and it contains panics, and Cliff is currently trying to remove all panics from the sys task.

hawkw:
Okay, as of 2f46737c53537c59ef41a8f363d4785a2bdc7145, this PR is now just counters, not ringbuf.

PR #1702, opened 2024-04-01
sensor: remove noop 1 Hz timer

cbiffle:
task/sensor was configuring a timer to wake it every 1 Hz. Each time it woke up, it would snooze its alarm clock and go back to sleep.

While I definitely sympathize with that desire, this is essentially dead code that looks like a bug, so this commit removes it. I'm not sure what the history here is.

In addition to removing noise from the code, this knocks 114 bytes off the sensor task text size.

### DIFF
--- a/task/sensor/src/main.rs
+++ b/task/sensor/src/main.rs
@@ -71,10 +71,7 @@ struct ServerImpl {
     err_time: SensorArray<u64>,
 
     nerrors: SensorArray<u32>,
-    deadline: u64,
 }
-
-const TIMER_INTERVAL: u64 = 1000;
 
 impl idl::InOrderSensorImpl for ServerImpl {
     fn get(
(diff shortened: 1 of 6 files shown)


task/sensor/src/main.rs
```diff
@@ -243,24 +240,20 @@ impl ServerImpl {
...
-    let deadline = sys_get_timer().now;
-
-    //
-    // This will put our timer in the past, and should immediately kick us.
-    //
-    sys_set_timer(Some(deadline), notifications::TIMER_MASK);
-
+    // N.B. if you are staring at this macro thinking that it looks like it
+    // doesn't do anything and might be obsolescent, the key is the :upper. This
+    // macro exists exclusively to uppercase the field names below.
```
cbiffle, line 256: (this macro confused me so I tossed in this documentation of it while I was here.)

hawkw (approved):
looks good to me (although it would be nice to hear from Bryan about the original motivation for this)

bcantrill (approved):
Sadly, I think that this was born vestigial. :(   Definitely rip it out, with my apologies!

PR #1703, opened 2024-04-01
Bump the toolchain to latest nightly.

cbiffle:
Also fixed _some_ new warnings reported by the toolchain, and many places where rustfmt has apparently changed its mind about some things.

Removed size constraints for tasks that increased in size rather than playing guess-the-number.

This leaves a number of warnings, which I'm hoping to divide up and not have to fix myself.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,3 @@
-cargo-features = ["resolver", "named-profiles"]
-
 [workspace]
 members = [
     "build/*",
--- a/app/gimlet/base.toml
+++ b/app/gimlet/base.toml
@@ -123,7 +123,7 @@ features = ["gimlet"]
 name = "task-thermal"
 features = ["gimlet"]
 priority = 5
-max-sizes = {flash = 32768, ram = 8192 }
+max-sizes = {flash = 65536, ram = 8192 }
 stacksize = 6000
 start = true
 task-slots = ["i2c_driver", "sensor", "gimlet_seq", "jefe"]
--- a/app/gimlet/base.toml
+++ b/app/gimlet/base.toml
@@ -123,7 +123,7 @@ features = ["gimlet"]
 name = "task-thermal"
 features = ["gimlet"]
 priority = 5
-max-sizes = {flash = 65536, ram = 8192 }
+max-sizes = {flash = 32768, ram = 8192 }
 stacksize = 6000
 start = true
 task-slots = ["i2c_driver", "sensor", "gimlet_seq", "jefe"]
--- a/sys/kern/src/fail.rs
+++ b/sys/kern/src/fail.rs
@@ -42,8 +42,14 @@ static mut KERNEL_EPITAPH: [u8; EPITAPH_LEN] = [0; EPITAPH_LEN];
 #[cfg(not(feature = "nano"))]
 fn begin_epitaph() -> &'static mut [u8; EPITAPH_LEN] {
     // We'd love to use an AtomicBool here but we gotta support ARMv6M.
-    let previous_fail =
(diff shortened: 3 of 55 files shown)


cbiffle:
After some discussion with upstream, I've gone ahead and bumped this to latest nightly to pick up some size fixes by @saethlin and others. This has caused a cavalcade of new warnings, most of which are valid, some of which are Clippy bugs.

Currently this PR has Humility and Clippy tests disabled:
- Humility, because a bug in its DWARF parser prevents it from parsing output of rust 1.75 and later, and
- Clippy, because it is complaining about valid things in code I didn't write and would prefer not to have to fix.

Cargo.toml
```diff
@@ -1,5 +1,3 @@
-cargo-features = ["resolver", "named-profiles"]
```
hawkw, line 1: are these no longer feature flagged in the current nightly?

app/gimlet/base.toml
```diff
@@ -123,7 +123,7 @@ features = ["gimlet"]
 name = "task-thermal"
 features = ["gimlet"]
 priority = 5
-max-sizes = {flash = 32768, ram = 8192 }
+max-sizes = {flash = 65536, ram = 8192 }
```
hawkw, line 126: what happened here?

sys/kern/src/fail.rs
```diff
@@ -42,8 +42,11 @@ static mut KERNEL_EPITAPH: [u8; EPITAPH_LEN] = [0; EPITAPH_LEN];
 #[cfg(not(feature = "nano"))]
 fn begin_epitaph() -> &'static mut [u8; EPITAPH_LEN] {
     // We'd love to use an AtomicBool here but we gotta support ARMv6M.
-    let previous_fail =
-        core::mem::replace(unsafe { &mut KERNEL_HAS_FAILED }, true);
+    // This could probably become SyncUnsafeCell in a future where it exists.
+    let previous_fail = core::mem::replace(
+        unsafe { &mut *core::ptr::addr_of_mut!(KERNEL_HAS_FAILED) },
+        true,
+    );
```
hawkw, line 49: Alternatively, as per the above comment re: `AtomicBool`, could we use `armv6m-atomic-hack` here instead? Or is there a reason not to?

hawkw (approved):
overall, this seems correct to me once the requisite Humility changes and clippy fixes are done!

app/gimlet/base.toml
cbiffle: It got bigger and I wasn't excited about removing the limit (because I want to investigate) or figuring out a precise new one (because I'm changing the toolchain version a lot right now).

sys/kern/src/fail.rs
cbiffle: That doesn't work in the kernel. It could probably be made to work in the kernel, but I haven't really looked into it yet.

saethlin: Are you sure you don't want to use `ptr::replace` here? Reborrowing into `&mut` after you've used `addr_of_mut!` is odd.

saethlin: Oh I see, the point of this function is to return `&'static mut`. So I suppose some friction with the lint is unavoidable.

cbiffle: Hi @saethlin! I think you're right actually, because the two blocks are operating on two different statics. I have replaced the replace with `core::ptr::replace`, which it appears I didn't know existed when I wrote this.

cbiffle:
Humility is back on after the fix there made it into the nightly build. Clippy is still off, will file bugs to track once we get this in (since that will mean the toolchain version stops changing).

flihp (approved):
LGTM

PR #1705, opened 2024-04-02
Fix whitespace inconsistency in Idol files

mkeeter:
There should be no non-whitespace changes in this PR!

- Always use 4x spaces for indentation
- Don't include empty `args: {}`
- Don't include a space before `:` or after `(`

### DIFF
--- a/idl/attest.idol
+++ b/idl/attest.idol
@@ -16,7 +16,7 @@ Interface(
         "cert": (
             doc: "Get a cert from the RoT-R",
             args: {
-				"index" : "u32",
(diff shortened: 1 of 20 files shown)


cbiffle:
I'm unconvinced that removing `args: {}` improves clarity. Convince me?

I feel like it makes it visually easier to determine that there are no args, as opposed to "it's several lines down and you missed it."

Otherwise I'm all for normalizing whitespace and where are we getting tabs anyway

cbiffle (approved):

PR #1706, opened 2024-04-02
stm32xx-sys: configure GPIO IRQ edge with an enum

hawkw:
The current `Sys::gpio_irq_configure` IPC interface is a bit confusing, as it takes two `bool`s for the rising and falling edge sensitivities, as in:

```rust
sys.gpio_irq_configure(mask, true, false);
```

This isn't particularly clear for a reader, who then has to remember the order of the arguments to determine which edge sensitivity is being enabled here. This is a case of ["boolean blindness"][1].

This commit improves the interface by changing it to take an `enum` with variants for `Rising`, `Falling`, and `Both`. Now, the same call to the IPC looks like this:

```rust
sys.gpio_irq_configure(mask, Edge::Rising);
```

which should be much more obvious to a reader, who no longer needs to remember the argument ordering to decypher calls to this function.

Also, this has the advantage of preventing the case where both rising and falling edge sensitivities are disabled at compile-time. Currently, this is a runtime error (reply fault), but with the new interface, a caller can provide an invalid configuration if they hand-write the input buffer rather than using the Idol client stub, so it should be harder to mess up.

[1]: https://runtimeverification.com/blog/code-smell-boolean-blindness

### DIFF
--- a/drv/stm32xx-sys-api/src/lib.rs
+++ b/drv/stm32xx-sys-api/src/lib.rs
@@ -37,6 +37,11 @@ pub enum RccError {
 #[derive(
     Copy, Clone, FromPrimitive, PartialEq, Eq, AsBytes, serde::Deserialize,
 )]
+// NOTE: This `repr` attribute is *not* necessary for
+// serialization/deserialization, but it is used to allow casting to `u8` in the
+// `Edge::{is_rising, is_falling}` methods. The current implementation of those
+// methods with bit-and tests generates substantially fewer instructions than
+// using `matches!` (see: https://godbolt.org/z/j5fdPfz3c).
 #[repr(u8)]
 pub enum Edge {
     /// The interrupt will trigger on the rising edge only.
--- a/drv/stm32xx-sys-api/src/lib.rs
+++ b/drv/stm32xx-sys-api/src/lib.rs
@@ -34,7 +34,9 @@ pub enum RccError {
 }
 
 /// Configures edge sensitivity for a GPIO interrupt
-#[derive(Copy, Clone, FromPrimitive, PartialEq, Eq, AsBytes)]
+#[derive(
+    Copy, Clone, FromPrimitive, PartialEq, Eq, AsBytes, serde::Deserialize,
+)]
 #[repr(u8)]
 pub enum Edge {
     /// The interrupt will trigger on the rising edge only.
--- a/drv/stm32xx-sys-api/src/lib.rs
+++ b/drv/stm32xx-sys-api/src/lib.rs
@@ -20,6 +20,7 @@ cfg_if::cfg_if! {
 
 use derive_idol_err::IdolError;
 use userlib::*;
+use zerocopy::AsBytes;
 
 pub use drv_stm32xx_gpio_common::{
     Alternate, Mode, OutputType, PinSet, Port, Pull, Speed,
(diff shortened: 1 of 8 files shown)


drv/stm32xx-sys-api/src/lib.rs
```diff
@@ -32,6 +33,18 @@ pub enum RccError {
     NoSuchPeripheral = 1,
 }
 
+/// Configures edge sensitivity for a GPIO interrupt
+#[derive(Copy, Clone, FromPrimitive, PartialEq, Eq, AsBytes)]
+#[repr(u8)]
```
cbiffle, line 38: Nit: not necessary, the use of `FromPrimitive` makes the interface repr-independent.

(If there's some other reason you want this to be `repr(u8)` specifically then carry on.

task/nucleo-user-button/src/main.rs
```diff
@@ -67,10 +67,26 @@ pub fn main() -> ! {
     let user_leds = drv_user_leds_api::UserLeds::from(USER_LEDS.get_task_id());
     let sys = drv_stm32xx_sys_api::Sys::from(SYS.get_task_id());
 
-    let (led, rising, falling) = if let Some(config) = TASK_CONFIG {
-        (config.led, config.rising, config.falling)
+    let (led, sensitivity) = if let Some(Config {
+        rising,
```
cbiffle, line 71: fwiw, if you were to `derive(Deserialize)` on `Edge` you could use the strings "Rising", "Falling", "Both" in the config.

task/nucleo-user-button/src/main.rs
```diff
@@ -67,10 +67,26 @@ pub fn main() -> ! {
...
+        falling,
+        led,
+    }) = TASK_CONFIG
+    {
+        let sensitivity = match (rising, falling) {
+            (true, true) => Edge::Both,
+            (false, true) => Edge::Rising,
+            (true, false) => Edge::Falling,
+            (false, false) => {
```
cbiffle, line 80: (also, you could move this into the build.rs if desired)

(I realize this is a demo task and it doesn't super matter, but demo tasks do tend to get copy-pasted.)

cbiffle (approved):
Mere nits and suggestions below.

task/nucleo-user-button/src/main.rs
hawkw: oh, good call. that would make the code here much simpler!

drv/stm32xx-sys-api/src/lib.rs
hawkw: the `repr(u8)` is used for casting to `u8` in these methods: https://github.com/oxidecomputer/hubris/pull/1706/files#diff-7d4882bb0fb1799e94fee962d08ea3adcf32a27babfd1c42bcefa60846faf897R302-R312

we could implement those using `matches!` instead, or something, but i thought this was cuter --- my naive assumption was that it would also generate less code, but this may not be true on thumb targets...

hawkw: update on that: i stuck it into Godbolt and it looks like the bitwise version does, in fact, generate much less code on `thumbv7m` and `thumbv6m`: https://godbolt.org/z/j5fdPfz3c

cbiffle: Cool -- maybe add a comment to that effect on the repr attribute, because otherwise it might get removed.

bcantrill (approved):
This looks great and will be less error prone to write and easier to review -- thank you!

PR #1709, opened 2024-04-02
userlib: add no-panic feature

cbiffle:
When enabled, this will turn any panics in a task into linker errors. That's not ideal, but it's better than expecting people to manually read disassembly.

This currently has no users in the source tree, because it's factored out of some other work I'm doing. Wanted to make this available so we can start turning it on opportunistically.

### DIFF
--- a/sys/userlib/Cargo.toml
+++ b/sys/userlib/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2021"
 [features]
 default = ["critical-section"]
 panic-messages = []
+no-panic = []
 critical-section = ["dep:critical-section"]
 
 [dependencies]
--- a/sys/userlib/src/lib.rs
+++ b/sys/userlib/src/lib.rs
@@ -1288,6 +1288,15 @@ pub unsafe extern "C" fn _start() -> ! {
     }
 }
 
+// Make the no-panic and panic-messages features mutually exclusive.
+#[cfg(all(feature = "no-panic", feature = "panic-messages"))]
+compile_error!(
+    "Both the userlib/panic-messages and userlib/no-panic feature flags \
+     are set! This doesn't make a lot of sense and is probably not what \
+     you wanted. (If you have a use case for this combination, update \
+     this check in userlib.)"
+);
+
 /// Panic handler for user tasks with the `panic-messages` feature enabled. This
 /// handler will try its best to generate a panic message, up to a maximum
 /// buffer size (configured below).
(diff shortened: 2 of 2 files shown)


sys/userlib/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [features]
 default = ["critical-section"]
 panic-messages = []
+no-panic = []
```
hawkw, line 9: Take it or leave it: it might be nice if the crate's top-level documentation mentioned both of these feature flags and their interactions. 

In particular, I notice that `no-panic` will take priority over `panic-messages`. I think this is probably a fine behavior, but it probably ought to be written down?

Alternatively, we could do something like this:
```rust
#[cfg(all(feature = "panic-messages", feature = "no-panic"))]
compile_error!(
    "both the `userlib/panic-messages` and `userlib/no-panic` feature flags \
    are enabled! this doesn't make a lot of sense and is probably not what \
    you wanted."
);
```

IMO, enabling both these features is _probably_ a mistake, and I might prefer to know that I made the mistake more than I'd want it to still build. On the other hand, because of cross-crate feature unification, it does seem possible to have a situation where both features are enabled for a task, because a dependency enables one of them. But, a library dependency generally shouldn't do that...I dunno.

This might be more relevant if we consider a mechanism for Idol-generated APIs to be marked as non-panicking, so that we can avoid generating retry code for non-idempotentent IPCs served by  tasks that are not able to panic?

hawkw (approved):
This looks good to me overall!

As I mentioned in an inline comment, we might want to consider documenting the interaction between the features that control panics, or make setting both of them an error?

And, I had a couple thoughts about potential future work:

- It might be nice to (eventually) have a way of including some metadata indicating that a task claims it will never panic in the task's build artifact. That way, the supervisor could read that metadata, and if it sees that a task which _claims_ never to panic has, in fact, panicked, it could try to raise a red flag about that. Potentially, this is also something that Humility commands like `humility tasks` could look at --- if `humility` sees a generation count > 0 for a non-panicking task, it could scream about it. But, this is all probably best left for a follow-up.
- It might also be worth investigating a way to make the `idol` code generator aware that a server task for an API won't panic, and if so, avoid generating server death handling code in the client stub. This would generate less code for downstream callers of Idol APIs served by non-panicking servers. However, this is potentially fraught in that not panicking is a characteristic of the server _implementation_ more than of the _IPC interface_; we could potentially have both panicking and non-panicking implementations of the same Idol API. I dunno whether this is actually a good idea or not...

sys/userlib/Cargo.toml
cbiffle: I like the build script suggestion to force them to be exclusive for now, until we find a reason for them not to be.

PR #1712, opened 2024-04-02
stm32xx-sys: add interrupt acking to `gpio_irq_control`

### LINKED ISSUE #1700: dealing with spurious GPIO notifications
While I was working on #1696, @cbiffle pointed out to me that a task that receives a notification mapped to an EXTI GPIO interrupt cannot rely on _just_ the notification bit to determine whether a GPIO pin has actually changed state, because any task can post any notification to any other task. This means that a buggy (or malicious, but what is malicious if not 'buggy, but on purpose?) task may post us the bit corresponding to the EXTI notification by mistake, and the task can't determine whether that notification actually came from the `stm32xx-sys` task. Thus, we must still check what state the GPIO pin is actually in if we want to be resilient against spurious notifications.

I can think of a few ways to deal with this:

1. **Make this the user's responsibility** --- the "null solution". User code is responsible for reading a GPIO pin on receipt of an EXTI interrupt. If we do this, we should probably update documentation and examples so that anyone handling GPIO IRQ notifications knows they should check the pin's state.
2. **Consider a higher-level API in `drv-stm32xx-sys-api`**. We could also consider adding a higher level `wait_for_gpio_pin_state` type API in `drv-stm32xx-sys-api`. This would be implemented on top of the lower-level `gpio_irq_control`  IPCs, and would take a notification mask and GPIO pin set, and provide an API for waiting for that pin to change state, actually check the pin's level, and continue waiting if it hasn't actually transitioned. We could also maybe have a type for doing this, where constructing that type does the appropriate `Sys::gpio_configure_input` and `Sys::gpio_irq_configure` calls before returning the value that allows actually waiting for the pin state change. 

   One potential downside of this approach, though, is that it doesn't compose nicely with code that waits for multiple notifications. For example, in `stm32xx-sprot-server`, we have code that either waits for an EXTI interrupt _or_ a timeout, which couldn't use an API that _just_ waits for a GPIO interrupt. We could potentially also add a variant of a "wait for pin state" function that also takes a timeout...but this still doesn't compose super well in cases where we want to, for example, wait for any of a set of GPIO notifications...

3. **Have `drv-stm32xx-sys` read a pin's state before dispatching an interrupt**. We could also have the `sys` task check whether a pin's level has actually transitioned before dispatching a notification to a user task. However, this only guards against one of the two potential failure modes: it handles the case where a task incorrectly posts the `exti-wildcard-irq` notification to `sys`, but it _does not_ handle the case where a task incorrectly posts the notification mapped to an EXTI interrupt to the end-user task. So, user tasks still have to look at pin states (since they might be notified by a task other than `sys`), and we would now be doing the check twice. I'm not sure whether this is actually all that useful.

There are also potential issues around whether a GPIO pin's state transition triggers an IRQ and then it goes _back_ to the previous state between when the `sys` task is notified and when the end-user task is notified. This is more of a problem for IRQs raised by stuff like buttons than for IRQs raised by a different MCU or similar device, where the IRQ line is often used as a "handshake" that only gets unset when the device receiving the IRQ line actually performs some operation. I'm not sure whether we should be concerned about this case or not...

### LINKED ISSUE #1714: stm32xx-sys `Sys::gpio_irq_control` arguments are easily swapped
The `drv-stm32xx-sys` IPC `Sys::gpio_irq_control(disable_mask, enable_mask)` interface is a bit confusing: it takes two notification masks, one to enable and the other to disable. The only way to distinguish these masks at the callsite is based on the order of the arguments in the function call. This means that someone reading or writing code that calls this API must remember the order of the arguments, and it's easy to get them backwards without knowing, resulting in buggy code.

We might want to consider changing this API to take a struct containing both masks as an argument, instead. This way, code that calls this IPC writes out the words `enable` and `disable`, and it's obvious which mask is which, regardless of ordering. Compare the following:

```rust
// current interface:
sys.gpio_irq_control(0, notifications::MY_NOTIF_MASK).unwrap_lite();

// possible new interface:
sys
   .gpio_irq_control(IrqMasks { disable: 0, enable: notifications::MY_NOTIF_MASK })
   .unwrap_lite();
```

Passing a struct rather than two positional arguments makes it easier to tell which is which, and prevents the caller from accidentally swapping them.

### Potential downsides

- The proposed new API is a bit more verbose than the old one --- you have to write the name of the struct, and import it. Personally, this is a cost I am quite willing to pay to reduce the risk of accidentally writing buggy code, but others may not agree.
- The proposed new API is less consistent with the existing `sys_irq_control` syscall in userlib, which takes arguments in order, rather than as a struct.  I do think consistency with existing APIs is important, especially because this `sys` IPC can be thought of as a relative of the kernel's `irq_control` syscall. 
  + We could consider also changing the `userlib::sys_irq_control` function to take an arguments struct as well (it could even be the same struct used by `Sys::gpio_irq_control`). The actual syscall would remain the same, just the wrapper function in `userlib` would take a struct. This is a much more painful breaking API change, though, since _lots_ of code uses `userlib::sys_irq_control`, while the only currently existing user of `Sys::gpio_irq_control` is the Nucleo user-button demo task...

hawkw:
Currently, there is no mechanism for a task woken by a notification
mapped to an EXTI GPIO interrupt to determine whether the pin it cares
about has actually changed state, rather than receiving a notification
posted by some other task (see #1700). In order to make this possible,
@cbiffle [suggested] that the `stm32xx-sys` task provide an IPC
interface that a task receiving an EXTI notification can call to ack
that the IPC actually was caused by a pin state transition.

This branch extends the `Sys::gpio_irq_control` IPC to return a bool if
any EXTI IRQs mapped to the notifications in a task's notification
maskhave triggered since the last time `gpio_irq_control` was called.

Adding this functionality to `gpio_irq_control` --- which a task that
wishes to receive another EXTI notification must call anyway --- allows
us to keep the number of IPC roundtrips the same when a task receives
such notifications in a loop. In order to support other cases, where a
task only cares about one such notification, the `gpio_irq_control`
interface has been changed to determine whether to enable or disable the
IRQs in the mask based on a second argument, rather than taking two
masks as it did previously, in order to allow a third operation which
*only* checks the state of the interrupt, leaving it enabled if it has
not been triggered but not re-enabling it if it has. We no longer have
the ability to both enable and disable disjunct sets of IRQs in one IPC,
but in practice, no task we have currently even *uses* more than one
GPIO IRQ anyway. And, getting rid of the double-mask interface also
fixes #1714.

I've implemented this by tracking a 16-bit bitmap in the `Sys` task that
mirrors the EXTI pending register. This is because `sys` must clear the
pending bit in order to receive another interrupt from _any_ EXTI
source.

Closes #1700
Closes #1714

[suggested]:
    https://github.com/oxidecomputer/hubris/issues/1700#issuecomment-2030458073

### DIFF
--- a/drv/stm32xx-sys-api/src/lib.rs
+++ b/drv/stm32xx-sys-api/src/lib.rs
@@ -68,7 +68,7 @@ pub enum IrqControl {
     ///
     /// If an interrupt is currently enabled, it will remain enabled, while if
     /// it is currently disabled, it will remain disabled.
-    Check = 0b11,
+    Check,
 }
 
 impl Sys {
--- a/drv/stm32xx-sys-api/src/lib.rs
+++ b/drv/stm32xx-sys-api/src/lib.rs
@@ -52,6 +52,25 @@ pub enum Edge {
     Both = 0b11,
 }
 
+/// Describes which operation is performed by the [`Sys::gpio_irq_control`] IPC.
+#[derive(
+    Copy, Clone, FromPrimitive, PartialEq, Eq, AsBytes, serde::Deserialize,
+)]
+// repr attribute is required for the derived `AsBytes` implementation
+#[repr(u8)]
+pub enum IrqControl {
+    /// Disable any interrupts mapped to the provided notification mask.
+    Disable = 0,
+    /// Enable any interrupts mapped to the provided notification mask.
+    Enable,
+    /// Check if any interrupts mapped to the provided notification mask have
+    /// been triggered, *without* enabling or disabling the interrupt.
+    ///
+    /// If an interrupt is currently enabled, it will remain enabled, while if
+    /// it is currently disabled, it will remain disabled.
+    Check = 0b11,
+}
+
 impl Sys {
     /// Requests that the clock to a peripheral be turned on.
     ///
@@ -328,6 +347,22 @@ impl core::ops::BitOr for Edge {
             (Edge::Falling, Edge::Falling) => Edge::Falling,
             _ => Edge::Both,
         }
+    }
+}
+
+impl From<bool> for IrqControl {
+    fn from(value: bool) -> Self {
+        if value {
+            IrqControl::Enable
+        } else {
+            IrqControl::Disable
+        }
+    }
+}
+
+impl From<Option<bool>> for IrqControl {
+    fn from(value: Option<bool>) -> Self {
+        value.map(Self::from).unwrap_or(Self::Check)
     }
 }
 
--- a/drv/stm32xx-sys/src/main.rs
+++ b/drv/stm32xx-sys/src/main.rs
@@ -250,12 +250,13 @@
 //!         Edge::Rising, // or `Edge::Falling`, `Edge::Both`
 //!     );
 //!
+//!     // First, enable the interrupt, so that we can receive our
+//!     // notification:
+//!     sys.gpio_irq_control(0, notifications::MY_GPIO_NOTIFICATION_MASK)
+//!         .unwrap_lite();
+//!
 //!     // Wait to recieve notifications for our GPIO interrupt in a loop:
 //!     loop {
-//!         // First, enable the interrupt, so that we can receive our
-//!         // notification:
-//!         sys.gpio_irq_control(0, notifications::MY_GPIO_NOTIFICATION_MASK);
-//!
 //!         // Wait for a notification:
 //!         //
 //!         // We only care about notifications, so we can pass a zero-sized
@@ -268,12 +269,17 @@
 //!         // Recv from the kernel never returns an error.
 //!         .unwrap_lite();
 //!
-//!         // Now, do ... whatever it is this task is supposed to do when
-//!         // the interrupt fires.
-//!         handle_interrupt();
-//!
-//!         // When the loop repeats, the call to `sys.gpio_irq_control()` will
-//!         // enable the interrupt again.
+//!         // Call `sys.gpio_irq_control()` to both re-enable the interrupt
+//!         // and ask the `sys` task to confirm for us that the interrupt has
+//!         // actually fired.
+//!         let fired = sys
+//!             .gpio_irq_control(0, notifications::MY_GPIO_NOTIFICATION_MASK)
+//!             .unwrap_lite()
+//!         if fired {
+//!             // If the sys task confirms that our interrupt has fired, do...
+//!             // whatever it is this task is supposed to do when that happens.
+//!             handle_interrupt();
+//!         }
 //!     }
 //! }
 //!```
@@ -547,6 +553,12 @@ struct ServerImpl<'a> {
     #[cfg(feature = "exti")]
     exti: &'a device::exti::RegisterBlock,
 
+    /// A bitfield tracking which EXTI interrupts have fired since the last time
+    /// their owners have called the `gpio_irq_control` IPC. This is necessary
+    /// as we must unset the sending bit in the *real* EXTI_CPUPR1 register on
+    /// receipt of an interrupt in order to receive another one, but we must
+    /// hang onto the pending state until the task that actually uses that
+    /// interrupt asks us for it.
     #[cfg(feature = "exti")]
     exti_cpupr_2: u16,
 }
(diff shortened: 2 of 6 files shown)


hawkw:
Hmm, so, after trying to integrate this change with my WIP branch #1696 (see https://github.com/oxidecomputer/hubris/pull/1696/commits/c29e5ef39ca20115a42e4cb896f43e45adbe92f8), I'm not actually sure if combining the "enable/disable IRQ" and "check if IRQ was triggered" behaviors in the same API is _always_ the Right Thing. 

It's nice for cases where code intends to receive an interrupt in a loop, as in the Nucleo user button demo: https://github.com/oxidecomputer/hubris/blob/1d44ab86a094793fa6a61d15c3c713a682c0410f/task/nucleo-user-button/src/main.rs#L88-L126 or (presumably) @bcantrill's change to record MAX5970 data when it raises its alert line, since such code will always _want_ to unmask the interrupt to receive it again after being notified, and it's nice for that code to only have to do one IPC round-trip, rather than two.

However, for code that only wants to wait for a _single_ interrupt and then go back to doing something else, such as `wait_rot_irq` in `drv-stm32h7-sprot-server`, the fact that checking whether an IRQ was triggered can only be done as part of an operation that either enables it or disables it is actually a bit awkward. In `stm32h7-sprot-server`, we also serve an Idol IPC interface, and doing this requires us to _occasionally_ wait for the IRQ. But, we don't want the notification to go off when we're _not_ in the code that actively waits for the interrupt, since we're trying to serve an IPC interface, and we don't want the code waiting for the interrupt to accidentally consume an IRQ notification that was posted a long time ago but not consumed. 

In that case, what we really want is a way to check if the IRQ was triggered, *leaving it disabled* if it was triggered, but leaving it *enabled* if the notification was spurious. Right now, because we don't have that, we always perform the check by disabling the IRQ, and then re-enable it if it was not actually received. This means we actually have to do *more* IPCs than we would otherwise. 

I think the interface from https://github.com/oxidecomputer/hubris/pull/1712/commits/a6c9163bbe55ea455881bf6f1b880dae4b1f4303, where I had a new `Sys::gpio_irq_ack` IPC that was separate from `Sys::gpio_irq_control`, was actually better for this use case. That design took a notification mask and a flag indicating whether to re-enable the IRQ if it had actually triggered, and returned whether it had triggered. This works well for both the "wait for multiple notifications in a loop" use-case, where we pass `true` to the re-enable argument, and in the "one-shot wait for a single notification" use case, where we can pass `false` to the re-enable argument, leaving the IRQ enabled if it didn't trigger, but *not* re-enabling it if it did. With this design, both the loop and one-shot use cases only need to perform one IPC, which seems ideal. 

If we want to go back to that design, we could still hang this off of `gpio_irq_control`, or go back to having an additional IPC. 

If we wanted to continue using `gpio_irq_control`, the potential approaches I can think of are:
-  adding a third argument of a mask of notifications to check but not change the enabled/disabled state of (eww, three arguments, all of which are masks...)
- special-casing notifications that appear in both the enable and disable masks to just leave them alone but check their state (which seems weird and hard to document)

On the other hand, an additional IPC has the downside of "adding another IPC", complicating the interface and probably resulting in more code, but I think it results in a somewhat less weird API surface. And, it has the side benefit of letting us leave the `gpio_irq_control` IPC marked as idempotent, if we didn't make it consume the "interrupt pending" state anymore.

@cbiffle, I'd love to get your opinion on this (and maybe also @bcantrill's, as a potential consumer of this API).

drv/stm32xx-sys/src/main.rs
```diff
@@ -690,17 +715,16 @@ impl idl::InOrderSysImpl for ServerImpl<'_> {
 
                             if mask & enable_mask != 0 {
                                 // Record that these bits meant something.
-                                used_bits |= mask;
+                                let bit = 1 << i;
```
cbiffle, line 718: Ought to be able to factor `bit` here out (it's duplicated on line 701)

cbiffle (approved):
Generally looks great! One opportunity for deduplication noted below.

I wish I could think of a better pattern to show in examples than `unwrap_lite` here, because I know people are just going to copy it, which will generate panic sites that aren't probably what they actually want. But I'm not sure what to suggest in the general case, so, this is probably a fine start.

cbiffle:
Restarting the client is generally a reliable way of ensuring things get reconfigured, and can sometimes be the right call -- however, it's also quite expensive, and we've had cases of programs relying on it for mundane errors. It's also a bit of a fault-multiplier in that, if the task in question is also a server (and most are), it propagates the panic out to _its_ callers, etc.

If we can conveniently show how to retry, that would be neat; if it would be a pain and make the example less useful, I'd say don't worry about it.

drv/stm32xx-sys/src/main.rs
hawkw: I wanted to avoid doing the bit shift in the case where we don't do anything with the entry. But, bit shifts are cheap...right?

drv/stm32xx-sys-api/src/lib.rs
```diff
@@ -52,6 +52,25 @@ pub enum Edge {
...
+    /// Disable any interrupts mapped to the provided notification mask.
+    Disable = 0,
+    /// Enable any interrupts mapped to the provided notification mask.
+    Enable,
+    /// Check if any interrupts mapped to the provided notification mask have
+    /// been triggered, *without* enabling or disabling the interrupt.
+    ///
+    /// If an interrupt is currently enabled, it will remain enabled, while if
+    /// it is currently disabled, it will remain disabled.
+    Check = 0b11,
```
cbiffle, line 71: The custom discriminator here is surprising; was this deliberate?

hawkw: i...don't remember typing that, actually. it must have been a copy-paste error. i agree that it's not necessary.

cbiffle (approved):

PR #1715, opened 2024-04-03
Convenient no-panic API for just waiting on notifications

cbiffle:
This is an alternate API for accessing the common "receive only kernel
notifications" pattern. That specific use of the sys_recv syscall is
defined as not being able to fail. Different users have handled that in
different ways -- some panic, some ignore the result. The latter embeds
the assumption of that behavior in many different places in the
codebase, a pattern I'm trying to keep an eye on.

This change adds a new sys_recv_notification userlib function that
bundles up this common pattern. Because of its central placement in
userlib, it seems like a more appropriate place to encode the assumption
about the syscall not failing, and so I've done that, using
unreachable_unchecked to explain it to the compiler.

This removes a panic site from every notification-wait in programs that
were previously using unwrap or unwrap_lite.

### DIFF
--- a/sys/userlib/src/lib.rs
+++ b/sys/userlib/src/lib.rs
@@ -237,12 +237,14 @@ unsafe extern "C" fn sys_send_stub(_args: &mut SendArgs<'_>) -> RcLen {
 /// let it, but it always receives _something_.
 #[inline(always)]
 pub fn sys_recv_open(buffer: &mut [u8], notification_mask: u32) -> RecvMessage {
-    // The open-receive version of the syscall is defined as being unable to
-    // fail, and so we should always get a success here. (This is not using
-    // `unwrap` because that generates handling code with formatting.)
     match sys_recv(buffer, notification_mask, None) {
         Ok(rm) => rm,
-        Err(_) => panic!(),
+        Err(_) => {
+            // Safety: the open-receive version of the syscall is defined as
+            // being unable to fail in the kernel ABI, so this path can't happen
+            // modulo a kernel bug.
+            unsafe { core::hint::unreachable_unchecked() }
+        }
     }
 }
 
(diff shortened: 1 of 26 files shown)


sys/userlib/src/lib.rs
```diff
@@ -259,20 +261,32 @@ pub fn sys_recv_open(buffer: &mut [u8], notification_mask: u32) -> RecvMessage {
...
 ) -> Result<RecvMessage, ClosedRecvError> {
-    sys_recv(buffer, notification_mask, Some(sender))
-        .map_err(|_| ClosedRecvError::Dead)
+    sys_recv(buffer, notification_mask, Some(sender)).map_err(|code| {
+        // We're not using the extract_new_generation function here because
+        // that has a failure code path for cases where the code is not a
+        // dead code. In this case, sys_recv is defined as being _only_
+        // capable of returning a dead code -- otherwise we have a serious
+        // kernel bug. So to avoid the introduction of a panic that can't
+        // trigger, we will do this manually:
```
hawkw, line 278: This comment is great, thanks for adding it!

sys/userlib/src/lib.rs
```diff
@@ -319,6 +334,25 @@ pub fn sys_recv(
...
+    match sys_recv(&mut [], notification_mask, Some(TaskId::KERNEL)) {
+        Ok(rm) => {
+            // The notification bits come back from the kernel in the operation
+            // code field.
+            rm.operation
+        }
+        Err(_) => {
+            // Because we passed Some(TaskId::KERNEL), this is defined as not
+            // being able to happen.
+            unsafe { core::hint::unreachable_unchecked() }
```
hawkw, line 351: tiny nit: should this be 
```suggestion
            // Safety: Because we passed Some(TaskId::KERNEL), this is defined
            // as notbeing able to happen.
            unsafe { core::hint::unreachable_unchecked() }
```

hawkw (approved):
This is lovely, I'm looking forward to using it!

PR #1717, opened 2024-04-03
stm32xx-sys: fix typo in EXTI pin configuration

cbiffle:
In the Giant Scary Nested Match we had to write to be able to address the EXTI register array as an array through the PAC, I typo'd the number base being used to select the field within the register, by reusing 2 from the line above. This is the _base 2 logarithm_ of the number we actually wanted, 4, because the operations are shift and mod, respectively.

Wheeee

Anyway, this fixes the situation where EXTI configuration of any pin p such that p % 4 = 2 or 3 on any port other than A mysteriously senses a different pin in port A instead.

### DIFF
--- a/drv/stm32xx-sys/src/main.rs
+++ b/drv/stm32xx-sys/src/main.rs
@@ -398,7 +398,7 @@ fn main() -> ! {
                 // Process entries that are filled in...
                 if let &Some(ExtiDispatch { port, .. }) = entry {
                     let register = i >> 2;
-                    let slot = i % 2;
+                    let slot = i & 0b11;
 
                     // This is an array of 4-bit fields spread across 4 32-bit
                     // registers. We're indexing them with i. There is really no


drv/stm32xx-sys/src/main.rs
```diff
@@ -398,7 +398,7 @@ fn main() -> ! {
                 // Process entries that are filled in...
                 if let &Some(ExtiDispatch { port, .. }) = entry {
                     let register = i >> 2;
-                    let slot = i % 2;
+                    let slot = i & 0b11;
```
hawkw, line 401: i waffled on whether `& 0b11` or `% 4` expresses intent more clearly --- i'm now mildly leaning towards `% 4` based on the observation that there are 4 bitfields per register. but, it doesn't actually matter.

I _do_ think that it might be nice if the `register` and `slot` calculation was below the comment describing what we're doing --- it's possible i didn't notice this in the review because i was too top-to-bottom in my reading habits, and didn't think to go back and look at the calculations after reading the comment explaining what they were doing. but, whether or not that would have prevented this is probably not really worth thinking too hard about :woman_shrugging:

hawkw (approved):
welp. good catch! assuming you've tested that looking at Pn14 on a Nucleo no longer results in us getting IRQs for `SWD_CLK`, this looks good to me!

PR #1720, opened 2024-04-04
build/i2c: generate range matches

cbiffle:
We've noticed the compiler tends to generate better code for integer range matches of the form

    match x {
        0..=7 => a,
        7..=10 => b,
        _ => c,
    }

as compared to

    match x {
        0
        | 1
        | 2
        | 3
        | 4
        | 5
        | 6
        | 7 => a,
        // and so on
    }

Also, Clippy in recent nightly has started objecting to the "cascade of digits" form. Rather than suppressing that warning, I thought I'd convert the biggest matches in our codegen output to use ranges.

This commit uses rangemap to automatically consolidate adjacent ranges of integers into a convenient match statement. I've also factored out the code from the two places it was used.


build/i2c/src/lib.rs
```diff
@@ -1770,3 +1747,33 @@ pub fn device_descriptions() -> impl Iterator<Item = I2cDeviceDescription> {
...
+{
+    for (controller, indices) in source {
+        let indices = indices
+            .iter()
+            .map(|&i| i..i + 1)
+            .collect::<RangeSet<usize>>();
+        let s: Vec<String> = indices
+            .iter()
+            .map(|range| format!("{}..={}", range.start, range.end - 1))
+            .collect::<_>();
```
hawkw, line 1767: nit, take it or leave it: why not just put the `.join(...)` here and have `s` be a string rather than a vec of strings?

hawkw (approved):

build/i2c/src/lib.rs
cbiffle: Done!

PR #1727, opened 2024-04-04
control-plane-agent: finish ComponentUpdater

cbiffle:
This change causes SpUpdate to impl ComponentUpdater, instead of sporting functions that look a lot _like_ ComponentUpdater but aren't actually. This fixes a warning in the newest toolchain, which was concerned that operations in the non-public trait were going unused (which they were!).

In order to adapt ComponentUpdater to be able to describe all three cases, I have added two associated types to it. UpdatePrepare parameterizes the difference between RoT/HostFlash updates (which want ComponentUpdatePrepare) and SP updates (which want a different SpUpdatePrepare type). SubComponent models the fact that only the SP seems to want a specifier for sub-components inside itself; the others set this to ().


jgallagher (approved):
LGTM - thanks!

PR #1729, opened 2024-04-05
transceivers-server: remove illusory loop

cbiffle:
Clippy pointed out that this loop can't, well, loop. At all.

I'm kind of surprised earlier versions missed this.


hawkw (approved):
Wow, the way GitHub's diff viewer renders this diff makes it totally impossible to tell what's going on! But, this seems correct to me.

Aaron-Hartwig (approved):
Huh, that loop sure doesn't loop. Thanks clippy and Cliff!

PR #1732, opened 2024-04-05
Add enums for LPC55 pin codegen

mkeeter:
This means that invalid pin configuration parameters will fail to parse, instead of making it through codegen and then failing at (Hubris) compile time.

I don't love leaning on the derived `Debug` here:
```rust
let mode = format_ident!("{}", format!("{:?}", self.get_mode()));
let slew = format_ident!("{}", format!("{:?}", self.get_slew()));
let invert = format_ident!("{}", format!("{:?}", self.get_invert()));
let digimode = format_ident!("{}", format!("{:?}", self.get_digimode()));
let od = format_ident!("{}", format!("{:?}", self.get_opendrain()));
tokens.append_all(final_pin);
tokens.append_all(quote::quote! {
    AltFn::#alt_num,
    Mode::#mode,
    Slew::#slew,
    Invert::#invert,
    Digimode::#digimode,
    Opendrain::#od,
});
```
but my other experiments were all worse; I'm open to suggestions for how to make this cleaner!

### DIFF
--- a/build/lpc55pins/src/lib.rs
+++ b/build/lpc55pins/src/lib.rs
@@ -40,11 +40,16 @@ impl ToTokens for Pin {
 pub struct PinConfig {
     pin: Pin,
     alt: usize,
-    mode: Option<Mode>,
-    slew: Option<Slew>,
-    invert: Option<Invert>,
-    digimode: Option<Digimode>,
-    opendrain: Option<Opendrain>,
+    #[serde(default)]
+    mode: Mode,
+    #[serde(default)]
+    slew: Slew,
+    #[serde(default)]
+    invert: Invert,
+    #[serde(default)]
+    digimode: Digimode,
+    #[serde(default)]
+    opendrain: Opendrain,
     direction: Option<Direction>,
     name: Option<String>,
 }
@@ -77,18 +82,18 @@ pub enum Invert {
 
 #[derive(Copy, Clone, Debug, Default, Deserialize)]
 #[serde(rename_all = "lowercase")]
-pub enum Opendrain {
+pub enum Digimode {
     #[default]
-    Normal,
-    Opendrain,
+    Digital,
+    Analog,
 }
 
 #[derive(Copy, Clone, Debug, Default, Deserialize)]
 #[serde(rename_all = "lowercase")]
-pub enum Digimode {
-    Analog,
+pub enum Opendrain {
     #[default]
-    Digital,
+    Normal,
+    Opendrain,
 }
 
 #[derive(Copy, Clone, Debug, Deserialize)]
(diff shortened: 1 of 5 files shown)


build/lpc55pins/src/lib.rs
```diff
@@ -98,12 +127,13 @@ impl ToTokens for PinConfig {
...
-        let invert = format_ident!("{}", self.get_invert());
-        let digimode = format_ident!("{}", self.get_digimode());
-        let od = format_ident!("{}", self.get_opendrain());
 
+        let mode = format_ident!("{}", format!("{:?}", self.get_mode()));
+        let slew = format_ident!("{}", format!("{:?}", self.get_slew()));
+        let invert = format_ident!("{}", format!("{:?}", self.get_invert()));
+        let digimode =
+            format_ident!("{}", format!("{:?}", self.get_digimode()));
+        let od = format_ident!("{}", format!("{:?}", self.get_opendrain()));
```
hawkw, line 136: can't these just be
```suggestion
        let mode = format_ident!("{:?}", self.get_mode());
        let slew = format_ident!("{:?}", self.get_slew());
        let invert = format_ident!("{:?}", format!("{:?}", self.get_invert());
        let digimode =
            format_ident!("{:?}", self.get_digimode());
        let od = format_ident!("{:?}", self.get_opendrain());
```
AFAICT, we shouldn't need the second string allocation?

hawkw: also, it occurs to me an alternative to using the `Debug` impl could be implementing `ToTokens` for the various enums, but...that would probably just internally use `syn::Ident::new` or something, which would end up looking similarish. But, you could hard-code the string, which is probably either better or worse...

hawkw: Oh, huh, I was apparently wrong about how `format_ident!` works!

I suppose another option would just be to write
```rust
syn::Ident::new(format!("{:?}", self.get_mode()), proc_macro2::Span::call_site())
```
to avoid the second allocation, although this is, admittedly, uglier...

mkeeter: Yeah, I'm on the fence here; if all solutions are equally ugly, then might as well pick the shorter one.

hawkw: Yeah...the impact on build time is probably not meaningful. Carry on.

build/lpc55pins/src/lib.rs
```diff
@@ -36,53 +36,82 @@ impl ToTokens for Pin {
...
+}
+
 impl PinConfig {
-    fn get_mode(&self) -> String {
-        match &self.mode {
-            None => "NoPull".to_string(),
-            Some(s) => s.to_string(),
-        }
+    fn get_mode(&self) -> Mode {
+        self.mode.unwrap_or(Mode::NoPull)
```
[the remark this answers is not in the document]
mkeeter: Nice, that comes out slightly shorter

PR #1734, opened 2024-04-05
stm32xx-sys: generate constants for EXTI pins

hawkw:
Currently, pins used for EXTI IRQs are defined in the `sys` task's config in `app.toml`. When a task wishes to actually use those pins, though, it must _also_ define them separately in order to call `sys.gpio_configure_input` (and if it wants to perform its own GPIO reads on that pin). This is generally done either by hard-coding the port and pin number in the task's source code, or by defining its own config for configuring which pin to use.

This is unfortunate, as it results in a duplicated pin definition, either between the source code and the config if hard-coded, or twice in the config, if the task also gets the definition from config. And, if the task configures the pin in its own config, the author of that task must re-implement the config handling.

This commit fixes this by adding new code generation to `drv-stm32xx-sys-api` to generate a module called `gpio_irq_pins` containing `PinSet` constants with the name of the IRQ pin and the pin number and port defined in the `sys` task's config section. This way, the same pin set used by `sys` when configuring the EXTI interrupt can be referenced by the task that consumes the interrupt, removing the need to duplicate the pin definition. I've edited the `nucleo-user-button` demo task to demonstrate using this.

### DIFF
--- a/build/stm32xx-sys/src/lib.rs
+++ b/build/stm32xx-sys/src/lib.rs
@@ -2,14 +2,16 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use anyhow::Context;
 use quote::{quote, TokenStreamExt};
 use serde::Deserialize;
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, io::Write};
 
 #[derive(Deserialize, Default)]
-#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
 pub struct SysConfig {
     /// EXTI interrupts
+    #[serde(default)]
     gpio_irqs: BTreeMap<String, GpioIrqConfig>,
 }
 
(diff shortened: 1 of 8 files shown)


build/stm32xx-sys/src/lib.rs
```diff
@@ -2,22 +2,24 @@
...
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use anyhow::Context;
 use quote::{quote, TokenStreamExt};
 use serde::Deserialize;
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, io::Write};
 
 #[derive(Deserialize, Default)]
-#[serde(rename_all = "kebab-case", deny_unknown_fields)]
```
cbiffle, line 10: ?

cbiffle (approved):

build/stm32xx-sys/src/lib.rs
hawkw: Using `build_util::other_task_config` requires removing the `deny_unknown_fields`, because the current implementation of that function attempts to interpret _every_ task's configs as this type, which panics.

PR #1747, opened 2024-04-11
kerncore: Add tests for bad or uncontained memory contained in a contiguous slice

aapoalas:
The recent Oxide and Friends episode mentioned this memory region spanning slices issue, and the code was interesting. I thought I saw a chance to optimise it by omitting the start address containment checking once the start region is found. (Based on benchmarking against the test impls, it isn't faster.) The obvious issue with that was of course validity of the intermediate regions but I realised that wouldn't be a problem since I still need to check the validity of each region between start and end.

I figured I'd write tests to show that the alternative solution works but then realised the actual issue: My "optimised" algorithm allows for uncontained (unmapped?) memory to be accessed as long as it is found between two good regions. I presume this makes my improvement worthless, but at least the tests are worthwhile additions. The second added test shows the error in my idea.

I've still left the original modification in this PR's commits to at least show what the idea was. There may still be some worth to it.

Side note: The idea for the two loops came from `Vec::retain_mut`'s code which uses a similar logic to find the first item that isn't retained.

### DIFF
--- a/sys/kerncore/src/lib.rs
+++ b/sys/kerncore/src/lib.rs
@@ -441,4 +441,38 @@ mod tests {
             "should NOT be able to access slice that starts and ends in good ranges but passes through bad one, but can",
         );
     }
+
+    #[test]
+    fn cannot_access_slice_spanning_over_uncontained_memory() {
+        let region_table = vec![
+            TestRegion {
+                base: 0x1238_5678,
+                size: 0x0001_0000,
+                label: "good".to_string(),
+            },
+            TestRegion {
+                // 65kb separated from previous region
+                base: 0x123A_5678,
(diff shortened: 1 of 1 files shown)


aapoalas:
After running some benchmarks, my version does not outperform the original loop even in its best-case scenario which is also probably highly unlikely: A long slice spanning over multiple contiguous good regions.

I'll revert the algorithm changes from the PR and leave behind the tests.

cbiffle:
Thanks for the tests! (And the experimentation on the algorithm, even if it wasn't clearly a win just yet.)

I've applied a couple small fixes and intend to merge this.

cbiffle (approved):

PR #1748, opened 2024-04-15
Enable buildomat

labbott:


.github/buildomat/jobs/sample.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+#:
+#: name = "build-and-test (ubuntu-22.04)"
+#: variety = "basic"
+#: target = "ubuntu-22.04"
+#: rust_toolchain = "nightly-2024-04-04"
```
hawkw, line 6: do you know if there's a way for this to be sourced from the `rust-toolchain.toml` file?

if not, it might be worth adding a comment in that file saying "hey, you have to update the Buildomat job definition if you change this". maybe also worth an upstream Buildomat feature request for `rust-toolchain.toml` awareness.

.github/buildomat/jobs/sample.sh
hawkw, line 1: since this actually does do a real build (gimlet-f), maybe it's worth naming the file something other than `sample`? not a big deal though, we can always rename it later.

cbiffle (approved):

.github/buildomat/jobs/sample.sh
labbott: decided to drop adding the job here. once buildomat is looking for jobs we can add a single job and see how it blends^Wworks

PR #1751, opened 2024-04-16
Can we build it?

labbott:


.github/buildomat/jobs/gimlet-e.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+#:
+#: name = "build-and-test gimlet-e"
+#: variety = "basic"
+#: target = "ubuntu-22.04"
+#: output_rules = [
+#:      "=/work/hubris/*.zip",
+#: ]
+
+exec .github/buildomat/build-one.sh app/gimlet/rev-e.toml gimlet-e default
```
labbott, line 10: Borrowed this technique of putting a common script from omicron

PR #1752, opened 2024-04-17
Begin removing/disabling wildcard imports.

cbiffle:
This starts with the "core" code in sys, test, and jefe, and works out from there.

Any package can opt into this at any time by adding this to its Cargo.toml:

    [lints]
    workspace = true

This will also cause it to pick up any future lints we decide on.

### DIFF
--- a/lib/multitimer/src/lib.rs
+++ b/lib/multitimer/src/lib.rs
@@ -213,8 +213,8 @@ mod fakes {
     use core::cell::Cell;
 
     thread_local! {
-        pub static CURRENT_TIME: Cell<u64> = Cell::new(0);
-        pub static TIMER_SETTING: Cell<(Option<u64>, u32)> = Cell::default();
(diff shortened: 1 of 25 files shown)


lib/multitimer/src/lib.rs
```diff
@@ -213,8 +213,8 @@ mod fakes {
     use core::cell::Cell;
 
     thread_local! {
-        pub static CURRENT_TIME: Cell<u64> = Cell::new(0);
-        pub static TIMER_SETTING: Cell<(Option<u64>, u32)> = Cell::default();
+        pub static CURRENT_TIME: Cell<u64> = const { Cell::new(0) };
+        pub static TIMER_SETTING: Cell<(Option<u64>, u32)> = const { Cell::new((None, 0)) };
```
cbiffle, line 217: (turns out we're not running clippy on multitimer, so these lines had warnings.)

hawkw (approved):

mkeeter (approved):

PR #1757, opened 2024-04-19
MB85RSxx FRAM bringup

hawkw:
This branch adds an initial implementation of a driver for the Fujitsu
MB85xx series of ferroelectric RAM (FRAM) devices, as seen in Power
Shelf Controller rev C boards. The FRAM chip is a SPI device with a
fairly simple protocol, detailed in [the Fujitsu datasheet][1]. It also
has additional pins for toggling the write protection feature (which can
also be controlled over SPI) and a "hold" pin to maintain a read/write
transaction when CS is deasserted, but my driver doesn't currently use
those pins because we don't *really* need them.

I've also included a simple demo task which writes some data to a FRAM
device and checks whether it can be read back. Eventually, I'd like to
replace this with a more general-purpose "read the contents of the FRAM"
task that can be used for debugging purposes, but I've left it as is for
now in the interests of getting the driver implementation ready to
merge. I'll revisit this later.

[1]:
    https://www.mouser.com/datasheet/2/1113/MB85RS64T_DS501_00051_2v0_E-2329177.pdf


drv/mb85rsxx-fram/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+        //  0x01 & 0b0001_1111 = 1
+        //  2^1 = 2
+        //  2 * 1024 = 2048 bytes
+        //
+        // Or, for the 8kb MB85RS64V and MB85RS64T, which have product IDs
+        // 0x0302 and 0x2303, respectively:
+        //  0x03 & 0b0001_1111 = 3
+        //  0x23 & 0b0001_1111 = 3
+        //  2^3 = 8
+        //  8 * 1024 * 8 = 8192 bytes
```
aapoalas, line 483: **nitpick:** I guess there's an extra ` * 8` here.

aapoalas (commented):
I'm not sure if outside reviews are welcome here: My apologies if I'm just being a nuisance.

I just wanted to say that this was really nice and easy to read, and all the code looked good to me.

drv/mb85rsxx-fram/src/lib.rs
hawkw: ah, thanks --- good catch!

hawkw:
> I just wanted to say that this was really nice and easy to read, and all the code looked good to me.

Thanks! I try :)

drv/mb85rsxx-fram/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+    const NEEDS_THREE_BYTE_ADDRS: bool = Self::SIZE > 64 * KIB;
+
+    /// Constructs a new `Fram` driver for the given SPI device.
+    ///
+    /// This function reads the FRAM's product ID and checks that it is the
+    /// device the driver thinks it's talking to, returning an error if it's not.
+    pub fn new(spi: SpiDevice<S>) -> Result<Self, FramInitError> {
+        // Look at the FRAM's ID to make sure it's the device we expect it to
+        // be. In particular, make sure it's the size we think it is.
+        let id = FramId::read(&spi).map_err(FramInitError::SpiError)?;
```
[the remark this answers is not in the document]
hawkw: Per the FRAM datasheets, I don't believe there is such a state; at the very least, the status register only contains bits related to various forms of write protection (block-level, global, and status-register write protect). There isn't anything in the datasheet related to a busy state, which I think indicates that the FRAM is fast enough that we don't need to worry about waiting for operations to complete. It's certainly not exposed in the protocol.

drv/mb85rsxx-fram/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+/// latch when it's dropped. This way, the FRAM remains in the write-protected
+/// state when you're not actively trying to write to it.
+///
+/// To write to the FRAM chip, use [`WritableFram::write`]. This type also
+/// exposes a [`WritableFram::read`] method, so the FRAM chip may also be read
+/// from while the write enable latch is set.
+#[must_use = "a WritableFram does nothing if constructed but not read from \
+    or written to"]
+pub struct WritableFram<'fram, S: SpiServer, const ID: u16>(
+    &'fram Fram<S, { ID }>,
```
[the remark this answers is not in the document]
hawkw: Yeah...my initial design for this took the `Fram` by value, but unfortunately, since `WritableFram` has a drop impl and `SpiDevice` is not `Copy` or `Clone`, there isn't a nice  way to get the `Fram` back _out_ of the `WritableFram` besides using `ManuallyDrop`, which I didn't really want to do...but maybe I should have. I'll change it, thanks.

drv/mb85rsxx-fram/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+        /// Block protect 0
+        const BP0 = 1 << 2;
+        /// Block protect 1
+        const BP1 = 1 << 3;
+        /// Status register write protect enabled
+        const WPEN = 1 << 7;
+    }
+}
+
+#[derive(Eq, PartialEq, Copy, Clone, counters::Count)]
```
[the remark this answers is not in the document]
hawkw: So, hmm. I found it quite useful for debugging the driver implementation, but I agree that it may not be that useful for debugging code depending on this. I do, however, think it's quite nice to have separate ringbuf events for "starting a read/write" and "did a read/write", so that there's a way to see if the FRAM driver is stuck while waiting for a read/write operation to complete (which is why the wrappers exist). But, higher-level code can do that.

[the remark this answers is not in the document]
hawkw: (i've since removed the ringbuf)

drv/mb85rsxx-fram/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+    pub const MB85RS1MT: u16 = 0x2703;
+    /// 256kb Fujitsu FRAM
+    pub const MB85RS2MTA: u16 = 0x4803;
+    /// 512kb Fujitsu FRAM
+    pub const MB85RS4MT: u16 = 0x4903;
+
+    /// Returns the size in bytes of the FRAM chip, based on its product ID.
+    pub(super) const fn size(product_id: u16) -> usize {
+        // The first 5 bits of the product ID give the density of the FRAM chip
+        // in multiples of 2KiB.
```
aapoalas, line 410: **question:** At the risk of sounding stupid: Is "multiples of" correct? I would read eg. 0x3 to become 3 * 2 KiB, not the 2 ^ 3 * 1 KiB that it actually means.

hawkw: yeah, i think i should have said..."exponents of two", or something, here? but "exponents of two KiB" also sounds wrong, it sounds like it's saying `2KiB ^ 3`, which is also not what it is...

aapoalas: Maybe "gives us the kilobyte density of [] in powers of two" or "gives us the density of [] in kilobyte powers of two"?

drv/mb85rsxx-fram/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+            // manufacturers' FRAM chips that use the same protocol do? Would
+            // have to read some more datasheets to be sure.
+            u16::from_be_bytes([mfg_id, cont])
+        };
+
+        Ok(Self { mfg_id, product_id })
+    }
+}
+
+impl<S: SpiServer, const ID: u16> Drop for WritableFram<'_, S, { ID }> {
```
mkeeter, line 369: Nit: move this to L338, so we're not interleaving `WritableFram` and `FrameId` impl blocks?

drv/mb85rsxx-fram/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+        // Or, for the 8kb MB85RS64V and MB85RS64T, which have product IDs
+        // 0x0302 and 0x2303, respectively:
+        //  0x03 & 0b0001_1111 = 3
+        //  0x23 & 0b0001_1111 = 3
+        //  2^3 = 8
+        //  8 * 1024 = 8192 bytes
+        const MASK: u8 = 0b0001_1111;
+        let [hi, _] = u16::to_be_bytes(product_id);
+        let density = hi & MASK;
+        2usize.pow(density as u32) * super::KIB
```
mkeeter, line 426: Weird personal opinion: even though `pow` is more mathematically obvious, I typically see conversion from log2 size → actual size written as a shift, e.g. `(1usize << density as u32)`

mkeeter (approved):

PR #1762, opened 2024-04-23
Enable notifications on any closed RECV.

### LINKED ISSUE #1761: Notifications should be able to break through closed RECV.
Currently, the reference manual says this about closed-RECV operations:

> During a closed receive, a task will receive messages only from the chosen task. The task will not receive notifications unless the chosen sender ID is the kernel’s task ID, 0xFFFF. (This behavior is a little odd because it predates notification masks, and may change.)

As prophesied, I would like to change this.

Currently we are in the odd situation where entering a closed-RECV with a non-zero notification mask just ignores the notification mask -- it's treated as equivalent to a notification mask of zero. We got here because of some implementation details in the _very early_ kernel ABI (like "first month" early).

This seems awkward. It would probably be better if it were orthogonal:
- Open RECV accepts messages from any task, Closed RECV accepts messages from only one task. (The kernel's virtual task ID would still be valid here, to filter out messages from _all_ user tasks.)
- Notification mask non-zero accepts (some subset of) notifications, zero does not.

Moreover, this may allow some useful behaviors that aren't currently possible. As I noted in #1654 I've been taking a hard look at our closed-RECV operation and thinking about how it interacts with priority. Currently, once a server enters a closed-RECV, it has no way of breaking back out -- timers won't work, interrupts won't work, etc. If we changed closed-RECV to honor the notification mask, it would let servers arrange to break out of a closed-RECV, which would allow them to do all sorts of useful things:
- Process periodic work despite waiting on a client
- Handle interrupts
- Impose a timeout and REPLY-FAULT a client that is taking too long

(I don't think imposing a timeout like that should happen very often, since timeouts are how we make flaky software, but it would be nice to have the ability!)

cbiffle:
Previously, a closed RECV against the kernel was special-cased as a "listen only for notifications" operation. Closed RECV against any other task would ignore the provided notification mask and treat it as zero.

This seemed weird and limiting.

This commit alters the semantics of RECV with a non-zero notification mask and enables notifications in that specific situation.

As far as I can tell, we have no code that relies on notifications being filtered out during closed-RECV, so this should be safe.

Fixes #1761.

### DIFF
--- a/doc/syscalls.adoc
+++ b/doc/syscalls.adoc
@@ -232,9 +232,10 @@ During an open receive, a task may receive messages sent by any other task, plus
 any notifications enabled by the notification mask.
 
 During a closed receive, a task will receive messages only from the chosen task.
-The task will *not* receive notifications unless the chosen sender ID is the
-kernel's task ID, `0xFFFF`. (This behavior is a little odd because it predates
-notification masks, and may change.)
+The task will still receive any notifications set in its notification mask.
+
+To listen *only* for notifications, a task can perform a closed receive against
+the kernel's task ID, `0xFFFF`.
 
 ==== Arguments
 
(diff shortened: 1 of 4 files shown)


hawkw (approved):
Looks straightforward to me!

PR #1765, opened 2024-04-30
PSC monitoring MVP

cbiffle:
This adds very basic monitoring of the presence and power-good lines from the PSUs at the PSC. It can manage remove/insertion of power supplies, delay restarts to ensure we don't cycle things too quickly, and respond to faults by resetting things. So far data is only logged to a ringbuf.

### DIFF
--- a/app/psc/base.toml
+++ b/app/psc/base.toml
@@ -6,7 +6,7 @@ fwid = true
 
 [kernel]
 name = "psc"
-requires = {flash = 32768, ram = 4880}
+requires = {flash = 32768, ram = 5080}
 features = ["dump"]
 
 [caboose]
@@ -56,6 +56,42 @@ port = "E"
 pin = 3
 owner = {name = "sprot", notification = "rot_irq"}
 
+# Note: present_l IRQs are missing because of a conflict with ROT_IRQ above;
+# these signals are currently polled.
+
+[tasks.sys.config.gpio-irqs.psu_pwr_ok_1]
+port = "J"
+pin = 6
+owner = {name = "sequencer", notification = "psu_pwr_ok_1"}
+
+[tasks.sys.config.gpio-irqs.psu_pwr_ok_2]
+port = "J"
+pin = 7
+owner = {name = "sequencer", notification = "psu_pwr_ok_2"}
+
+[tasks.sys.config.gpio-irqs.psu_pwr_ok_3]
+port = "J"
+pin = 8
+owner = {name = "sequencer", notification = "psu_pwr_ok_3"}
+
+[tasks.sys.config.gpio-irqs.psu_pwr_ok_4]
+port = "J"
+pin = 9
+owner = {name = "sequencer", notification = "psu_pwr_ok_4"}
+
+[tasks.sys.config.gpio-irqs.psu_pwr_ok_5]
+port = "J"
+pin = 10
+owner = {name = "sequencer", notification = "psu_pwr_ok_5"}
+
+[tasks.sys.config.gpio-irqs.psu_pwr_ok_6]
+port = "J"
+pin = 11
+owner = {name = "sequencer", notification = "psu_pwr_ok_6"}
+
+
+
+
 [tasks.i2c_driver]
 name = "drv-stm32xx-i2c-server"
 stacksize = 1048
@@ -86,6 +122,15 @@ priority = 4
 stacksize = 4096
 start = true
 task-slots = ["jefe", "packrat", "i2c_driver", "sys"]
+notifications = [
+    "psu_pwr_ok_1",
+    "psu_pwr_ok_2",
+    "psu_pwr_ok_3",
+    "psu_pwr_ok_4",
+    "psu_pwr_ok_5",
+    "psu_pwr_ok_6",
+    "timer",
+]
 
 [tasks.update_server]
 name = "stm32h7-update-server"
--- a/app/psc/base.toml
+++ b/app/psc/base.toml
@@ -119,24 +119,34 @@ task-slots = ["i2c_driver"]
 name = "task-net"
 stacksize = 8000
 priority = 4
-features = ["mgmt", "h753", "psc", "vlan", "vpd-mac", "use-spi-core", "spi2"]
+features = ["mgmt", "h753", "psc", "vlan", "vpd-mac", "spi2"]
 max-sizes = {flash = 131072, ram = 65536, sram1 = 16384}
 sections = {eth_bulk = "sram1"}
-uses = ["eth", "tim16", "spi2"]
+uses = ["eth", "tim16"]
 start = true
 notifications = [
     "eth-irq",
     "mdio-timer-irq",
     "wake-timer",
-    "spi-irq",
     "jefe-state-change",
 ]
-task-slots = ["sys", "packrat", "jefe"]
+task-slots = ["sys", "packrat", "jefe", {spi_driver = "spi2_driver"}]
 
 [tasks.net.interrupts]
 "eth.irq" = "eth-irq"
 "tim16.irq" = "mdio-timer-irq"
-"spi2.irq" = "spi-irq"
+
+[tasks.spi2_driver]
+name = "drv-stm32h7-spi-server"
+priority = 3
+features = ["spi2", "h753"]
+uses = ["spi2"]
+start = true
+interrupts = {"spi2.irq" = "spi-irq"}
+stacksize = 872
+task-slots = ["sys"]
+notifications = ["spi-irq"]
+
 
 [tasks.control_plane_agent]
 name = "task-control-plane-agent"
--- a/drv/psc-seq-server/src/main.rs
+++ b/drv/psc-seq-server/src/main.rs
@@ -3,6 +3,102 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 //! Server for managing the PSC sequencing process.
+//!
+//!
+//! # General notes on PSC power supply sequencing
+//!
+//! There are rules to follow here to avoid glitching the power supplies,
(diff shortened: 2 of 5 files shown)


drv/psc-seq-server/src/main.rs
```diff
@@ -3,48 +3,586 @@
...
+                    None
+                }
+            }
+
+            // yay!
+            (PsuState::Present(PresentState::On), _, Status::Good) => None,
+
+            (PsuState::Present(PresentState::On), _, Status::NotGood) => {
+                // The PSU appears to have pulled the OK signal into the "not
+                // OK" state to indicate an internal fault!
```
hawkw, line 509: _very_ nitpicky, to the extent that i wasn't sure if i should even mention this: elsewhere, you've put the comment describing what case the states represent _above_ the match arm, but here it's under it:
```suggestion
            // The PSU appears to have pulled the OK signal into the "not  OK" 
            // state to indicate an internal fault!
            (PsuState::Present(PresentState::On), _, Status::NotGood) => {
```

drv/psc-seq-server/src/main.rs
```diff
@@ -3,48 +3,586 @@
...
+                        Status::Good => {
+                            self.state = PsuState::Present(PresentState::On);
+                            Some(ActionRequired::EnableMe)
+                        }
+                    }
+                } else {
+                    // Remain in this state.
+                    None
+                }
+            }
```
hawkw, line 546: take it or leave it: I kind of feel like this might be a little bit easier to follow if we pulled these out into separate match arms? not a huge deal though, it's a marginal improvement at best.
```suggestion
            // Do it all again then, I guess.
            (
                PsuState::Present(PresentState::Faulted { turn_on_deadline }),
                _,
                Status::NotGood,
            ) if turn_on_deadline <= now => {
                let turn_on_deadline =
                    now.wrapping_add(FAULT_OFF_MS);
                self.state =
                    PsuState::Present(PresentState::Faulted {
                        turn_on_deadline,
                    });
                Some(ActionRequired::NoteStickyFault)
            },

            // The PSU has indicated that it is once again good!
            (
                PsuState::Present(PresentState::Faulted { turn_on_deadline }),
                _,
                Status::Good,
            ) if turn_on_deadline <= now => {
                self.state = PsuState::Present(PresentState::On);
                Some(ActionRequired::EnableMe)
            }

            // Remain in this state.
            _ => None
```

drv/psc-seq-server/src/main.rs
```diff
@@ -3,48 +3,586 @@
...
+    On,
+
+    /// The PSU has just appeared and we're waiting a bit to confirm that it's
+    /// stable before turning it on. (Waiting in this state provides some
+    /// debouncing for contact scrape.)
+    NewlyInserted { settle_deadline: u64 },
+    /// The PSU has unexpectedly deasserted the OK signal, or failed to assert
+    /// it within a reasonable amount of time after being turned on.
+    Faulted { turn_on_deadline: u64 },
+}
```
hawkw, line 597: nit, take it or leave it: I might consider defining these enums _before_ the logic that operates on them, rather than after --- the documentation here is valuable context when reading the implementations of those methods, so it might be nice if a top-to-bottom reader encountered it first? not a big deal though.

drv/psc-seq-server/src/main.rs
```diff
@@ -3,48 +3,586 @@
...
+    state: PsuState,
+}
+
+impl Psu {
+    /// Advances the PSU management state machine given the current time (`now`)
+    /// and the state of the `present` and `pwr_ok` inputs.
+    ///
+    /// This may be called at unpredictable intervals, and may be called more
+    /// than once for the same timestamp value. The implementation **must** use
+    /// `now` and the timer to control any time-sensitive operations.
```
hawkw, line 447: i kind of wonder if an ascii graphic flowchart depicting these state transitions might be useful?

app/psc/base.toml
```diff
@@ -119,24 +164,34 @@ task-slots = ["i2c_driver"]
...
+[tasks.spi2_driver]
+name = "drv-stm32h7-spi-server"
+priority = 3
+features = ["spi2", "h753"]
+uses = ["spi2"]
+start = true
+interrupts = {"spi2.irq" = "spi-irq"}
+stacksize = 872
+task-slots = ["sys"]
+notifications = ["spi-irq"]
```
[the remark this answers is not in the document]
cbiffle: Yeah I don't understand the TOML stuff well enough to do that. I can check with Matt.

drv/psc-seq-server/src/main.rs
```diff
@@ -3,48 +3,597 @@
...
+//! unusual. The PSU drives this open drain, so it will only go low if the PSU
+//! pulls it low to indicate a problem. This implies that, if the PSU is not
+//! detected as present, you cannot trust the `OK` signal.
+//!
+//!
+//! # Intended behavior
+//!
+//! The PSC is intended to be hot swappable. If the PSC gets plugged in, this
+//! task will start anew. It will also start anew if power is applied to a
+//! previously unpowered rack.
```
labbott, line 36: The "this task will start anew" seems to be in conflict with the sentence below "This task is not intended to restart under normal operation". I _think_ this paragraph is describing cold boot behavior vs below is runtime?

app/psc/base.toml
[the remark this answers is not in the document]
hawkw: It's not a big deal either way, just occurred to me that we could save some space on the downrev boards by avoiding the separate task. 🤷‍♀️

drv/psc-seq-server/src/main.rs
```diff
@@ -3,48 +3,586 @@
...
 task_slot!(JEFE, jefe);
 task_slot!(PACKRAT, packrat);
 
+#[derive(Copy, Clone, PartialEq, Eq)]
+enum Trace {
+    Empty,
+    /// Emitted at task startup when we find that a power supply is probably
+    /// already on. (Note that if the power supply is not present, we will still
+    /// detect it as "on" due to the pull resistors.)
+    FoundEnabled { psu: usize },
```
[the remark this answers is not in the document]
cbiffle: Fair point, fixed!

drv/psc-seq-server/src/main.rs
cbiffle: Interesting, these two classes of comments are very different things in my head, but not in a way I can easily explain or make actionable for future maintainers.

Basically, there's one class of comment explaining the _pattern,_ and one class explaining what to do after the pattern matches. e.g. the ones above the match arms are generally explaining why a field is removed or wildcard, the ones below talk strategy.

But I get that it just looks inconsistent, so I can make them all the same.

drv/psc-seq-server/src/main.rs
cbiffle: Hmmmm I'm not sure I feel the same on this.

I do think _all_ of these match arms were easier to read before rustfmt made them v
e
r
t
i
c
a
l
:\

drv/psc-seq-server/src/main.rs
cbiffle: Done!

drv/psc-seq-server/src/main.rs
hawkw: Ah, interesting, I see what you mean! To me, the comments on the patterns seemed to describe, in a general sense, "what this match arm represents", so this felt more like that. But, it's up to you --- do whatever you think is best!

drv/psc-seq-server/src/main.rs
cbiffle: Yeah, basically -- the PSC getting plugged in means _power is applied._ I have attempted to clarify this.

cbiffle:
Fixes pushed atop the original commit, PTAL

drv/psc-seq-server/src/main.rs
hawkw: My thought process was that to me, when the match arm captured the entire condition that represented a particular state transition, rather than having additional branching logic inside it, you could view the match as a mapping from "all the inputs" to "all the outputs", which felt maybe slightly more comprehensible to _my brain_ in particular...but I think this is a matter of taste, and I'm fine with the way you have it.

drv/psc-seq-server/src/main.rs
```diff
@@ -3,48 +3,603 @@
...
+                }
+                Some(ActionRequired::DisableMe { attempt_snapshot }) => {
+                    if attempt_snapshot {
+                        // TODO snapshot goes here
+                    }
+                    ringbuf_entry!((
+                        now,
+                        Trace::Disabling {
+                            psu: i as u8,
+                            present: attempt_snapshot,
```
hawkw, line 443: this is _very_ nitpicky, but it feels like the naming of this variable seems to be a bit in conflict over whether it represents a _fact_ about the state of the world ("is the PSU `present`?') or a _policy_ about whether or not to perform a particular action ("should we `attempt_snapshot`?"). 

it feels like there is a potential for future issues if we were to later change `step` to decide whether or not to attempt a snapshot based on more factors than _just_ whether the PSU is present, without updating this code. if we were to ever return `attempt_snapshot: false` for a present PSU for some reason, the ringbuffer would incorrectly indicate that the PSU is not present.

just a thought.

hawkw (approved):
no blockers from me!

drv/psc-seq-server/src/main.rs
cbiffle: I think it'll be clearer once I finish and merge the PMBus snapshot code, but I may rename it based on your suggestion at that point. The two names are a result of two different use cases:
- It was going to be named `attempt_snapshot` internally to reflect the implementation intent, but then
- I recognized that that name wasn't very useful in a ringbuf and came up with a different one for logging there. (This one's the user-visible one.)

PR #1766, opened 2024-05-01
Make enum encoded FPGA register fields more flexible

Aaron-Hartwig:
I'll start off with a bit of context since this is a corner of the system that doesn't get looked at often. We use SystemRDL to define registers in the FPGA. When we build an FPGA, one of the outputs is a JSON file that describes the registers. We commit that JSON file alongside the bitfile in this repo, which then gets turned into Rust by `build/fpga-regmap`. A neat part of this is that we can define enums as a register field in the RTL and carry them into the Rust code via a property we call `encode`.

For example, consider the [`STATUS_PORT0` register](https://github.com/oxidecomputer/hubris/blob/master/drv/sidecar-front-io/sidecar_qsfp_x32_controller_regs.json#L1154-L1213). Today, that would generate the following code:
```rust
...
    #[allow(non_snake_case)]
        pub mod STATUS_PORT0 {
            #[allow(dead_code)]
            #[allow(non_upper_case_globals)]
            pub const ERROR: u8 = 0b00001111;

            use num_derive::{ToPrimitive, FromPrimitive};
            #[derive(Copy, Clone, Eq, PartialEq, FromPrimitive, ToPrimitive)]
            #[allow(dead_code)]
            #[allow(non_camel_case_types, clippy::upper_case_acronyms)]
            pub enum Encoded {
                NoError = 0x00,
                NoModule = 0x01,
                NoPower = 0x02,
                PowerFault = 0x03,
                NotInitialized = 0x04,
                I2cAddressNack = 0x05,
                I2cByteNack = 0x06,
            }
            #[allow(dead_code)]
            #[allow(non_upper_case_globals)]
            pub const BUSY: u8 = 0b00010000;
        }
...
```

The presence of `encode` in the JSON yields an `Encoded` enum for the register in the Rust. This has a couple of sharp corners, both of which come from behavior that assumes an `encode` type is the only field in a register.

1. You can't have two `encoded` fields within the same register because we currently just name the enum `Encoded`.
2. If you `encoded` field lives alongside any other fields, such as is the case with `BUSY` and `ERROR` above, you need to make sure you're masking off the bits for the `encoded` field only when attempting to resolve to an `Encoded` variant.

This PR addresses both of these points by:

1. Naming the enum after the field it is encoding, i.e. `<field>_Encoded`. 
2. Implementing `TryFrom<u8>` for the enum and having that automatically handle the bit masking for the field so the programmer doesn't have to remember that.

After this PR, that would look like:
```rust
...
        #[allow(non_snake_case)]
        pub mod PORT0_STATUS {
            #[allow(dead_code)]
            #[allow(non_upper_case_globals)]
            pub const ERROR: u8 = 0b00001111;

            #[derive(Copy, Clone, Eq, PartialEq)]
            #[allow(dead_code)]
            #[allow(non_camel_case_types, clippy::upper_case_acronyms)]
            pub enum ERROR_Encoded {
                NoError = 0x00,
                NoModule = 0x01,
                NoPower = 0x02,
                PowerFault = 0x03,
                NotInitialized = 0x04,
                I2cAddressNack = 0x05,
                I2cByteNack = 0x06,
            }

            impl TryFrom<u8> for ERROR_Encoded {
                type Error = ();
                fn try_from(x: u8) -> Result<Self, Self::Error> {
                    use crate::Reg::QSFP::PORT0_STATUS::ERROR_Encoded::*;
                    let x_masked = x & ERROR;
                    match x_masked {
                        0x00 => Ok(NoError),
                        0x01 => Ok(NoModule),
                        0x02 => Ok(NoPower),
                        0x03 => Ok(PowerFault),
                        0x04 => Ok(NotInitialized),
                        0x05 => Ok(I2cAddressNack),
                        0x06 => Ok(I2cByteNack),
                        _ => Err(()),
                    }
                }
            }

            #[allow(dead_code)]
            #[allow(non_upper_case_globals)]
            pub const BUSY: u8 = 0b00010000;
        }
...
```

### DIFF
--- a/build/fpga-regmap/src/lib.rs
+++ b/build/fpga-regmap/src/lib.rs
@@ -136,7 +136,15 @@ impl Addr {{
 
 ////////////////////////////////////////////////////////////////////////////////
 
-fn write_reg_fields(children: &[Node], prefix: &str, output: &mut String) {
+fn write_reg_fields(
+    parents: Vec<String>,
+    children: &[Node],
+    prefix: &str,
+    output: &mut String,
+) {
+    // We need this to implement the u8 -> enum conversation
+    let parent_chain = parents.join("::");
+
     for child in children.iter() {
         if let Node::Field {
             inst_name,
@@ -155,19 +163,22 @@ fn write_reg_fields(children: &[Node], prefix: &str, output: &mut String) {
 {prefix}        pub const {inst_name}: u8 = 0b{mask:08b};",
             )
             .unwrap();
+
             // Deal with optional encoded Enums on this field
             match encode {
                 Some(x) => {
+                    let encode_name = inst_name.clone() + "_Encoded";
                     writeln!(
                         output,
                         "
-{prefix}        use num_derive::{{ToPrimitive, FromPrimitive}};
-{prefix}        #[derive(Copy, Clone, Eq, PartialEq, FromPrimitive, ToPrimitive)]
+{prefix}        #[derive(Copy, Clone, Eq, PartialEq)]
 {prefix}        #[allow(dead_code)]
 {prefix}        #[allow(non_camel_case_types, clippy::upper_case_acronyms)]
-{prefix}        pub enum Encoded {{"
+{prefix}        pub enum {encode_name} {{"
                     )
                     .unwrap();
+
+                    // unpack all of the enum variant -> u8 information
                     for item in x {
                         writeln!(
                             output,
@@ -176,7 +187,41 @@ fn write_reg_fields(children: &[Node], prefix: &str, output: &mut String) {
                         )
                         .unwrap();
                     }
+
                     writeln!(output, "{prefix}        }}").unwrap();
+
+                    // We want to implement TryFrom<u8> rather than From<u8>
+                    // because the u8 -> enum conversion can fail if the value
+                    // is not a valid enum variant. Additionally, we mask off
+                    // the supplied u8 with the mask of the field so encoded
+                    // fields could be colocated in a register with other fields
+                    writeln!(
+                        output,
+                        "
+{prefix}        impl TryFrom<u8> for {encode_name} {{
+{prefix}            type Error = ();
+{prefix}            fn try_from(x: u8) -> Result<Self, Self::Error> {{
+{prefix}                use crate::{parent_chain}::{encode_name}::*;
+{prefix}                let x_masked = x & {inst_name};
+{prefix}                match x_masked {{"
+                    )
+                    .unwrap();
+                    for item in x {
+                        writeln!(
+                            output,
+                            "{prefix}                    {1:#04x} => Ok({0}),",
+                            item.name, item.value
+                        )
+                        .unwrap();
+                    }
+                    writeln!(
+                        output,
+                        "{prefix}                    _ => Err(()),
+{prefix}                }}
+{prefix}            }}
+{prefix}        }}\n"
+                    )
+                    .unwrap();
                 }
                 None => {}
             }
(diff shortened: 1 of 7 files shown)


build/fpga-regmap/src/lib.rs
```diff
@@ -155,19 +163,22 @@ fn write_reg_fields(children: &[Node], prefix: &str, output: &mut String) {
 {prefix}        pub const {inst_name}: u8 = 0b{mask:08b};",
             )
             .unwrap();
+
             // Deal with optional encoded Enums on this field
             match encode {
                 Some(x) => {
+                    let encode_name = inst_name.clone() + "_Encoded";
```
hawkw, line 170: nitpicky, take it or leave it: the use of `+` on strings is pretty uncommon in ideomatic Rust. i probably would've written this, instead:
```suggestion
                    let encode_name = format!("{inst_name}_Encoded");
```
as a side note, this has the advantage of only allocating once, while `inst_name.clone() + "_Encoded"` will (likely) allocate twice, once for the clone and then again if the cloned string doesn't have enough capacity to add "_Encoded". this doesn't actually matter very much here, i'm sure that the impact of a second allocation on build times is basically un-noticeable, but i felt like i'd mention it anyway.

build/fpga-regmap/src/lib.rs
```diff
@@ -186,7 +231,13 @@ fn write_reg_fields(children: &[Node], prefix: &str, output: &mut String) {
...
 }
 
-fn write_node(node: &Node, prefix: &str, output: &mut String) {
+fn write_node(
+    parents: Vec<String>,
+    node: &Node,
+    prefix: &str,
+    output: &mut String,
+) {
+    let mut new_parents = parents.clone();
```
hawkw, line 240: AFAICT, this function never references `parents` again after cloning it. Since we take `parents` by value, we can just push the new parents into `parents`, because we have exclusive access to it --- I think the clone isn't necessary here, unless I've missed something.

hawkw: alternatively, since only some cases in `write_node` actually need to push to parents, the most efficient thing would be to change `write_node` to take `parents: &Vec<String>`,  remove the clone in `recurse_reg_map` and pass it by reference, and then only clone `parents` in the cases where we need to push a new parent to pass to a function (so that we avoid the unconditional clone in the `Node::Mem` case).

again, optimizing for performance here is not terribly important, but...that's what i'd do, so i figured i'd suggest it.

build/fpga-regmap/src/lib.rs
```diff
@@ -155,19 +163,22 @@ fn write_reg_fields(children: &[Node], prefix: &str, output: &mut String) {
...
                 Some(x) => {
+                    let encode_name = inst_name.clone() + "_Encoded";
                     writeln!(
                         output,
                         "
-{prefix}        use num_derive::{{ToPrimitive, FromPrimitive}};
-{prefix}        #[derive(Copy, Clone, Eq, PartialEq, FromPrimitive, ToPrimitive)]
+{prefix}        #[derive(Copy, Clone, Eq, PartialEq)]
 {prefix}        #[allow(dead_code)]
 {prefix}        #[allow(non_camel_case_types, clippy::upper_case_acronyms)]
```
hawkw, line 176: Also, re: naming, I kind of wonder if maybe we should just downcase the register name and emit `RegisternameEncoded`, instead of `REGISTERNAME_Encoded`? Up to you, though.

hawkw (approved):
Overall, this looks good to me! I had some minor style suggestions, but they're not blockers --- the core change seems correct!

build/fpga-regmap/src/lib.rs
Aaron-Hartwig: Thanks! Feel free to nit away, there's always stuff to learn!

build/fpga-regmap/src/lib.rs
Aaron-Hartwig: That makes sense to me. I shall make it so!

build/fpga-regmap/src/lib.rs
Aaron-Hartwig: I think this is a good suggestion to make this whole crate feel much more Rust-y rather than the SHOUTING AT US.

PR #1768, opened 2024-05-02
Add support for {Read,Write}NoAddrIncr SPI operations, leverage them in `transceivers`

Aaron-Hartwig:
Part 1) Adjustments to `fpga-api/src/lib.rs`, `fpga-server/src/main.rs`, and `fpga.idol` for the new SPI operations.

As of https://github.com/oxidecomputer/quartz/pull/139 the FPGA's SPI peripheral supports these operations. Prior to this PR the FPGA drivers did not take an operation argument for read operations because there was only a single type of read. Now that there are two (`Read` and `ReadAddrNoIncr`), the intended operation must be specified. Given that up until this point `Read` was the only option (and will probably remain the primary use 99.9% of the time) that has been made the default for the `read` operation which the majority of callers use. To utilize the new read operation, one can just call the underlying `read_bytes` function which takes a `ReadOp` parameter.

Part 2) Making use of the new SPI operations in the `transceivers` code

This also implements their use for transceivers I2C buffers for the QSFP FPGA design. All the `*_regs.*` files are generated from the FPGA build process and are brought in alongside the `.bit` file. The transceivers code has been reworked to leverage the SPI refactoring from https://github.com/oxidecomputer/quartz/pull/146. This code utilizes the non-address incrementing SPI operations to read/write to the I2C buffers in the FPGA.


drv/fpga-api/src/lib.rs
```diff
@@ -118,14 +118,28 @@ pub enum BitstreamType {
...
+    BitClear = 3
+    NoOp = 4
+    WriteNoAddrIncr = 5
+    ReadNoAddrIncr = 6
+
+    The {Write,Read}NoAddrIncr operations support sequential write/read to the
+    same address. These operations are useful for exposing FIFO-like interfaces
+    via a single memory-mapped address location.
+*/
+
```
cbiffle, line 135: I feel like the removed comment a couple lines down, explaining why the enum only contains a subset of the operations, was kinda useful. Adding a single line comment to each op enum saying something like

`/// This is just reads; for writes, see WriteOp`

would help avoid confusing readers, I think.

cbiffle (approved):

drv/sidecar-front-io/src/transceivers.rs
```diff
@@ -1025,114 +1025,86 @@ impl Transceivers {
...
+                }
+            } else {
+                error |= port.as_mask();
+            }
         }
-    }
+        // success is wherever we did not encounter an `FpgaError`
+        let success = mask & !error;
+        // only have an error where there was a requested module in mask
+        error &= mask;
```
hawkw, line 1105: I'm not totally sure if I get this --- IIUC, since we only write to the FPGA for ports in the mask, we shouldn't have to mask out the error bits again, right? Or am I missing something?

Aaron-Hartwig: You have not missed anything and are correct! This was a bit of copy/pasta on my part when reworking some similar code where the initial `error` mask could not be constructed intelligently (by port) and therefore I masked it off later. I'll clean this up!

Aaron-Hartwig:
My latest push rebases this, fixes [an FPGA bug](https://github.com/oxidecomputer/quartz/pull/146/commits/eb6c5e67e3270d98093c2877cb6cb937c684fc78), and addresses the review comments. Thank you both very much! I'll be queuing this up for auto-merge.

PR #1770, opened 2024-05-02
Remove array_methods flag

steveklabnik:
This *may* need to go in "nightly features that our dependencies use", as it doesn't appear in our code. Let's try that out!


labbott (approved):

steveklabnik:
I was looking over Hubris' unstable flags the other day, was like "hmm, what do we use array_methods for?" and then found this PR, that I forgot about. Given @labbott 's approval, I'm gonna merge; if there's problems, should be exceedingly easy to revert.

PR #1774, opened 2024-05-06
Fix various cfg-related crimes

cbiffle:
The nightly toolchain started doing checking for cfgs on Sunday, so I bumped our toolchain revision locally to see what it'd find.

In short, it got big mad.

This PR fixes All The Things. The commits provide detailed descriptions, but here's a summary:

- Boards (as in `target_board`) now have to be defined somewhere. I've added a `boards` directory containing (currently empty) config files per board. **Any reference to a `target_board` without a corresponding file will now be treated as a typo.**
- Renamed some boards that only mentioned their processor type and not the board itself.
- Removed vestiges of the `psc-a` and `sidecar-a` boards, which saw firmware support ended a while ago, but were never fully removed from the code.
- Removed the final bits of STM32H7B3 devel board support, something we haven't used in ~4 years at this point.
- Removed dead code implementing a `traptrace` feature in Sidecar, which isn't actually in the Cargo.toml and thus can't be built.
- Fixed `stm32xx-i2c`, which was attempting to key off of `target_board` without actually exposing the cfg in its build script. **Forgetting to do this is now a compiler warning.**
- Fixed `jefe`, which was attempting to key off of `armv8m` without actually exposing the cfg in its build script. **Also now a warning.**
- Fixed `build/i2c` to use its own features instead of generating code that tried to match on features in its _clients,_ which was working by accident.

In order to formally move us to a toolchain that prevents this from regressing, I had to do some small unrelated fixes, namely:

- Stop using the `core::uXX::MAX` constants, which are deprecated now.
- Bump a couple deps because their `derive` macros produced code the new compiler dislikes.
- Add a missing `syn` feature in `derive-idol-err`, which had apparently been picking it up from some dependency.


build/util/src/lib.rs
```diff
@@ -85,6 +90,47 @@ pub fn target_board() -> Option<String> {
...
+            "warning: boards directory not found, \
+                  target_board can't validate"
+        );
+    }
+    boards.sort();
+    boards.dedup();
+
+    let values = boards.join(",");
+
+    println!("cargo:rustc-check-cfg=cfg(target_board, values({values}))");
```
mkeeter, line 133: it took me a while to figure out what was going on here, but it's amazing

drv/stm32xx-sys-api/src/h7.rs
```diff
@@ -57,9 +57,7 @@ pub enum Peripheral {
     OctoSpi2 = periph(Group::Ahb3, 19), // B3 only
     Sdmmc1 = periph(Group::Ahb3, 16),
 
-    #[cfg(feature = "h7b3")]
-    OctoSpi1 = periph(Group::Ahb3, 14), // B3 only
-    #[cfg(any(feature = "h743", feature = "h747", feature = "h753"))]
+    #[cfg(any(feature = "h743", feature = "h753"))]
```
mkeeter, line 60: Is `any(feature = "h743", feature = "h753")` guaranteed to be true?  Now that we've removed `h7b3` support, `h743` and `h753` are the only remaining `h7*` features, so I think we can remove this condition (and others below).

drv/stm32xx-sys-api/src/h7.rs
```diff
@@ -158,12 +151,9 @@ pub enum Peripheral {
     Swp = periph(Group::Apb1H, 2),
     Crsen = periph(Group::Apb1H, 1),
 
-    #[cfg(feature = "h7b3")]
-    Dfsdm1 = periph(Group::Apb2, 30), // B3 differs from 43/47
-
     Hrtim = periph(Group::Apb2, 29), // 43/47 only
```
mkeeter, line 154: I'm confused by the comments in this file – there are many lines which was `43/47 only` but aren't gated on `h743`.

(we could also probably remove the references to the `B3` and `47` more generally, since they're both unsupported)

README.mkdn
```diff
@@ -152,9 +152,9 @@ application described by the TOML file.
 - `cargo xtask dist app/demo-stm32f4-discovery/app.toml` - stm32f4-discovery
 - `cargo xtask dist app/demo-stm32f4-discovery/app-f3.toml` - stm32f3-discovery
 - `cargo xtask dist app/lpc55xpresso/app.toml` - lpcxpresso55s69
-- `cargo xtask dist app/demo-stm32g0-nucleo/app-g031.toml` - stm32g031
-- `cargo xtask dist app/demo-stm32g0-nucleo/app-g070.toml` - stm32g070
-- `cargo xtask dist app/demo-stm32g0-nucleo/app-g0b1.toml` - stm32g0b1
+- `cargo xtask dist app/demo-stm32g0-nucleo/app-g031.toml` - stm32g031-nucleo
+- `cargo xtask dist app/demo-stm32g0-nucleo/app-g070.toml` - stm32g070-nucleo
+- `cargo xtask dist app/demo-stm32g0-nucleo/app-g0b1.toml` - stm32g0b1-nucleo
```
mkeeter, line 157: Naming nit: should `nucleo` be at the start or end?  We have both `stm32g0b1-nucleo` and `nucleo-ih743zi2` board names.

task/power/src/main.rs
```diff
@@ -387,11 +387,7 @@ macro_rules! mwocp68_controller {
     path = "bsp/gimlet_bcdef.rs"
 )]
 #[cfg_attr(
-    any(
-        target_board = "psc-a",
-        target_board = "psc-b",
-        target_board = "psc-c"
-    ),
+    any(target_board = "psc-b", target_board = "psc-c"),
     path = "bsp/psc_abc.rs"
```
mkeeter, line 391: Nit: rename to `psc_bc.rs` now that rev A is gone?

mkeeter:
Thanks for doing all of this cleanup!

With the toolchain bump, do you see any changes to `stackmargin` for our in-product images?

boards/README.mkdn
hawkw, line 1: <3

build/util/src/lib.rs
```diff
@@ -85,6 +90,47 @@ pub fn target_board() -> Option<String> {
...
+            break;
+        }
+    }
+    out_dir.push("boards");
+    println!("cargo:rerun-if-changed={}", out_dir.display());
+    if let Ok(dir) = std::fs::read_dir(&out_dir) {
+        for dirent in dir {
+            let Ok(dirent) = dirent else {
+                continue;
+            };
```
hawkw, line 109: nit: should we maybe print a warning on weird/bad dir entries?

build/util/src/lib.rs
```diff
@@ -85,6 +90,47 @@ pub fn target_board() -> Option<String> {
...
+            }
+        }
+    } else {
+        eprintln!(
+            "warning: boards directory not found, \
+                  target_board can't validate"
+        );
+    }
+    boards.sort();
+    boards.dedup();
```
aapoalas, line 129: **nitpick:** Duplicates would probably be a mistake of some sort, maybe worth a warning print as well?

drv/stm32xx-sys-api/src/h7.rs
```diff
@@ -158,12 +151,9 @@ pub enum Peripheral {
...
 
-    #[cfg(any(feature = "h743", feature = "h747", feature = "h757"))]
+    #[cfg(feature = "h743")]
     Dfsdm1 = periph(Group::Apb2, 28), // 43/47 differ from B3
 
     Sai3 = periph(Group::Apb2, 24), // 43/47 only
```
aapoalas, line 159: **question:** There's a bunch of "B3 only" commented lines below: Are those now entirely unused?

aapoalas (commented):
Few comments passing by

cbiffle:
> With the toolchain bump, do you see any changes to `stackmargin` for our in-product images?

I have not yet been able to test this. I'd be willing to remove the toolchain bump from this PR, but the changes may regress and I'd like to opt out of refixing them.

drv/stm32xx-sys-api/src/h7.rs
cbiffle: The lines aren't consistently conditional because I felt the danger of having someone accidentally name a peripheral that isn't implemented on their chip was small. I would like to at least keep the comments; we could comment out the lines that are only on chips we don't support, if you'd prefer.

README.mkdn
cbiffle: It appears that the vendor consistently puts nucleo at the front of the name, so, that's an argument for consistently doing that too.

cbiffle:
Okay, I have resuscitated this PR after getting distracted for several months.

I've fixed some of the things people suggested (thanks folks!) and left a few for future work -- basically in cases where this seems like a big improvement over what we had before, and the further improvement would take thought.

I'd appreciate a re-review; I've kept changes from code review at the top of the branch.

build/util/src/lib.rs
aapoalas: Wait, is it even possible to get duplicates since we take only toml files' stems and don't do eg. lowercasing? Shouldn't all the file stems be necessarily unique?

drv/stm32xx-i2c/src/lib.rs
```diff
@@ -400,21 +367,16 @@ impl I2cController<'_> {
             //   t_timeout = (TIMEOUTA + 1) x 2048 x t_i2cclk
             //
             // We want our t_timeout to be at least 25 ms: on h743 with a 10 ns
-            // t_i2cclk this yields 1219.7 (1220); on h7b3, this is 3416.9
-            // (3417); on g031, this is 195.88 (196).
+            // t_i2cclk this yields 1219.7 (1220); on g031, this is 195.88
+            // (196). Note that these numbers make assumptions about the
+            // system's clocking and clock tree configuration; TODO.
```
aapoalas, line 372: **question:** What is the TODO here about exactly? Is it about finishing this comment or something code related?

aapoalas (commented):
Went in, looking for a new outlook to life and programming: Found only questions.

Joking aside, I have no right to say this but the changes look good to me.

mkeeter:
The code changes look good to me, but I'd still like to see before / after `stackmargin` values for Gimlet / Sidecar / PSC, to make sure the compiler change hasn't made any Exciting New Choices About Inlining 😅

drv/stm32xx-i2c/src/lib.rs
cbiffle: The TODO is intended to be about the earlier part of the sentence: when visiting this file I noticed that it contains hardcoded assumptions about the system's clocking and clock tree configuration, which are under the control of the application main file. This should eventually be fixed.

cbiffle:
> The code changes look good to me, but I'd still like to see before / after `stackmargin` values for Gimlet / Sidecar / PSC, to make sure the compiler change hasn't made any Exciting New Choices About Inlining 😅

Agreed, will work on that today.

I may also bump the compiler to a more recent version, since it's been months since I picked this one. If we're testing stackmargin _anyway..._

drv/stm32xx-i2c/src/lib.rs
aapoalas: Thank you, that explains it really well.

boards/README.mkdn
```diff
@@ -0,0 +1,8 @@
+# Board definitions
```
labbott, line 1: Do we have a full list of "adding a new board" steps written down somewhere?

cbiffle: I don't believe we do. Eventually I'd like "add a file here" to be the whole list, but as you may have noticed, xtask flash and friends still have hardcoded match statements after this change. (I would love to fix that.)

cbiffle:
Alright! I have pushed commits addressing _most_ of the code review feedback. I have also bumped the toolchain rev to 2024-09-17, arbitrarily chosen as "the nightly that seemed to work and was closest to the date that I made the change." This corresponds to a post-1.82 version.

I've done stackmargin testing on all production boards, though only one rev of each, due to hardware availability (and, frankly, time). Some stacks are up, many are down -- in general the code appears to have gotten a little tighter with the updated toolchain. The only problems I found were

1. Gimlet's `hf` task was suspiciously low on stack, but I found 2 kiB of unnecessary data on its stack and fixed that in #1874, so it's good now.
2. PSC's `eeprom` task reported a stack margin of zero, but that turned out to be a kernel bug (#1876) that I will fix later. The task itself appears to be mostly okay, but should probably be removed from production images, as it appears to be a debug/test fixture.

So given that, how are people feeling about the change?

cbiffle:
Ugh Clippy has developed some new opinions about things.

mkeeter (approved):

mkeeter (approved):

README.mkdn
flihp, line 157: Nit: `app/demo-stm32g0-nucleo/app-g0b1.toml` was renamed to `app/demo-stm32g0-nucleo/app-g0b1.toml.noworky` back in c12d361ebf

flihp (approved):

PR #1780, opened 2024-05-10
`transceivers`: convert to `counted_ringbuf`s

### LINKED ISSUE #1778: Have the `transceivers` task utilize the ringbuf counters

Aaron-Hartwig:
I'm going to use a `counted_ringbuf` in the `transceivers` task because there is a lot of activity that happens, specifically in the UDP handler. I also bumped up the size from 16 -> 32 of that one because there's nearly constant activity and each UDP message involves at least three entries.

Fixes #1778


drv/transceivers-server/src/udp.rs
```diff
@@ -10,25 +10,28 @@
...
     ModuleId,
 };
-use userlib::UnwrapLite;
 
 ////////////////////////////////////////////////////////////////////////////////
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, PartialEq, Count)]
 enum Trace {
     None,
```
hawkw, line 36: in general, it's good to add `#[count(skip)]` to the `None`/`Empty` entries --- since we will never actually record them, we don't need to generate a counter for them.
```suggestion
    #[count(skip)]
    None,
```

drv/transceivers-server/src/main.rs
```diff
@@ -41,7 +44,7 @@ task_slot!(SENSOR, sensor);
 include!(concat!(env!("OUT_DIR"), "/i2c_config.rs"));
 
 #[allow(dead_code)]
-#[derive(Copy, Clone, PartialEq, Eq)]
+#[derive(Copy, Clone, PartialEq, Eq, Count)]
 enum Trace {
     None,
```
hawkw, line 49: similarly, we can avoid counting this 
```suggestion
    #[count(skip)]
    None,
```

drv/transceivers-server/src/main.rs
```diff
@@ -41,7 +44,7 @@ task_slot!(SENSOR, sensor);
...
     FrontIOBoardReady(bool),
```
hawkw, line 50: since `true` means that the front IO board is ready, and `false` means that it's _not_ ready, maybe we should generate separate counters for those states?
```suggestion
    FrontIOBoardReady(#[count(children)] bool),
```

hawkw (commented):
I left a couple suggestions. At some point we may also want to add `#[count(children)]` to some of the errors, so that we can generate a separate counter for each error variant, but it's up to you whether we ought to do that now, and which errors are worth counting individually.

drv/transceivers-server/src/main.rs
```diff
@@ -5,6 +5,10 @@
 #![no_std]
 #![no_main]
 
+use counters::Count;
```
aapoalas, line 8: **nitpick/question:** Here and in the other file, the imports seem to have been in an alphabetical order but now separate into two groups. Is this intended?

drv/transceivers-server/src/main.rs
Aaron-Hartwig: Agreed!

Aaron-Hartwig:
>At some point we may also want to add #[count(children)] to some of the errors, so that we can generate a separate counter for each error variant, but it's up to you whether we ought to do that now, and which errors are worth counting individually.

@hawkw I did add it in a few places, but some of the other places I'd like to add it are blocked by `#[derive(Count)]` being limited to `enum`s right now. Most of the places I'd like to add it would be keyed off an event happening at a particular port (currently a `usize` in the `Trace`, but could easily be made a `LogicalPort` which is a struct that wraps a `u8`). Thanks for the feedback as always!

drv/transceivers-server/src/main.rs
Aaron-Hartwig: I was separating out some of our common library code from the more task-specific stuff in these files.

hawkw:
> I did add it in a few places, but some of the other places I'd like to add it are blocked by `#[derive(Count)]` being limited to `enum`s right now. Most of the places I'd like to add it would be keyed off an event happening at a particular port (currently a `usize` in the `Trace`, but could easily be made a `LogicalPort` which is a struct that wraps a `u8`). Thanks for the feedback as always!

Yeah, I think that if you want to count events per `LogicalPort`, you would need to actually implement `Count` for that type manually. I would probably do it using an array of counters that's indexed by port, like this: 

```rust

impl Count for LogicalPort {
    type Counters = [AtomicU32; NUM_PORTS],;

    #[allow(clippy::declare_interior_mutable_const)]
    const NEW_COUNTERS: Self::Counters = {
        const COUNTER_INIT: AtomicU32 = AtomicU32::new(0);
        [COUNTER_INIT; NUM_PORTS]
    };

    fn count(&self, counters: &Self::Counters) {
        let Some(ctr) = counters.get(self.0 as usize) else {
            // `LogicalPort`s should never index the counters array out of
            // bounds, but if that does ever happen, let's just bail instead of
            // panicking...
            return;
        }
        armv6m_atomic_hack::AtomicU32Ext::fetch_add(ctr, 1, Ordering::Relaxed);
    }
}
```

Although, OTTOMH, I'm not sure how well the `humility` code for reading counters would handle an array, so I might have to go back and fix that if you did do something like this...

Aaron-Hartwig:
@hawkw should we do the `Count` impl for `LogicalPort` in this PR? I don't want to force Humility work upon you if you've something more pressing to work on. I can always write an issue for this and handle it in The Future.

hawkw:
> @hawkw should we do the `Count` impl for `LogicalPort` in this PR? I don't want to force Humility work upon you if you've something more pressing to work on. I can always write an issue for this and handle it in The Future.

My preference would be to go ahead and merge this as-is, and add something like that later --- I can think of a few other places where it would be nice to be able to use arrays for generating counters from a numeric index (e.g. `SensorId`s in `task-sensors`) if I get around to making it work nicely in humility!.

hawkw (approved):

PR #1782, opened 2024-05-15
Refresh FPGAs after toolchain bump

Aaron-Hartwig:
We updated the FPGA toolchain in https://github.com/oxidecomputer/quartz/pull/150. This refreshes the images we have as part of our applications with fresh ones from CI.

cc: @nathanaelhuffman @arjenroodselaar


labbott (approved):
LGTM

PR #1787, opened 2024-05-20
transceivers: fix I2C read bug

### LINKED ISSUE #1786: `transceivers`: `get_i2c_status_and_read_buffer` reading FIFO before I2C transaction finishes
Networking was misbehaving on the dogfood rack and the `qsfp` interfaces were not operating. While reading from and writing to modules appeared to be working, there was a bit of oddness when reading the vendor info:
```
root@oxz_switch0:~# xcvradm -i sidecar0 -t present vendor-info
Port Identifier               Vendor           Part             Rev  Serial           Mfg date
  18 Qsfp28 (0x11)            0905000000000000 0000000000000000 00   0000000000000000 00000000
  26 Qsfp28 (0x11)            FINISAR CORP     FTLC9558REPM     A1   X6HAW9Q          211029
  29 Qsfp28 (0x11)            FOIT             AFBR-89BDDZ           AF2132N000W      210806
  31 Qsfp28 (0x11)            Intel Corp       SPTSBP4LLCDF     01   CRLR2139028FP    21111700
```
Port 18's info incorrect (mostly zeroes).

On `niles` there was similar strangeness:
```
aaron@niles ~ $ ./xcvradm -i axf7 -t present vendor-info
Port Identifier               Vendor           Part             Rev  Serial           Mfg date
   0 Qsfp28 (0x11)            2203211111111111 1111111111111111 11   1111111111111111 11111111
   2 QsfpPlusCmis (0x1e)      Intel Corp       SPTSMP3CLCDA     03   CRFR2141020JP    21101500
   3 QsfpPlusCmis (0x1e)      Intel Corp       SPTSMP3CLCDA     03   CRFR213905JEP    21101800
   4 QsfpPlusCmis (0x1e)      FINISAR CORP.    FTCC1112E2PCL    A    X65BPQR          210901
   5 Qsfp28 (0x11)            FS               QSFP28-SR4-100G  1A   F2220590150      220615
   8 QsfpPlusCmis (0x1e)      FINISAR CORP.    FTCC1112E2PCL    A    X6QA1JC          220305
  16 Qsfp28 (0x11)            FS               QSFP28-SR4-100G  04   G2130484856      20220321
  24 Qsfp28 (0x11)            Intel Corp       AMQ28-SR4        01   IN100MC0040      221206
```

When modules were swapped, the problem seemed to stick with port 0:
```
aaron@niles ~ $ ./xcvradm -i axf7 -t present vendor-info
Port Identifier               Vendor           Part             Rev  Serial           Mfg date
   0 Qsfp28 (0x11)            0615
   2 QsfpPlusCmis (0x1e)      Intel Corp       SPTSMP3CLCDA     03   CRFR2141020JP    21101500
   3 QsfpPlusCmis (0x1e)      FINISAR CORP.    FTCC1112E2PCL    A    X65BPQR          210901
   4 QsfpPlusCmis (0x1e)      Intel Corp       SPTSMP3CLCDA     03   CRFR213905JEP    21101800
   5 Qsfp28 (0x11)            FS               QSFP28-SR4-100G  04   G2130484857      20220321
   8 Qsfp28 (0x11)            Intel Corp       AMQ28-SR4        01   IN100MC0040      221206
  16 Qsfp28 (0x11)            FS               QSFP28-SR4-100G  04   G2130484856      20220321
  24 QsfpPlusCmis (0x1e)      FINISAR CORP.    FTCC1112E2PCL    A    X6QA1JC          220305
```
When the module in port 0 was removed, the problem moved to the _next_ Qsfp28 module. This was very interesting. The reason it moved to the next Qsfp28 module is because the transceiver controller in dendrite splits up QSFP28 vs CMIS module communication and then talks to QSFP28 first (if applicable). So it moving to that next module was really it moving to the **first** module it was meant to talk to. This implied something between SP/FPGA I2C interaction.

This bug was introduced by https://github.com/oxidecomputer/hubris/pull/1768. That PR was a large change to how the SP/FPGA interaction happens for I2C transactions to the modules. The FPGA went from a memory-mapped interface to the data to a FIFO. The updates to the SP code failed to properly account for the fact that this interface was now a FIFO and continued to operate in the old model where it would grab the status byte, grab all the data, and then read the status byte to see if that data was valid yet or not (see https://github.com/oxidecomputer/hubris/blob/master/drv/sidecar-front-io/src/transceivers.rs#L1048-L1053). The "check the status byte" portion of the loop happened outside the part where we read the buffer, and since the buffer was just memory-mapped registers it could be repeatedly without consequence. Since the data was now in a FIFO, I was inadvertently draining the FIFO before the transaction was done.

The fix here will be to handle the "is I2C done yet" part of I2C reads within the function where we actually do the reading. Currently that loop is handled in the application code where we need the read data. It will be good to encapsulate this behavior in the transceivers layer so the user does not need to understand these implementation details.

Aaron-Hartwig:
#1768 did not properly account for the FIFO behavior of the FPGA's data buffers. The "check the status byte" portion of the loop happened outside the part where we read the buffer, and since the buffer was just memory-mapped registers it could be repeatedly without consequence. Since the data was now in a FIFO, I was inadvertently draining the FIFO before the transaction was done. This PR consolidates the "is I2C done yet" logic into the `get_i2c_status_and_read_buffer` so calling code can just deal with the status register and the data buffer.

Fixes #1786

### DIFF
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -270,9 +270,9 @@ impl ServerImpl {
             // increments of 1/256 degrees Celsius"
             //
             // - SFF-8636 rev 2.10a, Section 6.2.4
-            return Ok(Celsius(out.temperature.get() as f32 / 256.0));
+            Ok(Celsius(out.temperature.get() as f32 / 256.0))
         } else {
-            return Err(FpgaError::ImplError(status.as_bytes()[0]));
+            Err(FpgaError::ImplError(status.as_bytes()[0]))
         }
     }
 
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -16,7 +16,9 @@ use drv_fpga_api::FpgaError;
 use drv_i2c_devices::pca9956b::Error;
 use drv_sidecar_front_io::{
     leds::{FullErrorSummary, Leds},
-    transceivers::{LogicalPort, LogicalPortMask, Transceivers},
+    transceivers::{
+        FpgaI2CFailure, LogicalPort, LogicalPortMask, Transceivers,
+    },
     Reg,
 };
 use drv_sidecar_seq_api::{SeqError, Sequencer};
@@ -253,28 +255,24 @@ impl ServerImpl {
 
         #[derive(Copy, Clone, FromBytes, AsBytes)]
         #[repr(C)]
-        struct StatusAndTemperature {
-            status: u8,
+        struct Temperature {
             temperature: zerocopy::I16<zerocopy::BigEndian>,
         }
 
-        loop {
-            let mut out = StatusAndTemperature::new_zeroed();
-            self.transceivers
-                .get_i2c_status_and_read_buffer(port, out.as_bytes_mut())?;
-            if out.status & Reg::QSFP::PORT0_STATUS::BUSY == 0 {
-                if out.status & Reg::QSFP::PORT0_STATUS::ERROR != 0 {
-                    return Err(FpgaError::ImplError(out.status));
-                } else {
-                    // "Internally measured free side device temperatures are
-                    // represented as a 16-bit signed twos complement value in
-                    // increments of 1/256 degrees Celsius"
-                    //
-                    // - SFF-8636 rev 2.10a, Section 6.2.4
-                    return Ok(Celsius(out.temperature.get() as f32 / 256.0));
-                }
-            }
-            userlib::hl::sleep_for(1);
+        let mut out = Temperature::new_zeroed();
+        let status = self
+            .transceivers
+            .get_i2c_status_and_read_buffer(port, out.as_bytes_mut())?;
+
+        if status.error == FpgaI2CFailure::NoError {
+            // "Internally measured free side device temperatures are
+            // represented as a 16-bit signed twos complement value in
+            // increments of 1/256 degrees Celsius"
+            //
+            // - SFF-8636 rev 2.10a, Section 6.2.4
+            return Ok(Celsius(out.temperature.get() as f32 / 256.0));
+        } else {
+            return Err(FpgaError::ImplError(status.as_bytes()[0]));
         }
     }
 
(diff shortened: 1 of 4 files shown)


drv/transceivers-server/src/main.rs
```diff
@@ -253,28 +255,24 @@ impl ServerImpl {
...
+
+        if status.error == FpgaI2CFailure::NoError {
+            // "Internally measured free side device temperatures are
+            // represented as a 16-bit signed twos complement value in
+            // increments of 1/256 degrees Celsius"
+            //
+            // - SFF-8636 rev 2.10a, Section 6.2.4
+            return Ok(Celsius(out.temperature.get() as f32 / 256.0));
+        } else {
+            return Err(FpgaError::ImplError(status.as_bytes()[0]));
```
hawkw, line 275: since there's no longer a loop here , the `return` keyword is no longer necessary, and clippy is complaining about it: https://github.com/oxidecomputer/hubris/actions/runs/9162854238/job/25190734510?pr=1787

```suggestion
            Ok(Celsius(out.temperature.get() as f32 / 256.0))
        } else {
            Err(FpgaError::ImplError(status.as_bytes()[0]))
```

drv/transceivers-server/src/main.rs
```diff
@@ -287,28 +285,23 @@ impl ServerImpl {
...
-            match out[1] {
+        if status.error == FpgaI2CFailure::NoError {
+            match out[0] {
                 0x1E => Ok(ManagementInterface::Cmis),
                 0x0D | 0x11 => Ok(ManagementInterface::Sff8636),
                 i => Ok(ManagementInterface::Unknown(i)),
             }
         } else {
             // TODO: how should we handle this?
             // Right now, we'll retry on the next pass through the loop.
```
hawkw, line 303: this comment is no longer accurate, right?

drv/transceivers-server/src/main.rs
Aaron-Hartwig: I have a nasty habit of forgetting to run clippy locally. Thanks for pointing this out as well!

drv/transceivers-server/src/main.rs
Aaron-Hartwig: It actually is. The loop being referred to here is actually the `update_thermal_loop` function which will call `get_transceiver_interface` repeatedly (https://github.com/oxidecomputer/hubris/blob/master/drv/transceivers-server/src/main.rs#L361)

labbott (approved):
LGTM

drv/transceivers-server/src/main.rs
hawkw: ah, i see, yeah, that makes sense.

drv/transceivers-server/src/main.rs
```diff
@@ -253,28 +255,24 @@ impl ServerImpl {
...
+
+        if status.error == FpgaI2CFailure::NoError {
+            // "Internally measured free side device temperatures are
+            // represented as a 16-bit signed twos complement value in
+            // increments of 1/256 degrees Celsius"
+            //
+            // - SFF-8636 rev 2.10a, Section 6.2.4
+            Ok(Celsius(out.temperature.get() as f32 / 256.0))
+        } else {
+            Err(FpgaError::ImplError(status.as_bytes()[0]))
```
mkeeter, line 275: What is this doing?  I think it's going to return the value of `rdata_fifo_empty`, because it's casting an `PortI2CStatus` to a `&[u8]` and taking the first byte.

Aaron-Hartwig: Good catch! I actually don't care about the whole status here, I really just want to log the `error` observed. I'm going to make this line:
```suggestion
            Err(FpgaError::ImplError(status.error.as_bytes()[0]))
```
And get rid of the `AsBytes` deriviation on the `PortI2CStatus` type.

mkeeter: One more note: it looks like you can do `status.error as u8` instead of `status.error.as_bytes()[0]`, since it's already a single-byte `enum` type.

mkeeter (approved):

PR #1799, opened 2024-05-22
psc-seq: don't repeat StickyFault every time.

cbiffle:
We're currently in a situation where the dogfood rack has the power whip feeding PSUs 3-5 disconnected. This means the PSUs are sitting there in faulted state refusing to enable. Since we can't distinguish this from actual PSU failures, currently we're frantically chatting about this situation in the logs. This is annoying, and overwrites other data.

This change alters the behavior to note a sticky fault condition _once_ and then quietly keep trying to resolve it without the spam.

### DIFF
--- a/drv/psc-seq-server/src/main.rs
+++ b/drv/psc-seq-server/src/main.rs
@@ -259,7 +259,16 @@ enum PresentState {
     NewlyInserted { settle_deadline: u64 },
     /// The PSU has unexpectedly deasserted the OK signal, or failed to assert
     /// it within a reasonable amount of time after being turned on.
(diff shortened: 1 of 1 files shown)


hawkw:
This seems non-compiley --- it looks like you've forgotten the `already_sticky` field here: https://github.com/oxidecomputer/hubris/actions/runs/9198535640/job/25301455881?pr=1799#step:6:175

hawkw (approved):
This seems good to me, once it compiles! I think we probably ought to actually add a mechanism for reading out the *present* state of the rectifiers, as well.

PR #1802, opened 2024-05-24
psc: allow PSUs a grace period before requiring OK after enable.

cbiffle:
In fault injection testing I've seen _one time_ when the PSU failed to assert OK quickly after being enabled. This caused us to fault it and try again, at which point it worked. But, that has the effect of inserting an additional 5s delay on the fault recovery cycle, which in theory could repeat forever.

This change adds a grace period after taking a faulted PSU out of fault status, before we start expecting it to assert OK. This mirrors the grace period we provide for newly inserted PSUs.


drv/psc-seq-server/src/main.rs
```diff
@@ -250,6 +259,18 @@ enum PresentState {
         // with a "sticky fault" situation.
         turn_on_deadline: u64,
     },
+
+    /// We are allowing the ON signal to float active, as in the `On` state, but
+    /// we're not convinced the PSU is okay. We enter this state when bringing a
+    /// PSU out of an observed fault state, and it causes us to ignore its OK
+    /// output for a brief period (the deadline parameter).
```
hawkw, line 266: very minor nit: maybe worth referencing the duration constant here?
```suggestion
    /// output for a brief period ([`PROBATION_MS`]), which is represented
    /// by the deadline parameter.
```
(or something)

PR #1806, opened 2024-06-03
Adds a Medusa image

### LINKED ISSUE #1571: Create an application TOML for the QSFP tester

Aaron-Hartwig:
The primary purpose of this PR is to sketch out an initial stab at an image for Medusa, our QSFP front IO test fixture (details on that in RFD 405). The Medusa-specific logic lives in `medusa-seq-server`. The current goal is just to support hardware bringup, and thus that sequencer just attempts to power on the front IO board and then program its FPGAs. It also checks the status of the various power rails on Medusa and exposes an interface to control those rails (on Sidecar this functionality is handled by the FPGA).

An adjacent piece of work here is that Medusa does not have any fans to control and that broke the previous mold of "everything with `thermal` has fans". `thermal` logs `sensor` data _and_ wants to control fan speeds off that data. So the Medusa thermal BSP looks a bit odd compared to the others. Additionally, the `transceivers` task had a `thermal-control` feature flag added to gate the interactions with `thermal` (Sidecar will have this feature set, Medusa will not).

Lastly, there's some ugliness that I did for the sake of expediency given my personal situation. Until I can actually land #1805, Medusa simply copies some code from Sidecar. Once this PR lands, I will extend #1805 to cleanup Medusa as well.

Closes #1571


Aaron-Hartwig:
I'm opening this branch up as a draft PR just so we know where this work is at with my leave looming. There is a bunch of copy/paste here that can get addressed properly once https://github.com/oxidecomputer/hubris/pull/1805 lands.

task/monorail-server/src/bsp/medusa_a.rs
```diff
@@ -0,0 +1,528 @@
...
+
+            // Notify the sequencer about the state of the oscillator. If the
+            // oscillator is good any future resets of the PHY do not require a
+            // full power cycle of the front IO board.
+            self.seq
+                .set_front_io_phy_osc_state(osc_good)
+                .map_err(|e| VscError::ProxyError(e.into()))?;
+
+            if !osc_good {
+                ringbuf_entry!(Trace::FrontIoPhyOscillatorBad)
```
aapoalas, line 233: **question:** Should this only log once to avoid flooding the ringbuf in case resetting the sequencer isn't helping with the oscillator state? (And actually, should there be some limit where the software just gives up and decides that the oscillator is entirely unworkable?)

aapoalas (commented):
Comments from the peanut gallery.

drv/medusa-seq-server/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+        // The EN->PG time for this part was experimentally determined to be
+        // 35ms, so we roughly double that.
+        userlib::hl::sleep_for(75);
+
+        // Power is not good. Disable the rail and log that this happened.
+        if !self.power_control.v12_qsfp_out.check_power_good() {
+            return Err(MedusaError::FrontIOBoardPowerFault);
+        }
+
+        userlib::hl::sleep_for(25);
```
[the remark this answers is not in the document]
Aaron-Hartwig: I would say it is not! Good catch.

drv/medusa-seq-server/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+        Ok(false) => {
+            ringbuf_entry!(Trace::FrontIOBoardNotPresent);
+            server.power_control.v12_qsfp_out.set_enable(false);
+        }
+        Err(MedusaError::FrontIOBoardPowerFault) => {
+            ringbuf_entry!(Trace::FrontIOBoardPowerFault)
+        }
+        // Something went wrong getting the HSC status, eject.
+        Err(_) => panic!("unknown front IO board preinit failure"),
```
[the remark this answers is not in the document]
Aaron-Hartwig: This code is grafted over from how the `sidecar-seq-server` handles this currently and will be reworked heavily once I can land a large refactor. That being said, we actually should never reach this branch because `front_io_board_preinit` only returns a `MedusaError::FrontIOBoardPowerFault` error variant. I'll change this to an `unreachable` and update the comment accordingly.

drv/medusa-seq-server/src/power_control.rs
```diff
@@ -0,0 +1,159 @@
...
+        Self {
+            enable,
+            power_good,
+            name,
+        }
+    }
+
+    /// Sets the enable pin for the power rail to HIGH if `enabled` is true or
+    /// LOW if it is false. If `enabled` is true, this function will also
+    /// monitor the power good signal for the supply.
```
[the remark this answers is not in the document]
Aaron-Hartwig: It does not! An earlier iteration of the function did, but I pulled out the monitoring to be its own thing because I did not want to wait around for monitoring on each individual power supply enable as that would be problematic when there's a need for multiple supplies to come up within a certain time of one another. I've removed the stale portion of the comment.

drv/medusa-seq-server/src/power_control.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub v1p0_mgmt: PowerRail,
+    pub v1p2_mgmt: PowerRail,
+    pub v2p5_mgmt: PowerRail,
+    pub v1p0_phy: PowerRail,
+    pub v2p5_phy: PowerRail,
+}
+
+impl PowerControl {
+    pub fn new() -> Self {
+        // Wait a bit for it to ramp and then check that we can that it is happy
```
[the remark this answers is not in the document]
Aaron-Hartwig: Another stale comment!

task/monorail-server/src/bsp/medusa_a.rs
Aaron-Hartwig: This is not something I'm concerned about right now given our experience with the problematic part thus far. Sometime it will come up quickly, other times after several more power cycles. We've yet to encounter a case where it did not come up, and should that happen the impact is large enough that we'd need to take physical action with the switch anyway so I believe it is harmless to continue to cycle the board.

app/medusa/src/main.rs
```diff
@@ -0,0 +1,136 @@
...
+    let id_mask = (1 << 6) | (1 << 7) | (1 << 13);
+    let model = p.GPIOC.idr.read().bits() & id_mask;
+
+    cfg_if::cfg_if! {
+        if #[cfg(target_board = "medusa-a")] {
+            let expected_model = 0b000;
+        } else {
+            compile_error!("not a recognized medusa board");
+        }
+    }
```
hawkw, line 88: at first, i wasn't super convinced that the use of `cfg_if` was really pulling its weight here, given that this could just be 
```rust
#[cfg(target_board = "medusa-a")]
let expected_model = 0b000;
#[cfg(not(target_board = "medusa-a"))]
compile_error!(...);
```

but then it occurred to me that there's the possibility of more revisions in the future, and it's probably easier to handle them with `cfg_if`, so...carry on!

app/medusa/src/main.rs
hawkw, line 1: I'm assuming you know substantially more about the details of the Medusa clock tree than I do, so I didn't review this super closely, I'm assuming you've got it right :)

drv/medusa-seq-api/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#[derive(
+    Copy, Clone, Debug, FromPrimitive, Eq, PartialEq, IdolError, counters::Count,
+)]
+pub enum MedusaError {
+    FpgaError = 1,
+    NoFrontIOBoard,
+    FrontIOBoardPowerFault,
+    PowerFault,
```
hawkw, line 34: maybe a bit nitpicky but it might be nice to have a comment explaining the difference between these?

drv/medusa-seq-server/src/front_io.rs
```diff
@@ -0,0 +1,111 @@
...
+                });
+
+                controller.write_checksum()?;
+                (checksum, checksum_valid) = controller.checksum_valid()?;
+                ringbuf_entry!(Trace::FrontIOControllerChecksum {
+                    fpga_id: i,
+                    checksum,
+                    expected: FrontIOController::short_checksum(),
+                });
+            }
```
hawkw, line 104: style nit, take it or leave it: i  _might_ consider using `continue` here so that the rest of the code doesn't drift to the right; others may disagree with me on this though:
```suggestion
            if ident_valid && checksum_valid {
                ringbuf_entry!(Trace::SkipLoadingFrontIOControllerBitstream {
                    fpga_id: i
                });
                continue;
            }
  
            ringbuf_entry!(Trace::LoadingFrontIOControllerBitstream {
                fpga_id: i
            });

            if let Err(e) = controller.load_bitstream(self.auxflash_task) {
                ringbuf_entry!(Trace::FpgaBitstreamError(u32::from(e)));
                return Err(e);
            }

            (ident, ident_valid) = controller.ident_valid()?;
            ringbuf_entry!(Trace::FrontIOControllerIdent {
                fpga_id: i,
                ident
            });

            controller.write_checksum()?;
            (checksum, checksum_valid) = controller.checksum_valid()?;
            ringbuf_entry!(Trace::FrontIOControllerChecksum {
                fpga_id: i,
                checksum,
                expected: FrontIOController::short_checksum(),
            });
```

drv/medusa-seq-server/src/front_io.rs
```diff
@@ -0,0 +1,111 @@
...
+                    // Attempt to correct the invalid IDENT by reloading the
+                    // bitstream.
+                    controller.fpga_reset()?;
+                }
+            }
+
+            if ident_valid && checksum_valid {
+                ringbuf_entry!(Trace::SkipLoadingFrontIOControllerBitstream {
+                    fpga_id: i
+                });
```
hawkw, line 80: it might be nice to have a comment here saying that we don't need to do anything because we are already running the expected bitstream. or, we could rename the ringbuf entry to something like `FrontIOControllerBitstreamAlreadyLoaded` so that it's clear to the reader why we stop here.

drv/medusa-seq-server/src/front_io.rs
```diff
@@ -0,0 +1,111 @@
...
+                    ident
+                });
+
+                controller.write_checksum()?;
+                (checksum, checksum_valid) = controller.checksum_valid()?;
+                ringbuf_entry!(Trace::FrontIOControllerChecksum {
+                    fpga_id: i,
+                    checksum,
+                    expected: FrontIOController::short_checksum(),
+                });
```
hawkw, line 103: if we've loaded the expected bitstream and the ident/checksum are still invalid, that seems like a pretty bad error state, right? is there anything we ought to do to make it obvious that we're in a bad state?

this is fine to revisit later, though!

drv/medusa-seq-server/src/main.rs
```diff
@@ -0,0 +1,357 @@
...
+const TIMER_INTERVAL: u64 = 1000;
+
+struct ServerImpl {
+    power_control: PowerControl,
+    front_io_board: Option<FrontIOBoard>,
+}
+
+impl ServerImpl {
+    fn front_io_board_preinit(&self) -> Result<bool, MedusaError> {
+        // Enale the V12_QSFP_OUT rail
```
hawkw, line 74: typo:
```suggestion
        // Enable the V12_QSFP_OUT rail
```

drv/medusa-seq-server/src/main.rs
```diff
@@ -0,0 +1,357 @@
...
+            ringbuf_entry!(Trace::FrontIOBoardPhyPowerEnable(true));
+
+            while !front_io_board.phy().powered_up_and_ready()? {
+                userlib::hl::sleep_for(20);
+            }
+
+            Ok(())
+        } else {
+            Err(MedusaError::NoFrontIOBoard)
+        }
```
hawkw, line 166: style nit, take it or leave it: personally, i might consider writing this using `ok_or` rather than `if let`, so that the returned error is closer to the condition in which we return the error, and the code reads like a straight line:
```suggestion
        let front_io_board = self
            .front_io_board
            .as_mut()
            .ok_or(MedusaError::NoFrontIOBoard)?;

        // At this point the front IO board has either not yet been
        // initialized or may have been power cycled and should be
        // initialized.
        if !front_io_board.initialized() {
            front_io_board.init()?;
        }

        // The PHY is still powered down. Request the sequencer to power up
        // and wait for it to be ready.
        front_io_board.phy().set_phy_power_enabled(true)?;
        ringbuf_entry!(Trace::FrontIOBoardPhyPowerEnable(true));

        while !front_io_board.phy().powered_up_and_ready()? {
            userlib::hl::sleep_for(20);
        }

        Ok(())
```

drv/medusa-seq-server/src/main.rs
```diff
@@ -0,0 +1,357 @@
...
+                }
+            }
+        }
+
+        // Run preinit to check HSC status.
+        self.front_io_board_preinit()?;
+
+        if let Some(front_io_board) = self.front_io_board.as_mut() {
+            // At this point the front IO board has either not yet been
+            // initalized or may have been power cycled and should be
```
hawkw, line 148: typo:
```suggestion
            // initialized or may have been power cycled and should be
```

drv/medusa-seq-server/src/main.rs
```diff
@@ -0,0 +1,357 @@
...
+        Err(MedusaError::FrontIOBoardPowerFault) => {
+            ringbuf_entry!(Trace::FrontIOBoardPowerFault)
+        }
+        // `front_io_board_preinit` currently only returns a
+        // MedusaError::FrontIOBoardPowerFault
+        Err(_) => unreachable!(),
+    }
+
+    // The MGMT and PHY rails are enabled automatically by pullups, so we will
+    // check their power good signals and take action as appropriate.
```
hawkw, line 334: it looks like we don't actually take action if the MGMT or PHY rails are bad?

drv/medusa-seq-server/src/power_control.rs
```diff
@@ -0,0 +1,156 @@
...
+            PowerRail::new(Port::J.pin(6), Port::J.pin(5), RailName::V1P2Mgmt);
+        let v2p5_mgmt =
+            PowerRail::new(Port::J.pin(8), Port::J.pin(7), RailName::V2P5Mgmt);
+
+        // The VSC8562 rails are generated from the same LDO which shares an
+        // enable pin
+        let v1p0_phy =
+            PowerRail::new(Port::J.pin(10), Port::J.pin(11), RailName::V1P0Phy);
+        let v2p5_phy =
+            PowerRail::new(Port::J.pin(10), Port::J.pin(12), RailName::V2P5Phy);
```
hawkw, line 103: would it make sense for these pin numbers to be defined in the board config TOML, in anticipation of potential revisions? i don't think hardcoding them is a big problem, or if they're likely to ever change, just curious.

drv/transceivers-server/Cargo.toml
```diff
@@ -33,6 +33,7 @@ zerocopy = { workspace = true }
 
 [features]
 vlan = ["task-net-api/vlan"]
+thermal-control = []
```
hawkw, line 36: would it make sense to make the dependency on `task-thermal-api` optional?

drv/transceivers-server/src/main.rs
```diff
@@ -127,7 +129,7 @@ struct ServerImpl {
     consecutive_nacks: [u8; NUM_PORTS as usize],
 
     /// Handle to write thermal models and presence to the `thermal` task
-    thermal_api: Thermal,
+    thermal_api: Option<Thermal>,
```
hawkw, line 132: this means that we're checking at runtime whether thermal control is enabled, but whether or not it's enabled is determined at compile time by the feature flag. so, we could instead write
```suggestion
    #[cfg(feature = "thermal-control")]
    thermal_api: Thermal,
```
to avoid the runtime checks

drv/transceivers-server/src/main.rs
```diff
@@ -378,9 +380,11 @@ impl ServerImpl {
...
                 }
             } else if !operational && self.thermal_models[i].is_some() {
-                // This transceiver went away; remove it from the thermal loop
-                if let Err(e) = self.thermal_api.remove_dynamic_input(i) {
-                    ringbuf_entry!(Trace::ThermalError(i, e));
+                if let Some(thermal_api) = &self.thermal_api {
+                    // This transceiver went away; remove it from the thermal loop
+                    if let Err(e) = thermal_api.remove_dynamic_input(i) {
+                        ringbuf_entry!(Trace::ThermalError(i, e));
+                    }
```
hawkw, line 387: following my suggestion from above:
```suggestion
                #[cfg(feature = "thermal-control")]
                {
                    // This transceiver went away; remove it from the thermal loop
                    if let Err(e) = self.thermal_api.remove_dynamic_input(i) {
                        ringbuf_entry!(Trace::ThermalError(i, e));
                    }
```

drv/transceivers-server/src/main.rs
```diff
@@ -408,14 +412,16 @@ impl ServerImpl {
...
-            match self.thermal_api.update_dynamic_input(i, m.model) {
-                Ok(()) | Err(ThermalError::NotInAutoMode) => (),
-                Err(e) => ringbuf_entry!(Trace::ThermalError(i, e)),
+            if let Some(thermal_api) = &self.thermal_api {
+                // *Always* post the thermal model over to the thermal task, so
+                // that the thermal task still has it in case of restart.  This
+                // will return a `NotInAutoMode` error if the thermal loop is in
+                // manual mode; this is harmless and will be ignored (instead of
+                // cluttering up the logs).
+                match thermal_api.update_dynamic_input(i, m.model) {
```
hawkw, line 421: similarly, this doesn't really need to be a runtime check; if we followed my suggestion from above, this could instead be:
```suggestion
            #[cfg(feature = "thermal-control")]
            {
                // *Always* post the thermal model over to the thermal task, so
                // that the thermal task still has it in case of restart.  This
                // will return a `NotInAutoMode` error if the thermal loop is in
                // manual mode; this is harmless and will be ignored (instead of
                // cluttering up the logs).
                match self.thermal_api.update_dynamic_input(i, m.model) {
```

drv/transceivers-server/src/main.rs
```diff
@@ -620,6 +625,11 @@ fn main() -> ! {
         BUFS.claim()
     };
 
+    #[cfg(feature = "thermal-control")]
+    let thermal_api = Some(Thermal::from(THERMAL.get_task_id()));
+    #[cfg(not(feature = "thermal-control"))]
+    let thermal_api = None;
```
hawkw, line 631: this would also change a bit if we follow my above suggestion to remove the `Option`.

task/net/src/bsp/medusa_a.rs
```diff
@@ -0,0 +1,138 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! BSP for the Sidecar rev B
+//!
+//! Right now, this is identical to the rev A BSP, but that may change in the
+//! future, so they're kept separate.
```
hawkw, line 8: looks like this was copied and pasted and should, instead, say:
```suggestion
//! BSP for Medusa rev A
```

task/thermal/Cargo.toml
```diff
@@ -35,8 +35,9 @@ build-i2c = { path = "../../build/i2c" }
 build-util = { path = "../../build/util" }
 
 [features]
-gimlet = ["drv-gimlet-seq-api", "h753"]
-sidecar = ["drv-sidecar-seq-api", "drv-transceivers-api", "h753"]
+gimlet = ["dep:drv-gimlet-seq-api", "h753"]
+sidecar = ["dep:drv-sidecar-seq-api", "dep:drv-transceivers-api", "h753"]
```
hawkw, line 39: nitpicky, but personally i would prefer to avoid changing unrelated code if not necessary.

task/thermal/src/bsp/medusa_a.rs
```diff
@@ -0,0 +1,104 @@
...
+// External temperature inputs, which are provided to the task over IPC
+// In practice, these are our transceivers.
+pub const NUM_DYNAMIC_TEMPERATURE_INPUTS: usize =
+    drv_transceivers_api::NUM_PORTS as usize;
+
+// Number of individual fans - Medusa has none!
+pub const NUM_FANS: usize = 0;
+
+// Run the PID loop on startup
+pub const USE_CONTROLLER: bool = false;
```
hawkw, line 33: for some reason this code is extremely funny to me

not a suggestion, all the zeroes just made me giggle. carry on!

task/thermal/src/bsp/medusa_a.rs
```diff
@@ -0,0 +1,104 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+bitflags::bitflags! {
+    #[derive(Copy, Clone, Debug, Eq, PartialEq)]
+    pub struct PowerBitmask: u32 {
+    }
```
hawkw, line 40: style nit:
```suggestion
    pub struct PowerBitmask: u32 {}
```

hawkw (approved):
Overall, this looks good! You know a lot more about Medusa's design than I do, so I focused on the code rather than on what it's doing. I left some style suggestions, most of which aren't really a big deal. 

I do think it would be nice to make the `task-thermal-api` dependency in `drv-transcievers-server` optional, so that the thermal API doesn't have to be compiled in when thermal control is disabled. This could potentially make Medusa builds slightly faster, which is probably nice to have. Other than that, my suggestions were just minor style nits.

task/monorail-server/src/bsp/medusa_a.rs
hawkw: @aapoalas  it's also worth noting that our ringbufs have a de-duplication feature where, if we record entries that are equal multiple times in a row, it just increments the event's counter rather than recording a new entry. so, if there aren't any other entries recorded in between these, it isn't really a big concern.

app/medusa/src/main.rs
Aaron-Hartwig: Yeah, its a bit clunky currently. But it let makes adding future support a bit less cumbersome!

drv/medusa-seq-server/src/front_io.rs
Aaron-Hartwig: I hear you on this, but I'm going to opt to keep it the way it is right now. This file is just copy/paste from https://github.com/oxidecomputer/hubris/blob/master/drv/sidecar-mainboard-controller/src/front_io.rs to get a quick solution here. In #1805, which I've not attempted to land before I go on leave, this could get addressed properly.

drv/medusa-seq-server/src/front_io.rs
Aaron-Hartwig: As I state in the comment above, I'm going to leave this file as pure copy/paste as possible for now just to keep Sidecar/medusa as identical as possible until I can better unify the Front IO control stuff via #1805.

drv/medusa-seq-server/src/front_io.rs
Aaron-Hartwig: Same comment as above. Going to leave this file as-is for the present.

drv/medusa-seq-server/src/main.rs
Aaron-Hartwig: Stylistically I think I've tended to see the `if let Some` pattern more in Hubris, but I like your suggestion better!

drv/medusa-seq-server/src/main.rs
Aaron-Hartwig: The `power_control::{mgmt,phy}_power_check` functions check the power good pins and disable the supplies if anything is wrong.

drv/medusa-seq-server/src/power_control.rs
Aaron-Hartwig: I would not be against defining these pins in the TOML. That was just not something I knew how to do offhand, hence the hard coding. I also think it is low-risk that any of this changes for this design since we're not really pin constrained at all.

drv/transceivers-server/Cargo.toml
Aaron-Hartwig: It certainly could!

task/thermal/Cargo.toml
Aaron-Hartwig: Weird, I have no idea how this happened. I'll revert.

task/thermal/src/bsp/medusa_a.rs
Aaron-Hartwig: yeah, its weird! This is one of the ways this board is breaking our usual mold. Definitely makes one ponder how modularize certain pieces are split apart.

Issue #1755, opened 2024-04-18
Want to separate H753 image download from slot activate


rmustacc:
Today the H753 update server basically requires that the act of writing and activating a slot be joined together. Fundamentally we want to separate out the following operations into three things that can be run separately:

1. Write an image to a slot.
2. Activate a slot.
3. Reset the SP.

The most important thing is making sure that 1/2 are no longer joined as it means that if you want to rollback because you booted an old image that had a broken say Sidecar front I/O transceiver, then we could roll that back without doing a manual update. This likely needs changes to the h753 update server and then the control plane agent.

mkeeter:
I'd even suggest making steps 2+3 an atomic API; the ability to leave a slot activated _without_ resetting seems like a footgun.  (This would also play nicely with #1707)

rmustacc:
They were left separately on purpose and this was intentional. In particular @cbiffle brought up the desire to not have them joined together as well.

cbiffle:
Yeah, I feel like leaving them separate is safer in general. We've got a few cases where we've treated inter-system interfaces like this "cleverly" and bonded verbs and states together, and it winds up biting us. If there isn't a correctness issue in leaving them separate, I'd make each operation separate and well-defined on its own.

Switching the active slot does raise the spectre of a "surprise reboot" into the new image if the SP were to, like, power cycle before you sent the reboot -- I think this should be clearly documented on the operation, but we probably shouldn't attempt to prevent it.

labbott:
https://github.com/oxidecomputer/hubris/pull/1808 should take care of this. This should now work for setting it via the gateway API but actually using it in, say, wicket also requires gateway changes.

labbott:
Fixed by #1808 on the hubris side

PR #1808, opened 2024-06-05
Allow SP to set active slot

labbott:
It's useful for the SP to be able to change the active slot. Pull this through. The final finish step no longer implicitly swaps the bank.

### DIFF
--- a/drv/stm32h7-update-api/src/lib.rs
+++ b/drv/stm32h7-update-api/src/lib.rs
@@ -5,7 +5,9 @@
 #![no_std]
 
 use drv_caboose::CabooseError;
-use userlib::sys_send;
+use hubpack::SerializedSize;
+use serde::{Deserialize, Serialize};
+use userlib::{sys_send, FromPrimitive};
 
 pub use stage0_handoff::ImageVersion;
 
@@ -25,5 +27,36 @@ pub const FLASH_WORDS_PER_BLOCK: usize = 32;
 pub const BLOCK_SIZE_BYTES: usize = FLASH_WORD_BYTES * FLASH_WORDS_PER_BLOCK;
 
 pub const BLOCK_SIZE_WORDS: usize = BLOCK_SIZE_BYTES / 4;
+
+#[derive(
+    Clone,
+    Copy,
+    Eq,
+    PartialEq,
+    FromPrimitive,
+    Serialize,
+    Deserialize,
+    SerializedSize,
+)]
+pub enum SlotId {
+    Active = 0,
+    Inactive = 1,
+}
+
+impl TryFrom<u16> for SlotId {
+    type Error = ();
+    fn try_from(i: u16) -> Result<Self, Self::Error> {
+        Self::from_u16(i).ok_or(())
+    }
+}
+
+impl From<SlotId> for u16 {
+    fn from(id: SlotId) -> u16 {
+        match id {
+            SlotId::Active => 0,
+            SlotId::Inactive => 1,
+        }
+    }
+}
 
 include!(concat!(env!("OUT_DIR"), "/client_stub.rs"));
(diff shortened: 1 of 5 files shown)


drv/stm32h7-update-server/src/main.rs
```diff
@@ -279,6 +286,24 @@ impl<'a> ServerImpl<'a> {
...
+            self.swap_banks()?;
+        }
+        Ok(())
+    }
+
+    fn get_pending_boot_slot(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<SlotId, RequestError<Infallible>> {
+        Ok(self.pending)
```
lzrd, line 304: Is this state  retrievable from a HW register? Just thinking of the corner case where update_server restarts and having two sources of truth.

lzrd: Ok... I see the fetch of HW state below. Looks ok.

idl/stm32h7-update.idol
```diff
@@ -69,5 +69,24 @@ Interface(
             ),
             idempotent: true,
         ),
+        "get_pending_boot_slot": (
```
lzrd, line 72: The idl/lpc55-update-server.idol `switch_default_image` API doesn't fit well enough to reuse here.
Someday we could unify the boot policy APIs.

lzrd (approved):

labbott:
Fixes https://github.com/oxidecomputer/hubris/issues/1755

drv/stm32h7-update-api/src/lib.rs
```diff
@@ -26,4 +28,35 @@ pub const BLOCK_SIZE_BYTES: usize = FLASH_WORD_BYTES * FLASH_WORDS_PER_BLOCK;
...
+    Serialize,
+    Deserialize,
+    SerializedSize,
+)]
+pub enum SlotId {
+    Active = 0,
+    Inactive = 1,
+}
+
+impl TryFrom<u16> for SlotId {
```
cbiffle, line 46: If you're using this for something other than IPC, carry on.

For IPC specifically, your use of `encoding: Hubpack` relieves you from needing these instances (and explicit `= X` discriminator values). FYI.

drv/stm32h7-update-server/src/main.rs
```diff
@@ -519,9 +543,24 @@ impl NotificationHandler for ServerImpl<'_> {
 fn main() -> ! {
     let flash = unsafe { &*device::FLASH::ptr() };
 
+    // If the server restarts we need to fix our pending state
```
cbiffle, line 546: Nice, this looks like correct handling of task restart.

cbiffle (approved):
Not a lot of actual driver changes required, looks like!

drv/stm32h7-update-api/src/lib.rs
```diff
@@ -26,4 +28,35 @@ pub const BLOCK_SIZE_BYTES: usize = FLASH_WORD_BYTES * FLASH_WORDS_PER_BLOCK;
...
+    Copy,
+    Eq,
+    PartialEq,
+    FromPrimitive,
+    Serialize,
+    Deserialize,
+    SerializedSize,
+)]
+pub enum SlotId {
+    Active = 0,
```
aapoalas, line 42: **question:** From the peanut gallery: Does the naming of the slot IDs come from an outside source, or is it otherwise well understood within Oxide?

From an outsider perspective, it seems weird that apparently there are two banks that you can boot to, so either can be the "actively booted bank" but one is still called "inactive" and the other is "active", regardless of which one is actually currently booted into.

I guess these aren't actually slot/bank IDs but rather just a value that can be set as 1 to say "after restart, switch from current bank to other bank"? If that is the case, would this rather be a "pending boot slot value"?

Issue #1570, opened 2023-12-04
software anti-rollback epoch mechanism


morlandi7:
A software anti-rollback epoch mechanism implemented by update_server using an epoch stored in the caboose needs to be implemented. A well-behaved update_server would refuse to update and/or enable an image with an epoch numerically lower than its own.

see #1127 
related #1569
possibly could be solved by #1366

PR #1809, opened 2024-06-10
Implement a software rollback protection policy in RoT update_server.

lzrd:
Issue #1570 software anti-rollback
 
Implement a software rollback protection policy in RoT update_server using:
  - the `epoch` value in the optional ImageHeader for early rejection (1st block presented before any flash erase/write) or 
  - the `EPOC` tag in optional caboose for rejecting activation of an image.
 
"optional" because there are existent Bootleby images without ImageHeader or Caboose and Hubris images images in the future that may eliminate ImageHeader. In cases where an epoch value is not present, the image epoch is treated as equal to zero. 

Related PRs:
https://github.com/oxidecomputer/sprot-release/pull/11
https://github.com/oxidecomputer/hubtools/pull/32
https://github.com/oxidecomputer/permission-slip/pull/205
https://github.com/oxidecomputer/management-gateway-service/pull/240


drv/lpc55-update-server/src/images.rs
```diff
@@ -23,82 +31,569 @@ extern "C" {
...
+                })
+            } else {
+                None
+            }
+        }) {
+            Some(value) => Epoch { value },
+            None => Epoch::from(0),
+        }
+    }
+}
```
labbott, line 287: This seems like it's reimplementing https://doc.rust-lang.org/stable/core/primitive.u32.html#method.from_str_radix

lzrd: I had the EPOC format as 10 ASCII decimal digits and that I would be checking format and comparing u8's.
But, since adding support for using the u32 from ImageHeader for early checking during update, the EPOC value from the caboose needs to be converted to a u32 and the leading zeros don't add any value any more. I'm removing the requirement for the EPOC value being a [u8; 10].
`from_str_radix()` has a call to panic if the radix is outside 2..=36. That probably gets optimized out since the radix is a constant. It also requires a &str which means the &[u8] needs to be legal utf8. This is foreign data (pre-signature check), so we cannot avoid the check without risking UB.
This explicit conversion makes it easier, for me at least, to reason about the code.

drv/lpc55-update-server/src/images.rs
```diff
@@ -23,82 +31,605 @@ extern "C" {
...
+                }
+            } else {
+                None
+            }
+        }) {
+            Some(value) => Epoch { value },
+            None => Epoch::from(0u32),
+        }
+    }
+}
```
labbott, line 298: I know you said it makes it easier for you to reason about but I do not find this easy to review. I would really rather see us use a standard library function as a first pass and then write our own if we can't actually deal with the panics.

drv/lpc55-update-server/src/images.rs
```diff
@@ -23,82 +31,605 @@ extern "C" {
...
-    image_range(component, slot).0.start
+    flash_range(component, slot).store.start
         == unsafe { &__this_image } as *const _ as u32
 }
 
+pub struct FlashRange {
+    pub store: Range<u32>,
+    pub exec: Range<u32>,
+}
+
```
labbott, line 231: Can we break this part of the change out into a separate PR to review?

drv/lpc55-update-server/src/images.rs
```diff
@@ -23,82 +31,605 @@ extern "C" {
...
+        // True if this is a complete image.
+        complete: bool,
+    },
+    // The image is ram[0..] concatenated with flash span[ram.len()..]
+    Hybrid {
+        buffer: &'a [u8],
+        flash: &'a drv_lpc55_flash::Flash<'a>,
+        span: &'a Range<u32>,
+    },
+}
```
labbott, line 331: I think we should break this out into a separate PR to review as well to make sure we're getting the abstraction right

drv/lpc55-update-server/src/images.rs
lzrd: I'm fine with that.

drv/lpc55-update-server/src/images.rs
lzrd: Done PR #1813

drv/lpc55-update-server/src/images.rs
lzrd: Done in impl From<&[u8]> for Epoch { fn from(chars: &[u8]) -> Self { ... chars.parse::<u32>(). ... }

drv/lpc55-update-server/src/images.rs
lzrd: Done in POR #1813

drv/stm32h7-update-server/src/main.rs
```diff
@@ -416,13 +416,14 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
         Ok(BLOCK_SIZE_BYTES)
     }
 
+    /// Deprecated. The version and epoch values are in the Caboose.
```
mkeeter, line 419: Can we remove this entirely?  I don't see anyone calling it anymore.

app/gimletlet/app-mgmt.toml
```diff
@@ -49,7 +49,7 @@ task-slots = ["sys", "user_leds"]
 
 [tasks.net]
 name = "task-net"
-stacksize = 3000
+stacksize = 8000
```
mkeeter, line 52: I'm assuming this is an unrelated change which you discovered when building every `app-*.toml`?

build/xtask/src/dist.rs
```diff
@@ -1567,7 +1567,7 @@ fn build_kernel(
 /// Returns true if the header was found and updated,
 /// false otherwise.
 fn update_image_header(
-    cfg: &PackageConfig,
+    _cfg: &PackageConfig,
```
mkeeter, line 1570: Can we just remove this argument, since it's now unused?

drv/lpc55-update-api/src/lib.rs
```diff
@@ -242,6 +242,16 @@ impl From<RotSlot> for SlotId {
     }
 }
 
+impl SlotId {
+    pub fn other(&self) -> SlotId {
+        if *self == SlotId::A {
+            SlotId::B
+        } else {
+            SlotId::A
+        }
```
mkeeter, line 251:
```suggestion
        match self {
            SlotId::A => SlotId::B,
            SlotId::B => SlotId::A,
        }
```
slightly shorter, and will break loudly if someone adds a third slot

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
 
-// NXP LPC55's mixed header/vector table offsets
-const RESET_VECTOR_OFFSET: usize = 0x04;
-pub const LENGTH_OFFSET: usize = 0x20;
-pub const HEADER_OFFSET: u32 = 0x130;
-const MAGIC_OFFSET: usize = HEADER_OFFSET as usize;
+// An image may have an ImageHeader located after the
+// LPC55's mixed header/vector table.
+pub const IMAGE_HEADER_OFFSET: u32 = 0x130;
+pub const CABOOSE_TAG_EPOC: [u8; 4] = [b'E', b'P', b'O', b'C'];
```
mkeeter, line 53: Nit: this could be written as
```suggestion
pub const CABOOSE_TAG_EPOC: [u8; 4] = *b"EPOC";
```

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+                at_runtime: __IMAGE_STAGE0_BASE.as_ptr() as u32
+                    ..__IMAGE_STAGE0_END.as_ptr() as u32,
+            },
+        }
+    }
+}
+
+/// Does (component, slot) refer to the currently running Hubris image?
+pub fn is_current_hubris_image(component: RotComponent, slot: SlotId) -> bool {
+    // Safety: extern statics aren't controlled by Rust so poking them can
```
mkeeter, line 103: Why is there a `Safety` comment here?  There's no `unsafe`.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+    initial_pc: u32,                    // 0x04
+    _vector_table_0: [u32; 6],          // 0x08, 0c, 10, 14, 18, 1c
+    nxp_image_length: u32,              // 0x20
+    nxp_image_type: u32,                // 0x24
+    nxp_offset_to_specific_header: u32, // 0x28
+    _vector_table_1: [u32; 2],          // 0x2c, 0x30
+    nxp_image_executation_address: u32, // 0x32
+                                        // Additional trailing vectors are not
+                                        // interesting here.
+                                        // _vector_table_2[u32; 2]
```
mkeeter, line 125: Remove commented out `_vector_table_2`?

drv/lpc55-update-server/src/main.rs
```diff
@@ -781,105 +795,91 @@ impl ServerImpl<'_> {
...
-        )?;
-        if let Some(len) = round_up_to_flash_page(buf[0]) {
-            // The minimum image size should be further constrained
-            // but this is enough bytes for an NXP header and not
-            // bigger than the flash slot.
-            if len as usize <= span.len() && len >= HEADER_OFFSET {
-                return Ok(len as usize);
+    fn cached_stage0_image_len(&self) -> Result<usize, UpdateError> {
+        if let Ok(vectors) = ImageVectorsLpc55::try_from(
+            self.fw_cache[0..BLOCK_SIZE_BYTES].as_bytes(),
```
mkeeter, line 844: Slicing a `BLOCK_SIZE_BYTES` chunk then passing it into `read_from_prefix` doesn't make sense to me.

We should
1) Add a const assertion that `fw_cache` is long enough to fit `core::mem::size_of::<ImageVectorsLpc55>`
2) do something like 
```rust
let vectors = ImageVectorsLpc55::read_from(
    &self.fw_cache[..size_of::<ImageVectorsLpc55>()]
).unwrap_lite()
```
3) Remove `ImageVectorsLpc55`'s `TryFrom` implementation, since this is the only place it's used

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+        slot: SlotId,
+    ) -> ImageAccess<'_> {
+        let span = flash_range(component, slot);
+        ImageAccess {
+            accessor: Accessor::Ram { buffer, span },
+        }
+    }
+
+    pub fn new_hybrid<'a>(
+        flash: &'a drv_lpc55_flash::Flash<'a>,
```
mkeeter, line 381: Nit: if the order in memory is `[buffer, flash]`, then that should also be the argument order her (and constructor order below)

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+    if !vectors.is_image_type_signed_xip()
+        || vectors.nxp_offset_to_specific_header >= vectors.nxp_image_length
+    {
+        // Not a signed XIP image or no signature block.
+        // If we figure out a reasonable minimum size for the signature block
+        // we should test for that.
+        return Err(UpdateError::InvalidHeaderBlock);
+    }
+
+    // We don't rely on the ImageHeader, but if it is there, it needs to be valid.
```
mkeeter, line 210: This PR doesn't remove the `ImageHeader`, right?

I'm wondering why the logic change here; previously, we required the `ImageHeader` if the component is `RotComponent::Hubris` (but allowed for _bootloaders_ that have been released without an `ImageHeader`).

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+                let mut remainder = buffer.len();
+                let end_offset = start_offset
+                    .checked_add(remainder)
+                    .ok_or(UpdateError::OutOfBounds)?;
+                // Transfer data from the RAM portion of the image
+                if start_offset < ram.len() {
+                    let ram_end_offset = ram.len().min(end_offset);
+                    // Transfer starts within the RAM part of this image.
+                    let data = ram
+                        .get((start_offset)..ram_end_offset)
```
mkeeter, line 511:
```suggestion
                        .get(start_offset..ram_end_offset)
```

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+
+    /// Get the rounded up length of an LPC55 image if present.
+    pub fn padded_image_len(&self) -> Result<u32, UpdateError> {
+        let vectors = match self.accessor {
+            Accessor::Flash { .. } => {
+                let buffer =
+                    &mut [0u8; core::mem::size_of::<ImageVectorsLpc55>()];
+                self.read_bytes(0u32, buffer.as_bytes_mut())?;
+                ImageVectorsLpc55::read_from_prefix(&buffer[..])
+                    .ok_or(UpdateError::OutOfBounds)
```
mkeeter, line 550:
```suggestion
                let mut v = ImageVectorsLpc55::new_zeroed();
                self.read_bytes(0u32, v.as_bytes_mut())?;
                v
```

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+//
+// Invalid EPOC values converted to Epoch{value:0} include:
+//   - empty slice
+//   - any non-ASCII-digits in slice
+//   - any non-UTF8 in slice
+//   - leading '+' normally allowed by parse()
+//   - values greater than u32::MAX
+//
+//   Hand coding reduces size by about 950 bytes.
+impl From<&[u8]> for Epoch {
```
mkeeter, line 584: I would vote to return an `Option<Epoch>` instead, or `Result<Epoch, ()>` if you want to use `TryFrom`.  Snapping invalid epoch tags to 0 seems like a recipe for confusion, because it will allow updates to an invalid epoch if the initial epoch is 0.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+}
+
+/// Check a next image against an active image to determine
+/// if rollback policy allows the next image.
+/// If ImageHeader and Caboose are absent or Caboose does
+/// not have an `EPOC` tag, then the Epoch is defaulted to zero.
+/// This test is also used when the header block first arrives
+/// so that images can be rejected early, i.e. before any flash has
+/// been altered.
+pub fn check_rollback_policy(
```
mkeeter, line 613: It seems like this function is only ever useful when `complete` is `true`; otherwise, we can't read the epoch tag from the new image's caboose (so we have to keep loading more of the image).  Could we remove the `complete` flag and only call the function on complete images?

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +47,617 @@ extern "C" {
...
+) -> Result<(), UpdateError> {
+    let next_epoch = get_image_epoch(&next_image)?;
+    let active_epoch = get_image_epoch(&active_image)?;
+    match (active_epoch, next_epoch) {
+        // No active_epoch is treated as zero; update can proceed.
+        (None, _) => Ok(()),
+        (Some(active_epoch), None) => {
+            // If next_image is partial and HEADER_BLOCK has no ImageHeader,
+            // then there is no early rejection, proceed.
+            if !complete || active_epoch.can_update_to(Epoch::from(0u32)) {
```
mkeeter, line 626: Same vague concerns as before: being able to update to an invalid / missing epoch (but only from epoch 0) seems odd to me.

PR #1813, opened 2024-06-12
Add an image access abstraction to RoT update_server

lzrd:
The upcoming rollback protection feature needs to access partial or complete
images in flash, ram, or a mix.
The ImageAccess abstraction makes these operations more readable.


aapoalas (commented):
Comments from the peanut gallery. Looks like a very nice abstraction layer in general.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +46,499 @@ extern "C" {
...
+    }
+
+    pub fn _new_hybrid<'a>(
+        flash: &'a drv_lpc55_flash::Flash<'a>,
+        buffer: &'a [u8],
+        component: RotComponent,
+        slot: SlotId,
+    ) -> ImageAccess<'a> {
+        assert!((buffer.len() % BYTES_PER_FLASH_PAGE) == 0);
+        assert!(((buffer.as_ptr() as u32) % U32_SIZE) == 0);
```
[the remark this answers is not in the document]
aapoalas: Is an `if cond panic!()` cheaper?

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +46,499 @@ extern "C" {
...
+    pub exec: Range<u32>,
+}
+
+/// Get the flash storage address range and flash execution address range.
+pub fn flash_range(component: RotComponent, slot: SlotId) -> FlashRange {
+    // Safety: Linker defined symbols are trusted.
+    unsafe {
+        match (component, slot) {
+            (RotComponent::Hubris, SlotId::A) => FlashRange {
+                store: __IMAGE_A_BASE.as_ptr() as u32
```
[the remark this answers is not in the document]
lzrd: Most of the ARM hardware interfaces and all of the addressing are u32 while rust array indexing always uses usize. This is an area that could use general clean-up for readability.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +46,499 @@ extern "C" {
...
+                    ..__IMAGE_STAGE0NEXT_END.as_ptr() as u32,
+                exec: __IMAGE_STAGE0_BASE.as_ptr() as u32
+                    ..__IMAGE_STAGE0_END.as_ptr() as u32,
+            },
+        }
+    }
+}
+
+/// Does (component, slot) refer to the currently running Hubris image?
+pub fn same_image(component: RotComponent, slot: SlotId) -> bool {
```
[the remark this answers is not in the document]
lzrd: A good suggestion for readability.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +46,499 @@ extern "C" {
...
+        let span = flash_range(component, slot);
+        match (flash, buffer) {
+            (Some(flash), None) => ImageAccess {
+                accessor: Accessor::Flash { flash, span },
+            },
+            (None, Some(buffer)) => ImageAccess {
+                accessor: Accessor::Ram { buffer, span },
+            },
+            (Some(flash), Some(buffer)) => {
+                if buffer.len() < BYTES_PER_FLASH_PAGE {
```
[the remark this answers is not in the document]
lzrd: Currently, this buffer is passed from an Option<[u8; 512]> which is resulting in a non-u32-aligned array (address ended in 0x...5 last I checked.)

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +46,499 @@ extern "C" {
...
+                        .get((offset)..ram_end_offset)
+                        .ok_or(UpdateError::OutOfBounds)?;
+                    buffer.copy_from_slice(data);
+                    remainder -= data.len();
+                    offset = ram_end_offset;
+                }
+                // Transfer data from the flash-backed portion of the image.
+                if remainder > 0 {
+                    let start =
+                        offset.saturating_add(span.store.start as usize);
```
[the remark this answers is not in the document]
lzrd: That is more readable.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +46,499 @@ extern "C" {
...
+                self.read_bytes(0u32, buffer.as_bytes_mut())?;
+                ImageVectorsLpc55::read_from_prefix(&buffer[..])
+                    .ok_or(UpdateError::OutOfBounds)
+            }
+            Accessor::Ram { buffer, span: _ }
+            | Accessor::_Hybrid {
+                buffer,
+                flash: _,
+                span: _,
+            } => ImageVectorsLpc55::read_from_prefix(buffer)
```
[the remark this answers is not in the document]
lzrd: The RAM buffer is not aligned.

drv/lpc55-update-server/src/images.rs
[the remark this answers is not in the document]
aapoalas: I would be interested in contributing this sort of "grunt work" but I am working on a Linux machine; does there happen to be a way to cross compile Hubris on Linux?

I tried doing some cleaning a while ago but ran into some crates either not being cross compilable or requiring access to private repositories. Either that, or I managed to completely miss the "building on Linux" header in some readme or another, in which case I apologise for the dumb question :)

drv/lpc55-update-server/src/images.rs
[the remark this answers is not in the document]
lzrd: The asserts were in for testing and don't need to be there.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +46,499 @@ extern "C" {
...
+        buffer: &'a [u8],
+        span: FlashRange,
+    },
+    // Hybrid is used for later implementation of rollback protection.
+    // The buffer is used in place of the beginning of the flash range.
+    _Hybrid {
+        buffer: &'a [u8],
+        flash: &'a drv_lpc55_flash::Flash<'a>,
+        span: FlashRange,
+    },
```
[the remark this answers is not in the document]
lzrd: Yes. It's used in the next PR for rollback protection which will remove the underscores.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +46,499 @@ extern "C" {
...
-            ),
-            (RotComponent::Stage0, SlotId::B) => (
-                __IMAGE_STAGE0NEXT_BASE.as_ptr() as u32
-                    ..__IMAGE_STAGE0NEXT_END.as_ptr() as u32,
-                __IMAGE_STAGE0_BASE.as_ptr() as u32
-                    ..__IMAGE_STAGE0_END.as_ptr() as u32,
-            ),
+// Accessor keeps the implementation details of ImageAccess private
+enum Accessor<'a> {
+    // Flash driver, flash device range
```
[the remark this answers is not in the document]
lzrd: done

drv/lpc55-update-server/src/images.rs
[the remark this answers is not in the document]
lzrd: done

drv/lpc55-update-server/src/images.rs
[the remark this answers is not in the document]
lzrd: removed

drv/lpc55-update-server/src/images.rs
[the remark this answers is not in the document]
aapoalas: A dumb question indeed. I wonder how I missed the "Linux first, don't use cargo build directly" part...

lzrd:
This change adds 768 bytes to the image flash size
```
          master/xtask-oxide-rot-1_app-dev.log.ok:  flash:   0x39c80 (90%)
rot-image-access/xtask-oxide-rot-1_app-dev.log.ok:  flash:   0x39f80 (90%)
echo $(( 0x39f80 - 0x39c80 ))
768

              master/xtask-rot-carrier_app.log.ok:  flash:   0x35aa0 (83%)
    rot-image-access/xtask-rot-carrier_app.log.ok:  flash:   0x35da0 (84%)

echo $(( 0x35da0 - 0x35aa0 ))
768
```

New sizes for oxide-rot-1
```
flash   = 0x00050000..0x00090000
ram     = 0x20004000..0x2001c000
sram2   = 0x20020000..0x20030000
usbsram = 0x40102000..0x40103000
dice_certs = 0x40100000..0x40100a00
dice_alias = 0x40100a00..0x40101200
Used:
  flash:   0x39f80 (90%)
  ram:     0x13d20 (82%)
  sram2:   0x0 (0%)
  usbsram: 0x1000 (100%)
  dice_certs: extern region (attest)
  dice_alias: extern region (attest)
warning: memory allocation is sub-optimal
Suggested improvements:
kernel:
  ram:    3040  (currently 4096)
warning: memory allocation is sub-optimal
Suggested improvements:
kernel:
  ram:    3040  (currently 4096)
APP=app/oxide-rot-1/app-dev.toml
```

lzrd:
There is work to do to clear fmt and panics out of the the update_server:

On master:
```
$ arm-none-eabi-objdump -C -d target/oxide-rot-1-dev/dist/a/update_server  | tee update_server.s
$ grep '   b.*panic' update_server.s  | wc -l
50
```
On this PR:
```
$ arm-none-eabi-objdump -C -d target/oxide-rot-1-dev/dist/a/update_server  | tee update_server.s
$ grep '   b.*panic' update_server.s  | wc -l
59
```

drv/lpc55-update-server/src/images.rs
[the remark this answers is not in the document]
lzrd: If you're interested in following development here and if you don't have them already, get yourself one each of https://www.st.com/en/evaluation-tools/nucleo-h753zi.html and https://www.nxp.com/design/design-center/software/development-software/mcuxpresso-software-and-tools-/lpcxpresso-boards/lpcxpresso55s69-development-board:LPC55S69-EVK

labbott:
> This change adds 768 bytes to the image flash size
> 

I consider this reasonable

drv/lpc55-update-server/src/images.rs
```diff
@@ -2,12 +2,31 @@
...
+use drv_lpc55_update_api::{RawCabooseError, RotComponent, SlotId};
 use drv_update_api::UpdateError;
-use userlib::UnwrapLite;
+use zerocopy::{AsBytes, FromBytes};
 
+// Our layout of flash banks on the LPC55.
+// Addresses are from the linker.
+// The bootloader (`bootleby`) resides at __STAGE0_BASE and
+// only cares about IMAGE_A and IMAGE_B.
 // We shouldn't actually dereference these. The types are not correct.
```
cbiffle, line 30: Nit: after we switched these to zero-length arrays, dereferencing is actually safe and the types are arguably correct.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
-pub fn validate_header_block(
-    component: RotComponent,
-    slot: SlotId,
-    block: &[u8; BLOCK_SIZE_BYTES],
-) -> Result<(), UpdateError> {
-    let exec = image_range(component, slot).1;
+/// Address ranges that may contain an image during storage and active use.
+/// `store` and `exec` ranges are the same except for `stage0next`.
+pub struct FlashRange {
+    pub store: Range<u32>,
```
cbiffle, line 55: Not a problem in our current address space layout, but in general it seems like these should be RangeInclusive --- otherwise they can't model a range at the end of the u32 space.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+    pub exec: Range<u32>,
+}
+
+/// Get the flash storage address range and flash execution address range.
+pub fn flash_range(component: RotComponent, slot: SlotId) -> FlashRange {
+    // Safety: Linker defined symbols are trusted.
```
cbiffle, line 61: How about this for a safety comment?

```rust
// Safety: this block requires unsafe code to generate references to the extern "C" statics.
// Because we're only getting their addresses (all operations below are just as_ptr),
// we can't really trigger any UB here. The addresses themselves are assumed to
// be valid because they're produced by the linker, which we implicitly trust.
```

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+            },
+        }
+    }
+}
+
+/// Does (component, slot) refer to the currently running Hubris image?
+pub fn is_current_hubris_image(component: RotComponent, slot: SlotId) -> bool {
+    // Safety: We are trusting the linker.
+    flash_range(component, slot).store.start
+        == unsafe { &__this_image } as *const _ as u32
```
cbiffle, line 96: Consider using `addr_of!` for this and the other static address computation expressions --- I think it's clearer, and it's _much_ harder to get wrong.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+                exec: __IMAGE_STAGE0_BASE.as_ptr() as u32
+                    ..__IMAGE_STAGE0_END.as_ptr() as u32,
+            },
+        }
+    }
+}
+
+/// Does (component, slot) refer to the currently running Hubris image?
+pub fn is_current_hubris_image(component: RotComponent, slot: SlotId) -> bool {
+    // Safety: We are trusting the linker.
```
cbiffle, line 94: FWIW you can get more specific information about _why_ this requires `unsafe` by removing it. The current compiler says:

>note: extern statics are not controlled by the Rust type system: invalid data, aliasing violations or data races will cause undefined behavior

We can use that information to write a more precise safety comment that says something like

```rust
// Safety: extern statics aren't controlled by Rust so poking them can cause UB; in this case, it's
// zero length and we are only taking its numerical address, so we're not at risk.
```

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+    }
+
+    /// Image length padded to nearest page size.
+    pub fn padded_image_len(&self) -> Option<u32> {
+        round_up_to_flash_page(self.image_length()?)
+    }
+
+    /// Determine the bounds of an image assuming the given flash bank
+    /// addresses.
+    pub fn padded_image_range(&self, exec: &Range<u32>) -> Option<Range<u32>> {
```
cbiffle, line 150: Is Clippy cool with passing a range by-ref here? It's a small enough type that I'd expect a warning.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+        let image_start = exec.start;
+        let image_end = exec.start.saturating_add(self.padded_image_len()?);
```
cbiffle, line 152: Note that this will silently ignore cases where start+len overflows. If saturating arithmetic is correct here, please add a comment explaining why. (I expected `checked_add` to validate the image.)

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+        // ImageHeader is optional. Assuming code cannot be earlier than
+        // IMAGE_HEADER_OFFSET.
+        let initial_pc_start = image_start.saturating_add(IMAGE_HEADER_OFFSET);
+        let initial_pc_end =
+            image_start.saturating_add(self.nxp_offset_to_specific_header);
+        let pc = exec.start.saturating_add(self.normalized_initial_pc());
+
+        if !exec.contains(&image_end)
```
cbiffle, line 160: Because this isn't using `RangeInclusive` and you're adding the length to the start, an image that takes up the entire flash slot won't pass this check.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+    }
+
+    /// Determine the bounds of an image assuming the given flash bank
+    /// addresses.
+    pub fn padded_image_range(&self, exec: &Range<u32>) -> Option<Range<u32>> {
+        let image_start = exec.start;
+        let image_end = exec.start.saturating_add(self.padded_image_len()?);
+        // ImageHeader is optional. Assuming code cannot be earlier than
+        // IMAGE_HEADER_OFFSET.
+        let initial_pc_start = image_start.saturating_add(IMAGE_HEADER_OFFSET);
```
cbiffle, line 155: What does "pc" mean here? Because this is _not_ the image's initial program counter, which is at offset 4.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+        let initial_pc_end =
+            image_start.saturating_add(self.nxp_offset_to_specific_header);
+        let pc = exec.start.saturating_add(self.normalized_initial_pc());
```
cbiffle, line 158: The initial PC at offset 4 is an absolute address, adding it to exec.start here is probably not right?

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+        }
+        Some(image_start..image_end)
+    }
+}
+
+impl TryFrom<&[u8]> for ImageVectorsLpc55 {
+    type Error = ();
+
+    fn try_from(buffer: &[u8]) -> Result<Self, Self::Error> {
+        match ImageVectorsLpc55::read_from_prefix(buffer) {
```
cbiffle, line 177: Are you comfortable with the fact that this will only work on 32-bit aligned u8 arrays? If so, it needs a comment to that effect.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+    }
+}
+
+/// The update_server needs to deal with images that are:
+///   - Entirely in flash
+///   - Header block in RAM with the remainder unavailable
+///   - Header block in RAM with the remainder in flash.
+///   - Entire image in RAM (cached Stage0)
+/// Calls to methods use image relative addresses.
+/// ImageAccess::Flash{_, span} gives the physical flash addresses
```
cbiffle, line 335: What is `ImageAccess::Flash{_, span}`? I can't make sense of it.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+    },
+}
+
+impl Accessor<'_> {
+    fn exec(&self) -> &Range<u32> {
+        match self {
+            Accessor::Flash { flash: _, span }
+            | Accessor::Ram { buffer: _, span }
+            | Accessor::_Hybrid {
+                buffer: _,
```
cbiffle, line 321: FYI, consider `Acessor::_Hybrid { span, .. }` here

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+        flash: &'a drv_lpc55_flash::Flash<'a>,
+        span: FlashRange,
+    },
+    Ram {
+        buffer: &'a [u8],
+        span: FlashRange,
+    },
+    // Hybrid is used for later implementation of rollback protection.
+    // The buffer is used in place of the beginning of the flash range.
+    _Hybrid {
```
cbiffle, line 308: Why does this name start with an underscore, out of curiosity?

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+        component: RotComponent,
+        slot: SlotId,
+    ) -> ImageAccess<'_> {
+        let span = flash_range(component, slot);
+        ImageAccess {
+            accessor: Accessor::Ram { buffer, span },
+        }
+    }
+
+    pub fn _new_hybrid<'a>(
```
cbiffle, line 368: I assume this is underscored because the code isn't being used?

Could we not commit it if it's not tested instead?

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+        ImageAccess {
+            accessor: Accessor::_Hybrid {
+                flash,
+                buffer,
+                span,
+            },
+        }
+    }
+
+    fn exec(&self) -> &Range<u32> {
```
cbiffle, line 384: What is an exec? If I saw a call to this I would assume that you were asking it to execute something.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+            | Accessor::_Hybrid {
+                buffer: _,
+                flash: _,
+                span,
+            }
+            | Accessor::Ram { buffer: _, span } => {
+                span.store.end.saturating_sub(span.store.start)
+            }
+        };
+        offset < len && offset.saturating_add(U32_SIZE) <= len
```
cbiffle, line 401: So, I'm increasingly convinced as I read through this that every use of saturating arithmetic in this module should be replaced by checked. This does _not_ test the condition the comments suggest in the edge case where the integers could overflow.

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+/// Accessor keeps the implementation details of ImageAccess private
+enum Accessor<'a> {
+    // Flash driver, flash device range
+    Flash {
+        flash: &'a drv_lpc55_flash::Flash<'a>,
+        span: FlashRange,
+    },
+    Ram {
+        buffer: &'a [u8],
+        span: FlashRange,
```
cbiffle, line 304: So, a number of places in the access code below appear to rely on buffer.len() being equal to the length of this span. How is this invariant maintained? Would it make more sense to have span be a base address instead, so that the len isn't duplicated (and thus doesn't have to be kept in sync)?

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+    }
+
+    pub fn new_ram(
+        buffer: &[u8],
+        component: RotComponent,
+        slot: SlotId,
+    ) -> ImageAccess<'_> {
+        let span = flash_range(component, slot);
+        ImageAccess {
+            accessor: Accessor::Ram { buffer, span },
```
cbiffle, line 364: (Here is a place where the length-matches invariant is definitely not maintained, fwiw.)

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +45,494 @@ extern "C" {
...
+                let mut word = 0u32;
+                indirect_flash_read(flash, addr, word.as_bytes_mut())?;
+                Ok(word)
+            }
+            Accessor::Ram { buffer, span: _ } => {
+                match buffer
+                    .get(offset as usize..(offset as usize + SIZEOF_U32))
+                    .and_then(u32::read_from)
+                {
+                    Some(word) => Ok(word),
```
cbiffle, line 421: FWIW this is equivalent to `.ok_or(UpdateError::OutOfBounds)?`

cbiffle (commented):
I'm concerned about the image validation logic, would it be possible to pull this into a no_std crate so we can have unit tests?

drv/lpc55-update-server/src/images.rs
lzrd: Added a TODO for now. I'll come back to this.

drv/lpc55-update-server/src/images.rs
lzrd: LGTM

drv/lpc55-update-server/src/images.rs
lzrd: Done

drv/lpc55-update-server/src/images.rs
lzrd: Clippy does not complain about this.

drv/lpc55-update-server/src/images.rs
lzrd: Yeah. That's a poor choice of name.
`first_possible_pc` and `last_possible_pc` would be better names.
However, this check is off-topic for this function and is also redundant with `validate_header_block`. So, I'll remove these computations and tests. See `code_offset` in `validate_header_block` for correct similar tests.

There should be a thorough test before signing on images that checks that all of the vectors land in a text segment.

drv/lpc55-update-server/src/images.rs
lzrd: Removed for reasons above.

drv/lpc55-update-server/src/images.rs
lzrd: read_from_prefix copies into an aligned struct.

```
$ cat src/main.rs 
use anyhow::*;
use zerocopy::*;

#[derive(Debug, FromBytes, FromZeroes, Default)]
#[repr(C)]
struct Msg {
    n: u32,
    b: u16,
}

fn main() {
    let bytes: &[u8] = &[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
    println!("data: {:?}", bytes);

    for start in 0..(bytes.len() - core::mem::size_of::<Msg>()) {
        let msg = Msg::read_from_prefix(&bytes[start..])
            .ok_or_else(|| anyhow!("failed to parse Msg"))
            .unwrap();
        println!("msg @{}: {:#x?}", start, msg);
    }
}
$ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/struct-from-bytes`
data: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
msg @0: Msg {
    n: 0x3020100,
    b: 0x504,
}
msg @1: Msg {
    n: 0x4030201,
    b: 0x605,
}
msg @2: Msg {
    n: 0x5040302,
    b: 0x706,
}
```

drv/lpc55-update-server/src/images.rs
lzrd: Old comment fragments. Edited to make sense.

drv/lpc55-update-server/src/images.rs
lzrd: Thank you

drv/lpc55-update-server/src/images.rs
lzrd: It's a tradeoff between a complete ImageAccess implementation now vs adding it in on the next PR. The _Hybrid case isn't used yet, but I thought it made sense to show it here.

drv/lpc55-update-server/src/images.rs
lzrd: I changed these names: s/exec/at_runtime/ and s/store/stored/ which reads better.

drv/lpc55-update-server/src/images.rs
lzrd: I'm making that change which makes it easier to reason about the code.

lzrd:
> I'm concerned about the image validation logic, would it be possible to pull this into a no_std crate so we can have unit tests?

I would like to have a no_std library in the hubtools repo that can be used by any code concerned with Oxide LPC55 images. That would include Hubris lib/lpc55-rot-startup/src/images.rs if practical. Tests would be included there.

I'd rather not refactor that now and the `sprot-e2e` tests are successful for upgrade and rollback to and from the current master branch and the rot-image-access branch.

Created https://github.com/oxidecomputer/hubtools/issues/34

drv/lpc55-update-server/src/images.rs
lzrd: https://github.com/oxidecomputer/hubris/issues/1836

drv/lpc55-update-server/src/images.rs
lzrd: Done

drv/lpc55-update-server/src/images.rs
lzrd: done

drv/lpc55-update-server/src/images.rs
lzrd: Unless all of a slot is in use, it is expected that buffer.len() would always be less than the length of one of the spans.

Perhaps I should change `span` to  `slot_range` (since `Slot` is already in use).
Suggestions welcome.

Span is a FlashRange which describes both the stored and the at_runtime address ranges of the slot that the image can occupy.
`buffer` here is a portion of the image that currently resides in RAM. The rest of the image may not be present.
In the case of a Accessor::Ram variant, the entire image resides in ram. It's destination slot is described by `span.stored` and its runtime slot is described by `span.at_runtime` (new names).

drv/lpc55-update-server/src/images.rs
cbiffle: Ah, my mistake, I confused it with the one that does a direct ref. This is making a copy.

cbiffle:
Looks like all the saturating arithmetic is still in place. I need convincing that this handles bounds cases correctly. Tests would be the fastest way, short of that I'd like to see a laaaaarge rationale comment explaining why the arithmetic isn't checked instead.

drv/lpc55-update-server/src/images.rs
```diff
@@ -2,14 +2,32 @@
...
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+/// Implement LPC55_`update_server`'s knowledge about bits inside an image.
+///
+/// The update server needs to work with partial or otherwise invalid images.
+/// Signature checks are only performed at boot time. The update server
+/// does match FWIDs against the boot-time-info in some cases. But, because
+/// flash areas are mutated during update_server operations and stuff can
+/// happen, any data in the non-active Hubris image needs to be treated as
+/// untrusted (update_server does not alter its own image).
```
aapoalas, line 12: **question:** Idle curiosity and paranoia, but is the active image actually write-protected as well? Or would it theoretically be possible to modify it?

drv/lpc55-update-server/src/images.rs
```diff
@@ -26,79 +44,515 @@ extern "C" {
 // Location of the NXP header
 pub const HEADER_BLOCK: usize = 0;
 
-// NXP LPC55's mixed header/vector table offsets
-const RESET_VECTOR_OFFSET: usize = 0x04;
-pub const LENGTH_OFFSET: usize = 0x20;
-pub const HEADER_OFFSET: u32 = 0x130;
-const MAGIC_OFFSET: usize = HEADER_OFFSET as usize;
+// An image may have an ImageHeader located after the
```
aapoalas, line 47: **question:** Is this presence decided by eg. version of the header, where older images don't have the header? Might be useful information here.

drv/lpc55-update-server/src/images.rs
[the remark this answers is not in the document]
aapoalas: The variant comments are still non-doc comments, if you meant to change them.

aapoalas (commented):
I sort of accidentally came to read this PR anew and found some questions that I thought were interesting. Feel free to ignore these if need or want be; i ended up sitting on these questions for several days but figured eh, let's ask. They're just curiosity, basically.

drv/lpc55-update-server/src/images.rs
lzrd: The active image could be modified by the update_server task if there were a code path to do that.
There is a check testing to see if the slot to be updated is the currently executing Hubris slot. If so, the request is rejected.

drv/lpc55-update-server/src/images.rs
lzrd: The early versions of the bootloader did not have the header and we have had discussions in the past about removing the need for the header. `update_server` tolerates a missing ImageHeader.

Issue #1788, opened 2024-05-20
Hubris: grapefruit build and board toml


nathanaelhuffman:
schematic can be found in the latest grapefruit GH release: https://github.com/oxidecomputer/hardware-grapefruit/releases/

This should be considered a blocking issue for board bring-up, we'd like to be able to run some semblance of hubris and humility on the target hardawre day1 to facilitate hardware checkout by toggling pins, reading i2c etc.

eanders01:
Adding comments from [#1810 now closed as it was redundant to the original issue here]. Additional info for what is required:

Grapefruit connects to our Ruby dev board and is important to [unblock SW work for Cosmo](https://docs.google.com/presentation/d/1nyxET6XQDtX_13XzFpfj0eDYo29SV09l2peKvs1CneE/edit#slide=id.g2cfb6b75621_1_260). Context for the existence and purpose of Grapefruit is located in [RFD 461](https://rfd.shared.oxide.computer/rfd/0461). We are requesting software work to be completed to support board bringup and testing to unblock Cosmo SW work.

Features
- We need a basic image that can load the FPGA and has the I2C and UART ports enabled
- This will allow us to look at the I2C/PMBUS devices, connect to the SP5 UARTs, and program the FPGA with a blinking LED. This is the hubris functionality to load an FPGA binary. @nathanaelhuffman and @cbiffle have had conversations on this and can provide additional info.

Timing
- Boards are at Bench and will start build week of 10 Jun
- Bringup with @ericaasen @nathanaelhuffman week of 24 Jun  << we will need image in-place here

mkeeter:
Done in https://github.com/oxidecomputer/hubris/pull/1814

Issue #1810, opened 2024-06-10
Hubris Work for Grapefruit


eanders01:
Grapefruit connects to our Ruby dev board and is important to [unblock SW work for Cosmo](https://docs.google.com/presentation/d/1nyxET6XQDtX_13XzFpfj0eDYo29SV09l2peKvs1CneE/edit#slide=id.g2cfb6b75621_1_260). Context for the existence and purpose of Grapefruit is located in [RFD 461](https://rfd.shared.oxide.computer/rfd/0461). We are requesting software work to be completed to support board bringup and testing to unblock Cosmo SW work.

Features
- We need a basic image that can load the FPGA and has the I2C and UART ports enabled
- This will allow us to look at the I2C/PMBUS devices, connect to the SP5 UARTs, and program the FPGA with a blinking LED. This is the hubris functionality to load an FPGA binary. @nathanaelhuffman and @cbiffle Nathanael and Cliff L Biffle 

Timing
- Boards are at Bench and will start build week of 10 Jun
- Bringup with @ericaasen @nathanaelhuffman week of 24 Jun  << we will need image in-place here

eanders01:
Closing this out as redundant to #1788. Comments have been carried over to the original issue. Thanks for catching that @rmustacc.

PR #1814, opened 2024-06-13
Initial Grapefruit support

mkeeter:
See [oxidecomputer/hardware-grapefruit](https://github.com/oxidecomputer/hardware-grapefruit) for schematics

The Grapefruit board is a weird mix between Gimletlet, Gimlet, and Sidecar

- Like Gimletlet, it has a KSZ8463 going to ethernet jacks, instead of the full management network
- Like Sidecar, it stores an FPGA image in auxflash (instead of baking it into the image)
- Like Gimlet, there's only one FPGA programmed by the sequencer, so it's not as complicated as Sidecar

Unlike all of our previous boards, it's using a Xilinx FPGA (instead of Lattice), so programming is a bit different.

This PR adds the initial `app.toml`, a new BSP for `net` (due to different pins), and a new `grapefruit-seq` task (which configures the FPGA and does basically nothing else).


mkeeter:
Closes #1788 and #1810

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+use drv_stm32xx_sys_api as sys_api;
+use idol_runtime::{NotificationHandler, RequestError};
+use sha3::{Digest, Sha3_256};
+use userlib::{hl, sys_recv_notification, task_slot, RecvMessage};
+
+use ringbuf::{counted_ringbuf, ringbuf_entry, Count};
+
+#[derive(Copy, Clone, PartialEq, Count)]
+enum Trace {
+    Ice40Init(#[count(children)] bool),
```
nathanaelhuffman, line 20: Bad name maybe? no Ice40 on grapefruit

nathanaelhuffman (approved):
Thanks for looking at this! I'm on vacation so I only did a cursory read but seems reasonable.  One minor naming nit.

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        // Initializing the sequencer failed.
+        Err(e) => {
+            // Tell everyone that something's broken, as loudly as possible.
+            ringbuf_entry!(Trace::StartFailed(e));
+            sys.gpio_set(FAULT_PIN_L); // TODO should this be reset, not set?
+
+            // All these moments will be lost in time, like tears in rain...
+            // Time to die.
```
hawkw, line 64: me when i see my goofy comments copied into new code:
image

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+        sys.gpio_set(FPGA_PROGRAM_L);
+        sys.gpio_configure_output(
+            FPGA_PROGRAM_L,
+            sys_api::OutputType::OpenDrain,
+            sys_api::Speed::Low,
+            sys_api::Pull::None,
+        );
+
+        // Pulse PROGRAM_B low for 1 ms to reset the bitstream
+        // (T_PROGRAM is 250 ns, so this is fine)
```
hawkw, line 126: i'm assuming that this means that 250 ns is the _minimum_?

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+        loop {
+            let init = sys.gpio_read(FPGA_INIT_L) != 0;
+            ringbuf_entry!(Trace::Ice40Init(init));
+            if init {
+                break;
+            }
+
+            // Do _not_ burn CPU constantly polling, it's rude. We could also
+            // set up pin-change interrupts but we only do this once per power
+            // on, so it seems like a lot of work.
```
hawkw, line 142: This is decidedly *not* a big deal, but note that the code in Gimlet that also has the comment about not using pin-change interrupts because it's "a lot of work" long predates our implementation of EXTI support in `stm32xx-sys`, so I think it would have been a lot more work when that was written.

On the other hand, I don't think we currently have an interface for a task to *release* a claimed EXTI notification, so using EXTI for this would permanently occupy our ability to receive interrupts for pin changes on port B, so...we probably shouldn't do that. It might be a good idea to have a way for a task to release its configured EXTI interrupt, so that we can use EXTI for one shot stuff like this without tying up that port forever...

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+        let mut scratch_buf = [0u8; 128];
+        let mut pos = blob.start;
+        let mut sha = Sha3_256::new();
+        let mut decompressor = gnarle::Decompressor::default();
+        while pos < blob.end {
+            let amount = (blob.end - pos).min(scratch_buf.len() as u32);
+            let chunk = &mut scratch_buf[0..(amount as usize)];
+            if let Err(e) = aux.read_slot_with_offset(blob.slot, pos, chunk) {
+                return Err(SeqError::AuxReadError(e));
+            }
```
hawkw, line 174: nitpicky, not important: isn't this just:
```suggestion
            aux.read_slot_with_offset(blob.slot, pos, chunk).map_err(SeqError::AuxReadError)?;
```

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+        // This pin is by default an input, and we never change it
+        loop {
+            let done = sys.gpio_read(FPGA_CONFIG_DONE) != 0;
+            if done {
+                break;
+            } else {
+                ringbuf_entry!(Trace::WaitForDone);
+                hl::sleep_for(2);
+            }
+        }
```
hawkw, line 225: nit, not a big deal: isn't this a `while` loop?
```suggestion
        while sys.gpio_read(FPGA_CONFIG_DONE) != 0 {
            ringbuf_entry!(Trace::WaitForDone);
            hl::sleep_for(2);
        }
```

idl/grapefruit-seq.idol
```diff
@@ -0,0 +1,11 @@
+// Grapefruit Sequencer API
+Interface(
+    name: "Sequencer",
+    ops: {
+        "foo": (
+            args: {},
+            reply: Simple("u8"),
+            idempotent: true,
+        ),
```
hawkw, line 9: assuming this is a work in progress?

task/net/src/bsp/grapefruit.rs
```diff
@@ -0,0 +1,157 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#[cfg(not(feature = "ksz8463"))]
+compile_error!("this BSP requires the ksz8463 feature");
```
hawkw, line 6: this BSP is only enabled if the `"grapefruit"` feature flag is enabled, which itself enables the `"ksz8463"` feature, right? so this compile error should, AFAICT, never occur?

drv/grapefruit-seq-server/src/main.rs
mkeeter: Yup, fixed

app/grapefruit/README.md
```diff
@@ -0,0 +1,4 @@
+# Grapefruit Hubris image
+## References
```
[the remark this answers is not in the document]
mkeeter: I did not know that!

idl/grapefruit-seq.idol
mkeeter: Yeah, this is a placeholder to integrate the Idol compile + runtime stuff

drv/grapefruit-seq-server/src/main.rs
mkeeter: Yup, updated the comment

task/net/src/bsp/grapefruit.rs
mkeeter: Yes, this may also be copypasta
```console
$ rg "this BSP requires"
task/net/src/bsp/sidecar_bcd.rs
11:compile_error!("this BSP requires the ksz8463 and mgmt features");

task/net/src/bsp/grapefruit.rs
6:compile_error!("this BSP requires the ksz8463 feature");

task/net/src/bsp/gimlet_bcdef.rs
6:compile_error!("this BSP requires the ksz8463 and mgmt features");

task/net/src/bsp/gimletlet_mgmt.rs
6:compile_error!("this BSP requires the ksz8463 and mgmt features");

task/net/src/bsp/psc_bc.rs
6:compile_error!("this BSP requires the ksz8463 and mgmt features");

task/net/src/bsp/psc_a.rs
6:compile_error!("this BSP requires the ksz8463 and mgmt features");

task/net/src/bsp/gimletlet_nic.rs
6:compile_error!("this BSP requires the ksz8463 feature");
```

I'm fine deleting it and letting things fail naturally if someone messes it up.

aapoalas (commented):
Comments, or rather questions from the peanut gallery. Please excuse me.

app/grapefruit/app.toml
```diff
@@ -0,0 +1,351 @@
...
+chip = "../../chips/stm32h7"
+memory = "memory-large.toml"
+stacksize = 896
+epoch = 0
+version = 0
+fwid = true
+
+[kernel]
+name = "grapefruit"
+requires = {flash = 32768, ram = 8192}
```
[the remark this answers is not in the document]
mkeeter: We can autosize tasks, but can't autosize the kernel at the moment:

- the kernel needs to know the memory locations of all tasks at (kernel) build time
- tasks are located _after_ the kernel in flash / RAM
- therefore, we have to pick a kernel size _before_ building the kernel, to know task locations

This is all theoretically fixable, but has never been a top priority; wasting KiB here isn't critical.

app/grapefruit/src/main.rs
```diff
@@ -0,0 +1,130 @@
...
+    cortex_m::asm::dsb();
+    // Now, we can read the appropriately-named package register to find out
+    // what package we're on.
+    match p.SYSCFG.pkgr.read().pkg().bits() {
+        0b1000 => {
+            // TFBGA240, yay
+        }
+        _ => {
+            // uh
+            panic!();
```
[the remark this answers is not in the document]
mkeeter: We have a [custom panic handler](https://github.com/oxidecomputer/hubris/blob/b0d962e627e179e3bcd3298a07d535699dc0b5ec/sys/kern/src/fail.rs#L109), which records the panic info in variables with specific names.

[Humililty](https://github.com/oxidecomputer/humility) knows about these variables and will print special debug info if they're populated (see [`epitaph()`](https://github.com/oxidecomputer/humility/blob/9ef19bc397507e86c75dd83a79928d435202400f/humility-core/src/hubris.rs#L2073)).

app/grapefruit/.gitignore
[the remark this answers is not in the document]
hawkw: Maybe we should do a separate PR that just moves all this stuff into the global gitignore?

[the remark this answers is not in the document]
mkeeter: Opened #1816

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+            loop {
+                idol_runtime::dispatch(&mut buffer, &mut server);
+            }
+        }
+
+        // Initializing the sequencer failed.
+        Err(e) => {
+            // Tell everyone that something's broken, as loudly as possible.
+            ringbuf_entry!(Trace::StartFailed(e));
+            sys.gpio_set(FAULT_PIN_L); // TODO should this be reset, not set?
```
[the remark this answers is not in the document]
mkeeter: I pushed a fix to this branch, and did the Gimlet fix in #1817

Aaron-Hartwig:
This is looking good so far, Matt! Thanks for getting it rolling.

app/grapefruit/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+edition = "2021"
+readme = "README.md"
+name = "grapefruit"
```
hawkw, line 4: _suuuuuuper_ nitpicky, feel free to ignore me: it's a little bit weird to me that the `name` key is after other metadata...at first glance it looked like the crate didn't _have_ a name.

task/thermal/src/bsp/grapefruit.rs
```diff
@@ -0,0 +1,140 @@
...
+        &mut self,
+        mut fctrl: impl FnMut(FanControl<'_>),
+    ) -> Result<(), ControllerInitError> {
+        match self.fan_control(0.into()) {
+            Ok(c) => {
+                fctrl(c);
+                Ok(())
+            }
+            Err(e) => Err(e),
+        }
```
hawkw, line 82: style nit, take it or leave it: isn't this just 
```suggestion
        self.fan_control(0.into()).map(fctrl)
```

task/thermal/src/control.rs
```diff
@@ -300,6 +312,7 @@ pub(crate) struct Max31790State {
 }
 
 impl Max31790State {
+    #[allow(dead_code)]
```
hawkw, line 315: should we maybe feature-flag the two different fan-controllers, instead of allowing dead code?

not a blocker either way, i imagine rustc eliminates the dead code?

task/thermal/src/control.rs
```diff
@@ -345,6 +359,65 @@ impl Max31790State {
...
+            initialized: false,
+        };
+        // When we first start up, try to initialize the fan controller a few
+        // times, in case there's a transient I2C error.
+        for remaining in (0..3).rev() {
+            if this.initialize().is_ok() {
+                break;
+            }
+            ringbuf_entry!(Trace::FanControllerInitRetry { remaining });
+        }
```
hawkw, line 387: take it or leave it: maybe instead of just copying this from the `Max31790` implementation, maybe we should just have a 
```rust
fn retry_init(init: impl FnMut() -> Result<...>) {
    for remaining in (0..3).rev() {
        if init().is_ok() {
            break;
        }
        ringbuf_entry!(Trace::FanControllerInitRetry { remaining });
    }
}
```
that they can both use?

drv/i2c-devices/src/emc2305.rs
```diff
@@ -0,0 +1,334 @@
...
+) -> Result<(), ResponseCode> {
+    device.write(&[register as u8, val])
+}
+
+fn write_reg16(
+    device: &I2cDevice,
+    register: Register,
+    val: u16,
+) -> Result<(), ResponseCode> {
+    device.write(&[register as u8, (val >> 8) as u8, (val & 0xff) as u8])
```
hawkw, line 211: nit, take it or leave it: could also be
```suggestion
    let [lo, hi] = u16::to_le_bytes(val);
    device.write(&[register as u8,  lo, hi])
```

although i have no idea which one generates nicer assembly...

drv/i2c-devices/src/emc2305.rs
```diff
@@ -0,0 +1,334 @@
...
+        let device = &self.device;
+
+        // Enable the watchdog at all times
+        let mut config =
+            Configuration(read_reg8(device, Register::Configuration)?);
+        config.set_watchdog_enable(true);
+        write_reg8(device, Register::Configuration, config.0)?;
+
+        for fan in 0..fan_count {
+            let fan = Fan::try_from(fan).unwrap();
```
hawkw, line 245: nit: should this be `unwrap_lite()`?

drv/i2c-devices/src/emc2305.rs
```diff
@@ -0,0 +1,334 @@
...
+            ringbuf_entry!(Trace::ZeroTach(fan));
+            Err(ResponseCode::BadDeviceState)
+        } else {
+            let rpm = 7_680_000 / count;
+            let Ok(rpm) = rpm.try_into() else {
+                ringbuf_entry!(Trace::TachOverflow(count));
+                return Err(ResponseCode::BadDeviceState);
+            };
+            Ok(Rpm(rpm))
+        }
```
hawkw, line 306: very nitpicky, but: my preference would be to write the error condition as an early return, as:
```suggestion
            return Err(ResponseCode::BadDeviceState);
        }
        
        let rpm = 7_680_000 / count;
        let Ok(rpm) = rpm.try_into() else {
            ringbuf_entry!(Trace::TachOverflow(count));
            return Err(ResponseCode::BadDeviceState);
        };
        Ok(Rpm(rpm))
```

drv/i2c-devices/src/emc2305.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+}
+
+impl Validate<ResponseCode> for Emc2305 {
+    fn validate(device: &I2cDevice) -> Result<bool, ResponseCode> {
+        let pid = read_reg8(device, Register::ProductId)?;
+        let mfg = read_reg8(device, Register::MfgId)?;
+
+        // XXX The datasheet has ambiguity about whether PID should be 1011_0100
+        // or 0011_0100
```
hawkw, line 331: :clown_face:

drv/i2c-devices/src/emc2305.rs
```diff
@@ -0,0 +1,334 @@
...
+        // Our settings are the following:
+        //   poles = 2
+        //   n = edges measured (set to 5 in `cfg1` above)
+        //   m = range multiplier (2x, per configuration above)
+        //   f_TACH = tachometer measurement frequency (32 KHz)
+        //   count = reading
+        //
+        // Doing the math, we end up with
+        //   RPM = 7680000 / count
+        let count = ((val[0] as u32) << 5) | (val[1] >> 3) as u32;
```
hawkw, line 290: where do these shifts come from?  i assume this is defined in the datasheet, but it surprised me --- at a glance, i kind of assumed that we were just ORing the two bytes into a 16-bit value, but this is a bit more complex...

drv/i2c-devices/src/emc2305.rs
```diff
@@ -0,0 +1,334 @@
...
+        self.register(Register::Fan1Cfg2)
+    }
+
+    fn tach_count(&self) -> Register {
+        self.register(Register::Tach1ReadingHi)
+    }
+
+    fn pwm_target(&self) -> Register {
+        self.register(Register::Fan1Setting)
+    }
```
hawkw, line 181: this is cute! part of my diseased brain kinda wonders if it could be made `const fn`, but i guess there wouldn't really be any benefit from that because we're looping over the fans at runtime anyway...

hawkw (commented):
The EMC2305 driver looks nice! I left a few smallish suggestions, hope they're helpful!

drv/i2c-devices/src/emc2305.rs
mkeeter: Weirdly, the count is a 12-bit value that's shifted by 3.  We use 5 bits of the low byte (`val[1] >> 3`) and 8 bits of the high byte (`val[0] << 5`)

drv/i2c-devices/src/emc2305.rs
mkeeter: Alas, the `from_u8` function from `#[derive(FromPrimitive)]` is not `const`.  I trust that the compiler will probably inline these functions into a few instructions, though.

drv/i2c-devices/src/emc2305.rs
mkeeter: Sure (and fixed the other `unwrap` in here as well)

task/thermal/src/control.rs
mkeeter: Sure!

task/thermal/src/control.rs
mkeeter: I decided to leave the dead code, because it's at least type-checked (and yeah `rustc` should eliminate it)

PR #1819, opened 2024-06-21
Hiffy rng cleanup

flihp:


labbott (approved):
Deleting code is the best

PR #1820, opened 2024-07-11
Lib lpc55 rng seed

flihp:
This PR implements a new strategy to improve the quality of the initial seeding and reseeding of the PRNG hosted in the `rng_driver` task.


flihp:
I've marked this as a draft and it will remain a draft till:
- it currently crash loops if one of the `dice_*` features hasn't been enabled in the kernel
- has only been tested on rot-carrier & lpc55xpresso dev boards, has not yet been tested on a proper gimlet

The bulk of work here will remain unchanged as I address these last few points so reviewing this while still in draft won't be wasted time.

labbott:
current draft LGTM at a high level, I'll take another pass after testing

flihp:
I've updated the two commits that pull data from the `stage0-handoff` region. They now handle failures to pull data from these regions w/ an `Option` for the two seed components. If these end up being `None` then the error written in the `ringbuf` and that component is left out of the seed. I'm not sure if this is the right balance between flexibility and security but it'll keep the task from crash looping on lpc55s w/ DICE disabled.

andrewjstone:
> I've updated the two commits that pull data from the `stage0-handoff` region. They now handle failures to pull data from these regions w/ an `Option` for the two seed components. If these end up being `None` then the error written in the `ringbuf` and that component is left out of the seed. I'm not sure if this is the right balance between flexibility and security but it'll keep the task from crash looping on lpc55s w/ DICE disabled.

I understand why you did this. However, this strategy makes me somewhat nervous that we'll accidentally squash an error and leave ourselves vulnerable in production. 

Is DICE always going to sometimes be disabled on the lpc55s for dev use? Or is this a temporary state of affairs? If the latter, we should open an issue to revert this code when DICE is no longer ever disabled. If the former, we still need to prevent the crash looping. In that case, can we add an extra check where we return an option that matches on any specific error due to DICE being disabled and then confirms that DICE is actually disabled. If DICE is not disabled then we should panic rather than return `None`.

flihp:
Yes, that's what I meant when I said `I'm not sure if this is the right balance between flexibility and security`. I probably shouldn't have flipped off the `Draft` bit since I'm still trying to come up with reasonable guard rails here. That said we only ship this task on dev boards currently.

> Is DICE always going to sometimes be disabled on the lpc55s for dev use?

I spent some time these last few days trying to come to a decision on this point. My current thinking is that we can't assume that DICE is enabled at all. Most of this is because it can be turned on / off at will which (I think) means we have to make the strongest RNG seed the default and produce errors / panic *unless* the `app.toml` explicitly opts into a weaker RNG seed.

So my current thinking is that I'll introduce a new feature(s) to the RNG task that controls whether or not missing / unavailable components of the seed are hard errors. This will have to deal with conditional availability of the stage0-handoff region in `build.rs` as well as augmenting the `Option` stuff I added at the end of last week to reject `None` values (your feedback focuses mostly on this poitn). I've just starting to prototype this so it's not yet clear but I think this will work best if this feature is enabled by default, requiring selective disabling / explicit opt-in to weaken the seed.

labbott:
> Is DICE always going to sometimes be disabled on the lpc55s for dev use? 

These days it's less likely as everyone is probably working on a sled which has been setup or a rot-carrier which also usually gets embootleby'd . The only people who are likely to be running with DICE off are those of us who are working off of boards that have not been setup or testing something very specific.

flihp:
I've added a feature to gate inclusion of these components in the various RNG seeds. If this feature is enabled and the task can't access one of these memory regions then we record the failure in the ringbuf and panic. This isn't what we want, but it at least "fails closed" as they say. I'd prefer this feature were enabled by default so as to require the `app.toml` to disable it explicitly. This doesn't currently work (unless I screwed something up). Alternatively this flag could have been a negative one (enabling the flag disables the feature) but those get confusing.

I plan to take some time over the next day or so to get a rough idea of the level of effort associated with supporting default features for task `Cargo.toml`s since it would be convenient here. If folks think the current approach is sufficient I don't see a need to hold this PR up, but if there are lingering concerns over possible foot guns I'll be weighing the level of effort of task default features against turning the feature flag introduced here into a negative one.

drv/lpc55-rng/src/config.rs
```diff
@@ -0,0 +1,11 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#[derive(serde::Deserialize, Default, Debug)]
+#[serde(rename_all = "kebab-case")]
+#[cfg(feature = "dice-seed")]
+pub struct DataRegion {
+    pub address: u32,
```
aapoalas, line 9: **suggestion:** Use `usize` for these. For `address` you're doing `as *const X` so the type here doesn't really matter, and for `size` you're doing `as usize` anyway.

Of course, the end result is exactly the same either way.

drv/lpc55-rng/src/main.rs
```diff
@@ -197,14 +224,95 @@ impl NotificationHandler for Lpc55RngServer {
     }
 }
 
+/// Load a type implementing HandoffData (and others) from a config::DataRegion.
+/// Errors will be reported in the ringbuf and will return in None.
```
aapoalas, line 228: **nitpick:** Confused me for a second, I presume you meant to write "return None."

drv/lpc55-rng/src/main.rs
```diff
@@ -197,14 +224,95 @@ impl NotificationHandler for Lpc55RngServer {
...
+>(
+    region: &DataRegion,
+) -> Option<T> {
+    use core::slice;
+
+    // Safety: This memory is setup by code executed before hubris and
+    // exposed using the kernel `extern-regions` mechanism. The safety of
+    // this code is an extension of our trust in the hubris kernel / build.
+    let data = unsafe {
+        slice::from_raw_parts(region.address as *mut u8, region.size as usize)
```
aapoalas, line 241: **question:** Please excuse my dumb question / confirmation, but I presume "this memory" is loaded into actual memory, and is not MMIO or something like that? In which case it might not be safe to turn it into a slice for reading.

drv/lpc55-rng/src/main.rs
```diff
@@ -197,14 +224,95 @@ impl NotificationHandler for Lpc55RngServer {
...
+                    panic!();
+                },
+            }
+        } else {
+            None
+        }
+    }
+}
+
+/// Get the platform identifier / barcode string from the platform identity
```
aapoalas, line 277: **praise:** Great comments, really nice to read through this.

aapoalas (commented):
Comments from the peanut gallery. Really pleasant code to read <3

flihp:
This has been successfully tested on a gimlet, psc & sidecar. To do so required adding the `rng_driver` task to an `oxide-rot-1` image. The `app-dev.toml` / `oxide-rot-1-dev` image is lowest risk.

The remaining work to unblock this PR is in #1830

drv/lpc55-rng/src/main.rs
```diff
@@ -197,14 +225,96 @@ impl NotificationHandler for Lpc55RngServer {
...
+
+    // Safety: This memory is setup by code executed before hubris and
+    // exposed using the kernel `extern-regions` mechanism. The safety of
+    // this code is an extension of our trust in the hubris pre-main, kernel,
+    // and build process.
+    let data = unsafe {
+        slice::from_raw_parts(region.address as *mut u8, region.size)
+    };
+
+    // this can be replaced w/ .ok() if we get rid of the ringbuf entry
```
aapoalas, line 246: **suggestion:** `inspect_err` could be used to do the ringbuf, chained to `ok()`

aapoalas (commented):
More peanuts.

drv/lpc55-rng/src/main.rs
```diff
@@ -9,119 +9,124 @@
...
-    }
+#[derive(Copy, Clone, PartialEq)]
+enum Trace {
+    #[cfg(feature = "dice-seed")]
+    NoDiceSeed,
+    #[cfg(feature = "dice-seed")]
+    HandoffError(HandoffDataLoadError),
+    #[cfg(feature = "dice-seed")]
+    NoSeedPersonalization,
+    None,
```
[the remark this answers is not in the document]
flihp: AFAIK the position in the enum of the `None` variant shouldn't matter. But no reason it can't be at position 0 and I think it makes better sense as well so I've made this change.

`ringbuf` doesn't care that the enum changes and so long as `humility` is able to get symbols provided w/ the archive it should be fine. Removing the feature cfg stuff from these enum variants would probably only cost a few extra bytes and we can ignore the compilers warnings easy enough if we don't like mutating the type.

app/lpc55xpresso/app.toml
```diff
@@ -113,11 +113,11 @@ task-slots = ["gpio_driver", "syscon_driver"]
 [tasks.rng_driver]
 name = "drv-lpc55-rng"
 priority = 3
-max-sizes = {flash = 16384, ram = 4096}
 uses = ["rng", "pmc"]
 start = true
-stacksize = 2200
+stacksize = 4200
```
cbiffle, line 118: So I think this'd make it the highest stack usage in the system, above control-plane-agent. Do we have a sense for what caused this? (Asking because I've been winding up playing stack size whack-a-mole lately.)

lib/lpc55-rng/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+pub struct Lpc55Rng {
+    pub pmc: &'static pmc::RegisterBlock,
+    pub rng: &'static rng::RegisterBlock,
+}
+
+impl Lpc55Rng {
+    /// Create a new Lpc55Rng instance after powering on, enabling the clocks
+    /// and reseting the underlying HRNG.
+    pub fn new(syscon: &Syscon) -> Self {
+        let pmc = unsafe { &*PMC::ptr() };
```
cbiffle, line 25: I'd suggest having the PMC passed in (like the Syscon is) to make this dependency more obvious (and lose the unsafe!).

lib/lpc55-rng/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+        pmc.pdruncfg0.modify(|_, w| w.pden_rng().poweredon());
+
+        syscon.enable_clock(Peripheral::Rng);
+        syscon.enter_reset(Peripheral::Rng);
+        syscon.leave_reset(Peripheral::Rng);
+
+        Lpc55Rng {
+            pmc,
+            rng: unsafe { &*RNG::ptr() },
```
cbiffle, line 34: (same as on line 25)

lib/lpc55-rng/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+        self.try_fill_bytes(bytes).expect("fill_bytes")
+    }
+
+    /// Fill the provided buffer with output from the HRNG. If the HRNG
+    /// can't service the request an error is returned.
+    fn try_fill_bytes(&mut self, dst: &mut [u8]) -> Result<(), Error> {
+        let mut filled = 0;
+        while filled < dst.len() {
+            if self.pmc.pdruncfg0.read().pden_rng().bits() {
+                return Err(RngError::PoweredOff.into());
```
cbiffle, line 61: So, this looks like the only way this operation can fail. Given that the RNG is powered on in `new`, finding it powered off would be a surprising program bug. Perhaps this should be a panic? Otherwise I expect all callers to wind up unwrapping this at some point... could be wrong though.

drv/lpc55-rng/src/main.rs
```diff
@@ -9,119 +9,124 @@
...
+            usize::MAX
+        } else {
+            threshold
+        };
 
-        let threshold = if threshold == 0 { MAX } else { threshold };
+        let mut mixer = H::default();
+        if let Some(seed) = seed {
+            // mix platform unique seed derived by measured boot
+            Digest::update(&mut mixer, seed.as_bytes());
```
cbiffle, line 98: Curious about the explicit qualification on `Digest::` -- does `mixer.update(seed.as_bytes())` not work?

drv/lpc55-rng/src/main.rs
```diff
@@ -197,14 +225,91 @@ impl NotificationHandler for Lpc55RngServer {
...
+/// Errors will be reported in the ringbuf and will return None.
+#[cfg(feature = "dice-seed")]
+fn load_data_from_region<
+    T: for<'a> Deserialize<'a> + HandoffData + SerializedSize,
+>(
+    region: &DataRegion,
+) -> Option<T> {
+    use core::slice;
+
+    // Safety: This memory is setup by code executed before hubris and
```
cbiffle, line 238: This safety comment isn't actually correct, it assumes that `region` is valid (but the caller can pass any region it likes -- the fields are public). This function should probably be unsafe in its current form, since it's basically equivalent to `ptr::read`.

You could make it safe if the caller couldn't pass any arbitrary region -- maybe hardcode which region, or identify it with an enum, or....

app/lpc55xpresso/app.toml
cbiffle: Having read more of the code, I can answer this question at least partially: the `Server` type used in the RNG server is large, and it's being created on the stack and passed around.

It's large _primarily_ due to the `Sha3` embedded within it, which I recently noticed in an unrelated case was about 2 kiB.

The easiest way to fix this would be to allocate a static Sha3 instance using `ClaimOnceCell` (from `lib/static-cell`) and then pass it into the server constructor to take ownership of (via a `&'static mut`).

drv/lpc55-rng/src/main.rs
```diff
@@ -9,119 +9,124 @@
...
+            Digest::update(&mut mixer, pid);
+        }
+
+        // w/ 32 bytes from HRNG
+        let mut buf = Zeroizing::new(T::Seed::default());
+        reseeder.try_fill_bytes(buf.as_mut())?;
+        Digest::update(&mut mixer, buf.as_ref());
+
+        // create initial instance of the SeedableRng from the seed
+        let inner = T::from_seed(mixer.finalize_fixed_reset().into());
```
cbiffle, line 112: FYI, you might see lighter stack usage here by using one of the `finalize_into` operations, instead of returning the array. It might also not change anything. Hard to tell.

drv/lpc55-rng/src/main.rs
flihp: The type we're using for the digest implements multiple `update` functions. One's for the `Digest` trait like we're expecting here but the other is for the `Update` trait. So I think this is just a disambiguation.

lib/lpc55-rng/src/lib.rs
flihp: I went back and forth on whether to panic or return the error here a few times. Given the change requested in https://github.com/oxidecomputer/hubris/pull/1820#discussion_r1772082662 `new` will now take ownership of instances of the PMC & RNG types which makes it clear that we don't intend for them to change behind our backs. Returning the error was part of an attempt to be resilient to this, but it's so far from the common case that it's not worth making things less ergonomic.

lib/lpc55-rng/src/lib.rs
flihp: Using the `ptr()` function on PMC & RNG definitely qualifies as holding the lpc55-pac crate wrong. Should be addressed by [bce3889](https://github.com/oxidecomputer/hubris/pull/1820/commits/bce388971fce1d15496ef0bc7174bc4f4033a5e5).

lib/lpc55-rng/src/lib.rs
flihp: Addressed in [bce3889](https://github.com/oxidecomputer/hubris/pull/1820/commits/bce388971fce1d15496ef0bc7174bc4f4033a5e5) as well.

lib/lpc55-rng/src/lib.rs
flihp: addressed by: [c29dffa](https://github.com/oxidecomputer/hubris/pull/1820/commits/c29dffad19ac1a26bcfe4a058c66efe5e016bc83)

drv/lpc55-rng/src/main.rs
flihp: Yeah that's bad. Thanks for the recommendations. I went w/ hiding the details of the memory regions behind an enum. Seems to do what you suggest: [cf2704f](https://github.com/oxidecomputer/hubris/pull/1820/commits/cf2704fae6fab72f6178fb6ca768fc171c7820d0)

flihp: I didn't realize how badly this change affected stack usage ... working on that now.

app/lpc55xpresso/app.toml
flihp: Had to use `mutable-statics` because `Sha3` doesn't have a const constructor (per discussion in chat): [b43295c](https://github.com/oxidecomputer/hubris/pull/1820/commits/b43295cc9a8bf2f7668a17916d6c86d7d40becfc)

drv/lpc55-rng/src/main.rs
flihp: The majority of the stack bloat from the change in [cf2704f](https://github.com/oxidecomputer/hubris/pull/1820/commits/cf2704fae6fab72f6178fb6ca768fc171c7820d0) is mitigated in [bf092d3](https://github.com/oxidecomputer/hubris/pull/1820/commits/bf092d3ff4dbdfa4cd6a1f12b53eb74d2c9dc0b5).

drv/lpc55-rng/src/main.rs
```diff
@@ -283,13 +284,22 @@ fn main() -> ! {
         )
     };
 
+    let mixer = mutable_statics! {
+        static mut MIXER: [Sha3_256; 1] = [Sha3_256::new; _];
```
cbiffle, line 288: This is the change I had in mind -- I'm surprised it didn't reduce stack usage more! Oh well. It's still an improvement.

PR #1825, opened 2024-07-15
Bump sprot size for -dev variant

labbott:


flihp:
LGTM. Out of curiosity I checked the sizes for `sprot` on this image locally:

```
$ cargo xtask sizes app/oxide-rot-1/app-dev.toml
...
flash:
      ADDRESS  |    PROGRAM    |  USED |  SIZE | LIMIT
...
    0x0003c880 | sprot         | 34236 | 34240 | 54016
...
PROGRAM        REGION   USED   SIZE   LIMIT
...
sprot          flash    34236  34240  54016
```

If I'm reading this right it looks like there's a good amount of flash left between the `used` and `limit`. This may be space that was recently freed up when sprockets was removed?

labbott:
> 
> If I'm reading this right it looks like there's a good amount of flash left between the `used` and `limit`. This may be space that was recently freed up when sprockets was removed?

You are absolutely correct. I did this change then the sprockets change but that got merged first. This should no longer be necessary. I'm going to close it for now.

PR #1827, opened 2024-07-16
Add updated bootleby

labbott:
This brings bootleby up to v1.3.0


flihp (commented):
What's the relationship between these zip files and the existing bootleby bin files next to them? Like https://github.com/oxidecomputer/hubris/blob/99b7a384a7616831e1030c744603a008956bf9bd/app/oxide-rot-1/bootleby-oxide-rot-1-1.0.0.bartholomew.bin

flihp:
The merge commit here should probably be rebased away too?

labbott:
> What's the relationship between these zip files and the existing bootleby bin files next to them? Like https://github.com/oxidecomputer/hubris/blob/99b7a384a7616831e1030c744603a008956bf9bd/app/oxide-rot-1/bootleby-oxide-rot-1-1.0.0.bartholomew.bin

That's a single binary suitable for writing directly to memory (e.g. `lpc55_flash write-memory 0x0`). I don't think we've kept that up to date. I can just remove it since if you really want the single binary you can grab the bootleby bundle and unzip.

flihp:
Thanks for pulling those old bootleby images. Keeping those around is just inviting confusion.

PR #1830, opened 2024-07-16
xtask: Allow tasks & the kernel to specify default features.

### LINKED ISSUE #1829: support default features at the task / kernel level
This is mostly in the interest of keeping myself from adding features that aren't additive and is motivated by #1820. That PR adds some code to improve our seeding strategy. This includes passing a seed derived from the DICE measured boot through the `stage0-handoff` memory region to the RNG task. We want this feature enabled by default and to require `app.toml` files to *explicitly* disable this feature given its relevance to security & cryptography stuff. It's possible to accomplish this w/ a feature that *removes* the functionality but `cargo` discourages this. #1820 seems like enough of a reason to see if we can do this w/o too much work.

flihp:
tasks & kernels that want to disable default features can do so using the `no-default-features` boolean in the task / kernel config sections of the relevant `app.toml` files.

this resolves #1829

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1930,6 +1930,7 @@ name = "drv-stm32h7-fmc-demo-server"
 version = "0.1.0"
 dependencies = [
  "build-util",
+ "cfg-if",
  "cortex-m",
  "counters",
  "drv-stm32xx-sys-api",
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -222,14 +222,14 @@ impl Config {
         &self,
         verbose: bool,
         crate_name: &str,
+        no_default_features: bool,
         features: &[String],
         sysroot: Option<&'a Path>,
     ) -> BuildConfig<'a> {
-        let mut args = vec![
-            "--no-default-features".to_string(),
-            "--target".to_string(),
-            self.target.to_string(),
-        ];
+        let mut args = vec!["--target".to_string(), self.target.to_string()];
+        if no_default_features {
+            args.push("--no-default-features".to_string());
+        }
         if verbose {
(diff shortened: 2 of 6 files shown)


flihp:
I'm sure there was a reason for disabling this so this may be ill advised but I want default features for the rng task in #1820. I've built a few images for the RoT and basic tire kicking was encouraging. The output from the build (task & kernel elfs) are identical *except* for the `idle` task. I haven't dug deep into why but if anyone is interested in this feature or thinks this is a bad idea I'd appreciate the feedback good or bad.

flihp:
As a way to convince myself that enabling `default-features` doesn't have any unintended side effects the first commit in this PR removes all use of `default-features` from existing tasks. These were no-ops since we were passing `--no-default-features` to `rustc` anyway. The second commit here enables the use of `default-features` in tasks and because we removed all use of these in the previous commit the build output is identical both before and after [dc75180](https://github.com/oxidecomputer/hubris/pull/1830/commits/dc751802852885401e6f81ea1a4ce650fbe0a125) is applied. I have tested and confirmed that this is the case for every app.toml in the source tree.

cbiffle:
Historical context:

So, in the early days, we used outside-of-xtask builds in two situations:
1. Doing single-task iteration (in the days before `xtask build`) using `cargo check`
2. In editors using RLS, back when RLS was a thing

We used default features to fake a subset of the build system features to achieve this, which is part of what you're finding in the old default features lists.

Nowadays we've got the better rust-analyzer - build-system integration, and it's probably safe to stop doing this hack. So I'm generally positive about this change in concept, though I haven't finished reading the implementation yet.

cbiffle (approved):

PR #1832, opened 2024-07-23
Add tech port unlocking

mkeeter:
This PR implements tech port unlocking, as described in [RFD 492](https://rfd.shared.oxide.computer/rfd/492).  It corresponds to https://github.com/oxidecomputer/management-gateway-service/pull/254 and should be merged after that goes in.

There's a lot going on here, but it's tricky to split into smaller PRs, so I'll describe the different moving pieces

### Configuration
VLANs are no longer contiguous; instead, the configuration in the manifest specifies them by VID.  Each VLAN is also labelled as trusted or untrusted, which changes the behavior of the `net` task.  The mapping from VLAN to port has become many-to-one, because Sidecar now has 3x VLANs on `SpPort::One`.

### Runtime behavior
Packets to/from untrusted VLANs are dropped in the `net` task, unless the destination task has `accept-untrusted = true` in its socket configuration.  This may be disabled for a limited time with the new `Net.trust_vlan` Idol API.

## Monorail and KSZ8463 configuration changes
The VSC7448 (Monorail) and KSZ8463 switch have a new baseline configuration: packets between VSC7448 port 48 (to the Sidecar-local SP) and the KSZ8463 are now VLAN tagged per [this scheme](https://rfd.shared.oxide.computer/rfd/492#_fully_tagged_packets_on_port_48).

In addition, the VSC7448 is now configured to disable access to SPs (other than the Sidecar-local SP) from the tech port.  This is implemented in `configure_vlan_sidecar_locked`.  The Monorail API now has two new functions (`Monorail.unlock_vlans` and `Monorail.lock_vlans`) to change this behavior.

### General
Across all boards, `SpImpl` function signatures and implementations are changed to use the new `Sender` type where appropriate.

### Sidecar
On the Sidecar, we implement tech port unlocking based on SSH signatures.  For now the "private" key is well-known (and included in this commit); this is to establish workflows, and does not improve the system's security.

The build script is modified to generate `trusted_keys.rs` per the manifest, and we implement `MonorailComponentAction::*` in `mgs_sidecar.rs` to generate and verify challenges.

In addition, we add a (pre-existing) driver task for the STM32's RNG to generate a random nonce.


labbott:
Can you add flash/ram diff for reference?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +236,235 @@ impl MgsHandler {
...
+            (
+                UnlockChallenge::EcdsaSha2Nistp256(data),
+                UnlockResponse::EcdsaSha2Nistp256 { key, signature },
+            ) => verify_signature(&data, &key, &signature),
+        }?;
+        ringbuf_entry!(Trace::UnlockAuthSucceeded);
+
+        // Pick how long we'll trust things for
+        let now = sys_get_timer().now;
+        let trust_until = now + u64::from(time_sec) * 1000;
```
labbott, line 294: I did a thought experiment about setting the unlock time to 0 and I think everything should get handled okay?

task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -472,6 +500,15 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
         }
 
+        if let VLanMode::UnlockedUntil(t) = self.vlan_mode {
+            let now = userlib::sys_get_timer().now;
+            if now >= t {
+                ringbuf_entry!(Trace::AutomaticLock);
+                // Discard the error, because it's already logged in lock_vlans
+                let _ = self.lock_vlans();
+            }
+        }
```
labbott, line 510: I'm not familiar with the full monorail setup, when is `wake` getting called?

mkeeter:
> Can you add flash/ram diff for reference?

Before:
```
Used:
  flash:   0xa8100 (65%)
  ram:     0x4f800 (62%)
  sram1:   0x4000 (12%)
  sram2:   extern region (jefe, dump_agent)
  sram3:   extern region (jefe, dump_agent)
  sram4:   extern region (jefe, dump_agent)
  bank2:   extern region (update_server)
```

After:
```
Used:
  flash:   0xb8100 (71%)
  ram:     0x67800 (80%)
  sram1:   0x4000 (12%)
  sram2:   extern region (jefe, dump_agent)
  sram3:   extern region (jefe, dump_agent)
  sram4:   extern region (jefe, dump_agent)
  bank2:   extern region (update_server)
```

Here's the per-task size diff:
```diff
98,99c92,93
<                      ram     5384    6256    (fixed)
< jefe                 flash   8620    8704    16384
---
>                      ram     5552    6256    (fixed)
> jefe                 flash   8608    8704    16384
102a97,98
> rng_driver           flash   6244    6272    8192
>                      ram     452     512     512
107,108c103,104
< net                  flash   79048   81920   131072
<                      ram     44256   65536   65536
---
> net                  flash   83500   98304   131072
>                      ram     79104   131072  131072
110,112c106,108
< control_plane_agent  flash   105992  106496  131072
<                      ram     28476   32768   32768
< sprot                flash   45328   45344   65536
---
> control_plane_agent  flash   136520  136576  140672
>                      ram     35352   65536   65536
> sprot                flash   45336   45344   65536
114,119c110,115
< udpecho              flash   8952    8960    16384
<                      ram     4428    8192    8192
< udpbroadcast         flash   8460    8480    16384
<                      ram     4488    8192    8192
< monorail             flash   80148   80384   262144
<                      ram     9108    16384   16384
---
> udpecho              flash   9008    9024    16384
>                      ram     4460    8192    8192
> udpbroadcast         flash   8508    8512    16384
>                      ram     4520    8192    8192
> monorail             flash   81728   81920   262144
>                      ram     9236    16384   16384
130,131c126,127
< transceivers         flash   39236   39264   65536
<                      ram     11044   16384   16384
---
> transceivers         flash   39352   39360   65536
>                      ram     11076   16384   16384
140c136
< validate             flash   8792    8800    16384
---
> validate             flash   8796    8800    16384
146,147c142,143
< dump_agent           flash   18204   18432   32768
<                      ram     12916   16384   16384
---
> dump_agent           flash   18248   18432   32768
>                      ram     12948   16384   16384
```

task/control-plane-agent/Cargo.toml
```diff
@@ -14,6 +14,9 @@ ssmarshal.workspace = true
 static_assertions.workspace = true
 zerocopy.workspace = true
 
+p256 = { workspace = true, optional = true }
+sha2 = { workspace = true, optional = true }
```
lzrd, line 18: We have a mix of SHA2 and SHA3 in our implementations with some preference for SHA3.
If this is an arbitrary choice, then preference would be for SHA3.

task/control-plane-agent/src/mgs_gimlet.rs
```diff
@@ -462,24 +460,28 @@ impl SpHandler for MgsHandler {
     type BulkIgnitionStateIter = core::iter::Empty<IgnitionState>;
     type BulkIgnitionLinkEventsIter = core::iter::Empty<ignition::LinkEvents>;
 
+    fn is_message_trusted(
+        &mut self,
+        _kind: &MgsRequest,
+        _sender: Sender,
+    ) -> Result<(), SpError> {
+        // Gimlets are okay with everyone talking to them, since they're behind
+        // the management network.
```
lzrd, line 469: Someday we will revisit the assumption that the management network is trusted.

build/net/src/lib.rs
```diff
@@ -27,7 +27,8 @@ pub struct NetConfig {
     /// VLAN configuration, or None. This is checked against enabled features
     /// during the `net` build, so it must be present iff the `vlan` feature
     /// is turned on.
```
jgallagher, line 29: nit - the doc comment still refers to this as optional

task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -472,6 +500,15 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
             }
         }
 
+        if let VLanMode::UnlockedUntil(t) = self.vlan_mode {
+            let now = userlib::sys_get_timer().now;
```
lzrd, line 504: If the state (t) is lost due to task restart, are the VLANs locked on task restart?

task/control-plane-agent/src/main.rs
```diff
@@ -585,26 +589,25 @@ impl NetHandler {
...
 #[allow(dead_code)]
 fn vlan_id_from_sp_port(port: SpPort) -> u16 {
-    use task_net_api::VLAN_RANGE;
-    assert_eq!(VLAN_RANGE.len(), 2);
-
-    match port {
-        SpPort::One => VLAN_RANGE.start,
-        SpPort::Two => VLAN_RANGE.start + 1,
-    }
+    // TODO should we remove this and just use `Sender` everywhere?
```
jgallagher, line 601: Maaaybe? I think this is primarily used in the installinator / host phase 2 via MGS step, where the SP is initiating communication by sending to a multicast address that MGS is bound to. I wouldn't object to removing this but it might be kinda fiddly.

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -90,13 +142,31 @@ impl MgsHandler {
...
             Some(sys_get_timer().now + 1)
         } else {
-            None
+            self.locked
+                .iter()
+                .flat_map(|k| match k {
+                    LockState::UnlockedUntil(t) => Some(*t),
+                    _ => None,
+                })
+                .min()
```
jgallagher, line 151: I did a double-take here because `None < Some(_)`, but just checking I got it: this is `flat_map`, not `map`, so all the `None`s are discarded and we're only taking the min of the `Some(*t)`s?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +236,235 @@ impl MgsHandler {
...
+    }
+
+    fn lock(&mut self, vid: u16) -> Result<(), GwMonorailError> {
+        ringbuf_entry!(Trace::Locking(vid));
+
+        let net = task_net_api::Net::from(crate::NET.get_task_id());
+        if let Err(e) = net.distrust_vlan(vid) {
+            ringbuf_entry!(Trace::DistrustVLanFailed(e));
+            return Err(GwMonorailError::UnlockFailed);
+        } else if let Err(e) = self.monorail.lock_vlans() {
```
jgallagher, line 323: I stumbled on the `if let Err(_) = ... else if let Err(e) = ...`; these might be clearer as `.map_err(_)?`s? It wasn't immediately obvious to me that we run both the if and the else on the happy path.

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +236,235 @@ impl MgsHandler {
...
+            ringbuf_entry!(Trace::MonorailUnlockFailed(e));
+            return Err(GwMonorailError::UnlockFailed);
+        }
+
+        let i = VLANS.iter().position(|v| v.vid == vid).unwrap();
+        self.locked[i] = LockState::Locked;
```
jgallagher, line 329: If `net.distrust_vlan(_)` succeeds but `monorail.lock_vlans()` fails, should we update our state?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -341,16 +604,55 @@ impl SpHandler for MgsHandler {
                     }
                 }
                 .unwrap();
-                Ok(())
+                Ok(ComponentActionResponse::Ack)
+            }
+            (SpComponent::MONORAIL, ComponentAction::Monorail(action)) => {
+                match action {
+                    MonorailComponentAction::RequestChallenge => {
+                        // TODO is this sufficient?
```
jgallagher, line 612: Maybe a question for @flihp ?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -341,16 +604,55 @@ impl SpHandler for MgsHandler {
...
+                        let rng = drv_rng_api::Rng::from(RNG.get_task_id());
+                        let mut buf = [0u8; 32];
+                        if let Err(_e) = rng.fill(&mut buf) {
+                            // TODO log this?
```
jgallagher, line 616: As a `Trace` I assume? I'd vote yes

task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -548,6 +585,31 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
+        unlock_until: u64,
+    ) -> Result<(), MonorailError> {
+        ringbuf_entry!(Trace::UnlockUntil(unlock_until));
+        self.vsc7448
+            .configure_vlan_sidecar_unlocked()
+            .map_err(|e| {
+                ringbuf_entry!(Trace::UnlockError(e));
+                MonorailError::from(e)
+            })?;
+        self.vlan_mode = VLanMode::UnlockedUntil(unlock_until);
```
jgallagher, line 600: Nitpicky question, but - is it possible for `configure_vlan_sidecar_unlocked()` to partially succeed and leave us in some bad state while `self.vlan_mode` still thinks we're in `VLanMode::Locked`?

task/control-plane-agent/Cargo.toml
mkeeter: The current support keys are using `ecdsa-sha2-nistp256` certificates, so I think we're stuck with SHA2 for now.

jgallagher: FWIW I think this isn't just the support keys; the ssh-agent (our mechanism for talking to support yubikeys) protocol only lists SHA2-256 and SHA2-512 as options (https://datatracker.ietf.org/doc/html/draft-miller-ssh-agent-04#section-7.3).

task/monorail-server/src/bsp/sidecar_bcd.rs
mkeeter: Yes, when the `monorail` task initially starts up, the switch is configured to be locked.

Note that if just one of the lock-sensitive tasks (`control-plane-agent`, `monorail`, `net`) reboots, it may be locked while the rest of the system is unlocked.  I think this is fine, since it's failing into the more secure state.

jgallagher: > Note that if just one of the lock-sensitive tasks (`control-plane-agent`, `monorail`, `net`) reboots, it may be locked while the rest of the system is unlocked. I think this is fine, since it's failing into the more secure state.

I think this is minor, but could we land in a "system is unlocked but we can't lock it" or vice versa because of state disagreements? E.g., if CPA reboots and now thinks we're locked, will it refuse requests to lock even though it should pass them on?

Since all of this is time based it would clear itself up after at most an hour, and for that hour I agree we'd have failed in the correct direction.

mkeeter: If CPA reboots and thinks we're locked, it will ignore lock calls (because they're not on the list of messages accepted while locked), but _will_ accept calls to unlock (which will just unlock itself and update the unlock timer for other tasks).

My intent is for an unlock command to always unlock the system, even if some parts were already unlocked; let me know if you spot any places where that seems incorrect.

build/net/src/lib.rs
mkeeter: Updated to "must be non-empty" in 23b0a914ef2a

task/control-plane-agent/src/main.rs
mkeeter: Yeah, it ended up being fiddly because we swap between `SpPort::One/Two`, which is harder to do with `Sender` types; I've removed that TODO and inlined this whole function into `packet_to_mgs` (because that's the only caller at this point).

See 0c36fb5bd538

task/monorail-server/src/bsp/sidecar_bcd.rs
mkeeter: Every 500 ms, per [`WAKE_INTERVAL`](https://github.com/oxidecomputer/hubris/blob/24abf7292491069dc87e1e5a2930ab5822d26abb/task/monorail-server/src/bsp/sidecar_bcd.rs#L19-L20)

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: Added a comment in f6b3a5b8a17a to explain what it's doing 😄

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: Yup, done in 5d11fa049d64

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: I like `map_err(..)`; done in 5d11fa049d64

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +238,236 @@ impl MgsHandler {
...
+    fn unlock(
+        &mut self,
+        vid: u16,
+        challenge: UnlockChallenge,
+        response: UnlockResponse,
+        time_sec: u32,
+    ) -> Result<(), GwMonorailError> {
+        ringbuf_entry!(Trace::UnlockRequested(vid));
+
+        if time_sec > MAX_UNLOCK_TIME_SECS {
```
lzrd, line 252: We may want the signed hash to also cover the unlock duration.
e.g. when more sophisticated signing policies are implemented then a customer may be fine with unlock for 10 minutes but not MAX_UNLOCK_TIME_SECS.

mkeeter:
Closing in favor of #1835

PR #1834, opened 2024-07-29
Implement RotRequest for host-sp-comms

labbott:


task/host-sp-comms/src/main.rs
```diff
@@ -840,8 +847,36 @@ impl ServerImpl {
...
             }
             HostToSp::RotRequest => {
-                // TODO forward request to RoT
-                Some(SpToHost::RotResponse)
+                match attest_data::messages::parse_message(data) {
+                    Ok((command, data)) => {
+                        let n = usize::min(MAX_DATA_LEN, data.len());
+
+                        let mut data_buf: [u8; MAX_DATA_LEN] =
+                            [0; MAX_DATA_LEN];
```
jgallagher, line 855: Is `MAX_DATA_LEN` small enough that putting it on the stack is okay? Or should we statically allocate space up front? (Worried about a stack overflow that we only hit with certain messages)

task/host-sp-comms/src/main.rs
```diff
@@ -924,6 +959,109 @@ impl ServerImpl {
...
+                        let cert_len = self
+                            .sprot
+                            .cert_len(index)
+                            .map_err(AttestDataSprotError::from)?;
+                        for o in (0..cert_len).step_by(512) {
+                            let idx: usize = buf_idx + o as usize;
+                            let end = usize::min(512, (cert_len - o) as usize);
+
+                            self.sprot
+                                .cert(index, o, &mut buf[idx..idx + end])
```
jgallagher, line 987: This will panic if `buf` is too small for all the certs, right? Should we guard against that and return an error instead? (Same comment in the closures below)

task/host-sp-comms/src/main.rs
labbott: `MAX_DATA_LEN` is currently the size of a nonce which is 32 bytes (https://github.com/oxidecomputer/dice-util/pull/220/files#diff-d7aca782c921cfedc9fafda8a7258ce09c947b295ed7457bfc0956e0bbaab82dR12) . I'm expecting this to grow to 64 bytes for signing a hash but not more than that.

task/host-sp-comms/src/main.rs
labbott: That's unlikely to happen given the fixed size of the certs and the fixed size of the buffer for transfer based on the protocol. It might still be good to check because task crashes are not nice to debug.

task/host-sp-comms/src/main.rs
```diff
@@ -843,8 +851,36 @@ impl ServerImpl {
...
+                    Ok((command, data)) => {
+                        let n = usize::min(MAX_DATA_LEN, data.len());
+
+                        let mut data_buf: [u8; MAX_DATA_LEN] =
+                            [0; MAX_DATA_LEN];
+                        if n > 0 {
+                            data_buf[..n].copy_from_slice(&data[..n]);
+                        }
+
+                        self.handle_sprot(header.sequence, command, &data_buf);
```
jgallagher, line 864: Just making sure I understand - we pass `&data_buf` here instead of `&data[..n]` because attestation expects the data to always be `MAX_DATA_LEN` and zero padded if needed?

task/host-sp-comms/src/main.rs
```diff
@@ -927,6 +963,132 @@ impl ServerImpl {
...
+                            .sprot
+                            .cert_len(index)
+                            .map_err(AttestDataSprotError::from)?;
+                        for o in (0..cert_len).step_by(SPROT_TRANSFER_SIZE) {
+                            let idx: usize = buf_idx + o as usize;
+                            let end = usize::min(
+                                SPROT_TRANSFER_SIZE,
+                                (cert_len - o) as usize,
+                            );
+                            if idx >= buf.len() || end >= buf.len() {
```
jgallagher, line 995: Should this be 
```suggestion
                            if idx + end >= buf.len() {
```

instead (based on the `&mut buf[idx..idx+end]` below)?

task/host-sp-comms/src/main.rs
```diff
@@ -927,6 +963,132 @@ impl ServerImpl {
...
+                        .sprot
+                        .log_len()
+                        .map_err(AttestDataSprotError::from)?;
+                    for o in (0..measurement_len).step_by(SPROT_TRANSFER_SIZE) {
+                        let idx: usize = o as usize;
+                        let end = usize::min(
+                            SPROT_TRANSFER_SIZE,
+                            (measurement_len - o) as usize,
+                        );
+                        if idx >= buf.len() || end >= buf.len() {
```
jgallagher, line 1035: Same question about `idx + end`

jgallagher (approved):

task/host-sp-comms/src/main.rs
labbott: Yeah good catch

PR #1835, opened 2024-07-29
Tech port unlocking (simple version)

mkeeter:
This is a simplified version of #1832 that removes the SSH signature stuff.  Quoth the previous PR:

----------------------------------------------------------------

This PR implements tech port unlocking, as described in [RFD 492](https://rfd.shared.oxide.computer/rfd/492).  It corresponds to https://github.com/oxidecomputer/management-gateway-service/pull/259 and should be merged after that goes in.

There's a lot going on here, but it's tricky to split into smaller PRs, so I'll describe the different moving pieces

### Configuration
VLANs are no longer contiguous; instead, the configuration in the manifest specifies them by VID.  Each VLAN is also labelled as trusted or untrusted, which changes the behavior of the `net` task.  The mapping from VLAN to port has become many-to-one, because Sidecar now has 3x VLANs on `SpPort::One`.

### Runtime behavior
Packets to/from untrusted VLANs are dropped in the `net` task, unless the destination task has `accept-untrusted = true` in its socket configuration.  This may be disabled for a limited time with the new `Net.trust_vlan` Idol API.

## Monorail and KSZ8463 configuration changes
The VSC7448 (Monorail) and KSZ8463 switch have a new baseline configuration: packets between VSC7448 port 48 (to the Sidecar-local SP) and the KSZ8463 are now VLAN tagged per [this scheme](https://rfd.shared.oxide.computer/rfd/492#_fully_tagged_packets_on_port_48).

In addition, the VSC7448 is now configured to disable access to SPs (other than the Sidecar-local SP) from the tech port.  This is implemented in `configure_vlan_sidecar_locked`.  The Monorail API now has two new functions (`Monorail.unlock_vlans` and `Monorail.lock_vlans`) to change this behavior.

### General
Across all boards, `SpImpl` function signatures and implementations are changed to use the new `Sender` type where appropriate.

### Sidecar
On the Sidecar, we implement tech port unlocking based on a `Trivial` challenge, which calls into `monorail` and `net` to reconfigure them.


task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -548,6 +585,31 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
+        unlock_until: u64,
+    ) -> Result<(), MonorailError> {
+        ringbuf_entry!(Trace::UnlockUntil(unlock_until));
+        self.vsc7448
+            .configure_vlan_sidecar_unlocked()
+            .map_err(|e| {
+                ringbuf_entry!(Trace::UnlockError(e));
+                MonorailError::from(e)
+            })?;
+        self.vlan_mode = VLanMode::UnlockedUntil(unlock_until);
```
mkeeter, line 600: @jgallagher writes

> Nitpicky question, but - is it possible for configure_vlan_sidecar_unlocked() to partially succeed and leave us in some bad state while self.vlan_mode still thinks we're in VLanMode::Locked?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +237,197 @@ impl MgsHandler {
...
+            ringbuf_entry!(Trace::DistrustVLanFailed(e));
+            GwMonorailError::UnlockFailed
+        })?;
+        self.monorail.lock_vlans().map_err(|e| {
+            ringbuf_entry!(Trace::MonorailUnlockFailed(e));
+            GwMonorailError::UnlockFailed
+        })?;
+
+        let i = VLANS.iter().position(|v| v.vid == vid).unwrap();
+        self.locked[i] = LockState::Locked;
```
mkeeter, line 328: @jgallagher writes

> If `net.distrust_vlan(_)` succeeds but `monorail.lock_vlans()` fails, should we update our state?

labbott:
For any other reviewers here is the pertinent diff

```diff
diff --git a/task/control-plane-agent/build.rs b/task/control-plane-agent/build.rs
index d64ffd2b..06368f09 100644
--- a/task/control-plane-agent/build.rs
+++ b/task/control-plane-agent/build.rs
@@ -2,6 +2,17 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use anyhow::Context;
+use serde::Deserialize;
+use std::io::Write;
+use std::path::PathBuf;
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct Config {
+    trusted_keys: Option<Vec<PathBuf>>,
+}
+
 fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     build_util::build_notifications()?;
     idol::Generator::new()
@@ -14,5 +25,31 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
             idol::server::ServerStyle::InOrder,
         )?;
 
+    if let Some(tk) = build_util::task_maybe_config::<Config>()
+        .context("could not parse config.control_plane_agent")?
+        .and_then(|cfg| cfg.trusted_keys)
+    {
+        if tk.is_empty() {
+            panic!("cannot provide empty set of trusted keys");
+        }
+        let out_dir = build_util::out_dir();
+        let dest_path = out_dir.join("trusted_keys.rs");
+        let mut out = std::fs::File::create(&dest_path).with_context(|| {
+            format!("failed to create file '{}'", dest_path.display())
+        })?;
+        writeln!(&mut out, "const TRUSTED_KEYS: [[u8; 65]; {}] = [", tk.len())?;
+        for k in tk {
+            let key = ssh_key::PublicKey::read_openssh_file(&k)
+                .with_context(|| format!("failed to read public key: {k:?}"))?;
+            let pub_bytes = key
+                .key_data()
+                .ecdsa()
+                .expect("must be ECDSA key")
+                .as_sec1_bytes();
+            writeln!(&mut out, "    {pub_bytes:?},")?;
+        }
+        writeln!(&mut out, "];")?;
+    }
+
     Ok(())
 }
diff --git a/task/control-plane-agent/src/mgs_sidecar.rs b/task/control-plane-agent/src/mgs_sidecar.rs
index dcd68c28..7e349b93 100644
--- a/task/control-plane-agent/src/mgs_sidecar.rs
+++ b/task/control-plane-agent/src/mgs_sidecar.rs
@@ -41,6 +41,7 @@ use ignition_handler::IgnitionController;
 userlib::task_slot!(SIDECAR_SEQ, sequencer);
 userlib::task_slot!(MONORAIL, monorail);
 userlib::task_slot!(TRANSCEIVERS, transceivers);
+userlib::task_slot!(RNG, rng_driver);
 
 #[allow(dead_code)] // Not all cases are used by all variants
 #[derive(Clone, Copy, PartialEq, ringbuf::Count)]
@@ -66,6 +67,7 @@ enum Trace {
     ChallengeExpired,
     WrongKey,
     UntrustedResponse(GwMonorailError),
+    RngFillFailed(drv_rng_api::RngError),
 }
 counted_ringbuf!(Trace, 16, Trace::None);
 
@@ -284,6 +286,11 @@ impl MgsHandler {
         // possibility of new types being added.
         match (challenge, response) {
             (UnlockChallenge::Trivial, UnlockResponse::Trivial) => Ok(()),
+            (
+                UnlockChallenge::EcdsaSha2Nistp256(data),
+                UnlockResponse::EcdsaSha2Nistp256 { key, signature },
+            ) => verify_signature(&data, &key, &signature),
+            _ => Err(GwMonorailError::UnlockAuthFailed),
         }?;
         ringbuf_entry!(Trace::UnlockAuthSucceeded);
 
@@ -371,6 +378,63 @@ impl MgsHandler {
     }
 }
 
+fn verify_signature(
+    data: &[u8; 32],
+    key: &[u8; 65],
+    signature: &[u8; 64],
+) -> Result<(), GwMonorailError> {
+    if !TRUSTED_KEYS.iter().any(|t| t == key) {
+        ringbuf_entry!(Trace::WrongKey);
+        return Err(GwMonorailError::UnlockAuthFailed);
+    }
+
+    let sig = p256::ecdsa::Signature::from_bytes(signature.as_slice().into())
+        .map_err(|_| GwMonorailError::UnlockAuthFailed)?;
+
+    let v = p256::ecdsa::VerifyingKey::from_encoded_point(
+        &p256::EncodedPoint::from_bytes(key)
+            .map_err(|_| GwMonorailError::UnlockAuthFailed)?,
+    )
+    .map_err(|_| GwMonorailError::UnlockAuthFailed)?;
+
+    // Build an SSH signature blob to be verified
+    //
+    // See https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.sshsig
+    // for the signature blob format
+    #[rustfmt::skip]
+    let mut buf = [
+        // MAGIC_PREAMBLE
+        b'S', b'S', b'H', b'S', b'I', b'G',
+
+        // namespace
+        0, 0, 0, 15,
+        b'm', b'o', b'n', b'o', b'r', b'a', b'i', b'l', b'-',
+        b'u', b'n', b'l', b'o', b'c', b'k',
+
+        // reserved
+        0, 0, 0, 0,
+
+        0, 0, 0, 6, // hash type
+        b's', b'h', b'a', b'2', b'5', b'6',
+
+        0, 0, 0, 32, // hash of our actual data
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0,
+    ];
+
+    let mut hasher = sha2::Sha256::new();
+    use sha2::Digest;
+    hasher.update(data);
+    let hash = hasher.finalize();
+    buf[43..].copy_from_slice(&hash);
+
+    use p256::ecdsa::signature::Verifier;
+    v.verify(&buf, &sig)
+        .map_err(|_| GwMonorailError::UnlockAuthFailed)
+}
+
 impl SpHandler for MgsHandler {
     type BulkIgnitionStateIter = ignition_handler::BulkIgnitionStateIter;
     type BulkIgnitionLinkEventsIter =
@@ -572,7 +636,16 @@ impl SpHandler for MgsHandler {
             (SpComponent::MONORAIL, ComponentAction::Monorail(action)) => {
                 match action {
                     MonorailComponentAction::RequestChallenge => {
-                        let c = UnlockChallenge::Trivial;
+                        // TODO is this sufficient?
+                        let rng = drv_rng_api::Rng::from(RNG.get_task_id());
+                        let mut buf = [0u8; 32];
+                        rng.fill(&mut buf).map_err(|e| {
+                            ringbuf_entry!(Trace::RngFillFailed(e));
+                            SpError::Monorail(
+                                GwMonorailError::GetChallengeFailed,
+                            )
+                        })?;
+                        let c = UnlockChallenge::EcdsaSha2Nistp256(buf);
 
                         // Store the new challenge, which expires in 60 seconds
                         let now = sys_get_timer().now;
@@ -1013,3 +1086,5 @@ fn sp_error_from_ignition_error(err: IgnitionError) -> SpError {
     };
     SpError::Ignition(err)
 }
+
+include!(concat!(env!("OUT_DIR"), "/trusted_keys.rs"));
```

mkeeter:
For reference, I'm also maintaining the version with SSH-based unlock in the [`mkeeter/tech-port-unlock-ssh`](https://github.com/oxidecomputer/hubris/tree/mkeeter/tech-port-unlock-ssh), which is staged on top of this branch.  Once this branch is merged, I'll rebase it onto master and keep it up to date.

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: For both this and the above question, I've made `control-plane-agent` err on the side of being locked (e.g. it locks itself before trying to calling any RPC functions).

This can still leave the system in a partially-consistent state, but the caller will always get an error indicating that something went wrong.  There's no real way to guarantee a consistent state; attempts to recover could _also_ fail (if we believe the initial call could fail).

task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -32,9 +33,20 @@ enum Trace {
...
     Reinit,
+    UnlockUntil(u64),
+    LockingVLans,
+    AutomaticLock,
+    LockError(#[count(children)] VscError),
+    UnlockError(#[count(children)] VscError),
 }
 ringbuf!(Trace, 16, Trace::None);
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
hawkw, line 44: we shouldn't need to derive a `Debug` implementation for this, right? 
that's just extra strings in the binary we'll never use...
```suggestion
#[derive(Copy, Clone, PartialEq, Eq)]
```

task/net-api/src/lib.rs
```diff
@@ -50,6 +50,26 @@ pub enum RecvError {
     ServerRestarted = 2,
 }
 
+/// Errors that can occur when trying to set a VLAN as trusted
+#[derive(
+    Copy, Clone, Debug, PartialEq, Eq, FromPrimitive, IdolError, counters::Count,
```
hawkw, line 55: does this need a `Debug` impl? I don't see why it would unless it makes it to MGS?

task/net-api/src/lib.rs
```diff
@@ -265,5 +285,30 @@ impl From<Ipv6Address> for smoltcp::wire::Ipv6Address {
...
+/// Values are based on the KSZ8463's numbering (1-3); port 3 is connected to
+/// the SP itself and cannot be used as a source / destination.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum SpPort {
+    One,
+    Two,
+}
+
+/// Configuration for a SP VLAN
+#[derive(Copy, Clone, Debug)]
```
hawkw, line 299: again with the deriving `Debug`, always with the `Debug`... :)

hawkw (commented):
Some quick notes, not a complete review of the actual functionality implemented here, which I'd like to review in more detail.

task/monorail-server/src/bsp/sidecar_bcd.rs
cbiffle: FWIW I'd be surprised if the debug impl were compiled into the image, but it's worth testing! Saving space is nice.

cbiffle:
Couple high level comments from an initial scan and an IM conversation with Matt:

- I would argue that the server receiving an invalid VLAN should cause a reply-fault. I'm having a hard time coming up with a case where this ought to be tolerated.
- I wonder if it would be worth moving to VLAN indices, to contain the specific constants? Similar to how sockets are indexed instead of using raw port numbers. I'm not convinced this is a good idea, but I wanted to float it.
- It looks like the trust status will get reset if the net task crashes. Currently, the net task is designed so that a crash only causes packet loss, and affects no other externally observable state; storing the trust status in net would change this. Should the state be in Packrat and restored on net task restart? (It's possible that you've already done that and I missed it, I have not yet completed a thorough reading.)

app/sidecar/base.toml
```diff
@@ -78,10 +78,10 @@ task-slots = ["sys"]
 
 [tasks.net]
 name = "task-net"
-stacksize = 8000
+stacksize = 10000
```
cbiffle, line 81: What happened here?

app/sidecar/base.toml
```diff
@@ -78,10 +78,10 @@ task-slots = ["sys"]
...
 priority = 5
 features = ["mgmt", "h753", "sidecar", "vlan", "vpd-mac", "use-spi-core", "spi3"]
-max-sizes = {flash = 131072, ram = 65536, sram1 = 16384}
+max-sizes = {flash = 131072, ram = 131072, sram1 = 16384}
```
cbiffle, line 84: I assume this is a result of ~doubling the number of netstack instances required?

app/sidecar/base.toml
```diff
@@ -96,24 +96,24 @@ task-slots = ["sys", "packrat", { seq = "sequencer" }, "jefe"]
 [tasks.control_plane_agent]
 name = "task-control-plane-agent"
 priority = 7
-max-sizes = {flash = 131072, ram = 32768 }
-stacksize = 6144
+max-sizes = {flash = 140672, ram = 65536 }
+stacksize = 12000
```
cbiffle, line 100: Do we know what doubled the stack requirement here?

app/sidecar/base.toml
```diff
@@ -1114,6 +1119,7 @@ owner = {name = "control_plane_agent", notification = "socket"}
 port = 11111
 tx = { packets = 3, bytes = 2048 }
 rx = { packets = 3, bytes = 2048 }
+accept-untrusted = true
```
cbiffle, line 1122: So, how would you feel about `allow-untrusted`? I bring this up because
- this setting applies to both transmit and receive, whereas normally in a firewall config I think of "accept" as applying to inbound traffic only
- "allow" puts it more in line with allow/deny terminology in packet filters

build/net/src/lib.rs
```diff
@@ -82,23 +90,65 @@ pub fn load_net_config() -> Result<NetConfig> {
...
 
+pub fn generate_port_consts(
+    config: &NetConfig,
+    mut out: impl std::io::Write,
+) -> Result<(), std::io::Error> {
+    let ports = config
+        .vlans
+        .iter()
+        .map(|v| v.port)
+        .collect::<HashSet<_>>()
```
cbiffle, line 101: I had to stare at this for a bit to convince myself that you don't wind up relying on the iteration order of HashSet. (Because using HashSet in the build system in general is a Big Red Flag.)

Very weak preference for BTreeSet here just to avoid looking sus.

If you were open to `cargo add itertools` you could rephrase this as `.unique().count()` (you'd have to `use itertools::Itertools`).

build/net/src/lib.rs
```diff
@@ -82,23 +90,65 @@ pub fn load_net_config() -> Result<NetConfig> {
...
+
 pub fn generate_vlan_consts(
     config: &NetConfig,
     mut out: impl std::io::Write,
 ) -> Result<(), std::io::Error> {
-    let vlan = config.vlan.unwrap();
-    let end = vlan.start + vlan.count;
-    if end > 0xFFF {
-        panic!("Invalid VLAN range (must be < 4096)");
+    if config.vlans.iter().any(|v| v.vid > 0xFFF) {
```
cbiffle, line 111: Is there code that ensures that the vids are unique (I don't see it)? Is it important that the vids be unique (I'm guessing so)?

Would it be worth representing this as a vid-keyed map in the TOML to make this clear? (I'm not sure.)

build/net/src/lib.rs
```diff
@@ -82,23 +90,65 @@ pub fn load_net_config() -> Result<NetConfig> {
...
     }
     writeln!(
         out,
         "
-pub const VLAN_RANGE: core::ops::Range<u16> = {:#x}..{:#x};
-pub const VLAN_COUNT: usize = {};
-",
-        vlan.start, end, vlan.count
-    )
+    pub const VLAN_COUNT: usize = {0};
```
cbiffle, line 117: It'd be really nice to use quote here instead of the dedented string literals. Not mandatory.

build/net/src/lib.rs
```diff
@@ -144,5 +194,16 @@ pub fn generate_socket_enum(
     }
     writeln!(out, "];")?;
 
+    writeln!(
+        out,
+        "#[allow(unused)]\
```
cbiffle, line 199: The backslash causes this to be generated on the same line as the pub const, was that deliberate?

drv/ksz8463/src/registers.rs
```diff
@@ -1074,32 +1074,43 @@ pub enum Register {
...
     }
+
+    // Helper function to dispatch between two registers
+    #[inline(always)]
+    fn select3(i: u8, r1: Register, r2: Register, r3: Register) -> Register {
+        match i {
+            1 => r1,
+            2 => r2,
+            3 => r3,
+            _ => panic!("Invalid port {}", i),
```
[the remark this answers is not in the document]
cbiffle: The integer formatting code that this pulls in is also quite expensive, over and above the panic message.

drv/monorail-api/src/lib.rs
```diff
@@ -130,6 +130,9 @@ pub enum MonorailError {
     /// The given port does not have a PHY associated with it
     NoPhy,
 
+    /// The given function is not implemented
+    NotImplemented,
```
cbiffle, line 134: Under what circumstances would a correct program call an unimplemented function in monorail? (i.e. should this be replaced by reply-fault?)

drv/ksz8463/src/registers.rs
```diff
@@ -1074,32 +1074,43 @@ pub enum Register {
...
     #[inline(always)]
     pub fn PxCR2(i: u8) -> Self {
-        Self::select(i, Self::P1CR2, Self::P2CR2)
+        Self::select3(i, Self::P1CR2, Self::P2CR2, Self::P3CR2)
     }
 
     // Helper function to dispatch between two registers
     #[inline(always)]
-    fn select(i: u8, r1: Register, r2: Register) -> Register {
+    fn select2(i: u8, r1: Register, r2: Register) -> Register {
```
cbiffle, line 1098: So it kinda feels like `i` should be an enum, eliminating the need for runtime integer checking? But I haven't followed all the uses to _their_ use sites so there may be a reason why it's an arbitrary byte.

task/control-plane-agent/src/main.rs
```diff
@@ -585,25 +589,10 @@ impl NetHandler {
...
-#[allow(dead_code)]
-fn vlan_id_from_sp_port(port: SpPort) -> u16 {
-    use task_net_api::VLAN_RANGE;
-    assert_eq!(VLAN_RANGE.len(), 2);
-
-    match port {
-        SpPort::One => VLAN_RANGE.start,
-        SpPort::Two => VLAN_RANGE.start + 1,
+    use task_net_api::VLANS;
+    match VLANS.iter().find(|v| v.vid == meta.vid).unwrap().port {
```
cbiffle, line 593: Please use `unwrap_lite` if this must panic.

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
```diff
@@ -181,11 +180,23 @@ impl HostPhase2Requester {
...
 
+        let vid = {
+            let target = match port {
+                SpPort::One => task_net_api::SpPort::One,
+                SpPort::Two => task_net_api::SpPort::Two,
+            };
+            task_net_api::VLANS
+                .iter()
+                .find(|v| v.port == target)
+                .unwrap()
```
cbiffle, line 191: Please use `unwrap_lite` if this must panic. (An `EnumMap` might be nice here if you expect there to always be a mapping for every variant in `SpPort`...)

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -58,12 +90,23 @@ pub(crate) type BorrowedUpdateBuffer = update_buffer::BorrowedUpdateBuffer<
...
+}
+
 pub(crate) struct MgsHandler {
     common: MgsCommon,
     sequencer: Sequencer,
     monorail: Monorail,
     transceivers: Transceivers,
     ignition: IgnitionController,
+
+    last_challenge: Option<(UnlockChallenge, u64)>,
```
cbiffle, line 107: Should `last_challenge` be per-port?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -90,13 +142,32 @@ impl MgsHandler {
...
     }
 
     pub(crate) fn handle_timer_fired(&mut self) {
         // This is a no-op if we're not preparing for an SP update.
         self.common.sp_update.step_preparation();
+
+        let now = sys_get_timer().now;
+        for (i, k) in self.locked.clone().iter().enumerate() {
+            if let LockState::UnlockedUntil(lock_at) = k {
+                if now >= *lock_at {
```
cbiffle, line 163: fwiw, this has recreated most of multitimer. (You'd need an enum rather than an integer naming your vlans to use multitimer.)

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +237,204 @@ impl MgsHandler {
...
+            ringbuf_entry!(Trace::UnlockTimeTooLong(time_sec));
+            return Err(GwMonorailError::TimeIsTooLong);
+        }
+
+        let Some(vlan_index) = VLANS.iter().position(|v| v.vid == vid) else {
+            return Err(GwMonorailError::InvalidVLAN);
+        };
+
+        if VLANS[vlan_index].trusted {
+            return Err(GwMonorailError::AlreadyTrusted);
```
cbiffle, line 261: Should this be an _error,_ or should it bump the unlock time window forward?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +237,204 @@ impl MgsHandler {
...
+        }
+
+        let Some((last_challenge, challenge_time)) =
+            core::mem::take(&mut self.last_challenge)
+        else {
+            ringbuf_entry!(Trace::NoChallenge);
+            return Err(GwMonorailError::UnlockAuthFailed);
+        };
+
+        if challenge != last_challenge {
```
cbiffle, line 271: Note that this allows an intervening message to make a valid challenge unwinnable (because you `take` the challenge info and don't put it back on error paths). Perhaps that's deliberate, but if so I would suggest a comment stating the intent.

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +237,204 @@ impl MgsHandler {
...
+        ringbuf_entry!(Trace::UnlockAuthSucceeded);
+
+        // Pick how long we'll trust things for
+        let now = sys_get_timer().now;
+        let trust_until = now + u64::from(time_sec) * 1000;
+
+        // Reconfigure the management network for arbitrary SP access
+        if let Err(e) = self.monorail.unlock_vlans(trust_until) {
+            ringbuf_entry!(Trace::MonorailUnlockFailed(e));
+            return Err(GwMonorailError::UnlockFailed);
```
cbiffle, line 301: Should this take the underlying error as a parameter, so that the caller can inspect it without needing a copy of our ringbuffer?

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +237,204 @@ impl MgsHandler {
...
+        }
+
+        // Reconfigure the net task to temporarily trust this port's VLAN
+        let net = task_net_api::Net::from(crate::NET.get_task_id());
+        if let Err(e) = net.trust_vlan(vid, trust_until) {
+            ringbuf_entry!(Trace::TrustVLanFailed(e));
+            return Err(GwMonorailError::UnlockFailed);
```
cbiffle, line 308: (same question, right now these two failures are indistinguishable to the caller)

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +237,204 @@ impl MgsHandler {
...
+    }
 }
 
 impl SpHandler for MgsHandler {
     type BulkIgnitionStateIter = ignition_handler::BulkIgnitionStateIter;
     type BulkIgnitionLinkEventsIter =
         ignition_handler::BulkIgnitionLinkEventsIter;
 
+    /// Checks whether we trust the given message
+    fn is_request_trusted(
```
cbiffle, line 387: Having a function called `is_condition` that doesn't return a `bool` is weird. Maybe `ensure_request_trusted`?

This formulation also has "parse don't validate" implications. It would be neat if a trusted message had a different type via newtype or something; this would clearly distinguish which points in the code need to have a trusted message vs not. Not mandatory, but something to think about.

task/monorail-server/src/bsp/sidecar_bcd.rs
cbiffle: Yeah, that seems like a reasonable concern -- need to ensure that `configure_vlan_sidecar_unlocked` reliably restores state on any error, which I bet it doesn't.

task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -548,6 +585,31 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
+        Ok(())
+    }
+
+    pub(crate) fn lock_vlans(&mut self) -> Result<(), MonorailError> {
+        ringbuf_entry!(Trace::LockingVLans);
+        self.vsc7448.configure_vlan_sidecar_locked().map_err(|e| {
+            ringbuf_entry!(Trace::LockError(e));
+            MonorailError::from(e)
+        })?;
+        self.vlan_mode = VLanMode::Locked;
```
cbiffle, line 610: Same concern here.

Is there a way we could eliminate `vlan_mode` and ask the actual configured hardware whether we're locked or not, so they can't go out of sync?

task/net/src/server.rs
```diff
@@ -8,22 +8,37 @@ use crate::notifications;
...
+#[allow(dead_code)]
+#[derive(Copy, Clone, Eq, PartialEq, counters::Count)]
+enum Trace {
+    #[count(skip)]
+    None,
+    SetTrust {
+        vid: u16,
+        t: VLanTrust,
+    },
+    TrustExpired(u16),
```
cbiffle, line 31: As a person who mostly interprets other people's ringbufs, it would be great if all primitive fields like this had names (like it does for SetTrust above)

task/net/src/server.rs
```diff
@@ -577,18 +710,26 @@ where
...
+        // Convert from absolute VID to an index in our VLAN array
         #[cfg(feature = "vlan")]
-        let vlan_index = {
-            // Convert from absolute VID to an index in our VLAN array
-            if !VLAN_RANGE.contains(&metadata.vid) {
-                return Err(ClientError::BadMessageContents.fail());
-            }
-            usize::from(metadata.vid - VLAN_RANGE.start)
+        let Some(vlan_index) = VLANS.iter().position(|v| v.vid == metadata.vid) else {
+            return Err(ClientError::BadMessageContents.fail());
```
cbiffle, line 716: So it looks like in some cases, we _are_ reply-faulting in response to an invalid vlan tag? But in others (like monorail) we aren't

task/net/src/server.rs
```diff
@@ -577,18 +710,26 @@ where
...
 
         let vlan = &mut self.vlan_state[vlan_index];
+
+        // Refuse to send messages directed to an untrusted VLAN, silently
+        // dropping them.
+        let now = userlib::sys_get_timer().now;
+        let trust = vlan.check_trust(now)
+            | generated::SOCKET_ACCEPT_UNTRUSTED[socket_index];
+        if !trust {
+            return Ok(());
```
cbiffle, line 730: This should hit a counter.

cbiffle (commented):
I really appreciate the parameter/interface cleanup on the internal MGS trait.

drv/vsc7448/src/lib.rs
```diff
@@ -725,40 +732,80 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
     /// - The uplink port (49) sends and receives packets with one VLAN tag, and
     ///   uses that packet to select which VLAN (i.e. which downstream port)
     ///   should receive that packet.  The VLAN tag is stripped on egress.
-    fn configure_vlan_with_mask(
+    fn configure_vlan_with_mask_and_tagged<F: Fn(u8) -> (u64, bool)>(
         &self,
-        f: fn(u8) -> u64,
+        f: F,
```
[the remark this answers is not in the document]
mkeeter: This changed a bunch (splitting VLAN configuration versus port configuration), but both now take a `fn(u8) -> ??`

drv/ksz8463/src/lib.rs
```diff
@@ -438,13 +455,65 @@ impl<S: SpiServer> Ksz8463<S> {
...
+                self.write_vlan_table(3, 0b110, 0x302)?;
+                self.write_vlan_table(4, 0b000, 0x3FF)?;
+                for i in 5..16 {
+                    self.disable_vlan(i)?;
+                }
+
+                // Assign default VLAN tags to each port
+                self.write(Register::P1VIDCR, 0x3FF)?;
+                self.write(Register::P2VIDCR, 0x302)?;
+                self.write(Register::P3VIDCR, 0x3FF)?;
```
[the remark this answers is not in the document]
mkeeter: Done in f4d80936

drv/ksz8463/src/registers.rs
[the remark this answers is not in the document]
mkeeter: This was a pre-existing condition above, but I fixed it everywhere with `enum KszPort` (for `select3`) and `enum KszPhyPort` (for `select2`); see e31bff0db3e0

drv/vsc7448/src/lib.rs
```diff
@@ -806,23 +853,48 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
...
+            }
+        })
+    }
+
+    /// Implements the VLAN scheme described in RFD 492
+    pub fn configure_vlan_sidecar_locked(&self) -> Result<(), VscError> {
+        const LOCAL_SP: u8 = 48; // Connection to the local SP
+        const UPLINK: u8 = 49; // Uplink to the Tofino 2
+        const TECHNICIAN_1: u8 = 44;
+        const TECHNICIAN_2: u8 = 45;
```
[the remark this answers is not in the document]
mkeeter: Yup, done in the grand reorganization

task/monorail-server/src/bsp/sidecar_bcd.rs
mkeeter: Yeah, removing the derive doesn't change the code size at all here.

task/net-api/src/lib.rs
mkeeter: This is a case of preexisting condition copypasta – every other type in here also derives `Debug` 😉 

I removed it, although I don't believe it makes a difference since the compiler will strip it.

(There are a [_bunch_ of things](https://gist.github.com/mkeeter/ce77b889fd12cb26e4a10d2b5f16ca06) that derive debug; we could do a separate pass to clean them out)

task/net-api/src/lib.rs
mkeeter: this is my emotional support `#[derive(Debug)]`, it's allowed to go into the embedded system

task/net/src/server.rs
mkeeter: Good idea, done in 6359a7b5

app/sidecar/base.toml
mkeeter: I like it, done in 677894e28891

build/net/src/lib.rs
mkeeter: I am against bad vibes and looking sus; changed to `BTreeSet` in 1528c24be825

drv/ksz8463/src/registers.rs
mkeeter: Yup, done in e31bff0d + 47c1bc4f (lots of plumbing was required)

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: I changed the name of this flag to `always_trusted`; it should be an error, because you can't change the trust state of an always-trusted VLAN (which in this case is the Tofino upstream).

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: Not easily; `GwMonorailError` is an error type from `gateway-messages`, which is in a separate crate and can't use the local `MonorailError` type.  We could try to keep them one-to-one, but that would be annoying.

build/net/src/lib.rs
mkeeter: Done in the Grand VLAN refactoring!

build/net/src/lib.rs
mkeeter: Also done in the Grand VLAN refactoring!

build/net/src/lib.rs
mkeeter: "Intentional" is a strong word for it; all the other writes in this function also do it.

drv/vsc7448/src/lib.rs
```diff
@@ -725,40 +732,80 @@ impl<'a, R: Vsc7448Rw> Vsc7448<'a, R> {
...
+                self.modify(port.VLAN_CTRL(), |r| {
+                    r.set_vlan_pop_cnt(1);
+                    r.set_vlan_aware_ena(1);
+                })?;
+                self.modify(port.VLAN_TPID_CTRL(), |r| {
+                    // Only accept 0x8100 as a valid TPID, to keep things
+                    // simple. This is the designated EtherType for "Customer
+                    // VLAN tag" in IEEE 802.1Q; the register's polarity
+                    // requires us to **clear** bit 0 to accept this TPID.
+                    r.set_basic_tpid_aware_dis(0b1110);
```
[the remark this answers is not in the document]
mkeeter: Done in 68f8519e4d6f, but it's true that most of the configuration here only makes sense if read alongside the datasheet.

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -166,19 +237,204 @@ impl MgsHandler {
...
+            } else {
+                ringbuf_entry!(Trace::TimedRelock(VLANS[i].vid));
+                self.lock(VLANS[i].vid)?;
+                ringbuf_entry!(Trace::MessageNotTrusted(VLANS[i].vid));
+                Err(GwMonorailError::ManagementNetworkLocked)
+            }
+        } else {
+            ringbuf_entry!(Trace::MessageNotTrusted(VLANS[i].vid));
+            Err(GwMonorailError::ManagementNetworkLocked)
+        }
```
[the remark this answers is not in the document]
mkeeter: This has changed in the Grand VLAN Refactoring!

task/control-plane-agent/src/main.rs
mkeeter: Obviated in the Grand VLAN Refactoring

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
mkeeter: Also fixed in the Grand VLAN Refactoring

task/net/src/server.rs
mkeeter: Fixed in the Grand VLAN Refactoring; these functions now take an `enum VLanId`.

drv/ksz8463/src/lib.rs
```diff
@@ -438,13 +455,65 @@ impl<S: SpiServer> Ksz8463<S> {
...
+                    // tags (and untagged frames will be assigned 0x3FF then
+                    // unceremoniously dropped).
+                    self.modify(Register::PxCR2(i), |r| *r |= 1 << 14)?;
+                }
+
+                // There's a secret bonus register which _actually_ enables
+                // PVID tagging, despite not being mentioned in the VLAN
+                // tagging section of the datasheet.  We enable tagging when
+                // frames come from Port 2 and go to Port 3.
+                self.write(Register::SGCR9, 1 << 3)?;
```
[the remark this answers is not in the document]
mkeeter: Refactored into helper functions in 60bc07d201cf

drv/monorail-api/src/lib.rs
mkeeter: Switched to reply-fault in 45a93d961a14

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: Same thing here, we don't want to / can't easily expose internal `MonorailError` types to MGS.

app/sidecar/base.toml
mkeeter: Not sure!  `MAXDEPTH` goes from 6160 → 8392, so it's a noticeable chance.

app/sidecar/base.toml
mkeeter: I believe this was adding the crypto verification code, which has since been pulled out into a separate branch.  We can revert the `stacksize` changes, or just leave them as-is, since we'll need it eventually.

app/sidecar/base.toml
mkeeter: Yup!  RAM doesn't quite double – it goes from 79420 to 98304 – but we're limited to power-of-two sizes for the RAM region (only flash gets to use spare regions, since it's a tighter commodity).

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -632,17 +881,13 @@ impl SpHandler for MgsHandler {
...
-        _sender: SocketAddrV6,
-        _port: SpPort,
     ) -> Result<gateway_messages::StartupOptions, SpError> {
         ringbuf_entry_root!(Log::MgsMessage(MgsMessage::GetStartupOptions));
         Err(SpError::RequestUnsupportedForSp)
     }
 
     fn set_startup_options(
         &mut self,
-        _sender: SocketAddrV6,
```
aapoalas, line 644: **suggestion:** Could these removals of unused parameters be done in a trivial side-PR?

build/net/src/lib.rs
```diff
@@ -82,23 +98,129 @@ pub fn load_net_config() -> Result<NetConfig> {
...
-    let vlan = config.vlan.unwrap();
-    let end = vlan.start + vlan.count;
-    if end > 0xFFF {
-        panic!("Invalid VLAN range (must be < 4096)");
+    let ports = config
+        .vlans
+        .values()
+        .map(|v| v.port)
+        .collect::<BTreeSet<_>>()
+        .len();
```
hawkw, line 110: it took me a sec to realize that the reason we did this instead of using `Iterator::count()` was that we wanted to count unique ports, which...is obvious in hindsight, so maybe it's not really worth adding a comment on, i dunno...

hawkw: (minor allocation golf that almost certainly doesn't matter: it might be nicer to use a `HashSet` here, since we don't care about order, just uniqueness, and we won't have to allocate for each entry?)

build/net/src/lib.rs
```diff
@@ -82,23 +98,129 @@ pub fn load_net_config() -> Result<NetConfig> {
...
+    let port_count = ports.max(1);
+    let s = quote! {
+        pub const PORT_COUNT: usize = #port_count;
+    };
+    writeln!(out, "{s}")
+}
+
+fn check_vlan_config(config: &NetConfig) {
+    if config.vlans.values().any(|v| v.vid > 0xFFF) {
+        panic!("Invalid VLAN VID (must be < 4096)");
```
hawkw, line 121: nitpicky: maybe it would be better to use `Iterator::find` to actually return the first VID that's greater than 4096, so we can print what the value was?
```suggestion
    if let Some(v) = config.vlans.values().find(|v| v.vid > 0xFFF) {
        panic!("Invalid VLAN VID {} (must be < 4096)", v.vid);
```

build/net/src/lib.rs
```diff
@@ -82,23 +98,129 @@ pub fn load_net_config() -> Result<NetConfig> {
...
+    config: &NetConfig,
+    mut out: impl std::io::Write,
+) -> Result<(), std::io::Error> {
+    check_vlan_config(config);
+
+    let vlan_count = config.vlans.len();
+    let vids = config.vlans.values().map(|cfg| cfg.vid).collect::<Vec<_>>();
+    let s = quote! {
+        #[allow(unused)]
+        pub const VLAN_COUNT: usize = #vlan_count;
```
hawkw, line 142: out of curiosity: what's the motivation for having a separate constant for this, instead of just using `VLAN_VIDS.len()`?

build/net/src/lib.rs
```diff
@@ -82,23 +98,129 @@ pub fn load_net_config() -> Result<NetConfig> {
...
+    check_vlan_config(config);
+
+    let s = if config.vlans.is_empty() {
+        quote! {
+            #[derive(
+                Copy, Clone, Eq, PartialEq,
+                enum_map::Enum,
+                serde::Serialize, serde::Deserialize,
+                hubpack::SerializedSize,
+            )]
```
hawkw, line 167: nitpicky, feel free to ignore me: maybe the code inside the `if` could *just* make the list of varriants, and we could `quote!` that list of variants outside of the if with the rest of the enum definition, like:
```rust
let vlans = if config.vlans.is_empty() {
    quote! { None, }
} else {
   // ... the rest of the stuff ...
};
let s = quote! {
    #[derive(
        Copy, Clone, Eq, PartialEq,
        enum_map::Enum,
        serde::Serialize, serde::Deserialize,
        hubpack::SerializedSize,
    )]
    pub enum VLanId {
        #(vlans)
    }
};
```
that way, we don't have the opportunity for annoying bugs where we add an additional derive, and forget to add it to the "no VLANs" case, or something.

this is probably not super important, but it's the kind of thing i think too hard about...

drv/ksz8463/src/registers.rs
```diff
@@ -1073,33 +1074,52 @@ pub enum Register {
...
+    fn select2(i: KszPhyPort, r1: Register, r2: Register) -> Register {
+        match i {
+            KszPhyPort::One => r1,
+            KszPhyPort::Two => r2,
+        }
+    }
+
+    // Helper function to dispatch between three registers
+    #[inline(always)]
+    fn select3(
```
hawkw, line 1113: nit, take it or leave it: now that we have enums for `KszPort` and `KszPhyPort`, maybe these ought to be renamed to `select_phy_port` and `select_port` or something?
```suggestion
    fn select_phy_port(i: KszPhyPort, r1: Register, r2: Register) -> Register {
        match i {
            KszPhyPort::One => r1,
            KszPhyPort::Two => r2,
        }
    }

    // Helper function to dispatch between three registers
    #[inline(always)]
    fn select_port(
```

hawkw (commented):
some small and unimportant notes on the codegen stuff

build/net/src/lib.rs
mkeeter: lol Cliff literally suggested switching from `HashSet` to `BTreeSet` in [this comment](https://github.com/oxidecomputer/hubris/pull/1835#discussion_r1699203911)

hawkw: welp, disregard that.

task/monorail-server/src/bsp/sidecar_bcd.rs
mkeeter: I'm on the fence here – yes, we can read back configuration registers, but they too could be in an inconsistent state – or the read could fail, since it's going to an external device!

I think the combination of

1) Fail towards the locked state
2) Return an error to the user if anything goes wrong (and the state could be inconsistent)

is okay for a first draft, since we'll be doing this manually and won't ignore the errors

build/net/src/lib.rs
mkeeter: I'm going to pass on this because the two branches don't just define the `enum VLanId`; the non-empty branch also adds an impl block, so splitting them is trickier.

drv/ksz8463/src/registers.rs
mkeeter: Good idea, done

build/net/src/lib.rs
mkeeter: Historical inertia; I switched to using `VLanId::LENGTH` (which is generated from `enum_map::Enum`)

build/net/src/lib.rs
mkeeter: Sure, done

task/net/src/server.rs
mkeeter: I added a ringbuf message and made it a counted ringbuf, so we'll now get counters for free!

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -40,6 +42,36 @@ userlib::task_slot!(SIDECAR_SEQ, sequencer);
 userlib::task_slot!(MONORAIL, monorail);
 userlib::task_slot!(TRANSCEIVERS, transceivers);
 
+#[allow(dead_code)] // Not all cases are used by all variants
+#[derive(Clone, Copy, PartialEq, ringbuf::Count)]
```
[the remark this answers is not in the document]
mkeeter: The IPC client counters should cover all of the error variants here, but I did add per-`VLanId` counting

mkeeter:
> Couple high level comments from an initial scan and an IM conversation with Matt:
> 
>     * I would argue that the server receiving an invalid VLAN should cause a reply-fault. I'm having a hard time coming up with a case where this ought to be tolerated.
> 
>     * I wonder if it would be worth moving to VLAN indices, to contain the specific constants? Similar to how sockets are indexed instead of using raw port numbers. I'm not convinced this is a good idea, but I wanted to float it.
> 
>     * It looks like the trust status will get reset if the net task crashes. Currently, the net task is designed so that a crash only causes packet loss, and affects no other externally observable state; storing the trust status in net would change this. Should the state be in Packrat and restored on net task restart? (It's possible that you've already done that and I missed it, I have not yet completed a thorough reading.)

The VLAN handling has been updated to generate an `enum VLanId`, and make generous use of `EnumMap` objects, so everything is now more type-safe.

Packrat is write-once, so we couldn't use it to store `net` state that changes.  We could change this in packrat, of course, but I'm okay with `net` failing towards locked if it manages to crash, because it should only ever be unlocked during active debugging work (so we'll notice).

jgallagher:
> Packrat is write-once, so we couldn't use it to store `net` state that changes.

FWIW this varies by API. Mostly this is true (as evidenced by all the [`Self::set_once()`](https://github.com/oxidecomputer/hubris/blob/a3e3b7b83bd0bb99170015523dbb4d962a0aa157/task/packrat/src/main.rs#L180) calls in the idol server), but we also store host startup options which are allowed to change multiple times ([no set_once call](https://github.com/oxidecomputer/hubris/blob/a3e3b7b83bd0bb99170015523dbb4d962a0aa157/task/packrat/src/main.rs#L226-L227)), because we wanted those to persist in the face of a control-plane-agent crash.

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
```diff
@@ -181,11 +180,16 @@ impl HostPhase2Requester {
 
         let n = gateway_messages::serialize(tx_buf, &message).unwrap_lite();
 
+        let vid = match port {
+            SpPort::One => task_net_api::VLanId::Sidecar2,
```
cbiffle, line 184: What determines this, and is the 1=>2; 2=>1 here deliberate? (If so, a comment should do it.)

task/control-plane-agent/src/mgs_sidecar.rs
cbiffle: Okay. I'm on the lookout for improved patterns for situations like this, because debugging this sort of situation in a prod system can be hard. No ideas off the top of my head for this case, though.

task/control-plane-agent/src/mgs_sidecar.rs
cbiffle: I think in general this would have made the diff easier to follow, but chained PRs on Github are hard to manage so I sympathize with Matt's motivation for not doing so.

task/monorail-server/src/bsp/sidecar_bcd.rs
cbiffle: (determination on this is on line 614 for the record)

task/monorail-server/src/server.rs
```diff
@@ -792,6 +792,23 @@ impl<'a, R: Vsc7448Rw> idl::InOrderMonorailImpl for ServerImpl<'a, R> {
...
+        self.bsp
+            .unlock_vlans_until(unlock_until)
+            .map_err(RequestError::from)
+    }
+
+    fn lock_vlans(
+        &mut self,
+        _mgs: &userlib::RecvMessage,
+    ) -> Result<(), RequestError<MonorailError>> {
+        self.bsp.lock_vlans().map_err(RequestError::from)
```
cbiffle, line 810: Whichever syntax you prefer is fine, but as an FYI, I'm pretty sure this is equivalent to `Ok(self.bsp.lock_vlans()?)` which is shorter.

task/net/src/mgmt.rs
```diff
@@ -219,15 +222,19 @@ impl Bsp {
...
 
-        for i in 0..2 {
+        for (i, port) in
+            [KszPhyPort::One, KszPhyPort::Two].into_iter().enumerate()
+        {
             // The KSZ8463 numbers its ports starting at 1 (e.g. P1MBSR)
-            let port = i as u8 + 1;
             match self.ksz8463.read(KszRegister::PxMBSR(port)) {
                 Ok(sr) => {
                     s.ksz8463_100base_fx_link_up[i] = (sr & (1 << 2)) != 0
```
cbiffle, line 231: If I'm not mistaken, the need to enumerate() this iterator could be avoided if this link_up array were an enum map keyed by KszPort.

task/net/src/server.rs
```diff
@@ -346,19 +449,71 @@ where
...
-        storage: &'static mut [Storage; N],
-        sockets: generated::Sockets<'static, N>,
-        mut mkdevice: impl FnMut(usize) -> E,
+        storage: &'static mut [Storage; VLanId::LENGTH],
+        sockets: generated::Sockets<'static, { VLanId::LENGTH }>,
+        mut mkdevice: impl FnMut(VLanId) -> E,
     ) -> Self {
         // Local storage; this will end up owned by the returned ServerImpl.
-        let mut vlan_state: Vec<VLanState<E>, N> = Vec::new();
+        let mut vlan_state: Vec<VLanState<E>, { VLanId::LENGTH }> = Vec::new();
```
cbiffle, line 457: In the code below, it kinda feels like this would be simplified if this started life as a mutable `EnumMap` instead of getting converted at the end?

cbiffle (commented):
Generally looks good! Matt and I are going to talk through my comments at 1400 PDT and get them resolved.

cbiffle (approved):
Alright, of the things I highlighted, the only one that might be a bug is the case where our mapping between port enums swaps 1=>2 / 2=>1. Matt's going to check whether it's correct and if so add a comment.

We should be clear to merge this after that happens.

task/net/src/server.rs
mkeeter: We talked about this in person; we can't build a default `VLanState`, so we're using `Vec` to incrementally build the array

task/net/src/mgmt.rs
mkeeter: I looked at this, but it gets messy – Humility uses this type for the `net status` command, so changing it would require teaching Humility how to decode an `EnumMap`.

task/control-plane-agent/src/mgs_gimlet/host_phase2.rs
mkeeter: Yup, this is wrong; fixed

PR #1843, opened 2024-08-13
grapefruit: add fmc-demo IP interface

cbiffle:
This adds a UDP interface for messing with Grapefruit's FPGA. The protocol is documented in `drv/stm32h7-fmc-demo-server/README.md`.

### DIFF
--- a/app/grapefruit/app.toml
+++ b/app/grapefruit/app.toml
@@ -244,8 +244,9 @@ name = "drv-stm32h7-fmc-demo-server"
 features = ["h753"]
 priority = 4
 start = true
-task-slots = ["sys"]
+task-slots = ["sys", "net"]
 uses = ["fmc", "fmc_nor_psram_bank_1"]
+notifications = ["socket"]
 
 [tasks.fmc_nor_flash]
 name = "drv-fmc-nor-flash"
@@ -293,6 +294,13 @@ owner = {name = "dump_agent", notification = "socket"}
 port = 11113
 tx = { packets = 3, bytes = 1024 }
 rx = { packets = 3, bytes = 1024 }
+
+[config.net.sockets.fmc_test]
+kind = "udp"
+owner = {name = "fmc_demo", notification = "socket"}
+port = 11114
+tx = { packets = 3, bytes = 4096 }
+rx = { packets = 3, bytes = 4096 }
 
 ################################################################################
 
--- a/drv/stm32h7-fmc-demo-server/README.md
+++ b/drv/stm32h7-fmc-demo-server/README.md
@@ -2,3 +2,37 @@
 
 This task maps in a 16-bit multiplexed PSRAM bus and lets you do incredibly
 ill-advised things to it. It is intended for interface testing only, do not eat.
+
+
+## Network protocol
+
+This task can be bound to a UDP socket to allow ill-advised things over the
+network. Here's the UDP protocol.
+
+Packets sent to the device always begin with a two-byte header:
+- `version` (1 byte): currently 0
+- `command` (1 byte)
+
+The only currently defined `command` is 0.
+
+Following this is a series of arguments to command 0, which describe a sequence
+of memory operations. Arguments are simply concatenated together. Each starts
+with an identifier byte, and some are followed by additional bytes. Any sequence
+of the following is acceptable:
+
+- `address` (0) followed by a little-endian 32-bit address. This loads the
+  address into the internal address register used by subsequent peeks and pokes.
+- `peek` (1/2/3/4) reads an 8/16/32/64-bit (respectively) value from memory at
+  the current address and concatenates it (in little endian order) onto the
+  response packet.
+- `peek_advance` (5/6/7/8) operate like the corresponding `peek` operations, but
+  also advance the address register by the size of data being operated on.
+- `poke` (9/10/11/12), followed by an 8/16/32/64-bit (respectively)
+  little-endian value, writes that value into memory at the address in the
+  address register. Nothing is appended to the response.
+- `poke_advance` (13/14/15/16) operate like the corresponding `poke` operations,
+  but also advance the address register by the size of data being operated on.
+
+The response starts with a single byte, which is zero for success and non-zero
+for failure. On success, the results of all peek operations are concatenated
+after the first byte, in order.
--- a/drv/stm32h7-fmc-demo-server/src/main.rs
+++ b/drv/stm32h7-fmc-demo-server/src/main.rs
@@ -7,8 +7,13 @@
 #![no_std]
 #![no_main]
 
-use core::convert::Infallible;
+use core::{convert::Infallible, mem::take};
+use counters::{count, Count};
+use static_cell::ClaimOnceCell;
 use sys_api::{Alternate, OutputType, Peripheral, Port, Pull, Speed};
+use task_net_api::{
+    LargePayloadBehavior, Net, RecvError, SendError, SocketName,
+};
 use userlib::*;
 
 #[cfg(feature = "h743")]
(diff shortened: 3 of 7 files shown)


drv/stm32h7-fmc-demo-server/src/main.rs
```diff
@@ -226,15 +251,57 @@ fn main() -> ! {
...
 impl NotificationHandler for ServerImpl {
     fn current_notification_mask(&self) -> u32 {
-        0
+        notifications::SOCKET_MASK
     }
 
-    fn handle_notification(&mut self, _bits: u32) {
-        unreachable!()
+    fn handle_notification(&mut self, bits: u32) {
+        if bits & notifications::SOCKET_MASK != 0 {
```
aapoalas, line 265: **suggestion:** Invert the condition and do an early return to get one less level of indentation for the loop?

aapoalas (commented):
Comment from the peanut gallery.

That's a nice little debug API <3

app/grapefruit/app.toml
```diff
@@ -294,6 +295,13 @@ port = 11113
 tx = { packets = 3, bytes = 1024 }
 rx = { packets = 3, bytes = 1024 }
 
+[config.net.sockets.fmc_test]
+kind = "udp"
+owner = {name = "fmc_demo", notification = "socket"}
+port = 11114
```
mkeeter, line 301: This should be reserved in [`oxidecomputer/oana`](https://github.com/oxidecomputer/oana?tab=readme-ov-file#service-processor-udp-ports)

drv/stm32h7-fmc-demo-server/src/main.rs
```diff
@@ -216,8 +232,17 @@ fn main() -> ! {
     // space, which is fine even if it aliases (which it doesn't).
     let fmc = unsafe { &*device::FMC::ptr() };
 
+    let net = Net::from(NET.get_task_id());
+
     // Fire up a server.
-    let mut server = ServerImpl { fmc };
+    let rx_packet = RX_PACKET.claim();
```
mkeeter, line 238: What's the intuition for when we should use `ClaimOnceCell` versus `mutable_statics!`?

drv/stm32h7-fmc-demo-server/src/main.rs
```diff
@@ -226,15 +251,57 @@ fn main() -> ! {
...
+                                count!(Event::RequestRejected);
+                                self.tx_packet[0] = e as u8;
+                                meta.size = 1;
+                            }
+                        }
+                        let response = &self.tx_packet[..meta.size as usize];
+                        // We're going to send best-effort because the only
+                        // thing that can really go wrong here is filling our
+                        // own queue, which, well, send slower.
+                        let _ = self.net.send_packet(SOCKET, meta, response);
```
mkeeter, line 291: This seems like it could be confusing for users, who will see packets being dropped if they send too fast.  Is there a downside to looping here?

drv/stm32h7-fmc-demo-server/src/main.rs
```diff
@@ -392,6 +459,158 @@ impl idl::InOrderFmcDemoImpl for ServerImpl {
...
+    Ok(first)
+}
+
+fn write_byte(byte: u8, buf: &mut &mut [u8]) -> Result<(), NetworkError> {
+    let tbuf = take(buf);
+    let (first, rest) =
+        tbuf.split_first_mut().ok_or(NetworkError::ResponseTooBig)?;
+    *first = byte;
+    *buf = rest;
+    Ok(())
```
mkeeter, line 589: What about delegating to `write_chunk`, e.g.
```suggestion
    write_chunk([byte], buf)
```

drv/stm32h7-fmc-demo-server/src/main.rs
```diff
@@ -392,6 +459,158 @@ impl idl::InOrderFmcDemoImpl for ServerImpl {
...
+        }
+    }
+
+    Ok(orig_response_len - response.len())
+}
+
+fn read_byte(buf: &mut &[u8]) -> Result<u8, NetworkError> {
+    let (&first, rest) = buf.split_first().ok_or(NetworkError::Truncated)?;
+    *buf = rest;
+    Ok(first)
```
mkeeter, line 571:
```suggestion
    Ok(read_chunk::<1>(buf)?[0])
```

drv/stm32h7-fmc-demo-server/README.md
```diff
@@ -2,3 +2,37 @@
...
+
+Following this is a series of arguments to command 0, which describe a sequence
+of memory operations. Arguments are simply concatenated together. Each starts
+with an identifier byte, and some are followed by additional bytes. Any sequence
+of the following is acceptable:
+
+- `address` (0) followed by a little-endian 32-bit address. This loads the
+  address into the internal address register used by subsequent peeks and pokes.
+- `peek` (1/2/3/4) reads an 8/16/32/64-bit (respectively) value from memory at
+  the current addressand concatenates it (in little endian order) onto the
```
taspelund, line 26: missing a space: "addressand"

drv/stm32h7-fmc-demo-server/src/main.rs
cbiffle: `mutable_statics!` leaves its backing memory uninitialized, so it can be filled in with things that aren't legal consts. This may also provide a benefit in cases where the initializer isn't zeroes but the initialization code is relatively simple (because a non-zeroes const initializer ends up fully expanded in the data image).

 `ClaimOnceCell` requires a legal const in its initializer but has lower complexity (no macros) and runtime cost (no initialization).

drv/stm32h7-fmc-demo-server/src/main.rs
cbiffle: FWIW, this isn't what will cause users to see packets getting dropped. Because the incoming and outgoing queues are the same size, sending too fast is far more likely to drop on receipt than response.

That being said, I can add the loop here. I was being lazy and felt the indentation was getting a little deep. :-)

drv/stm32h7-fmc-demo-server/src/main.rs
cbiffle: Huh, why not

drv/stm32h7-fmc-demo-server/src/main.rs
cbiffle: A good suggestion in general -- I like early returns -- but in case this notification handler should grow a second notification in the future, I'd rather follow this pattern, so that the second bit we need to test causes less churn.

app/grapefruit/app.toml
cbiffle: Done! Also added gimlet-inspector.

drv/stm32h7-fmc-demo-server/src/main.rs
cbiffle: I found a different phrasing of this that avoids having quite so much punctuation.

`let [byte] = read_chunk(buf)?; Ok(byte)`

mkeeter (approved):

drv/stm32h7-fmc-demo-server/src/main.rs
cbiffle: So this has caused me to realize that there's a small subtle liveness bug in our normal retry loop pattern that is unlikely to occur in practice, but is theoretically possible:
- A task attempts to send
- It finds its queue full
- It begins waiting for the net stack's obligatory "your queue is no longer full" notification
- The net stack _crashes_ and is restarted
- The net stack observes all queues as having space available and goes on processing packets.

In this case, the transmitting task will sit and wait until it receives a net notification, which in this case is likely to be the next packet addressed to one of its sockets.

I have recorded this as #1844.

drv/stm32h7-fmc-demo-server/README.md
cbiffle: Fixed! Thanks!

PR #1845, opened 2024-08-14
Add APIs to check if the RoT is dev or release

mkeeter:
We want to make policy decisions based on whether a system is in development or release mode.  For example, development machines may have more lax policy on tech port unlocking; see [RFD 492](https://rfd.shared.oxide.computer/rfd/506) and [506](https://rfd.shared.oxide.computer/rfd/506)

This PR adds a `SpRot` message and Idol function to perform a "development or release" check.  The policy decision is made on the RoT, which returns either `Development` or `Release`.

The RoT uses two pieces of information to make this policy decision.

First, it checks which keys are active in the Root of Trust's CFPA region, which has four slots:

- Release is `[valid, valid, invalid, invalid]`
- Development is `[revoked, revoked, valid, valid]`

The LPC55's firmware only lets keys go from `invalid → valid → revoked`; if any of slots 2/3 are `valid`, then we know that the machine is in Development mode.

However, we also distribute images with a CFPA that only enables key 0 (`[valid, invalid, invalid, invalid]`).  These images use the Bartholomew certificate ([`fake_certs` in this repository](https://github.com/oxidecomputer/hubris/tree/master/support/fake_certs)), which is not _actually_ secret and is used for convenience when developing.

To detect this case, we check whether the CMPA is unlocked.  If the CMPA is unlocked, then we treat the system as in Development.

### DIFF
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -7,11 +7,11 @@ use attest_api::Attest;
 use crc::{Crc, CRC_32_CKSUM};
 use drv_lpc55_update_api::{RotComponent, RotPage, SlotId, Update};
 use drv_sprot_api::{
-    AttestReq, AttestRsp, CabooseReq, CabooseRsp, DumpReq, PolicyDevOrRelease,
-    PolicyReq, PolicyRsp, ReqBody, Request, Response, RotIoStats, RotPageRsp,
-    RotState, RotStatus, RspBody, SprocketsError, SprotError,
-    SprotProtocolError, SwdReq, UpdateReq, UpdateRsp, CURRENT_VERSION,
-    MIN_VERSION, REQUEST_BUF_SIZE, RESPONSE_BUF_SIZE,
+    AttestReq, AttestRsp, CabooseReq, CabooseRsp, DumpReq, ReqBody, Request,
+    Response, RotIoStats, RotPageRsp, RotState, RotStatus, RspBody,
+    SprocketsError, SprotError, SprotProtocolError, StateDevOrRelease,
+    StateError, StateReq, StateRsp, SwdReq, UpdateReq, UpdateRsp,
+    CURRENT_VERSION, MIN_VERSION, REQUEST_BUF_SIZE, RESPONSE_BUF_SIZE,
 };
 use lpc55_romapi::bootrom;
 use ringbuf::ringbuf_entry_root as ringbuf_entry;
@@ -568,37 +568,45 @@ impl<'a> Handler {
                 )?;
                 Ok((RspBody::Ok, None))
             }
-            ReqBody::Policy(PolicyReq::DevOrRelease) => {
-                let mut buf = [0u8; 512];
+            ReqBody::State(StateReq::DevOrRelease) => {
+                let out = self.state_dev_or_release();
 
-                // If the SHA-256 Digest is zeros, the CMPA is unlocked
-                self.update.read_rot_page(RotPage::Cmpa, &mut buf)?;
-                let unlocked = buf[480..].iter().all(|b| *b == 0);
-
-                // Otherwise, check which root keys are enabled
-                let dev = if unlocked {
-                    true
-                } else {
-                    self.update.read_rot_page(RotPage::CfpaActive, &mut buf)?;
-                    // Look at the ROTKH_REVOKE byte
-                    let revoke = buf[24];
-                    // Check if any of the development keys (slots 2 and 3) are
-                    // in the Valid state (0b01)
-                    [2, 3]
-                        .iter()
-                        .any(|slot| (revoke >> (slot * 2)) & 0b11 == 0b01)
-                };
-
-                Ok((
-                    RspBody::Policy(PolicyRsp::DevOrRelease(if dev {
-                        PolicyDevOrRelease::Development
-                    } else {
-                        PolicyDevOrRelease::Release
-                    })),
-                    None,
-                ))
+                Ok((RspBody::State(out.map(StateRsp::DevOrRelease)), None))
             }
         }
+    }
+
+    fn state_dev_or_release(
+        &mut self,
+    ) -> Result<StateDevOrRelease, StateError> {
+        let mut buf = [0u8; 512];
+
+        // If the SHA-256 Digest is zeros, the CMPA is unlocked
+        self.update
+            .read_rot_page(RotPage::Cmpa, &mut buf)
+            .map_err(StateError::ReadCmpa)?;
+        let unlocked = buf[480..].iter().all(|b| *b == 0);
+
+        // Otherwise, check which root keys are enabled
+        let dev = if unlocked {
+            true
+        } else {
+            self.update
+                .read_rot_page(RotPage::CfpaActive, &mut buf)
+                .map_err(StateError::ReadCfpa)?;
+            // Look at the ROTKH_REVOKE byte
+            let revoke = buf[24];
+            // Check if any of the development keys (slots 2 and 3) are
+            // in the Valid state (0b01)
+            [2, 3]
+                .iter()
+                .any(|slot| (revoke >> (slot * 2)) & 0b11 == 0b01)
+        };
+        Ok(if dev {
+            StateDevOrRelease::Development
+        } else {
+            StateDevOrRelease::Release
+        })
     }
 }
 
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -7,11 +7,11 @@ use attest_api::Attest;
 use crc::{Crc, CRC_32_CKSUM};
 use drv_lpc55_update_api::{RotComponent, RotPage, SlotId, Update};
 use drv_sprot_api::{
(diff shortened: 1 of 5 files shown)


drv/sprot-api/src/lib.rs
```diff
@@ -499,6 +510,9 @@ pub enum RspBody {
     Attest(Result<AttestRsp, AttestError>),
 
     Page(Result<RotPageRsp, UpdateError>),
+
+    // Added in sprot protocol version 6
+    Policy(PolicyRsp),
```
mkeeter, line 515: One thing I'm not sure of: why do some messages use the `SprotError` as the return code, but others return a nested `Result<ThingyRsp, ThingyError>`?

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -1341,6 +1341,24 @@ impl<S: SpiServer> idl::InOrderSpRotImpl for ServerImpl<S> {
...
+        &mut self,
+        _msg: &userlib::RecvMessage,
+    ) -> Result<
+        drv_sprot_api::PolicyDevOrRelease,
+        idol_runtime::RequestError<SprotError>,
+    > {
+        let body = ReqBody::Policy(PolicyReq::DevOrRelease);
+        let tx_size = Request::pack(&body, self.tx_buf);
+        let rsp =
+            self.do_send_recv_retries(tx_size, TIMEOUT_LONG, DEFAULT_ATTEMPTS)?;
```
aapoalas, line 1355: **question:** Shouldn't the timeout here (probably) be `TIMEOUT_QUICK`? I'd imagine the reading of the data doesn't take that long. If memory serves, the `component_switch_default_image` function uses `TIMEOUT_LONG` because it may include rebooting the device (or something like that).

aapoalas (commented):
Peanut gallery here again.

drv/stm32h7-sprot-server/src/main.rs
mkeeter: Fixed in f10d7be9565c

drv/sprot-api/src/lib.rs
andrewjstone: IIRC, for some errors, like `DumpError` in particular, we don't want to expose them up through the control plane agent. But also, it doesn't really make sense to embed unrelated errors inside `SprotError`. This was likely a mistake I made early on that we are living with. I'd stick with embedding a result for new messages. The result can also be embedded in the singleton type and parsed out later, so no need to add one here since it's not immediately needed.

drv/sprot-api/src/lib.rs
```diff
@@ -349,6 +349,8 @@ pub enum ReqBody {
     RotPage { page: RotPage },
     // Added in sprot protocol version 5
     Swd(SwdReq),
+    // Added in sprot protocol version 6
```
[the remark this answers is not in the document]
mkeeter: Renamed to `State` (there's some overlap with `RotState`, but I still think it's a better name)

drv/sprot-api/src/lib.rs
mkeeter: Alright, I switched to a dedicated `Result<StateRsp, StateError>`

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -568,6 +568,36 @@ impl<'a> Handler {
...
+                let unlocked = buf[480..].iter().all(|b| *b == 0);
+
+                // Otherwise, check which root keys are enabled
+                let dev = if unlocked {
+                    true
+                } else {
+                    self.update.read_rot_page(RotPage::CfpaActive, &mut buf)?;
+                    // Look at the ROTKH_REVOKE byte
+                    let revoke = buf[24];
+                    // Check if any of the development keys (slots 2 and 3) are
```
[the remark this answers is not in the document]
mkeeter: I think it's impossible to add new root keys with a locked CMPA, because the 4 key hashes are stored in CMPA (which is locked).

CFPA lets us switch key _state_ for each of the 4 slots, which is a forward-only transition from invalid → valid → revoked.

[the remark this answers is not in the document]
andrewjstone: Makes sense. Thanks.

andrewjstone (approved):
Looks great. Thanks @mkeeter

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -568,8 +568,46 @@ impl<'a> Handler {
...
+            true
+        } else {
+            self.update
+                .read_rot_page(RotPage::CfpaActive, &mut buf)
+                .map_err(StateError::ReadCfpa)?;
+            // Look at the ROTKH_REVOKE byte
+            let revoke = buf[24];
+            // Check if any of the development keys (slots 2 and 3) are
+            // in the Valid state (0b01)
+            [2, 3]
```
cbiffle, line 601: huh, cute

cbiffle (approved):

drv/lpc55-sprot-server/src/handler.rs
lzrd: Do we have symbolic names for those slots anywhere?

mkeeter: Not that I can find, but I think the comment above is sufficient to explain their meaning.

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -568,8 +568,48 @@ impl<'a> Handler {
...
+
+        // If the SHA-256 Digest is zeros, the CMPA is unlocked.  The SHA-256
+        // digest is located in the last 4 words of the CMPA.
+        self.update
+            .read_rot_page(RotPage::Cmpa, &mut buf)
+            .map_err(StateError::ReadCmpa)?;
+        let unlocked = buf[CMPA_SIZE - 32..].iter().all(|b| *b == 0);
+
+        // Otherwise, check which root keys are enabled
+        let dev = if unlocked {
```
flihp, line 593: The enum of RoT lifecycle states in RFD 286 identifies this configuration as `Unprogrammed`. AFAIK this work is motivated by a need to implement some policy in the SP and that policy seems to have leaked into this interface?
Regardless, I'd prefer that this interface expose the platform lifecycle states per the RFD and we can have the policy layer treat the `Unprogrammed` and `Development` states as equivalent or whatever makes sense for the policy you're trying to enforce.

drv/sprot-api/src/lib.rs
```diff
@@ -499,6 +510,9 @@ pub enum RspBody {
     Attest(Result<AttestRsp, AttestError>),
 
     Page(Result<RotPageRsp, UpdateError>),
+
+    // Added in sprot protocol version 6
+    State(Result<StateRsp, StateError>),
```
flihp, line 515: We call these states `lifecycle states` in rfd 286. If it's not too many characters that help differentiate it from the `RotState` variant.

flihp (changes requested):

drv/lpc55-sprot-server/src/handler.rs
mkeeter: Switched to `LifecycleState` in 370c89cc

flihp (approved):
Thanks!

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -568,7 +568,62 @@ impl<'a> Handler {
...
+        let mut buf = [0u8; CMPA_SIZE];
+
+        // If the SHA-256 Digest is zeros, the CMPA is unlocked.  The SHA-256
+        // digest is located in the last 4 words of the CMPA.
+        self.update
+            .read_rot_page(RotPage::Cmpa, &mut buf)
+            .map_err(StateError::ReadCmpa)?;
+        if buf[CMPA_SIZE - 32..].iter().all(|b| *b == 0) {
+            // We use Unprogrammed as a catchall if the CMPA is unlocked
+            return Ok(LifecycleState::Unprogrammed);
```
labbott, line 590: No need to do this here but I think we may want to consider abstracting this out for future use

drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -568,7 +568,62 @@ impl<'a> Handler {
...
+            | [_, S::Enabled, S::Invalid, S::Invalid] => {
+                LifecycleState::Release
+            }
+            [S::Revoked, S::Revoked, S::Enabled, _]
+            | [S::Revoked, S::Revoked, _, S::Enabled] => {
+                LifecycleState::Development
+            }
+            [S::Revoked, S::Revoked, S::Revoked, S::Revoked] => {
+                LifecycleState::EndOfLife
+            }
```
labbott, line 623: This should be an impossible state. If we've revoked all keys we can't boot any secure boot code and we're locked per the above check so we can't even run this code.

labbott: I think `unreachable!` is a better idea but if that would add in surprise panics we can go with a comment

mkeeter: (we talked in chat, and are going with the comment to avoid a panic path)

labbott (approved):
LGTM

PR #1847, opened 2024-08-14
net: wake all sockets on restart.

### LINKED ISSUE #1844: net: on startup, treat all tasks as waiting_to_send to avoid liveness bug
We currently have a potential (albeit unlikely) liveness bug in the net stack task protocol. It goes like this:

1. A task attempts to send a packet.
2. The net stack reports that the socket's outgoing queue is full and the task should try again later.
3. The task begins waiting for the queue-available notification.
4. The net stack _crashes and restarts,_ losing its table of which tasks are owed a notification.
5. It sees all queues as having available space and goes on about its business.

The task will only be woken up when it next receives a packet.

After discussion with @mkeeter I think the simplest way of addressing this is to change the protocol slightly. Specifically:

**On `net` restart, all tasks should be treated as `waiting_to_send`.**

This will cause `net` to distribute a notification to all sockets on restart. Since we treat notifications as potentially spurious, the worst that this will do is cause the clients to wake up briefly and do unnecessary checking before going back to sleep. But, if a task was waiting for available TX queue space, this would cause it to discover its queues empty and make forward progress again.

cbiffle:
This avoids a potential liveness issue described in #1844, wherein a task waiting for tx queue space would not notice net restarting.

Fixes #1844.

### DIFF
--- a/task/net/src/main.rs
+++ b/task/net/src/main.rs
@@ -242,6 +242,14 @@ fn main() -> ! {
         );
     }
 
+    // Ensure that sockets are woken at least once at startup, so that anyone
+    // who was waiting to hear back on their TX queue becoming non-full will
+    // snap out of it.
+    //
+    // This only works because we've set waiting_to_send to true for all sockets
+    // above.
+    server.wake_sockets();
+
     // Go!
     loop {
         count!(Event::Iter);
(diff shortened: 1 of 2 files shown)


cbiffle:
Tested as working on Grapefruit, at least.

mkeeter (approved):

PR #1851, opened 2024-08-20
MVP implementation of `HostFlash` for Cosmo

mkeeter:
This PR implements the `HostFlash` API for the Cosmo / Grapefruit flash IC, mediated by an FPGA.

`gimlet-hf` is renamed to `hf` in a few places, because we're now using the same API for both Gimlet and Cosmo systems.  There are no changes to the API itself.

`drv-fmc-nor-flash` is renamed to `drv-cosmo-hf` for symmetry with Gimlet.  It is split into `main.rs` (implementing the actual chip wrangling) and `hf.rs` (implementing the `HostFlash` Idol API).

Many functions are left as stubs, to be implemented in later PRs.  However, this PR is sufficient to program and verify the flash chip with `humility qspi`!


out.raw
nathanaelhuffman, line 1: I assume this was added by mistake?

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,209 @@
...
+                } else {
+                    end
+                }
+            }
+            None => {
+                return Err(HfError::HashBadRange.into());
+            }
+        };
+        // If we knew the flash part size, we'd check against those limits.
+        let mut buf = [0u8; PAGE_SIZE_BYTES];
```
aapoalas, line 160: **question:** Is the idea that `being` is aligned at the start of a page and this can then read the flash part page-by-page (not counting the last, possibly not full page)? Is `start` guaranteed to be page-aligned?

Not sure if it would really make much of a difference, but you could split the range into full pages (with or without a not-full prefix, depending on if `start` is aligned or not) plus a not-necessarily-full suffix. That would mean you don't need to calculate `size` for the full pages.

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,294 @@
...
+            if i >= 32 {
+                sleep_for(1);
+            }
+        }
+    }
+
+    /// Wait until the SPI flash is idle
+    fn wait_flash_busy(&self, t: Trace) {
+        // Wait for the busy flag to be unset
+        while (self.read_flash_status() & 1) != 0 {
```
aapoalas, line 168: **nitpick:** Other loops do `let status = act(); if act & compare == 0 { break; }`, but this uses a `while (act & compare != 0`. Oh the horror! :)

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,294 @@
...
+    fn flash_write_enable(&self) {
+        self.write_reg(reg::DATA_BYTES, 0);
+        self.write_reg(reg::ADDR, 0);
+        self.write_reg(reg::DUMMY_CYCLES, 0);
+        self.write_reg(reg::INSTR, instr::WRITE_ENABLE);
+        self.wait_fpga_busy();
+    }
+
+    fn clear_fifos(&self) {
+        self.write_reg(reg::SPICR, 0x8080);
```
aapoalas, line 193: **question:** Other operations generally need to wait for the FPGA to become idle after acting; that's presumably not the case here?

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub const SECTOR_ERASE: u32 = 0x20;
+    pub const READ_JEDEC_ID: u32 = 0x9f;
+    pub const BLOCK_ERASE_64KB: u32 = 0xd8;
+    pub const BLOCK_ERASE_64KB_4B: u32 = 0xdc;
+    pub const QUAD_INPUT_PAGE_PROGRAM: u32 = 0x32;
+    pub const QUAD_INPUT_PAGE_PROGRAM_4B: u32 = 0x34;
+}
+
+impl ServerImpl {
+    fn flash_read_id(&self) -> [u8; 20] {
```
aapoalas, line 111: **praise:** For an outsider like me, it's really cool seeing this lowest-of-the-low level register work happening here.

**suggestion:** A mountain of comments wouldn't maybe go amiss here and would be a terrific pleasure to read :)

out.raw
mkeeter: Yup, fixed!

aapoalas (commented):
Peanut gallery checking in.

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    fn page_program(
+        &mut self,
+        _: &RecvMessage,
+        addr: u32,
+        protect: HfProtectMode,
+        data: LenLimit<Leased<R, [u8]>, PAGE_SIZE_BYTES>,
+    ) -> Result<(), RequestError<HfError>> {
+        if addr as usize / SECTOR_SIZE_BYTES == 0
```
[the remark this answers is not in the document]
cbiffle: I am still confused by this line tbh. Is this `addr < SECTOR_SIZE_BYTES`? I'd suggest writing it that way (or using a range if you prefer) so that it doesn't look like a mistake.

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,209 @@
...
+        _: &RecvMessage,
+        addr: u32,
+        len: u32,
+    ) -> Result<[u8; SHA256_SZ], RequestError<HfError>> {
+        let hash_driver = drv_hash_api::Hash::from(HASH.get_task_id());
+        if hash_driver.init_sha256().is_err() {
+            return Err(HfError::HashError.into());
+        }
+        let begin = addr as usize;
+        // TODO: Begin may be an address beyond physical end of
```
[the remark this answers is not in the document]
cbiffle: I think the checked add in this case is only checking for _integer wrap_ -- the flash part is likely to be much smaller than `2**32` so the flash addressing will wrap before the integer will.

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,209 @@
...
+        // flash part and may wrap around.
+        let end = match begin.checked_add(len as usize) {
```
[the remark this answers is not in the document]
cbiffle: For better or worse, a lot of existing Hubris code assumes a 32-bit platform, yeah.

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,209 @@
...
+                // TODO: End may be beyond physical end of flash part.
+                //       Use that limit rather than maximum 4-byte address.
+                if end > u32::MAX as usize {
+                    return Err(HfError::HashBadRange.into());
+                } else {
+                    end
+                }
+            }
+            None => {
+                return Err(HfError::HashBadRange.into());
```
cbiffle, line 156: Tip: I would find this clearer, personally, if it were expressed as

```rust
let end = begin.checked_add(len as usize)
    .ok_or(HfError::HashBadRange)?;
```

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,209 @@
...
+            }
+        };
+        // If we knew the flash part size, we'd check against those limits.
+        let mut buf = [0u8; PAGE_SIZE_BYTES];
+        for addr in (begin..end).step_by(buf.len()) {
+            let size = (end - addr).min(buf.len());
+            self.flash_read(addr as u32, &mut buf[..size]).unwrap();
```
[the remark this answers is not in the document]
cbiffle: Tracing the data flow, this unwrap is a client panic multiplier.

`flash_read` only fails if the client-supplied borrow writer returns an error, which happens if the buffer provided is e.g. too small or otherwise violates some assumption. The length of the buffer is checked, so the only practical way the buffer operation could fail is if the client were restarted during the IPC.

This is generally hard to achieve, but can happen in practice through supervisor action, and there's no reason to kill the flash driver in response to it. This unwrap should be removed.

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,209 @@
...
+        }
+        self.flash_sector_erase(addr);
+        Ok(())
+    }
+
+    fn get_mux(
+        &mut self,
+        _: &RecvMessage,
+    ) -> Result<HfMuxState, RequestError<HfError>> {
+        todo!()
```
cbiffle, line 105: There's a lot of TODOs here providing remote panic opportunities. Are these operations going to make sense for Cosmo? I ask because, if not, we should have two different idl files, since we can't cfg things in IDL files at the moment.

drv/cosmo-hf/src/hf.rs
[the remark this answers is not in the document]
mkeeter: `flash_read` is generic across `idol_runtime::BufWriter`, so this call is writing to a local buffer (`&mut buf[..size]`), **not** a lease.  I believe this means that `BufWriter` can't fail, so this is actually fine (but agreed that a comment would be good here).

drv/cosmo-hf/src/hf.rs
cbiffle: I think the comment on line 159 is slightly misplaced. What this code appears to be doing is just ensuring that reads are under 256 bytes in size by reusing a buffer; it doesn't appear to be sensitive to alignment, afaict.

drv/cosmo-hf/src/hf.rs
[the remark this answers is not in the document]
cbiffle: I see. Yeah, a comment would be good, and on the off chance that this generates a panic site, please use `unwrap_lite`.

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,305 @@
...
+        self.write_reg(reg::DATA_BYTES, 0);
+        self.write_reg(reg::ADDR, addr);
+        self.write_reg(reg::DUMMY_CYCLES, 0);
+        self.write_reg(reg::INSTR, instr::BLOCK_ERASE_64KB_4B);
+
+        // Wait for the busy flag to be unset
+        self.wait_flash_busy(Trace::SectorEraseBusy);
+    }
+
+    fn flash_read<'a, B: idol_runtime::BufWriter<'a>>(
```
cbiffle, line 207: To avoid generating multiple copies of this function, I'd strongly suggest switching this to use `&mut dyn BufWriter` instead of a type parameter.

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,305 @@
...
+        self.write_reg(reg::DATA_BYTES, len as u32);
+        self.write_reg(reg::ADDR, addr);
+        self.write_reg(reg::DUMMY_CYCLES, 0);
+        for i in 0..len.div_ceil(4) {
+            let mut v = [0u8; 4];
+            for (j, byte) in v.iter_mut().enumerate() {
+                let k = i * 4 + j;
+                if k < len {
+                    let Some(d) = data.read() else {
+                        return Err(());
```
cbiffle, line 251: Should this use an error type? I'm surprised Clippy isn't mad about this tbh.

drv/cosmo-hf/src/hf.rs
mkeeter: From talking with Nathanael, we do think that the existing API makes sense for Cosmo!

There are PRs coming down the line that implement these functions, e.g. #1853, but I didn't want to do too much in the MVP PR.

cbiffle (commented):
For me the highest priority things to look at here are:
1. Switch the generic flash read functions to use `dyn` so we don't generate multiple copies of them in flash. (The traits appear object-safe.)
2. The use of `Result<(), ()>` seems suspicious, consider whether the error type could be something more useful. (I think this might have been inherited from `BufWriter`, but I think it's arguably wrong there too; no reason to propagate the wrongness when your IPC error types are right there.)
3. Comment that unwrap please, it's subtle
4. Make sure those todo!s will get todone, otherwise the operations shouldn't be in the IPC protocol (and we need to fork it).
5. The `flash_foo` functions toward the end could use doc comments explaining any assumptions and failure cases.

drv/cosmo-hf/src/hf.rs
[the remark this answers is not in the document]
mkeeter: Yes, `addr < SECTOR_SIZE_BYTES` is more clear (and I'll add a comment as well)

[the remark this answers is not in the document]
aapoalas: Yeah, I think that's what it is in the end: I think clippy would even complain about the range comparison because the lower end is guaranteed by the type.

Later there's a second line where we compare the division result to 1, so there we'd need to use `BYTES..BYTES*2` range inclusion check. But you're also absolutely right: A single less-than comparison is all we need here. (I think.)

drv/cosmo-hf/src/hf.rs
mkeeter: Yup, that comment isn't quite relevant here (we're limited by FPGA buffers, not the flash chip itself).  Deleted!

drv/cosmo-hf/src/main.rs
mkeeter: There's a reason for the difference: the other loops actually log the status in the ringbuf, so we need it in a variable.

drv/cosmo-hf/src/main.rs
aapoalas: Mildly off-topic, but I've had good(?) use of `Result<T, ()>` in `TryFrom<ParentEnum> for SubEnum` places. In that case the error case is just "it wasn't in the subgroup you checked for". So I'm not _that_ surprised clippy doesn't complain about this: There are some minor use cases where it's not an entirely bad choice.

drv/cosmo-hf/src/main.rs
nathanaelhuffman: There is no wait required here. The FPGA FIFO clears take single-digit clock cycles to complete, as the reset may have to cross clock domains to clear read/write pointers, but the SP is hanging off the FMC interface and can't even finish executing another read/write transaction to the FPGA before the FIFO clear has completed so this can be treated as if it happens instantly.

drv/cosmo-hf/src/main.rs
aapoalas: Oh, indeed. My bad.

drv/cosmo-hf/src/main.rs
mkeeter: done!

drv/cosmo-hf/src/hf.rs
mkeeter: nice, done!

drv/cosmo-hf/src/hf.rs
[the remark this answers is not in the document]
mkeeter: commented and `unwrap_lite`'d

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+enum Trace {
+    #[count(skip)]
+    None,
+
+    FpgaBusy(u32),
+    SectorEraseBusy,
+    WriteBusy,
+
+    HashUpdateError(drv_hash_api::HashError),
+    HashFinalizeError(drv_hash_api::HashError),
```
hawkw, line 32: do we care about having counters for the individual `HashError` variants here or not? i'll note that we probably generate IPC counters for them already, so may not need to count them here as well...

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+use ringbuf::{counted_ringbuf, ringbuf_entry};
+
+mod hf; // implementation of `HostFlash` API
+
+#[derive(Debug, Clone, Copy, PartialEq, counters::Count)]
+enum Trace {
+    #[count(skip)]
+    None,
+
+    FpgaBusy(u32),
```
hawkw, line 27: What does the `u32` mean here? It might be nice if this was a struct-like variant with a named field, so that someone inspecting the ring buffer in `humility` immediately knows what the integer refers to?

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+///
+/// **Note:** the datasheet refers to a "sector" as a 4K block, but also
+/// supports 64K block erases, so we call the latter a sector to match the
+/// behavior of the Gimlet host flash driver.
+pub const SECTOR_SIZE_BYTES: u32 = 65_536;
+
+#[export_name = "main"]
+fn main() -> ! {
+    // Wait for the FMC to be configured
+    userlib::hl::sleep_for(1000);
```
hawkw, line 60: how do we know that it will take less than 1000 ticks to do this? What happens if we sleep and the FMC still hasn't been configured?

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+            ringbuf_entry!(Trace::FpgaBusy(status));
+            sleep_for(1);
+        }
+    }
+
+    /// Wait until a word is available in the FPGA's RX buffer
+    fn wait_fpga_rx(&self) {
+        for i in 0.. {
+            let status = self.read_reg(reg::SPISR);
+            // Wait for the
```
hawkw, line 152: Wait for the...?

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+                self.wait_fpga_rx();
+                let v = self.read_reg(reg::RX_FIFO);
+                let v = v.to_le_bytes();
+                for (j, byte) in v.iter().enumerate() {
+                    let k = i * 4 + j;
+                    if k < len {
+                        dest.write(*byte)?;
+                    }
+                }
+            }
```
hawkw, line 237: wondered a bit if this structure, which is repeated a few times, is easily generalizable to both the `flash_read` and `flash_write` methods, but honestly...trying to factor it out is probably more trouble than it's worth...

mkeeter:
> For me the highest priority things to look at here are:
> 
> 1. Switch the generic flash read functions to use `dyn` so we don't generate multiple copies of them in flash. (The traits appear object-safe.)
> 2. The use of `Result<(), ()>` seems suspicious, consider whether the error type could be something more useful. (I think this might have been inherited from `BufWriter`, but I think it's arguably wrong there too; no reason to propagate the wrongness when your IPC error types are right there.)
> 3. Comment that unwrap please, it's subtle
> 4. Make sure those todo!s will get todone, otherwise the operations shouldn't be in the IPC protocol (and we need to fork it).
> 
> 5. The `flash_foo` functions toward the end could use doc comments explaining any assumptions and failure cases.

This is all done except 2, because I opened https://github.com/oxidecomputer/idolatry/pull/55 to fix it upstream.

drv/cosmo-hf/src/main.rs
mkeeter: Renamed to `spi_sr`

drv/cosmo-hf/src/main.rs
mkeeter: BALEETED

drv/cosmo-hf/src/main.rs
mkeeter: > What happens if we sleep and the FMC still hasn't been configured?

Panic! at the ID check

(I added a basic `ping` function to the sequencer API, so `cosmo-hf` now waits for that to return before starting up)

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,206 @@
...
+        &mut self,
+        _: &RecvMessage,
+        addr: u32,
+        protect: HfProtectMode,
+    ) -> Result<(), RequestError<HfError>> {
+        if addr < SECTOR_SIZE_BYTES
+            && !matches!(protect, HfProtectMode::AllowModificationsToSector0)
+        {
+            return Err(HfError::Sector0IsReserved.into());
+        }
```
hawkw, line 101: nit, take it or leave it: we could maybe have a function like:
```rust
fn check_addr_writable(addr: u32, protect: HfProtectMode) -> Result<(), HfError> {
    if addr < SECTOR_SIZE_BYTES && 
        !matches!(protect, HfProtectMode::AllowModificationsToSector0)
    {
          return Err(HfError::Sector0IsReserved);
    }
    Ok(())
}
```
since we seem to repeat this in a few places. potentially, factoring it out it would mean that a future change adding other checks against `HfProtectMode`s wouldn't have to add the requisite checks in every function that currently checks if sector 0 is writable? 

not a big deal either way though.

drv/cosmo-hf/src/hf.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    fn hash(
+        &mut self,
+        _: &RecvMessage,
+        addr: u32,
+        len: u32,
+    ) -> Result<[u8; SHA256_SZ], RequestError<HfError>> {
+        let hash_driver = drv_hash_api::Hash::from(HASH.get_task_id());
+        if hash_driver.init_sha256().is_err() {
+            return Err(HfError::HashError.into());
```
hawkw, line 144: should failures to initialize the hash driver also be ringbuf'd, so that they can be distinguished from `Trace::HashUpdateError`?

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,312 @@
...
+/// **Note:** the datasheet refers to a "sector" as a 4K block, but also
+/// supports 64K block erases, so we call the latter a sector to match the
+/// behavior of the Gimlet host flash driver.
+pub const SECTOR_SIZE_BYTES: u32 = 65_536;
+
+#[export_name = "main"]
+fn main() -> ! {
+    // Wait for the FPGA to be configured
+    let seq = drv_grapefruit_seq_api::Sequencer::from(SEQ.get_task_id());
+    seq.ping(); // waits until the sequencer has completed configuration
```
hawkw, line 64: Cute, I like this approach. :)

drv/cosmo-hf/src/hf.rs
mkeeter: Sure, done in f239a0a7b7cc

drv/cosmo-hf/src/main.rs
mkeeter: I don't feel strongly either way, but my vibe is that the IPC counters are sufficient for per-error counting.

hawkw: I think so too, let's stick with that.

drv/cosmo-hf/src/main.rs
```diff
@@ -0,0 +1,313 @@
...
+        }
+    }
+
+    /// Reads the STATUS1 register from flash
+    fn read_flash_status(&self) -> u8 {
+        self.clear_fifos();
+        self.write_reg(reg::DATA_BYTES, 1);
+        self.write_reg(reg::ADDR, 0);
+        self.write_reg(reg::DUMMY_CYCLES, 0);
+        self.write_reg(reg::INSTR, instr::READ_STATUS_1);
```
hawkw, line 189: there is a part of me that wonders if it's worth taking all this and putting it in a function that takes the number of bytes and the instruction value, but, honestly, I'm not sure if it would be worth doing, because you'd still have to provide the number of bytes, address, dummy cycles, and instructions to that function so i don't really think it would be abstracting over anything meaningful...and this feels more explicit, i guess?

drv/cosmo-hf/src/main.rs
hawkw: also, it could maybe be nice if there was a comment explaining that this kinda weird loop is because we read bytes out of the FIFO in 32-bit words, and depending on the length to read, we might overshoot the requisite length and have to bail early? i was able to figure it out fairly quickly but it might not be totally obvious to everyone...

hawkw (approved):
Looks good to me, no blockers! I agree with Cliff that it would be good to prioritize checking off the `todo!`s relatively soon, but everything looks good to me!

drv/cosmo-hf/src/main.rs
aapoalas: **opinion:** It's not too difficult code to read or understand, but a quick comment wouldn't be out of place. Words are still generally easier to understand. Also I personally love the comment-rich coding style I often see in Oxide's code, so more comments! :)

PR #1852, opened 2024-08-20
Split `ServerImpl` versus `FlashDriver`, add `FailServer`

mkeeter:
Staged on top of #1851

This PR splits the FMC NOR flash into `FlashDriver` (talking to the FPGA and flash) and `ServerImpl` (implementing the `HostFlash` API).  In addition, it adds a non-panicking failure path.  In this new failure path, there's a small Idol server which replies with an error to any message.

### DIFF
--- a/drv/cosmo-hf/src/main.rs
+++ b/drv/cosmo-hf/src/main.rs
@@ -74,7 +74,7 @@ fn main() -> ! {
 
     // Check the flash chip's ID against Table 7.3.1 in the datasheet
     let id = drv.flash_read_id();
-    if &id[0..3] != [0xef, 0x40, 0x21] {
+    if id[0..3] != [0xef, 0x40, 0x21] {
         fail(drv_hf_api::HfError::BadChipId);
     }
 
--- a/drv/cosmo-hf/src/main.rs
+++ b/drv/cosmo-hf/src/main.rs
@@ -65,18 +65,29 @@ fn main() -> ! {
     seq.ping(); // waits until the sequencer has completed configuration
 
     let id = unsafe { reg::BASE.read_volatile() };
-    assert_eq!(id, 0x1de);
+    if id != 0x1de {
+        fail(drv_hf_api::HfError::FpgaNotConfigured);
+    }
 
-    // Fire up a server.
-    let mut server = ServerImpl;
-    server.flash_set_quad_enable();
+    let drv = FlashDriver;
+    drv.flash_set_quad_enable();
+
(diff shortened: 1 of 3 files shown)


drv/cosmo-hf/src/main.rs
```diff
@@ -291,4 +302,13 @@ impl ServerImpl {
     }
 }
 
+/// Failure function, running an Idol response loop that always returns an error
+fn fail(err: drv_hf_api::HfError) {
```
aapoalas, line 306: **praise:** This seems like a great idea for (at least) debugging.

drv/cosmo-hf/src/main.rs
```diff
@@ -291,4 +302,13 @@ impl ServerImpl {
...
+    let mut buffer = [0; hf::idl::INCOMING_SIZE];
+    let mut server = hf::FailServer { err };
+    loop {
+        idol_runtime::dispatch(&mut buffer, &mut server);
+    }
+}
```
hawkw, line 312: This is a nice approach to having a driver go into "failure mode" when hardware can't be initialized, I wonder if we might want to do something similar in e.g. `gimlet_seq`, here:
https://github.com/oxidecomputer/hubris/blob/6990578c7f09bfa75784e6cfa923bbd1dace649f/drv/gimlet-seq-server/src/main.rs#L159-L176
rather than just waiting for a non-existent notification?

Obviously not something to do in this PR, but could be nice to do later.

drv/cosmo-hf/src/main.rs
```diff
@@ -60,18 +60,29 @@ fn main() -> ! {
     userlib::hl::sleep_for(1000);
 
     let id = unsafe { reg::BASE.read_volatile() };
-    assert_eq!(id, 0x1de);
+    if id != 0x1de {
+        fail(drv_hf_api::HfError::FpgaNotConfigured);
```
hawkw, line 64: What's the semantics of this error? Is it possible that the FPGA could later be configured, and the host-flash driver will need to be kicked back into actually working? Or is that a problem to solve later?

hawkw (approved):
This seems good to me! I had a couple smallish notes, but nothing major.

drv/cosmo-hf/src/main.rs
mkeeter: I'm putting that into the "problems to be solved later" bucket!

hawkw: Cool, carry on then!

PR #1853, opened 2024-08-20
Split Cosmo host flash into two virtual devices

mkeeter:
Staged on top of #1852

Unlike Gimlet, Cosmo / Grapefruit only has a single 128 MiB flash chip.  This PR allows us to treat it as two virtual 64 MiB devices, preserving the existing `HostFlash` API semantics.


drv/cosmo-hf/src/hf.rs
```diff
@@ -52,7 +68,15 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         if !matches!(protect, HfProtectMode::AllowModificationsToSector0) {
             return Err(HfError::Sector0IsReserved.into());
         }
-        todo!()
+        // Don't use the bulk erase command, because it will erase the entire
+        // chip.  Instead, use the sector erase to erase the currently-active
+        // virtual device.
+        for offset in 0..SLOT_SIZE_BYTES / SECTOR_SIZE_BYTES as u32 {
```
aapoalas, line 74: **suggestion:** Would this be clearer / more straightfoward as `(0..SLOT_SIZE_BYTES).iter().step_by(SECTOR_SIZE_BYTES)`?

drv/cosmo-hf/src/hf.rs
```diff
@@ -35,6 +39,17 @@ impl ServerImpl {
             Ok(())
         }
     }
+
+    fn flash_base(&self) -> u32 {
+        match self.dev {
+            HfDevSelect::Flash0 => 0,
+            HfDevSelect::Flash1 => SLOT_SIZE_BYTES,
```
hawkw, line 46: it occurs to me that we _could_ be cute about it and have an enum in this crate whose variants are represented by the values 0 and `SLOT_SIZE_BYTES`,  like
```rust
enum CosmoDevSelect {
    Flash0 = 0,
    Flash1 = SLOT_SIZE_BYTES,
}
```
and then just use `self.dev as u32`, and convert between the API `HfDevSelect` at the API boundary, so that individual reads/writes don't have to  do this conversion. But, honestly? This is simpler and that's probably not worth the trouble...

hawkw (approved):
This looks good to me! I was going to ask whether the logic for checking whether a write would touch sector 0 needed to be changed, but --- since we perform the check with the "virtual" address rather than the "physical" address calculated by `ServerImpl::flash_addr`, we don't have to worry about that.

PR #1855, opened 2024-08-20
chore: Cleanup userlib namespace imports

aapoalas:
Didn't quite have the time to get rid of all of these, but I removed a bunch of userlib namespace imports.

These were mentioned in a PR nearly half a year ago and I "helpfully" went through and pointed out those that were in the PR, so this is my atonement.

Each task is done in a separate commit here. All others except `drv-lpc55-spi-server` is tested to cause no build errors, or I think anyway. I'm randomly building tasks based on searches into the repository with the device name, and testing that they fail if I remove the `userlib::*` import entirely, and then add imports until build works.

### DIFF
--- a/drv/auxflash-api/src/lib.rs
+++ b/drv/auxflash-api/src/lib.rs
@@ -9,7 +9,7 @@
 use derive_idol_err::IdolError;
 use sha3::{Digest, Sha3_256};
 use tlvc::{TlvcRead, TlvcReader};
-use userlib::*;
+use userlib::{sys_send, FromPrimitive};
 use zerocopy::{AsBytes, FromBytes};
 
 pub use drv_qspi_api::{PAGE_SIZE_BYTES, SECTOR_SIZE_BYTES};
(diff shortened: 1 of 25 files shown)


cbiffle (approved):
Wow, thanks! This is great.

cbiffle:
(Whoever merges this, please squash before merge and take the PR text so there's more explanation of the purpose)

aapoalas:
I'm planning on removing the rest of them as well; given that these changes actually change nothing in the resulting code, it is fine to me merge this and make a second PR for the rest or keep this waiting until finish the cleanup. Either is okay.

cbiffle:
Alright, rebased and set to autosquash on success.

cbiffle:
> I'm planning on removing the rest of them as well; given that these changes actually change nothing in the resulting code, it is fine to me merge this and make a second PR for the rest or keep this waiting until finish the cleanup. Either is okay.

Naw, no reason to carry the PR longer than is necessary. If you want to clean up more things, that's great, but this PR is a marginal improvement and I'd like to go ahead and merge it.

aapoalas:
> Naw, no reason to carry the PR longer than is necessary. If you want to clean up more things, that's great, but this PR is a marginal improvement and I'd like to go ahead and merge it.

Sure thing!

PR #1857, opened 2024-08-23
chore: Cleanup remaining userlib namespace imports

aapoalas:
This PR ends my work of walking through the hubris code base, removing userlib namespace imports.

This time I did a search-and-replace on all of these to clear them out. I currently see some build issue about notifications with `app/lpc55xpresso/app-sprot.toml`; how the `notifications` should get built is not quite clear to me.

Another interesting file is `task/template/src/main.rs` where we have this comment:
```
// NOTE: you will probably want to remove this when you write your actual code;
// we need to import userlib to get this to compile, but it throws a warning
// because we're not actually using it yet!
#[allow(unused_imports)]
use userlib::{};
```
I'm not sure if I should add some imports here or not.


drv/stm32xx-i2c/src/lib.rs
```diff
@@ -923,7 +923,7 @@ impl I2cController<'_> {
 
         #[rustfmt::skip]
         i2c.oar1.modify(|_, w| { w
-            .oa1en().clear_bit()                    // own-address disable
```
aapoalas, line 926: **note:** Zed editor did these format fixes automatically, so they're kind of "slipping in" on the side.

aapoalas:
@cbiffle My apologies for the ping: Would you be amenable to taking a look at this? All tests have been fully green (so far anyway), only `app/lpc55xpresso/app-sprot.toml` has some sort of build failure about missing `notifications` that seemed like it might be a configuration failure in that board's configuration. It is not being tested by CI so I am not sure if that is even relevant.

aapoalas:
The remaining lpc55xpresso `app-sprot.toml` dist xtask passes now as well. I had to bump some stack sizes and copy SP_TO_ROT_JTAG_DETECT_L and ROT_TO_SP_RESET_L pin definitions from oxide-rot-1's `app.toml`. I have no idea if these are correct, but it seems fairly likely that if the xpresso build is still being used then it is somehow accidentally pulling in the pin definitions from the rot-1 configuration, in which case they should still be correct.

Though oops, I only now actually checked whether `app-sprot.toml` works on the master branch and it doesn't, so I guess that was wasted work.

jamesmunns (approved):
Thank you, and apologies for the delay @aapoalas!

PR #1859, opened 2024-08-27
Add key-based tech port unlocking

mkeeter:
Hubris counterpart to https://github.com/oxidecomputer/management-gateway-service/pull/274; see that PR or [RFD 492 § Sketch of an unlock policy](https://rfd.shared.oxide.computer/rfd/492#_sketch_of_an_unlock_policy) for more details.

We use the new `SpRot::lifecycle_state` function to decide whether to send `UnlockChallenge::Trivial` or `UnlockChallenge::EcdsaSha2Nistp256`.

(opening as a draft pending testing)


app/sidecar/base.toml
```diff
@@ -119,6 +130,9 @@ task-slots = [
 features = ["sidecar", "vlan", "auxflash", "vpd"]
 notifications = ["socket", "usart-irq", "timer"]
 
+[tasks.control_plane_agent.config]
+trusted-keys = ["../../support/fake_keys/id_ecdsa.pub"]
```
mkeeter, line 134: TODO: switch to the actual support keys from https://github.com/oxidecomputer/support-piv-tokens/

lzrd: I assume you plan to have all keys/modes compiled in and then select based on the RoT providing information to the SP as to the mode (prod-release, etc. per your previous PR) with the default being to fail safe.
Maybe you're thinking of a way to compile in just the keys needed for the ultimate signed image.

lzrd (approved):

app/sidecar/base.toml
mkeeter: Basically!  There are two modes, based on what the RoT reports:

- Release, which sends `UnlockChallenge::EcdsaSha2Nistp256`, and the reply must match one of these keys
- Not-release, which sends `UnlockChallenge::Trivial` (so the reply is also trivial)

In other words, there's no "fake keys" mode (and I'm going to delete the `fake_keys` subfolder before moving this out of draft).  Otherwise, we have the Illusion of Security, and have to explain why we've committed private keys to the repo.

The two-mode logic is already implemented in `control-plane-agent`; it's just the actual keys that need to change.

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -393,6 +405,65 @@ impl MgsHandler {
...
+    let sig = p256::ecdsa::Signature::from_bytes(signature.as_slice().into())
+        .map_err(|_| GwMonorailError::UnlockAuthFailed)?;
+
+    let v = p256::ecdsa::VerifyingKey::from_encoded_point(
+        &p256::EncodedPoint::from_bytes(key)
+            .map_err(|_| GwMonorailError::UnlockAuthFailed)?,
+    )
+    .map_err(|_| GwMonorailError::UnlockAuthFailed)?;
+
+    // Build an SSH signature blob to be verified
```
labbott, line 428: Is there not a rust crate to do this or is there one but it's not `no-std`

mkeeter: [`ssh_key::SshSig::signed_data`](https://docs.rs/ssh-key/latest/ssh_key/struct.SshSig.html#method.signed_data) is what we want, but it returns a `Vec<u8>`.  This is because the namespace (`monorail-unlock` here) is of unknown size.

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -597,14 +668,34 @@ impl SpHandler for MgsHandler {
...
+                        use drv_sprot_api::{LifecycleState, SpRot};
+                        let sprot =
+                            SpRot::from(crate::mgs_common::SPROT.get_task_id());
+                        let challenge = match sprot.lifecycle_state() {
+                            Ok(
+                                LifecycleState::Development
+                                | LifecycleState::Unprogrammed
+                                | LifecycleState::EndOfLife,
+                            )
+                            | Err(_) => {
```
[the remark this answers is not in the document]
mkeeter: That was intentional, but I could be persuaded otherwise.

We have checksums / retries in the `SpRot` protocol, so we shouldn't see glitches / spurious failures; if we can't talk to the RoT, something is Very Wrong, and failing into a state where we can debug the system seemed reasonable.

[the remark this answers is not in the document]
labbott: The problem is we both want authorized users to be able to debug the system and not open up a back door if something Very Bad has happened with the RoT. otoh, if we can't actually communicate with the RoT at all we're probably not actually going to be able to do anything about that over the tech port.

Can you add a comment that failing open is an intentional choice? My gut is telling me we may need to re-evaluate this in the future.

Cargo.toml
```diff
@@ -118,7 +120,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2"] }
 # Oxide forks and repos
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"], branch = "mkeeter/tech-port-unlock-ssh" }
```
mkeeter, line 123: TODO switch this back before merging

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -1042,3 +1133,36 @@ fn sp_error_from_ignition_error(err: IgnitionError) -> SpError {
...
+    static_assertions::const_assert!(
+        HW_ID_LEN >= core::mem::size_of::<VpdIdentity>()
+    );
+    hw_id[..core::mem::size_of::<VpdIdentity>()]
+        .copy_from_slice(identity.as_bytes());
+
+    let now = sys_get_timer().now;
+    Ok(EcdsaSha2Nistp256Challenge {
+        hw_id,
+        sw_id: [0, 0, 0, 1],
```
aapoalas, line 1162: **question:** Is this equivalent to something like `SOFTWARE_IDENTITY.to_??_bytes()` for some (potentially inexistent) constant?

aapoalas (commented):
Peanut gallery checking in from my morning coffee.

task/control-plane-agent/src/mgs_sidecar.rs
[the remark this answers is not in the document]
mkeeter: Sure, I added a comment to that effect

labbott (approved):

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: This is mostly a placeholder for now, added a comment to that effect

support/support_tokens/README.md
```diff
@@ -0,0 +1,6 @@
+This folder contains the public keys for our support PIV tokens.
+
+`authorized_keys` is from [`oxidecomputer/support-piv-tokens@36abeb82`](https://github.com/oxidecomputer/support-piv-tokens/blob/36abeb82aacb3640be2f6c988e49844356cf65a8/authorized_keys).
```
jgallagher, line 3: Should we note that these are also in omicron? (And maybe also note in omicron that they're repeated here, in case someone is changing them?)

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -312,6 +316,14 @@ impl MgsHandler {
                 UnlockChallenge::Trivial { timestamp: ts1 },
                 UnlockResponse::Trivial { timestamp: ts2 },
             ) if ts1 == ts2 => Ok(()),
+            (
```
jgallagher, line 319: Tiny nit - The comment just outside the diff above this is out of date ("right now there's only one option..."). It's maybe also wrong - since we use a `_ => Err(_)` the match doesn't really check for new types?

jgallagher (approved):

support/support_tokens/README.md
mkeeter: Sure, added a comment and opened https://github.com/oxidecomputer/omicron/pull/6494

task/control-plane-agent/src/mgs_sidecar.rs
mkeeter: Fixed!

PR #1860, opened 2024-08-27
Implement persistent data for Cosmo `hf` slot selection

mkeeter:
Staged on top of #1853

This PR implement persistent storage in host flash for Cosmo, using the same strategy as Gimlet: reserve sector 0 and tile it with a small data payload, using a CRC to ensure validity and a monotonic counter to find the newest payload.

This only works because the AMD boot sequence ignores sector 0 (other than word 0, which must be all 1s).

`RawPersistentData` is moved to `drv-hf-api` and renamed `HfRawPersistentData` for consistency.  In addition, it now implements `cmp`, making a few operations cleaner.

The new code in `drv-cosmo-hf` is _mostly_ copypasta from `drv-gimlet-hf`, but with a few tweaks (both virtual devices are on the same flash chip, the QSPI driver has a slightly different API, etc).

### DIFF
--- a/drv/cosmo-hf/src/hf.rs
+++ b/drv/cosmo-hf/src/hf.rs
@@ -62,23 +62,32 @@ impl ServerImpl {
     }
 
     fn flash_base(&self) -> u32 {
-        match self.dev {
+        Self::flash_base_for(self.dev)
+    }
+
+    fn flash_addr(&self, offset: u32) -> u32 {
+        Self::flash_addr_for(offset, self.dev)
+    }
+
+    fn flash_addr_for(offset: u32, dev: HfDevSelect) -> u32 {
+        offset + Self::flash_base_for(dev)
+    }
+
+    fn flash_base_for(dev: HfDevSelect) -> u32 {
+        match dev {
             HfDevSelect::Flash0 => 0,
             HfDevSelect::Flash1 => SLOT_SIZE_BYTES,
         }
     }
 
-    fn flash_addr(&self, offset: u32) -> u32 {
-        offset + self.flash_base()
-    }
-
-    /// Scans the currently active (virtual) device for persistent data
+    /// Scans the provided (virtual) device for persistent data
     ///
     /// Returns a tuple containing
-    /// - newest persistent data record
+    /// - newest valid persistent data record
     /// - address of first empty slot
     fn persistent_data_scan(
         &mut self,
+        dev: HfDevSelect,
     ) -> (Option<HfRawPersistentData>, Option<u32>) {
         let mut best: Option<HfRawPersistentData> = None;
         let mut empty_slot: Option<u32> = None;
@@ -86,11 +95,12 @@ impl ServerImpl {
             let addr = i * HF_PERSISTENT_DATA_STRIDE as u32;
             let mut data = HfRawPersistentData::new_zeroed();
             self.drv
-                .flash_read(self.flash_addr(addr), &mut data.as_bytes_mut())
+                .flash_read(
+                    Self::flash_addr_for(addr, dev),
+                    &mut data.as_bytes_mut(),
+                )
                 .unwrap_lite(); // flash_read is infallible when using a slice
-            if data.is_valid() && best.map(|b| data > b).unwrap_or(true) {
-                best = Some(data);
-            }
+            best = best.max(Some(data).filter(|d| d.is_valid()));
             if empty_slot.is_none()
                 && data.as_bytes().iter().all(|b| *b == 0xFF)
             {
@@ -101,11 +111,15 @@ impl ServerImpl {
     }
 
     fn get_persistent_data(&mut self) -> Result<HfPersistentData, HfError> {
-        let out = self.get_raw_persistent_data()?;
-        Ok(HfPersistentData {
+        self.get_raw_persistent_data().map(|out| HfPersistentData {
             dev_select: match out.dev_select {
                 0 => HfDevSelect::Flash0,
-                _ => HfDevSelect::Flash1,
+                1 => HfDevSelect::Flash1,
+
+                // get_raw_persistent_data only returns data with a valid
+                // HfDevSelect. TODO add an intermediate type for a validated
+                // HfRawPersistentData?
+                _ => unreachable!(),
             },
         })
     }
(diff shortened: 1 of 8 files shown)


drv/cosmo-hf/src/hf.rs
```diff
@@ -50,6 +68,134 @@ impl ServerImpl {
...
+    ) -> (Option<HfRawPersistentData>, Option<u32>) {
+        let mut best: Option<HfRawPersistentData> = None;
+        let mut empty_slot: Option<u32> = None;
+        for i in 0..SECTOR_SIZE_BYTES / HF_PERSISTENT_DATA_STRIDE as u32 {
+            let addr = i * HF_PERSISTENT_DATA_STRIDE as u32;
+            let mut data = HfRawPersistentData::new_zeroed();
+            self.drv
+                .flash_read(self.flash_addr(addr), &mut data.as_bytes_mut())
+                .unwrap_lite(); // flash_read is infallible when using a slice
+            if data.is_valid() && best.map(|b| data > b).unwrap_or(true) {
```
aapoalas, line 88: **thought:** I spent so long pondering: "Why on earth is bigger better?" :D Took me way too long to scroll down and find that there's a custom Ord impl.

drv/cosmo-hf/src/main.rs
```diff
@@ -259,6 +257,9 @@ impl FlashDriver {
     }
 
     /// Writes data from a `BufReader` into the flash
+    ///
+    /// This function will only return an error if it fails to write to a
```
aapoalas, line 261: **praise:** Good addition.

I wonder if it would be a good worth it to make an explicit infallible version to avoid the need for `.unwrap_lite()`? Would that help remove some unnecessary codegen?

aapoalas (commented):
Peanut gallery!

drv/cosmo-hf/src/hf.rs
```diff
@@ -50,6 +68,134 @@ impl ServerImpl {
...
+    ///
+    /// Returns a tuple containing
+    /// - newest persistent data record
+    /// - address of first empty slot
+    fn persistent_data_scan(
+        &mut self,
+    ) -> (Option<HfRawPersistentData>, Option<u32>) {
+        let mut best: Option<HfRawPersistentData> = None;
+        let mut empty_slot: Option<u32> = None;
+        for i in 0..SECTOR_SIZE_BYTES / HF_PERSISTENT_DATA_STRIDE as u32 {
```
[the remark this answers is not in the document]
hawkw: @aapoalas note that your Godbolt is compiled for x86, while Hubris only runs on small ARM targets, and compiled with optimization settings that prefer code size over speed. So, in the future, you may want to try passing compiler flags like `-C opt-level=z --target=thumbv7em-none-eabihf` or similar when you're curious about code generation in Hubris.

That said, the current implementation also generates less code for a realistic hubris target, too: https://godbolt.org/z/M67d1EqjK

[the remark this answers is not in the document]
aapoalas: Yeah, I figured the compile target wouldn't really have much effect in this case as it seemed like the assembly output indicated that the compiler didn't "fully get it" (as much as it ever can, of course).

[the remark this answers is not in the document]
aapoalas: Oh, I didn't know or think about that regarding Cortex-M and SIMD. Thank you, that's good to know.

drv/cosmo-hf/src/hf.rs
```diff
@@ -53,6 +71,134 @@ impl ServerImpl {
...
+
+        self.dev = prev_slot;
+        let (b_data, b_next) = self.persistent_data_scan();
+
+        match (a_data, b_data) {
+            (Some(a), Some(b)) => {
+                match a.cmp(&b) {
+                    core::cmp::Ordering::Less => {
+                        self.dev = !prev_slot;
+                        self.write_raw_persistent_data_to_addr(a_next, &b);
```
hawkw, line 177: similarly to my comments on the calls to `persistent_data_scan`, what do you think about just passing the `HfDevSelect` to this function rather than mutating state we have to then remember to go back and un-mutate?

drv/cosmo-hf/src/hf.rs
```diff
@@ -218,15 +364,54 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
+        let data = HfPersistentData { dev_select };
+
+        let prev_slot = self.dev;
+
+        // Scan both slots for persistent data
+        self.dev = !prev_slot;
+        let (a_data, a_next) = self.persistent_data_scan();
+
+        self.dev = prev_slot;
+        let (b_data, b_next) = self.persistent_data_scan();
```
hawkw, line 384: aaaaaand my comments about the toggling of `self.dev` apply to this function as well, of course.

hawkw (commented):
Overall, this looks great. There were basically two things that stuck out to me that could, potentially, be worth considering:

1. I noticed a bunch of code that could be simplified by using the `Ord` impl for `Option<T> where T: Ord`, rather than `match`ing on the two `Option`s and comparing their inner `HfRawPersistentData`s. The resultant code feels a bit nicer to me, but if you think the current `Option` handling is more explicit, my objection to it is not a blocker
2. I don't love the code that toggles the value of `self.dev` and then calls `self.persistent_data_scan()` or `self.write_raw_persistent_data_to_addr()` which read the value of `self.dev`...I don't think there are any bugs that could result from it, thanks to Rust's shared XOR mutable aliasing rules (and, also, the fact that all Hubris tasks are inherently single-threaded!) but, it _feels_ a little unfortunate to me. What do you think about just passing the `HfDevSelect` into those functions, instead? The only downside is passing more stuff on the stack. Which, if you think that's worth optimizing for, I'm fine with leaving this code as-is as well.

drv/hf-api/src/lib.rs
```diff
@@ -101,4 +101,92 @@ pub struct HfPersistentData {
...
+    oxide_magic: u32,
+
+    /// Must always be `HF_PERSISTENT_DATA_HEADER_VERSION` (for now)
+    header_version: u32,
+
+    /// Monotonically increasing counter
+    pub monotonic_counter: u64,
+
+    /// Either 0 or 1; directly translatable to [`HfDevSelect`]
+    pub dev_select: u32,
```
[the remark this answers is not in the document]
mkeeter: I'm going to not touch this, because it's already used for Gimlet host-flash stuff (and we can't derive `FromBytes` on a `HfDevSelect`, because it could have invalid bit paterns)

drv/cosmo-hf/src/hf.rs
```diff
@@ -114,24 +128,11 @@ impl ServerImpl {
...
-            (Some(_), None) => a,
-            (None, Some(_)) => b,
-            (None, None) => None,
-        }
-        .ok_or(HfError::NoPersistentData)
+        // Read the two slots
+        let (data0, _) = self.persistent_data_scan(HfDevSelect::Flash0);
+        let (data1, _) = self.persistent_data_scan(HfDevSelect::Flash1);
+
+        data0.max(data1).ok_or(HfError::NoPersistentData)
```
hawkw, line 135: lovely :)

drv/cosmo-hf/src/hf.rs
```diff
@@ -161,41 +163,29 @@ impl ServerImpl {
...
-            }
-            (Some(a), None) => {
-                self.write_raw_persistent_data_to_addr(b_next, &a)
-            }
-            (None, Some(b)) => {
-                self.dev = !prev_slot;
-                self.write_raw_persistent_data_to_addr(a_next, &b);
-                self.dev = prev_slot;
+        let (data0, next0) = self.persistent_data_scan(HfDevSelect::Flash0);
+        let (data1, next1) = self.persistent_data_scan(HfDevSelect::Flash1);
```
hawkw, line 167: as an aside, i like that we have gone from `a`/`b` naming here to `0`/`1`, because it matches the `HfDevSelect` terminology! :)

hawkw (approved):
lovely, the way you've applied my suggestions looks great! thanks! :)

PR #1866, opened 2024-09-09
Include the Hubris measurement in DICE

lzrd:
Include the Hubris measurement in DICE even though it is done in Hubris pre-main.

### DIFF
--- a/lib/lpc55-rot-startup/src/lib.rs
+++ b/lib/lpc55-rot-startup/src/lib.rs
@@ -163,32 +163,15 @@ pub fn startup(
 
     apply_memory_protection(mpu);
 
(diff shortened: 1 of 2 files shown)


lib/lpc55-rot-startup/src/lib.rs
```diff
@@ -201,6 +184,30 @@ pub fn startup(
         panic!();
     };
 
+    #[cfg(any(feature = "dice-mfg", feature = "dice-self"))]
```
labbott, line 187: I might make a separate change to drop this cfg check as all our RoT targets now enable one of `dice-mfg` or `dice-self`

labbott (approved):

PR #1869, opened 2024-09-16
Fix Sidecar fan speed sensor ordering in base.toml

### LINKED ISSUE #1865: Sidecar Mis-Reports bad Fans at incorrect location
In https://github.com/oxidecomputer/mfg-troubleshooting/issues/23, Fan2 rotor 0 was reported as having 0RPM.  During testing, Fan 2 was running just fine but Fan 3 seemed to have a very slow rotor.  Swapping Fan 2 and Fan 3 made the issue move (as reported using `dashboard` while connected to the `Sam` programming station.

Aaron-Hartwig:
We have Sidecar's fans wired up to the controllers in a non-intuitive way. That leads to a couple places we need to have this accounted for properly in the code: speed sensor naming and controller mapping. The speed sensor naming was incorrect and fixed by this commit.

fixes #1865

### DIFF
--- a/app/sidecar/base.toml
+++ b/app/sidecar/base.toml
@@ -483,7 +483,7 @@ address = 0b0100_011
 device = "max31790"
 name = "East"
 description = "Fan 0/1 controller"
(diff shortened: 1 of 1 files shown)


ericaasen (approved):
looks good

PR #1873, opened 2024-09-19
Drop workflows caching

labbott:
Somewhere along the way this broke. Remove it until someone is inspired/annoyed enough to add it back.


labbott:
Turns out we do still cache, we just update our Cargo.toml enough that it frequently doesn't hit

PR #1875, opened 2024-09-20
xtask lsp: add HUBRIS_TASK environment variable

cbiffle:
Setting HUBRIS_TASK causes xtask lsp to only search tasks of a specific name, across configured apps (either the default set, or the one chosen by HUBRIS_APP).

This can be useful to get the cfgs right for a specific case.

### DIFF
--- a/build/xtask/src/lsp.rs
+++ b/build/xtask/src/lsp.rs
@@ -355,31 +355,56 @@ fn inner(file: &PathBuf, clients: &[LspClient]) -> Result<LspConfig> {
             "app/psc/rev-b.toml".to_string(),
         ]
     };
-    for app_name in preferred_apps {
-        let file = root.join(&app_name);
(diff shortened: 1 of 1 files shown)


build/xtask/src/lsp.rs
```diff
@@ -355,31 +355,52 @@ fn inner(file: &PathBuf, clients: &[LspClient]) -> Result<LspConfig> {
...
+                return Ok(out);
+            }
         }
     }
-    Err(anyhow!("could not find {package_name} used in any apps"))
+
+    // Try to be specific about the error condition.
+    let apps = preferred_apps.join(", ");
+    if let Some(taskname) = preferred_task {
+        Err(anyhow!("task {taskname} not found or {package_name} is not used; \
```
mkeeter, line 399: Nit: `bail!(...)` is incrementally shorter than `Err(anyhow!(...))`

mkeeter (approved):

PR #1877, opened 2024-09-23
Fix EMC2305 manufacturer ID

mkeeter:
...and remove disclaimer, now that we've tested things


mkeeter:
We're going to leave this open while Nathanael fights with the power sequencing on the FPGA, to avoid committing a bunch of FPGA images to the repo history.

hawkw (approved):

nathanaelhuffman (approved):
I think we can get rid of some of the TODO-ish comments in grapefruit-seq-server's main

```
// Do we have to send the synchronization word ourself, or is it built
        // into the bitstream?
        // Same with device ID check
```
at line 164 and

the TODO at line 177/178 also.

GH won't let me comment there since you didn't touch them.

Otherwise lgtm!

PR #1878, opened 2024-09-25
Perform releases on the Oxide colo

labbott:
Also add an attestation for use later


.github/workflows/build-one.yml
```diff
@@ -84,10 +86,17 @@ jobs:
         run: |
```
lzrd, line 86: Is a clippy run supposed to be exclusive to 'ubuntu-latest' (not also 'oxide-colo-builder')?

labbott: we could add it but it's not strictly necessary. I forget why we restricted it to linux only.

mkeeter (approved):

PR #1883, opened 2024-09-30
Add min and max output bounds to `PidConfig`

mkeeter:
The EMC2305 on Grapefruit doesn't turn on the fans at 0 PWM, and we want some airflow to get better readings from the temperature sensor.  Might as well make the lower and upper bounds fully parameterized by the PID config!


nathanaelhuffman (approved):
LGTM, emperical tests show the fans start to get flaky on the ruby at <12%PWM so this seems fine

PR #1884, opened 2024-10-01
Add cosmo hf mux control; minor cleanup

mkeeter:
- Implements (virtual) mux control for `cosmo-hf`
- Adds checks on reads and writes that the flash is muxed to the SP
- Moves some magic numbers into named constants in the `regs` module

### DIFF
--- a/drv/cosmo-hf/src/main.rs
+++ b/drv/cosmo-hf/src/main.rs
@@ -94,7 +94,15 @@ mod reg {
 
     pub const NOR: *mut u32 = BASE.wrapping_add(0x40);
     pub const SPICR: *mut u32 = NOR.wrapping_add(0x0);
+    pub mod spisr {
+        pub const RX_EMPTY_BIT: u32 = 1 << 6;
+    }
     pub const SPISR: *mut u32 = NOR.wrapping_add(0x1);
+    pub mod spicr {
+        pub const SP5_OWNS_FLASH: u32 = 1 << 31;
+        pub const RX_FIFO_RESET: u32 = 1 << 15;
+        pub const TX_FIFO_RESET: u32 = 1 << 7;
+    }
     pub const ADDR: *mut u32 = NOR.wrapping_add(0x2);
     pub const DUMMY_CYCLES: *mut u32 = NOR.wrapping_add(0x3);
     pub const DATA_BYTES: *mut u32 = NOR.wrapping_add(0x4);
@@ -149,6 +157,11 @@ impl FlashDriver {
         unsafe { reg.write_volatile(v) };
     }
 
+    fn modify_reg<F: Fn(u32) -> u32>(&self, reg: *mut u32, f: F) {
+        let prev = self.read_reg(reg);
+        self.write_reg(reg, f(prev));
+    }
+
     /// Wait until the FPGA is idle
     fn wait_fpga_busy(&self) {
         loop {
@@ -165,8 +178,7 @@ impl FlashDriver {
     fn wait_fpga_rx(&self) {
         for i in 0.. {
             let spi_sr = self.read_reg(reg::SPISR);
-            const RX_EMPTY_BIT: u32 = 1 << 6;
-            if spi_sr & RX_EMPTY_BIT == 0 {
+            if spi_sr & reg::spisr::RX_EMPTY_BIT == 0 {
                 break;
             }
             ringbuf_entry!(Trace::FpgaBusy { spi_sr });
@@ -179,7 +191,9 @@ impl FlashDriver {
 
     /// Clears the FPGA's internal FIFOs
     fn clear_fifos(&self) {
-        self.write_reg(reg::SPICR, 0x8080);
+        self.modify_reg(reg::SPICR, |v| {
+            v | reg::spicr::RX_FIFO_RESET | reg::spicr::TX_FIFO_RESET
+        });
     }
 
     /// Wait until the SPI flash is idle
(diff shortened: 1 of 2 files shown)


drv/cosmo-hf/src/main.rs
```diff
@@ -327,6 +341,36 @@ impl FlashDriver {
...
+        match self.get_flash_mux_state() {
+            drv_hf_api::HfMuxState::SP => Ok(()),
+            drv_hf_api::HfMuxState::HostCPU => {
+                Err(drv_hf_api::HfError::NotMuxedToSP)
+            }
+        }
+    }
+
+    fn set_flash_mux_state(&self, ms: drv_hf_api::HfMuxState) {
+        match ms {
```
aapoalas, line 364: **suggestion:** If `modify_reg` took `F: FnOnce(u32)` then (I believe) you could match on the HfMuxState to get the right side of the bitwise or, and write the rest of the function as a single `modify_reg` that uses the result of that match.

I didn't check if the compiler output is the same either way, but the code might be clearer that way: There wouldn't be two mostly equal function call paths with just one load-bearing exclamation mark hidden somewhere in there.

drv/cosmo-hf/src/main.rs
```diff
@@ -94,7 +94,15 @@ mod reg {
 
     pub const NOR: *mut u32 = BASE.wrapping_add(0x40);
     pub const SPICR: *mut u32 = NOR.wrapping_add(0x0);
+    pub mod spisr {
+        pub const RX_EMPTY_BIT: u32 = 1 << 6;
+    }
     pub const SPISR: *mut u32 = NOR.wrapping_add(0x1);
+    pub mod spicr {
```
aapoalas, line 101: **praise:** Very nice idea to pull these into constants. <3

aapoalas (commented):
Peanut gallery checking in.

drv/cosmo-hf/src/main.rs
cbiffle: Hmm, I think you could do this refactoring even with the current type for `modify_reg` -- and I bet it would generate less code, just because the way things are currently written, each `modify_reg` is a monomorphization site (due to the closure types being disjoint) and that tends to encourage more code generation.

mkeeter: Done in c2461ba96956a5c7ff08cd153f3cb46c5713ab29

drv/cosmo-hf/src/main.rs
```diff
@@ -94,7 +94,15 @@ mod reg {
...
+        pub const SP5_OWNS_FLASH: u32 = 1 << 31;
+        pub const RX_FIFO_RESET: u32 = 1 << 15;
+        pub const TX_FIFO_RESET: u32 = 1 << 7;
```
hawkw, line 104: I don't suppose there's an easily referenceable citation for where these bits come from we could point to in a comment?

hawkw (approved):
No concerns from me, at least!

drv/cosmo-hf/src/main.rs
nathanaelhuffman: so "yes" and future work required here. Over in fpga-land our register specs are defined and as part of the build we generate these: https://github.com/oxidecomputer/hubris/blob/master/drv/grapefruit-seq-server/spi_nor_reg_map.json which are carried in here or you can read the .html memory map at that same location.

For gimlet/sidecar, these had minimal re-usable parts so a flat memory map for the whole fpga worked fine, and there is xtask code to transform this json into something usable in rust. As we expand our use of FPGAs I think this isn't the right path, we'd ideally like to have these blocks collocated with the drivers. I'm now generating these .json files on a per-module basis but some further discussion needs to happen to better transform these into usable rust structures.  I'm happy to skip the json step and generate rust directly also but getting some examples of what we want this to look like and how to generate it in a friendly, idiomatic way would be helpful and are really needed to move this forward.

For the minimal drivers we're creating here, the decision was to punt on the fancy generation work for expediency I think but I'm certainly interested in getting to a better place here long-term and will put work in on our side to make that happen if we have a strategy and direction!

nathanaelhuffman (approved):
Do we have a hiffy interface with which to "flip" this mux?

It's currently unclear who will "own" controlling this mux long-term, clearly some logic in the SP eventually.  Since this code doesn't exist right now, it might make sense to just flip back the mux to your own ownership rather than just erroring out, but if we have hiffy functionality for flipping that might be fine also so long as we can flip it after the error and continue on. 

Additionally, when setting the slot to page0/page1 we're going to need to set the correct flash addresses in the SP5FlashOffset register.  The SP5 processor will always read from address 0x0 when booting, so this offset allows the FPGA to translate to the correct physical flash addresses corresponding to the two different pages.

As it is with no offsets set there, no matter the page selected, the SP5 will always start fetching at physical address 0.  This may also be problematic as we use sector0 for special things, and so may need to skip sector0 in the slot0 case as a general rule.

mkeeter:
> Do we have a hiffy interface with which to "flip" this mux?

Yes, `HostFlash.set_mux/get_mux`

> Additionally, when setting the slot to page0/page1 we're going to need to set the correct flash addresses in the SP5FlashOffset register. The SP5 processor will always read from address 0x0 when booting, so this offset allows the FPGA to translate to the correct physical flash addresses corresponding to the two different pages.

We're already doing this in `set_dev`'s implementation, but this reminded me to check whether we were also doing it at startup (from the persistent slot selection in flash); we were not!  Fixed in 604d300ae6df

PR #1887, opened 2024-10-01
Make `oxide-rot-1-dev` a completely different board

labbott:
`oxide-rot-1/app.toml` expects to be run on a machine that has been through the full manufacturing flow. If it is run on one that hasn't had that happen it will fail to boot (loop forever). This is mostly a concern for engineering development but it keeps tripping people up so attempt to differentiate the correct image to use by changing the board. It's not strictly incompatible but going from one to another kind is a sort of cheap and relatively easy to rework.

### DIFF
--- a/.github/workflows/build-boards.yml
+++ b/.github/workflows/build-boards.yml
@@ -99,14 +99,14 @@ jobs:
     name: build-rot
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


cbiffle (approved):
Please add `boards/oxide-rot-1-selfsigned.toml` so that we can make board-specific `cfg`s if required.

PR #1890, opened 2024-10-02
Fail to build if tasks may overflow

mkeeter:
(details to come, but opening as a draft to see if any of the CI-checked builds fail)

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -478,7 +478,7 @@ pub fn package(
         let mut possible_stack_overflow = vec![];
         for task_name in cfg.toml.tasks.keys() {
             if tasks_to_build.contains(task_name.as_str()) {
-                if task_can_overflow(&cfg, task_name)? {
+                if task_can_overflow(&cfg, task_name, verbose)? {
                     possible_stack_overflow.push(task_name);
                 }
 
@@ -1080,7 +1080,11 @@ fn build_task(cfg: &PackageConfig, name: &str) -> Result<()> {
         .context(format!("failed to build {}", name))
 }
 
-fn task_can_overflow(cfg: &PackageConfig, task_name: &str) -> Result<bool> {
+fn task_can_overflow(
+    cfg: &PackageConfig,
+    task_name: &str,
+    verbose: bool,
+) -> Result<bool> {
     // Open the statically-linked ELF file
     let f = cfg.dist_file(format!("{task_name}.tmp"));
     let data = std::fs::read(f).context("could not open ELF file")?;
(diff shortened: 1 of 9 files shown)


lzrd (approved):
I only made one pass over this, but it looks good and is a valuable addition to the build process.

build/xtask/src/dist.rs
```diff
@@ -1068,6 +1079,272 @@ fn build_task(cfg: &PackageConfig, name: &str) -> Result<()> {
...
+            let instrs = cs
+                .disasm_all(&chunk, addr.into())
+                .map_err(|e| anyhow!("disassembly failed: {e:?}"))?;
+            for instr in instrs.iter() {
+                let detail = cs.insn_detail(instr).map_err(|e| {
+                    anyhow!("could not get instruction details: {e}")
+                })?;
+                if detail.groups().iter().any(|g| {
+                    g == &InsnGroupId(InsnGroupType::CS_GRP_CALL as u8)
+                }) {
```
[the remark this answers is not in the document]
mkeeter: Not specifically – I was using `yaxpeax` earlier and was fine looking for `BL` opcodes, but we're using enough weird instructions that I was hitting `DecodeError::Incomplete` in a bunch of tasks.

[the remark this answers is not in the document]
iximeow: huh! i'll have to take a look later.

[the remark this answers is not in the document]
mkeeter: It should be easy enough to test by checking out one of the earlier commits, e.g. d7fc2eb8deb12744b03f51e18030be9430981ef0 uses `yaxpeax`

build/xtask/src/dist.rs
```diff
@@ -1656,6 +1933,7 @@ fn build(
             "-C link-arg=-z -C link-arg=common-page-size=0x20 \
              -C link-arg=-z -C link-arg=max-page-size=0x20 \
              -C llvm-args=--enable-machine-outliner=never \
+             -Z emit-stack-sizes \
```
labbott, line 1936: do you know if there is any work tracking this not being nightly only? We still have a goal (dream?) of being able to build hubris with a stable toolchain.

build/xtask/src/dist.rs
```diff
@@ -1064,6 +1075,272 @@ fn build_task(cfg: &PackageConfig, name: &str) -> Result<()> {
         .context(format!("failed to build {}", name))
 }
 
+fn task_can_overflow(
```
labbott, line 1078: Are there any concerns about false positives / false negatives? I guess for a false positive we'd increase the stack size unnecessarily and for a false negative, well, we're back to our old methods.

labbott (approved):
I think this should be good to go!

build/xtask/src/dist.rs
mkeeter: The path to stabilization appears rocky, unfortunately:
https://github.com/rust-lang/rust/issues/54192

build/xtask/src/dist.rs
mkeeter: Added more comments in 9471544fbf62e939a01757b91d3d2f6fd868e738 about the possibility of false positives (possible but unlikely) and negatives (possible if using dynamic dispatch).

PR #1904, opened 2024-10-17
Add buildomat jobs for select images

labbott:
It's useful to have some images that have gone through a `permslip sign`. Do so in a very limited fashion.

### DIFF
--- a/.github/buildomat/force-git-over-https.sh
+++ b/.github/buildomat/force-git-over-https.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+#
+# The token authentication mechanism that affords us access to other private
+# repositories requires that we use HTTPS URLs for GitHub, rather than SSH.
+#
+override_urls=(
+    'git://github.com/'
+    'git@github.com:'
+    'ssh://github.com/'
+    'ssh://git@github.com/'
+    'git+ssh://git@github.com/'
+)
+for (( i = 0; i < ${#override_urls[@]}; i++ )); do
+	git config --add --global url.https://github.com/.insteadOf \
+	    "${override_urls[$i]}"
+done
+
+#
+# Require that cargo use the git CLI instead of the built-in support.  This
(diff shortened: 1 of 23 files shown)


.github/buildomat/force-git-over-https.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+override_urls=(
+    'git://github.com/'
+    'git@github.com:'
+    'ssh://github.com/'
+    'ssh://git@github.com/'
+    'git+ssh://git@github.com/'
+)
+for (( i = 0; i < ${#override_urls[@]}; i++ )); do
+	git config --add --global url.https://github.com/.insteadOf \
```
mkeeter, line 14: Dumb question: why don't we need quotes around "https://github.com", i.e.

```
git config --add --global url."https://github.com/".insteadOf
```
?

labbott: I uh borrowed this setup from omicron https://github.com/oxidecomputer/omicron/blob/main/tools/include/force-git-over-https.sh

.github/buildomat/permslip-setup.sh
```diff
@@ -0,0 +1,78 @@
...
+
+$PERMSLIP_DIR/target/release/permslip-server import-ssh-key /tmp/id_p256.pub
+$PERMSLIP_DIR/target/release/permslip-server import-private-key "UNTRUSTED bart" rsa "$BART_KEY"
+$PERMSLIP_DIR/target/release/permslip-server start-server &
+
+sleep 5
+
+$PERMSLIP_DIR/target/release/permslip --url=http://localhost:41340 list-keys
+
+# Gimlet
```
mkeeter, line 49: Why are we generating the gimlet / sidecar / psc keys, if we're only signing with `UNTRUSTED SP`?

labbott: That was from some earlier work that never went forward. I'll remove this

mkeeter (approved):

PR #1905, opened 2024-10-18
Clear SP flash errors we should never see

labbott:
The STM32H7 is haunted/contains an errata where the CPU may speculate into a system area (`0x1ff0_0000`) and trigger a flash error normally only seen when security features are enabled. This is not an official errata but seems to be well(?) reported

https://community.st.com/t5/stm32-mcus-products/spurious-rdperr-and-rdserr-when-all-protection-and-security/td-p/279852 https://github.com/zephyrproject-rtos/zephyr/issues/60449

One suggested workaround is to add an extra MPU region which marks The Danger Zone as NX. For the way hubris is designed this isn't plausible. Instead clear any errors before we attempt to write to flash.


cbiffle (commented):
ugh github won't let me comment on line 136 of update server main.rs, but: how about removing lines 136-142 where rdserr/rdperr are converted into a failure?

Otherwise looks good.

mkeeter (approved):
Approved, but is there a reason why we're both clearing the errors _and_ ignoring the errors?

cbiffle (approved):

labbott:
> Approved, but is there a reason why we're both clearing the errors _and_ ignoring the errors?

Seemed more comprehensive

PR #1911, opened 2024-10-29
Remove lease limit from `qspi_read`

mkeeter:
Tested on Grapefruit (where I found that we weren't updating the offset address!); eyeballed on Gimlet.

### DIFF
--- a/drv/cosmo-hf/src/main.rs
+++ b/drv/cosmo-hf/src/main.rs
@@ -241,11 +241,11 @@ impl FlashDriver {
 
     /// Reads data from the given address into a `BufWriter`
     ///
-    /// This function will only return an error if it fails to read from a
+    /// This function will only return an error if it fails to write to a
     /// provided lease; when given a slice, it is infallible.
     fn flash_read(
         &mut self,
-        offset: u32,
+        mut offset: u32,
         dest: &mut dyn idol_runtime::BufWriter<'_>,
     ) -> Result<(), ()> {
         loop {
@@ -269,6 +269,7 @@ impl FlashDriver {
                     }
                 }
             }
+            offset += len as u32;
         }
         Ok(())
     }
--- a/drv/gimlet-hf-server/src/main.rs
+++ b/drv/gimlet-hf-server/src/main.rs
@@ -505,14 +505,21 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
     fn read(
         &mut self,
         _: &RecvMessage,
-        addr: u32,
-        dest: LenLimit<Leased<W, [u8]>, PAGE_SIZE_BYTES>,
+        mut addr: u32,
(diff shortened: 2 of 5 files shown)


drv/gimlet-hf-server/src/main.rs
```diff
@@ -505,14 +505,21 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
         _: &RecvMessage,
-        addr: u32,
-        dest: LenLimit<Leased<W, [u8]>, PAGE_SIZE_BYTES>,
+        mut addr: u32,
+        dest: Leased<W, [u8]>,
     ) -> Result<(), RequestError<HfError>> {
         self.check_muxed_to_sp()?;
-        self.qspi.read_memory(addr, &mut self.block[..dest.len()]);
+        let mut offset = 0;
+        for i in (0..dest.len()).step_by(self.block.len()) {
```
aapoalas, line 513: **suggestion:** Since `i` is only used to calculate `n` which itself is either the block length or smaller than that, would it perhaps be clearer to rewrite this as a `k = dest.len() % block.len()` loop, each reading `block.len()` bytes, followed by a final step of reading the remainder? Alternatively, combine the two and do the min calculation on every loop of course.

I had to wrack my brain a good while to figure out what the `n` calculation was doing.

**suggestion:** Maybe perform `self.block.len()` only once (and `dest.len()` as well?) and reuse the result: I didn't check with Godbolt but I wouldn't be surprised if the mutable access to `dest` and `self.block` wouldn't cause the compiler to think it has to reload the lengths, even though they presumably never do change.

aapoalas (commented):
Peanut gallery checking in.

drv/gimlet-hf-server/src/main.rs
aapoalas: **note:** While looking at my folly, I noticed that it seems like `step_by` is stopping some compiler work from happening: https://godbolt.org/z/6Phz5WEcG

It's a shame that such a useful API drops code off a (probably minor) cliff :(

I hope this at least is interesting and potentially useful information to you.

drv/cosmo-hf/src/main.rs
```diff
@@ -241,11 +241,11 @@ impl FlashDriver {
 
     /// Reads data from the given address into a `BufWriter`
     ///
-    /// This function will only return an error if it fails to read from a
+    /// This function will only return an error if it fails to write to a
```
hawkw, line 244: whoops :P

drv/cosmo-hf/src/main.rs
```diff
@@ -269,6 +269,7 @@ impl FlashDriver {
                     }
                 }
             }
+            offset += len as u32;
```
hawkw, line 272: should we be doing something to avoid a panic site from the overflow check here?

PR #1916, opened 2024-11-07
draft PR for commenting on snitch implementation

cbiffle:


hawkw (commented):
Overall, this is lovely, although --- as we discussed out of band --- I think the ereport ingestion protocol will be easier to implement if there's a cursor-style interface for reads, so that you can reply to a "give me everything you've got" message and then handle a single ack for every ereport you replied with. Essentially exposing slightly more of the ringbuffer guts to the protocol.

This implementation is lovely though. I picked a few inconsequential nits.

PR #1917, opened 2024-11-11
Support the NUCLEO-STM32F429ZI board.

jonathanpallant:
On this new board, we get the user-led task to drive the LEDs via the `stm32xx-sys` task, like it did on the STM32H7 based boards.

This PR also updates the existing STM32F4-DISCOVERY support in the `user-led` task (which uses the STM32F407) to also use the `stm32xx-sys` task for driving LEDs. This leaves just the STM32F3-DISCOVERY using the direct-access approach.

The STM32 UART support also now understands both the F429 and F407, but I didn't yet hook up a terminal to see if it works.

The demo was tested on a NUCLEO-STM32F429ZI. Unfortunately I don't have a STM32F4-DISCOVERY to hand to check I didn't break anything there.

This was literally my first introduction to Hubris, so hopefully I got it all mostly right. The changes took about five hours, all told. Much of that was diagnosing the `stm32xx-sys` constantly restarting - it was because it didn't have permission to write to the GPIO address range.

### DIFF
--- a/drv/stm32xx-gpio-common/src/server.rs
+++ b/drv/stm32xx-gpio-common/src/server.rs
@@ -38,6 +38,17 @@ cfg_if::cfg_if! {
                 compile_error!("unsupported or missing SoC model feature");
             }
         }
+    } else if #[cfg(feature = "family-stm32f4")] {
+        use stm32f4 as pac;
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "model-stm32f407")] {
+                use pac::stm32f407 as device;
+            } else if #[cfg(feature = "model-stm32f429")] {
+                use pac::stm32f429 as device;
+            } else {
+                compile_error!("unsupported or missing SoC model feature");
+            }
+        }
     } else {
         compile_error!("unsupported or missing SoC family feature");
     }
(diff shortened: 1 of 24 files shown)


drv/stm32xx-sys-api/src/lib.rs
```diff
@@ -12,7 +12,10 @@ cfg_if::cfg_if! {
         pub use self::g0::*;
     } else if #[cfg(feature = "family-stm32h7")] {
         mod h7;
-        pub use self::h7::*;
+        pub use self::f4::*;
```
labbott, line 15: I think this was a mismerge

jonathanpallant: Oh no. Let me fix that.

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -259,6 +270,10 @@ impl_gpio_periph!(gpioa);
 #[cfg(feature = "has-gpiob-type")]
 impl_gpio_periph!(gpiob);
 
+// At least F4 distinguish gpiok from other ports.
```
hawkw, line 273: Tiny grammar edit:
```suggestion
// At least F4 distinguishes gpiok from other ports.
```

drv/user-leds/src/main.rs
```diff
@@ -575,6 +537,100 @@ fn led_toggle(led: Led) {
     sys.gpio_toggle(pinset.port, pinset.pin_mask).unwrap_lite();
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// The STM32F4 specific bits.
```
hawkw, line 541: Take it or leave it: I wonder if we might want to move the board-specific bits into submodules with `#[cfg(feature = <board feature>]` on them, instead of sticking a `cfg` attribute on each function and constant? We could do the same for the existing F3 stuff. The code might read a little nicer that way. But, it's up to you.

jonathanpallant: That sounds like a good follow-up PR. I just wanted to share what I'd done to make it run on the board I had.

PR #1919, opened 2024-11-17
Demonstrate GPIO interrupts on the LPC55S69-EVK board.

lzrd:
Pressing the USER button generates an interrupt to the button task.

A single press will increment the RGB pattern.

Quick (~1.5s) successive presses 2nd, 3rd, and subsequent will respecitvely
  - turn off LEDs
  - blink LEDs
  - cycle through RBG pattern including all off.

Minor updates to other app.toml files:
  - add `pint` and `inputmux` peripherals to their `gpio_driver` tasks.
  - when compiling all targets, some needed slight allocation adjustments not related to this PR.


drv/lpc55-gpio/src/main.rs
```diff
@@ -137,8 +146,150 @@ impl idl::InOrderPinsImpl for ServerImpl<'_> {
...
+            | PintSlot::Slot2
+            | PintSlot::Slot3
+            | PintSlot::Slot4
+            | PintSlot::Slot5
+            | PintSlot::Slot6
+            | PintSlot::Slot7 => {
+                // Ensure that the pin is configured as an input during
+                // PINT setup. XXX remove if not needed.
+                let (tport, tpin) = gpio_port_pin_validate(pin);
+                self.set_pin_direction(tport, tpin, Direction::Input);
```
lzrd, line 170: The data sheet says that the pin must be an input. But we are going to be turning the SP_RESET line back and forth between input and output. The minimal effective procedure needs to be worked out to avoid spurious interrupts when driving SP_RESET and not losing them when detecting SP_RESET.

task/button/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+        quick: 0,
+        last_button_press: 0,
+        on: true, // LEDs are on
+        on_ms: 0,
+        off_ms: 0,  // timer inactive, with on_ms > 0, increment
+        rgb: 0b111, // start with all LEDs on
+    };
+
+    // XXX for now at least, setting up a PINT pin will set dir to Input
+    server.gpio.disable_rising(BUTTON_PINT_MASK);
```
lzrd, line 245: Some of these calls are redundant with the reset state of the PINT peripheral and can be removed.

lzrd:
Note: This implementation only implements edge-detection interrupts and ignores level-detection.

drv/lpc55-gpio/src/main.rs
```diff
@@ -51,6 +51,19 @@ task_slot!(SYSCON, syscon_driver);
...
     gpio: &'a device::gpio::RegisterBlock,
+    pint: &'a device::pint::RegisterBlock,
+    inputmux: &'a device::inputmux::RegisterBlock,
+}
+
+impl ServerImpl<'_> {
+    fn set_pin_direction(&self, port: usize, pin: usize, dir: Direction) {
+        match dir {
+            Direction::Input => self.gpio.dirclr[port]
+                .write(|w| unsafe { w.dirclrp().bits(1 << pin) }),
```
aapoalas, line 62: **question:** Would adding safety comments while moving this code be worthwhile, or is it immediately obvious?

task/button/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+            .set_val(BLUE_LED, if b { Value::Zero } else { Value::One });
+        Ok(leds)
+    }
+
+    fn timer_expiry(&mut self) {
+        // then:
+        //   - toggle cycle
+        //   - speed up
+        // else
+        //   - inrement
```
aapoalas, line 82: **type:**
```suggestion
        //   - increment
```

aapoalas (commented):
:peanuts:

task/button/src/main.rs
lzrd: didn't need that comment anyway :)

app/gimletlet/app-mgmt.toml
```diff
@@ -49,7 +49,7 @@ task-slots = ["sys", "user_leds"]
 
 [tasks.net]
 name = "task-net"
-stacksize = 3000
```
labbott, line 52: Was this strictly necessary or can this be split out into a separate change?

build/lpc55pins/src/lib.rs
```diff
@@ -140,21 +207,27 @@ impl ToTokens for PinConfig {
...
     let dest_path = out_dir.join("pin_config.rs");
-    let mut file = std::fs::File::create(dest_path)?;
-
+    let mut file = std::fs::File::create(&dest_path)?;
     let mut buf = BufWriter::new(Vec::new());
+    let mut used_slots = 0u8;
+
     if pins.iter().any(|p| p.name.is_some()) {
-        writeln!(&mut buf, "use drv_lpc55_gpio_api::Pin;")?;
+        writeln!(&mut file, "use drv_lpc55_gpio_api::Pin;")?;
```
labbott, line 215: This doesn't look write, the goal was to write everything into the buf first and then write to the file at the end

drv/lpc55-gpio/src/main.rs
```diff
@@ -137,8 +145,150 @@ impl idl::InOrderPinsImpl for ServerImpl<'_> {
...
+                self.set_pin_direction(tport, tpin, Direction::Input);
+
+                // The INPUTMUX only needs to be turned on during configuration.
+                let syscon = Syscon::from(SYSCON.get_task_id());
+                syscon.enable_clock(Peripheral::Mux);
+                syscon.leave_reset(Peripheral::Mux);
+                unsafe {
+                    self.inputmux.pintsel[pint_slot as usize]
+                        .write(|w| w.bits(pin as u32));
+                }
```
labbott, line 178: Can we use the `intpin` API which would let us wrap the unsafe a bit tighter or does that not work with our hubris design?

drv/lpc55-gpio/src/main.rs
labbott: The unsafe is a quirk of how the lpc55-pac is generated. Adding a safety comment for this is unlikely to be useful.

drv/lpc55-gpio/src/main.rs
lzrd: Removed and tested. Not needed.

build/lpc55pins/src/lib.rs
lzrd: I'll revert this line.

PR #1921, opened 2024-11-18
Read SP task crash dumps over the network

mkeeter:
Hubris counterpart to https://github.com/oxidecomputer/management-gateway-service/pull/316, allowing us to read task dumps over the network via MGS.

These MGS messages – unlike `dump-agent` – require more intelligence into the Hubris side.  The external API is just "start read" and "continue read"; Hubris is responsible for knowing how `humpty` writes stuff into RAM.

We allow for up to 4 simultaneous readers, distinguished by caller-provided keys.  If there are more than four simultaneous readers, then (1) what the heck, and (2) some readers will return an error indicating that their key is no longer valid.

There are also minor tweaks to `dump-agent` and `jefe` implementations:
- `dump-agent` learns how to read dump data into a lease (instead of returning a 256-byte value)
- `jefe` caches the most recent dump area → memory address lookup, so we don't always need to start from the beginning of the linked list.  This isn't useful if we have multiple readers, but makes the single-reader case less Accidentally Quadratic.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,7 +124,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2", "deflate
 # Oxide forks and repos
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.4.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"], branch = "mkeeter/sp-task-dump" }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 gimlet-inspector-protocol = { git = "https://github.com/oxidecomputer/gimlet-inspector-protocol", version = "0.1.0" }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
 humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.3" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,8 +124,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2", "deflate
 # Oxide forks and repos
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.4.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-#gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
-gateway-messages = { path = "../management-gateway-service/gateway-messages", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"], branch = "mkeeter/sp-task-dump" }
 gimlet-inspector-protocol = { git = "https://github.com/oxidecomputer/gimlet-inspector-protocol", version = "0.1.0" }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
 humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.3" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,7 +124,8 @@ zip = { version = "0.6", default-features = false, features = ["bzip2", "deflate
 # Oxide forks and repos
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.4.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+#gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = { path = "../management-gateway-service/gateway-messages", default-features = false, features = ["smoltcp"] }
 gimlet-inspector-protocol = { git = "https://github.com/oxidecomputer/gimlet-inspector-protocol", version = "0.1.0" }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
 humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.3" }
--- a/app/gimlet/base.toml
+++ b/app/gimlet/base.toml
@@ -267,6 +267,7 @@ start = true
 uses = ["usart1"]
 task-slots = [
     "jefe",
+    "dump_agent",
     "net",
     "update_server",
     "sys",
--- a/app/gimlet/base.toml
+++ b/app/gimlet/base.toml
@@ -230,7 +230,7 @@ name = "task-host-sp-comms"
 features = ["stm32h753", "uart7", "baud_rate_3M", "hardware_flow_control", "vlan", "gimlet"]
 uses = ["uart7", "dbgmcu"]
 interrupts = {"uart7.irq" = "usart-irq"}
-priority = 7
+priority = 8
 max-sizes = {flash = 65536, ram = 65536}
 stacksize = 5080
 start = true
@@ -259,7 +259,7 @@ notifications = ["socket"]
 
 [tasks.control_plane_agent]
 name = "task-control-plane-agent"
-priority = 6
+priority = 7
 max-sizes = {flash = 131072, ram = 32768}
 # This can probably overkill and can be tuned later
 stacksize = 6000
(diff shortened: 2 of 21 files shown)


labbott (commented):
I'm a little bit iffy about failure modes for more than 4 clients. It seems like the only result should be bad data on the client end? A sufficiently malicious client would potentially only cause a DoS?

app/gimlet/base.toml
```diff
@@ -348,7 +349,7 @@ stacksize = 800
 
 [tasks.idle]
 name = "task-idle"
-priority = 8
+priority = 9
```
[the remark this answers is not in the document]
mkeeter: `control-plane-agent` now needs to call into `dump-agent`; they were previously the same priority (and then ripple effects from things above `control-plane-agent`)

task/control-plane-agent/src/dump.rs
```diff
@@ -0,0 +1,221 @@
...
+            .read_dump_into(state.area_index, 0, header.as_bytes_mut())
+            .map_err(|_e| SpError::Dump(DumpError::ReadFailed))?;
+
+        // Make sure the header is still valid
+        if header.contents != humpty::DumpContents::SingleTask.into() {
+            return Err(SpError::Dump(DumpError::NoLongerValid));
+        }
+
+        // Move along to the next area if we're at the end
+        if state.offset + size_of::<humpty::DumpSegmentData>() as u32
```
hawkw, line 163: considered obnoxious, feel free to ignore: `size_of::<humpty::DumpSegmentData>()` and `size_of::<humpty::DumpAreaHeader>()` are a little bit hard to tell apart in this function, because they're both long-ish expressions and the type parameter to `size_of` starts with `humpty::Dump` and my eyes glazed over at that point. I _might_ consider doing something like
```rust
const HEADER_SIZE: u32 = size_of::<humpy::DumpAreaHeader>() as u32;
const SEGMENT_DATA_SIZE: u32 = size_of::<humpy::DumpSegmentData>() as u32;
```
and then referring to them by those constants elsewhere, so that it's more obvious that they're different things? but maybe this is just a "me thing", in which case you're welcome to ignore me.

task/jefe/src/main.rs
```diff
@@ -86,8 +90,14 @@ struct ServerImpl<'s> {
     task_states: &'s mut [TaskStatus; NUM_TASKS],
     deadline: u64,
     reset_reason: ResetReason,
+
+    /// Base address for a linked list of dump areas
     #[cfg(feature = "dump")]
     dump_areas: u32,
+
+    /// Cache of most recently checked dump area
```
hawkw, line 98: might be nice to mention what this is used for?

task/jefe/src/main.rs
```diff
@@ -161,8 +171,33 @@ impl idl::InOrderJefeImpl for ServerImpl<'_> {
...
-                dump::get_dump_area(self.dump_areas, index)
-                    .map_err(|e| e.into())
+                // If we have cached a dump area, then use it to accelerate
+                // lookup by jumping partway through the linked list
+                let d = if let Some(prev) = self.last_dump_area {
+                    if index == prev.index {
+                        // Easy case: we've already looked up this area
+                        Ok(prev)
+                    } else if let Some(offset) = index.checked_sub(prev.index) {
+                        // Slight tricker: the requested area is after our
```
hawkw, line 181: nit: i think you meant
```suggestion
                        // Slightly tricker: the requested area is after our
```

task/jefe/src/main.rs
```diff
@@ -161,8 +171,33 @@ impl idl::InOrderJefeImpl for ServerImpl<'_> {
...
             ) -> Result<DumpArea, RequestError<DumpAgentError>> {
-                dump::get_dump_area(self.dump_areas, index)
-                    .map_err(|e| e.into())
+                // If we have cached a dump area, then use it to accelerate
+                // lookup by jumping partway through the linked list
+                let d = if let Some(prev) = self.last_dump_area {
+                    if index == prev.index {
+                        // Easy case: we've already looked up this area
+                        Ok(prev)
+                    } else if let Some(offset) = index.checked_sub(prev.index) {
```
hawkw, line 180: i was going to suggest that this expression might be better represented as a `match`, since the two default cases could be combined, but...i think you can't write a match guard like
```rust
Some(prev) if let Some(offset) = index.checked_sub(prev.index) =>
```
which is a bit of a bummer. I suppose you could instead write
```rust
Some(prev) if prev.index >= index => 
```
and then do the subtraction in the body of the match arm, but using `checked_sub` both as the check and the actual subtraction feels much nicer, and the case where the `checked_sub` returns `None` would still have to be handled despite never actually happening, which feels quite sad...

task/control-plane-agent/src/dump.rs
```diff
@@ -0,0 +1,221 @@
...
+    area_index: u8,
+
+    /// Offset into the dump area
+    ///
+    /// This should point to a `DumpSegmentData` header, followed by compressed
+    /// data.
+    offset: u32,
+}
+
+const MAX_DUMP_CLIENTS: usize = 4;
```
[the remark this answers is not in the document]
mkeeter: Mostly vibes, but I was thinking something like "max of two MGSes plus two tech port connections".

task/jefe/src/main.rs
mkeeter: Yup, done

mkeeter:
> I'm a little bit iffy about failure modes for more than 4 clients. It seems like the only result should be bad data on the client end? A sufficiently malicious client would potentially only cause a DoS?

Yeah, that's correct – if I open 5x terminals and run
```
faux-mgs --log-level=debug --interface en0 dump read --index 0
```
in all of them, the first one stops with `Error: Error response from SP: dump: invalid key`

Cargo.toml
```diff
@@ -124,7 +124,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2", "deflate
 # Oxide forks and repos
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.4.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"], branch = "mkeeter/sp-task-dump" }
```
bcantrill, line 127: Presumably that's not intended to be a branch by the time this integrates?

task/control-plane-agent/src/dump.rs
```diff
@@ -0,0 +1,276 @@
...
+        DumpState {
+            clients: [None; MAX_DUMP_CLIENTS],
+            agent: DumpAgent::from(DUMP_AGENT.get_task_id()),
+        }
+    }
+
+    pub(crate) fn get_task_dump_count(&mut self) -> Result<u32, SpError> {
+        let mut count = 0;
+        for index in 0.. {
+            // XXX accidentally quadratic!
```
bcantrill, line 63: I... think it's not accidental? ;) Worth turning this into a comment explaining that it's quadratic and not a problem or deleting it? (As it stands, it reads like a note to oneself.)

bcantrill, line 63: I... think it's not accidental? ;) Worth turning this into a comment explaining that it's quadratic and not a problem or deleting it? (As it stands, it reads like a note to oneself.)

task/control-plane-agent/src/dump.rs
```diff
@@ -0,0 +1,276 @@
...
+        dump_index: u32,
+        key: [u8; 16],
+    ) -> Result<DumpTask, SpError> {
+        // Find the area where this dump starts
+        let mut count = 0;
+        let mut found = None;
+        let mut data = [0u8; 256];
+        let mut header = humpty::DumpAreaHeader::new_zeroed();
+        for index in 0.. {
+            // XXX accidentally quadratic!
```
bcantrill, line 95: Same comment about the comment.

task/control-plane-agent/src/dump.rs
```diff
@@ -0,0 +1,276 @@
...
+            .read_dump_into(
+                pos.area_index,
+                pos.offset,
+                &mut buf[..ds.compressed_length as usize],
+            )
+            .map_err(|_e| SpError::Dump(DumpError::ReadFailed))?;
+        pos.offset += ds.compressed_length as u32;
+        while pos.offset & 3 != 0 {
+            // pad to the nearest u32
+            // XXX why is `humpty::DUMP_SEGMENT_MASK` private?
```
bcantrill, line 259: I think the answer there was purely historical (this happened to be the first constant in that file); will push a fix to that to `humpty` and then you can use the constant here.

bcantrill (approved):
Looks great, modulo very small comments. Will push the fix to `humpty` now to allow you to use the proper constant.

task/control-plane-agent/src/dump.rs
mkeeter: In fact, this is no longer quadratic (in the typical case); `jefe` now caches its position in the linked list and will resume from that point!

I've deleted the comment here and below.

task/control-plane-agent/src/dump.rs
mkeeter: Bumped and using the constant now

Cargo.toml
mkeeter: Yup, done

PR #1922, opened 2024-11-18
Enable GPIO Pin Interrupts

lzrd:
Add interrupt-related API calls to the LPC55 `gpio_driver`.

Also:
Fix app/gimletlet/app-mgmt.toml stack size which was found while building all targets.

---

A task on an LPC55 can now configue and use GPIO interrupts.

app.toml example shows Pin Interrupt configuration:

[tasks.foo]
...
interrupts = { "pint.irq0" = "button-irq" }
...
task-slots = ["gpio_driver", ...]

[tasks.foo.config]
pins = [
  { name="BUTTON', pin={ port=1, pin=9}, alt=0, pint=0, direction="input", opendrain="normal" }
]


drv/lpc55-gpio/src/main.rs
```diff
@@ -137,8 +145,143 @@ impl idl::InOrderPinsImpl for ServerImpl<'_> {
...
+        // If the GPIO pin is configured for interrupts, then
+        // interrupt configuration is done before pin configuration.
+        // INPUTMUX from UM11126:
+        // Once set up, no clocks are required for the input multiplexer to
+        // function. The system clock is needed only to write to or read from
+        // the INPUT MUX registers. Once the input multiplexer is configured,
+        // disable the clock to the INPUT MUX block in the AHBCLKCTRL register.
+        // See: Section 4.5.17 “AHB clock control 0”.
+        match pint_slot {
+            PintSlot::None => {}
```
labbott, line 157: This basically feels like 
```
match pint_slot {
Some(slot) => ...
_ => {}
}

File: build/lpc55pins/src/lib.rs [hunk 2]
```diff
@@ -183,21 +256,33 @@ pub fn codegen(pins: Vec<PinConfig>) -> Result<()> {
...
+                    &name, pin.0, pin.1
                 )?;
+
+                let mut ignore = 0u8;
+                let slot = p.get_pint_slot(&mut ignore);
+                if slot != PintSlot::None {
+                    writeln!(&mut buf, "#[allow(unused)]")?;
+                    writeln!(
+                        &mut buf,
+                        "pub const {}_PINT_MASK: u32 = 1 << {};",
```
labbott, line 274: Passing the raw mask around doesn't seem right and you have a comment below about wanting to validate the mask. Can we pass the `PintSlot` enum around instead? That should always be valid.

build/lpc55pins/src/lib.rs
```diff
@@ -98,6 +99,53 @@ pub enum Opendrain {
     Opendrain,
 }
 
+#[derive(Copy, Clone, Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+#[repr(u32)]
+pub enum PintSlot {
+    None = 8,
```
labbott, line 106: Having this `PintSlot::None` representation feels very strange. Can we make the API take an `Option<PintSlot>` instead so we don't have to worry about this case?

drv/lpc55-gpio/src/main.rs
lzrd: Cleaned this up using Hubpack which then simplifies everything else.

build/lpc55pins/src/lib.rs
lzrd: Changing the idl call to Hubpack encoding allows use of the enum which made this easy to clean up.

chips/lpc55/chip.toml
```diff
@@ -77,3 +77,12 @@ size = 0x1000
 [bootrom]
 address = 0x03000000
 size = 0x10000
+
+[pint]
```
cbiffle, line 81: FYI, to keep region count down (which is less critical on v8M but still), IOCON, PINT, and two currently-unused-but-relevant-looking blocks (GINT0/1) are adjacent. So you could merge them into one block at 0x4000_1000 with length 0x4000.

hawkw (commented):
I left some minor notes, hopefully they're useful.

chips/lpc55/chip.toml
lzrd: Should we have automatic coalescing of adjacent regions when used by the same task? (We don't already?)

build/lpc55pins/src/lib.rs
```diff
@@ -113,6 +162,33 @@ impl PinConfig {
 
         self.alt
     }
+
+    fn get_pint_slot(&self, used: &mut u32) -> Option<PintSlot> {
+        if let Some(slot_number) = self.pint {
+            if self.pin.port > 1 || self.pin.pin > 32 {
+                panic!("Invalid gpio pin for interrupt");
```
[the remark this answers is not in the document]
lzrd: Change made. A quick test shows that a different panic is hit before this one but defense in depth is good.

build/lpc55pins/src/lib.rs
```diff
@@ -137,67 +213,177 @@ impl ToTokens for PinConfig {
...
+        writeln!(
+            buf,
+            "iocon.set_val({} {});",
+            pin_tokens,
+            if v { "Value::One" } else { "Value::Zero" }
+        )?;
+    }
+    if let Some(d) = p.direction {
+        writeln!(buf, "iocon.set_dir({} Direction::{:?});", pin_tokens, d)?;
+    }
```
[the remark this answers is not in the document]
lzrd: quote!() is good, but I'm going to leave this as is for now.

idl/lpc55-pins.idol
```diff
@@ -41,5 +43,17 @@ Interface(
...
+               "pint_slot": "PintSlot",
+                "op": "PintOp",
+                "cond": "PintCondition",
+            },
+            reply: Result(
+                ok: "Option<bool>",
+                err: ServerDeath,
+            ),
+            encoding: Hubpack,
+        ),
```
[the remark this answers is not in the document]
lzrd: That is a good question, but this can stand for now. GPIO interrupts seem to be an exercise left for the ARM chip designer and they aren't consistent (EXTI vs PINT). I'm not opposed to having a unified GPIO interrupt API. At one point, I was checking this against Zephyr's abstractions to see if I might have missed anything. I did not enjoy going through their layers of abstraction vs the LPC55 docs. This is a good base for some later project that could unify the APIs.

chips/lpc55/chip.toml
lzrd: It's not obvious from the data that the regions can be combined. If we had a hint in the chips file
that it was ok to do that, then we could make it automatic and clear to the reader that it was intended.

In `chips/lpc55/chip.toml`, the regions are:
```
bootrom          0x03000000..0x03010000 gap=0x3cff0000
syscon           0x40000000..0x40001000 gap=0x00000000
iocon            0x40001000..0x40002000 gap=0x00011000
anactrl          0x40013000..0x40014000 gap=0x0000c000
pmc              0x40020000..0x40021000 gap=0x00013000
flash_controller 0x40034000..0x40035000 gap=0x00005000
rng              0x4003a000..0x4003b000 gap=0x0004b000
flexcomm0        0x40086000..0x40087000 gap=0x00002000
flexcomm3        0x40089000..0x4008a000 gap=0x00000000
flexcomm4        0x4008a000..0x4008b000 gap=0x00001000
gpio             0x4008c000..0x4008e484 gap=0x00007b7c
flexcomm5        0x40096000..0x40097000 gap=0x00008000
flexcomm8        0x4009f000..0x400a0000 gap=0x00004000
hash_crypt       0x400a4000..0x400a5000 gap=0x0005c200
dice_spmeasure   0x40101200..0x40101a00 gap=0x00000000
dice_rng         0x40101a00..0x40101b00 gap=0x0fefe500
secure_syscon    0x50000000..0x50001000 gap=0x00033000
```

lzrd: Created issue https://github.com/oxidecomputer/hubris/issues/2008 for further discussion but will not act on it here.

labbott (approved):

PR #1923, opened 2024-11-18
Bump stacksize in .../app-mgmt.toml so building all apps works.

lzrd:
Building with all app.toml files fails on this one. So, fix it.


cbiffle (approved):
That's a really big stack, lemme make sure we've got an issue filed to look into that. Thanks for fixing it.

PR #1930, opened 2024-11-22
gimlet-seq-server: use set_bytes not write_bytes for sequencing control

### LINKED ISSUE #1929: `gimlet_seq` should use `set_bytes`, not `write_bytes` when setting individual bits
Right now the sequencer kicks off A1: https://github.com/oxidecomputer/hubris/blob/master/drv/gimlet-seq-server/src/main.rs#L726

Later, it kicks off A0: https://github.com/oxidecomputer/hubris/blob/master/drv/gimlet-seq-server/src/main.rs#L782

However since `write_bytes` uses the `Write` operation, we effectively clear the A1 bit when setting the A0 bit. Using `set_bytes` uses a `BitSet` operation, allowing us to set A0 while leaving A1 in place. The FPGA has papered over this by ignoring the A1 bit while it is in A0.

Aaron-Hartwig:
Fixes #1929

Since `write_bytes` uses the `Write` operation, we effectively clear the A1 bit when setting the A0 bit. Using `set_bytes` uses a `BitSet` operation, allowing us to set A0 while leaving A1 in place. The FPGA has papered over this by ignoring the A1 bit while it is in A0.

### DIFF
--- a/drv/gimlet-seq-server/src/main.rs
+++ b/drv/gimlet-seq-server/src/main.rs
@@ -723,7 +723,7 @@ impl<S: SpiServer> ServerImpl<S> {
                 // failing to sequence.
                 //
                 let a1 = Reg::PWR_CTRL::A1PWREN;
-                self.seq.write_bytes(Addr::PWR_CTRL, &[a1]).unwrap_lite();
+                self.seq.set_bytes(Addr::PWR_CTRL, &[a1]).unwrap_lite();
 
                 loop {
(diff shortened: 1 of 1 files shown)


nathanaelhuffman (approved):
I concur. Without this, we're actually commanding the A1 state machine back to IDLE, but it holds the A1 rails up because A0 rails are still up and we have sequencing requirements so everything "works" even though this was currently wrong.

PR #1931, opened 2024-11-22
Add kipc::find_faulted_task.

cbiffle:
This resolves a four-year-old TODO in El Jefe asking for a way to process faulted tasks without making so many kipcs. The original supervisor kipc interface was, by definition, designed before we knew what we were doing. Now that we have some miles on the system, some things are more clear:

1. The supervisor doesn't use the TaskState data to make its decisions.
2. The TaskState data is pretty expensive to serialize/deserialize, and produces code containing panic sites.
3. Panic sites in the supervisor are bad, since it is not allowed to panic.

The new find_faulted_task operation can detect all N faulted tasks using N+1 kipcs, instead of one per potentially faulted task, and the request and response messages are trivial to serialize (one four-byte integer each way). This has allowed me to write (out-of-tree) "minisuper," a supervisor in 256 bytes that cannot panic.

In-tree, this has the advantage of knocking 33% off Jefe's flash size and reducing statically-analyzable max stack depth by 20%.

### DIFF
--- a/doc/kipc.adoc
+++ b/doc/kipc.adoc
@@ -290,6 +290,69 @@ returned by a call to `get_task_dump_region` for the specified task.
 A copy of the memory referred to by the specified region, starting
 at `base` and running for `size` bytes.
 
+=== `find_faulted_task` (8)
+
+Scans forward from a given task index searching for a faulted task. If a faulted
+task is found, returns its index. Otherwise, returns zero, the index of the
+supervisor task, which is by definition not faulted.
+
+To simplify supervisor implementations, the given task index may equal the
+number of tasks in the system. In this case, there are no tasks to search, and
+you always get zero back.
+
+We do _not_ permit greater indices than that, because that would strongly
+suggest a bug.
+
+==== Request
+
+[source,rust]
+----
+struct FindFaultedTaskRequest {
+    starting_index: u32,
+}
+----
+
+==== Preconditions
+
+The `starting_index` must be a valid index for this system, or one greater.
+
+==== Response
+
+[source,rust]
+----
+struct FindFaultedTaskResponse {
+    fault_index: u32,
+}
+----
+
+==== Notes
+
+This is intended for the common case of a supervisor scanning the task table in
+search of a fault. Compared to doing this manually with `read_task_status`,
+`find_faulted_task` has several advantages:
+
+1. Generates one kipc per failed task, plus one, rather than one kipc per
+   potentially failed task.
+
+2. Avoids serializing/transferring/deserializing the relatively complex
+   `TaskState` type, which supervisors rarely make use of in practice.
+
+3. Can be implemented in a way that doesn't potentially panic the supervisor.
+
+The "task ID or zero" return value is represented as an `Option<NonZeroUsize>`
+in the Rust API, so a typical use of this kipc looks like:
+
+[source,rust]
+----
+let mut next_task = 1; // skip supervisor
+while let Some(fault) = kipc::find_faulted_task(next_task) {
+    let fault = usize::from(fault);
+    // do things with the faulted task
+
+    next_task = fault + 1;
+}
+----
+
 == Receiving from the kernel
 
 The kernel never sends messages to tasks. It's simply not equipped to do so.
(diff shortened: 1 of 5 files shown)


sys/kern/src/kipc.rs
```diff
@@ -465,3 +468,44 @@ fn software_irq(
...
+            return Ok(NextTask::Same);
+        }
+    }
+
+    let response_len =
+        serialize_response(&mut tasks[caller], response, &0_u32)?;
+    tasks[caller]
+        .save_mut()
+        .set_send_response_and_length(0, response_len);
+    Ok(NextTask::Same)
```
mkeeter, line 510: Take it or leave it, but this could be written more concisely as
```suggestion
    let i = tasks[index..]
        .iter()
        .position(|task| matches!(task.state(), TaskState::Faulted { .. }))
        .unwrap_or(0);

    let response_len =
        serialize_response(&mut tasks[caller], response, &(i as u32))?;
    tasks[caller]
        .save_mut()
        .set_send_response_and_length(0, response_len);
    return Ok(NextTask::Same);
```
(which IMHO is also more clear)

mkeeter (approved):

doc/kipc.adoc
```diff
@@ -290,6 +290,62 @@ returned by a call to `get_task_dump_region` for the specified task.
...
+[source,rust]
+----
+let mut next_task = 1; // skip supervisor
+while let Some(fault) = kipc::find_faulted_task(next_task) {
+    let fault = usize::from(fault);
+    // do things with the faulted task
+
+    next_task = fault + 1;
+}
+----
```
hawkw, line 347: This is lovely :)

sys/userlib/src/kipc.rs
```diff
@@ -24,6 +26,20 @@ pub fn read_task_status(task: usize) -> abi::TaskState {
     ssmarshal::deserialize(&response[..len]).unwrap_lite().0
 }
 
+pub fn find_faulted_task(task: usize) -> Option<NonZeroUsize> {
+    // Coerce `task` to a known size (Rust doesn't assume that usize == u32)
+    let task = task as u32;
+    let mut response = 0_u32;
+    let (_, _) = sys_send(
+        TaskId::KERNEL,
+        Kipcnum::ReadTaskStatus as u16,
```
hawkw, line 35: uhh...this looks like the wrong `Kipcnum` variant?
```suggestion
        Kipcnum::FindFaultedTask as u16,
```

sys/userlib/src/kipc.rs
```diff
@@ -24,6 +26,20 @@ pub fn read_task_status(task: usize) -> abi::TaskState {
     ssmarshal::deserialize(&response[..len]).unwrap_lite().0
 }
 
+pub fn find_faulted_task(task: usize) -> Option<NonZeroUsize> {
```
hawkw, line 29: Perhaps this ought to have a doc comment? I know there's docs for the KIPC interface in the Hubris reference, but might we want to include the same documentation in the actual Rust API?

hawkw (changes requested):
Overall, this is lovely, but it looks like the binding in userlib is actually incorrect! It's using the KIPC number for `ReadTaskStatus` instead of for the new IPC!

Should be easy to fix but I wanted to flag it now before this merges :)

sys/kern/src/kipc.rs
cbiffle: Into it, thanks.

sys/userlib/src/kipc.rs
cbiffle: Good catch! I've mostly tested this with exhubris where I used the right number. The perils of a fork!

sys/userlib/src/kipc.rs
hawkw: In fairness, it looks like the rest of `userlib::kipc` lacks them...except for the one I added ;-)

https://github.com/oxidecomputer/hubris/blob/d6a6e67e2377a986ea78006324c108cf55154f8a/sys/userlib/src/kipc.rs#L130-L131

sys/kern/src/kipc.rs
cbiffle: Rejiggered this a little to use `enumerate` and `find`, to ensure that we don't get a silly overflow check added to the `i+index` expression.

sys/userlib/src/kipc.rs
cbiffle: No reason not to improve things while I'm in here!

hawkw (approved):
looks great!

sys/kern/src/kipc.rs
cbiffle: Heh, amusingly, that totally broke it, due to a brain-o on my part. Switched back to your method. :-)

PR #1934, opened 2024-11-22
add support for PSU firmware update

bcantrill:


drv/i2c-devices/src/mwocp68.rs
```diff
@@ -62,6 +115,43 @@ impl From<pmbus::Error> for Error {
...
+}
+
+impl UpdateState {
+    fn delay_ms(&self) -> u64 {
+        match self {
+            Self::WroteBootLoaderKey | Self::WroteProductKey => 3_000,
+            Self::BootedBootLoader => 1_000,
+            Self::StartedProgramming => 2_000,
+            Self::WroteBlock { .. } | Self::WroteLastBlock { .. } => 100,
+            Self::SentChecksum => 2_000,
```
hawkw, line 147: suuuper pedantic, feel free to ignore me: for other variants that have the same delay, we use `|` to combine them into the same match arm, but we don't for `StartedProgramming` and `SentChecksum`; for consistency's sake, perhaps:
```suggestion
            Self::StartedProgramming | Self::SentChecksum => 2_000,
            Self::WroteBlock { .. } | Self::WroteLastBlock { .. } => 100,
```

(I will freely admit that this is a goofy thing to care about, which is why i should clarify that I don't actually care.)

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -62,6 +115,43 @@ impl From<pmbus::Error> for Error {
     }
 }
 
+// const MWOCP68_MFR_ID: &str = "Murata-PS";
+// const MWOCP68_MFR_MODEL: &str = "MWOCP68-3600-D-RM";
+
+const MWOCP68_REVISION_LEN: usize = 14;
```
mkeeter, line 121: Nit: these could live in `firmware_revision` instead of the global namespace, since that's where they're used.

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -350,6 +440,262 @@ impl Mwocp68 {
...
+
+        //
+        // Per ACAN-114, we are expecting this to be of the format:
+        //
+        //    XXXX-YYYY-0000
+        //
+        // Where XXXX is the firmware revision on the primary MCU (AC input
+        // side) and YYYY is the firmware revision on the secondary MCU (DC
+        // output side).  We aren't going to be rigid about the format of
+        // either revision, but we will be rigid about the rest of the format.
```
hawkw, line 475: nitpicky: perhaps the `MWOCP_REVISION_FORMAT` constant ought to be defined near this comment? I notice that it's not referenced anywhere else, so we could just move it in here?

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -350,6 +440,262 @@ impl Mwocp68 {
...
+    }
+
+    pub fn power_good(&self) -> Result<bool, Error> {
+        use commands::mwocp68::STATUS_WORD::*;
+
+        let status = pmbus_read!(self.device, STATUS_WORD)?;
+        Ok(status.get_power_good_status() == Some(PowerGoodStatus::PowerGood))
+    }
+
+    pub fn firmware_revision(&self) -> Result<Mwocp68FirmwareRev, Error> {
```
hawkw, line 458: if i understand correctly, this function returns the revision of the primary MCU only (`XXXX` in the format string)? perhaps that should be documented here?

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -350,6 +440,262 @@ impl Mwocp68 {
...
+
+        Ok(Mwocp68FirmwareRev([data[0], data[1], data[2], data[3]]))
+    }
+
+    fn get_boot_loader_status(
+        &self,
+    ) -> Result<BOOT_LOADER_STATUS::CommandData, Error> {
+        use pmbus::commands::mwocp68::CommandCode;
+        let cmd = CommandCode::BOOT_LOADER_STATUS as u8;
+        let mut data = [0u8; 1];
```
mkeeter, line 499: Optional: you could write this as
```rust
        let mut data = 0u8;

        let len = self
            .device
            .read_block(cmd, core::slice::from_mut(&mut data))
            .map_err(|code| Error::BadRead { cmd, code })?;
```
which saves having to index `data[0]` elsewhere in the function.

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -350,6 +440,262 @@ impl Mwocp68 {
...
+
+            let checksum = match state {
+                Some(UpdateState::WroteLastBlock { checksum }) => checksum,
+                _ => panic!(),
+            };
+
+            data[0] = CommandCode::IMAGE_CHECKSUM as u8;
+            data[1] = 2;
+            data[2] = (checksum & 0xff) as u8;
+            data[3] = ((checksum >> 8) & 0xff) as u8;
```
hawkw, line 617: take it or leave it, but why not:
```suggestion
            let checksum = match state {
                Some(UpdateState::WroteLastBlock { checksum }) => checksum,
                _ => panic!(),
            };
            
            let data: [u8; 4] = [
                CommandCode::IMAGE_CHECKSUM as u8,
                2,
                (checksum & 0xff) as u8,
                ((checksum >> 8) & 0xff) as u8,
            ];
```

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+                ringbuf_entry!(Trace::FirmwareRevFailed(ndx, err));
+                return false;
+            }
+        }
+
+        if let (Some(started), Some(backoff)) =
+            (self.update_started, self.update_backoff)
+        {
+            if started + backoff > now {
+                ringbuf_entry!(Trace::BackingOff(ndx));
```
hawkw, line 179: i wonder if we might want to include the number of remaining ticks to back off for in the ringbuf entry? i don't _think_ that would make the entry enum bigger, given that we have other variants that include an index and error code?

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -350,6 +440,262 @@ impl Mwocp68 {
...
+            }
+        };
+
+        let send_checksum = || -> Result<UpdateState, Error> {
+            let mut data = [0u8; 4];
+
+            let checksum = match state {
+                Some(UpdateState::WroteLastBlock { checksum }) => checksum,
+                _ => panic!(),
+            };
```
mkeeter, line 612:
```suggestion
            let Some(UpdateState::WroteLastBlock { checksum }) = state else {
                panic!();
            };
```

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -27,15 +31,64 @@ pub struct Mwocp68 {
...
+    UpdateNotInBootLoader,
+    UpdateAlreadySuccessful,
+    BadBootLoaderStatus {
+        data: u8,
+    },
+    BadBootLoaderCommand {
+        cmd: BootLoaderCommand,
+        code: ResponseCode,
+    },
+    ChecksumNotSuccessful,
```
hawkw, line 81: here we've added a bunch of error variants that are only ever returned by specific functions, but are part of the error enum that's returned by _all_ functions in this module, forcing callers of those functions to handle the new update-specific variants. i'm not sure how important this is in practice, since presently, I think `sensor-polling` is the only caller of other methods on the `Mwocp68` type, and it handles any unexpected variant by just recording a `DeviceError` (rather than e.g. panicking)...so, it's probably fine.  but i figured it was worth mentioning.

hawkw (commented):
this looks good to me! i had some very minor drive-by suggestions which may or may not be useful.

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -350,6 +440,262 @@ impl Mwocp68 {
...
+        };
+
+        if self.get_boot_loader_mode()? != expected {
+            return Err(match expected {
+                Mode::BootLoader => Error::UpdateNotInBootLoader,
+                Mode::NotBootLoader => Error::UpdateInBootLoader,
+            });
+        }
+
+        let next = match state {
```
mkeeter, line 682: This is an interesting architecture!

Putting stuff into helper functions makes this block less wordy, but it means that there's potential panics in the helper functions (e.g. if `send_checksum` is called from a state other than `WroteLastBlock`).  In practice, I expect the compiler to inline + optimize out the panic paths.

One alternative would be extracting the key data in this `match`, then passing it into the helper functions (so that they don't need to unpack from `state`), e.g.
```rust
let next = match state {
    Some(UpdateState::WroteLastBlock { checksum }) => send_checksum(checksum)?,
    // ... etc
```

(very optional)

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+use drv_i2c_devices::Validate;
+use ringbuf::*;
+use static_cell::ClaimOnceCell;
+use userlib::*;
+
+use core::ops::Add;
+
+task_slot!(I2C, i2c_driver);
+
+const TIMER_INTERVAL: u64 = 10000;
```
mkeeter, line 21:
```suggestion
const TIMER_INTERVAL: u64 = 10_000;
```
(consider renaming to `TIMER_INTERVAL_MS` as well?)

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    UpdateFailedState(Option<UpdateState>),
+    UpdateFailure(Mwocp68Error),
+    UpdateState(UpdateState),
+    WroteBlock,
+    UpdateSucceeded,
+    UpdateDelay(u64),
+}
+
+const MWOCP68_FIRMWARE_REV: Mwocp68FirmwareRev = Mwocp68FirmwareRev(*b"0762");
+const MWOCP68_FIRMWARE_PAYLOAD: &'static [u8] =
```
mkeeter, line 68: Nit: I believe `'static` is unnecessary here.

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+                    ringbuf_entry!(Trace::UpdateFailedState(state));
+                    ringbuf_entry!(Trace::UpdateFailure(err));
+
+                    self.update_failure = Some((Ticks::now(), state, err));
+                    break;
+                }
+
+                Ok((UpdateState::UpdateSuccessful, _)) => {
+                    ringbuf_entry!(Trace::UpdateSucceeded);
+                    self.update_succeeded = Some(Ticks::now());
```
mkeeter, line 214: I think we should reset `self.update_backoff` here, to catch the case of someone updating a PSU (possibly with failures, leading to backoff winding up), then swaps a new PSU into that slot in the 10-second window.

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+        self.firmware_revision = None;
+
+        if !dev.present() {
+            self.present = Some(false);
+
+            //
+            // If we are seeing our device as not present, we will clear our
+            // backoff value: if/when a PSU is plugged back in, we want to
+            // attempt to update it immediately if the firmware revision
+            // doesn't match our payload.
```
mkeeter, line 136: This doesn't seem totally robust: if someone swaps a PSU during the 10-second wait between checks, we won't notice that it's gone (and therefore will use the original `update_backoff`).  Should we be storing the serial number (`MFR_SERIAL`) to catch this edge case?

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -350,6 +440,262 @@ impl Mwocp68 {
...
+                }
+            }
+        };
+
+        let reboot_psu = || -> Result<UpdateState, Error> {
+            self.boot_loader_command(BootLoaderCommand::BootPSUFirmware)?;
+            Ok(UpdateState::RebootedPSU)
+        };
+
+        let verify_success = || -> Result<UpdateState, Error> {
```
mkeeter, line 646: Should we read the power-good line here, to make sure the rectifier is back up?

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    });
+
+    loop {
+        hl::sleep_for(TIMER_INTERVAL);
+
+        for (ndx, psu) in psus.iter_mut().enumerate() {
+            let dev = &devs[ndx];
+
+            if psu.update_should_be_attempted(dev, ndx as u8) {
+                psu.update_firmware(dev, ndx as u8);
```
mkeeter, line 256: If a firmware update completes, but the rectifier doesn't report power-good (https://github.com/oxidecomputer/hubris/pull/1934/files#r1854779375), we should probably stop rather than rolling it out to all of the PSUs.  This would look like returning a status code from `update_firmware`, and bailing out (into an infinite sleep?) under certain error conditions.

drv/i2c-devices/src/mwocp68.rs
bcantrill: I actually went the other way on this, and broke everything (except for the two `WroteBlock` variants) out as their own delay.

drv/psc-psu-update/src/main.rs
bcantrill: I want to avoid flooding the ringbuf with entries (the remaining ticks will change over time). Will add a comment to that effect.

drv/i2c-devices/src/mwocp68.rs
bcantrill: I had actually coded it this way originally, but felt it was too cluttered.  As you say, I would expect the compiler to eliminate the potential panics...

drv/i2c-devices/src/mwocp68.rs
bcantrill: I tried this on, but it left it out of step with the way the other commands are structured.

drv/i2c-devices/src/mwocp68.rs
bcantrill: Yeah, I think it's okay -- and I really want to have the additional context that we log in the update task.

drv/psc-psu-update/src/main.rs
bcantrill: Yes, but also:

2024-11-22-19-59-46_1041x31

At least they define the length! And hard to see how it will be worse to track it...

drv/psc-psu-update/src/main.rs
bcantrill: Added logic here. This really (really) Shouldn't Happen, so if we see PG not set on a PSU we just successfully updated, we will wait until we do see PG become set.

drv/psc-psu-update/src/main.rs
bcantrill: Agreed -- fixed.

drv/i2c-devices/src/mwocp68.rs
bcantrill: Actually, I changed my mind: I like it better your way. ;)

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -350,6 +440,262 @@ impl Mwocp68 {
...
+        ringbuf_entry!(Trace::BootLoaderMode(data[0]));
+
+        match BOOT_LOADER_STATUS::CommandData::from_slice(&data[0..]) {
+            Some(status) => Ok(status),
+            None => Err(Error::BadBootLoaderStatus { data: data[0] }),
+        }
+    }
+
+    fn get_boot_loader_mode(&self) -> Result<BOOT_LOADER_STATUS::Mode, Error> {
+        Ok(self.get_boot_loader_status()?.get_mode().unwrap())
```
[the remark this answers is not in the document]
bcantrill: Done

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -27,13 +27,68 @@ pub struct Mwocp68 {
...
+pub struct SerialNumber(pub [u8; 12]);
+
+//
+// The boot loader command -- sent via BOOT_LOADER_CMD -- is unfortunately odd
+// in that its command code is overloaded with BOOT_LOADER_STATUS.  (That is,
+// a read to the command code is BOOT_LOADER_STATUS, a write is
+// BOOT_LOADER_CMD.)  This is behavior that the PMBus crate didn't necessarily
+// envision, so it can't necessarily help us out; we define the single-byte
+// payload codes here rather than declaratively in the PMBus crate.
+//
```
mkeeter, line 43: Nit: triple `///` to actually generate a docstring, which is highlighted differently for those of us using syntax highlighting

(Also, leading and trailing empty lines are unidiomatic, but I know it will take some prying to remove them from your hands)

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -62,6 +117,66 @@ impl From<pmbus::Error> for Error {
...
+
+    /// The boot loader has been booted
+    BootedBootLoader,
+
+    /// Programming of firmware has been indicated to have started
+    StartedProgramming,
+
+    /// A block has been written; the next offset is at [`offset`], and the
+    /// running checksum is in [`checksum`]
+    WroteBlock { offset: usize, checksum: u64 },
```
mkeeter, line 140: Nit: it looks like we only use a `u16` for the checksum, so we could store a `u16` and use `wrapping_add` (instead of unchecked arithmetic and trusting that we'll never overflow a `u64`).

drv/i2c-devices/src/mwocp68.rs
```diff
@@ -62,6 +117,66 @@ impl From<pmbus::Error> for Error {
...
+
+    /// The last block has been written; the checksum is in [`checksum`]
+    WroteLastBlock { checksum: u64 },
```
mkeeter, line 143: Same here

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,389 @@
...
+                                ringbuf_entry!(Trace::PGError(ndx, err));
+                            }
+                        }
+
+                        hl::sleep_for(TIMER_INTERVAL_MS);
+                    }
+
+                    ringbuf_entry!(Trace::UpdateSucceeded);
+                    self.update_succeeded = Some(Ticks::now());
+                    self.update_backoff = None;
```
mkeeter, line 343: Clear `serial_number` here as well, to avoid confusion when moving to the next PSU?

mkeeter: Whoops, never mind, this is tracked on a per-PSU basis

mkeeter (approved):

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,389 @@
...
+#[derive(Copy, Clone, Debug, PartialEq, counters::Count)]
+enum Trace {
+    None,
+    PowerGoodFailed(u8, drv_i2c_devices::mwocp68::Error),
+    FirmwareRevFailed(u8, drv_i2c_devices::mwocp68::Error),
+    AttemptingUpdate(u8),
+    BackingOff(u8),
+    UpdateFailed,
+    UpdateFailedState(Option<UpdateState>),
+    UpdateFailure(Mwocp68Error),
```
hawkw, line 84: Perhaps we ought to put a `#[count(skip)]` here to count the error variants?

drv/psc-psu-update/src/main.rs
```diff
@@ -0,0 +1,389 @@
...
+        update_started: None,
+        update_succeeded: None,
+        update_failure: None,
+        update_backoff: None,
+    }; 6],
+);
+
+#[derive(Copy, Clone, Debug, PartialEq, counters::Count)]
+enum Trace {
+    None,
```
hawkw, line 77: should probably skip this.
```suggestion
    #[count(skip)]
    None,
```

drv/psc-psu-update/src/main.rs
hawkw: Good point, thanks for adding the comment!

hawkw (approved):

drv/i2c-devices/src/mwocp68.rs
bcantrill: It's definitely not intended to be a doc comment, just an implementation comment -- which hopefully gives me a stay of execution on the added vertical spacing for readability. ;)

drv/i2c-devices/src/mwocp68.rs
bcantrill: I don't want to chuck the higher bits of the checksum even though the part only wants the low 16 -- but agreed on the unchecked arithmetic; what about using a `wrapping_add` on the checksum?

drv/psc-psu-update/src/main.rs
bcantrill: Slightly confused; do you mean to add a ` #[count(children)]` to the members?

drv/i2c-devices/src/mwocp68.rs
mkeeter: Yeah, sounds good

PR #1936, opened 2024-11-25
build/util: prevent infinite loop if target/ not found

cbiffle:
Previously, if you attempted to use certain functions in the build_util crate...and your output directory happened to not include any parent directories with the exact name "target"...

...the build script would burn 100% of a core forever, until you killed it.

This doesn't affect Hubris-proper but _does_ affect out of tree builds. The original code was arguably wrong anyway, so, this fixes that.


hawkw (approved):
whoops! good catch!

labbott (approved):

PR #1937, opened 2024-11-25
userlib: remove panics in kipc implementation

cbiffle:
The kipc stubs tend to assert on the rc value received from the kernel. This isn't ideal: we generally have to trust the kernel, and we can't do anything useful if it misbehaves. So the assert is merely costing a bunch of flash for a condition that can never happen. This removes these asserts.

The system_restart kipc needs to indicate a "noreturn" send, but of course we have no way of doing that right now. So, it panicked if the send returned. This created _another_ can't-happen panic site in flash. I've replaced it with an infinite loop that compiles to a two-byte branch instruction, and in practice will never even be reached (but the compiler doesn't know that).

This further reduces text and stack size in El Jefe.

### DIFF
--- a/sys/userlib/src/kipc.rs
+++ b/sys/userlib/src/kipc.rs
@@ -11,7 +11,7 @@
 //! IPC is defined as able to fail. We have no choice but to trust the kernel,
 //! since it controls everything.
 //!
-//! As a result, asserting on return codes when they can only be zero just
+//! As a result, asserting on return codes and lengths when they can only be zero just
 //! wastes flash space in the supervisor.
 
 use core::num::NonZeroUsize;
--- a/sys/userlib/src/kipc.rs
+++ b/sys/userlib/src/kipc.rs
@@ -3,6 +3,16 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 //! Operations implemented by IPC with the kernel task.
+//!
+//! # On checking return values
+//!
+//! All the functions in this module send IPCs to the kernel directly. It's not
+//! generally useful for us to check the return codes, except in cases where the
+//! IPC is defined as able to fail. We have no choice but to trust the kernel,
+//! since it controls everything.
+//!
+//! As a result, asserting on return codes when they can only be zero just
+//! wastes flash space in the supervisor.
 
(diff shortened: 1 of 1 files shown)


sys/userlib/src/kipc.rs
```diff
@@ -3,6 +3,16 @@
...
+//!
+//! # On checking return values
+//!
+//! All the functions in this module send IPCs to the kernel directly. It's not
+//! generally useful for us to check the return codes, except in cases where the
+//! IPC is defined as able to fail. We have no choice but to trust the kernel,
+//! since it controls everything.
+//!
+//! As a result, asserting on return codes when they can only be zero just
+//! wastes flash space in the supervisor.
```
hawkw, line 15: considered pedantic: perhaps 
```suggestion
//! As a result, asserting on return codes and lengths when they can only be zero just
//! wastes flash space in the supervisor.
```
or similar, given that we've also removed some assertions about response lengths?

hawkw (commented):
The rationale for removing these assertions makes sense to me, thanks for documenting it!

Out of curiosity: can Jefe stack size allocations reasonably be disembiggened in any of our app.tomls now?

hawkw (approved):
whoops, forgot to click "approve"!

cbiffle:
> Out of curiosity: can Jefe stack size allocations reasonably be disembiggened in any of our app.tomls now?

@hawkw probably, but the process of finding them and right-sizing them is unpleasant and I'm currently not up for it.

hawkw:
> > Out of curiosity: can Jefe stack size allocations reasonably be disembiggened in any of our app.tomls now?
> 
> 
> 
> @hawkw probably, but the process of finding them and right-sizing them is unpleasant and I'm currently not up for it.

That's entirely reasonable; I wasn't going to say you had to do it now, or anything. And, I think it can be justified by saying that, in the absence of better automated processes for estimating max stack depths, giving the supervisor a bigger stack *margin* is probably more useful unless we are actively trying to scrounge up a few more bytes of RAM for some specific purpose, which we're not at present!

cbiffle:
@hawkw if you like getting stack back, you're going to love my next PR. ;-)

PR #1939, opened 2024-12-02
oxide-barcode: allow shorter serial and part numbers

### LINKED ISSUE #1893: Prepare oxide-barcode for Terra, v2 serial numbers
In the OXV2 barcode format, there are a series of `:` separated fields that are used to distinguish between the part number, revision, and serial number respectively. Hubris parses this VPD encoded line using the internal `oxide-barcode` crate.

While this crate does split on the `:` characters, it is making internal assumptions about the number of characters that will and won't be present in various parts here. See for example, some of the logic in [VpdIdentity parse function](https://github.com/oxidecomputer/hubris/blob/eeaf69f2c4fb880c2f2773d89f53acf6ee4c2c11/lib/oxide-barcode/src/lib.rs#L52-L84).

The changes that are happening with these is that Terra CPNs are going to be shorter and the v2 serial numbers are also shorter. From looking at where these are all consumed in the system through both the inventory data and the SP status messages, it seems that those all are using larger fields for the serial and part numbers, such that shorter ones should be fine. The main issue that we'll need to sort through is how we want to change the parsing logic and the `VpdIdentity` structure.

Thinking out loud, it seems likely that we can just retain the existing structure and mostly change the logic in [this case](https://github.com/oxidecomputer/hubris/blob/eeaf69f2c4fb880c2f2773d89f53acf6ee4c2c11/lib/oxide-barcode/src/lib.rs#L77-L82) to deal with dynamic lengths and perhaps error if the length is larger than the current set aside field, but otherwise zero pad the rest.

cbiffle:
This retains the original maximum length for both fields, but permits parsed strings to contain shorter sections. Shorter sections are right-padded with NULs for consistency with IPCC.

Fixes #1893

### DIFF
--- a/lib/oxide-barcode/src/lib.rs
+++ b/lib/oxide-barcode/src/lib.rs
@@ -60,6 +60,8 @@ impl VpdIdentity {
             return Err(ParseError::UnexpectedFields);
         }
 
+        // Note: the fact that this is created _zeroed_ is important for the
+        // variable length field handling below.
         let mut out = VpdIdentity::new_zeroed();
 
         match version {
@@ -75,10 +77,12 @@ impl VpdIdentity {
             }
             // V2 part number includes the hyphen; copy it as-is.
             b"OXV2" | b"0XV2" => {
-                if part_number.len() != out.part_number.len() {
+                if part_number.len() > out.part_number.len() {
                     return Err(ParseError::WrongPartNumberLength);
                 }
-                out.part_number.copy_from_slice(part_number);
+                out.part_number[..part_number.len()]
+                    .copy_from_slice(part_number);
+                // tail is already zeroed due to use of new_zeroed above
             }
             _ => return Err(ParseError::UnknownVersion),
         }
@@ -88,11 +92,11 @@ impl VpdIdentity {
             .and_then(|rev| rev.parse().ok())
             .ok_or(ParseError::BadRevision)?;
 
-        if serial.len() != out.serial.len() {
+        if serial.len() > out.serial.len() {
             return Err(ParseError::WrongSerialLength);
         }
-
-        out.serial.copy_from_slice(serial);
+        out.serial[..serial.len()].copy_from_slice(serial);
+        // tail is already zeroed
 
         Ok(out)
     }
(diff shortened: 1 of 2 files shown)


mkeeter (approved):

jgallagher (approved):

hawkw (commented):
Handful of *very* unimportant nits, feel free to disregard me.

lib/oxide-barcode/src/lib.rs
```diff
@@ -102,39 +106,82 @@ impl VpdIdentity {
...
+
+        // We accept barcode strings that start with both leading zero and
+        // leading capital-O. Permute our input from one of these to the other
+        // to make sure both forms parse equivalently.
+        let mut copy = input.to_vec();
+        match copy[0] {
+            b'0' => copy[0] = b'O',
+            b'O' => copy[0] = b'0',
+            c => {
+                panic!("unexpected leading character in string: {}", c as char)
```
hawkw, line 126: nit: this wording feels slightly confusing, as the assertion a few lines above says "parsing string: {the input string}". this panic message kind of reads like the unexpected character was the _whole_ string, and we never print the whole string here. IMO, it might make more sense to say something like
```suggestion
                panic!("unexpected leading character {} in string: {}", c as char,, String::from_utf8_lossy(input))
```
or, just drop the "in string:" part if we don't think showing the whole string is important here (and it's probably not, since the `#[track_caller]` will make it clear which test failed...):
```suggestion
                panic!("unexpected leading character: {}", c as char)
```

hawkw: (this was supposed to be a review comment but GitHub seems to have eaten it somehow...)

citrus-it:
I flashed these changes to a rev B bench gimlet which has an unlocked VPD, then replaced the VPD contents with a v2 barcode containing short part and serial numbers:

```
% pfexec humility vpd -i 10 -r
[
    ("FRU0", [
        ("BARC", [
            "0XV2:9131234:000:EVT0006",
        ]),
        ("MAC0", [
            [
                168,
                64,
                37,
                1,
                1,
                3,
                8,
                0,
                8,
            ],
        ]),
    ]),
```

and booted. The board properties (and their length) printed as the kernel starts up look good:

```
Boot properties:
        0xb3e8 baseboard-revision = len=4 00000000
        0xb398 baseboard-model = len=8 9131234
        0xb348 baseboard-identifier = len=8 EVT0006
```

and sending an IPCC ident command from the running OS also shows the expected results:

```
# /usr/platform/oxide/bin/ipcc ident
Serial: 'EVT0006'
Model:  '9131234'
Rev:    0x0
```

The `t6init` service detected that the data encoded in the T6 SROM was no longer correct, and replaced it. That also worked correctly:

```
# t6mfgadm srom vpd show -d 1
OFFSET  FLAGS   ID              PN              SN              MAC
0x0400  CIPS-   T62100-KR       9131234         EVT0006         -
0x0800  CIPSM   T62100-KR       9131234         EVT0006         a8:40:25:1:1:13
0x0c00  CIPS-   T62100-KR       9131234         EVT0006         -
0x1000  CIPSM   T62100-KR       9131234         EVT0006         a8:40:25:1:1:13
0x1400  CIPS-   T62100-KR       9131234         EVT0006         -
0x1800  CIPSM   T62100-KR       9131234         EVT0006         a8:40:25:1:1:13
0x1c00  CIPS-   T62100-KR       9131234         EVT0006         -
0x2000  CIPSM   T62100-KR       9131234         EVT0006         a8:40:25:1:1:13
0x2400  CIPS-   T62100-KR       9131234         EVT0006         -
0x2800  CIPSM   T62100-KR       9131234         EVT0006         a8:40:25:1:1:13
0x2c00  CIPS-   T62100-KR       9131234         EVT0006         -
0x3000  CIPSM   T62100-KR       9131234         EVT0006         a8:40:25:1:1:13
0x3400  CIPS-   T62100-KR       9131234         EVT0006         -
0x3800  CIPSM   T62100-KR       9131234         EVT0006         a8:40:25:1:1:13
0x3c00  CIPS-   T62100-KR       9131234         EVT0006         -
0x4000  CIPSM   T62100-KR       9131234         EVT0006         a8:40:25:1:1:13
```

lib/oxide-barcode/src/lib.rs
cbiffle: I've taken the latter diff, since this is checking correctness of test fixtures only.

PR #1942, opened 2024-12-09
Sidecar FPFAs: pick up I2C SCL stretch fix

Aaron-Hartwig:
Pickup https://github.com/oxidecomputer/quartz/commit/d4c2c020ceff20e77dcae2c82af276f426cf0201.

I'm updating the Front IO FPGAs as they're the primary thing impacted here, but I'm also updating the mainboard FPGA as it has an I2C interface to the Tofino 2.

The FPGAs are no longer built with Buildomat so I've updated their READMEs accordingly. Additionally, I am pulling in an updated front IO register map after the fix in https://github.com/oxidecomputer/quartz/commit/39d93cd4640a1a8a60d00dad554e1c1224a3171f which results in a bunch of erroneous code getting pulled out of the HTML file.


nathanaelhuffman (approved):
lgtm. My only comment is that the artifacts at the GHA links you're referencing aren't permanent and will be rolled over within 90 days.  If you want a permanent link to the artifacts, I think we'd have to make a quartz gh release and upload the build results there.  I have open issues to improve this process but we don't have an automated solve here yet so FYI.

Aaron-Hartwig:
I updated the readme to point to the commit itself for now. We can always recreate the same image given the git rev and OS/toolchain.

PR #1945, opened 2024-12-12
stm32h7-rng: rework driver, add seed error recovery

cbiffle:
We have a machine in manufacturing that gets periodic seed errors. The manual specifies a recovery procedure for seed errors... which we haven't implemented.

This commit makes us more tolerant of seed errors by attempting recovery, but not twice in a row. Two seed errors in a row will still return an error.

This also adds a ringbuf so we can monitor frequency.

### DIFF
--- a/drv/stm32h7-rng/src/main.rs
+++ b/drv/stm32h7-rng/src/main.rs
@@ -13,6 +13,8 @@ use drv_rng_api::RngError;
 use drv_stm32xx_sys_api::{Peripheral, Sys};
 use idol_runtime::{ClientError, NotificationHandler, RequestError};
 
+use ringbuf::{counted_ringbuf, ringbuf_entry};
+
 #[cfg(feature = "h743")]
 use stm32h7::stm32h743 as device;
 
@@ -23,10 +25,19 @@ use userlib::*;
 
 task_slot!(SYS, sys);
 
+counted_ringbuf!(Trace, 32, Trace::Blank);
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, counters::Count)]
+enum Trace {
+    #[count(skip)]
+    Blank,
+    ClockError,
+    SeedError,
+    Recovered,
+}
+
 struct Stm32h7Rng {
-    cr: &'static device::rng::CR,
-    dr: &'static device::rng::DR,
-    sr: &'static device::rng::SR,
+    regs: &'static device::rng::RegisterBlock,
     sys: Sys,
 }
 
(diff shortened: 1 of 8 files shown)


drv/stm32h7-rng/src/main.rs
```diff
@@ -23,73 +23,134 @@ use userlib::*;
 
 task_slot!(SYS, sys);
 
+ringbuf::ringbuf!(Trace, 32, Trace::Blank);
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+enum Trace {
+    Blank,
```
mkeeter, line 30: Nit: do you prefer `Blank` to `None`?  `Trace::None` is very commonly used as the default type for other ringbufs.

drv/stm32h7-rng/src/main.rs
```diff
@@ -23,73 +23,134 @@ use userlib::*;
...
-
-    fn is_clock_error(&self) -> bool {
-        // if clock error detection is disabled, CECS & CEIS won't be valid
-        if self.is_clock_error_detect() {
-            return self.sr.read().cecs().bits()
-                || self.sr.read().ceis().bits();
+    fn attempt_recovery(&mut self) -> Result<(), RngError> {
+        // The recovery procedure is in section 34.3.7 of the manual.
+        //
+        // First, we clear the SEIS flag. The CEIS and SEIS flags are both
```
mkeeter, line 132: Nit: this is done twice (both before returning `SeedError` and here).  I'd vote to only clear it here; the explanation of "Clear it so we don't repeat ourselves" doesn't make sense when we're about to return a `SeedError`.

drv/stm32h7-rng/src/main.rs
```diff
@@ -109,23 +170,47 @@ impl idl::InOrderRngImpl for Stm32h7RngServer {
...
+                    made_progress = false;
+                }
+                Err(e) => {
+                    // We have no way to recover from the other cases.
+                    return Err(e.into());
+                }
+            }
         }
 
         Ok(cnt)
```
mkeeter, line 216: It seems a little silly to return `Ok(cnt)` when it's structurally guaranteed be the same as `len`, but dunno if it's worth changing here.

drv/stm32h7-rng/src/main.rs
```diff
@@ -23,73 +23,134 @@ use userlib::*;
...
+                // want to write CEIS to 1 here.
+                self.regs.sr.modify(|_, w| {
+                    w.ceis().set_bit();
+                    w.seis().clear_bit();
+                    w
+                });
+                return Err(RngError::SeedError);
+            }
+            // If data is ready, we can yield it.
+            if sr.drdy().bit_is_set() {
```
mkeeter, line 103: When the RNG is ready, we can theoretically read the `DR` up to 4x.  I'm not sure if we want to, though!

drv/stm32h7-rng/src/main.rs
```diff
@@ -23,73 +23,134 @@ use userlib::*;
...
+        // rc_w0, meaning writing 1 has no effect, and 0 clears. So to clear
+        // SEIS, we have to _set_ CEIS too.
+        self.regs.sr.write(|w| {
+            w.seis().clear_bit();
+            w.ceis().set_bit();
+            w
+        });
+
+        // Next, we read 12 words to clear the pipeline. The number of words is
+        // 12. Not 11, not 10.
```
hawkw, line 142: i assume 12 comes from the manual?

drv/stm32h7-rng/src/main.rs
```diff
@@ -109,23 +170,47 @@ impl idl::InOrderRngImpl for Stm32h7RngServer {
...
+
+                    // Note that we successfully did a thing, for recovery
+                    // purposes.
+                    made_progress = true;
+                }
+                Err(RngError::SeedError) if made_progress => {
+                    // It may be worth attempting recovery. Recovery may not
+                    // succeed, in which case this'll just bomb right out of
+                    // here and send an error to the client.
+                    self.rng.attempt_recovery()?;
```
hawkw, line 203: i wonder if it's worth including a ringbuf entry for successful recoveries, so that we can distinguish between two subsequent seed errors in service of the same request, which result in a failure observed by the client, and seed errors that we successfully recovered from and kept going?

hawkw: as it stands currently, i think the ringbuf will basically just say
```
SeedError, SeedError, SeedError, SeedError
```
which doesn't make it very obvious whether we've encountered seed errors and recovered from them, or whether the RNG is just totally wedged

drv/stm32h7-rng/src/main.rs
cbiffle: I know, I really dislike this since None is imported in the prelude from Option. What I actually want is for ringbufs to be uninitialized until filled in so we can stop doing this, but for want of that...

drv/stm32h7-rng/src/main.rs
cbiffle: Yeah, I agree, the API is weird. I chose not to alter the API in this commit.

drv/stm32h7-rng/src/main.rs
cbiffle: I'm not actually clear on whether we can, if we (say) read 1 from it in a previous call. So I figured I'd just read one every time. It should leave DRDY set as I read repeatedly.

drv/stm32h7-rng/src/main.rs
```diff
@@ -23,73 +23,134 @@ use userlib::*;
...
+                //
+                // > When data is not ready (DRDY = 0) RNG_DR returns zero. It
+                // > is recommended to always verify that RNG_DR is different
+                // > from zero. Because when it is the case a seed error
+                // > occurred between RNG_SR polling and RND_DR output reading
+                // > (rare event).
+                //
+                // Why it's not sufficient to check SR.SEIS after reading DR, I
+                // couldn't tell you. But we'll implement their weird
+                // recommendation.
```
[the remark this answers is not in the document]
cbiffle: I agree. It's weird and underspecified. I'm following the manual until that's proven wrong.

drv/stm32h7-rng/src/main.rs
cbiffle: It does -- all of this is from the recovery procedure in the section I cited above.

drv/stm32h7-rng/src/main.rs
cbiffle: This is a good point, I think. Noting resumption of service is good. I specifically _did not_ include a "successful transaction" ringbuf entry because that tends to cause errors to get overwritten (looking at you, i2c_driver) -- but noting a successful recovery seems wise.

drv/stm32h7-rng/src/main.rs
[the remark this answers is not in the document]
hawkw: Yeah, fair enough.

drv/stm32h7-rng/src/main.rs
hawkw: Yup, I agree that we probably don't want to spam the ringbuf with every successful request!

drv/stm32h7-rng/src/main.rs
```diff
@@ -23,73 +23,134 @@ use userlib::*;
...
+            // and clock errors can only occur if the RNG kernel clock is
+            // waaaaay below the AHB clock, which we don't do. But, we're
+            // interested in finding out they _were happening_ if they happen,
+            // so:
+            if sr.ceis().bit_is_set() {
+                // TODO this should be an ereport
+                ringbuf::ringbuf_entry!(Trace::ClockError);
+                // Clear it so we don't repeat ourselves. The two writable bits
+                // in this register are write-one-to-preserve, so weirdly we
+                // want to write SEIS to 1 here.
```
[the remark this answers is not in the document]
cbiffle: There are a lot of registers on ST peripherals like this, for better or worse.

cbiffle:
Addressed the comments, PTAL.

drv/stm32h7-rng/src/main.rs
```diff
@@ -129,24 +132,20 @@ impl Stm32h7Rng {
...
+        // 12. Not 11, not 13. 12 is the number given in the manual with no
+        // explanation.
         for _ in 0..12 {
             let _ = self.regs.dr.read();
         }
 
         // Finally, we check whether SEIS is clear.
         if self.regs.sr.read().seis().bit_is_clear() {
             // Yay!
+            ringbuf_entry!(Trace::Recovered);
```
hawkw, line 148: yay!

drv/stm32h7-rng/src/main.rs
```diff
@@ -23,73 +25,131 @@ use userlib::*;
 
 task_slot!(SYS, sys);
 
+counted_ringbuf!(Trace, 32, Trace::Blank);
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Count)]
+enum Trace {
+    Blank,
```
hawkw, line 32: since this is a `counted_ringbuf` now, the `Blank` variant ought to have 
```suggestion
    #[count(skip)]
    Blank,
```

hawkw (approved):
huh, CI seems sad for some reason: https://github.com/oxidecomputer/hubris/actions/runs/12323310764/job/34398699576?pr=1945

Once that's resolved, this looks good to me!

PR #1946, opened 2024-12-13
Measure the SP on SP_RESET signal interrupt

### LINKED ISSUE #2020: Potential Attest.record changes
## Background

With #1946, the `swd` task will record a measurement of the active SP flash bank whenever the SP_RESET pin is asserted.

If the RoT `swd` task is not finished initializing, making it ready to detect SP_RESET, then there will be no measurement recorded until the next time the SP is reset and the RoT is ready.

If the control plane wants an initial or fresher measurement, it can request that the SP reset itself. 
This scheme prevents interrupting the SP at some critical time that the RoT would not be aware of.
The control plane, or an individual host, may reboot itself if it needs to be measured by the SP.
In the future, we may choose to carefully sequence RoT and SP power-on boot to avoid this dance but that still leaves the case where RoT firmware has been updated with the SP necessarily still running to help orchestrate the RoT update.

## The issue

Given that:
  -  the RoT and SP can normally be up and running for some time without any entries in the attestation log,
  -  there is support for the SP to also send measurements to the attestation log,
  -  the log currently does not identify the source of a measurement

it is important to prevent a buggy or malicious SP image from recording a known good measurement that would imply that the SP was running a trustworthy image. A malicious SP image can boot, query the attestation log, reset the RoT if there is an entry, then record a fake entry into the first slot and proceed, a wolf in sheep's clothing.

## Implementation options

To that end we have a few choices (and please add any alternatives that you think of)
  0. keep existing implementation 
  1. tag attestation log entries with their source
  2. assign slots to sources
  3. prevent any source but the `swd` task from recording the first measurement

### No change
We can delay making changes but I don't see how to avoid changing log semantics to secure the system. 

### Tagged Entries
In choce one, the attest task labels entries with their source. We have not determined all the measurements that will ever go into the log. There may be motivation for a source-supplied tag as well.
 
### Slot Indices Are Assigned
Number two changes log semantics from being append style to being an array of `Option<Measurement>`. This choice incurs the burden of managing slot assignments which we may want to change over time.

### Assign first slot
Number three is a form of slot assignment and says that the SP is not permitted to make measurements unless it itself has been measured.  The `record` function would check the log size and refuse all but the `swd` task if the log was empty. (The allowed task list is configurable in `app.toml`.)

## How to limit attestation slot usage

It is highly desirable, even required, to have a trustworthy measurement of every bit of mutable state that affects the behavior of the system at power-on boot or reset at least up to the point of executing applications in the host OS. It is not necessary or even practical for all but of the few early measurements to be recorded with a RoT. That said, there may still be enough to complicate the management of the RoT's attestation log.

The number of attestation log entries ever used can be limited by having SP and higher layers, if any, only record a hash of their set of measurements. Then the attestation log does not need to manage a complex namespace. Each layer would need to be able to produce the measurement set when queried. The measurements are presumed to have been taken at a time when the reporting system can be trusted and has been measured by its predecessor in the chain of trust.

# Recommendations

Tentatively: Only the `swd` task should be allowed to record the first attestation log entry. If any other source attempts to record the first entry, it will be rejected.

However, the problem needs to be discussed to make sure we are properly covering our use cases and security concerns.

lzrd:
On SP_RESET asserted, the RoT will measure the entire active SP Hubris bank (Hubris plus 0xff padding from 0x800_0000 to 0x80f_ffff).
The measurement reliably takes about 0.482 seconds since neither the SP nor the RoT are doing anything else during this time. This low measurement time and the existing timeouts and retries used by the control plane mitigate concerns that were present when the duration was measured in multiple seconds.

This time could be further reduced. It is currently taking 0.245 seconds to inject the measurement program into the SP. Each u32 is being sent and read back separately and retried if there is any error (no errors have been seen as of yet).

If all is well with the SP, the measurement will match the `Sha3_256` value represented in the SP Hubris archive in the file named "img/final.fwid".
 
# Testing SP measurement

Checkout hubris branch `attest-sp`

```sh
APP=$PWD/app/oxide-rot-1/app-dev.toml
ARCHIVE=$(cargo -q xtask print --archive --image-name a $APP)
cargo xtask flash $APP
```

## Measure on SP Reset:

Hit the reset button on the SP or use the SpCtrl.db_reset_sp function:
```sh
humility --archive=$ARCHIVE hiffy -c SpCtrl.db_reset_sp -a delay_ms=10
sleep 13
```

## See traces
The ringbuf trace in `swd` and `attest` tasks have the interesting information including the time expended:

```sh
humility --archive=$ARCHIVE ringbuf
```

## Dump the attestation log:

```sh
humility --archive=$ARCHIVE hiffy -c Attest.log -a offset=0 -n 256 -o out
hexdump -C out
```
The hexdump is not hard to read directly. The first four bytes are a LE u32 giving
the number of valid entries. Each entry is a 1-byte discriminator that is 0_u8 for `Sha3_256` followed by the 32-byte measurement (FWID). There is space for 16 measurements.

## Notes

  - The RoT only attempts a measurement if it detects an SP reset.
  - Existing measurements are cleared if an STLINK power-on is detected or if an STLINK is on during a measurement attempt. It is observed that powered-off STLINKs have no effect.
  - No SP reset can be captured until the `swd` task has initialized. If needed, the control plane can request the SP to reset itself which will result in the RoT measuring the SP.
  - The `attest` task has a `reset_log_and_record` function usable only by the `swd` task (and hiffy for debug builds).
  - Closes #2020 by reserving the first slot of the attestation log for the `swd` task.


drv/lpc55-swd/build.rs
```diff
@@ -88,6 +97,163 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
...
+    for name in [
+        // Load address of image
+        (LOAD_SYMBOL, "LOAD"),
+        // Address of endoscope output struct
+        (SHARED_STRUCT_SYMBOL, "SHARED"),
+    ]
+    .iter()
+    {
+        interesting.insert(name.0, name.1);
+    }
```
mkeeter, line 133: Nit: you can do this immutably with something like
```suggestion
    let interesting: std::collections::BTreeMap<_, _> = [
        // Load address of image
        (LOAD_SYMBOL, "LOAD"),
        // Address of endoscope output struct
        (SHARED_STRUCT_SYMBOL, "SHARED"),
    ]
    .into_iter()
    .collect();
```

drv/lpc55-swd/build.rs
```diff
@@ -88,6 +97,163 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
...
+
+    for sym in elf.syms.iter() {
+        if let Some(str) = elf.strtab.get_at(sym.st_name) {
+            let name = format!("{:#}", demangle(str));
+            if let Some(myname) = interesting.get(name.as_str()) {
+                writeln!(
+                    &mut file,
+                    "pub const {}:u32 = 0x{:x};",
```
mkeeter, line 141: Formatting nit:
```suggestion
                    "pub const {}: u32 = {:#x};",
```
(I know we run `rustfmt` later, but this was confusing on first glance)

I also changed the second format string, because you can get the leading `0x` automatically with the `#` directive.

drv/lpc55-swd/build.rs
```diff
@@ -88,6 +97,163 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
...
+
+    // We're also assuming that `vm_range`s go from 0 to bin_size.
+    // So, expect a panic if we now try to read outside those limits.
+    let mut bin = vec![0u8; bin_size];
+
+    for section in sections {
+        let file_range = section.file_range().unwrap();
+        let vm_range = section.vm_range();
+        reader.seek(SeekFrom::Start(file_range.start as u64))?;
+        if reader.read(&mut bin[vm_range.start..vm_range.end]).is_err() {
```
mkeeter, line 247:
```suggestion
        if reader.read_exact(&mut bin[vm_range.start..vm_range.end]).is_err() {
```
(the current code would silently allow for partial reads)

lib/endoscope/README.md
```diff
@@ -0,0 +1,201 @@
...
+  - Deassert the SP RESET signal
+  - ensure that the STM32 halted due to VC_CORERESET
+  - Inject the program into the STM32's RAM at the address "__vector_table".
+  - Set The STM32 core register PC/DR to the value in `__vector_table[1]`.
+  - Set The STM32 core register MSP to the value in `__vector_table[0]`.
+  - Set The STM32 VTOR to the value of the symbol "__vector_table".
+  - Continue execution
+  - Poll DHCSR for S_HALT state or timeout.
+  - The attestation log will be reset.
+  - If halted && `SHARED.magic == Shared::MAGIC` && `SHARED.state == State::Done`
```
mkeeter, line 37: Formatting consistency:
```suggestion
  - Inject the program into the STM32's RAM at the address `__vector_table`.
  - Set The STM32 core register PC/DR to the value in `__vector_table[1]`.
  - Set The STM32 core register MSP to the value in `__vector_table[0]`.
  - Set The STM32 VTOR to the value of the symbol `__vector_table`.
  - Continue execution
  - Poll DHCSR for S_HALT state or timeout.
  - The attestation log will be reset.
  - If halted && `SHARED.magic == Shared::MAGIC` && `SHARED.state == State::Done`
```

drv/lpc55-swd/src/main.rs
```diff
@@ -919,6 +1105,38 @@ impl ServerImpl {
...
+        // It is consuming about 0.25 seconds to inject the `endoscope` code.
+        let mut addr = addr;
+        if data.len() % 4 != 0 {
+            ringbuf_entry!(Trace::BadLen);
+            return Err(Ack::Fault);
+        }
+        let mut limit = 37;
+        for slice in data.chunks(4) {
+            let word =
+                u32::from_le_bytes(slice[0..=3].try_into().unwrap_lite());
```
mkeeter, line 1121:
```suggestion
            let word = u32::from_le_bytes(slice.try_into().unwrap_lite());
```
Hopefully the compiler is smart enough to treat this as infallible; if not, using `data.chunks_exact(4)` would probably work.

drv/lpc55-swd/src/main.rs
```diff
@@ -919,6 +1105,38 @@ impl ServerImpl {
...
+            let readback = self.read_single_target_addr(addr)?;
+            if readback != word {
+                ringbuf_entry!(Trace::Data {
+                    addr,
+                    data: readback,
+                    src: word
+                });
+                return Err(Ack::Fault);
+            } else if limit > 0 {
+                limit -= 1;
```
mkeeter, line 1132: Does `limit` do anything?  It looks like it just counts down from 37 to 0.

drv/lpc55-swd/src/sp.rs
```diff
@@ -0,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+//
+
+// Stuff we know about the SP and its Hubris images but would rather get
+// without copying it here.
+// We could extract these values from a representive SP hubris ELF file.
+pub const FLASH_START: u32 = 0x0800_0000;
+pub const FLASH_END: u32 = 0x0810_1000;
```
mkeeter, line 10: Why is this 4K above the actual end of flash?  A comment is probably warranted.

lib/endoscope-abi/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            let shared: &Shared = layout.into_ref();
+            Some(shared)
+        } else {
+            None
+        }
+    }
+}
+
+// Symbols relied on in the endoscope.elf file.
```
mkeeter, line 59: It feels a little weird to have these symbol strings in the `endoscope-abi` crate, since they're not actually part of the embedded ABI.  Should they be constants in the `build.rs` instead?

lib/endoscope/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+anyhow = { workspace = true }
+
+[[bin]]
+name = "endoscope"
+test = false
+doctest = false
+bench = false
+
+[features]
+# running in SRAM with default clocking takes about 6 seconds
```
mkeeter, line 41: I'd suggest deleting the SRAM code entirely to shrink down the PR (removing this feature, deleting the `-sram` linker script, hardcoding the remaining linker script name, etc).

lib/endoscope/src/main.rs
```diff
@@ -0,0 +1,115 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#![no_std]
+#![no_main]
+
+// We have to do this if we don't otherwise use it to ensure its vector table
```
mkeeter, line 8: What's with all of the commented out code here?

drv/lpc55-swd/build.rs
```diff
@@ -88,6 +97,163 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
...
+        (SHARED_STRUCT_SYMBOL, "SHARED"),
+    ]
+    .iter()
+    {
+        interesting.insert(name.0, name.1);
+    }
+
+    for sym in elf.syms.iter() {
+        if let Some(str) = elf.strtab.get_at(sym.st_name) {
+            let name = format!("{:#}", demangle(str));
```
mkeeter, line 137: I don't think `demangle` is needed here, because the relevant symbols are either from the linker script or marked with `#[no_mangle]`

lib/endoscope/src/shared.rs
```diff
@@ -0,0 +1,84 @@
...
+        unsafe {
+            assert_eq!((*shared_wrapper.shared.get()).state, 42);
+            assert_eq!((*shared_wrapper.shared.get()).digest[0], 1);
+        }
+    }
+}
+
+// Mark as used so that symbol remains in symbol table
+#[used]
+#[no_mangle]
```
mkeeter, line 83: I believe that `#[no_mangle]` implies `#[used]`, so you don't need both here.

lib/endoscope/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+#[entry]
+fn main() -> ! {
+    // Note: The RoT does not examine results until the SP is halted.
+    SHARED.set_state(State::Running);
+    let _p = system_init(CLOCK_CONFIG);
+    SHARED.set_flash_area();
+    let mut hash = Sha3_256::new();
+    let image = unsafe {
+        core::slice::from_raw_parts(SHARED.get_start(), SHARED.get_len())
+    };
```
mkeeter, line 81: I believe this is UB, because the slice data is not guaranteed to be initialized.  However, I'm not sure of the Right Way to do this; even `core::ptr::read_volatile` requires the pointed-to data be initialized.

task/attest/src/main.rs
```diff
@@ -272,6 +274,54 @@ impl idl::InOrderAttestImpl for AttestServer {
...
+        ringbuf_entry!(Trace::ResetRecord(algorithm));
+
+        if !PERMIT_LOG_RESET.iter().any(|x| *x == msg.sender.0) {
+            return Err(ClientError::AccessViolation.fail());
+        }
+        ringbuf_entry!(Trace::ResetLog);
+        self.measurements = Log::default();
+
+        let measurement = match algorithm {
+            HashAlgorithm::Sha3_256 => {
```
mkeeter, line 306: Should we instead make the `algorithm` argument into a data-bearing enum, e.g. `HashMeasurement::Sha3_256([u8; 32])`?

task/attest/build.rs
```diff
@@ -60,5 +67,24 @@ pub const ALIAS_DATA: DataRegion = DataRegion {{
...
+    // Get list of those tasks allowed to reset the attestation log
+    if let Ok(task_config) = build_util::task_config::<TaskConfig>() {
+        writeln!(
+            out,
+            "pub const PERMIT_LOG_RESET: [u16; {}] = [",
+            task_config.permit_log_reset.len()
+        )?;
+        let tasks = build_util::task_ids();
+        for task_name in task_config.permit_log_reset {
+            let id = tasks.get(task_name.as_str()).context(format!(
```
mkeeter, line 79: Nit: it's typical to use `with_context` if the context operation does work (like formatting a string).
```suggestion
            let id = tasks.get(task_name.as_str()).with_context(|| format!(
```

lib/endoscope-abi/src/lib.rs
lzrd: Thanks!

This is some speculative generalization that can be eliminated.

My thinking is that part of the ABI is where to inject the endoscope blob into the SP's memory, what its entry address is and where to find the results.
That allows good flexibility for developing the blob independent from the RoT code. So, the "private" symbol names need to be bound somehow to those used by the `swd` task.

Since drv/lpc55-swd/build.rs is the only user and endoscope is the only injected blob used in this way, I'm happy to move these to build.rs.

lib/endoscope/Cargo.toml
lzrd: I'm ok with deleting the *-sram linker script.
There is talk of someday using something other than the STM32H7 which may change the linker script. The overhead in lib/endoscope/build.rs is pretty small and makes it easier on anyone adding support for a new SP.

lib/endoscope/src/shared.rs
lzrd: You are correct: https://doc.rust-lang.org/reference/abi.html#the-no_mangle-attribute

task/attest/build.rs
lzrd: Thanks

task/attest/src/main.rs
lzrd: I think that filhp was thinking about the various formats that other attestation schemes need to support and that we might want to support in the future. That doesn't mean it could not be done with a data-bearing enum. I'm not motivate to change this right now but can check with @flihp

Cargo.toml
```diff
@@ -20,6 +20,30 @@ opt-level = "z" # smaller optimizations
 [profile.dev]
 opt-level = 1 # no optimizations was just too painful in terms of flash size
 
+# Per-package profiles do not currently allow lto, panic, or rpath.
+# This costs us an extra 456 bytes in the endoscope blob that is injected
+# into the SP to measure its active flash bank.
```
labbott, line 25: How close are we to hitting our limit? Do we really need to be concerned about this right now?

app/oxide-rot-1/app.toml
```diff
@@ -68,6 +69,7 @@ task-slots = ["syscon_driver"]
 [tasks.sprot]
 name = "drv-lpc55-sprot-server"
 priority = 6
+max-sizes = {flash = 54016, ram = 32768}
```
labbott, line 72: Do we really need to reintroduce these or can we rely on autosizing?

drv/lpc55-swd/build.rs
```diff
@@ -88,6 +107,165 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
...
+        reader.seek(SeekFrom::Start(file_range.start as u64))?;
+        reader
+            .read_exact(&mut bin[vm_range.start..vm_range.end])
+            .with_context(|| {
+                format!(
+                    "cannot read elf[{}..] into bin[{}..{}]",
+                    file_range.start, vm_range.start, vm_range.end
+                )
+            })?;
+    }
```
labbott, line 265: Was this code used to catch an actual bug? Under what circumstances would we expect to hit this?

drv/lpc55-swd/src/armv7debug.rs
```diff
@@ -0,0 +1,189 @@
...
+        const DWTTRAP = 1 << 2;
+        // Breakpoint
+        const BKPT = 1 << 1;
+        // Halt request debug event.
+        // • A C_HALT or C_STEP request, triggered by a write to the DHCSR,
+        //   see Debug Halting Control and Status Register, DHCSR.
+        // • A step request triggered by setting DEMCR.MON_STEP to 1,
+        //   see Debug monitor stepping on page C1-696.
+        const HALTED = 1 << 0;
+        const _ = !0;
```
labbott, line 143: Is this a quirk of bitflags?

drv/lpc55-swd/src/main.rs
```diff
@@ -919,6 +1103,40 @@ impl ServerImpl {
...
+        // TODO: Performance could be improved here. This is using
+        // single_read/single_write to write and verify every u32.
+        // It is consuming about 0.25 seconds to inject the `endoscope` code.
+        let mut addr = addr;
+        const U32_SIZE: usize = core::mem::size_of::<u32>();
+        if data.len() % U32_SIZE != 0 {
+            ringbuf_entry!(Trace::BadLen);
+            return Err(Ack::Fault);
+        }
+        let mut limit = 37;
```
labbott, line 1117: Can you add a comment explaining this limit?

drv/lpc55-swd/src/main.rs
```diff
@@ -954,6 +1172,27 @@ impl ServerImpl {
...
+            .map(|r| T::from_bits_retain(r))
+    }
+
+    fn dp_write_bitflags<T>(&mut self, val: T) -> Result<(), Ack>
+    where
+        T: bitflags::Flags<Bits = u32> + DpAddressable,
+    {
+        self.write_single_target_addr(T::addr(), val.bits())
+    }
+
```
labbott, line 1195: This is a really nice improvement to the interface

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+        // Set SP's Stack Pointer
+        let sp_initial_sp =
+            u32::from_le_bytes(ENDOSCOPE_BYTES[0..=3].try_into().unwrap_lite());
+        if self
+            .do_write_core_register(Reg::Sp.into(), sp_initial_sp)
+            .is_err()
+        {
+            ringbuf_entry!(Trace::WroteSpRegisterFail);
+            return Err(());
+        }
```
labbott, line 1364: Can you add a citation or short note about where the `0..=3` and `4..=7` come from for those who don't have vector tables memorized? Looks like B1.5.2 and B1.5.3 in the ARMv7m manual? 

With `sp_reset_vector | 1`, isn't the `| 1` redundant? I think the compiler should be generating proper code to set the thumb bit.

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+
+        // It takes about 0.25 seconds (236 RoT systicks) for `endoscope` to run.
+        // Allow about twice that time for the measurement to complete.
+        // endoscope executes a BKPT instruction on completion.
+        // We observe an S_HALT state if all goes well.
+        if self.halt_wait(512).is_err() {
+            ringbuf_entry!(Trace::DidNotHalt);
+            return Err(());
+        };
```
labbott, line 1405: What do we observe if endoscope either loops forever or hits a fault?

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+            ringbuf_entry!(Trace::Digest1(
+                shared.digest[0x08..=0x0f].try_into().unwrap_lite()
+            ));
+
+            ringbuf_entry!(Trace::Digest2(
+                shared.digest[0x10..=0x17].try_into().unwrap_lite()
+            ));
+            ringbuf_entry!(Trace::Digest3(
+                shared.digest[0x18..=0x1f].try_into().unwrap_lite()
+            ));
```
labbott, line 1429: That's a lot of unwraps, do we see any improvement in code size if we remove them?

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+                    endoscope::FLASH_BASE + sp::IMAGE_HEADER_LENGTH_OFFSET,
+                    total_image_size.as_bytes_mut(),
+                )
+                .is_err()
+            || !(sp::MIN_IMAGE_SIZE..(endoscope::FLASH_SIZE as usize))
+                .contains(&(total_image_size as usize))
+        {
+            ringbuf_entry!(Trace::SpHubrisFailsSanityCheck);
+            return Err(());
+        }
```
labbott, line 1522: What's the benefit to checking if the hubris image is valid? What is this going to look like from the higher levels?

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+    }
+
+    fn do_halt(&mut self) -> Result<(), SpCtrlError> {
+        ringbuf_entry!(Trace::DoHalt);
+        self.dp_write_bitflags::<Dhcsr>(Dhcsr::halt())
+            .map_err(|_| SpCtrlError::Fault)?;
+        self.halt_wait(5000)
+    }
+
+    fn halt_wait(&mut self, timeout: u64) -> Result<(), SpCtrlError> {
```
labbott, line 1621: I think I confused myself because `halt_wait` means wait for the system to halt, not attempt to halt and then wait. Can you either add a comment to this effect or change the name?

labbott (commented):
This looks really nice. Thanks for putting in the work to finally make the RoT a reality!

Either before we merge or right afterwards I think we should makes sure this gets run through a sample manufacturing flow to make sure we don't have any surprises there.

Cargo.toml
lzrd: I'll delete this. The same information is in endoscope's README.md and it's not a concern right now.

app/oxide-rot-1/app.toml
lzrd: They don't need to be here. Deleted.

drv/lpc55-swd/build.rs
lzrd: No bug was encountered. This just makes the assumptions clear.

drv/lpc55-swd/src/armv7debug.rs
lzrd: Yes. It avoids surprises with unspecified bits.

See https://docs.rs/bitflags/latest/bitflags/#externally-defined-flags

Cargo.toml
```diff
@@ -20,6 +20,30 @@ opt-level = "z" # smaller optimizations
...
+[profile.release.package.endoscope]
+codegen-units = 1
+debug = 2
+#lto = true
```
mkeeter, line 29: Nit: remove commented out parameters here?

drv/lpc55-swd/src/armv7debug.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub fn end_debug() -> Self {
+        Self::DBGKEY
+    }
+    pub fn is_halted(self) -> bool {
+        self & Self::S_HALT == Self::S_HALT
+    }
+    pub fn is_regrdy(self) -> bool {
+        self & Self::S_REGRDY == Self::S_REGRDY
+    }
+    pub fn _is_lockup(self) -> bool {
```
mkeeter, line 88: Why is there a leading underscore here?

drv/lpc55-swd/src/armv7debug.rs
```diff
@@ -0,0 +1,189 @@
...
+        self & Self::S_LOCKUP == Self::S_LOCKUP
+    }
+}
+
+// Debug Core Register Selector Register
```
mkeeter, line 93: I like the new `DpAddressable` + `bitflags!` design, so it's awkward that these two registers continue to use hard-coded values...

drv/lpc55-swd/src/armv7debug.rs
```diff
@@ -0,0 +1,189 @@
...
+// For the STM32H7xx, see ARMv7-M Architecture Reference Manual
+// Part 3 Debug Arch.
+// https://developer.arm.com/documentation/ddi0403/d/Debug-Architecture?lang=en
+
+use bitflags::bitflags;
+
+pub trait DpAddressable {
+    // Address for accessing a DP Register
+    fn addr() -> u32;
+}
```
mkeeter, line 15: Nit: you can put the `ADDRESS` constant into the trait, e.g.
```suggestion
pub trait DpAddressable {
    /// Address for accessing a DP Register
    const ADDRESS: u32;
}
```

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+            ringbuf_entry!(Trace::DidNotHalt);
+            return Err(());
+        };
+
+        let mut shared = Shared {
+            state: State::Preboot as u32,
+            digest: [0u8; 32],
+        };
+
+        if self
```
mkeeter, line 1412: Vibes: this section is hard to follow, because it logs digests even if the state is invalid.  I'd suggest reorganizing it to put the error handling first, e.g.
```rust
        if let Err(e) = self
            .read_buf_from_addr(endoscope::SHARED, shared.as_bytes_mut())
            .is_ok()
        {
            ringbuf_entry!(Trace::ReadSharedFailed(e));
            Err(())
        } else if shared.state != State::Done as u32 {
            ringbuf_entry!(Trace::InvalidState(shared.state))
            Err(())
        } else {
            ringbuf_entry!(Trace::SharedState(shared.state));
            ringbuf_entry!(Trace::Digest0(
                shared.digest[0x00..=0x07].try_into().unwrap_lite()
            ));
            ringbuf_entry!(Trace::Digest1(
                shared.digest[0x08..=0x0f].try_into().unwrap_lite()
            ));

            ringbuf_entry!(Trace::Digest2(
                shared.digest[0x10..=0x17].try_into().unwrap_lite()
            ));
            ringbuf_entry!(Trace::Digest3(
                shared.digest[0x18..=0x1f].try_into().unwrap_lite()
            ));
            Ok(shared.digest)
        }
```
(adding new made-up ringbuf entries, feel free to adjust them)

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+        Err(())
+    }
+
+    fn do_write_core_register(
+        &mut self,
+        register: u16,
+        value: u32,
+    ) -> Result<(), SpCtrlError> {
+        // C1.6 Debug system registers
+        let r = match register {
```
mkeeter, line 1443: Vibes: should we make the `enum Reg` include all of these values?  Then, we could make this function take a `register: Reg` instead of having to check here.

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+        addr: u32,
+        buf: &mut [u8],
+    ) -> Result<(), SpCtrlError> {
+        let start = addr;
+        let end = addr + buf.len() as u32;
+        self.start_read_transaction(start, ((end - start) as usize) / 4)
+            .map_err(|_| SpCtrlError::Fault)?;
+
+        let cnt = buf.len();
+        if cnt % 4 != 0 {
```
mkeeter, line 1551: Nit: should we do this check _before_ starting a read transaction, to avoid leaving the SP in a weird state?

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+        let gpio = Pins::from(self.gpio);
+        gpio.set_val(ROT_TO_SP_RESET_L_OUT, Value::Zero);
+    }
+
+    /// Cleanup is true if we don't want to be notified of this reset
+    fn sp_reset_leave(&mut self, squelch_notify: bool) {
+        let gpio = Pins::from(self.gpio);
+        setup_rot_to_sp_reset_l_in(self.gpio);
+        gpio.set_val(ROT_TO_SP_RESET_L_IN, Value::One); // should be a no-op
+        if squelch_notify {
```
mkeeter, line 1321: I'm confused by `squelch_notify` here; wouldn't the IRQ have already fired in `sp_reset_enter`?

drv/lpc55-swd/src/main.rs
```diff
@@ -512,28 +598,130 @@ impl idl::InOrderSpCtrlImpl for ServerImpl {
...
+            self.watchdog_ms = None;
+        }
+
+        if (bits & notifications::SP_RESET_IRQ_MASK) != 0 {
+            ringbuf_entry!(Trace::SpResetFired);
+            if !invalidate && !self.do_handle_sp_reset() {
+                ringbuf_entry!(Trace::InvalidateSpMeasurement);
+                invalidate = true;
+            }
+            // else something something {
```
mkeeter, line 701: Remains of an else state here? (see also `//}` at L708)

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+        // In any case, keep/force the SP into a reset condition.
+        // Though AIRCR::SYSRESETREQ can be used to effect a local reset.
+        // that does not affect the whole SP SoC.
+        // So, use the SP_RESET GPIO.
+        //
+
+        // Setting up to inject the measurement program into the SP
+        // has several potential failures. Use this `prep` closure
+        // and `need_undo` state to keep from indenting too much.
+        let mut prep = |undo: &mut Undo| -> Result<(), ()> {
```
mkeeter, line 1735: Nit: you can capture `need_undo` automatically and don't need to pass it as an argument (so `*undo` in the function body would just become `need_undo`)

drv/lpc55-swd/src/main.rs
labbott: I like this idea. I found it difficult to reason about the switch statement.

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+            self.sp_reset_enter();
+            *undo |= Undo::RESET;
+
+            // Asserting SP_RESET for >1ms here works.
+            hl::sleep_for(1);
+
+            if self.dp_write_bitflags::<Dhcsr>(Dhcsr::resume()).is_err() {
```
mkeeter, line 1742: `resume()` is a confusing name for the function which sets debug enable

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+        };
+
+        // To ensures that any cleanup is done and the SP hardware is left
+        // running properly, there can only be one return at the end.
+
+        let digest = if prep(&mut need_undo).is_ok() {
+            self.do_measure_sp()
+        } else {
+            Err(())
+        };
```
mkeeter, line 1808:
```suggestion
        let digest = prep().and_then(|()| self.do_measure_sp());
```
(this assumes you take my suggestion above of capturing `need_undo` instead of passing it in)

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+                ringbuf_entry!(Trace::DemcrWriteError);
+            }
+            if let Ok(r) = self.dp_read_bitflags::<Dhcsr>() {
+                ringbuf_entry!(Trace::Dhcsr(r));
+            }
+        }
+
+        // This read of DHCSR should never be useful given the code above.
+        // Read back to make sure that VC_CORERESET is clear.
+        // XXX remove
```
mkeeter, line 1843: Is it time to remove this?

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+            error = true;
+        }
+
+        if !need_undo.is_empty() {
+            ringbuf_entry!(Trace::IncompleteUndo(need_undo));
+            error = true;
+        }
+
+        // The SP is still halted.
+        // Get it running again by toggling its RESET pin.
```
mkeeter, line 1880: It would also be nice to leave a comment here explaining that the caller is responsible for clearing the pending IRQ, so we don't measure it again.

drv/lpc55-swd/src/main.rs
lzrd: We don't need to have any Hubris sanity check here. It was reassuring at the start of this effort but it adds no value.

drv/lpc55-swd/src/main.rs
lzrd: `wait_for_sp_halt` would be better.

lib/endoscope/src/main.rs
lzrd: This is how we've been dealing with ranges of flash. In this case we're measuring flash content. We may encounter ECC errors which are handled properly. The addresses are from the STM32H753 data sheets and provided through the linker.

task/attest/src/main.rs
lzrd: There will be changes to the attestation API when it is finally plumbed through. This PR won't try to anticipate those changes.

drv/lpc55-swd/src/main.rs
lzrd: done

drv/lpc55-swd/src/main.rs
lzrd: The unwraps are gone but there are bounds checks and libraries that keep panics and their bloat in the binary.

drv/lpc55-swd/src/main.rs
lzrd: deleted

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+        // write start of image/vector table address(0x20000000) to VTOR (at 0xe000ed08)
+        // write the program to RAM
+        // write DHCSR to RUN (MATIC + C_DEBUGEN)
+        // poll for S_HALT or timeout
+
+        // Set SP's Program Counter
+        let sp_reset_vector =
+            u32::from_le_bytes(ENDOSCOPE_BYTES[4..=7].try_into().unwrap_lite());
+        if self
+            .do_write_core_register(Reg::Dr.into(), sp_reset_vector | 1)
```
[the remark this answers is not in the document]
lzrd: done.

drv/lpc55-swd/src/main.rs
lzrd: thanks. done

drv/lpc55-swd/src/main.rs
```diff
@@ -1339,26 +1351,48 @@ impl ServerImpl {
...
+
+        const_assert!(ENDOSCOPE_BYTES.len() > 2 * 1024);
+
         // Set SP's Program Counter
-        let sp_reset_vector =
-            u32::from_le_bytes(ENDOSCOPE_BYTES[4..=7].try_into().unwrap_lite());
-        if self
-            .do_write_core_register(Reg::Dr.into(), sp_reset_vector | 1)
-            .is_err()
+        if let Some(sp_reset_vector) = slice_to_le_u32(&ENDOSCOPE_BYTES[4..=7])
```
mkeeter, line 1370: Why did this change?  I found the previous version to be more clear:
```rust
let sp_reset_vector =
    u32::from_le_bytes(ENDOSCOPE_BYTES[4..=7].try_into().unwrap_lite());
```
versus the new code, which hides the `unreachable!` in the `else` of a conditional which is always taken

drv/lpc55-swd/src/main.rs
lzrd: gone

drv/lpc55-swd/src/main.rs
```diff
@@ -512,28 +598,130 @@ impl idl::InOrderSpCtrlImpl for ServerImpl {
...
+
+            const SLOT: PintSlot = ROT_TO_SP_RESET_L_IN_PINT_SLOT;
+            let _ = gpio.pint_op(SLOT, PintOp::Clear, PintCondition::Status);
+            // Get rid of spurious interrupts cause by do_handle_sp_reset()
+            // toggling SP_RESET.
+            sys_irq_control_clear_pending(
+                notifications::SP_RESET_IRQ_MASK,
+                true,
+            );
+            ringbuf_entry!(Trace::EndOfNotificationHandler);
```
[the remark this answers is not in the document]
lzrd: thanks. moved.

drv/lpc55-swd/src/main.rs
mkeeter: Doesn't the kernel queue the notification as soon as the pin changes state?  I would expect that we need a call to `sys_irq_control_clear_pending` here as well.

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+
+        ringbuf_entry!(Trace::SpResetAsserted);
+
+        // This notification handler should be compatible with watchdog but
+        // will result in the invalidation of any dumps held in the SP if successful.
+
+        // TODO: confirm that bank-flipping SP update watchdog is working.
+
+        self.init = true; // we don't want GPIO pin reconfiguration.
+        if self.do_setup_swd().is_ok() {
```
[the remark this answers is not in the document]
lzrd: I pushed changes that include a fix to this. The GPIO pins are statically configured except when handling an SP_RESET and all code paths return the SP_RESET pin  to its input/detection configuration. That includes the RoT being reset where `swd`'s `main()` will call pin_setup().

drv/lpc55-swd/src/main.rs
lzrd: Comments added to document the various cases.

drv/lpc55-swd/src/main.rs
lzrd: Changed

drv/lpc55-swd/src/main.rs
lzrd: Done

drv/lpc55-swd/src/main.rs
lzrd: You are correct. Moved the extra interrupt suppression into the notification handler which is the only place that generates them.

drv/lpc55-swd/src/main.rs
lzrd: Done

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1297,626 @@ impl ServerImpl {
...
+
+            if self
+                .dp_write_bitflags::<Demcr>(Demcr::VC_CORERESET)
+                .is_err()
+            {
+                ringbuf_entry!(Trace::DemcrWriteError);
+                *undo |= Undo::VC_CORERESET;
+                return Err(());
+            }
+            *undo |= Undo::VC_CORERESET;
```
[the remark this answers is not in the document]
lzrd: yes. In this failure case we don't know why we failed to set VC_CORERESET so we're going to try to put it back to normal, which will probably fail also. But since we're setting that flag unconditionally, it only needs to be done once.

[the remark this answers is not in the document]
lzrd: Fixed for DEBUGEN as well.

drv/lpc55-swd/src/main.rs
lzrd: yes

drv/lpc55-swd/src/main.rs
lzrd: LGTM

drv/lpc55-swd/src/main.rs
lzrd: I wanted to get rid of calls to unwrap_lite().

How do you feel about this:

```rust
        // Set SP's Stack Pointer
        if slice_to_le_u32(&ENDOSCOPE_BYTES[0..=3])
            .and_then(|initial_sp| {
                self.do_write_core_register(Reg::Sp, initial_sp).ok()
            })
            .is_some()
        {
            ringbuf_entry!(Trace::WroteSpRegisterFail);
            return Err(());
        }
```

drv/lpc55-swd/src/main.rs
lzrd: This check is on the leased buffer. The calling task can mess this up and this is the place to deal with it. This PR doesn't try to change these APIs.

drv/lpc55-swd/src/main.rs
lzrd: Sort of. What it is doing is clearing the C_HALT bit though and keeping the other required bits needed to maintain debug control.

lzrd: added comment to that effect.

drv/lpc55-swd/src/main.rs
lzrd: I refactored the pending interrupt management.

drv/lpc55-swd/src/main.rs
```diff
@@ -498,12 +600,28 @@ impl idl::InOrderSpCtrlImpl for ServerImpl {
...
         // already running; set the new deadline based on current time.
         set_timer_relative(time_ms, notifications::TIMER_MASK);
+
+        // The common case is that there will be a RESET before the watchdog
+        // can fire.
+        // That will kick off an SP image measurement which takes under a second.
+        //
+        // At the time of writing this comment, the watchdog timer value used
+        // in omicron is 2000ms. With different startup times for the various
+        // SP Hubris applications, it is important to test for watchdog failuew
```
flihp, line 622: `s/failuew/failure`

drv/lpc55-swd/src/main.rs
mkeeter: Vibes: I prefer the explicit `unwrap_lite`: we should never fail to convert a 4-byte slice into a `u32` and I think it's fine to be explicit about it.  It will be compiled down to nothing ([Godbolt example](https://rust.godbolt.org/z/E4feeq4qr)), so there's no runtime cost.

lzrd: Ok. I'm fine with that.

lib/endoscope/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+# running in ITCM/DTCM with tuned clocking takes about 0.5 seconds
+soc_stm32h753 = ["stm32h7/stm32h753", "drv-stm32h7-startup/h753"]
+
+#[build]
```
flihp, line 45: What are the benefits to keeping this section (and the ones above) that are commented out?

lib/endoscope/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+    // Safety: The bounds of the device's flash area are link-time constants.
+    let image = unsafe {
+        core::slice::from_raw_parts(
+            FLASH_BASE.as_ptr() as u32 as *const u8,
+            FLASH_SIZE.as_ptr() as u32 as usize,
+        )
+    };
+
+    hash.update(image[..].as_ref());
+    let digest: [u8; 256 / 8] = hash.finalize().into();
```
flihp, line 85: Getting the digest size from the `Sha3_256` type instead of hardcoding `256 / 8` will make switching hash functions much easier for the next person who comes along. I struggle to find the traits from the RustCrypto crates sometimes but something like:

```rust
use digest::{core_api::OutputSizeUser, typenum::Unsigned};
...
const DIGEST_SIZE: usize = <Sha3_256 as OutputSizeUser>::OutputSize::USIZE;
...
let digest: [u8; DIGEST_SIZE] = hash.finalize().into();
```

should work.

lib/endoscope/build.rs
```diff
@@ -0,0 +1,34 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use anyhow::Result;
```
flihp, line 5: NIT: Nothing in this file requires `anyhow::Result` so I don't think there's any reason to bring `anyhow` in as a dev dependency.

lib/endoscope/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+cortex-m = { workspace = true, default-features = false }
+cortex-m-rt = { workspace = true, default-features = false, features = ["device"] }
+panic-halt = { workspace = true, default-features = false }
+sha3 = { workspace = true, default-features = false }
+endoscope-abi = { path = "../endoscope-abi" }
+stm32h7 = { workspace = true }
+drv-stm32h7-startup = { path = "../../drv/stm32h7-startup" }
+
+[build-dependencies]
+anyhow = { workspace = true }
```
flihp, line 32: see https://github.com/oxidecomputer/hubris/pull/1946/files#r1987833120

lib/endoscope/src/shared.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+unsafe impl Sync for SharedWrapper {}
+
+impl SharedWrapper {
+    pub const fn new() -> Self {
+        SharedWrapper {
+            shared: UnsafeCell::new(Shared {
+                state: State::Preboot as u32,
+                digest: [0xff_u8; 32],
```
flihp, line 21: same as https://github.com/oxidecomputer/hubris/pull/1946/files#r1987774415: get the digest length from the digest type being used

lib/endoscope/Cargo.toml
lzrd: It's a reminder that if lto is ever enabled in bindeps, that it should be used in order to save some time and space.

lib/endoscope/Cargo.toml
lzrd: thanks

lib/endoscope/Cargo.toml
lzrd: Changing to an explicit reminder.

lib/endoscope/src/main.rs
lzrd: That adapts nicely to get rid of the hard-coded constants. Thanks.

lib/endoscope/src/shared.rs
lzrd: done

drv/lpc55-swd/build.rs
```diff
@@ -88,6 +103,161 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
     Ok(())
 }
 
+fn prepare_endoscope() -> Result<(), anyhow::Error> {
+    let key = "CARGO_BIN_FILE_ENDOSCOPE";
+    let elf_path = PathBuf::from(
+        std::env::var(key)
+            .with_context(|| format!("Cannot read env var '{}'", key))?,
+    );
```
mkeeter, line 111: To make sure that this is rebuilt if the environmental variables changes, you should add
```rust
    println!("cargo:rerun-if-env-changed={key}");
```

lzrd: Done. Thanks!

drv/lpc55-swd/src/main.rs
```diff
@@ -392,91 +525,43 @@ impl idl::InOrderSpCtrlImpl for ServerImpl {
...
-        if self.write_single_target_addr(DCRSR, r as u32).is_err() {
-            return Err(SpCtrlError::Fault.into());
-        }
-
+        Reg::from_u16(register)
+            .ok_or(SpCtrlError::InvalidCoreRegister)
+            .and_then(|r| {
+                self.write_single_target_addr(DCRSR, r as u32)
+                    .map_err(|_| SpCtrlError::Fault)
+            })?;
```
mkeeter, line 558: vibes: I would break this into two operations, instead of chaining with `and_then`
```suggestion
        let r = Reg::from_u16(register)
            .ok_or(SpCtrlError::InvalidCoreRegister)?;
        self.write_single_target_addr(DCRSR, r as u32)
            .map_err(|_| SpCtrlError::Fault)?;
```

drv/lpc55-swd/src/main.rs
```diff
@@ -512,28 +613,174 @@ impl idl::InOrderSpCtrlImpl for ServerImpl {
         _msg: &userlib::RecvMessage,
     ) -> Result<(), RequestError<core::convert::Infallible>> {
         ringbuf_entry!(Trace::DisabledWatchdog);
+        self.watchdog_ms = None;
```
mkeeter, line 616: I don't understand the point of `watchdog_ms`; as far as I can tell, it is written but never used?

drv/lpc55-swd/src/main.rs
```diff
@@ -747,7 +994,6 @@ impl ServerImpl {
     }
 
     fn swd_read(&mut self, port: Port, reg: RawSwdReg) -> Result<u32, Ack> {
-        ringbuf_entry!(Trace::SwdRead(port, reg));
```
mkeeter, line 750: Is removing this logging intentional (also below in `swd_write`)?

drv/lpc55-swd/src/main.rs
lzrd: You are correct. It isn't needed. It is left over from me tracking the interaction of dump, watchdog, and measurement activities.

drv/lpc55-swd/src/main.rs
lzrd: Yes. In the case of code injection, this is a very noisy point to log.

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1358,588 @@ impl ServerImpl {
...
+        self.sp_reset_leave();
+    }
+
+    fn sp_reset_enter(&mut self) {
+        setup_rot_to_sp_reset_l_out(self.gpio);
+        let gpio = Pins::from(self.gpio);
+        gpio.set_val(ROT_TO_SP_RESET_L_OUT, Value::Zero);
+    }
+
+    /// Cleanup is true if we don't want to be notified of this reset
```
mkeeter, line 1377: Outdated comment, there's no `cleanup` argument any more

lzrd: done

drv/lpc55-swd/src/main.rs
mkeeter: I agree that the calling task can mess this up.  I think we should check for that condition before calling `start_read_transaction`.  Right now, if the calling task messes it up, then we start the read transaction _before_ returning an error, which leaves the SP in a weird state.

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1358,588 @@ impl ServerImpl {
...
+        // Not ok yet: A reset happened. If we don't get a measurement then
+        // make sure that the old one is invalidated.
+        let mut error = false;
+
+        ringbuf_entry!(Trace::SpResetAsserted);
+
+        // This notification handler should be compatible with watchdog but
+        // will result in the invalidation of any dumps held in the SP if successful.
+
+        // TODO: confirm that bank-flipping SP update watchdog is working.
```
mkeeter, line 1731: Can we remove this TODO?

drv/lpc55-swd/src/main.rs
```diff
@@ -1058,21 +1358,588 @@ impl ServerImpl {
...
+    }
+}
+
+fn slice_to_le_u32(slice: &[u8]) -> Option<u32> {
+    if slice.len() == core::mem::size_of::<u32>() {
+        if let Ok(data) = core::convert::TryInto::<[u8; 4]>::try_into(slice) {
+            return Some(u32::from_le_bytes(data));
+        }
+    }
+    None
```
mkeeter, line 1934:
```suggestion
   slice.try_into().map(u32::from_le_bytes).ok()
```

task/attest/src/main.rs
```diff
@@ -252,6 +254,61 @@ impl idl::InOrderAttestImpl for AttestServer {
             return Err(AttestError::LogFull.into());
         }
 
+        // The first measurement can only be made by a privileged task.
+        if self.measurements.is_empty()
+            && !PERMIT_LOG_RESET.iter().any(|x| *x == msg.sender.0)
+        {
+            return Err(ClientError::AccessViolation.fail());
```
mkeeter, line 261: This will fault the client; I'm not sure if that's the behavior that we want here.

Specifically, if the SP tries to record a measurement into an empty log, the RoT's side of the `sprot-server` will be the one killed when it tries to call `Attest.record`.

drv/lpc55-swd/src/main.rs
lzrd: done

drv/lpc55-swd/src/main.rs
lzrd: I'll re-run my fault injection test to make sure all is well then remove the comment.

task/hiffy/src/lpc55.rs
```diff
@@ -75,6 +75,8 @@ pub enum Functions {
     ReadFromSp((u32, u32), drv_sp_ctrl_api::SpCtrlError),
     #[cfg(feature = "spctrl")]
     SpCtrlInit((), drv_sp_ctrl_api::SpCtrlError),
+    #[cfg(feature = "spctrl")]
+    MeasureSp((), drv_sp_ctrl_api::SpCtrlError),
```
mkeeter, line 79: It's confusing that the function is `db_reset_sp` but this is `MeasureSp`, maybe rename both sides to `ResetAndMeasureSp`?

lzrd: The measurement is a side-effect and won't happen if there is a powered STLINK attached to the SP. I can put comments in to that effect.

mkeeter (approved):

task/attest/src/main.rs
lzrd: There are a few of these that indicate coding errors in the client (they should not ask) and some that should not have `.fail()`. I've reviewed them all and will fix the unnecessary/unwanted penalties. Thanks!

drv/lpc55-swd/src/main.rs
lzrd: Thanks!

task/attest/src/main.rs
lzrd: Added a non-fatal error and used it where appropriate.

drv/lpc55-swd/src/main.rs
lzrd: done

labbott (approved):

Issue #1943, opened 2024-12-10
Kernel refuses to start if task's initial stack frame spans regions


cbiffle:
The `exhubris` tools are turning out to be a pretty good kernel torture test. 😄

During the `reinitialize` operation that starts (and restarts) a task, the kernel (on ARM M-profile at least) needs to deposit an initial stack frame in the task's stack memory, from which the hardware will load the task's initial register values. Before we do that, we "zap" the stack, overwriting it with a predictable pattern to help monitor a task's stack usage high-watermark.

The code in question is here: https://github.com/oxidecomputer/hubris/blob/8437fbd30aa129bb6e6fb66fe274be719c62ee8b/sys/kern/src/arch/arm_m.rs#L298-L322

That code assumes that the memory region containing the topmost word in the stack (`initial_stack_ptr.saturating_sub(4)`) also contains the the entire stack, because the kernel uses `region.base` to define the place to start zapping.

If the initial stack frame (104 bytes on Cortex-M4, thanks largely to the FPU) does not _fit_ in the region that contains the topmost word of the stack, you instead get an invalid `USlice` constructed using `region.base` and a large negative number. (If you're building with overflow checks enabled, you get a subtraction with overflow panic. I disabled overflow checks to help find the real problem.) `USlice` rightfully rejects this as nonsense, which the code here _really wasn't expecting,_ since it calls `unwrap_lite` on the result. Panic.

This code dates to the early days of Hubris when a task would always have a single contiguous RAM region containing the stack. In mainline "Oxide Hubris" that's still true in practice: we mostly apply the region-splitting allocator to flash.

The exhubris tools split both RAM and flash, deciding which to split next based on the potential cost of fragmentation given their sizes. This means I wound up with a task with a 256 byte stack split over no fewer than _five regions_ (sizes: 32, 64, 64, 64, 32). In that case, it takes the last _three_ regions to cover the 104-byte initial stack frame, and hilarity ensues. Where by "hilarity" I mean "the application refusing to start."

Interestingly, the actual code for writing the frame (just below) works fine if I disable the stack zap code, because it doesn't try to be clever with regions. So, the zap code itself needs to be improved to handle more elaborate RAM region splits.

I may land an initial fix that just makes the zap code optional, since the stack zapping method is inexact anyway (#1872).

mkeeter:
https://github.com/oxidecomputer/hubris/pull/2458 has made this an issue for stock Hubris as well, which bumps the priority of a fix.

[...]

PR #1947, opened 2024-12-13
kern: Make stack zap failure less catastrophic

cbiffle:
The kernel really wants to scribble task memory. Like, really really. Like, contains `unwrap()`s about it.

This is silly. It's a diagnostic tool. The kernel doesn't care if it works or not, and the application should not fail to start because of it.

So, this commit makes it optional and tolerates failure.

Is this me papering over a kernel bug? Yes, #1943 to be precise. But it's reasonable paper.

### DIFF
--- a/sys/kern/src/arch/arm_m.rs
+++ b/sys/kern/src/arch/arm_m.rs
@@ -309,15 +309,21 @@ pub fn reinitialize(task: &mut task::Task) {
         .iter()
         .find(|region| region.contains(initial_stack.saturating_sub(4)))
     {
-        let mut uslice: USlice<u32> = USlice::from_raw(
+        // If the slice doesn't fit in the region, this will fail. Should this
+        // occur, don't crash the entire system, since this is a diagnostic tool
+        // -- just skip filling the stack.
+        if let Ok(mut uslice) = USlice::<u32>::from_raw(
             region.base as usize,
             (initial_stack - frame_size - region.base as usize) >> 2,
-        )
-        .unwrap_lite();
-
-        let zap = task.try_write(&mut uslice).unwrap_lite();
-        for word in zap.iter_mut() {
-            *word = 0xbaddcafe;
+        ) {
+            // This one, we're unwrapping rather than tolerating failure. This
+            // is because try_write failing would indicate an invalid region
+            // descriptor for the task (read-only stack area) which would bite
+            // us later.
+            let zap = task.try_write(&mut uslice).unwrap_lite();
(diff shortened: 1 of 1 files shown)


sys/kern/src/arch/arm_m.rs
```diff
@@ -309,15 +309,14 @@ pub fn reinitialize(task: &mut task::Task) {
         .iter()
         .find(|region| region.contains(initial_stack.saturating_sub(4)))
     {
-        let mut uslice: USlice<u32> = USlice::from_raw(
+        if let Ok(mut uslice) = USlice::<u32>::from_raw(
```
hawkw, line 312: mmmaybe worth a comment here to the effect of "if this fails, don't crash the whole system, as it's just for diagnostic purposes"? also, perhaps we should mention why it might fail here?

sys/kern/src/arch/arm_m.rs
```diff
@@ -309,15 +309,14 @@ pub fn reinitialize(task: &mut task::Task) {
...
             region.base as usize,
             (initial_stack - frame_size - region.base as usize) >> 2,
-        )
-        .unwrap_lite();
-
-        let zap = task.try_write(&mut uslice).unwrap_lite();
-        for word in zap.iter_mut() {
-            *word = 0xbaddcafe;
+        ) {
+            let zap = task.try_write(&mut uslice).unwrap_lite();
```
hawkw, line 316: AFAICT, the `try_write` call failing here would be Bad News --- if a task's stack region isn't actual memory or that memory isn't writable, something has gone seriously wrong, so I think unwrapping this is reasonable. but, that might not be immediately obvious to a reader, who would see that we allow the `Uslice::from_raw` call to fail, but if we can't write to it, we throw up our hands and kill the whole system, which might look weird at a glance. Maybe a comment describing the difference between the two failure modes would be worth including here?

hawkw (approved):
I agree that a failure to scribble over the stack should probably not take down the whole system, so this looks good to me! It might be nice if there were some comments explaining why we allow this to fail in the source, for future readers? Not a huge deal though.

sys/kern/src/arch/arm_m.rs
cbiffle: Yeah, that was approximately my thinking with leaving the unwrap, but you're right that a comment would improve things.

cbiffle:
Comments added, PTAL.

hawkw (approved):
comments look good to me, ship it!

PR #1948, opened 2024-12-13
kern: let drivers disable double-interrupt behavior

### LINKED ISSUE #536: IRQ double-fires, but only when handled by Hubris
Hey all! I'm seeing some odd behavior when servicing interrupts in Hubris: I'm seeing interrupts triggering twice, but _only_ when handled through hubris's `DefaultHandler` mechanism. I know this smacks of the usual _"You just returned too fast and forgot to allow the IRQ acknowledge to go through the bus"_, but hear me out 😅 

## Custom ISR: single-firing

For a baseline test, I override the `TIM_CC` interrupt (technically `TIM1_CC`, but the SVD/PAC seems to have a typo) and trigger at the very center of the PWM signal. When in the interrupt, I set a GPIO pin, `nop` for a bit, acknowledge the pending interrupt in the timer, ensure it's propagated, then drop the GPIO pin.

```rust
#[interrupt]
fn TIM_CC() {
    let gpiof = unsafe { &*device::GPIOF::ptr() };
    gpiof.bsrr.write(|w| w.bs11().set_bit());
    for _ in 0..100 {
        cortex_m::asm::nop();
    }
    let timer = unsafe { &*device::TIM1::ptr() };
    timer.sr.write(|w| w.cc4if().clear());
    cortex_m::asm::dmb(); // Overkill; triple-checking to be sure that the IRQ
    cortex_m::asm::dsb(); // acknowledge went through. Reading back the 
    cortex_m::asm::isb(); // register below should be sufficient.
    while timer.sr.read().cc4if().bit_is_set() {}
    gpiof.bsrr.write(|w| w.br11().set_bit());
}
```
image

Cyan: PWM signal
Blue: GPIO toggling inside ISR

1) Interrupt trigger location
2) GPIO high during ISR

Seems to work as expected.

## Hubris `notification`: double-firing

Now if I take the ***exact same code*** and put it in an Idolatry `handle_notificaiton`, I'm seeing the interrupt fire a second time between the end of the `syscall_entry` and when the `sys_irq_control` call returns.

```rust
const TIM1_CC_MASK: u32 = 1 << 2;

        TIM1_CC_MASK
    }

        let gpiof = unsafe { &*device::GPIOF::ptr() };
        gpiof.bsrr.write(|w| w.bs11().set_bit());
        for _ in 0..100 {
            cortex_m::asm::nop();
        }
        let timer = unsafe { &*device::TIM1::ptr() };
        timer.sr.write(|w| w.cc4if().clear());
        cortex_m::asm::dmb();
        cortex_m::asm::dsb();
        cortex_m::asm::isb();
        while timer.sr.read().cc4if().bit_is_set() {}
        sys_irq_control(TIM1_CC_MASK, true);  // <--- only new line
        gpiof.bsrr.write(|w| w.br11().set_bit());
    }
}
```

image

Cyan: PWM signal
Blue: GPIO toggling inside ISR
Magenta: `event_isr_enter` from the new kernel profiling code from last week (enter/exit on `DefaultHandler`)
Yellow: `event_syscall_enter` from the new kernel profiling code (enter/exit on `syscall_entry`)

A bit busy of an image, sorry. Here's the sequence of events:

0) Timer IRQ is triggered
1) It's handled by Hubris' `DefaultHandler` (magenta)
2) The Hubris notification has reached `handle_notification`, and a GPIO pin is toggled high (blue)
3) A call to `sys_irq_control(TIM1_CC_MASK, true)` to handle the `irq_control` sycall (yellow)
4) A **second** timer IRQ is fired. Note that this happens _prior to the control path returning from the `sys_irq_control` call_
5) The `sys_irq_control` call returns, the GPIO pin is lowered, and the original Hubris notification call to `handle_notification` is complete.
6) A second Hubris notification is triggered, driving the GPIO pin high again in `handle_notification`
7) The call to `sys_irq_control` to re-enable the notification handler. Note that this does _not_ cause a third interrupt to fire.
8) Post-notification syscall (not sure what, but I assume it's the `sys_recv` being called again within the Idolatry server's `handle_n`.

## Thoughts

* I'm running this in an Idolatry server and haven't gotten a chance to test it out via raw syscalls, but I can't see anything special within `dispatch_n` that would make me think the Idolatry framework would be the cause. I can also confirm that the second interrupt's `IPSR` is the same IRQ number as the first (26). 
* I also tried poking the corresponding `NVIC[ICPR]` to manually acknowledge the IRQ at the NVIC level but that caused a memory fault since `handle_notification` is in a task and thus unprivileged code (A Good Thing:tm:).
* I double-checked that I acknowledged the timer's ISR bit by attempting to panic if it was still set just prior to re-enabling the notification:

  ```rust
        ...
        while timer.sr.read().cc4if().bit_is_set() {}
        cortex_m::asm::dmb();
        cortex_m::asm::dsb();
        cortex_m::asm::isb();
        if timer.sr.read().cc4if().bit_is_set() {
            panic!("It still was set :C");
        }
        sys_irq_control(TIM1_CC_MASK, true);
  ```

  No panic :/
* Both the syscall _and_ the second interrupt occur before `sys_irq_control` has returned to the original task's code, so not sure what else I could/should be doing to prevent it.
* AFAICT the `#[interrupt]` proc macro from `cortex_m_rt` doesn't do anything fancy aside from defining an `$IRQ_trampoline`
* Not sure what the difference is between the first and second, but when the second iteration of `handle_notification` calls `sys_irq_control` the interrupt is nowhere to be found after the syscall.

There's still a good chance this is a case of PEBKAC, but considering that the same code block in a bare IRQ handler does not fire twice I'm at a loss as to why the Hubris notification would be coming through twice. Any ideas what may be happening here?

cbiffle:
Currently drivers tend to get a spurious interrupt notification just after handling a real one. This is because the interrupt controller notices the kernel generic ISR exit while the interrupt condition is still asserted, because it doesn't expect us to be handling it in task code. So, it sets the pending bit so the 'new' interrupt condition isn't forgotten.

This causes an interrupt to occur immediately when we re-enable.

This adds an additional bit for IRQ_CONTROL that drivers can use to request any pending interrupt status be cleared.

Fixes #536.

### DIFF
--- a/doc/syscalls.adoc
+++ b/doc/syscalls.adoc
@@ -469,7 +469,9 @@ Collects information about one entry in a sender's lease table.
 ==== Arguments
 
 - 0: notification bitmask corresponding to the interrupt
-- 1: desired state (0 = disabled, 1 = enabled)
+- 1: desired state
+** bit 0: 0 = disabled, 1 = enabled
+** bit 1: 0 = leave pending, 1 = clear pending
 
 ==== Return values
 
@@ -497,6 +499,13 @@ their notification bits. However, this is quite deliberate, for two reasons:
 
 2. It makes it impossible for a task to mess with other tasks' interrupts,
    since it can only refer to its _own_ mapped interrupts, by construction.
+
+The concept of a "pending" interrupt is inherently specific to a particular
+architecture and interrupt controller. On an architecture without a concept of
+pending interrupts, bit 1 has no effect. However, on architectures with
+level-triggered interrupts from peripherals and a concept of "pending"
+interrupts, clearing the pending status when re-enabling may be important for
+avoiding a duplicate notification.
 
 === `PANIC` (8)
 
--- a/sys/abi/src/lib.rs
+++ b/sys/abi/src/lib.rs
@@ -549,3 +549,15 @@ bitflags::bitflags! {
         const POSTED = 1 << 2;
     }
 }
+
+bitflags::bitflags! {
+    /// Bitflags that can be passed into the `IRQ_CONTROL` syscall.
+    #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+    pub struct IrqControlArg: u32 {
+        /// Enables the interrupt if present, disables if not present.
+        const ENABLED = 1 << 0;
+        /// If present, requests that any pending instance of this interrupt be
+        // cleared.
+        const CLEAR_PENDING = 1 << 1;
+    }
+}
--- a/sys/kern/src/syscalls.rs
+++ b/sys/kern/src/syscalls.rs
@@ -31,8 +31,8 @@
 use core::sync::atomic::{AtomicBool, Ordering};
 
 use abi::{
-    FaultInfo, IrqStatus, LeaseAttributes, SchedState, Sysnum, TaskId,
-    TaskState, ULease, UsageError,
+    FaultInfo, IrqControlArg, IrqStatus, LeaseAttributes, SchedState, Sysnum,
+    TaskId, TaskState, ULease, UsageError,
 };
 use unwrap_lite::UnwrapLite;
 
@@ -753,15 +753,19 @@ fn irq_control(
 ) -> Result<NextTask, UserError> {
     let args = tasks[caller].save().as_irq_args();
 
-    let operation = match args.control {
-        0 => crate::arch::disable_irq,
-        1 => crate::arch::enable_irq,
-        _ => {
-            return Err(UserError::Unrecoverable(FaultInfo::SyscallUsage(
-                UsageError::NoIrq,
-            )))
-        }
+    // TODO: our use of NoIrq here is conventional but is getting increasingly
+    // weird. Arguably it's always been wrong, since it's validating the control
+    // argument.
+    let control = IrqControlArg::from_bits(args.control).ok_or(
+        UserError::Unrecoverable(FaultInfo::SyscallUsage(UsageError::NoIrq)),
+    )?;
+
+    let operation = if control.contains(IrqControlArg::ENABLED) {
+        crate::arch::enable_irq
+    } else {
+        crate::arch::disable_irq
     };
+    let also_clear_pending = control.contains(IrqControlArg::CLEAR_PENDING);
 
     let caller = caller as u32;
 
@@ -774,7 +778,7 @@ fn irq_control(
             UsageError::NoIrq,
         )))?;
     for i in irqs.iter() {
-        operation(i.0);
+        operation(i.0, also_clear_pending);
     }
     Ok(NextTask::Same)
 }
--- a/sys/userlib/src/lib.rs
+++ b/sys/userlib/src/lib.rs
@@ -964,8 +964,30 @@ unsafe extern "C" fn sys_borrow_info_stub(
 
 #[inline(always)]
 pub fn sys_irq_control(mask: u32, enable: bool) {
+    let mut arg = IrqControlArg::empty();
+    if enable {
+        arg |= IrqControlArg::ENABLED;
+    }
+
     unsafe {
-        sys_irq_control_stub(mask, enable as u32);
+        sys_irq_control_stub(mask, arg.bits());
+    }
+}
+
+/// Variation on [`sys_irq_control`] that also clears any pending interrupt.
+///
+/// This sets the interrupt enable status based on `enable`, and also cancels a
+/// pending instance of this interrupt in the interrupt controller, if the
+/// interrupt controller supports such a concept (ARM M-profile NVIC does, for
+/// instance).
+#[inline(always)]
+pub fn sys_irq_control_clear_pending(mask: u32, enable: bool) {
+    let mut arg = IrqControlArg::CLEAR_PENDING;
+    if enable {
+        arg |= IrqControlArg::ENABLED;
+    }
+    unsafe {
+        sys_irq_control_stub(mask, arg.bits());
     }
 }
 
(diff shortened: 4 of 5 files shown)


mkeeter (approved):

doc/syscalls.adoc
```diff
@@ -498,6 +500,13 @@ their notification bits. However, this is quite deliberate, for two reasons:
 2. It makes it impossible for a task to mess with other tasks' interrupts,
    since it can only refer to its _own_ mapped interrupts, by construction.
 
+The concept of a "pending" interrupt is inherently specific to a particular
+architecture and interrupt controller. On an architecture without a concept of
```
hawkw, line 504: this might just be one of my pet hobby-horses but i might cite some examples of such architectures here?

sys/userlib/src/lib.rs
```diff
@@ -969,6 +969,13 @@ pub fn sys_irq_control(mask: u32, enable: bool) {
     }
 }
 
+#[inline(always)]
```
hawkw, line 972: Should this have a doc comment (perhaps reproducing what's in `syscalls.adoc`)?

sys/kern/src/syscalls.rs
```diff
@@ -753,15 +753,16 @@ fn irq_control(
 ) -> Result<NextTask, UserError> {
     let args = tasks[caller].save().as_irq_args();
 
-    let operation = match args.control {
+    let operation = match args.control & 1 {
```
hawkw, line 756: feel free to ignore me, but i would personally prefer to write these masks as 
```suggestion
    let operation = match args.control & 0b01 {
```
as it feels a bit more obvious which bit we are picking out

sys/kern/src/syscalls.rs
```diff
@@ -753,15 +753,16 @@ fn irq_control(
...
-        1 => crate::arch::enable_irq,
-        _ => {
-            return Err(UserError::Unrecoverable(FaultInfo::SyscallUsage(
-                UsageError::NoIrq,
-            )))
-        }
+        _ => crate::arch::enable_irq,
     };
+    let also_clear_pending = args.control & 2 != 0;
+    if args.control & !3 != 0 {
```
hawkw, line 761: similarly, this is just a personal style preference and you're welcome to tell me i'm wrong, but i'd have done
```suggestion
    let also_clear_pending = args.control & 0b10 != 0;
    if args.control & !0b11 != 0 {
```

sys/userlib/src/lib.rs
```diff
@@ -969,6 +969,13 @@ pub fn sys_irq_control(mask: u32, enable: bool) {
...
+pub fn sys_irq_control_clear_pending(mask: u32, enable: bool) {
+    unsafe {
+        sys_irq_control_stub(mask, enable as u32 | 0b10);
```
hawkw, line 975: it seems like it could be nicer if the bitflags for this argument were defined in `sys/abi` so that the kernel and userlib could depend on the same definition, instead of having magic numbers. this isn't a huge deal, and if it's too annoying, we're probably not going to go and mess with them any time soon, so it's fine-as is. but, it would make me a little happier.

hawkw (approved):
all of my comments are just obnoxious style nitpicks, so feel free to merge whenever you're happy with it!

doc/syscalls.adoc
cbiffle: I can't, sorry. I'd take suggestions, but, all the architectures where I'm deeply familiar with the interrupt controller have this bit. I think maybe AVRs don't, but they also don't have a Rust port.

cbiffle:
> all of my comments are just obnoxious style nitpicks, so feel free to merge whenever you're happy with it!

You are too hard on yourself, these are good comments. :-)

doc/syscalls.adoc
hawkw: hm, honestly, maybe we should just say "such as all the architectures Hubris currently targets"...but we would have to go and change that if hubris gets ported to some weird thing someday...

sys/userlib/src/lib.rs
cbiffle: Probably, though note that none of the stubs in the Oxide userlib have comments. I'll add one to this one.

sys/kern/src/syscalls.rs
cbiffle: Added a bitflags type and removed the constants.

cbiffle:
Alright, addressed, PTAL.

sys/abi/src/lib.rs
```diff
@@ -549,3 +549,15 @@ bitflags::bitflags! {
...
+bitflags::bitflags! {
+    /// Bitflags that can be passed into the `IRQ_CONTROL` syscall.
+    #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+    pub struct IrqControlArg: u32 {
+        /// Enables the interrupt if present, disables if not present.
+        const ENABLED = 1 << 0;
+        /// If present, requests that any pending instance of this interrupt be
+        // cleared.
+        const CLEAR_PENDING = 1 << 1;
+    }
```
hawkw, line 562: you love to see it!

sys/kern/src/syscalls.rs
```diff
@@ -753,16 +753,19 @@ fn irq_control(
...
+    } else {
+        crate::arch::disable_irq
     };
-    let also_clear_pending = args.control & 2 != 0;
-    if args.control & !3 != 0 {
-        return Err(UserError::Unrecoverable(FaultInfo::SyscallUsage(
-            UsageError::NoIrq,
-        )));
-    }
+    let also_clear_pending = control.contains(IrqControlArg::CLEAR_PENDING);
```
hawkw, line 768: this looks lovely!

hawkw (approved):

PR #1949, opened 2024-12-13
Syscall / context switch performance improvements

cbiffle:
There's a lot of stuff in here, and I may break it up before sending it for review. Feedback welcome regardless!

These changes together reduce the time spent in syscalls in my test application by about 26%. Some calls are boosted significantly more than that (30% off large sends, 40% off empty sends).

I recommend reviewing the commits rather than the combined diff.


sys/kern/src/arch/arm_m.rs
```diff
@@ -1474,16 +1473,15 @@ unsafe extern "C" fn handle_fault(task: *mut task::Task) {
     // switch to a task to run.
     with_task_table(|tasks| {
         let next = match task::force_fault(tasks, idx, fault) {
-            task::NextTask::Specific(i) => i,
+            task::NextTask::Specific(i) => &tasks[i],
             task::NextTask::Other => task::select(idx, tasks),
-            task::NextTask::Same => idx,
+            task::NextTask::Same => &tasks[idx],
```
hawkw, line 1555: It occurs to me that, because the length of the task table is always `HUBRIS_TASK_COUNT` and task indices are always in bounds, we could probably shave off a few instructions (and an unreachable panic site) by using `get_unchecked` when indexing the task table here and elsewhere in the scheduler. I'm not sure whether this optimization is worth the unsafe code, but it's worth thinking about...

sys/kern/src/arch/arm_m.rs
```diff
@@ -1474,16 +1473,15 @@ unsafe extern "C" fn handle_fault(task: *mut task::Task) {
...
         };
 
-        if next == idx {
+        if next as *const _ == &tasks[idx] as *const _ {
```
hawkw, line 1558: isn't it better to do this comparison by index rather than by address, because writing `&tasks[idx]` will do a bounds check when accessing the task table, which could be avoided if we just compared the integers? OTTOH then we are using an additional register or stack slot to store both the index and the address... :woman_shrugging:

sys/kern/src/arch/arm_m.rs
```diff
@@ -1688,16 +1686,15 @@ unsafe extern "C" fn handle_fault(
...
```
hawkw, line 1771: same comments as above

sys/kern/src/task.rs
```diff
@@ -820,47 +820,55 @@ pub fn check_task_id_against_table(
...
-        if !pred(&tasks[i]) {
+) -> Option<(usize, &Task)> {
+    let mut pos = previous;
+    let mut choice: Option<(usize, &Task)> = None;
+    for _step_no in 0..tasks.len() {
+        pos = pos.wrapping_add(1);
+        if pos >= tasks.len() {
+            pos = 0;
+        }
+        let t = &tasks[pos];
```
hawkw, line 857: similarly, there's probably a few instructions and a panic site that could be shaved off here by using `get_unchecked`, given that the `if` immediately above this line ensures that `pos` is always in bounds. On the other hand, LLVM might be smart enough to figure that out here --- the code that ensures it's in bounds is straightforward enough that I wouldn't be shocked if the compiler could figure it out...

hawkw (commented):
quick review of 2a56849d74e46bddd4d78e1dfe8d6332369a659b

aapoalas (commented):
:peanuts: checking in; really cool and interesting stuff! I left a few comments, hopefully of some potential use to you <3

sys/kerncore/src/lib.rs
```diff
@@ -159,35 +176,51 @@ where
 
     // Per the function's preconditions, the region table is sorted in ascending
     // order of base address, and the regions within it do not overlap. This
-    // lets us use a one-pass algorithm.
+    // lets us use a binary search followed by a short scan
```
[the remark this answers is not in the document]
cbiffle: Indeed!

sys/kerncore/src/lib.rs
```diff
@@ -82,6 +84,17 @@ pub trait MemoryRegion {
     fn contains(&self, addr: usize) -> bool;
     fn base_addr(&self) -> usize;
     fn end_addr(&self) -> usize;
+
+    #[inline(always)]
+    fn compare(&self, addr: usize) -> Ordering {
```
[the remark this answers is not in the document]
cbiffle: Done in upcoming commit, thanks!

sys/kern/src/arch/arm_m.rs
cbiffle: > we could probably shave off a few instructions (and an unreachable panic site) by using get_unchecked

That's probably true but we'd want to be _very_ careful about where we do it.

> why don't all these methods work with &(mut) [Task; HUBRIS_TASK_COUNT] references?

There's a historical reason why we don't use fixed-length arrays for the task table, but it's not really valid anymore. (The number of tasks was not originally available to the kernel at compile time.)

In practice, however, all of these functions get specialized by the compiler to fixed-length arrays. When disassembling the kernel you see a lot of `cmp r0, #6` and stuff for validation. Making the types explicit might be an interesting simplification -- relying on the compiler for stuff like this can be tricky.

I'd prefer not to pile either of those changes on _this_ PR if possible.

sys/kern/src/arch/arm_m.rs
cbiffle: As of this commit, we no longer _have_ the index, and `slice::elem_offset` isn't stable. Part of the performance improvement from this change is returning `&mut Task` from `select` instead of an index.

However, I note that we actually derive `idx` from the `task: *mut Task` value passed in from the ISR wrapper. So I will change the code to compare with that.

This code does not appear to generate a novel bounds check in practice, because `&tasks[idx]` winds up being done a lot of other places in this function implicitly.

I'll do a before and after comparison here in a bit, once I'm done addressing the rest of this feedback.

sys/kern/src/task.rs
cbiffle: Yeah, this (and basically all other cases where an access is dominated by a naive `i < slice.len()`) is optimized out. I'm trying to only apply unsafe in cases where I can't figure out how to get the panics removed from the safe code, and this code specifically does not produce a panic rn.

sys/kern/src/arch/arm_m.rs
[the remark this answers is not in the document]
cbiffle: I'd be happy to evaluate a change here. It'd come down to whether the added complexity (new types and invariants) produced a useful change in the generated code in practice.

sys/kern/src/task.rs
```diff
@@ -820,47 +820,55 @@ pub fn check_task_id_against_table(
...
     previous: usize,
     tasks: &[Task],
     pred: impl Fn(&Task) -> bool,
-) -> Option<usize> {
-    uassert!(previous < tasks.len());
-    let search_order = (previous + 1..tasks.len()).chain(0..previous + 1);
-    let mut choice = None;
-    for i in search_order {
-        if !pred(&tasks[i]) {
+) -> Option<(usize, &Task)> {
```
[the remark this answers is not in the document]
cbiffle: From a scan of the use sites, it looks like most want a scheduling hint (`NextTask`) as their return value. I hesitate to add a lifetime parameter to `NextTask`, as it's quite commonly used and would extend a lot of borrows in ways that might prove awkward.

The intent here would be to eliminate possible bounds checks in the consumption of the `NextTask` value it currently returns? I'd want to double-check that there are bounds checks being generated today before adding the complexity, but it might be an improvement.

sys/kern/src/arch/arm_m.rs
```diff
@@ -349,6 +350,99 @@ pub fn reinitialize(task: &mut task::Task) {
...
+    // If the size is zero or one, this subtraction will underflow. This
+    // should not occur in a valid image, but could occur due to runtime
+    // flash corruption. Any region size under 32 bytes is illegal on
+    // ARMv7-M anyway, so panicking is better than triggering possibly
+    // undefined hardware behavior.
+    //
+    // On ARMv6-M, there is no CLZ instruction either. This winds up
+    // generating decent intrinsic code for `leading_zeros` so we'll live
+    // with it.
+    let l2size = 30 - size.leading_zeros();
```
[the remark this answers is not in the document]
cbiffle: That comment might have been unclear, I'm using `leading_zeros` there _despite_ it not having native support on v6M, because the code generation is pretty okay. You are correct, of course, that there should no longer _be any_ code generation.

`leading_zeros` and `trailing_zeros` here are not interchangeable; `trailing_zeros` would measure the smallest power-of-two alignment of the size, not the magnitude of the size itself.

sys/kern/src/arch/arm_m.rs
```diff
@@ -350,11 +350,17 @@ pub fn reinitialize(task: &mut task::Task) {
     task.save_mut().exc_return = EXC_RETURN_CONST;
 }
 
+/// PMSAv6/7-style precomputed region data.
+///
+/// This struct is `repr(C)` to preserve the order of its fields, which happens
```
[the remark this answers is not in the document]
cbiffle: Data cache concerns are valid here, actually: while the M4 has no cache, M4 _implementations_ often have caches to accelerate data loads from flash, and the region descriptors are in flash. On the STM32L4, misses in this cache caused this change to improve performance by less than I expected when running at high enough speeds. (It's still a net win.)

Today, region information is separate from tasks to conserve flash in the case where a single region appears in multiple tasks. The most common case of this is the "null" region that we use to defend against null pointer dereferences in the kernel: every task has it mapped in at least one slot. Moving the region data into the tasks would expand each slot from one pointer to two words minimum (the size of a current `RegionDescExt` on ARM-M), at a cost of 32 bytes of flash per task.

I considered this, but I'm currently not convinced it's worthwhile -- the improvement of this change is significant, and the kernel is already bigger than I'd like in flash.

But, perhaps there's a data structure I haven't considered here. As long as it's relatively simple I'm open to suggestions!

sys/kern/src/arch/arm_m.rs
```diff
@@ -349,6 +350,104 @@ pub fn reinitialize(task: &mut task::Task) {
...
+#[cfg(any(armv6m, armv7m))]
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+#[repr(C)]
+pub struct RegionDescExt {
+    rbar: u32,
+    rasr: u32,
+}
+
+#[cfg(any(armv6m, armv7m))]
+pub const fn compute_region_extension_data(
```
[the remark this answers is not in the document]
cbiffle: Thanks! I've been wanting to do this for a very long time.

sys/kern/src/arch/arm_m.rs
cbiffle: FWIW, I have made this change and will fold it into the PR shortly, but it has no impact on the generated code on v7M or v6M.

sys/kerncore/src/lib.rs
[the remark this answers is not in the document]
cbiffle: As expected for moving something from a trait to a fn, this has no impact on the generated code. (I wanted to make sure!)

sys/kern/src/task.rs
hawkw: Ah, in that case, let's leave it as is! Good job, LLVM!

cbiffle:
Alright! I have...
- Addressed the comments that I didn't explicitly argue with (feel free to argue back, of course!)
- Fixed the rustfmt issues
- **Shortened the PR to just the two changes y'all have most thoroughly reviewed.** I'll put up a sequel PR with the rest.

PTAL!

sys/kern/src/arch/arm_m.rs
hawkw: Entirely reasonable on both counts, we can think more about this later.

sys/kerncore/src/lib.rs
```diff
@@ -159,35 +178,53 @@ where
...
+    // Perform fast checks on the initial region. In practical testing this
+    // provides a ~1% performance improvement over only using the loop below.
+    let first_region = &table[index];
+    if !region_ok(first_region) {
+        return false;
+    }
+    // Advance to the end of the first region
+    scan_addr = first_region.end_addr();
+    if scan_addr >= end_addr {
+        // That was easy
```
hawkw, line 201: a super tiny, unimportant copyedit nit that i feel quite silly for commenting on: all the other comments in this function are terminated by periods, but these aren't (exclamation point optional):
```suggestion
    // Advance to the end of the first region.
    scan_addr = first_region.end_addr();
    if scan_addr >= end_addr {
        // That was easy!
```

sys/kerncore/src/lib.rs
```diff
@@ -159,35 +178,53 @@ where
...
+    // Scan adjacent regions.
+    for region in &table[index + 1..] {
+        if !region.contains(scan_addr) {
+            // We've hit a hole without finishing our scan.
+            break;
+        }
+        // Make sure the region is permissible!
+        if !region_ok(region) {
+            // bail to the fail handling code at the end.
             break;
```
hawkw, line 214: hum, the "fail handling code at the end" is just `false`...so couldn't these just be `return false;`? I suppose it makes  sense to have them not be if we anticipate probably adding additional code that runs when no valid region is found, but...I dunno. Not a big deal, as the current thing is fine.

hawkw (approved):
The two changes that you've extracted out here look good to me! I'm looking forward to reviewing the others separately.

cbiffle:
Aaaand I apparently borked the rebase and the code doesn't build. Let me fix that.

sys/kerncore/src/lib.rs
cbiffle: ¡but what if I want it at the front though

sys/kern/src/arch/arm_m.rs
cbiffle: Update: so I actually measured that incorrectly! As shown by the resulting branch not actually building when I uploaded it. Whoops!

This is a net win, it knocks 16 bytes off the kernel by eliminating one redundant check.

sys/kern/src/arch/arm_m.rs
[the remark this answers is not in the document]
aapoalas: Oops, yeah I got thrown off and started thinking that this was a case of `30 - size.leading_zeros() == size.trailing_zeros()`.

sys/kern/src/task.rs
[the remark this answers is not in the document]
aapoalas: I was indeed thinking of eliminating bounds checks.

I'll see about checking this option out (at a lower priority), and if it proves to be an easy change then I'll check the asm output to see if there is any potential benefit there.

sys/kern/src/arch/arm_m.rs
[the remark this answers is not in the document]
aapoalas: I'll take a look at this and open a PR if it seems to make some sense.

sys/kern/src/arch/arm_m.rs
aapoalas: Agreed: Since I already have the changes locally, I'll try my hand at figuring out the asm output at play and see if this helps LLVM or just duplicates its work.

aapoalas: Alright, I couldn't find any difference in output (except my mind or different checked out commit from this PR playing tricks on me).

I can still open a PR, if you'd like: The full patch file is only 434 lines long, with 44 actual lines changed I think. Very small, guarantees the optimisation and it might have _some_ effect somewhere, even though I was unable to find that place (I managed to only really analyse `select` and `safe_copy`.)

sys/kern/src/arch/arm_m.rs
aapoalas: If you have the time to answer or perhaps have a ready link available: Are there any flags I can turn on to eg. get debug info included in Hubris builds? Or other ways to make viewing the assembly output easier? Running gdb raw on the output of `cargo xtask dist app.toml` proved to be somewhat convoluted a way to do so, especially with it being a 50/50 chance of `info address func` and `disass func` working or not.

sys/kern/src/arch/arm_m.rs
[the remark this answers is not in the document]
aapoalas: Seems to be a case of penny-wise-pound-foolish: The change does shave off 8 bytes from `force_fault` but adds double that to `syscall_entry`. Percentage wise it's a a victory, `force_fault` goes from 74 to 66 bytes while `syscall_entry` goes from 4198 to 4214. And `unsafe extern "C" fn handle_fault` doesn't seem to change at all.

sys/kern/src/task.rs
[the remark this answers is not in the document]
aapoalas: This probably also makes no difference, or is negative. The `unsafe extern "C" fn handle_fault` function includes a ton of `handle_fault` calls but already does not seemingly have a single bounds check panic site. Trying to remove from that would obviously only cause an unsigned integer overflow panic.

sys/kerncore/src/lib.rs
hawkw: :woman_shrugging: well, it's not terribly important to me --- i just went hunting for what the fault handling code at the end might be, and then, when I saw that it was just `false`, kind of went, "wait, what am I missing here?" for a bit.

PR #1952, opened 2024-12-13
website: note the existence of kernel bug #1943

cbiffle:


hawkw (approved):
DAYS SINCE LAST KERNEL BUG: :zero:

PR #1953, opened 2024-12-14
gimlet-seq: Record why the power state changed

### LINKED ISSUE #1950: Gimlet sequencer: Add ringbuf logging for source/cause of A2 transitions
Today in manufacturing we had a situation where 5 sleds had rebooted from the host's perspective, but the SPs had not reset. We were unable to find any evidence on the host side about what caused the reboots. Particularly curiously, the uptimes on all 5 sleds were almost identical, so it appears all five made A2 -> A0 transitions at roughly the same time.

@rmustacc suggested an easy extension to the sequencer ringbufs that would help in a situation like this: whenever the sequencer gets a request to change the power state, can we require callers to pass in something that describes the source or cause (e.g., request via control plane agent vs request from ipcc vs ...)?

hawkw:
Currently, when the Gimlet CPU sequencer changes the system's power
state, no information about *why* the power state changed is recorded by
the SP firmware. A system may power off or reboot for a variety of
reasons: it may be requested by the host OS over IPCC, by the control
plane over the management network, or triggered by the thermal task due
to an overheat condition. This makes debugging an unexpected reboot or
power off difficult, as the SP ringbuffers and other diagnostics do not
indicate why an unexpected power state change occurred. See #1950 for a
motivating example.

This commit resolves this as described in #1950 by adding a new field to
the `SetState` variant in the `drv-gimlet-seq-server` ringbuffer, so
that the reason a power state change occurred can be recorded. A new IPC
function, `Sequencer.set_state_with_reason`, is added to the `cpu_seq`
IPC API. This is equivalent to `Sequencer.set_state` but with the
addition of a `StateChangeReason` argument in addition to the desired
power state, and the sequencer task will record the provided reason in
its ringbuffer. This way, we can distinguish between the various reasons
a power state change may have occurred when debugging such issues.

All Hubris-internal callers of `Sequencer.set_state` are updated to
instead use `Sequencer.set_state_with_reason`. In particular,
`host-sp-comms` will record a variety of different `StateChangeReason`s,
allowing us to indicate whether the host requested a normal
power-off/reboot, the host OS panicked or failed to boot, or the host
CPU reset itself. Other callers like `control-plane-agent` and `thermal`
are simpler and just say "it was the control plane" or "overheat",
respectively. For backwards compatibility with existing callers of
`Sequencer.set_state` via `hiffy`, the `set_state` IPC is left as-is,
and will be recorded in the ringbuffer with `StateChangeReason::Other`.
Since all Hubris tasks now use the new API, `Other` *basically* just
means `hiffy`.

The `StateChangeReason` enum also generates counters, so that the total
number of power state changes can be tracked.

Also, while I was here, I've changed the `Trace::SetState` entry in the
`drv-gimlet-seq-server` ringbuf from a tuple-like enum variant to a
struct-like enum variant with named fields. This entry includes two
`PowerState` fields, one recording the previous power state and the
other recording the new power state that has been set. IMHO, using a
tuple-like variant to represent this is a bit unfortunate, as in
Humility, we'll see two values of the same type and it's not immediately
obvious which is the previous state and which is the new state. This
must be determined based on the order of the fields in the ringbuf
entry, which requires referencing the Hubris code to determine.

I felt like it was nicer to just use a struct-like variant with named
fields for this. That way, the semantic meaning of the two `PowerState`s
is actually encoded in the debug info, and Humility can just indicate
which is the previous state and which is the new state when displaying
the ring buffer. I also think it's a bit nicer to name the timestamp
field --- otherwise, it just looks like some arbitrary integer, and you
need to look at the code to determine that it's the timestamp of the
power state change.

Fixes #1950


hawkw:
argh, I [forgot that Gimletlet also implements the sequencer API](https://github.com/oxidecomputer/hubris/actions/runs/12332689182/job/34420831773?pr=1953)

drv/cpu-seq-api/src/lib.rs
```diff
@@ -31,6 +32,22 @@ pub enum SeqError {
...
+    InitialPowerOn = 1,
+    /// A power state change was requested by the control plane.
+    ControlPlane,
+    /// The host OS requested that the system power off without rebooting.
+    HostPowerOff,
+    /// The host OS requested that the system reboot.
+    HostReboot,
+    /// The system powered off because a component has overheated.
+    Overheat,
+}
```
hawkw, line 49: I'm very open to suggestions about the naming of these variants, if anyone dislikes the ones I came up with...

drv/cpu-seq-api/src/lib.rs
```diff
@@ -31,6 +32,22 @@ pub enum SeqError {
...
+pub enum StateChangeReason {
+    /// TThe system has just received power, so the sequencer has booted the
+    /// host CPU.
+    InitialPowerOn = 1,
+    /// A power state change was requested by the control plane.
+    ControlPlane,
+    /// The host OS requested that the system power off without rebooting.
+    HostPowerOff,
+    /// The host OS requested that the system reboot.
+    HostReboot,
```
hawkw, line 46: I also considered making these one thing, like
```rust
HostPowerOff { reboot: bool }
```
but that makes this enum two bytes, and probably also embiggens the ringbuf entry...

drv/mock-gimlet-seq-server/src/main.rs
```diff
@@ -58,6 +58,7 @@ impl idl::InOrderSequencerImpl for ServerImpl {
         &mut self,
         _: &RecvMessage,
         state: PowerState,
+        _: StateChangeReason,
```
hawkw, line 61: Perhaps we should add a ringbuffer to the mock sequencer implementation on Gimletlet, to aid in debugging in its callers?

drv/cpu-seq-api/src/lib.rs
```diff
@@ -31,6 +32,22 @@ pub enum SeqError {
     ServerRestarted,
 }
 
+#[derive(Copy, Clone, Debug, FromPrimitive, Eq, PartialEq, AsBytes, Count)]
+#[repr(u8)]
+pub enum StateChangeReason {
+    /// TThe system has just received power, so the sequencer has booted the
```
aapoalas, line 38: typo:
```suggestion
    /// The system has just received power, so the sequencer has booted the
```

drv/cpu-seq-api/src/lib.rs
aapoalas: I think you might get the desired effect with another enum; `HostPowerOff(PowerOffStyle)`

```rs
enum PowerOffStyle {
  Shutdown,
  Reboot
}
```

As I understand, as long as there aren't multiple enum variants with a value, and the only value is a non-repr enum then Rust will smash the two together. At least, that's what I've seen happening in one place in code I work with.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -90,7 +90,13 @@ enum Trace {
     RailsOn,
     UartEnabled,
     A0(u16),
-    SetState(PowerState, PowerState, u64),
+    SetState {
```
aapoalas, line 93: **praise:** Seems like a really good change to add the field names here.

drv/cpu-seq-api/src/lib.rs
hawkw: Huh, I didn't realize Rust would do that when the outermost enum has a `repr` attribute. That's kinda nice! I'll look into it, thanks!

drv/gimlet-seq-server/src/main.rs
hawkw: Yeah, tuple-like enum variants with two fields of the same type are almost always a bad move IMHO...

drv/cpu-seq-api/src/lib.rs
hawkw: Honestly, though, I'm not sure if that's really any nicer than the current thing, in this case --- the only reason I would've done it with a `bool` is so that the caller that uses this variant, the `power_off_host` function in `task-host-sp-comms`, could just pass its `reboot` field rather than having to say `if reboot { ... }`:
https://github.com/oxidecomputer/hubris/blob/63995e0cdd0e5087244bbd413db7d00b6654bfc5/task/host-sp-comms/src/main.rs#L376-L386

But, if we add another enum here, it still has to do that, so. I'm not sure if this is worth changing.

aapoalas: I was really confused and surprised when I found out as well, but it's pretty nice at times.

You could implement a `From<bool>` for the inner enum to hide the if-else away. That being said, it's arguably not that much nicer.

hawkw: Given that a `From<bool>` implementation would only be used in one place (`power_off_host` in `task-host-sp-comms`) but would have to be defined in `cpu-seq-api`, I think I'll leave it as-is. I think it's easier to understand when the conversion from the `reboot: bool` to a `StateChangeReason` occurs in the one place where it actually happens, instead of hiding it behind a `From` impl in a separate crate that isn't used anywhere else.

Thanks for the suggestions, though!

drv/cpu-seq-api/src/lib.rs
```diff
@@ -31,6 +32,22 @@ pub enum SeqError {
...
+pub enum StateChangeReason {
+    /// The system has just received power, so the sequencer has booted the
+    /// host CPU.
+    InitialPowerOn = 1,
+    /// A power state change was requested by the control plane.
+    ControlPlane,
+    /// The host OS requested that the system power off without rebooting.
+    HostPowerOff,
+    /// The host OS requested that the system reboot.
+    HostReboot,
```
rmustacc, line 46: While I realize it's a bit tricky, I think it'd be useful to distinguish between a reboot due to an IPCC panic and a reboot. Unfortunately the panic command doesn't itself indicate a reboot, but a subsequent one from the host with no additional action taken will indicate a panic. It may not be relevant at this particular point, but it depends a bit on where we see us generating an alert that the system has rebooted from. That is, it's not clear to me that we'd want to do that from all of the different callers versus use the sequencer as a generator.

I also somewhat expected to see a distinction between overheat and MAPO related actions, but the latter may be because if the FPGA sequencer triggers a MAPO event we are seeing that in a different way. 

If someone uses humility to manually turn on/off the host power how is that noted through here?

hawkw: > While I realize it's a bit tricky, I think it'd be useful to distinguish between a reboot due to an IPCC panic and a reboot. Unfortunately the panic command doesn't itself indicate a reboot, but a subsequent one from the host with no additional action taken will indicate a panic.

Definitely agree that it would be good to differentiate between these, I'll see about having the host-sp-comms task track a "was there just a panic" bit so we can thread that through here.

> I also somewhat expected to see a distinction between overheat and MAPO related actions, but the latter may be because if the FPGA sequencer triggers a MAPO event we are seeing that in a different way.

As I understand it, MAPO events won't go through the `Sequencer.set_state` IPC interface, and won't record the `SetState` ringbuf entry. Instead, I think Hubris' understanding of MAPO is entirely internal to the sequencer task --- I think the relevant ringbuf entries are the ones we record in [`a0_failure` here](https://github.com/oxidecomputer/hubris/blob/63995e0cdd0e5087244bbd413db7d00b6654bfc5/drv/gimlet-seq-server/src/main.rs#L937-L949). @bcantrill or @cbiffle would know better, but that's my understanding, at least.

drv/mock-gimlet-seq-server/src/main.rs
hawkw: @jgallagher yeah, that's why I didn't do it in this branch, it seemed straightforward enough to add later if someone actually needed it for some reason.

task/host-sp-comms/src/main.rs
```diff
@@ -250,6 +250,11 @@ struct ServerImpl {
     reboot_state: Option<RebootState>,
     host_kv_storage: HostKeyValueStorage,
     hf_mux_state: Option<HfMuxState>,
+    /// Set to true when the host OS panics, and unset when the system reboots.
+    ///
+    /// This is used to determine whether a host-triggered power-off is due to a
+    /// kernel panic or was a normal power-off.
+    host_just_panicked: bool,
```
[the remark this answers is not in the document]
hawkw: Yeah, I think that's a good idea. I'll add boot failure as well.

task/host-sp-comms/src/main.rs
```diff
@@ -374,11 +380,24 @@ impl ServerImpl {
     // basically only ever succeed in our initial set_state() request, so I
     // don't know how we'd test it
     fn power_off_host(&mut self, reboot: bool) {
+        let reason = if self.host_just_panicked {
+            // Clear the panic flag, so that subsequent power state transitions
+            // are not marked as panics unless the host OS once again panics.
```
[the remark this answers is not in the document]
hawkw: Ah, yeah, I had wondered about that --- didn't realize that the `host-sp-comms` task would get notified when somebody else changes the power state. I'll go add that.

task/host-sp-comms/src/main.rs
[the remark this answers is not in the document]
citrus-it: They are two different reasons but a host boot failure will be immediately followed by a panic. I still think it's useful to distinguish the two (you'd have to just treat boot failure as more relevant).

Of course a panic may not be immediately followed by a request for a reboot - if the kernel debugger is loaded we'll end up at a debugger prompt until somebody intervenes.

task/host-sp-comms/src/main.rs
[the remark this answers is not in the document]
hawkw: Okay, as of https://github.com/oxidecomputer/hubris/pull/1953/commits/2650ced4a3bcfaf0ef1384ee64b1f0961e794e2b, the `host-sp-comms` task is now aware of host boot failures and CPU resets in A0, as well as host panics and reboot/power off requests. And, it clears its understanding of the last power-off whenever the system reaches A2 per Jefe. I think this should be sufficient to address this as well as https://github.com/oxidecomputer/hubris/pull/1953#discussion_r1887496763?

drv/cpu-seq-api/src/lib.rs
hawkw: We're now able to distinguish between host boot failures, host panics, and host reboot/power-off requests, as of https://github.com/oxidecomputer/hubris/pull/1953/commits/2650ced4a3bcfaf0ef1384ee64b1f0961e794e2b. The change I made in https://github.com/oxidecomputer/hubris/pull/1953/commits/7e6b0d31f80a9821e6a72ec22821b8fe6611cbb3 to leave the IPCs currently used by Hiffy alone to avoid breaking scripts etc should cause any Hiffy-triggered power state change to have the reason `Other`. I considered explicitly making it `Hiffy`, but :woman_shrugging:

task/host-sp-comms/src/main.rs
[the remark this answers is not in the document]
hawkw: We should now understand host boot failure as a separate thing from host panics in https://github.com/oxidecomputer/hubris/pull/1953/commits/2650ced4a3bcfaf0ef1384ee64b1f0961e794e2b. We don't set the panic flag if the boot failure flag was already set, so I think we should be able to differentiate between them correctly.

If there's an easy way to force a boot failure on a bench system, I'd love to test that Hubris is able to understand all of these events correctly?

drv/gimlet-seq-server/src/main.rs
```diff
@@ -1029,11 +1045,21 @@ impl<S: SpiServer> idl::InOrderSequencerImpl for ServerImpl<S> {
     }
 
     fn set_state(
+        &mut self,
```
aapoalas, line 1048: **suggestion:** Consider marking this as deprecated if the intention is that this should no longer be used.

task/host-sp-comms/src/main.rs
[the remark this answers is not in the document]
jgallagher: > If there's an easy way to force a boot failure on a bench system

I've done this on accident many times! If you flash a mismatched phase 1 ROM to whatever is in the phase 2 BSU of the host, it'll go into a boot failure loop.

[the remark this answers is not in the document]
hawkw: Okay, I did some testing on a bench gimlet that has no host phase 2 image, so booting will fail predictably (thanks @rmustacc!), and it looks like the current code is treating the boot failure as a host panic:
```
  44  683        1        1 SetState { prev: A0PlusHP, next: A2, why: HostPanic, now: 0xa3941 }
  45 1422        1        1 V3P3SysA0VOut(Volts(3.3105469))
  46  692        1        1 PGStatus { b_pg: 0xff, c_pg: 0x3, nic: 0x7f }
  47  698        1        1 SMStatus { a1: 0x5, a0: 0xc }
  48  703        1        1 PowerControl(0x3)
  49  707        1        1 InterruptFlags(0x0)
  50  670        1        1 UpdateState(A2)
  51 1422        1        1 V3P3SysA0VOut(Volts(3.3105469))
  52  920        1        1 A2
  53 1422        1        1 V3P3SysA0VOut(Volts(1.1875))
  54 1422        1        1 V3P3SysA0VOut(Volts(0.9765625))
  55  534        1        1 Status { ier: 0x0, ifr: 0x0, amd_status: 0xb, amd_a0: 0xb }
  56  683        1        1 SetState { prev: A2, next: A0, why: HostReboot, now: 0xa4d9d }
```
so I've clearly messed something up in here...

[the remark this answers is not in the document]
hawkw: hahahaha, whoops: https://github.com/oxidecomputer/hubris/blob/2650ced4a3bcfaf0ef1384ee64b1f0961e794e2b/task/host-sp-comms/src/main.rs#L827-L830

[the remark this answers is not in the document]
hawkw: There's also a question about whether it's nicer for `host_sp_comms` to continue to use the same `reason` value when it sets the system back to A0 after a host boot failure/panic. Currently, we always set the reason for transitions back to A0 driven by `host_sp_comms` to `HostReboot`, because we clear the last power off reason when Jefe tells us that we've gone to A2. But perhaps we should instead wait to clear that until we get to A0 again?

drv/gimlet-seq-server/src/main.rs
hawkw: Hm, unfortunately, that's a bit tricky. We can't set an _implementation_ of a trait method as deprecated; instead, the trait itself must put the deprecated annotation on the definition of the trait method. Deprecating the method only in an `impl` block has no effect, and in fact, [results in a compiler error](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=20e207ebba3693a11c78832253974bc7). Because this is an Idol IPC interface, the trait `InOrderSequencerImpl` is not defined in Rust code anywhere, but is instead generated by Idol's codegen, so we can't easily add the deprecated attribute to it without changing the code generator to allow adding attributes to trait methods. And, *consumers* of this IPC interface don't actually interact with the `InOrderSequencerImpl` trait. Instead, that's used by the Idol server code to implement the IPC *server*. Clients of this interface interact with it through different generated code, so we would actually need to put the deprecated attribute on *that*.

It would be nice to have Idol support adding attributes such as `#[deprecated]` to generated client code, but that's something we'd have to go implement separately...

task/host-sp-comms/src/main.rs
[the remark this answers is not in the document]
hawkw: Okay, here we go; *now* it does the right thing (as of 108894718aec404173b363f5c6f77658e3af8dbb), here's an excerpt from the sequencer ringbuffer when a host boot fails:
```
  44  683        1        1 SetState { prev: A0PlusHP, next: A2, why: HostBootFailure, now: 0xb580a }
  45 1422        1        1 V3P3SysA0VOut(Volts(3.3105469))
  46  692        1        1 PGStatus { b_pg: 0xff, c_pg: 0x3, nic: 0x7f }
  47  698        1        1 SMStatus { a1: 0x5, a0: 0xc }
  48  703        1        1 PowerControl(0x3)
  49  707        1        1 InterruptFlags(0x0)
  50  670        1        1 UpdateState(A2)
  51 1422        1        1 V3P3SysA0VOut(Volts(3.2480469))
  52  920        1        1 A2
  53 1422        1        1 V3P3SysA0VOut(Volts(1.1953125))
  54 1422        1        1 V3P3SysA0VOut(Volts(0.9921875))
  55  534        1        1 Status { ier: 0x0, ifr: 0x0, amd_status: 0xb, amd_a0: 0xb }
  56  683        1        1 SetState { prev: A2, next: A0, why: HostBootFailure, now: 0xb6c66 }
```

The whole ringbuffer is also interesting, as you can also see the `SetState` call with `why: Other` when I manually poke the machine into A0 using `humility hiffy`:

<details>
<summary>Full sequencer ringbuffer contents</summary>

```
   0  273        1        1 Ice40Rails(true, true)
   1  302        1        1 Ice40PowerGoodV1P2(true)
   2  323        1        1 Ice40PowerGoodV3P3(true)
   3  365        1        1 IdentValid(false)
   4  368        1        1 ChecksumValid(false)
   5  371        1        1 Reprogram(true)
   6  385        1        1 Programming
   7  414        1        1 Programmed
   8  418        1        1 RailsOff
   9  421        1        1 Ident(0xde01)
  10  427        1        1 A2Status(0x0)
  11  450        1      446 ClockConfigWrite
  12  461        1        1 ClockConfigSuccess
  13 1422        1        1 V3P3SysA0VOut(Volts(0.10546875))
  14  463        1        1 A2
  15 1235        1        1 SpdDimmsFound(0x10)
  16  683        1        1 SetState { prev: A2, next: A0, why: Other, now: 0x865dd }
  17 1422        1        1 V3P3SysA0VOut(Volts(0.091796875))
  18  692        1        1 PGStatus { b_pg: 0x0, c_pg: 0x0, nic: 0x0 }
  19  698        1        1 SMStatus { a1: 0x0, a0: 0x0 }
  20  703        1        1 PowerControl(0x0)
  21  707        1        1 InterruptFlags(0x0)
  22  749        1        1 A1Status(0x2)
  23  749        1      100 A1Status(0x3)
  24  749        1        1 A1Status(0x5)
  25  767        1        1 CPUPresent(true)
  26  777        1        1 Coretype { coretype: true, sp3r1: true, sp3r2: false }
  27  805        1       11 A0Status(0x1)
  28  805        1      284 A0Status(0x2)
  29  805        1        1 A0Status(0x4)
  30  805        1        2 A0Status(0x6)
  31  805        1        1 A0Status(0x7)
  32  827        1        1 RailsOn
  33  839        1        1 A0Power(0x7)
  34  839        1        9 A0Power(0xa)
  35  839        1        4 A0Power(0xb)
  36  839        1        1 A0Power(0xc)
  37  864        1        1 UartEnabled
  38  868        1        1 A0(0x33e)
  39  670        1        1 UpdateState(A0)
  40  534        1    16649 Status { ier: 0x0, ifr: 0x0, amd_status: 0xc, amd_a0: 0x4 }
  41  564        1        1 NICPowerEnableLow(false)
  42  670        1        1 UpdateState(A0PlusHP)
  43  534        1      257 Status { ier: 0x0, ifr: 0x0, amd_status: 0xc, amd_a0: 0x4 }
  44  683        1        1 SetState { prev: A0PlusHP, next: A2, why: HostBootFailure, now: 0xb580a }
  45 1422        1        1 V3P3SysA0VOut(Volts(3.3105469))
  46  692        1        1 PGStatus { b_pg: 0xff, c_pg: 0x3, nic: 0x7f }
  47  698        1        1 SMStatus { a1: 0x5, a0: 0xc }
  48  703        1        1 PowerControl(0x3)
  49  707        1        1 InterruptFlags(0x0)
  50  670        1        1 UpdateState(A2)
  51 1422        1        1 V3P3SysA0VOut(Volts(3.2480469))
  52  920        1        1 A2
  53 1422        1        1 V3P3SysA0VOut(Volts(1.1953125))
  54 1422        1        1 V3P3SysA0VOut(Volts(0.9921875))
  55  534        1        1 Status { ier: 0x0, ifr: 0x0, amd_status: 0xb, amd_a0: 0xb }
  56  683        1        1 SetState { prev: A2, next: A0, why: HostBootFailure, now: 0xb6c66 }
  57 1422        1        1 V3P3SysA0VOut(Volts(0.09765625))
  58  692        1        1 PGStatus { b_pg: 0x0, c_pg: 0x0, nic: 0x0 }
  59  698        1        1 SMStatus { a1: 0x0, a0: 0x0 }
  60  703        1        1 PowerControl(0x0)
  61  707        1        1 InterruptFlags(0x0)
  62  749        1        1 A1Status(0x2)
  63  749        1      100 A1Status(0x3)
  64  749        1        1 A1Status(0x5)
  65  767        1        1 CPUPresent(true)
  66  777        1        1 Coretype { coretype: true, sp3r1: true, sp3r2: false }
  67  805        1       11 A0Status(0x1)
  68  805        1      306 A0Status(0x2)
  69  805        1        1 A0Status(0x4)
  70  805        1        2 A0Status(0x6)
  71  805        1        1 A0Status(0x7)
  72  827        1        1 RailsOn
  73  839        1        1 A0Power(0x7)
  74  839        1        9 A0Power(0xa)
  75  839        1        4 A0Power(0xb)
  76  839        1        1 A0Power(0xc)
  77  864        1        1 UartEnabled
  78  868        1        1 A0(0x36a)
  79  670        1        1 UpdateState(A0)
  80  534        1      291 Status { ier: 0x0, ifr: 0x0, amd_status: 0xc, amd_a0: 0x4 }
```

</details>

hawkw:
@jgallagher Alright, I think all the review feedback has been addressed and I'd love another review whenever you've got the chance!

jgallagher (approved):
LGTM :+1:

drv/gimlet-seq-server/src/main.rs
```diff
@@ -1029,11 +1045,21 @@ impl<S: SpiServer> idl::InOrderSequencerImpl for ServerImpl<S> {
...
+        msg: &RecvMessage,
+        state: PowerState,
+    ) -> Result<(), RequestError<SeqError>> {
+        self.set_state_with_reason(msg, state, StateChangeReason::Other)
```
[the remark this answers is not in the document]
hawkw: Ah, hmm, that's a fair point even if it isn't an issue in practice. Perhaps I should change this back to my original approach of having a private method that both call into...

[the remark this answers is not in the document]
hawkw: Changed in 007470bdac2b56c18e959144b5566205691d81f8

PR #1956, opened 2024-12-16
kern: stack watermark support

cbiffle:
Each task has grown a new field, `low_stack_pointer`, that tracks the lowest stack pointer value seen so far (because stacks grow down). We update this value on any kernel entry (interrupt, timer, fault, syscall), so while we will miss very brief excursions of stack _between_ syscalls or interrupts, we should still get a useful value.

In addition, the field `past_low_stack_pointer` tracks the lowest stack pointer value across _all previous incarnations_ of the task. This way if the task is periodically restarting, we can still get useful stack stats, including at stack overflow.

All of this stuff is conditional on the `stack-watermark` feature, since it increases the kernel RAM requirement, which is explicit in Oxide's Hubris build system.

### DIFF
--- a/sys/kern/src/arch/arm_m.rs
+++ b/sys/kern/src/arch/arm_m.rs
@@ -936,6 +936,17 @@ static TICKS: [AtomicU32; 2] = {
 #[no_mangle]
 pub unsafe extern "C" fn SysTick() {
     crate::profiling::event_timer_isr_enter();
+
+    // We don't need the current task pointer in this routine, but we'll access
+    // it briefly to update the stack watermark:
+    {
+        let current = CURRENT_TASK_PTR.load(Ordering::Relaxed);
+        // Safety: `CURRENT_TASK_PTR` is valid once the kernel is started, and
+        // this interrupt is only enabled once the kernel is started.
+        let t = unsafe { &mut *current };
+        t.update_stack_watermark();
+    }
+
     with_task_table(|tasks| {
         // Load the time before this tick event.
         let t0 = TICKS[0].load(Ordering::Relaxed);
@@ -1106,6 +1117,15 @@ pub unsafe extern "C" fn DefaultHandler() {
             options(pure, nomem, preserves_flags, nostack),
         );
         ipsr & 0x1FF
+    };
+
+    let current = CURRENT_TASK_PTR.load(Ordering::Relaxed);
+    uassert!(!current.is_null()); // irq before kernel started?
+
+    let current_prio = {
+        let t = unsafe { &mut *current };
+        t.update_stack_watermark();
+        t.priority()
     };
 
     // The first 16 exceptions are architecturally defined; vendor hardware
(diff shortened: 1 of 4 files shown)


sys/kern/src/arch/arm_m.rs
```diff
@@ -1108,6 +1119,15 @@ pub unsafe extern "C" fn DefaultHandler() {
         ipsr & 0x1FF
     };
 
+    let current = CURRENT_TASK_PTR.load(Ordering::Relaxed);
+    uassert!(!current.is_null()); // irq before kernel started?
+
+    let current_prio = {
```
aapoalas, line 1125: **question:** The current_prio here seems like it's not being used at all? Is it intentional?

sys/kern/src/arch/arm_m.rs
```diff
@@ -936,6 +936,17 @@ static TICKS: [AtomicU32; 2] = {
 #[no_mangle]
 pub unsafe extern "C" fn SysTick() {
     crate::profiling::event_timer_isr_enter();
+
+    // We don't need the current task pointer in this routine, but we'll access
+    // it briefly to update the stack watermark:
+    {
```
hawkw, line 942: i  wonder if this whole block should be conditional on the feature flag? I see that `update_stack_watermark` is a nop if the feature isn't enabled, and the compiler is *probably* smart enough to eliminate the `CURRENT_TASK_PTR` access when the function that's called on it doesn't do anything, but... I just kinda don't trust it...

hawkw (approved):
Overall, very straightforward! Looks good to me once Clippy is made happy again.

sys/kern/src/arch/arm_m.rs
cbiffle: I was trying to make as little code conditional as possible, so that it always gets compiled; this is why it's only the body of update_watermark that's cfg'd. The compiler does appear to optimize this load out, fwiw.

sys/kern/src/task.rs
```diff
@@ -49,6 +49,20 @@ pub struct Task {
     /// Pointer to the ROM descriptor used to create this task, so it can be
     /// restarted.
     descriptor: &'static TaskDesc,
+
+    /// Tracks the lowest stack pointer value (e.g. fullest stack) observed on
```
mkeeter, line 53: Vibes – I'd vote to put this into a separate object with helper functions, e.g.

```
#[cfg(feature = "stack-watermark")]
struct StackWatermark {
    /// Tracks the lowest stack pointer value (e.g. fullest stack) observed on
    /// any kernel entry for this instance of this task.
    ///
    /// Initialized to `u32::MAX` if the task has not yet run.
    #[cfg(feature = "stack-watermark")]
    stack_pointer_low: u32,

    /// Tracks the lowest stack pointer value (e.g. fullest stack) observed on
    /// any kernel entry across *any* instance of this task.
    ///
    /// Initialized to `u32::MAX` if the task has not yet run.
    #[cfg(feature = "stack-watermark")]
    past_stack_pointer_low: u32,
}

#[cfg(feature = "stack-watermark")]
impl Default for StackWatermark {
    fn default() -> Self {
        StackWatermark {
            stack_pointer_low: u32::MAX,
            past_stack_pointer_low: u32::MAX,
        }
    }
}

#[cfg(feature = "stack-watermark")]
impl StackWatermark {
    fn reinitialize() {
        self.past_stack_pointer_low =
            u32::min(self.past_stack_pointer_low, self.stack_pointer_low);
        self.stack_pointer_low = u32::MAX;
    }

    fn update(&mut self, stack_pointer: u32) {
        self.stack_pointer_low =
            u32::min(self.stack_pointer_low, self.save().stack_pointer());
    }
}
```

cbiffle: Sold.

PR #1959, opened 2024-12-17
auxflash: make startup a lot faster.

cbiffle:
The auxflash server was computing a SHA3 of every potentially occupied slot in the QSPI flash, only to compare it to _the stored SHA3_ and then compare it to _the expected SHA3_ and then throw it away. This has been causing Sidecar startup to be linear in the number of valid images that have ever been flashed to auxflash, up to 16.

This change rearranges the logic, at least for startup. For each chunk, we now see if it even claims to have the right SHA. Only then do we compute the actual SHA to validate. This reduces an 11.6s delay observed on one Sidecar to just over 1s, and knocks 6s off the boot (something else is still delaying for 5s).

I haven't changed the behavior of the externally exposed `read_slot_checksum` operation, because it has no documentation and I can't figure out what it's used for, so I'm not sure I would get the semantics right or know how to test it.

### DIFF
--- a/drv/auxflash-server/src/main.rs
+++ b/drv/auxflash-server/src/main.rs
@@ -161,7 +161,7 @@ impl ServerImpl {
         &self,
         slot: u32,
     ) -> Result<AuxFlashChecksum, AuxFlashError> {
-        read_slot_checksum(&self.qspi, slot)
+        read_and_check_slot_checksum(&self.qspi, slot)
     }
 
     /// Checks that the matched slot in this even/odd pair also has valid data.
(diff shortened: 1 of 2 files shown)


drv/auxflash-server/src/main.rs
```diff
@@ -434,16 +434,36 @@ impl NotificationHandler for ServerImpl {
...
+            continue;
+        }
+
+        let Ok(actual) = handle.calculate_checksum() else {
+            // TODO: this ignores I/O errors, but, this is how the code has
+            // always been structured...
+            continue;
+        };
+
+        if chck == actual {
```
jgallagher, line 459: Not sure if there are already ringbuf logs in this task, but - should we log if this condition is false? It seems like that would be very surprising.

jgallagher (approved):
LGTM :+1:

Aaron-Hartwig (approved):
Thanks for this improvement, Cliff!

drv/auxflash-server/src/main.rs
cbiffle: I'm not sure that it's that surprising; a difference between actual and claimed checksum could occur if a write is aborted partway through, for example.

PR #1960, opened 2024-12-17
Improve context switch / MPU loading performance

cbiffle:
After discovering that MPU loading was a _significant_ fraction of the time we were spending in the kernel, I took a pass over it. The primary changes here are:

1. The existing functions we use to compute the RASR/RBAR/RLAR and other MPU registers, all of which sound suspiciously like a mumbling tyrannosaur, have been made `const fn` so they can be evaluated at compile time.
2. The build system now generates code for the region tables that call these `const fn`s, so that the MPU code is no longer being evaluated at runtime.
3. In support of this, the `RegionDesc` internal kernel structure has grown a new field to hold _unspecified architecture specific data,_ which for our current architectures happens to be a pair of words to be loaded into the MPU (RASR/RBAR or RBAR/RLAR depending on tyrannosaur version).
4. Finally, the actual code for _loading_ the MPU was underperforming on v6m/v7m. I simplified it. Now we turn off the MPU before loading just like we do on v8m, and don't have to do multiple writes with bit manipulation anymore.


cbiffle:
Looks like I get to run around adjusting all the kernel size allocations. I'll push a commit for that here in a bit.

sys/kern/src/arch/arm_m.rs
```diff
@@ -360,84 +459,102 @@ pub fn apply_memory_protection(task: &task::Task) {
...
+    // Safety: same as above, has no safety implications really.
+    unsafe {
+        mpu.ctrl.write(0b101);
+    }
+}
+
+#[cfg(armv8m)]
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub struct RegionDescExt {
+    rlar_disabled: u32,
```
labbott, line 500: Can you add add comment explaining this is the RLAR with everything except the `enable` bit set?  It took me a bit to figure out that was what `rlar_disabled` meant

sys/kern/src/arch/arm_m.rs
```diff
@@ -360,84 +459,102 @@ pub fn apply_memory_protection(task: &task::Task) {
         &*cortex_m::peripheral::MPU::PTR
     };
 
+    // Turn off the MPU.
+    //
+    // Safety: this has no actual memory safety implications, except for
+    // potentially exposing the kernel to a NULL dereference that succeeds.
+    unsafe {
+        mpu.ctrl.write(0);
+    }
```
labbott, line 468: I really wish ARMv7m gave a sample sequence for the MPU. IIRC the reason we turn off/on on v8m is because that was the sequence specified.

sys/kern/src/arch/arm_m.rs
```diff
@@ -73,6 +73,7 @@
 use core::arch::{self, global_asm};
 use core::sync::atomic::{AtomicBool, AtomicPtr, AtomicU32, Ordering};
 
+use serde::{Deserialize, Serialize};
```
labbott, line 76: It feels a little weird to be bringing these into the kernel, mostly because those seem much higher level. I don't have a better idea.

sys/kern/src/arch/arm_m.rs
```diff
@@ -349,6 +350,104 @@ pub fn reinitialize(task: &mut task::Task) {
...
+    // On v6/7-M the MPU expresses size of a region in log2 form _minus
+    // one._ So, the minimum allowed size of 32 bytes is represented as 4,
+    // because `2**(4 + 1) == 32`.
+    //
+    // We store sizes in the region table in an architecture-independent
+    // form (number of bytes) because it simplifies basically everything
+    // else but this routine. Here we must convert between the two -- and
+    // quickly, because this is called on every context switch.
+    //
+    // The image-generation tools check at build time that region sizes are
```
labbott, line 418: Might be worth unifying where this check occurs later to remove some code from big scary xtask

labbott (commented):
Minor comments but LGTM

sys/kern/src/arch/arm_m.rs
cbiffle: I just had to go convince myself that we _need_ `rlar_disabled` so I could explain it ... I think we actually don't, but I'll leave it in in the interest of not breaking things. I will add an explanatory comment and a TODO.

sys/kern/src/arch/arm_m.rs
cbiffle: Let me remind myself where this is actually used nowadays, maybe we can remove it.

sys/kern/src/arch/arm_m.rs
cbiffle: We could actually move the checks into this function, now that we expect it to be evaluated only in const context.....

sys/kern/src/arch/arm_m.rs
cbiffle: Yeah, agreed.

For those following along, @labbott did manage to track down this: https://github.com/ARM-software/m-profile-user-guide-examples/tree/main/Memory_model/rtos_context_switch

That is, unfortunately, wrong, and were we to follow that example we would recreate the crash previously fixed by #1193.

sys/kern/src/arch/arm_m.rs
hawkw: It seems it's required here because the (cross-platform) `RegionDesc` in `kern::descs` derives `Serialize` and `Deserialize`, and the `RegionDescExt` struct must also to not break that. I don't actually see any obvious places where we serialize or deserialize `RegionDesc`s --- I wonder if this is a historical artifact. Did we once use these types in the build code when deserializing an application's config file? It looks like we now use a different type in `build_kconfig`: https://github.com/oxidecomputer/hubris/blob/874c3b899af5833a501580bd5ad10f8acb3fe3d0/build/kconfig/src/lib.rs#L96-L118

cbiffle: Good news! I can remove it, and will push the commit shortly.

sys/kern/src/arch/arm_m.rs
cbiffle: I agree. We could actually do the check from the const fn nowadays if we wanted. Followup maybe.

aapoalas:
Just to bring visibility, I took a look at splitting the `RegionDescExt` data being added here away from the existing `RegionDesc` data, and I think I got a pretty nice solution out of it in the end. I opened it as a PR against this branch in Cliff's fork: https://github.com/cbiffle/hubris-fork/pull/1 (It seemed like a good idea at the time.)

I'm not convinced it is the best choice, but I'm fairly convinced it's not an asinine choice. It may not be worth bringing into this PR, but maybe it would be worth a follow-up?

labbott (approved):
LGTM

cbiffle:
> Just to bring visibility, I took a look at splitting the `RegionDescExt` data being added here away from the existing `RegionDesc` data, and I think I got a pretty nice solution out of it in the end. I opened it as a PR against this branch in Cliff's fork: [cbiffle#1](https://github.com/cbiffle/hubris-fork/pull/1) (It seemed like a good idea at the time.)

Neat, I'm happy to take a look at that @aapoalas. It may be a little while, schedules are starting to get weird due to the encroaching holidays. But I can test the cache effects of that pretty easily, I think.

PR #1961, opened 2024-12-26
kern: fix client-controlled kernel panic in kipc

cbiffle:
It turns out it's possible to trigger a bounds check in kipc handling code by submitting an invalid IRQ number. This fixes that, causing the bounds check to be optimized out.

In several syscall paths (and one kipc) we manipulate interrupts without accepting an interrupt number from the client. In these cases, a bad interrupt number would indicate a problem in our dispatch table, so I've made them explicit panics. (They were panics before, just implicit ones in the array access.)

This causes a small increase in kernel text size (40-64 bytes), but kipc should not be able to crash the kernel!

### DIFF
--- a/sys/kern/src/arch/arm_m.rs
+++ b/sys/kern/src/arch/arm_m.rs
@@ -83,7 +83,7 @@ use crate::time::Timestamp;
 use crate::umem::USlice;
 #[cfg(any(armv7m, armv8m))]
 use abi::FaultSource;
-use abi::{FaultInfo, InterruptNum};
+use abi::{FaultInfo, InterruptNum, UsageError};
 #[cfg(armv8m)]
 use armv8_m_mpu::{disable_mpu, enable_mpu};
 use unwrap_lite::UnwrapLite;
@@ -1233,7 +1233,10 @@ pub unsafe extern "C" fn DefaultHandler() {
                 .unwrap_or_else(|| panic!("unhandled IRQ {irq_num}"));
 
             let switch = with_task_table(|tasks| {
-                disable_irq(irq_num, false);
+                // This can only fail if the IRQ number is out of range, which
+                // in this case would mean the hardware is conspiring against
+                // us. So ignore it to ensure we don't generate a bogus check.
+                disable_irq(irq_num, false).ok();
 
                 // Now, post the notification and return the
                 // scheduling hint.
@@ -1250,22 +1253,29 @@ pub unsafe extern "C" fn DefaultHandler() {
     crate::profiling::event_isr_exit();
 }
 
-pub fn disable_irq(n: u32, also_clear_pending: bool) {
+pub fn disable_irq(n: u32, also_clear_pending: bool) -> Result<(), UsageError> {
     // Disable the interrupt by poking the Interrupt Clear Enable Register.
     let nvic = unsafe { &*cortex_m::peripheral::NVIC::PTR };
     let reg_num = (n / 32) as usize;
     let bit_mask = 1 << (n % 32);
     unsafe {
-        nvic.icer[reg_num].write(bit_mask);
+        nvic.icer
+            .get(reg_num)
+            .ok_or(UsageError::NoIrq)?
+            .write(bit_mask);
     }
     if also_clear_pending {
         unsafe {
-            nvic.icpr[reg_num].write(bit_mask);
+            nvic.icpr
+                .get(reg_num)
+                .ok_or(UsageError::NoIrq)?
+                .write(bit_mask);
         }
     }
+    Ok(())
 }
 
-pub fn enable_irq(n: u32, also_clear_pending: bool) {
+pub fn enable_irq(n: u32, also_clear_pending: bool) -> Result<(), UsageError> {
     // Enable the interrupt by poking the Interrupt Set Enable Register.
     let nvic = unsafe { &*cortex_m::peripheral::NVIC::PTR };
     let reg_num = (n / 32) as usize;
@@ -1273,17 +1283,24 @@ pub fn enable_irq(n: u32, also_clear_pending: bool) {
     if also_clear_pending {
         // Do this _before_ enabling.
         unsafe {
-            nvic.icpr[reg_num].write(bit_mask);
+            nvic.icpr
+                .get(reg_num)
+                .ok_or(UsageError::NoIrq)?
+                .write(bit_mask);
         }
     }
     unsafe {
-        nvic.iser[reg_num].write(bit_mask);
+        nvic.iser
+            .get(reg_num)
+            .ok_or(UsageError::NoIrq)?
+            .write(bit_mask);
     }
+    Ok(())
 }
 
 /// Looks up an interrupt in the NVIC and returns a cross-platform
 /// representation of that interrupt's status.
-pub fn irq_status(n: u32) -> abi::IrqStatus {
+pub fn irq_status(n: u32) -> Result<abi::IrqStatus, UsageError> {
     let mut status = abi::IrqStatus::empty();
 
     let nvic = unsafe { &*cortex_m::peripheral::NVIC::PTR };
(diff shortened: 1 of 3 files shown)


sys/kern/src/arch/arm_m.rs
```diff
@@ -1210,7 +1210,10 @@ pub unsafe extern "C" fn DefaultHandler() {
                 .unwrap_or_else(|| panic!("unhandled IRQ {irq_num}"));
 
             let switch = with_task_table(|tasks| {
-                disable_irq(irq_num, false);
+                // This can only fail if the IRQ number is out of range, which
+                // in this case would mean the hardware is conspiring against
+                // us. So ignore it to ensure we don't generate a bogus check.
+                disable_irq(irq_num, false).ok();
```
hawkw, line 1216: <s>I know you prefer this over `let _` because the latter is a bit of a footgun in other contexts, but there is a part of me that kind of wonders if 
```rust
fn_returning_result().ok();
```
optimizes worse than 
```rust
let _ = fn_returning_result();
```
because it's a function call, though hopefully LLVM can eliminate it...I might have to go check.</s>

**EDIT**: After spending some time in Compiler Explorer, I have been proven definitively wrong about this, in fact, the `opt-level=2` compiler is so smart about this that it correctly realizes that these are actually *the same function* and [doesn't even generate two separate functions for them](https://godbolt.org/z/hM75ThrYj):
```rust
#[no_mangle]
#[inline(never)]
pub extern fn ignore_err_ok() {
    returns_result().ok();
    // pretend the function then does other stuff
    core::hint::black_box(1);
}

#[no_mangle]
#[inline(never)]
pub extern fn ignore_err_underscore() {
    let _ = returns_result();
    // pretend the function then does other stuff
    core::hint::black_box(1);
}
```

So, I have seen the error of my ways and should not have doubted the use of `.ok()` to ignore `Result`s!.

sys/kern/src/syscalls.rs
```diff
@@ -918,9 +920,10 @@ fn irq_status(
         )))?;
 
     // Combine the platform-level status of all the IRQs in the notification set.
-    let mut status = irqs.iter().fold(IrqStatus::empty(), |status, irq| {
-        status | crate::arch::irq_status(irq.0)
-    });
+    let mut status =
+        irqs.iter().try_fold(IrqStatus::empty(), |status, irq| {
+            crate::arch::irq_status(irq.0).map(|n| status | n)
+        })?;
```
hawkw, line 926: it's nice how this was so trivially converted to fallible with `try_fold`! :)

sys/kern/src/arch/arm_m.rs
```diff
@@ -1269,25 +1286,30 @@ pub fn irq_status(n: u32) -> abi::IrqStatus {
...
     status.set(abi::IrqStatus::PENDING, pending);
 
-    status
+    Ok(status)
 }
 
-pub fn pend_software_irq(InterruptNum(n): InterruptNum) {
+pub fn pend_software_irq(
+    InterruptNum(n): InterruptNum,
+) -> Result<(), UsageError> {
```
hawkw, line 1303: I was about to say "hmm, shouldn't we update the documentation for the `software_irq` KIPC to reflect that it will fault you if you ask for a nonexistent IRQ...but then, I realized that the Hubris reference docs for KIPCs [doesn't even mention `software_irq`, and instead suggests that `find_faulted_task` is KIPC 8](https://hubris.oxide.computer/reference/#_find_faulted_task_8), which it isn't (it's KIPC 9, and `software_irq` is 8). This is almost certainly my fault back when I added the software IRQ KIPC, so I should probably fix that...

sys/kern/src/arch/arm_m.rs
```diff
@@ -1227,40 +1230,54 @@ pub unsafe extern "C" fn DefaultHandler() {
...
     }
     if also_clear_pending {
         unsafe {
-            nvic.icpr[reg_num].write(bit_mask);
+            nvic.icpr
+                .get(reg_num)
+                .ok_or(UsageError::NoIrq)?
+                .write(bit_mask);
         }
     }
```
hawkw, line 1251: if we wanted to be really fastidious about code size, we _could_ potentially do a single check that the IRQ number is in range, and then make both of the actual accessess unchecked; if rustc isn't smart enough to know that if it's in range for ICER, it's also in range for ICPR? but, I dunno if it's worth the more manual checking to save 40 bytes of kernel --- feels sketchier for sure.

sys/kern/src/kipc.rs
```diff
@@ -462,7 +463,9 @@ fn software_irq(
         )))?;
 
     for &irq in irqs.iter() {
-        crate::arch::pend_software_irq(irq);
+        // Any error here would be a problem in our dispatch table, not the
+        // caller, so we panic because we want to hear about it.
+        crate::arch::pend_software_irq(irq).unwrap_lite();
```
hawkw, line 468: Hmm --- this is the only call site of `pend_software_irq`, right? So, if we're going to unwrap here, we _could_ just leave that function using infallible array indexing and let it panic, instead. I dunno if that has a meaningful impact on binary size, though, and this is maybe nicer in the event that we add other uses of `pend_software_irq` later, which _don't_ want to panic. I think having it be the caller's decision whether or not to panic is almost certainly the ideologically nicer factoring...

hawkw (approved):
This looks good to me, great catch! I commented on some ways we could potentially golf the binary size impact down a bit but I don't really think any of them are actually worth pursuing...

hawkw:
Incidentally --- and this is *extremely* pedantic of me --- I don't think the title/commit message message for this PR is quite right. The client-controlled panic is actually in the syscalls, rather than the `software_irq` KIPC. The KIPC implementation is the one that still `unwrap_lite()`s and will panic, as it gets the notification-mask-to-IRQ mappings from the kernel's dispatch table.

sys/kern/src/arch/arm_m.rs
cbiffle: It's always worth checking in cases like this! The best codegen changes over time, too, so the ground truth back when I developed this habit might be different from today. So, thanks for chasing it down.

sys/kern/src/syscalls.rs
cbiffle: Yeah, I don't reach for the fold family of functions often, but sometimes they're just the ticket!

sys/kern/src/arch/arm_m.rs
cbiffle: Interesting. I'd be curious how much that'd actually save -- probably both the error return paths for these expressions get merged into a single epilogue, but I could be wrong.

I feel like the result is potentially more fragile, though, so I hesitate.

sys/kern/src/kipc.rs
cbiffle: We'd want to provide callers with some way of checking if an IRQ is valid for the architecture then, since it's in the `arch` module. The current factoring makes _it_ responsible for the check, using its architecture-specific knowledge, which I suspect results in simpler code at the use sites.

But as you say, there is only one use site atm, and I'm not sure how likely that is to change.

mkeeter:
Should we add this to the kernel test suite?

sys/kern/src/arch/arm_m.rs
hawkw: Yeah, I think this is fine! It's just a thought that occurred to me.

hawkw:
> > Incidentally --- and this is _extremely_ pedantic of me --- I don't think the title/commit message message for this PR is quite right. The client-controlled panic is actually in the syscalls, rather than the `software_irq` KIPC. The KIPC implementation is the one that still `unwrap_lite()`s and will panic, as it gets the notification-mask-to-IRQ mappings from the kernel's dispatch table.
> 
> I think we're just using the terms a little differently -- the client-controlled panic here is in the kernel implementation of the KIPC operation, which is called from the syscall dispatcher (like all KIPC implementations). So the panic is not available to tasks that don't use KIPC.
> 
> Is there a wording of that you'd find clearer?

Ah, I see what you mean, nevermind, the confusion is purely on my end here.

PR #1962, opened 2024-12-26
kern: reply buffer capacity checking

cbiffle:
The syscall docs say that attempting to reply with a message that's too big for your client to accept is a programming error. Programming errors are expected to produce faults in the program with the error. This fault is documented as being of type `ReplyTooLarge`.

However, there is no such fault defined in the ABI, and the actual behavior of the kernel is to deliver the prefix of the reply that fits, and hand the client the length of that prefix. In our current world, this means the client will generally die shortly thereafter with a deserialization error.

In a potential future world where messages can contain variable length portions, the client might not even hit an error when the reply is truncated.

The code in the kernel has been marked with a TODO since forever. This commit resolves the TODO by adding a check and fault reason.

### DIFF
--- a/doc/syscalls.adoc
+++ b/doc/syscalls.adoc
@@ -332,7 +332,7 @@ There is only one way to break `REPLY`, and that's with a bogus slice.
 | `MemoryAccess`
 
 | Reply message is longer than recipient requested.
-| `ReplyTooLarge`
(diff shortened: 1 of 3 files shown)


cbiffle:
Whoops, this is a duplicate.

PR #1963, opened 2024-12-26
kern: rework ARMv8-M context switch, fixing a bug?

cbiffle:
I was playing with reducing ARMv8-M context switch time, by looking into the same sort of changes to MPU loading that I recently did on v6/7.

I noticed that we only ever bitwise-OR values into the MAIR registers. In other words, the first task that gets activated loads its exact attributes into MAIR for each region; the _second_ task combines its attributes using bitwise OR; and so on until the registers contain the bitwise OR of all regions in all tasks. (Note that these are the cacheability/sharability attributes, _not_ the access permissions, which are set correctly. The main implication of this would be accidentally caching device memory or breaking DMA.)

That seemed bad, so I removed it and reworked the MPU loading loop. Now we build up the MAIR contents in a local before writing them -- no more read-modify-write.

Halting the processor in task code, I was able to observe a meaningful distinction in MAIR contents: previously tasks were unable to set memory as device (which involves more 1 bits than not-device), and now they can. The fact that this hasn't bitten us speaks to the relative simplicity of our current ARMv8-M parts!

I don't have detailed measurements of the code, but this knocks 32 bytes off the routine, which is likely a performance win too.

### DIFF
--- a/sys/kern/src/arch/arm_m.rs
+++ b/sys/kern/src/arch/arm_m.rs
@@ -494,15 +494,22 @@ pub fn apply_memory_protection(task: &task::Task) {
 }
 
 /// ARMv8-M specific MPU accelerator data.
+///
+/// This is `repr(C)` only to make the field order match the register order in
+/// the hardware, which improves code generation. We do not actually rely on the
+/// in-memory representation of this struct otherwise.
 #[cfg(armv8m)]
 #[derive(Copy, Clone, Debug)]
+#[repr(C)]
 pub struct RegionDescExt {
-    /// This is the contents of the RLAR register, but without the enable bit
-    /// set. We write that first, and then set the enable bit.
-    rlar_disabled: u32,
-
+    /// Contents of the RBAR register.
     rbar: u32,
-    mair: u32,
+
+    /// Contents of the RLAR register.
+    rlar: u32,
+
+    /// This region's portion of the four-byte MAIR register.
+    mair: u8,
 }
 
 #[cfg(armv8m)]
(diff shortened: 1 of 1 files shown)


sys/kern/src/arch/arm_m.rs
```diff
@@ -571,30 +576,37 @@ pub fn apply_memory_protection(task: &task::Task) {
...
 
         let ext = &region.arch_data;
 
+        mairs[i] = ext.mair;
+
+        // Set the attridx field of the RLAR to just choose the attributes with
+        // the same index as the region. This lets us treat MAIR as an array
+        // corresponding to the regions.
+        //
+        // We unfortunately can't do this at compile time, because regions can
```
aapoalas, line 594: **thought:** This seems to call for a `TaskDescExt` kind of thing, possibly stored ouf-of-band in a separate array, which could then contain the MAIR0 and MAIR1 values precomputed.

Right now, the `mair` field for each region takes up 4 bytes (due to padding) while storing only a single byte of data. With an unrealistic assumption that each task only uses, on average, two unique regions that already means that the mair fields add 8 bytes of data to the flash. I assume something like 4-5 unique regions might be a more reasonable average, meaning 16-20 bytes are added for each task to, eventually, store only the 4-5 bytes of data. Moving this data to be task-specific would duplicate data but still lead to a reduction in flash size due to the removal of the padding.

sys/kern/src/arch/arm_m.rs
```diff
@@ -571,30 +576,37 @@ pub fn apply_memory_protection(task: &task::Task) {
...
             mpu.rbar.write(ext.rbar);
-            mpu.rlar.write(rlar_disabled | 1); // enable the region
+            mpu.rlar.write(rlar);
         }
     }
 
     unsafe {
+        // Load the MAIR registers.
+        mpu.mair[0].write(u32::from_le_bytes(mairs[..4].try_into().unwrap()));
+        mpu.mair[1].write(u32::from_le_bytes(mairs[4..].try_into().unwrap()));
```
hawkw, line 609: hm, the `try_into()` is a bit of a bummer here, since these are always going to be fixed-size arrays, and it would be nice to be able to reason about that statically and avoid checking that they actually are 4 bytes at runtime. but, i guess the alternative is to declare two 4-byte arrays and make the `for` loop above do some extra logic to reason about which one it's in, which also feels kind of sad. i dunno...

hawkw (commented):
This feels like it deserves a review from @labbott or someone else who knows the deep Cortex-M lore, but I was happy to give it a look anyway. Commented on a few things, but nothing major --- I'll leave that to someone who actually knows more about what these registers do :)

labbott (approved):
LGTM, thanks for fixing this

sys/kern/src/arch/arm_m.rs
cbiffle: I agree that this would be an interesting thing to try, and might speed things up / make things smaller.

sys/kern/src/arch/arm_m.rs
cbiffle: The original code (a couple of commits ago) did the for-loop-with-smarts thing, and it generated really bad code; using a single array here gets much better code.

The try_into to an `&[T; N]` succeeds statically, no checks or panics are generated. Breaking this would require nudging that 4 around. So while I don't love the _syntax_, and I'd really like to have a function that splits an array of 8 into two arrays of 4 statically, this is the best we've got in core rn.

sys/kern/src/arch/arm_m.rs
```diff
@@ -494,15 +494,22 @@ pub fn apply_memory_protection(task: &task::Task) {
...
-    /// This is the contents of the RLAR register, but without the enable bit
-    /// set. We write that first, and then set the enable bit.
-    rlar_disabled: u32,
-
     rbar: u32,
-    mair: u32,
+
+    /// This is the contents of the RLAR register with the enable bit set. We
+    /// can write it with the enable bit set directly, since we disable the MPU
+    /// before doing so.
```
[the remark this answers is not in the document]
cbiffle: Comments reworked.

PR #1964, opened 2024-12-26
Another batch of syscall perf enhancements

cbiffle:
This is unspooling more commits from my performance fix branch in my fork. As with the previous performance PR, this is a collection of semi-related commits that may be easier to review commit-by-commit than as a big diff. Boy I miss Gerrit.

This is only a ~10% overall improvement because most of the truly low hanging fruit has been eaten. But, this significantly improves the performance of kernel memory copies and SEND with an empty message (unusual but not unheard of). This also knocks a few hundred bytes off a typical kernel by removing some panic sites.

### DIFF
--- a/sys/abi/src/lib.rs
+++ b/sys/abi/src/lib.rs
@@ -220,7 +220,7 @@ pub const fn extract_new_generation(code: u32) -> Option<Generation> {
 pub const DEFECT: u32 = 1;
 
 /// State used to make scheduling decisions.
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
 pub enum TaskState {
     /// Task is healthy and can be scheduled subject to the `SchedState`
     /// requirements.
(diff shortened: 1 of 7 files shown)


sys/kern/src/umem.rs
```diff
@@ -168,6 +168,21 @@ impl<T> USlice<T> {
...
     }
+
+    /// Adjusts `a` and `b` to have the same length, which is the shorter of the
+    /// two.
+    ///
+    /// Returns the new common length.
+    ///
+    /// Shortening `a` and `b` ensures that the slices returned from `try_read`
+    /// / `try_write` have the same length, without the need for further
+    /// slicing.
```
hawkw, line 179: Neat! Love to see panic sites disappear!

sys/abi/src/lib.rs
```diff
@@ -252,7 +252,10 @@ impl TaskState {
     /// Checks if a task in this state is trying to deliver a message to
     /// `target`.
     pub fn is_sending_to(&self, target: TaskId) -> bool {
-        self == &TaskState::Healthy(SchedState::InSend(target))
+        match self {
+            TaskState::Healthy(SchedState::InSend(t)) => *t == target,
```
hawkw, line 256: Using matches instead of generating whole `PartialEq`/`Eq` implementations is a smart move when optimizing for binary size, i'll have to remember that.

hawkw (approved):
very nice! nothing too complicated in this one.

PR #1965, opened 2024-12-26
docs: add missing `software_irq` KIPC docs

hawkw:
I [recently discovered][1] that the `software_irq` KIPC call is missing
reference documentation. In fact, the KIPC documentation doesn't even
mention it (in contrast to other KIPC calls that have no documentation
but are still listed in the reference). Instead, the reference
documentation currently incorrectly states that `find_faulted_task` is
KIPC number 8, when it's actually KIPC 9 and `software_irq` is number 8.

Not adding this KIPC to the docs was an oversight on my part in PR #1661
--- my bad, sorry!

This commit adds documentation for the `software_irq` KIPC to the
reference.

[1]: https://github.com/oxidecomputer/hubris/pull/1961#discussion_r1898058287

### DIFF
--- a/doc/kipc.adoc
+++ b/doc/kipc.adoc
@@ -329,7 +329,9 @@ type SoftwareIrqResponse = ();
 
 This interface is primarily intended for testing interrupt handling, and can
 only be called by a task running as the supervisor. When this KIPC is called,
-an actual machine interrupt is triggered and dispatched by the kernel to the
+an actual machine interrupt is triggered for the relevant hardware interrupt
+source. The kernel then responds to the interrupt using its normal
+interrupt-handling infrastructure, which dispatches a notification to the 
 subscribed task.
 
 This KIPC is *not* intended for use as a general-purpose inter-task signalling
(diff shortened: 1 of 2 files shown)


hawkw:
Ugh, the Windows CI job is [yet again failing due to a tiny binary size difference that's totally unrelated to this change...](https://github.com/oxidecomputer/hubris/actions/runs/12507585005/job/34894347794) :upside_down_face:

doc/kipc.adoc
```diff
@@ -290,7 +290,53 @@ returned by a call to `get_task_dump_region` for the specified task.
...
+[source,rust]
+----
+type SoftwareIrqResponse = ();
+----
+
+==== Notes
+
+This interface is primarily intended for testing interrupt handling, and can
+only be called by a task running as the supervisor. When this KIPC is called,
+an actual machine interrupt is triggered and dispatched by the kernel to the
```
mkeeter, line 332: I would split this up to make it more obvious that the KIPC only triggers the interrupt, something like

> When this KIPC is called, an actual machine interrupt is triggered for the relevant hardware interrupt source.  The kernel then responds to the interrupt using its normal interrupt-handling infrastructure, which dispatches a notification to the subscribed task.

mkeeter (approved):

doc/kipc.adoc
hawkw: Thanks, I agree that's nicer.

PR #1977, opened 2025-01-16
Add IMX3112 mux and APML devices to Grapefruit

mkeeter:
# TODO
- [ ] Test on actual hardware (`standish`)

### DIFF
--- a/app/grapefruit/app.toml
+++ b/app/grapefruit/app.toml
@@ -403,6 +403,11 @@ sda.pin = 11
 scl.pin = 10
 af = 4
 
+[[config.i2c.controllers.ports.B.muxes]]
+driver = "imx3112"
+address = 0x70
+# U6 on AMD Ruby schematic
+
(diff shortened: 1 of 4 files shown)


mkeeter:
This is currently failing on real hardware.  We can talk to the mux and read back its version / ID / etc, but once we select mux channel 0 (`0x40`), we don't actually see anything talking back to us on that bus (other than the mux itself).

It's unclear whether this is a mux issue or whether our SP5 isn't configured for SB-TSI; @nathanaelhuffman is going to investigate the feasibility of soldering to SDA / SCL after the mux to narrow it down.

nathanaelhuffman:
Should we abandon this as overcome by events since we never got it working on ruby+grapefruit and there's not any real reason to push it now?

mkeeter:
Yeah, that makes sense to me – we've got our own hardware to fight with now!

PR #1981, opened 2025-01-16
QSFP: Pull in Quartz 0bc3093

Aaron-Hartwig:
This commit pulls in an updated QSFP FPGA which has an increased timeout on the module power supply power good signal assertion. Additionally, it removes internal pull-down resistors on those signals. See https://github.com/oxidecomputer/quartz/pull/261.


Aaron-Hartwig:
FYI - @labbott this is the commit I was talking about in chat.

nathanaelhuffman (approved):

PR #1982, opened 2025-01-17
Working eSPI for IPCC UART packets using oob message mechanism

nathanaelhuffman:
New FPGA, with FPGA-internal changes only to support IPCC over eSPI.  No interface or register map changes, boots fine on ruby with grapefruit!


Aaron-Hartwig (commented):
I know grapefruit is fast and loose, but I think there is value in pointing to where the collateral comes from ([example](https://github.com/oxidecomputer/hubris/blob/master/drv/sidecar-front-io/README.md)). Otherwise right now we don't really know the origin. At very least we should have the Quartz commit somewhere. Ideally it would be a commit off `main`, but I'm not going to die on that hill for a dev system.

Aaron-Hartwig (approved):
Thanks!

nathanaelhuffman:
> I know grapefruit is fast and loose, but I think there is value in pointing to where the collateral comes from ([example](https://github.com/oxidecomputer/hubris/blob/master/drv/sidecar-front-io/README.md)). Otherwise right now we don't really know the origin. At very least we should have the Quartz commit somewhere. Ideally it would be a commit off `main`, but I'm not going to die on that hill for a dev system.

Fixed with a link to the commit on quartz main, image was built off that commit by ci.

PR #1988, opened 2025-01-22
Disable periodic IPCC zero bytes on Grapefruit

citrus-it:
In the current IPCC design, the SP sends zero bytes periodically when the channel is idle to guard against the possibility of losing a frame terminator. On gimlet this causes the UART FIFO to quickly fill up and so when the channel wakes up there are around 16 bytes to drain and things move along. On grapefruit/cosmo, however, there is a 4K FIFO in the FPGA upstream of the SP here so the natural saturation doesn't (yet) occur as we'd like. The extra zero bytes also make tracing the bus harder. Disable them for now.

### DIFF
--- a/task/host-sp-comms/src/tx_buf.rs
+++ b/task/host-sp-comms/src/tx_buf.rs
@@ -139,8 +139,14 @@ impl TxBuf {
     }
 
     /// Should we be sending periodic 0 bytes?
+    #[cfg(feature = "gimlet")]
     pub(crate) fn should_send_periodic_zero_bytes(&self) -> bool {
         matches!(self.state, State::Idle)
+    }
+
+    #[cfg(feature = "grapefruit")]
+    pub(crate) fn should_send_periodic_zero_bytes(&self) -> bool {
+        false
     }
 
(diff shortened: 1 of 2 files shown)


task/host-sp-comms/src/tx_buf.rs
```diff
@@ -139,10 +139,16 @@ impl TxBuf {
     }
 
     /// Should we be sending periodic 0 bytes?
+    #[cfg(feature = "gimlet")]
     pub(crate) fn should_send_periodic_zero_bytes(&self) -> bool {
         matches!(self.state, State::Idle)
     }
 
+    #[cfg(not(feature = "gimlet"))]
+    pub(crate) fn should_send_periodic_zero_bytes(&self) -> bool {
```
mkeeter, line 148: Nit: maybe `feature = "grapefruit"` instead, so that we notice when we add a new target?

(e.g. after adding Cosmo support, it will become `any(feature = "grapefruit", feature = "cosmo")`)

mkeeter (approved):

PR #1990, opened 2025-01-22
Cleanup to `grapefruit-seq` in preparation for Cosmo

mkeeter:
Three relatively standalone commits:

- Add a helper function `AuxFlash::get_compressed_blob_streaming` to do streaming decompression of a blob in auxflash.  We'll use this for both FPGAs on Cosmo.
- Move Spartan7 initialization to a separate crate, since we need to do it on both Grapefruit and Cosmo.  This is mostly modeled after `drv-ice40-spi-program`.
- Remove `fmc` peripheral from the `fmc_demo` task, because it's now initialized pre-Hubris.


drv/auxflash-api/src/lib.rs
```diff
@@ -171,6 +171,55 @@ where
...
+impl AuxFlash {
+    /// Reads a compressed blob, streaming it to a callback function
+    pub fn get_compressed_blob_streaming<
+        F: Fn(&[u8]) -> Result<(), E>,
+        E: From<AuxFlashError>,
+    >(
+        &self,
+        tag: [u8; 4],
+        f: F,
+    ) -> Result<[u8; 32], E> {
```
hawkw, line 184: suuuuper nitpicky (sorry), but this is enough generic bounds that i would strongly recommend putting them in a `where` clause:
```suggestion
    pub fn get_compressed_blob_streaming<F, E>(
        &self,
        [tag: u8; 4],
        f: F,
    ) -> Result<u32, E>
    where 
        F: Fn(&[u8]) -> Result<(), E>,
        E: From<AuxFlashError>,
    {
```

personally, i find that somewhat more readable; it took me a second to figure out this wasn't a function returning a weird tuple or something...

drv/auxflash-api/src/lib.rs
```diff
@@ -171,6 +171,55 @@ where
...
+            let mut decompress_buffer = [0; 512];
+
+            while !chunk.is_empty() {
+                let decompressed_chunk = gnarle::decompress(
+                    &mut decompressor,
+                    &mut chunk,
+                    &mut decompress_buffer,
+                );
+
+                // The compressor may have encountered a partial run at the
```
hawkw, line 209: nit: shouldn't this say
```suggestion
                // The decompressor may have encountered a partial run at the
```

drv/spartan7-spi-program/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+//!    / `CONFIG_DONE` pins yourself
+//! 3. Call `begin_bitstream_load` once.
+//! 4. Call `continue_bitstream_load` as many times as you need to, passing
+//!    chunks of data each time.
+//! 5. Call `finish_bitstream_load` once to complete the process and check the
+//!    result.
+//!
+//! If any of the operations fail, the intention is that you restart the process
+//! from `begin_bitstream_load` -- it should handle the reset and clean up from
+//! the earlier failure. However, this is only _somewhat_ tested.
```
hawkw, line 25: I'll be the first to say that I don't really know if this suggestion is worth the effort, but it occurs to me that we *could* perhaps have an API surface that ensures you do this properly. What if `begin_bitstream_load` returned some kind of `BitstreamLoader` struct, and `continue_bitstream_load` and `finish_bitstream_load` were methods like:
```rust
pub struct BitstreamLoader<'a, S> {
    spi: &'a SpiDevice<S>,
    sys: &'a Sys,
    config: &'a Config,
}

impl BitstreamLoader<'_, S> {
    pub fn begin_bitstream_load(
        sys: &Sys,
        config: &Config,
        spi: &SpiDevice<S>,
    ) -> Result<Self, Spartan7Error> {
        // ...
    }

     pub fn continue_bitstream_load(self, data: &[u8])-> Result<Self, Spartan7Error>
     where S: SpiServer, {
          // ...
     }

     pub fn finish_bitstream_load(self) -> Result<(), Spartan7Error>
     where S: SpiServer {
         // ...
     }
}
```

Because the `continue_bitstream_load` method here takes the `BitstreamLoader` by value and returns it only on `Ok`, it ensures that the user of the API *must* restart the process from `begin_bitstream_load` on error, and prevents the user from calling `continue_bitstream_load` unless they've called `begun_bitstream_load`. Similarly, `finish_bitstream_load` consumes the loader struct and therefore stops you from continuing a bitstream load unless you start over again. This also has the nice side benefit of ensuring that the same `SpiDevice` is used for every chunk of bitstream, although I'll freely admit that code calling this is unlikely to have enough random `SpiDevice`s flying around for that to be a common error.

I doubt we'll be using this module frequently enough to justify spending a lot of time on ensuring it has a misuse-resistant API --- we're basically going to use it in the grapefruit and cosmo initialization code, and basically nowhere else. But, I think this change is actually not all that much work, so it might be worth it even though it's not actually that important. Just a thought.

drv/spartan7-spi-program/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut wait_time_ms = 0;
+    while sys.gpio_read(config.config_done) == 0 {
+        hl::sleep_for(DELAY_MS);
+        wait_time_ms += DELAY_MS;
+        if wait_time_ms > TIMEOUT_MS {
+            return Err(Spartan7Error::ConfigDidNotComplete);
+        }
+    }
+
+    // Send 64 bonus clocks to complete the startup sequence (see "Clocking
```
hawkw, line 154: I love the naming "bonus clocks" :D

hawkw (approved):
This is pretty straightforward and seems good to me! I had one suggestion about the API for the `spartan7-spi-program` crate, which could make things a bit nicer but is also completely ignorable. :)

drv/auxflash-api/src/lib.rs
mkeeter: thanks, fixed

drv/auxflash-api/src/lib.rs
mkeeter: Reasonable, fixed!

mkeeter:
I applied @hawkw 's comments, here's the [inter-diff](https://github.com/oxidecomputer/hubris/compare/dfd0d5e000634a8c8ca88758b9db0fa521b6ac76..207302e230b258f576515da6aa615b5ebb7d53e1)

drv/spartan7-spi-program/src/lib.rs
mkeeter: Seems reasonable, done

hawkw:
> I applied @hawkw 's comments, here's the [inter-diff](https://github.com/oxidecomputer/hubris/compare/dfd0d5e000634a8c8ca88758b9db0fa521b6ac76..207302e230b258f576515da6aa615b5ebb7d53e1)

Lovely!

Issue #1995, opened 2025-01-29
Sidecar Front I/O QSFP LEDs stay on while system is in A2


rmustacc:
@leftwo was working on a Sidecar and had commanded the system go to A2. While it did successfully do so, it left all the front I/O LEDs in the last state that they were in when dendrite had operated it. Given that the switch zone and dendrite only generally expect to operate while Sidecar is in A0 (though this may be something we need to consider and change), we likely should be defaulting to turning off LEDs when in A2.

This definitely requires some deeper thought about how we want the service and control points to operate while we're doing troubleshooting and in-field service, but given that this can happen independent of dendrite existing, they likely shouldn't remain on. I expect the flow of control and the ordering of different requests via the broader control plane and dendrite is something that will still want to continue to evolve, but at the moment what we have for behavior is rather confusing.

PR #1998, opened 2025-01-31
transceivers-server: keep LEDs off while not in A0

Aaron-Hartwig:
While Sidecar is not in A0 the transceivers-server will now keep the LEDs off. The current behavior is that the LEDs only change state when Dendrite tells them to, and while the Sidecar is not in Dendrite isn't doing much at all! An A0->A2 transition left the LEDs in their current state which is confusing for folks who may have eyes on the switch as it would indicate Dendrite is running when it is in fact not. The system attention LED is unaffected by this commit.

fixes https://github.com/oxidecomputer/hubris/issues/1995

### DIFF
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -78,6 +78,7 @@ enum Trace {
     DisablingPorts(LogicalPortMask),
     DisableFailed(usize, LogicalPortMask),
     ClearDisabledPorts(LogicalPortMask),
+    SeqError(SeqError),
 }
 
 counted_ringbuf!(Trace, 16, Trace::None);
@@ -194,12 +195,12 @@ impl ServerImpl {
         self.system_led_state
     }
 
-    fn update_leds(&mut self) {
+    fn update_leds(&mut self, seq_state: TofinoSeqState) {
         let mut next_state = LogicalPortMask(0);
 
-        // We only turn transceiver LEDs on when Sidecar is in A0 since that is when there can be
-        // meaningful link activity happening. When outside of A0 we default the LEDs to off.
-        if self.in_a0 {
+        // We only turn transceiver LEDs on when Sidecar is in A0, since that is when there can be
+        // meaningful link activity happening. When outside of A0, we default the LEDs to off.
+        if seq_state == TofinoSeqState::A0 {
             for (i, state) in self.led_states.0.into_iter().enumerate() {
                 let i = LogicalPort(i as u8);
                 match state {
@@ -518,9 +519,9 @@ impl ServerImpl {
         // the `sensors` and `thermal` tasks.
     }
 
-    fn handle_i2c_loop(&mut self) {
+    fn handle_i2c_loop(&mut self, seq_state: TofinoSeqState) {
         if self.leds_initialized {
-            self.update_leds();
+            self.update_leds(seq_state);
             let errors = match self.leds.error_summary() {
                 Ok(errs) => errs,
                 Err(e) => {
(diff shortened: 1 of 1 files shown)


Aaron-Hartwig:
I tested this.

Leds on in A0:
```
aaron@niles ~ $ pfexec humility -t sidecar-b-aaron hiffy -c Sequencer.tofino_seq_state
Sequencer.tofino_seq_state() => A0
```
ima_6d452d6

Set sytem to A2 and observe LEDs go off:
```
aaron@niles ~ $ pfexec humility -t sidecar-b-aaron hiffy -c Sequencer.set_tofino_seq_policy -a policy=Disabled
Sequencer.set_tofino_seq_policy() => ()
aaron@niles ~ $ pfexec humility -t sidecar-b-aaron hiffy -c Sequencer.tofino_seq_state
Sequencer.tofino_seq_state() => A2
```
ima_233142a

Aaaand they come back on in A0 when Dendrite is running again:
```
aaron@niles ~ $ pfexec humility -t sidecar-b-aaron hiffy -c Sequencer.set_tofino_seq_policy -a policy=LatchOffOnFault
Sequencer.set_tofino_seq_policy() => ()
aaron@niles ~ $ pfexec humility -t sidecar-b-aaron hiffy -c Sequencer.tofino_seq_state
Sequencer.tofino_seq_state() => A0
```
ima_ebdd51d

drv/transceivers-server/src/main.rs
```diff
@@ -194,20 +195,25 @@ impl ServerImpl {
...
-        for (i, state) in self.led_states.0.into_iter().enumerate() {
-            let i = LogicalPort(i as u8);
-            match state {
-                LedState::On => next_state.set(i),
-                LedState::Blink => {
-                    if self.blink_on {
-                        next_state.set(i)
+
+        // We only turn transceiver LEDs on when Sidecar is in A0 since that is when there can be
+        // meaningful link activity happening. When outside of A0 we default the LEDs to off.
```
hawkw, line 201: nitpicky
```suggestion
        // We only turn transceiver LEDs on when Sidecar is in A0, since that is when there can be
        // meaningful link activity happening. When outside of A0, we default the LEDs to off.
```

drv/transceivers-server/src/main.rs
```diff
@@ -715,6 +722,12 @@ fn main() -> ! {
             }
         }
 
+        server.in_a0 = match seq.tofino_seq_state() {
+            Ok(TofinoSeqState::A0) => true,
+            Ok(_) => false,
+            Err(_) => false,
+        };
```
hawkw, line 729: If I understand correctly, we only need to check the sequencer state on timer ticks that call `handle_i2c_loop` (which then calls `update_leds`). But, here we're performing the IPC to the sequencer task to get the power state on _every_ iteration of the loop, even on ones where we won't be messing with the LEDs.

What do you think about changing this so that, instead of having an `in_a0` field on `ServerImpl`, we instead passed the `TofinoSeqState` into `handle_i2c_loop` as an argument, which in turn passed it to `update_leds`? That way, we would only call the sequencer's `tofino_seq_state` IPC when we actually need to know what power state we're in, and the flow of that information to the place where it's actually consumed would be a bit more obvious.

hawkw (commented):
Some drive-by suggestions; I hope they're useful.

drv/transceivers-server/src/main.rs
Aaron-Hartwig: Your point about getting the IPC out of the main loop and into the slower loop is well taken! I've also reworked things to just pass the current sequencer state into the I2C loop as you suggest.

Aaron-Hartwig:
@hawkw I always welcome your feedback! Drive-by or solicited 😄

bnaecker (approved):

PR #2000, opened 2025-02-04
add LM5066 support

bcantrill:
Draft comment:

```
This adds support for the LM5066 hotswap conntroller. Note that this
also adds a device for Gimletlet boards to allow for the code to be
tested, but Gimletlets will operate correctly without the device
physically connected.
```


drv/i2c-devices/src/adm1272.rs
```diff
@@ -85,7 +85,7 @@ enum Trace {
     None,
 }
 
-ringbuf!(Trace, 32, Trace::None);
+ringbuf!(Trace, 8, Trace::None);
```
mkeeter, line 88: Is this change intentional?

drv/i2c-devices/src/lm5066.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+}
+
+impl From<Error> for ResponseCode {
+    fn from(err: Error) -> Self {
+        match err {
+            Error::BadRead { code, .. } => code,
+            Error::BadWrite { code, .. } => code,
+            Error::BadValidation { code, .. } => code,
+            _ => panic!(),
```
mkeeter, line 50: Without more context, this seems like it's leaving a footgun lying around.  Maybe implement `TryFrom<Error> for ResponseCode` instead, so that the caller has to unwrap it?

drv/i2c-devices/src/lm5066.rs
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        let device_setup = pmbus_read!(self.device, lm5066::DEVICE_SETUP)?;
+        ringbuf_entry!(Trace::DeviceSetup(device_setup));
+        self.device_setup.set(Some(device_setup));
+
+        Ok(device_setup)
+    }
+
+    //
```
mkeeter, line 128: Nit: `///` so that this is recognized as a docstring

drv/i2c-devices/src/adm1272.rs
bcantrill: Yeah -- that ring buffer was **way** oversized -- and I want to pay for a similar ring buffer in the lm5066 code. I have sized them both at a much more rational size, so it's net-minus on RAM usage.

drv/i2c-devices/src/lm5066.rs
bcantrill: For better or for ill, this is a pattern we have in a bunch of drivers. I would like to fix all of them in the same way; happy to do it as part of this work (indeed, now may be the time for it?) but it will increase the scope quite a bit. Thoughts?

drv/i2c-devices/src/lm5066.rs
bcantrill: Does that make sense for a non-public function? (I have always used docstrings for stuff that I want to show up in rustdoc.)

drv/i2c-devices/src/lm5066.rs
mkeeter: Yeah, it looks like this is only used in `validate`, so we could be safer about it.

mkeeter: On further inspection, this is a preexisting condition and is not covered by my code review insurance, so I'm fine leaving the footgun lying around (until we rewrite the whole I2C driver).

drv/i2c-devices/src/lm5066.rs
mkeeter: I find the distinction meaningful because my editor highlights them differently, so it's easier to spot the beginnings of functions:

Screenshot 2025-02-04 at 4 12 55 PM

drv/i2c-devices/src/lm5066.rs
bcantrill: Here are all of the errors that currently result in a `panic!` when coerced into a `ResponseCode`:

```
adm1272::Error::InvalidData
adm1272::Error::InvalidConfig
adt7420::Error::BadID
at24csw080::Error::InvalidAddress
at24csw080::Error::InvalidEndAddress
at24csw080::Error::InvalidObjectSize
at24csw080::Error::MisalignedPage
at24csw080::Error::InvalidPageSize
at24csw080::Error::InvalidSecurityRegisterReadByte
at24csw080::Error::InvalidSecurityRegisterWriteByte
bmr491::Error::InvalidData
isl68224::Error::InvalidData
lm5066::Error::InvalidData
lm5066::Error::InvalidDeviceSetup
mwocp68::Error::BadData
mwocp68::Error::InvalidData
mwocp68::Error::BadFirmwareRev
mwocp68::Error::BadFirmwareRevLength
mwocp68::Error::UpdateInBootLoader
mwocp68::Error::UpdateAlreadySuccessful
mwocp68::Error::BadBootLoaderStatus
mwocp68::Error::BadBootLoaderCommand
mwocp68::Error::ChecksumNotSuccessful
raa229618::Error::BadData
raa229618::Error::InvalidData
sbrmi::Error::BadThreadId
sbrmi::Error::BadCpuidInput
sbrmi::Error::BadCpuidLength
sbrmi::Error::CpuidFailed
sbrmi::Error::BadRdmsrLength
sbrmi::Error::BadRdmsr   
tps56b24a::Error::BadData
tps56b24a::Error::InvalidData
```

Of these, one is just wrong (`sbrmi::Error::BadRdmsr`), many are due to bad/unexpected device state, and the balance are due to programmer error. I think my bias would be to do as `nvme_bmc` does and convert the bad device state into a `ResponseCode::BadDeviceState`, and then leave the ones that are programmer error as a `panic!`? (Note that this will clean up the new errors introduced here, which fall into the category of bad device state.)  Thoughts on that?

mkeeter: This makes sense to me!

app/gimletlet/app.toml
```diff
@@ -259,3 +259,21 @@ pins = [
...
+sensors = { voltage = 1, current = 1, temperature = 1 }
+
+[tasks.power]
+name = "task-power"
+priority = 6
+max-sizes = {flash = 65536, ram = 16384 }
+stacksize = 3800
+start = true
+task-slots = ["i2c_driver", "sensor"]
+notifications = ["timer", "external_badness"]
```
mkeeter, line 278: ... `"external_badness"`?

bcantrill: Stowaway turd in the Gimlet app definition from 4f29df05b455490636ed67a87940ab77a518986f and then copy-pasta'd here; removed both.

PR #2001, opened 2025-02-05
Add RAA229620A rail driver

mkeeter:
Another standalone building block for Cosmo.

The RAA229620A is identical to the RAA229618 (for our purposes), other than its device ID code.

### DIFF
--- a/drv/i2c-devices/src/raa229620a.rs
+++ b/drv/i2c-devices/src/raa229620a.rs
@@ -1,0 +1,169 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use core::cell::Cell;
+
+use crate::{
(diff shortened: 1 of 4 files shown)


drv/i2c-devices/src/raa229620a.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    pub fn turn_on(&mut self) -> Result<(), Error> {
+        let mut op = pmbus_rail_read!(self.device, self.rail, OPERATION)?;
+        op.set_on_off_state(OPERATION::OnOffState::On);
+        pmbus_rail_write!(self.device, self.rail, OPERATION, op)
+    }
+
+    pub fn set_vout(&mut self, value: Volts) -> Result<(), Error> {
+        if value > Volts(3.050) {
```
labbott, line 101: This is intended to be `3.050` and not say `3.500`

labbott (approved):
LGTM modulo one double check that something isn't a typo

drv/i2c-devices/src/raa229620a.rs
mkeeter: yes, thanks for checking!

PR #2002, opened 2025-02-05
ereport storage data structure

cbiffle:


lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+            })
+        })
+    }
+
+    /// Reads records starting from `first_ena`, inclusive.
+    pub fn read_from(
+        &mut self,
+        first_ena: u64,
+    ) -> impl Iterator<Item = Record<'_>> + '_ {
+        // Attempt to recover and insert a loss record if necessary.
```
mkeeter, line 132: Why does this function also attempt to recover, versus `iter_contents` which is immutable?

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+
+            let size = OVERHEAD + len;
+            // So it's weird, but, heapless::Deque only lets you pop individual
+            // bytes. Hopefully this is not too expensive.
+            for _ in 0..size {
+                self.storage.pop_front();
+            }
+
+            self.earliest_ena += 1;
+        }
```
mkeeter, line 167: Would it make sense to call `recover_if_required` here (or even inline its contents), instead of in `insert_impl`?  This appears to be the only point where we become able to recover.

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// If, after recovery, we are `Collecting` and there is enough space for a
+    /// header (`OVERHEAD`) plus `data`, the record is stored, implicitly
+    /// assigned the next ENA.
+    ///
+    /// If recovery is not possible, the existing loss count is advanced.
+    ///
+    /// If we are `Collecting` but `data` plus a header won't fit, we enter the
+    /// `Losing` state with a count of 1.
+    fn insert_impl(&mut self, sender: u16, timestamp: u64, data: &[u8]) {
+        let data_len = u16::try_from(data.len()).unwrap_lite();
```
mkeeter, line 184: This is a caller-controlled panic; should we be inserting a loss record instead?

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+                }
+            }
+            InsertState::Losing { count, .. } => {
+                *count = count.saturating_add(1);
+            }
+        }
+    }
+
+    /// Internal utility routine for storing a record header.
+    fn write_header(&mut self, task: u16, timestamp: u64, data_len: u16) {
```
mkeeter, line 209: Nit: is there a reason arguments here are in a different order than the header?

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+
+impl Record<'_> {
+    /// Iterates over the body bytes, hiding the fact that they be
+    /// discontiguous.
+    pub fn body_bytes(&self) -> impl Iterator<Item = u8> + '_ {
+        self.slices
+            .0
+            .iter()
+            .copied()
+            .chain(self.slices.1.iter().copied())
```
mkeeter, line 296: ⛳ 
```suggestion
        self.slices
            .0
            .iter()
            .chain(self.slices.1.iter())
            .copied()
```

lib/snitch-core/src/lib.rs
cbiffle: I actually switched the behavior of `iter_contents` after writing tests against it, it turns out to be rather annoying to have it not attempt recovery.

lib/snitch-core/src/lib.rs
cbiffle: Hm. So my initial thinking was that this is an explicit precondition that could be checked by the caller (packrat, currently), so a panic made sense. But I think you have a point. I'll try changing it and see how the client code feels.

lib/snitch-core/src/lib.rs
cbiffle: Path dependence. ;-) This was actually the original order of fields in the header, until it became clear that len-first made the flush impl simpler. I'll swap 'em.

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+
+        for _ in 0..=index {
+            // Discard the lead record in the queue.
+            let mut slices = self.storage.as_slices();
+            let len = u16::from_le_bytes(take_array(&mut slices).unwrap_lite())
+                as usize;
+
+            let size = OVERHEAD + len;
+            // So it's weird, but, heapless::Deque only lets you pop individual
+            // bytes. Hopefully this is not too expensive.
```
hawkw, line 194: looks like it does [this](https://docs.rs/heapless/0.8.0/src/heapless/deque.rs.html#280-287)...

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+        }
+        for byte in timestamp.to_le_bytes() {
+            self.storage.push_back(byte).unwrap_lite();
+        }
+        self.stored_record_count += 1;
+    }
+
+    /// Checks if we're losing data and attempts to stop, by generating a loss
+    /// record in the queue.
+    fn recover_if_required(&mut self, space_required: Option<usize>) -> bool {
```
hawkw, line 258: This method returns a `bool` but, as far as I can tell, no callers of it actually consume that `bool`. Is this vestigial, or are you expecting to need it later?

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+    n: usize,
+) -> Option<(&'a [u8], &'a [u8])> {
+    let first_n = usize::min(slices.0.len(), n);
+    let (first, rest) = slices.0.split_at(first_n);
+    slices.0 = rest;
+
+    let rest_n = n - first_n;
+    if slices.1.len() < rest_n {
+        return None;
+    }
```
hawkw, line 354: So, I'm not sure if this is an issue in practice, but it occurs to me that, in the case where `slices.1.len() < rest_n`, we will have mutated `slices.0` but return `None`, which potentially leaves the two slices in a weird-ish state. Should we instead consider performing the check that there are enough bytes to take _before_ we mutate either slice, as in:
```suggestion
    let rest_n = n - first_n;
    if slices.1.len() < rest_n {
        return None;
    }
    
    let (first, rest) = slices.0.split_at(first_n);
    slices.0 = rest;
```

lib/snitch-core/src/lib.rs
cbiffle: So, this at least needs more comments, because it took me a bit to figure out what I intended.

There is a reason why recovery is done at insert, instead of here, though I need to re-convince myself that it's a _good_ reason. It's to try and prevent a series of loss records, in the case where

- we try to insert something big, and enter losing state with a count of 1.
- Something gets flushed, which opens up enough space for the _loss record_ but not any useful-sized data. If we recovered during flush, we'd emit a loss record here.
- we try to insert something big again, re-entering losing state with a count of 1.
- this generates another loss record
- and so forth

By recovering at insert, and in particular by considering the inserted size plus the recovery requirement when deciding to recover, we can avoid generating sequences of loss records in this corner case.

Does that make sense?

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+    /// contains the ENA that _will be_ assigned to the next pushed record.
+    earliest_ena: u64,
+
+    /// Current state of the record insertion state machine. See the enum for
+    /// details.
+    insert_state: InsertState,
+
+    /// To keep this crate platform-independent (for testing) and to keep this
+    /// type in bss, we expect the task that creates a `Store` to provide its
+    /// task ID to `initialize`; it then gets stored here.
```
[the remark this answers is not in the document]
cbiffle: Done!

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+    /// current system `timestamp`, to be used in generating loss records.
+    pub fn initialize(&mut self, tid: u16, timestamp: u64) {
+        self.our_task_id = tid;
+        // If the queue has never been touched, insert our "arbitrary data loss"
+        // record into the stream to consume ENA 0.
+        if self.earliest_ena == 0 && self.storage.is_empty() {
+            self.insert_impl(
+                self.our_task_id,
+                timestamp,
+                &[0xA1, 0x64, 0x6C, 0x6F, 0x73, 0x74, 0xF6],
```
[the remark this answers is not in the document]
cbiffle: Yeah, I'm kinda hoping to replace both of them with uses of the `cbor!` macro, but... it's starting to look like I can't, because it doesn't allow this sort of partial record / template creation. So, more comments are needed.

lib/snitch-core/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+    /// Sets up the `Store` with data from the environment. Must be called once
+    /// before other functions.
+    ///
+    /// This stores the bits of the current task's `TaskId` (`tid`) and the
+    /// current system `timestamp`, to be used in generating loss records.
+    pub fn initialize(&mut self, tid: u16, timestamp: u64) {
+        self.our_task_id = tid;
+        // If the queue has never been touched, insert our "arbitrary data loss"
+        // record into the stream to consume ENA 0.
+        if self.earliest_ena == 0 && self.storage.is_empty() {
```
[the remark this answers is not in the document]
cbiffle: lol, yes, it is. Fixed!

lib/snitch-core/src/lib.rs
cbiffle: I keep expecting to need it, and yet I haven't... I will flag it for reconsideration.

lib/snitch-core/src/lib.rs
cbiffle: I think this is a great point -- while it doesn't make a high level difference in the behavior right now, I think it's important to get the semantics of the utility functions right, or they'll break something subtle during maintenance later.

lib/snitch-core/src/lib.rs
cbiffle: Alright, I've switched this logic to automatically "lose" any such packets, not that they will exist in practice.

cbiffle:
Went ahead and marked this as non-draft since we're through like three rounds of reviews.

lib/snitch-core/src/lib.rs
cbiffle: Yeah, the question is what the code looks like when called in a loop, in practice, on our target. I haven't evaluated that yet. But it'll probably be fine.

lib/snitch-core/src/lib.rs
cbiffle: I've added a trio of comments attempting to express this rationale in the source code for posterity.

mkeeter: This explanation + the new comments make sense to me!

hawkw: I like the new comments, thanks for that!

lib/snitch-core/src/lib.rs
hawkw: Yup, that was my thinking as well, it just felt like a sharp edge that we weren't *currently* grabbing but might as well sand down in case we reach for it later. Thanks for fixing it!

hawkw (approved):
I took another look at this, focusing on the tests, and I feel quite good about it! It could be worthwhile to consider some kind of "generate a bunch of random operations" style property testing, but...honestly, we don't exactly have a whole bunch of important invariants to assert in such a test, so constructing a double of the expected behavior might not be worth the effort. You've certainly covered all the cases I can immediately anticipate being important in practice based on the rest of the ereport ingestion design. This looks great overall!

PR #2003, opened 2025-02-05
Standalone spartan-7 loader

mkeeter:
For Grapefruit and Cosmo, the Spartan-7 FPGA is used as a memory-mapped peripheral once its bitstream is loaded.  It's important that it not die or be reset, because that will be Very Confusing for any tasks writing directly into its memory space.

This PR moves loading that FPGA into a standalone task, separate from `grapefruit-seq` (and the eventual `cosmo-seq`).  The new task is infallible once it enters the Idol runtime loop, and can provide unforgeable tokens that prove the FPGA was loaded; eventually, I'd like for FPGA-based peripherals to take those tokens in their constructors.

The code changes in this PR are mostly reorganization and boilerplate, pulling stuff from `grapefruit-seq` into `drv-spartan7-loader` and writing `drv-spartan7-loader-api` / Idol files.

Tested on Grapefruit (standalone).

### DIFF
--- a/drv/grapefruit-seq-server/src/main.rs
+++ b/drv/grapefruit-seq-server/src/main.rs
@@ -7,31 +7,21 @@
 #![no_std]
 #![no_main]
 
-use core::num::NonZeroUsize;
 use drv_cpu_seq_api::{PowerState, StateChangeReason};
-use drv_spartan7_spi_program::{BitstreamLoader, Spartan7Error};
-use drv_spi_api::{SpiDevice, SpiServer};
+use drv_spartan7_loader_api::Spartan7Loader;
 use drv_stm32xx_sys_api as sys_api;
 use idol_runtime::{NotificationHandler, RequestError};
 use task_jefe_api::Jefe;
 use task_packrat_api::{CacheSetError, MacAddressBlock, Packrat, VpdIdentity};
-use userlib::{
-    hl, sys_recv_notification, task_slot, FromPrimitive, RecvMessage,
-    UnwrapLite,
-};
+use userlib::{hl, task_slot, FromPrimitive, RecvMessage, UnwrapLite};
 
 use ringbuf::{counted_ringbuf, ringbuf_entry, Count};
 
 task_slot!(JEFE, jefe);
+task_slot!(LOADER, spartan7_loader);
 
 #[derive(Copy, Clone, PartialEq, Count)]
 enum Trace {
-    FpgaInit,
-    FpgaInitFailed(#[count(children)] Spartan7Error),
-    StartFailed(#[count(children)] SeqError),
-    ContinueBitstreamLoad(usize),
-    WaitForDone,
-    Programmed,
     MacsAlreadySet(MacAddressBlock),
     IdentityAlreadySet(VpdIdentity),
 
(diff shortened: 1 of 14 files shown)


drv/spartan7-loader/src/main.rs
```diff
@@ -0,0 +1,230 @@
...
+    ContinueBitstreamLoad(usize),
+    WaitForDone,
+    Programmed,
+
+    #[count(skip)]
+    None,
+}
+
+#[derive(Copy, Clone, PartialEq, Count)]
+enum SeqError {
```
hawkw, line 62: nitpick, feel free to ignore me: I know this was just copied from `grapefruit-sesq-server`, but...does it really make sense to call this `SeqError` now that it's in the `spartan7-loader` driver? Should it instead be called `LoaderError` or something now?

drv/spartan7-loader/src/main.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+}
+
+fn init() -> Result<(), SeqError> {
+    let sys = sys_api::Sys::from(SYS.get_task_id());
+    let dev = claim_spi(&sys).device(drv_spi_api::devices::SPARTAN7_FPGA);
+    let aux = drv_auxflash_api::AuxFlash::from(AUXFLASH.get_task_id());
+
+    init_spartan7_fpga(&sys, &dev, &aux)?;
```
hawkw, line 126: hm, i don't totally get why `init_spartan7_fpga` is a separate function, since (AFAIK) it only gets called here, and we don't use it as a separate scope for the `?` operator or similar.  is there any reason for factoring it out into a separate function i'm missing?

drv/spartan7-loader-api/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+//! API crate for the Spartan7 loader.
+
+#![no_std]
+
+use userlib::sys_send;
+
+/// Token that indicates that the Spartan-7 is running
+///
+/// This token can be passed to peripheral constructors.
+pub struct Spartan7Token(core::marker::PhantomData<()>);
```
hawkw, line 14: nitpicky, feel free to ignore me: i don't actually think `PhantomData` is necessary here, this could just be
```suggestion
pub struct Spartan7Token(());
```
as [they're both zero-sized](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=116058fa412d4f13c41b6a3c1cc4750a).

idl/spartan7-loader.idol
```diff
@@ -0,0 +1,13 @@
+// Interface to Spartan-7 FPGA loader.
+
+Interface(
+    name: "Spartan7Loader",
+    ops: {
+        "ping": (
```
hawkw, line 6: take it or leave it: i might have named this something more descriptive, like "wait_for_completion" or something...?

drv/spartan7-loader/src/main.rs
mkeeter: Historical artifact; fixed in 1572e138

drv/spartan7-loader-api/src/lib.rs
mkeeter: Nice, done in 01c2dd1a

drv/spartan7-loader/src/main.rs
mkeeter: Sure, done in 73b9724b

idl/spartan7-loader.idol
mkeeter: I'd prefer to stick with `ping` – the operation is so fundamentally basic that I don't want the op name to imply any complexity.

hawkw: reasonable, carry on!

hawkw (approved):
Looks good to me! I dunno if you want to wait for another review, but I think this is pretty uncontroversial...

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -96,59 +63,23 @@ fn main() -> ! {
...
-    fn init(
-        sys: &sys_api::Sys,
-        spi: S,
-        aux: drv_auxflash_api::AuxFlash,
-    ) -> Result<Self, SeqError> {
+impl ServerImpl {
+    fn init(sys: &sys_api::Sys) -> Self {
         // Ensure the SP fault pin is configured as an open-drain output, and
         // pull it low to make the sequencer restart externally visible.
+        const FAULT_PIN_L: sys_api::PinSet = sys_api::Port::A.pin(15);
```
lzrd, line 82: Pins can be named in  app.toml files which helps port tasks to different hardware. e.g. app/gimletlet/app.toml
```
[config.sprot]
pins = [
    # ROT_IRQ (af=15 for EXTI)
    { name = "ROT_IRQ", pin = { port = "D", pin = 0, af = 15, direction = "input"}},
    # SPI6 CS repurposed for debugging
    { name = "DEBUG", pin = { port = "G", pin = 8, af = 0, direction = "output"}}
]
```

lzrd: I see that the only use of this on STM32 is in my code. I still encourage use of the feature but changing it could wait for some cleanup round.

lzrd (approved):

drv/grapefruit-seq-server/src/main.rs
mkeeter: Switched to `task_config!` in ec67ba343e689f8a535fb166084247f6be2d83a8

PR #2006, opened 2025-02-07
Add APOB messages to host_sp_comms

mkeeter:
See https://github.com/oxidecomputer/stlouis/issues/707, https://github.com/oxidecomputer/rfd/pull/855 , https://github.com/oxidecomputer/amd-host-image-builder/pull/222


task/host-sp-comms/src/main.rs
```diff
@@ -995,6 +1034,51 @@ impl ServerImpl {
...
+    fn apob_write_page(
+        hf: &HostFlash,
+        offset: u32,
+        data: &[u8],
+    ) -> Result<(), APOBError> {
+        if offset as usize % drv_hf_api::SECTOR_SIZE_BYTES == 0 {
+            hf.bonus_sector_erase(offset)
+                .map_err(|err| APOBError::EraseFailed { offset, err })?;
+        } else {
+            // Read back the page and confirm that it's all empty
```
jgallagher, line 1070: Could this fail if there's a blip in the IPCC path and the host resends an `APOB` request? (I'm not sure what the expectations are for the offsets the host is providing.)

hawkw (commented):
Some mostly kind of annoying nitpicks.

task/host-sp-comms/src/main.rs
```diff
@@ -134,6 +134,11 @@ enum Trace {
         #[count(children)]
         message: SpToHost,
     },
+    APOBWriteError {
+        offset: u64,
+        #[count(children)]
+        err: APOBError,
```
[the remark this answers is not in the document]
mkeeter: Renamed to `page_offset`

task/host-sp-comms/src/main.rs
```diff
@@ -1021,6 +1073,87 @@ impl ServerImpl {
...
     }
 
+    /// Write data to the bonus region of flash
+    ///
+    /// This does not take `&self` because we need to force a split borrow
+    fn apob_write(
+        hf: &HostFlash,
+        mut offset: u64,
+        data: &[u8],
+    ) -> Result<(), ApobError> {
```
rmustacc: So we need something in the broader API to discover what the erase size granularity is or at least make sure that we're sending stuff that is page size aligned. This gets to what @jgallagher gets at below. But if the host sent things that wasn't page aligned then we'd erase the entire page because our API is not doing a read-modify-write.

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,660 @@ impl From<HfError> for ApobError {
...
+    /// Must always be `APOB_PERSISTENT_DATA_HEADER_VERSION` (for now)
+    header_version: u32,
+
+    /// Monotonically increasing counter
+    pub monotonic_counter: u64,
+
+    /// Either 0 or 1; directly translatable to [`ApobSlot`]
+    pub slot_select: u32,
+
+    /// CRC-32 over the rest of the data using the iSCSI polynomial
```
labbott, line 317: Any significance to the choice of polynomial? (curiosity more than anything)

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,660 @@ impl From<HfError> for ApobError {
...
+            header_version: APOB_PERSISTENT_DATA_HEADER_VERSION,
+            monotonic_counter,
+            slot_select: match slot {
+                ApobSlot::Slot0 => 0,
+                ApobSlot::Slot1 => 1,
+            },
+            checksum: 0, // dummy value
+        };
+        out.checksum = out.expected_checksum();
+        assert!(out.is_valid());
```
labbott, line 350: `assert` becomes panic, is that the behavior we want here?

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,660 @@ impl From<HfError> for ApobError {
...
+    oxide_magic: u32,
+
+    /// Must always be `APOB_PERSISTENT_DATA_HEADER_VERSION` (for now)
+    header_version: u32,
+
+    /// Monotonically increasing counter
+    pub monotonic_counter: u64,
+
+    /// Either 0 or 1; directly translatable to [`ApobSlot`]
+    pub slot_select: u32,
```
labbott, line 315: There are a couple of places where we have `unreachable` because of using the `u32`, maaaybe an enum would be cleaner and reduce a few checks? Or is the issue this needs to match exactly what the host is expecting?

drv/cosmo-hf/src/hf.rs
```diff
@@ -98,21 +102,6 @@ impl ServerImpl {
         }
     }
 
-    /// Converts a relative address to an absolute address in bonus space
-    fn bonus_addr(offset: u32, len: u32) -> Result<FlashAddr, HfError> {
```
labbott, line 102: rip `bonus_addr` and friends

drv/cosmo-hf/src/main.rs
```diff
@@ -429,7 +433,7 @@ impl FlashDriver {
 /// Failure function, running an Idol response loop that always returns an error
 fn fail(err: drv_hf_api::HfError) {
     let mut buffer = [0; hf::idl::INCOMING_SIZE];
-    let mut server = hf::idl::FailServer::new(err);
+    let mut server = hf::FailServer(err);
```
labbott, line 436: Hmmm, do we lose the `idl` generation of this?

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,660 @@ impl From<HfError> for ApobError {
...
+                    Err(ApobBeginError::InvalidState)
+                }
+            }
+        }
+    }
+
+    pub(crate) fn write(
+        &mut self,
+        drv: &mut FlashDriver,
+        offset: u64,
```
labbott: Do we need the `u64` everywhere? It seems like everything gets converted/checked against `u32` anyway

labbott (commented):
Will probably do another pass later

drv/cosmo-hf/src/main.rs
mkeeter: Yeah – I switched to finer-grained error types for a few methods, which means that the IDL generator can't make a `FailServer` (which assumes a single error type).

drv/cosmo-hf/src/apob.rs
mkeeter: This isn't seen at all by the host, but we're reading / writing this object directly to disk, so we need `zerocopy`-friendly types.

drv/cosmo-hf/src/apob.rs
mkeeter: I copied what we used for [`HfRawPersistentData`](https://github.com/oxidecomputer/hubris/blob/2dc17d22155df3e999bc4ba24534ea69881ad3f1/drv/hf-api/src/lib.rs#L141-L159) 😄

drv/cosmo-hf/src/hf.rs
mkeeter: 🫡

drv/cosmo-hf/src/apob.rs
mkeeter: The initial `u64` is sent by the host, but I pushed the `u32` conversion upstream into `host_sp_comms`.

drv/cosmo-hf/src/apob.rs
hawkw: FWIW, `zerocopy::TryFromBytes` can be derived for enums (though I'm not sure how annoying that would be to actually use here)

drv/cosmo-hf/src/apob.rs
mkeeter: I think so; the only way this should panic is if someone has broken the code in a dramatic way (e.g. editing the implementation of `is_valid` so that previously valid data is no longer valid).

mkeeter: But this is also copied from `HfRawPersistentData`, so I didn't think about it too much!

citrus-it:
@mkeeter - here are the IPCC messages which are expected to be seen before the host is finished with APOB. Anything else incoming from the host should trigger the lockdown.

```
   TOTAL VARIANT
      99 Request(ApobRead)
      96 Request(ApobData)
       2 Request(KeyLookup)
       1 Request(GetBootStorageUnit)
       1 Request(GetIdentity)
       1 Request(GetStatus)
       1 Request(AckSpStart)
       1 Request(ApobBegin)
       1 Request(ApobCommit)
```

mkeeter:
@citrus-it Great, thanks!  I've pushed this list to Hubris and to RFD 593.

task/host-sp-comms/src/main.rs
mkeeter: This has changed a bunch since February; messages should now all be idempotent.

task/host-sp-comms/src/main.rs
mkeeter: This has all changed now – we ensure that the to-be-written page is erased when the state machine starts, and writes are idempotent.

drv/cosmo-hf/src/apob.rs
mkeeter: I think it would be usable, but the docs seem to imply that you shouldn't it when round-tripping through bytes (?!).

I've opened https://github.com/google/zerocopy/issues/2722 to ask for clarification

app/gimlet/base.toml
```diff
@@ -240,7 +240,7 @@ features = ["stm32h753", "uart7", "baud_rate_3M", "hardware_flow_control", "vlan
 uses = ["uart7", "dbgmcu"]
 interrupts = {"uart7.irq" = "usart-irq"}
 priority = 8
-max-sizes = {flash = 65952, ram = 65536}
+max-sizes = {flash = 67648, ram = 65536}
```
hawkw, line 243: Huh, feels a little unfortunate that this change impacts Gimlet at all, given that (IIUC) the APOB stuff is ~SP5-specific. Is this just due to the addition of new IPCC messages?

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+    BadApobSize {
+        expected: u32,
+        actual: u32,
+    },
+    BadApobWalk {
+        expected: u32,
+        actual: u32,
+    },
+}
+ringbuf!(Trace, 16, Trace::None);
```
hawkw: 16 entries is a pretty small ringbuf, might we want to also generate counters for the various operations like slot erase/sector erase/persistent data write, and/or the total number of errors?

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+        let base = self.base_addr();
+        if offset >= APOB_SLOT_SIZE {
+            return None;
+        }
+        base.0.checked_add(offset).and_then(FlashAddr::new)
+    }
+}
+
+#[derive(Copy, Clone, PartialEq)]
+pub(crate) enum ApobState {
```
hawkw, line 275: nitpicky, feel free to ignore: it might be nice to have a doc comment on this referencing the state diagram in https://rfd.shared.oxide.computer/rfd/593?expand=1#_production_strength_implementation ?

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+    }
+
+    fn expected_checksum(&self) -> u32 {
+        static CRC: crc::Crc<u32> = crc::Crc::<u32>::new(&crc::CRC_32_ISCSI);
+        let mut c = CRC.digest();
+        // We do a CRC32 of everything except the checksum, which is positioned
+        // at the end of the struct and is a `u32`
+        let size = core::mem::size_of::<ApobRawPersistentData>()
+            - core::mem::size_of::<u32>();
+        c.update(&self.as_bytes()[..size]);
```
hawkw, line 361: turbo nit, feel free to ignore: it occurs to me that this could also be achieved by having two `#[repr(C)] #[derive(IntoBytes)]` structs and nesting the "used to calculate CRC32" parts inside the outer struct, like:

```rust
#[derive(
    Copy, Clone, Eq, PartialEq, IntoBytes, FromBytes, Immutable, KnownLayout,
)]
#[repr(C)]
pub struct ApobRawPersistentData {
    pub inner: ApobRawPersistentDataInner,
    checksum: u32,
}

#[derive(
    Copy, Clone, Eq, PartialEq, IntoBytes, FromBytes, Immutable, KnownLayout,
)]
#[repr(C)]
pub struct ApobRawPersistentDataInner {
    oxide_magic: u32,
    header_version: u32,
    pub monotonic_counter: u64,
    pub slot_select: u32,
}

impl ApobRawPersistentData {

    // ...

    fn expected_checksum(&self) -> u32 {
        static CRC: crc::Crc<u32> = crc::Crc::<u32>::new(&crc::CRC_32_ISCSI);
        let mut c = CRC.digest();
        // We do a CRC32 of everything except the checksum, which is positioned
        // at the end of the struct and is a `u32`
        c.update(&self.inner.as_bytes());
        c.finalize()
    }
    
    // ...
}
```

is this worth it? well...perhaps not!

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+        }
+        base.0.checked_add(offset).and_then(FlashAddr::new)
+    }
+}
+
+impl ApobState {
+    /// Initializes the `ApobState`
+    ///
+    /// Searches for an active slot in the metadata regions, updating the offset
+    /// in the FPGA driver if found, and erases unused or invalid slots.
```
hawkw, line 399: where does the erasure happen in this function? i don't _think_ that happens in `set_apob_offset`, right?

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+    ) -> Option<ApobRawPersistentData> {
+        let a = Self::slot_scan(drv, Meta::Meta0);
+        let b = Self::slot_scan(drv, Meta::Meta1);
+
+        match (a, b) {
+            (Some(a), Some(b)) => Some(a.max(b)),
+            (Some(_), None) => a,
+            (None, Some(_)) => b,
+            (None, None) => None,
+        }
```
hawkw: FWIW, there's a `core::cmp::PartialOrd` impl for `Option`s where `None` always compares less than `Some`, so I'm pretty sure this reduces to just
```suggestion
        a.max(b)
```

drv/cosmo-hf/src/apob.rs
hawkw: **Update**: OH it's `get_slot` that does that. this naming was shocking and disturbing to me, i would not expect `get_thing` to also delete all the other `thing`s.

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+    }
+
+    /// Finds the currently active APOB slot, erasing any unused slots
+    fn get_slot(drv: &mut FlashDriver) -> Option<ApobSlot> {
```
hawkw, line 441: naming quibble: i really don't love that `get_slot` (which i expected would just be some kinda lookup) not only `get`s a `slot`, but also erases all the other slots...

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+        static_assertions::const_assert!(
+            (SECTOR_SIZE_BYTES as usize).is_multiple_of(PAGE_SIZE_BYTES)
+        );
+        let size = size.next_multiple_of(SECTOR_SIZE_BYTES);
+        assert!(size <= APOB_SLOT_SIZE);
+
+        // Read back each sector and decide whether to erase it.  We round up
+        // here to the nearest sector
+        for sector_offset in (0..size).step_by(SECTOR_SIZE_BYTES as usize) {
+            let mut buf = [0u8; PAGE_SIZE_BYTES];
```
hawkw: might we consider using a statically allocated buffer with `ClaimOnceCell` for this, so that we don't `PAGE_SIZE_BYTES` on the stack?

since it mutably borrows the `FlashDriver`, if the driver owned an `&'static mut [u8; PAGE_SIZE_BYTES]`, it could safely reuse that buffer for every operation that reads a whole page into memory and not ever have to push a `PAGE_SIZE_BYTES`-sized frame

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+        drv: &mut FlashDriver,
+        meta: Meta,
+    ) -> Option<ApobRawPersistentData> {
+        let mut best: Option<ApobRawPersistentData> = None;
+        for offset in (0..APOB_META_SIZE).step_by(APOB_PERSISTENT_DATA_STRIDE) {
+            let mut data = ApobRawPersistentData::new_zeroed();
+            let addr = meta.flash_addr(offset).unwrap_lite();
+            // flash_read is infallible when using a slice
+            drv.flash_read(addr, &mut data.as_mut_bytes()).unwrap_lite();
+            best = best.max(Some(data).filter(|d| d.is_valid()));
```
hawkw: this...feels a bit too clever. it took me a minute to figure out what it was doing. the one liner is cute, but i think it would be much easier to understand if it was like this:
```suggestion
            if data.is_valid() {
                 best = best.max(Some(data));
            }
```

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+
+        // Check that the end of the data range is within our expected length
+        if offset
+            .checked_add(data.len() as u32)
+            .is_none_or(|d| d > expected_length)
+        {
+            return Err(ApobWriteError::InvalidSize);
+        }
+        let mut out_buf = [0u8; PAGE_SIZE_BYTES];
+        let mut scratch_buf = [0u8; PAGE_SIZE_BYTES];
```
hawkw: similarly to my above comment, we could probably move one or perhaps both of these buffers off the stack and into statics.

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+
+        // Check that the end of the data range is within a slot size
+        if offset
+            .checked_add(data.len() as u32)
+            .is_none_or(|d| d > APOB_SLOT_SIZE)
+        {
+            return Err(ApobReadError::InvalidSize);
+        }
+
+        let mut out_buf = [0u8; PAGE_SIZE_BYTES];
```
hawkw: could also probably not be on the stack

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+                    all_matches = false;
+                    if *a != 0xFF {
+                        return Err(ApobWriteError::NotErased);
+                    }
+                }
+            }
+            if !all_matches {
+                drv.flash_write(addr, &mut &out_buf[..n])
+                    .map_err(|_| ApobWriteError::WriteFailed)?;
+            }
```
hawkw: i _nearly_ suggested that we break out of the loop here upon encountering the first non-matching byte, but i realized that's not actually correct; specifically we have to always loop over the entire current contents in case there is a byte that is both _not erased_ and _non-matching_. so finding a `ff` byte should still keep going. it took a little bit of thinking to figure that out but perhaps i'm just stupid.

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+        }
+
+        let mut out_buf = [0u8; PAGE_SIZE_BYTES];
+        for i in (0..data.len()).step_by(PAGE_SIZE_BYTES) {
+            // Read data from the lease into local storage
+            let n = (data.len() - i).min(PAGE_SIZE_BYTES);
+            let addr = read_slot.flash_addr(i as u32 + offset).unwrap_lite();
+
+            // Read back the current data, then write it to the lease
+            drv.flash_read(addr, &mut &mut out_buf[..n])
```
hawkw: `&mut &mut out_buf[..n]` :eyes:

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+            return r;
+        }
+
+        // We will write persistent data to flash which selects our new slot
+        let perst = Self::get_raw_persistent_data(drv);
+        let new_counter = perst
+            .map(|p| p.monotonic_counter)
+            .unwrap_or(1)
+            .wrapping_add(1);
+        let new_perst = ApobRawPersistentData::new(write_slot, new_counter);
```
hawkw: the string "perst" makes me think of [a very different thing](https://www.intel.com/content/www/us/en/docs/programmable/683059/22-3-8-3-0/pci-express-resets.html) and i got confused. perhaps we ought to call it something slightly different?

drv/cosmo-hf/src/apob.rs
```diff
@@ -122,3 +129,680 @@ impl From<HfError> for ApobError {
...
+    }
+
+    fn apob_validate(
+        drv: &mut FlashDriver,
+        expected_length: u32,
+        expected_hash: ApobHash,
+        write_slot: ApobSlot,
+    ) -> Result<(), ApobCommitError> {
+        // Confirm that the hash of data matches our expectations
+        let mut buf = [0u8; PAGE_SIZE_BYTES];
```
hawkw: also could be not on the stack if you blah blah blah

drv/cosmo-hf/src/hf.rs
```diff
@@ -17,21 +17,23 @@ use userlib::{set_timer_relative, task_slot, RecvMessage, UnwrapLite};
...
-const SLOT_SIZE_BYTES: u32 = 1024 * 1024 * 32;
-const BONUS_SIZE_BYTES: u32 = 1024 * 1024 * 64;
+pub(crate) const SLOT_SIZE_BYTES: u32 = 1024 * 1024 * 32;
 
 pub struct ServerImpl {
     pub drv: FlashDriver,
     pub dev: HfDevSelect,
     hash: HashData,
+
+    // Used in the `apob` module for implementation
```
hawkw: for implementation?

drv/hf-api/src/lib.rs
```diff
@@ -271,4 +271,118 @@ pub struct HfChipId {
...
+    Deserialize,
+    Serialize,
+    SerializedSize,
+    IdolError,
+    FromPrimitive,
+    counters::Count,
+)]
+pub enum ApobBeginError {
+    /// APOB is not implemented on this hardware
+    NotImplemented = 1,
```
hawkw, line 301: `ImAGimlet` :)

drv/hf-api/src/lib.rs
```diff
@@ -271,4 +271,118 @@ pub struct HfChipId {
...
+    PartialEq,
+    Eq,
+    Deserialize,
+    Serialize,
+    SerializedSize,
+    FromPrimitive,
+    IdolError,
+    counters::Count,
+)]
+pub enum ApobCommitError {
```
hawkw, line 377: thank you for having separate error types for each operation <3

drv/spartan7-loader/cosmo-seq/README.md
```diff
@@ -1,3 +1 @@
-FPGA images and collateral are generated from:
-[this sha](https://github.com/oxidecomputer/quartz/commit/bdc5fb31e1905a1b66c19647fe2d156dd1b97b7b)
-[release](https://api.github.com/repos/oxidecomputer/quartz/releases/242278756)
\ No newline at end of file
+FPGA images and collateral are generated from some hw guy's local build
```
hawkw: i assume this is going to be changed before merging this? :)

task/host-sp-comms/src/main.rs
```diff
@@ -1052,6 +1116,128 @@ impl ServerImpl {
...
+                        Err(ApobBeginError::InvalidState) => {
+                            ApobBeginResult::InvalidState
+                        }
+                        Err(ApobBeginError::BadDataLength) => {
+                            ApobBeginResult::BadDataLength
+                        }
+                    }
+                } else {
+                    ApobBeginResult::BadHashLength
+                }
```
hawkw, line 1153: take it or leave it, might be stupid, but we could make this less rightward-drift-y
```suggestion
                let Ok(d) = data.try_into() else {
                    return ApobBeginResult::BadHashLength;
                };
                let hash = ApobHash::Sha256(d);
                match hf.apob_begin(length, hash) {
                    Ok(()) => ApobBeginResult::Ok,
                    Err(ApobBeginError::NotImplemented) => {
                        ApobBeginResult::NotImplemented
                    }
                    Err(ApobBeginError::InvalidState) => {
                        ApobBeginResult::InvalidState
                    }
                    Err(ApobBeginError::BadDataLength) => {
                        ApobBeginResult::BadDataLength
                    }
                }
```

app/gimlet/base.toml
mkeeter: Yeah, the Gimlet / Cosmo-specific stuff is all in `cosmo-hf` (or stubbed out in `gimlet-hf`); this lets `host-sp-comms` be generic across boards, but means that it's got more code in Gimlet as well.

drv/cosmo-hf/src/apob.rs
mkeeter: it's a mutable mutable buffer, nothing to see here

task/host-sp-comms/src/main.rs
mkeeter: I'm ambivalent here, since reducing rightward drift doesn't let us pack this into fewer lines (and adds an explicit `return`).

drv/hf-api/src/lib.rs
mkeeter: At great cost!  (we can no longer use the autogenerated `idl::FailServer`, which requires the same error type for all ops)

drv/cosmo-hf/src/apob.rs
mkeeter: I agree with the Counters Evangelism Strike Force, and made this counted in 8d7c2698 (and added sub-counts for slot / meta variants).

drv/cosmo-hf/src/apob.rs
mkeeter: Done in 8d7c2698

drv/cosmo-hf/src/apob.rs
mkeeter: Neat, done in 8d7c2698

drv/cosmo-hf/src/apob.rs
mkeeter: Probably, done in 8d7c2698

drv/cosmo-hf/src/apob.rs
mkeeter: I would normally say this is a you problem, but we probably have enough other PCIe Aficionados at the company that it's worth changing 😛 

Renamed to `meta_data` in 8d7c2698

drv/cosmo-hf/src/apob.rs
mkeeter: I kinda like this, but couldn't get it to work: `ApobRawPersistentDataInner` has an alignment of 8 bytes (due to its `u64`), which adds padding to `ApobRawPersistentData`, so we can't use `zerocopy`.  We _could_ label `ApobRawPersistentDataInner` as `packed` to remove the padding, but that gives it a `struct` alignment of 1, so the compiler complains:
```
error[E0793]: reference to packed field is unaligned
  --> src/lib.rs:29:9
   |
29 | /         self.inner
30 | |             .monotonic_counter
   | |______________________________^
   |
   = note: packed structs are only aligned by one byte, and many modern architectures penalize unaligned field accesses
   = note: creating a misaligned reference is undefined behavior (even if that reference is never dereferenced)
   = help: copy the field contents to a local variable, or replace the reference with a raw pointer and use `read_unaligned`/`write_unaligned` (loads and stores via `*p` must be properly aligned even when using raw pointers)
```

I expected `packed(4)` to work, but it fails in the same way (??), despite `core::mem::align_of::<ApobRawPersistentDataInner>()` returning 4.

drv/cosmo-hf/src/apob.rs
mkeeter: Sure, done in ee29443fe425

This knocks 472 off our max stack depth, roughly what I'd expect (two `PAGE_SIZE` buffers)
```
hf: 2392 bytes (limit is 4000)
     [+0] _start
  [+1568] main
   [+616] drv_cosmo_hf::apob::ApobState::slot_erase_range
    [+80] drv_cosmo_hf::FlashDriver::flash_read
     [+8] <core::iter::adapters::enumerate::Enumerate<I> as core::iter::traits::iterator::Iterator>::next
    [+32] core::panicking::panic_const::panic_const_add_overflow
    [+24] core::panicking::panic_fmt
    [+56] __rustc[b0c4d0eb67640fa4]
     [+8] core::fmt::Write::write_fmt
```

```
hf: 1920 bytes (limit is 4000)
     [+0] _start
  [+1328] main
    [+56] drv_cosmo_hf::apob::ApobState::init
    [+80] drv_cosmo_hf::apob::ApobState::get_raw_persistent_data
   [+248] drv_cosmo_hf::apob::ApobState::slot_scan
    [+80] drv_cosmo_hf::FlashDriver::flash_read
     [+8] <core::iter::adapters::enumerate::Enumerate<I> as core::iter::traits::iterator::Iterator>::next
    [+32] core::panicking::panic_const::panic_const_add_overflow
    [+24] core::panicking::panic_fmt
    [+56] __rustc[b0c4d0eb67640fa4]
     [+8] core::fmt::Write::write_fmt
```

drv/cosmo-hf/src/apob.rs
mkeeter: Ah, `packed(4)` shouldn't necessarily work, because that could leave the `u64` aligned to 4 bytes rather than 8.  I'm going to punt on this, though I appreciate the thought!

drv/cosmo-hf/src/apob.rs
mkeeter: I switched the polarity of this conditional to be `needs_write` instead of `all_matches`, which makes it a little more clear.  I also added a big comment (4a12383ec34f)

mkeeter: ...and a more explicit comment telling people not to add a `break` in 54061dc35390

drv/cosmo-hf/src/hf.rs
mkeeter: I have no idea what this was intended to say, baleeted.

drv/cosmo-hf/src/apob.rs
```diff
@@ -237,6 +237,32 @@ impl ApobSlot {
...
+        [0; PAGE_SIZE_BYTES],
+        [0; PAGE_SIZE_BYTES],
+    ));
+    let (persistent_data, page, scratch) = BUFS.claim();
+    ApobBuf {
+        persistent_data,
+        page,
+        scratch,
+    }
+}
```
hawkw: turbo nits: why not:
```suggestion
pub(crate) struct ApobBufs {
    persistent_data: &'static mut [u8; APOB_PERSISTENT_DATA_STRIDE],
    page: &'static mut [u8; PAGE_SIZE_BYTES],
    scratch: &'static mut [u8; PAGE_SIZE_BYTES],
}

impl ApobBufs {
     /// Grabs references to the static buffers.  Can only be called once.
     pub fn claim_statics() -> Self {
         use static_cell::ClaimOnceCell;
         static BUFS: ClaimOnceCell<(
             [u8; APOB_PERSISTENT_DATA_STRIDE],
             [u8; PAGE_SIZE_BYTES],
             [u8; PAGE_SIZE_BYTES],
         )> = ClaimOnceCell::new((
             [0; APOB_PERSISTENT_DATA_STRIDE],
             [0; PAGE_SIZE_BYTES],
             [0; PAGE_SIZE_BYTES],
         ));
         let (persistent_data, page, scratch) = BUFS.claim();
         Self {
             persistent_data,
             page,
             scratch,
         }
}
```

drv/cosmo-hf/src/apob.rs
```diff
@@ -367,8 +393,8 @@ impl ApobState {
     ///
     /// Searches for an active slot in the metadata regions, updating the offset
     /// in the FPGA driver if found, and erases unused or invalid slots.
-    pub(crate) fn init(drv: &mut FlashDriver) -> Self {
-        if let Some(s) = Self::get_slot(drv) {
+    pub(crate) fn init(drv: &mut FlashDriver, buf: &mut ApobBuf) -> Self {
```
hawkw: super unimportant nit: what do you think about perhaps factoring all this into an `Apob` struct that owns both the `ApobState` and the bufs, so they need not be passed in to every single thing? not a big deal

app/gimlet/base.toml
hawkw: yeah. probably not worth the effort to feature flag out a lot of the APOB stuff in IPCC to save some flash on Gimlet --- i'm not sure how possible that would even be.

drv/cosmo-hf/src/apob.rs
hawkw: Ah well. It was worth a shot, thanks for trying!

drv/cosmo-hf/src/apob.rs
hawkw: yay!

drv/cosmo-hf/src/apob.rs
```diff
@@ -145,15 +145,18 @@ pub const APOB_META1_ADDR: u32 = APOB_META0_ADDR + APOB_META_SIZE;
 pub const APOB_SLOT0_ADDR: u32 = APOB_META1_ADDR + APOB_META_SIZE;
 pub const APOB_SLOT1_ADDR: u32 = APOB_SLOT0_ADDR + APOB_SLOT_SIZE;
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, PartialEq, counters::Count)]
```
hawkw, line 148: counters look nice --- thanks!

drv/cosmo-hf/src/apob.rs
```diff
@@ -225,7 +237,39 @@ impl ApobSlot {
...
+    }
+}
+
+/// State machine class, which implements the logic from RFD 593
+///
+/// See rfd.shared.oxide.computer/rfd/593#_production_strength_implementation
+/// for details on the states and transitions.  Note that the diagram in the RFD
+/// includes fine-grained states (e.g. writing), which the actual implementation
+/// never dwells in; these states are not explicit in `ApobState`.
+#[derive(Copy, Clone, PartialEq, counters::Count)]
```
hawkw, line 274: since `ApobSlot` is countable, we could probably put `#[count(children)]` on the `write_slot` fields in `Waiting` and `Ready` --- not sure if that's worth the extra few bytes of counters, but maybe it's useful?

drv/cosmo-hf/src/apob.rs
```diff
@@ -225,7 +237,39 @@ impl ApobSlot {
...
+    ));
+    let (persistent_data, page, scratch) = BUFS.claim();
+    ApobBuf {
+        persistent_data,
+        page,
+        scratch,
+    }
+}
+
+/// State machine class, which implements the logic from RFD 593
```
hawkw: super nit: i might not have said "class" here.

drv/cosmo-hf/src/apob.rs
```diff
@@ -225,7 +237,39 @@ impl ApobSlot {
...
+///
+/// See rfd.shared.oxide.computer/rfd/593#_production_strength_implementation
+/// for details on the states and transitions.  Note that the diagram in the RFD
+/// includes fine-grained states (e.g. writing), which the actual implementation
+/// never dwells in; these states are not explicit in `ApobState`.
```
hawkw, line 273: <3

drv/spartan7-loader/cosmo-seq/README.md
mkeeter: Done, thanks Aaron

drv/cosmo-hf/src/apob.rs
```diff
@@ -554,33 +611,42 @@ impl ApobState {
...
+            // non-matching byte that is *also* not erased.
+            let mut needs_write = false;
+            for (a, b) in buf.scratch[..n].iter().zip(buf.page[..n].iter()) {
                 if *a != *b {
-                    all_matches = false;
+                    // You may be tempted to insert a `break` here, but that
+                    // would be incorrect: there could be subsequent bytes which
+                    // do not match *and* are not erased, in which case we must
+                    // return `NotErased`.
+                    needs_write = true;
```
hawkw, line 641: comments here aer lovely, thank you for adding them --- this is all much clearer now.

drv/cosmo-hf/src/apob.rs
mkeeter: Sure, done in 7d5c7574757f

drv/cosmo-hf/src/apob.rs
mkeeter: Fixed in 7d5c7574757f

drv/cosmo-hf/src/apob.rs
mkeeter: Sure – I actually put it on the `read_slot` field in `Waiting`, because that's more informative (it tells us whether there was valid data available at all; `write_slot` is always populated).

drv/cosmo-hf/src/apob.rs
mkeeter: Fixed so that `get_slot` just returns the slot, and `init(..)` does the erasing itself

drv/cosmo-hf/src/apob.rs
mkeeter: Fixed so that `get_slot` just returns the slot, and `init(..)` does the erasing itself, which is less shocking and disturbing

drv/cosmo-hf/src/apob.rs
mkeeter: I poked at this, but it ends up being pretty messy / not worth it (there are functions which are called during `init` without a fully constructed state, so you end up still passing around `&mut ApobBufs` in a bunch of places).

drv/cosmo-hf/src/apob.rs
hawkw: ah, yup, that's nicer. thanks!

drv/cosmo-hf/src/apob.rs
hawkw: cool, carry on!

drv/cosmo-hf/src/apob.rs
```diff
@@ -393,22 +397,35 @@ impl ApobState {
...
+
             ApobState::Waiting {
                 read_slot: Some(s),
                 write_slot: !s,
             }
         } else {
+            // Erase both slots
+            Self::slot_erase(drv, buf, ApobSlot::Slot0);
+            Self::slot_erase(drv, buf, ApobSlot::Slot1);
+
```
hawkw, line 417: this is a lot clearer than hiding the erasure in `get_slot`, thank you!

drv/cosmo-hf/src/apob.rs
hawkw: lovely, thank you!

PR #2009, opened 2025-02-10
transceivers: pull in power enable status fix

Aaron-Hartwig:
This pulls in an update QSFP front IO FPGA at https://github.com/oxidecomputer/quartz/commit/9209c858a8c91d7ab4d81e146803dcc4f4b8c280. This FPGA change makes it more clear if power is actually enabled to a module or not by splitting out the software power enable register from a hardware readback power enable register. In practice the FPGA gated on the enable bit being set **and** module presence, but that was not reflected in the `power_enable` field of module status. The only adjustment needed in the server is that we must write to the new `QSFP_SW_POWER_ENx` registers instead of the old `QSFP_POWER_ENx` registers which are now read only.


Aaron-Hartwig:
With the current code you can see that all the `power_enable` bits are set as that is the default state:
```
aaron@niles ~ $ pfexec humility -t sidecar-b hiffy -c Transceivers.get_module_status
Transceivers.get_module_status() => ModuleStatus { power_enable: 0xffffffff, power_good: 0x101213d, power_good_timeout: 0x0, power_good_fault: 0x0, resetl: 0x101213d, lpmode_txdis: 0x101213d, modprsl: 0xfefedec2, intl_rxlosl: 0x0 }
```

With this update, `power_enabled` is set for only the ports we have a module inserted **and** the SW power enable bit set:
```
aaron@niles ~ $ pfexec humility -t sidecar-b-aaron hiffy -c Transceivers.get_module_status
Transceivers.get_module_status() => ModuleStatus { power_enable: 0x101213d, power_good: 0x101213d, power_good_timeout: 0x0, power_good_fault: 0x0, resetl: 0x101213d, lpmode_txdis: 0x101213d, modprsl: 0xfefedec2, intl_rxlosl: 0x0 }
```

nathanaelhuffman (approved):
lgtm

PR #2010, opened 2025-02-10
Peripheral codegen for Grapefruit / Cosmo

mkeeter:
This PR adds a new target to `build-fpga-regmap`, generating Rust code for a memory-mapped peripheral.  For Grapefruit and Cosmo, the Spartan-7 FPGA is connected over FMC, so we can treat peripherals implemented in the FPGA as native(ish).  This PR uses the new codegen paths for `cosmo-hf` and `grapefruit-seq`, replacing raw addresses.  The peripheral types take a `drv_spartan7_loader_api::Spartan7Token`, which proves that the FPGA is loaded and running.

Other changes that came along for the ride:

- Grapefruit FPGA files (bitstream and JSON) are moved to `drv/spartan7-loader/grapefruit/` subfolder.  I also updated to a more recent build (latest `main`), and removed the HTML file (since the JSON is smaller + has the same information).
- `syn` is bumped to version 2, which requires a few miscellaneous changes


build/fpga-regmap/src/lib.rs
```diff
@@ -33,11 +38,31 @@ enum Node {
...
 #[derive(Debug, Deserialize)]
-struct EnumEncode {
+pub struct EnumEncode {
     name: String,
     value: u8,
 }
 
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum SwAccess {
```
nathanaelhuffman, line 48: As a note, there are a few more possible types here: https://systemrdl-compiler.readthedocs.io/en/stable/api/types.html#systemrdl.rdltypes.builtin_enums.AccessType
that could show up.

We also don't do much with the OnRead and OnWrite types described below that section but am open to doing so if there's code-gen value.

nathanaelhuffman (approved):
Left a minor comment about some other access types we might see.  We have a lot of other stuff that could be exposed from the RDL side, but probably best to deal with it along-and-along as we start experiencing them.

Excited about making this more 1st class!

labbott (approved):
Minor questions but LGTM

build/fpga-regmap/src/lib.rs
mkeeter: Interesting, I'm going to leave those as unimplemented so that they'll fail loudly if we ever end up using them!

drv/cosmo-hf/src/main.rs
```diff
@@ -67,14 +65,16 @@ fn main() -> ! {
     // Idol loop after the FPGA has been brought up.
     let seq =
         drv_spartan7_loader_api::Spartan7Loader::from(LOADER.get_task_id());
-    seq.ping();
```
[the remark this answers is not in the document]
mkeeter: Yes, the call to `seq.get_token()` is what ensures that the FPGA is loaded.

PR #2021, opened 2025-02-28
Split Grapefruit into Ruby vs standalone images

mkeeter:
- Splits Grapefruit into `app-dev.toml` (dev boards only) and `app-ruby.toml` (attached to Ruby)
- Switch `app-dev` to using UART8, which is brought out on a connector

This lets us test IPCC on a desktop, connecting to Grapefruit with a USB-to-serial cable.


citrus-it (approved):
This looks fine to me.

PR #2024, opened 2025-03-05
FMC-attached peripherals

mkeeter:
This lets us specify a set of FMC-attached **virtual peripherals**, which are implemented in an FPGA.

A manifest may specify an `[fmc]` configuration object, with a base address and register map (which is a JSON file).  If present, `xtask` parses the JSON file and adds its peripherals to the `chip.toml`'s peripheral list, prefixed with `fmc_`.  Tasks can include these peripherals in their `uses = [...]` list.

Instead of copying `chip.toml` into the output archive, we re-serialize the `peripherals` map, so that it includes these virtual peripherals.  This means that `humility map` works fine:

```
DESC       LOW          HIGH          SIZE ATTR   ID TASK
...elided...
0x08005820 0x60000000 - 0x600000ff     256 rw-d-- 24 [fmc_base] fmc_demo, hf
0x0800585c 0x60000100 - 0x600001ff     256 rw-d-- 24 [fmc_spi_nor] fmc_demo, hf
0x08005834 0x60000200 - 0x600002ff     256 rw-d-- 24 [fmc_espi] fmc_demo
0x08005848 0x60000300 - 0x600003ff     256 rw-d-- 12 [fmc_sgpio] grapefruit_seq, fmc_demo
```

(it does lose the comments in the original `chip.toml`; raise an objection if you think that's a big deal)

I've removed the `fmc_nor_psram_bank_1` peripheral, because Humility doesn't play nice with multiple regions at the same start address.  Instead, `drv-stm32h7-fmc-demo-server` is configured to use each of the FPGA's peripherals.  If we are still using this demo task with > 8 peripherals, then we can reconsider.

### DIFF
--- a/app/grapefruit/base.toml
+++ b/app/grapefruit/base.toml
@@ -7,6 +7,10 @@ epoch = 0
 version = 0
 fwid = true
 
+[mmio]
+peripheral-region = "fmc_nor_psram_bank_1"
+register-map = "../../drv/spartan7-loader/grapefruit/gfruit_top_map.json"
+
 [kernel]
 name = "grapefruit"
 requires = {flash = 32768, ram = 8192}
@@ -177,7 +181,7 @@ max-sizes = {flash = 131072, ram = 16384 }
 stacksize = 2600
 start = true
 task-slots = ["sys", "jefe", "packrat", "spartan7_loader"]
-uses = ["fmc_nor_psram_bank_1"]
+uses = ["mmio_sgpio"]
 
 [tasks.spartan7_loader]
 name = "drv-spartan7-loader"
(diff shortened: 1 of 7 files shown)


app/grapefruit/base.toml
```diff
@@ -7,6 +7,10 @@ epoch = 0
 version = 0
 fwid = true
 
+[fmc]
+base-address = 0x60000000
+register-map = "../../drv/spartan7-loader/grapefruit/gfruit_top_map.json"
+
```
labbott, line 13: It feels off that we have the base address here in the `base.toml` file instead of the `chip.toml` file.

mkeeter: Good call, I've updated the PR to keep `fmc_nor_psram_bank_1` in `chip.toml` – with a comment telling you not to use it directly – then specifying that region by name in the manifest.

build/xtask/src/config.rs
```diff
@@ -23,6 +23,7 @@ struct RawConfig {
     target: String,
     board: String,
     chip: String,
+    fmc: Option<FmcConfig>,
```
cbiffle, line 26: So, "fmc" is a term specific to certain ST microcontrollers... can we think of a more general name for this in the shared config?

labbott (approved):

build/xtask/src/config.rs
mkeeter: switched to `mmio`

PR #2027, opened 2025-03-10
Removed unneeded used_with_arg

steveklabnik:
I'm actually not 100% sure this isn't needed here, but it appears vestigial, so let's give it a CI run.


labbott:
I_think_ this is supposed to be used in conjunction with https://github.com/oxidecomputer/hubris/blob/master/test/test-suite/src/main.rs#L50-L65

steveklabnik:
@labbott yeah but there's already https://github.com/oxidecomputer/hubris/blob/master/test/test-suite/src/main.rs#L21 on that file

PR #2028, opened 2025-03-11
Grapefruit update and peripheral codegen improvements

mkeeter:
The top-level FPGA peripheral map now has more regular structure:

- `orig_type_name` is guaranteed to be the same as the JSON file for the individual peripheral registers
- `inst_name` is a per-peripheral instance name (which will be relevant if we have > 1 copy of the same peripheral)

This means that we don't need to guess the peripheral register file name from the `inst_name`, which was requiring more and more special cases.

In addition, we now define the top-level MMIO map in the manifest, so individual tasks shouldn't also specify it.  `build_fpga_regmap::fpga_peripheral` now just takes the peripheral's `inst_name`, and loads the peripheral map + base address from environment variables set by the build script.


labbott (commented):
LGTM at least from a hubris build perspective, I'll give others a chance to review too

Issue #1958, opened 2024-12-16
Need to separate out active slot for writing from active slot for booting


rmustacc:
In Gimlet the host flash driver manages an analog mux to determine whether we are booting from QSPI 0 or QSPI 1.  This is the `HostFlash.set_dev` RPC as opposed to the `HostFlash.set_mux` RPC. When we perform an update of the host flash we are generally writing to the slot opposite of the one we're booting from (but not necessarily). This will cause a switch to device that will remain until another switch happens or the SP restarts and we read from the in-memory state.

Consider the following sequence:

* control-plane-agent switches the dev from QSPI 0 to QSPI 1
* control-plane-agent is being sent an image to write to the SPI flash in chunks
* The host OS reboots intentionally or unintentionally (e.g. panic, MAPO, etc.)

At this point in time we will end up in a spot where the host will end up getting an incomplete image from QSPI 1. We probably want to keep track of two logical bits in the host flash driver:

1. What is the image I intend to boot from?
2. What is the image that I intend to write to?

Generally once the host successfully boots to the point of issuing an IPCC command we will mux the QSPI flash back to the SP at which point something like this could continue. My assumption is that during an online update of the control plane we will end up wanting to write and stage this update, but the host will be active meaning that we may also want to make sure that we have a clear error message that it is muxed to the host and how long it has been since then so the control plane can deal with that. I expect parts of that already exist.

mkeeter:
I chatted with @nathanaelhuffman about this, and wanted to provide some additional color.

First, this is not Cosmo-specific.  It's easier to explain on Gimlet, where stuff is implemented in hardware:

![Image](https://github.com/user-attachments/assets/9e759c3a-7774-4345-a697-c3fced0961ef)

(see [RFD 163](https://rfd.shared.oxide.computer/rfd/163) for background)

`set_mux` configures the first 2:1 mux, routing QSPI from either the SP3 or the SP to the flash chips.  `set_dev` configures the 1:2 demux, mapping the chip-select line to either flash 1 or flash 2.

The specific mux and dev selections are controlled by various bits of logic in Hubris.

At power-up, flash is initially muxed to the SP.  When we do the A2 → A0 transition, Hubris switches the mux to the host so that it can boot.  After the host boots, it sends the `GetStatus` message over IPCC to Hubris; as a side effect of receiving that message, Hubris swaps the mux back to the SP.

[...]

PR #2042, opened 2025-04-15
Make `flash_dev_select` mandatory in `gimlet-hf-server`

mkeeter:
This is initial cleanup to make #1958 easier.


labbott (approved):
One question from outside eyes but LGTM

drv/gimlet-hf-server/src/main.rs
```diff
@@ -666,75 +640,48 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
-        } else {
-            // Single-flash case is less complicated, because we don't have to
-            // track and maintain the active device.
-            let (prev_data, next) = self.persistent_data_scan();
-            let prev_monotonic_counter = match prev_data {
-                Some(a) => a.monotonic_counter,
-                None => 0,
-            };
+        self.set_dev(prev_slot).unwrap();
+        let (b_data, b_next) = self.persistent_data_scan();
```
labbott, line 651: Doesn't have to be here but do we actually need  the `unwraps` here? We also call `check_muxed_to_sp` in `set_dev` too.

mkeeter: Good catch, we could either discard the value (`let _ = self.set_dev(...);`) or do normal error propagation without changing the resulting behavior.  I'm going to leave the unwraps here to minimize changes, but we could revisit in the future; the downside would be that we'd lose an implicit assertion about expected behavior.

PR #2043, opened 2025-04-15
Don't modify active slot while updating host flash

mkeeter:
We don't want programming the host flash to _also_ set the slot used at boot (see #1958 for details), because interruptions could leave us in an unfortunate state.

This PR adds new `page_program_dev` and `sector_erase_dev` functions, which operate on a particular slot (passed as an argument), but are guaranteed to leave the active slot unchanged when they exit.  In other words, they switch to a particular slot, perform an action, then restore the previous slot before returning the result.

In `control-plane-agent`, we now use those functions during host flash updates (previously, we set the host flash slot during `HostFlashUpdate::prepare` then assumed it wouldn't change).

There's one behavioral change: at the end of the update, we have not (implicitly) switched to the new slot.  Looking at Omicron, it looks like we don't rely on this behavior, but I'd appreciate a second opinion.

Tested on `niles` and it works as expected (I see `HostFlash.get_dev` stay the same during the update).

### DIFF
--- a/drv/cosmo-hf/src/hf.rs
+++ b/drv/cosmo-hf/src/hf.rs
@@ -305,6 +305,21 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
             .map_err(|()| RequestError::went_away())
     }
 
+    fn page_program_dev(
+        &mut self,
+        msg: &RecvMessage,
+        dev: HfDevSelect,
+        addr: u32,
+        protect: HfProtectMode,
+        data: LenLimit<Leased<R, [u8]>, PAGE_SIZE_BYTES>,
+    ) -> Result<(), RequestError<HfError>> {
+        let prev = self.dev;
+        self.set_dev(msg, dev)?; // makes subsequent set_dev infallible
+        let r = self.page_program(msg, addr, protect, data);
+        self.set_dev(msg, prev).unwrap_lite(); // this is infallible!
+        r
+    }
+
     /// Reads a page from the currently selected `dev`
     fn read(
         &mut self,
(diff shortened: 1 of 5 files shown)


drv/cosmo-hf/src/hf.rs
```diff
@@ -339,6 +354,20 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
+        &mut self,
+        msg: &RecvMessage,
+        dev: HfDevSelect,
+        addr: u32,
+        protect: HfProtectMode,
+    ) -> Result<(), RequestError<HfError>> {
+        let prev = self.dev;
+        self.set_dev(msg, dev)?;
+        let r = self.sector_erase(msg, addr, protect);
+        self.set_dev(msg, prev).unwrap();
```
andrewjstone, line 367: Should this be `unwrap_lite()`. What is the difference these days?

andrewjstone (commented):
@mkeeter This looks good. It's my first time looking at hubris code in a while though, so take that with a grain of salt ;)

IIUC, this works because each SP operation resets the device mux after each call, and the sequencer call to `set_mux` will only run after this call returns since it's in the same server. 

However, what happens if the SP crashes or resets before `set_dev` is called at the bottom of these functions? Will we still end up with the updated target slot set? Is this the best we can do?

drv/cosmo-hf/src/hf.rs
mkeeter: They do the same thing, but `unwrap_lite()` doesn't add as much code because it skips formatting the backtrace.  We use one or the other depending on vibes (e.g. if we're trying to keep a task under a particular size boundary).

mkeeter:
> IIUC, this works because each SP operation resets the device mux after each call, and the sequencer call to set_mux will only run after this call returns since it's in the same server.

That's correct, no one can interrupt us mid-operation

> However, what happens if the SP crashes or resets before set_dev is called at the bottom of these functions? Will we still end up with the updated target slot set? Is this the best we can do?

If the SP or `hf` task resets, it will active a slot based on the "persistent" slot stored in the flash.  Conveniently, the `do_stuff_dev(..)` functions are stateless, so if the crash was limited to the `hf` task, the caller won't notice.

andrewjstone:
> If the SP or `hf` task resets, it will active a slot based on the "persistent" slot stored in the flash. Conveniently, the `do_stuff_dev(..)` functions are stateless, so if the crash was limited to the `hf` task, the caller won't notice.

Oh, so this is just a task local being set and reset? It's just in the prior case, the task local remained and overrode the persistent state when setting the mux?

andrewjstone (approved):

labbott (approved):

PR #2048, opened 2025-04-24
minibar: automatically setup the PCIe refclk buffer

Aaron-Hartwig:
Most of the noise here is me adjusting the case of some register fields. The real change here is to some of the reset values of fields in the PCIE_REFCLK_CTRL register in the FPGA to get the refclk from the sled to Minibar's slot. On startup the SP clears the "power down" line to the clock buffer, at which point it samples its strappings and begins operation.


Aaron-Hartwig:
Testing was captured in https://github.com/oxidecomputer/hardware-minibar/issues/194

mkeeter (approved):

PR #2050, opened 2025-04-28
Transient boot selection

lzrd:
Automated update should take advantage of transient boot selection. This is the last piece to make that feature available.
See RFD 374 for a description and [bootleby source](https://github.com/oxidecomputer/bootleby/blob/bdf56dd950b934360df596ed5b2d8b8813c92168/src/lib.rs#L433C1-L457C2) for the logic being driven.


drv/lpc55-update-server/src/main.rs
```diff
@@ -503,6 +508,12 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
         ringbuf_entry!(Trace::State(self.state));
         self.next_block = None;
         self.fw_cache.fill(0);
+        // The sequence: [update, set transient preference, update] is legal.
+        // Clear any stale transient preference before update.
+        // Stage0 doesn't support transient override.
+        if component == RotComponent::Hubris {
+            set_hubris_transient_override(None);
+        }
```
labbott, line 516: I think I see what this is trying to do but I'm not sure this is the right place. This would make `[set transient preference, update]` quietly not work as expected. I'm also not sure about the example `[update, set transient preference, update]` being legal but not supported. Can we clarify what the semantics of transient boot preference are supposed to be? Having it be effective for exactly one boot and clearing it on task startup seems like the least confusing semantics.

labbott (commented):
We need to update `fn boot_preferences` to read our transient preference (GitHub won't let me comment on code not in the PR)

drv/lpc55-update-server/src/main.rs
labbott: Okay I re-read the RFD and double checked the bootleby source, the transient command is supposed to be nuked on boot by bootleby. I'm still not quite sure what makes this a 'stale' choice.

chips/lpc55/memory.toml
```diff
@@ -93,6 +93,22 @@ write = true
 execute = false
 dma = true
 
+[[override]]
+name = "a"
+address = 0x3003ffe0
```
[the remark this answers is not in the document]
lzrd: Yes.

drv/lpc55-update-server/src/main.rs
lzrd: It's not actually stale until some flash has been altered. Given the sequence where SlotId::B is active:
  - prep_image_update SlotId::A
  -  flash to hubris-a with version 1
  - switch_default_hubris_image(SlotId::A, SwitchDuration::Once)
  - prep_image_update SlotId::A
  - // abandon or abort the update before flashing anything
  - // reset the RoT
What image is preferred to boot?
 
This is a corner case since we normally expect update, set-preference, reset.
I'm thinking that in this case, since the RoT update server saw an intent to boot once into SlotId::A and then saw intent to update A again, that should cancel the boot once directive.

We could wait until the first block of an image is received or that block is written to cancel the boot once directive.

labbott: I agree it's a corner case. Fixing it in this way is a pretty poor developer experience because there's no indication that the boot setting has changed which will frustrate everyone. The transient boot setting should not be cancelled without some indication. We also have the exact same issue with the persistent settings. I get the theory is to avoid booting into a potentially bad image because we're now relying on bootleby to correctly determine that the slot is unbootable but this feels like it causes more problems than it solves. I think this issue of boot settings would be better tracked as a separate issue.

lzrd: I think that an alternate approach will work; If current boot preference points to a flash bank that would be altered by some operation, then the operation fails. The remedy would be to abort the update or set the boot preference away from the flash bank to be altered.
So, Active and Alt start with valid images, Alt is then set as next boot preference, attempt to update Alt fails.
Setting boot preference to the Active image works, then updating Alt works, then setting preference to Alt works.
Appropriate errors are generated on the failures as feedback.

labbott: I think that's a very solid approach

lzrd: Need to merge https://github.com/oxidecomputer/management-gateway-service/pull/398 and fix the Cargo.toml here before merging.

lzrd: > Need to merge [oxidecomputer/management-gateway-service#398](https://github.com/oxidecomputer/management-gateway-service/pull/398) and fix the Cargo.toml here before merging.

done

chips/lpc55/memory.toml
```diff
@@ -93,6 +93,22 @@ write = true
...
+name = "a"
+address = 0x2003ffe0
+size = 32
+read = true
+write = true
+execute = false
+
+[[override]]
+name = "b"
+address = 0x3003ffe0
```
labbott: this doesn't match `a`

drv/lpc55-update-server/src/main.rs
```diff
@@ -493,12 +498,48 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+                let active = match bootstate()
+                    .map_err(|_| UpdateError::MissingHandoffData)?
+                    .active
+                {
+                    stage0_handoff::RotSlot::A => SlotId::A,
+                    stage0_handoff::RotSlot::B => SlotId::B,
+                };
+                if active == slot {
+                    return Err(UpdateError::InvalidSlotIdForOperation.into());
+                }
```
labbott: I'm not opposed to adding this check here but I thought we were already doing this somewhere in hubris (maybe not...)

drv/lpc55-update-server/src/main.rs
```diff
@@ -493,12 +498,48 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
-        self.image = match (component, slot) {
+        let image = match (component, slot) {
             (RotComponent::Hubris, SlotId::A)
-            | (RotComponent::Hubris, SlotId::B)
-            | (RotComponent::Stage0, SlotId::B) => Some((component, slot)),
+            | (RotComponent::Hubris, SlotId::B) => {
+                let active = match bootstate()
+                    .map_err(|_| UpdateError::MissingHandoffData)?
+                    .active
+                {
```
labbott: This means our update path now has a dependency on having valid bootstate and means we can't simply update out of an error with our bootstate. Is that something we're comfortable with?

drv/lpc55-update-server/src/main.rs
labbott: https://github.com/oxidecomputer/hubris/blob/master/drv/lpc55-update-server/src/main.rs#L1160-L1163 Yes we already are doing a similar check in the write block path. I think having both is probably appropriate since they are checking slighlty different things.

drv/lpc55-update-server/src/main.rs
```diff
@@ -493,12 +498,48 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
             UpdateState::Finished | UpdateState::NoUpdate => (),
         }
 
-        self.image = match (component, slot) {
+        let image = match (component, slot) {
```
labbott: any reason to do the `let` binding here?

drv/lpc55-update-server/src/main.rs
```diff
@@ -940,6 +1001,19 @@ impl ServerImpl<'_> {
...
+                //   - A is active
+                //   - persistent switch to B without reset
+                //       - scratch page is updated
+                //       - return Ok(())
+                //   - persistent switch to A without reset
+                //       - A is already active, no work performed
+                //       - return Ok(())
+                //   - reset
+                //   - B is active
+
```
labbott: Ahhh I understand the issue better with this example! Can you add this to hubris #2093 ?

lzrd: done

lzrd:
> We need to update `fn boot_preferences` to read our transient preference (GitHub won't let me comment on code not in the PR)

done

chips/lpc55/memory.toml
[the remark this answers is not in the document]
lzrd: thanks, done.

chips/lpc55/memory.toml
lzrd: fixed

drv/lpc55-update-server/src/main.rs
lzrd: Nope. Fixed.

drv/lpc55-update-server/src/main.rs
lzrd: You are correct, the bootstate represents history, not the future. The `same_image` function has the needed information.

lzrd: I added comments to make this area clearer and used `same_image` instead of `bootstate`.
I've also tested upgrade and rollback with and without the transient preference feature between the current master branch and this branch using test scripts and test plan found in [MGS/scripts on the `stoltz/rhai-fault-injection` branch](https://github.com/oxidecomputer/management-gateway-service/tree/stoltz/rhai-fault-injection/scripts)

Cargo.lock
```diff
@@ -2987,15 +2987,16 @@ checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 [[package]]
 name = "gateway-messages"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/management-gateway-service#e71c87a6ee63e8f8d6a1940cfeda00e0c312fa19"
+source = "git+https://github.com/oxidecomputer/management-gateway-service#ea3225658f718b12cd58d611960e0a936be47d0d"
```
labbott: This needs to be the final commit we merge and I don't think the MGS side is ready yet

labbott (commented):
Once the MGS side is ready I will hit approve

drv/lpc55-update-server/src/main.rs
```diff
@@ -905,10 +949,32 @@ impl ServerImpl<'_> {
...
+                // Get the currently active slot.
+                let active_slot = match bootstate()
+                    .map_err(|_| UpdateError::MissingHandoffData)?
+                    .active
+                {
+                    stage0_handoff::RotSlot::A => SlotId::A,
+                    stage0_handoff::RotSlot::B => SlotId::B,
+                };
+
+                // If the requested slot is the same as the active slot,
```
mkeeter: These semantics seem confusing / non-obvious.  I think it would make more sense to explicitly return an error (`InvalidPreferredSlotId`?) if someone tries to set the transient override to be the same as the active slot.

drv/lpc55-update-server/src/main.rs
```diff
@@ -495,8 +500,49 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+
+                let (persistent, pending_persistent, transient) =
+                    self.boot_preferences()?;
+
+                // The transient preference must not select the update target.
+                if let Some(pref) = transient {
+                    if slot == pref {
+                        return Err(UpdateError::InvalidPreferredSlotId.into());
+                    }
+                }
```
mkeeter:
```suggestion
                if transient == Some(pref) {
                    return Err(UpdateError::InvalidPreferredSlotId.into());
                }
```
?

drv/lpc55-update-server/src/main.rs
```diff
@@ -495,8 +500,49 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+                // If there is a pending persistent preference, it must
+                // not select the update target.
+                if let Some(pref) = pending_persistent {
+                    if slot == pref {
+                        return Err(UpdateError::InvalidPreferredSlotId.into());
+                    }
```
mkeeter: Same here:
```suggestion
                if pending_persistent == Some(slot) {
                    return Err(UpdateError::InvalidPreferredSlotId.into());
```

mkeeter:
The use of `MaybeUninit` here seems sketchy; it's only defined in terms of Rust behavior, so using `assume_init` to grab data that happens to be in RAM could be fragile.

(This is admittedly not a new problem introduced in the PR, but the PR adds another case of it)

I think the right way to handle this is using an `extern static`, although it's still a subject of active discussion (https://github.com/rust-lang/unsafe-code-guidelines/issues/397#issuecomment-2619633222)

Here's an (untested) patch which does just that:

```diff
diff --git a/app/lpc55xpresso/app.toml b/app/lpc55xpresso/app.toml
index cc03b0de53..48b479bacb 100644
--- a/app/lpc55xpresso/app.toml
+++ b/app/lpc55xpresso/app.toml
@@ -51,7 +51,8 @@
 max-sizes = {flash = 26720, ram = 16704}
 stacksize = 8192
 start = true
-sections = {bootstate = "usbsram", transient_override = "override"}
+sections = {bootstate = "usbsram"}
+extern-regions = ["transient_override"]
 uses = ["flash_controller", "hash_crypt"]
 notifications = ["flash-irq", "hashcrypt-irq"]
 interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
diff --git a/app/oxide-rot-1/app-dev.toml b/app/oxide-rot-1/app-dev.toml
index 50a945c016..9d45ad973f 100644
--- a/app/oxide-rot-1/app-dev.toml
+++ b/app/oxide-rot-1/app-dev.toml
@@ -55,7 +55,8 @@
 priority = 3
 stacksize = 8192
 start = true
-sections = {bootstate = "usbsram", transient_override = "override"}
+sections = {bootstate = "usbsram"}
+extern-regions = ["transient_override"]
 uses = ["flash_controller", "hash_crypt"]
 notifications = ["flash-irq", "hashcrypt-irq"]
 interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
diff --git a/app/oxide-rot-1/app.toml b/app/oxide-rot-1/app.toml
index 9e12953bad..0c45eb254d 100644
--- a/app/oxide-rot-1/app.toml
+++ b/app/oxide-rot-1/app.toml
@@ -43,7 +43,8 @@
 priority = 3
 stacksize = 8192
 start = true
-sections = {bootstate = "usbsram", transient_override = "override"}
+sections = {bootstate = "usbsram"}
+extern-regions = ["transient_override"]
 uses = ["flash_controller", "hash_crypt"]
 notifications = ["flash-irq", "hashcrypt-irq"]
 interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
diff --git a/app/rot-carrier/app.toml b/app/rot-carrier/app.toml
index 3072a96e1c..74c20d7d64 100644
--- a/app/rot-carrier/app.toml
+++ b/app/rot-carrier/app.toml
@@ -43,7 +43,8 @@
 # TODO size this appropriately
 stacksize = 8192
 start = true
-sections = {bootstate = "usbsram", transient_override = "override"}
+sections = {bootstate = "usbsram"}
+extern-regions = ["transient_override"]
 uses = ["flash_controller", "hash_crypt"]
 notifications = ["flash-irq", "hashcrypt-irq"]
 interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
diff --git a/chips/lpc55/memory.toml b/chips/lpc55/memory.toml
index 0125777412..c6322ef47d 100644
--- a/chips/lpc55/memory.toml
+++ b/chips/lpc55/memory.toml
@@ -93,7 +93,7 @@
 execute = false
 dma = true
 
-[[override]]
+[[transient_override]]
 name = "a"
 address = 0x2003ffe0
 size = 32
@@ -101,7 +101,7 @@
 write = true
 execute = false
 
-[[override]]
+[[transient_override]]
 name = "b"
 address = 0x2003ffe0
 size = 32
diff --git a/drv/lpc55-update-server/src/main.rs b/drv/lpc55-update-server/src/main.rs
index 8f78b91c42..3cc2898d46 100644
--- a/drv/lpc55-update-server/src/main.rs
+++ b/drv/lpc55-update-server/src/main.rs
@@ -41,10 +41,6 @@
 #[link_section = ".bootstate"]
 static BOOTSTATE: MaybeUninit<[u8; 0x1000]> = MaybeUninit::uninit();
 
-#[used]
-#[link_section = ".transient_override"]
-static mut TRANSIENT_OVERRIDE: MaybeUninit<[u8; 32]> = MaybeUninit::uninit();
-
 #[derive(Copy, Clone, PartialEq)]
 enum UpdateState {
     NoUpdate,
@@ -1416,22 +1412,39 @@
     RotBootStateV2::load_from_addr(addr)
 }
 
+extern "C" {
+    // Symbols injected by the linker.
+    //
+    // This requires adding `extern-regions = ["transient_override"]` to the
+    // task config
+    pub static mut __REGION_TRANSIENT_OVERRIDE_BASE: [u32; 0];
+}
+
 fn set_transient_override(preference: [u8; 32]) {
-    // Safety: Data is consumed by Bootleby on next boot.
-    // There are no concurrent writers possible.
-    // Calling this function multiple times is ok.
-    // Bootleby is careful to vet contents before acting.
-    unsafe {
-        TRANSIENT_OVERRIDE.write(preference);
+    // SAFETY: populated by the linker, getting the address is fine
+    let override_addr =
+        unsafe { __REGION_TRANSIENT_OVERRIDE_BASE.as_ptr() } as *mut u8;
+    for (i, p) in preference.iter().enumerate() {
+        // SAFETY: this points to a valid region of RAM that is otherwise unused
+        // by Rust, so we can write to it.
+        unsafe {
+            core::ptr::write_volatile(override_addr.wrapping_add(i), *p);
+        }
     }
 }
 
 fn get_transient_override() -> [u8; 32] {
-    // Safety: Data is consumed by Bootleby on next boot.
-    // There are no concurrent writers possible.
-    // Bootleby consumes and resets TRANSIENT_OVERRIDE.
-    // The client may be verifying state set during update flows.
-    unsafe { TRANSIENT_OVERRIDE.assume_init() }
+    // SAFETY: populated by the linker, getting the address is fine
+    let override_addr =
+        unsafe { __REGION_TRANSIENT_OVERRIDE_BASE.as_ptr() } as *const u8;
+    let mut out = [0u8; 32];
+    for (i, p) in out.iter_mut().enumerate() {
+        // SAFETY: values are guaranteed to be initialized by stage0
+        unsafe {
+            *p = core::ptr::read_volatile(override_addr.wrapping_add(i));
+        }
+    }
+    out
 }
 
 // Preference constants are taken from bootleby:src/lib.rs
```

drv/lpc55-update-server/src/main.rs
lzrd: This is for the case where a previous update attempt has been abandoned or failed and left a transient preference that the new session does not want. We want to enable resetting an existing transient preference.

lzrd: This semantic came out of a fault insertion test where this type of abandoned session was simulated.

labbott: This still seems non ideal for the developer user case. I brought up not clearing the transient setting without some indication before https://github.com/oxidecomputer/hubris/pull/2050#discussion_r2068562888 albeit in a slightly different place.  I'd rather have an explicit way to clear the transient setting if we want that behavior.  This would require MGS changes but I'd argue this is a gap in how the original API was designed.

drv/lpc55-update-server/src/main.rs
```diff
@@ -905,10 +945,32 @@ impl ServerImpl<'_> {
...
+                };
+
+                // If the requested slot is the same as the active slot,
+                // treat this as a request to CLEAR the transient override.
+                //
+                // If we did allow setting the active slot as the transient
+                // preference, then we get this confusing 2-boot scenario when
+                // pending persistent preference is also used:
+                // the next reset returns us to the original image and the 2nd
+                // reset has us running the new/alternate image.
```
labbott: This statement seems strange in contrast to "This is for the case where a previous update attempt has been abandoned or failed and left a transient preference that the new session does not want" from above. If we are running active slot `A` and alternate slot is `B`, an incomplete/aborted update will always be targeting `B`. Why do we have a persistent setting for `B` if the update was aborted?

lzrd:
> The use of `MaybeUninit` here seems sketchy; it's only defined in terms of Rust behavior, so using `assume_init` to grab data that happens to be in RAM could be fragile.
> 
> (This is admittedly not a new problem introduced in the PR, but the PR adds another case of it)
> 
> I think the right way to handle this is using an `extern static`, although it's still a subject of active discussion ([rust-lang/unsafe-code-guidelines#397 (comment)](https://github.com/rust-lang/unsafe-code-guidelines/issues/397#issuecomment-2619633222))
> 
> Here's an (untested) patch which does just that:
> 
> ```diff
> diff --git a/app/lpc55xpresso/app.toml b/app/lpc55xpresso/app.toml
> index cc03b0de53..48b479bacb 100644
> --- a/app/lpc55xpresso/app.toml
> +++ b/app/lpc55xpresso/app.toml
> @@ -51,7 +51,8 @@
>  max-sizes = {flash = 26720, ram = 16704}
>  stacksize = 8192
>  start = true
> -sections = {bootstate = "usbsram", transient_override = "override"}
> +sections = {bootstate = "usbsram"}
> +extern-regions = ["transient_override"]
>  uses = ["flash_controller", "hash_crypt"]
>  notifications = ["flash-irq", "hashcrypt-irq"]
>  interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
> diff --git a/app/oxide-rot-1/app-dev.toml b/app/oxide-rot-1/app-dev.toml
> index 50a945c016..9d45ad973f 100644
> --- a/app/oxide-rot-1/app-dev.toml
> +++ b/app/oxide-rot-1/app-dev.toml
> @@ -55,7 +55,8 @@
>  priority = 3
>  stacksize = 8192
>  start = true
> -sections = {bootstate = "usbsram", transient_override = "override"}
> +sections = {bootstate = "usbsram"}
> +extern-regions = ["transient_override"]
>  uses = ["flash_controller", "hash_crypt"]
>  notifications = ["flash-irq", "hashcrypt-irq"]
>  interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
> diff --git a/app/oxide-rot-1/app.toml b/app/oxide-rot-1/app.toml
> index 9e12953bad..0c45eb254d 100644
> --- a/app/oxide-rot-1/app.toml
> +++ b/app/oxide-rot-1/app.toml
> @@ -43,7 +43,8 @@
>  priority = 3
>  stacksize = 8192
>  start = true
> -sections = {bootstate = "usbsram", transient_override = "override"}
> +sections = {bootstate = "usbsram"}
> +extern-regions = ["transient_override"]
>  uses = ["flash_controller", "hash_crypt"]
>  notifications = ["flash-irq", "hashcrypt-irq"]
>  interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
> diff --git a/app/rot-carrier/app.toml b/app/rot-carrier/app.toml
> index 3072a96e1c..74c20d7d64 100644
> --- a/app/rot-carrier/app.toml
> +++ b/app/rot-carrier/app.toml
> @@ -43,7 +43,8 @@
>  # TODO size this appropriately
>  stacksize = 8192
>  start = true
> -sections = {bootstate = "usbsram", transient_override = "override"}
> +sections = {bootstate = "usbsram"}
> +extern-regions = ["transient_override"]
>  uses = ["flash_controller", "hash_crypt"]
>  notifications = ["flash-irq", "hashcrypt-irq"]
>  interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
> diff --git a/chips/lpc55/memory.toml b/chips/lpc55/memory.toml
> index 0125777412..c6322ef47d 100644
> --- a/chips/lpc55/memory.toml
> +++ b/chips/lpc55/memory.toml
> @@ -93,7 +93,7 @@
>  execute = false
>  dma = true
>  
> -[[override]]
> +[[transient_override]]
>  name = "a"
>  address = 0x2003ffe0
>  size = 32
> @@ -101,7 +101,7 @@
>  write = true
>  execute = false
>  
> -[[override]]
> +[[transient_override]]
>  name = "b"
>  address = 0x2003ffe0
>  size = 32
> diff --git a/drv/lpc55-update-server/src/main.rs b/drv/lpc55-update-server/src/main.rs
> index 8f78b91c42..3cc2898d46 100644
> --- a/drv/lpc55-update-server/src/main.rs
> +++ b/drv/lpc55-update-server/src/main.rs
> @@ -41,10 +41,6 @@
>  #[link_section = ".bootstate"]
>  static BOOTSTATE: MaybeUninit<[u8; 0x1000]> = MaybeUninit::uninit();
>  
> -#[used]
> -#[link_section = ".transient_override"]
> -static mut TRANSIENT_OVERRIDE: MaybeUninit<[u8; 32]> = MaybeUninit::uninit();
> -
>  #[derive(Copy, Clone, PartialEq)]
>  enum UpdateState {
>      NoUpdate,
> @@ -1416,22 +1412,39 @@
>      RotBootStateV2::load_from_addr(addr)
>  }
>  
> +extern "C" {
> +    // Symbols injected by the linker.
> +    //
> +    // This requires adding `extern-regions = ["transient_override"]` to the
> +    // task config
> +    pub static mut __REGION_TRANSIENT_OVERRIDE_BASE: [u32; 0];
> +}
> +
>  fn set_transient_override(preference: [u8; 32]) {
> -    // Safety: Data is consumed by Bootleby on next boot.
> -    // There are no concurrent writers possible.
> -    // Calling this function multiple times is ok.
> -    // Bootleby is careful to vet contents before acting.
> -    unsafe {
> -        TRANSIENT_OVERRIDE.write(preference);
> +    // SAFETY: populated by the linker, getting the address is fine
> +    let override_addr =
> +        unsafe { __REGION_TRANSIENT_OVERRIDE_BASE.as_ptr() } as *mut u8;
> +    for (i, p) in preference.iter().enumerate() {
> +        // SAFETY: this points to a valid region of RAM that is otherwise unused
> +        // by Rust, so we can write to it.
> +        unsafe {
> +            core::ptr::write_volatile(override_addr.wrapping_add(i), *p);
> +        }
>      }
>  }
>  
>  fn get_transient_override() -> [u8; 32] {
> -    // Safety: Data is consumed by Bootleby on next boot.
> -    // There are no concurrent writers possible.
> -    // Bootleby consumes and resets TRANSIENT_OVERRIDE.
> -    // The client may be verifying state set during update flows.
> -    unsafe { TRANSIENT_OVERRIDE.assume_init() }
> +    // SAFETY: populated by the linker, getting the address is fine
> +    let override_addr =
> +        unsafe { __REGION_TRANSIENT_OVERRIDE_BASE.as_ptr() } as *const u8;
> +    let mut out = [0u8; 32];
> +    for (i, p) in out.iter_mut().enumerate() {
> +        // SAFETY: values are guaranteed to be initialized by stage0
> +        unsafe {
> +            *p = core::ptr::read_volatile(override_addr.wrapping_add(i));
> +        }
> +    }
> +    out
>  }
>  
>  // Preference constants are taken from bootleby:src/lib.rs
> ```

Let's make this a separate PR to resolve all of these cases. Your comment copied to [issue #2110](https://github.com/oxidecomputer/hubris/issues/2110)

drv/lpc55-update-server/src/main.rs
lzrd: The current API is a bit deficient. The user may reasonably expect that A or B can be selected at any time and that since we are in an update flow, that that is enough. When you consider overall symmetry and failed update recovery, then a boot preference clearing operation is also needed.

drv/lpc55-update-server/src/main.rs
lzrd: MGS PR# 433 provides an explicit means to clear pending preferences using a new message. The implicit modifications are gone and multiple setting of conflicting or already existing state is now an error.

lzrd:
> The use of `MaybeUninit` here seems sketchy; it's only defined in terms of Rust behavior, so using `assume_init` to grab data that happens to be in RAM could be fragile.
> 
> (This is admittedly not a new problem introduced in the PR, but the PR adds another case of it)
> 
> I think the right way to handle this is using an `extern static`, although it's still a subject of active discussion ([rust-lang/unsafe-code-guidelines#397 (comment)](https://github.com/rust-lang/unsafe-code-guidelines/issues/397#issuecomment-2619633222))
> 
> Here's an (untested) patch which does just that:
> 
> ```diff
> diff --git a/app/lpc55xpresso/app.toml b/app/lpc55xpresso/app.toml
> index cc03b0de53..48b479bacb 100644
> --- a/app/lpc55xpresso/app.toml
> +++ b/app/lpc55xpresso/app.toml
> @@ -51,7 +51,8 @@
>  max-sizes = {flash = 26720, ram = 16704}
>  stacksize = 8192
>  start = true
> -sections = {bootstate = "usbsram", transient_override = "override"}
> +sections = {bootstate = "usbsram"}
> +extern-regions = ["transient_override"]
>  uses = ["flash_controller", "hash_crypt"]
>  notifications = ["flash-irq", "hashcrypt-irq"]
>  interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
> diff --git a/app/oxide-rot-1/app-dev.toml b/app/oxide-rot-1/app-dev.toml
> index 50a945c016..9d45ad973f 100644
> --- a/app/oxide-rot-1/app-dev.toml
> +++ b/app/oxide-rot-1/app-dev.toml
> @@ -55,7 +55,8 @@
>  priority = 3
>  stacksize = 8192
>  start = true
> -sections = {bootstate = "usbsram", transient_override = "override"}
> +sections = {bootstate = "usbsram"}
> +extern-regions = ["transient_override"]
>  uses = ["flash_controller", "hash_crypt"]
>  notifications = ["flash-irq", "hashcrypt-irq"]
>  interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
> diff --git a/app/oxide-rot-1/app.toml b/app/oxide-rot-1/app.toml
> index 9e12953bad..0c45eb254d 100644
> --- a/app/oxide-rot-1/app.toml
> +++ b/app/oxide-rot-1/app.toml
> @@ -43,7 +43,8 @@
>  priority = 3
>  stacksize = 8192
>  start = true
> -sections = {bootstate = "usbsram", transient_override = "override"}
> +sections = {bootstate = "usbsram"}
> +extern-regions = ["transient_override"]
>  uses = ["flash_controller", "hash_crypt"]
>  notifications = ["flash-irq", "hashcrypt-irq"]
>  interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
> diff --git a/app/rot-carrier/app.toml b/app/rot-carrier/app.toml
> index 3072a96e1c..74c20d7d64 100644
> --- a/app/rot-carrier/app.toml
> +++ b/app/rot-carrier/app.toml
> @@ -43,7 +43,8 @@
>  # TODO size this appropriately
>  stacksize = 8192
>  start = true
> -sections = {bootstate = "usbsram", transient_override = "override"}
> +sections = {bootstate = "usbsram"}
> +extern-regions = ["transient_override"]
>  uses = ["flash_controller", "hash_crypt"]
>  notifications = ["flash-irq", "hashcrypt-irq"]
>  interrupts = {"flash_controller.irq" = "flash-irq", "hash_crypt.irq" = "hashcrypt-irq"}
> diff --git a/chips/lpc55/memory.toml b/chips/lpc55/memory.toml
> index 0125777412..c6322ef47d 100644
> --- a/chips/lpc55/memory.toml
> +++ b/chips/lpc55/memory.toml
> @@ -93,7 +93,7 @@
>  execute = false
>  dma = true
>  
> -[[override]]
> +[[transient_override]]
>  name = "a"
>  address = 0x2003ffe0
>  size = 32
> @@ -101,7 +101,7 @@
>  write = true
>  execute = false
>  
> -[[override]]
> +[[transient_override]]
>  name = "b"
>  address = 0x2003ffe0
>  size = 32
> diff --git a/drv/lpc55-update-server/src/main.rs b/drv/lpc55-update-server/src/main.rs
> index 8f78b91c42..3cc2898d46 100644
> --- a/drv/lpc55-update-server/src/main.rs
> +++ b/drv/lpc55-update-server/src/main.rs
> @@ -41,10 +41,6 @@
>  #[link_section = ".bootstate"]
>  static BOOTSTATE: MaybeUninit<[u8; 0x1000]> = MaybeUninit::uninit();
>  
> -#[used]
> -#[link_section = ".transient_override"]
> -static mut TRANSIENT_OVERRIDE: MaybeUninit<[u8; 32]> = MaybeUninit::uninit();
> -
>  #[derive(Copy, Clone, PartialEq)]
>  enum UpdateState {
>      NoUpdate,
> @@ -1416,22 +1412,39 @@
>      RotBootStateV2::load_from_addr(addr)
>  }
>  
> +extern "C" {
> +    // Symbols injected by the linker.
> +    //
> +    // This requires adding `extern-regions = ["transient_override"]` to the
> +    // task config
> +    pub static mut __REGION_TRANSIENT_OVERRIDE_BASE: [u32; 0];
> +}
> +
>  fn set_transient_override(preference: [u8; 32]) {
> -    // Safety: Data is consumed by Bootleby on next boot.
> -    // There are no concurrent writers possible.
> -    // Calling this function multiple times is ok.
> -    // Bootleby is careful to vet contents before acting.
> -    unsafe {
> -        TRANSIENT_OVERRIDE.write(preference);
> +    // SAFETY: populated by the linker, getting the address is fine
> +    let override_addr =
> +        unsafe { __REGION_TRANSIENT_OVERRIDE_BASE.as_ptr() } as *mut u8;
> +    for (i, p) in preference.iter().enumerate() {
> +        // SAFETY: this points to a valid region of RAM that is otherwise unused
> +        // by Rust, so we can write to it.
> +        unsafe {
> +            core::ptr::write_volatile(override_addr.wrapping_add(i), *p);
> +        }
>      }
>  }
>  
>  fn get_transient_override() -> [u8; 32] {
> -    // Safety: Data is consumed by Bootleby on next boot.
> -    // There are no concurrent writers possible.
> -    // Bootleby consumes and resets TRANSIENT_OVERRIDE.
> -    // The client may be verifying state set during update flows.
> -    unsafe { TRANSIENT_OVERRIDE.assume_init() }
> +    // SAFETY: populated by the linker, getting the address is fine
> +    let override_addr =
> +        unsafe { __REGION_TRANSIENT_OVERRIDE_BASE.as_ptr() } as *const u8;
> +    let mut out = [0u8; 32];
> +    for (i, p) in out.iter_mut().enumerate() {
> +        // SAFETY: values are guaranteed to be initialized by stage0
> +        unsafe {
> +            *p = core::ptr::read_volatile(override_addr.wrapping_add(i));
> +        }
> +    }
> +    out
>  }
>  
>  // Preference constants are taken from bootleby:src/lib.rs
> ```

I've made similar changes to what you propose to make it compile. I kept the type as a pointer to a [u8;32] which lets the compiler use its favorite way to transfer the data.

drv/lpc55-update-server/src/main.rs
lzrd: I've added an explicit "cancel pending preference" message and a `-c` flag to cancel/clear a previous `-s` set flag in the faux-mgs CLI. There are errors for setting a preference that is already set or clearing a preference that is not set. This is much nicer than trying to wedge new behavior into the existing APIs.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -429,6 +429,52 @@ impl MgsCommon {
...
+                    return Err(GwSpError::RequestUnsupportedForComponent);
+                };
+                self.update_sp
+                    .set_pending_boot_slot(slot)
+                    .map_err(|err| GwSpError::UpdateFailed(err as u32))?;
+                Ok(())
+            }
+            // Other components might also be served someday.
+            _ => Err(GwSpError::RequestUnsupportedForComponent),
+        }
```
labbott, line 475: I don't think this is correct, this isn't doing a cancel, it's setting the slot which is not the behavior I would expect if I was using this API.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -662,6 +708,12 @@ impl MgsCommon {
             HIGHEST_KNOWN_ROT_VERSION
         );
 
+        // Force update if the MGS and ROT default mapping is not correct.
+        static_assertions::const_assert_eq!(
+            GwRotBootInfo::HIGHEST_KNOWN_VERSION - 1,
+            SpVersionedRotBootInfo::HIGHEST_KNOWN_VERSION,
+        );
+
```
labbott, line 716: It's not clear how this is related to to this change. If it's not  can you remove it?

task/control-plane-agent/src/mgs_compute_sled.rs
```diff
@@ -956,6 +956,29 @@ impl SpHandler for MgsHandler {
...
+            MgsMessage::ComponentCancelPendingActiveSlot {
+                component,
+                slot,
+                persist,
+            }
+        ));
+        match component {
+            SpComponent::HOST_CPU_BOOT_FLASH => {
+                Err(SpError::RequestUnsupportedForComponent)
+            }
```
labbott, line 975: It feels a little strange to have the check for `HOST_CPU_BOOT_FLASH` in all three specific mgs files since we also have the fallback in `mgs_common.rs` . I think this should probably just be removed.

drv/lpc55-update-server/src/main.rs
```diff
@@ -1027,6 +1100,71 @@ impl ServerImpl<'_> {
...
+                    if transient_selection == slot {
+                        set_hubris_transient_override(None);
+                        Ok(())
+                    } else {
+                        Err(UpdateError::InvalidPreferredSlotId.into())
+                    }
+                } else {
+                    Err(UpdateError::NonePending.into())
+                }
+            }
```
labbott, line 1127: I'm not a huge fan of requiring the exact slot for cancellation. I could see an argument that it's more robust but at the same time it's just as easy to make a mistake passing in an incorrect slot. I'd argue that it's safer to require fewer options vs having to handle your cancellation failing. I'd really prefer to just pass the component and the persistent option but I'm not going to block the PR over it.

drv/lpc55-update-server/src/main.rs
```diff
@@ -1027,6 +1100,71 @@ impl ServerImpl<'_> {
 
         Ok(())
     }
+
+    /// Cancel pending transient or persistent Hubris image selection.
+    fn switch_default_hubris_image_cancel_pending(
```
labbott, line 1105: This name confused me because it sounds like it's switching the default image, not clearing the pending. Maybe `fn clear_pending_image_switch`?

drv/lpc55-update-server/src/main.rs
```diff
@@ -529,6 +574,23 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+        slot: SlotId,
+        duration: SwitchDuration,
+    ) -> Result<(), RequestError<UpdateError>> {
+        match component {
+            RotComponent::Hubris => {
+                self.switch_default_hubris_image_cancel_pending(slot, duration)
+            }
+            RotComponent::Stage0 => {
+                Err(UpdateError::InvalidSlotIdForOperation.into())
+            }
```
labbott, line 591: Hmmm, It might be cleaner to just error out in `mgs_common.rs` because the issue is the component, not the Slot ID.

drv/lpc55-update-server/src/main.rs
```diff
@@ -905,12 +965,25 @@ impl ServerImpl<'_> {
         slot: SlotId,
         duration: SwitchDuration,
     ) -> Result<(), RequestError<UpdateError>> {
+        // Note: Rollback policy will be checking epoch values before activating.
+
         match duration {
             SwitchDuration::Once => {
-                // TODO deposit command token into buffer
-                return Err(UpdateError::NotImplemented.into());
+                if get_hubris_transient_override().is_some() {
```
labbott, line 972: Can we use `boot_preferences` here for consistency?

drv/lpc55-update-server/src/main.rs
```diff
@@ -495,8 +503,45 @@ impl idl::InOrderUpdateImpl for ServerImpl<'_> {
...
+                //
+                // After the update, the boot image selection will be based on:
+                //   - there being only one properly signed image, or
+                //   - transient boot selection (highest priority), or
+                //   - pending persistent selection (altering the persistent
+                //     selection)
+                //   - persistent preference if neither of the above was used.
+
+                let (persistent, pending_persistent, transient) =
+                    self.boot_preferences()?;
```
labbott, line 527: I went back and forth on this but I think since we're now actually checking the transient and persistent settings more places we should make `fn boot_preferences` return a struct e.g.
```
struct BootPreferences{
  persistent: SlotId
  pending_persistent: Option<SlotId>
  transient: Option<SlotId>
}
fn boot_preferences() -> BootPreferences
```
That will make parts of it easier to review

drv/lpc55-update-server/src/main.rs
```diff
@@ -1027,6 +1100,71 @@ impl ServerImpl<'_> {
...
+        slot: SlotId,
+        duration: SwitchDuration,
+    ) -> Result<(), RequestError<UpdateError>> {
+        // Note: Rollback policy will be checking epoch values before activating.
+        match duration {
+            SwitchDuration::Once => {
+                // Is there a pending transient boot selection?
+                // If it matches the one being cancelled
+                if let Some(transient_selection) =
+                    get_hubris_transient_override()
```
labbott, line 1116: Same here `boot_preferences()`

PR #2053, opened 2025-04-30
"humility qspi -D" is sluggish on...everything else

hawkw:
_No description provided._

So, `humility qspi` and auxflash commands are Really Slow on Cosmo. It turns out that this is because Hiffy has a magical list of boards that are allowed to go fast, and we never put Cosmo on it. Boards on the go-faster list are allowed to use a much larger buffer for data transferred through Hiffy than boards not on the go faster list. 

Controlling this based on `target_board` seems quite unfortunate. Subsequent revisions of the same board need to be added to the list, and it's easy to forget to do that when there's a new board revision and regress hiffy performance again (c.f. #1498). Instead, let's control this via a Cargo feature, which can at least be set in the `base.toml` for a board and inherited by every subsequent revision of the board without having to remember to do that.

I had wanted to just always enable go-faster mode on STM32H7 but I let @cbiffle talk me out of it.

### DIFF
--- a/task/hiffy/Cargo.toml
+++ b/task/hiffy/Cargo.toml
@@ -55,7 +55,11 @@ h753 = ["stm32h7", "drv-stm32xx-i2c/h753", "build-i2c/h753"]
 g031 = ["stm32g0", "drv-stm32xx-i2c/g031", "build-i2c/g031"]
 g030 = ["stm32g0", "drv-stm32xx-i2c/g030", "build-i2c/g030"]
 micro = ["no-ipc-counters"]
-# enables large memory buffer to make hiffy not slow
+# enables large memory buffer to make data transfers through through hiffy (e.g.
+# `humility qspi` and such) not slow. this conflicts with the "micro" feature,
+# which makes the buffer smaller for memory constrained-targets, and is off by
+# default, because it makes the buffer much larger, and you may not need that if
+# you're not using `humility qspi` etc.
 turbo = []
 panic-messages = ["userlib/panic-messages"]
 spctrl = ["drv-sp-ctrl-api"]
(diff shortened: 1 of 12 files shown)


cbiffle:
You know we might want to add Cosmo here...

task/hiffy/Cargo.toml
```diff
@@ -55,6 +55,8 @@ h753 = ["stm32h7", "drv-stm32xx-i2c/h753", "build-i2c/h753"]
 g031 = ["stm32g0", "drv-stm32xx-i2c/g031", "build-i2c/g031"]
 g030 = ["stm32g0", "drv-stm32xx-i2c/g030", "build-i2c/g030"]
 micro = ["no-ipc-counters"]
+# enables large memory buffer to make hiffy not slow
```
aapoalas, line 58: **nitpick:** I assume this is a matter of expedience, but I'd still comment at least somewhere why this feature is not always on; being not-slow sounds like an immediate win after all. Presumably there is some downside/hardware feature/??? required for this to work that would be great to document instead of leaving it as silent institutional knowledge.

hawkw: it uses another order of magnitude more RAM for the buffer, you're right that the comment could be a bit more explicit about it

cbiffle (approved):

task/hiffy/Cargo.toml
```diff
@@ -55,6 +55,12 @@ h753 = ["stm32h7", "drv-stm32xx-i2c/h753", "build-i2c/h753"]
 g031 = ["stm32g0", "drv-stm32xx-i2c/g031", "build-i2c/g031"]
 g030 = ["stm32g0", "drv-stm32xx-i2c/g030", "build-i2c/g030"]
 micro = ["no-ipc-counters"]
+# enables large memory buffer to make data transfers through through hiffy (e.g.
```
aapoalas, line 58:
```suggestion
# enables large memory buffer to make data transfers through hiffy (e.g.
```

Mea culpa :(

PR #2054, opened 2025-05-01
Update to `zerocopy` 0.8

hawkw:
This branch updates our dependency on `zerocopy` from v0.6.x to v0.8.x.
The primary motivation for this change is that I had wanted to use
`zerocopy` v0.8's support for data-bearing `enum`s in the
`gateway-ereport-messages` crate I added in
oxidecomputer/management-gateway-service#370, and...I hadn't realized
how painful taking the `zerocopy` update in Hubris would be. :) But,
it's also just good to stay on top of new dependency versions
regardless.

This is a _very_ large change, since pretty much every place where we
derive or use `zerocopy`'s traits needed to be changed slightly to use
the new APIs, but for the most part, it's not actually that
*interesting*, so reviewing it should be pertty straightforward. The
main API changes that are worth noting are:

- `AsBytes` is now called `IntoBytes`, which was an easy update.
- All the `_from_prefix`/`_from_suffix` APIs now return the rest of the
  slice, which is a nice improvement --- previously we would manually
  split off the rest of the slice when using those functions.
- Conversions from bytes now return `Result`s instead of `Option`s,
  which very trivial changes in a few places.
- `LayoutVerified` is replaced by a new `Ref` type, but other API
  changes mean that you now basically never need to use it, which rocks!
- Some methods were renamed, which was also a pretty trivial
  find-and-replace.
- `zerocopy` adds new `Immutable` and `KnownLayout` marker traits.
  `Immutable` is required for `IntoBytes::as_bytes()`; types which do
  not derive it are now assumed to have interior mutability and only
  provide `IntoBytes::as_bytes_mut()`. So, basically everything now
  needs to derive `Immutable`. `KnownLayout` isn't required for APIs we
  use as commonly but I added it on everything anyway. This was most of
  what made this update annoying.
- `FromBytes::new_zeroed` is now provided by a new `FromZeros` trait,
  but the `FromBytes` derive implements that trait as well.

There may be some places where we can now make better use of new
`zerocopy` features. In particular, we can now use `zerocopy` on
data-bearing enums, which might allow us to replace `hubpack` with
`zerocopy` in several places. I didn't do that in this PR, but it's
worth looking into in follow-up changes --- I just wanted to get
everything building with the new API and felt that improving our usage
of it would be better off done in smaller commits.

One other important thing to note is that updating the
`gateway-messages` dependency increased stack depth in
`control-plane-agent` from 6000B to 6136B for non-sidecar targets, so I
bumped them up to 6256B. This, in turn, increases RAM to 65525B for
`control-plane-agent`, which exceeds the `max-sizes` config, per
@cbiffle's previous advice, I just deleted all the `max-sizes` for
`control-plane-agent` tasks.

Furthermore, this branch requires the following changes to other crates
to pick up the latest `zerocopy`:

- oxidecomputer/management-gateway-service#384
- oxidecomputer/idolatry#57
- oxidecomputer/humpty#8

Those ought to merge first so we can point our Git dependencies on those
repos back at the `main` branch.


Cargo.toml
```diff
@@ -145,10 +145,10 @@ dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-fe
 gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
 gimlet-inspector-protocol = { git = "https://github.com/oxidecomputer/gimlet-inspector-protocol", version = "0.1.0" }
 hif = { git = "https://github.com/oxidecomputer/hif", default-features = false }
-humpty = { git = "https://github.com/oxidecomputer/humpty", default-features = false, version = "0.1.3" }
+humpty = { git = "https://github.com/oxidecomputer/humpty", branch = "eliza/zerocopy-0.8", default-features = false, version = "0.1.3" }
 hubtools = { git = "https://github.com/oxidecomputer/hubtools", default-features = false, version = "0.4.6" }
-idol = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
-idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", default-features = false }
+idol = { git = "https://github.com/oxidecomputer/idolatry.git", branch = "eliza/zerocopy-0.8", default-features = false }
+idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", branch = "eliza/zerocopy-0.8", default-features = false }
```
labbott, line 151: Do you want to merge with that branch or are you going to rebase when those get approved?

labbott (commented):
Overall LGTM

cbiffle:
It looks like you've addressed the comments, so I'll second the LGTM. Do remember to squash. :-) (speaking as someone who sometimes forgets)

cbiffle (approved):

PR #2055, opened 2025-05-02
cosmo-hf: add limited busy-polls in all wait paths

hawkw:
Presently, the `cosmo-hf` driver has three functions that wait for some status to change before continuing: `wait_fpga_busy`, `wait_fpga_rx`, and `wait_flash_busy`. All of these functions check the desired condition in a loop and sleep for 1ms when it is not satisfied. Currently, `wait_fpga_busy` will first perform a limited number of busy-polls before starting to sleep between polls, but the other functions do not. While investigating HF write performance on Grapefruit,
@cbiffle and I found that commenting out the sleeps in these polling functions improved HF write performance significantly: writing 2MiB of random data using `humility qspi -D` went from taking about a minute to about 36 seconds --- that's a 40% performance improvement.

This branch is like a slightly more principled version of commenting out the `sleep_for` calls. I've generalized the pervious `wait_fpga_busy` of doing 32 busy-polls before starting to sleep to something that can be used by all three wait-for-condition type functions. I did change the code for implementing this a bit from the previous thing, which implemented the counter using a `for i in 0..` loop that checked `i > 32` This was a bit more concise and kind  of a cute trick, but had a somewhat surprising behavior: the loop wasn't actually infinite, and would terminate after `u32::MAX` polls. Of course, polling for 4,294,967,295ms means we would have waited about 50 days for a write to complete before terminating the loop unexpectedly, and if that happened, something would have to be REALLY broken. But, it's still a pretty wrong behavior, so the implementation I've done --- using a `while` loop with an explicitly incremented counter --- will never terminate unless the condition has actually changed, which seems a little less anxiety-inducing.


hawkw:
I don't have a Cosmo or Grapefruit readily available for testing this, so I've opened it as a draft for now. If @mkeeter or @cbiffle wants to give this a quick smoke test, I can un-draft it, otherwise I'll go into the office some time next week and try it out on a Grapefruit from the executive bathroom.

drv/cosmo-hf/src/main.rs
```diff
@@ -177,13 +162,47 @@ impl FlashDriver {
...
+        // When polling the FPGA or flash chips, this number of polls are
+        // attempted *without* sleeping between polls. If the FPGA/flash's
+        // status has not changed after this number of polls, the driver will
+        // begin to sleep for a short period between subsequent polls.
+        //
+        // This is intended to improve copy performance for operations where the
+        // desired status transition occurs in less than 1ms, avoiding a 1-2ms
+        // sleep and round-trip through the scheduler. status transitions
+        // quickly.
+        const MAX_BUSY_POLLS: u32 = 32;
```
hawkw, line 180: 32 may be more busy-polls than are *really* necessary, at least for `wait_fpga_rx`; @cbiffle and I saw it take about one poll before becoming ready pretty consistently, on Grapefruit. But, I figured we'd keep the current number for now.

drv/cosmo-hf/src/main.rs
```diff
@@ -177,13 +162,47 @@ impl FlashDriver {
         });
     }
 
+    /// Wait for a condition represented by the provided `poll` function.
+    ///
+    /// The driver will wait until the `poll` function returns `true`. Each time
+    /// `poll` returns `false`, the provided `trace` will be recorded in the
+    /// ring buffer.
+    fn poll_wait(&self, poll: fn(&Self) -> bool, trace: Trace) {
```
cbiffle, line 170: Have you looked at whether this gets inlined in practice?

If it does, you could make the function more general with no added code size implications by loosening the `poll` type to `impl FnMut(&self) -> bool`. That'd allow closures with mutable state.

If the function is not currently inlined then this would cause it to be duplicated, potentially at a small code size cost. Personally, I'd probably still do it unless it is clearly a serious problem, because taking general purpose closures is nice.

hawkw: Ah, yeah, good point. I used the function pointer as I wanted to avoid monomorphization in case it was *not* being inlined, and none of the _current_ code needed `FnMut`. But, it occurs to me that we could also mark the function as `#[inline(always)]`...

mkeeter (approved):
I tested reading, writing and hashing using `humility qspi`; everything seem fine.  For a 32 MiB image, it took 2 minutes to erase + 11 minutes to flash:

```console
humility: erased 32.00MB in 2 minutes
humility: flashed 32.00MB in 11 minutes
```

hawkw:
> I tested reading, writing and hashing using `humility qspi`; everything seem fine.  For a 32 MiB image, it took 2 minutes to erase + 11 minutes to flash:
> 
> 
> 
> ```console
> 
> humility: erased 32.00MB in 2 minutes
> 
> humility: flashed 32.00MB in 11 minutes
> 
> ```

@mkeeter While you're testing things, I'd be curious to hear how that compares to `master`. When @cbiffle and I were testing a less fancy version of this change on Grapefruit we saw a noticeable performance improvement versus `master`; IIRC a 2MiB write went from about 60 seconds to about 35 seconds. It would be nice to make sure that this change results in a similarly noticeable improvement, since that's kind of the point of making it...

mkeeter:
@hawkw it's now faster!

Here's the performance with old code:
> humility: erased 32.00MB in 2 minutes
> humility: flashed 32.00MB in 20 minutes

hawkw:
> Here's the performance with old code:

Yay okay great thank you for checking! That's about the delta I was hoping to see!

PR #2057, opened 2025-05-05
Improve swd errors

labbott:
`SpCtrlError::Fault` is not specific enough to determine errors in some cases. Specifically, if `swd_setup` fails there's not much to go on. Improve our error reporting everywhere by indicating if the error is a SWD Fault, Wait, or Protocol error. `swd_setup` gets dedicated errors since failures there are very tricky to triage on live systems.

Fixes hubris#2049


hawkw:
Huh, the "Fixes hubris#2049" in the PR description seems not to have been recognized by GitHub for some reason...maybe it's the "hubris" prefix. You may want to either change that or make sure to manually close the issue after this merges.

drv/lpc55-swd/src/main.rs
```diff
@@ -550,7 +571,10 @@ impl idl::InOrderSpCtrlImpl for ServerImpl {
         let r =
             Reg::from_u16(register).ok_or(SpCtrlError::InvalidCoreRegister)?;
         self.write_single_target_addr(DCRSR, r as u32)
-            .map_err(|_| SpCtrlError::Fault)?;
+            .map_err(|e| {
+                let err: SpCtrlError = e.into();
+                err
+            })?;
```
hawkw, line 577: nitpick, take it or leave it: could also be 
```suggestion
            .map_err(|e| SpCtrlError::from(e))?;
```

labbott:
> Huh, the "Fixes hubris#2049" in the PR description seems not to have been recognized by GitHub for some reason...maybe it's the "hubris" prefix. You may want to either change that or make sure to manually close the issue after this merges.

I can never remember the syntax and am reminded why I usually just drop full URLs in commit messages

drv/lpc55-swd/src/main.rs
```diff
@@ -560,15 +584,15 @@ impl idl::InOrderSpCtrlImpl for ServerImpl {
                         break;
                     }
                 }
-                Err(_) => {
-                    return Err(SpCtrlError::Fault.into());
+                Err(e) => {
+                    return Err(e.into());
```
hawkw, line 588: it occurs to me that now that the error case here is just to return `e.into()`, this whole loop could be simplified to just:
```rust
        loop {
            let dhcsr = self.dp_read_bitflags::<Dhcsr>()?;
            ringbuf_entry!(Trace::Dhcsr(dhcsr));
            if dhcsr.is_regrdy() {
                break;
            }
        }
```
but, that's a bigger diff than the previous code, so...up to you whether it's worth the change.

drv/lpc55-swd/src/main.rs
```diff
@@ -1534,7 +1561,7 @@ impl ServerImpl {
                     limit -= 1;
                     hl::sleep_for(1);
                 }
-                Err(_) => return Err(SpCtrlError::Fault),
+                Err(e) => return Err(e.into()),
```
hawkw, line 1564: similarly, if we don't mind the much bigger diff, the whole `match` could just be:
```rust
let dhcsr = self.dp_read_bitflags::<Dhcsr>()?;
if dhcsr.is_regrdy() {
    // Trace retries used
    if limit != RETRY_LIMIT {
        ringbuf_entry!(Trace::LimitRemaining(limit));
    }
    return Ok(());
}
if limit == 0 {
    ringbuf_entry!(Trace::LimitRemaining(limit));
    return Err(SpCtrlError::Fault);
}
limit -= 1;
hl::sleep_for(1);
```

hawkw (commented):
Some pretty insignificant nitpicks, feel free to ignore me.

drv/lpc55-swd/src/main.rs
labbott: Heh I think this is from an earlier pass where the compiler was refusing to do the conversion. I'll clean it up.

hawkw: it's also possible that i am just wrong! :)

drv/lpc55-swd/src/main.rs
labbott: Good suggestion on both paths

labbott: This inspired me to make similar changes elsewhere. I think it cleans up the code nicely!

lzrd:
This is very nice. This appears to give us a unique error code at every place it matters. Do you know if that is the case?

hawkw (approved):
looks good to me, dunno if you also wanna get Ben's eyes on it before merging?

drv/lpc55-swd/src/main.rs
```diff
@@ -1511,31 +1515,27 @@ impl ServerImpl {
...
-                Err(_) => return Err(SpCtrlError::Fault),
+                return Ok(());
             }
+            if limit == 0 {
+                ringbuf_entry!(Trace::LimitRemaining(limit));
+                return Err(SpCtrlError::Fault);
+            }
+            limit -= 1;
+            hl::sleep_for(1);
         }
```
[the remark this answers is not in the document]
labbott: I really like these suggestions, I think I'll aim to put them into a separate PR though.

labbott:
> This is very nice. This appears to give us a unique error code at every place it matters. Do you know if that is the case?

To the extent that it matters, yes. I think it's most important to get specific error information early in `swd_setup`. Everywhere else we now at least get details about a fault vs protocol vs wait. The rest of the APIs that are called are pretty limited so the goal here is to get enough information to be able to reason about the probable causes e.g. if we're seeing protocol errors this is a sign something has gone wrong with the connection and we probably need to do a hard cycle to reset.

Issue #2056, opened 2025-05-02
`app/lpc55xpresso/app-sprot.toml` is busted on `master`


hawkw:
Discovered while going through all the `app.toml`s and testing if they build with the `zerocopy` update as part of #2054, it looks like this TOML doesn't build (and isn't checked on CI). It seems the SWD driver is missing some notifications, I _assume_ adding them will fix it:
```
error: could not compile `drv-lpc55-swd` (bin "drv-lpc55-swd") due to 27 previous errors
Error: failed to build swd

Caused by:
    Missing notifications; do you need to add them to your TOML file?
    - swd is missing ["jtag-detect-irq", "sp-reset-irq", "timer"]
```

<details>
<summary>Full build output (lengthy)</summary>

```console
eliza@theseus ~/Code/oxide/hubris $ cargo xtask dist app/lpc55xpresso/app-sprot.toml
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.22s
     Running `target/debug/xtask dist app/lpc55xpresso/app-sprot.toml`
app.toml has changed; rebuilding all tasks
cleaning ["lpc55xpresso", "task-jefe", "task-hiffy", "task-idle", "lpc55-update-server", "drv-lpc55-syscon", "drv-lpc55-gpio", "drv-user-leds", "drv-lpc55-usart", "drv-lpc55-i2c", "drv-lpc55-rng", "task-pong", "drv-lpc55-swd", "task-dumper", "drv-lpc55-sprot-server", "task-attest"]
     Removed 223 files, 509.4MiB total
building crate task-jefe
   Compiling task-jefe v0.1.0 (/home/eliza/Code/oxide/hubris/task/jefe)
   Compiling hubris-num-tasks v0.1.0 (/home/eliza/Code/oxide/hubris/sys/num-tasks)
    Finished `release` profile [optimized + debuginfo] target(s) in 0.97s
target/thumbv8m.main-none-eabihf/release/task-jefe -> target/lpc55xpresso-sprot/dist/jefe.elf
building crate task-hiffy
   Compiling drv-lpc55-gpio-api v0.1.0 (/home/eliza/Code/oxide/hubris/drv/lpc55-gpio-api)
   Compiling hubris-num-tasks v0.1.0 (/home/eliza/Code/oxide/hubris/sys/num-tasks)
   Compiling task-hiffy v0.1.0 (/home/eliza/Code/oxide/hubris/task/hiffy)
[...]

PR #2059, opened 2025-05-07
Remove `app-sprot.toml`

labbott:
This was a setup for testing RoT to SP communication before we had many boards available. This has bit rotted so keeping it around doesn't seem useful.


labbott:
Fixes #2056

mkeeter (approved):

PR #2063, opened 2025-05-09
thermal: remove set_fan_pwm

Aaron-Hartwig:
I'm removing `set_fan_pwm` from the `thermal` task interface because its existence no longer serves a purpose. Fans can either be in `auto` control mode, where the thermal model decides the PWM, or `manual` control mode, which effectively hardcodes the PWM to the `initial_pwm` argument of `set_mode_manual`. I say "effectively hardcodes" because it calls the [`maintain_pwm`](https://github.com/oxidecomputer/hubris/blob/master/task/thermal/src/main.rs#L333) function each loop iteration. This means that when `set_fan_pwm` was setting the PWM value, it would just be overwritten the next iteration. In addition to that, we just don't control single fans really, so the fact that `set_fan_pwm` required a single fan to be specified just doesn't match reality any more.

An alternative path here would be to remove the `index` argument from `set_fan_pwm` and change its function to update the `last_pwm` value read by `maintain_pwm` each iteration.

### DIFF
--- a/task/thermal/src/control.rs
+++ b/task/thermal/src/control.rs
@@ -110,6 +110,7 @@ impl Fans<{ bsp::NUM_FANS }> {
     pub fn new() -> Self {
         Self([None; bsp::NUM_FANS])
     }
+    #[allow(dead_code)]
     pub fn len(&self) -> usize {
         self.0.len()
     }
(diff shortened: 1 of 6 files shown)


task/thermal/src/control.rs
```diff
@@ -110,6 +110,7 @@ impl Fans<{ bsp::NUM_FANS }> {
     pub fn new() -> Self {
         Self([None; bsp::NUM_FANS])
     }
+    #[allow(dead_code)]
```
mkeeter, line 113: Should we just delete this too?

mkeeter: Okay, I don't feel strongly about it in that case.

mkeeter (approved):

Aaron-Hartwig:
@mkeeter I opted to just have the BSPs point to `NUM_FANS` since they know what it is and the `Fans` type relies on `bsp::NUM_FANS` during initialization anyway. It lets us remove a few more lines and what was basically a little indirection.

PR #2064, opened 2025-05-11
cpu-seq: Legalize the identity transition

hawkw:
Currently, the `cpu-seq` API has the sequencer return an
`IllegalTransition` error in the case where the requested power state is
the same as the current power state. This is also the error variant
returned when the requested state differs from the current state, but a
transition between those two states cannot be performed. This is
unfortunate, as the caller cannot distinguish between cases where the
system *is* in the requested state and everything is basically fine, and
cases where the caller asked the sequencer to do something that will
never succeed.

This commit changes the `cpu-seq` Idol API to return a `Transition` type
from the `set_power_state` and `set_power_state_with_reason` IPCs that
indicates whether or not a power state change occurred. Sequencer
implementations are changed to return `Ok(Transition::NoChange)` instead
of `Err(IllegalTransition)` when no state change occurred because the
current and requested states are the same.

I considered alternatively having the `IllegalTransition` error return
the current state so that the caller can distinguish between illegal
transitions that are totally disallowed and cases where no transition
occurs. However, this requires changing the entire IPC from using
`zerocopy` to `hubpack`, or adding an empty byte of padding to every
*other* `SeqError` variant so that we can use the support for returning
`Complex` error types with `zerocopy` I added in
oxidecomputer/idolatry#59. That gets a bit trickier, as we also convert
the `SeqError` code into a `u32` to send it upstack, so I figured
keeping it as a C-like enum was better than adding logic to convert the
error into a `u32` error code rather than just `as` casting it. Also,
it felt a bit wrong for this to be an "error", in my opinion. Doing
nothing because we were already in the requested state feels more like a
different type of "success" to me...

This is a first step towards fixing
oxidecomputer/management-gateway-service#270. Fully fixing that will
also require adding a way to indicate to MGS whether a state transition
occurred or not, which also requires a change to the
`gateway-sp-messages` crate.


hubris.testout.0
```diff
@@ -0,0 +1,638 @@
+==== Test archive details
```
mkeeter, line 1: This looks accidentally committed?

task/host-sp-comms/src/main.rs
```diff
@@ -394,26 +402,40 @@ impl ServerImpl {
...
+                        state: PowerState::A2,
+                    });
+                    if reboot {
+                        self.reboot_state = Some(RebootState::WaitingForA2);
+                    }
+                    return;
+                }
+                Ok(Transition::NoChange) => {
+                    // We're already in A2.
+                    ringbuf_entry!(Trace::SetState {
```
mkeeter, line 419: Should this be `Trace::AlreadyInState`?

jgallagher (commented):
:heavy_plus_sign: to Matt's comments; LGTM otherwise.

hubris.testout.0
hawkw: Agh, my bad! Will remove

drv/grapefruit-seq-server/src/main.rs
```diff
@@ -96,7 +96,11 @@ impl ServerImpl {
             jefe: Jefe::from(JEFE.get_task_id()),
             sgpio: fmc_periph::Sgpio::new(loader.get_token()),
         };
-        server.set_state_impl(PowerState::A2);
+
+        // Note that we don't use `Self::set_state_impl` here, as that will
+        // first attempt to get the current power state from `jefe`, and we
+        // haven't set it yet!
+        server.jefe.set_state(PowerState::A2 as u32);
```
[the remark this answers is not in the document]
hawkw: No, it's just a consequence of the refactoring I did in 40a867b9be14b8cbfdbbcb3e10d4dad30cadbf85. Previously the mock sequencer implementations had a a `set_state_impl` function that just called `jefe.set_state`, and a `validate_state_change` function that checked the transition against the current state. I combined this into `set_state_impl` in 40a867b9be14b8cbfdbbcb3e10d4dad30cadbf85. I thought this was worth doing mostly becausethe need to determine whether or not a change is a no-op and only call the `jefe.set_state` IPC if it isn't a no-op (to accurately simulate the real sequencer's behavior WRT Jefe) made the IPC server's `set_power_state` and `set_power_state_with_reason` repeat a bit more logic and I thought centralizing it was better. On the other hand, having a `set_state_impl` that's called both in initialization and in the IPC methods was basically only abstracting over the cast from `PowerState` to `u32`, which seemed less important to have an abstraction for.

Honestly I think this might be explained a bit better by looking at that diff, rather than my rambly attempt to describe it.

drv/cpu-seq-api/src/lib.rs
```diff
@@ -73,6 +73,28 @@ pub enum StateChangeReason {
...
+    PartialEq,
+    IntoBytes,
+    Immutable,
+    KnownLayout,
+    Count,
+)]
+#[repr(u8)]
+pub enum Transition {
+    /// The sequencer successfully transitioned to the desired state.
+    Done = 1,
```
mkeeter, line 92: Nit: if we're deliberately skipping value 0, there should be a comment explaining why.

(I'm also not sure if it's necessary, since this isn't deriving `IdolError`)

hawkw: Yeah, I don't actually think the representation of this matters, since it's never niche-optimized...I can just remove the `= 1`.

hawkw: removed it: 6454bb2697e6da700a7165715d64a505a4dfd345

mkeeter (approved):
This looks reasonable to me!

I also tested `humility dashboard` and `humility hiffy -f Sequencer.set_state`, and they all work out of the box:
```console
➜  hubris jj:(nsn (empty)) h hiffy -c Sequencer.set_state -a state=A0
Sequencer.set_state() => NoChange
➜  hubris jj:(nsn (empty)) h hiffy -c Sequencer.set_state -a state=A2
Sequencer.set_state() => Done
➜  hubris jj:(nsn (empty)) h hiffy -c Sequencer.set_state -a state=A2
Sequencer.set_state() => NoChange
➜  hubris jj:(nsn (empty)) h hiffy -c Sequencer.set_state -a state=A1
Sequencer.set_state() => Err(IllegalTransition)
```

azaleacolburn:
:transgender_flag:

PR #2071, opened 2025-05-19
Improve gimlet qspi error reporting

labbott:
We currently ignore any errors when working with the QSPI driver. The biggest place this is noticable is if we happen to give a bad address for reading we will loop forever. Fix this up to actually handle errors.


drv/gimlet-hf-server/src/main.rs
```diff
@@ -211,8 +228,10 @@ impl ServerImpl {
 
     fn page_program_raw(&self, addr: u32, data: &[u8]) -> Result<(), HfError> {
         self.set_and_check_write_enable()?;
-        self.qspi.page_program(addr, data);
-        self.poll_for_write_complete(None);
+        self.qspi
+            .page_program(addr, data)
+            .map_err(|x| qspi_to_hf(x))?;
```
mkeeter, line 233: Nit: you can write this without the lambda, e.g. `map_err(qspi_to_hf)`

(same note applies in a bunch of other places in the PR)

drv/gimlet-hf-server/src/main.rs
```diff
@@ -108,7 +116,16 @@ fn main() -> ! {
...
-        qspi.read_id(&mut idbuf);
+        let result = qspi.read_id(&mut idbuf);
+        if result.is_err() {
+            // If we can't read the id there's a good chance nothing else is going to
+            // work. `panic` would probably just be a crash loop. looping forever
+            // is slightly more polite.
+            loop {
+                // We are dead now.
+                hl::sleep_for(1000);
+            }
```
hawkw, line 127: elsewhere, we have tasks that die permanently by calling `sys_recv_notification` with an empty notification mask, like this: https://github.com/oxidecomputer/hubris/blob/7a24b884ba1907c1d45432760ad189d1a606e4a2/drv/gimlet-seq-server/src/main.rs#L174-L184

i think this is a bit nicer as we don't get periodically woken up by the timer, remember that we are just looping on it, and go back to sleep. not that this case is particularly worth optimizing for, though.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -61,6 +61,14 @@ struct Config {
     pub clock: u8,
 }
 
+// There isn't a great crate to do `From` implementation so do this manually
```
hawkw, line 64: ah, i suppose the alternative is giving `drv-hf-api` a dependency on `drv_stm32h7_qspi`? which, yeah, seems sad...

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -258,6 +282,18 @@ impl Qspi {
...
+                return Err(QspiError::Timeout);
+            }
+            if self.reg.sr.read().tef().bit_is_set() {
+                self.reg.fcr.modify(|_, w| w.ctef().set_bit());
+                return Err(QspiError::TransferError);
+            }
+            // Make sure our errors are enabled
+            self.reg
+                .cr
+                .modify(|_, w| w.teie().set_bit().toie().set_bit());
```
mkeeter, line 296: Is it okay to leave these flags set all the time?

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -258,6 +282,18 @@ impl Qspi {
         // perform transfers.
         let mut out = out;
         while !out.is_empty() {
+            if self.reg.sr.read().tof().bit_is_set() {
```
mkeeter, line 285: How about reading `self.reg.sr` just once, then checking `tof`, `tef`, and `flevel` (below) on that value?

drv/gimlet-hf-server/src/main.rs
mkeeter, line 64: I have a _slight_ bias towards making `drv-hf-api` depend on `drv-stm32h7-qspi`, so that it can define the `From` implementation, but I don't feel strongly about it.

hawkw: I suppose the downside of that is that it probably makes `drv-hf-api` not compile for target MCUs other than STM32H7. But, we don't currently have any reason to implement this API on other MCUs, so :woman_shrugging:

labbott: We also use `drv-hf-api` on cosmo which is still an STM32H7 but does not use `drv-stm32h7-qspi` for host flash so it felt a little weird to add that there. I don't love any option here tbh.

mkeeter: The Industrial Strength version would be to add a `stm32` feature to `drv-hf-api` which then brings in the QSPI crate and adds the `From` implementation, but that's Even More Shenanigans 🤷🏻

drv/gimlet-hf-server/src/main.rs
mkeeter: IIRC Humility shows tasks in `sys_recv_notification(0)` as `(DEAD)`, so there's precedent for treating it as a canonically dead state.

drv/gimlet-hf-server/src/main.rs
hawkw: alternatively `QspiError` could go in some kinda `qspi-types` crate or something that both the `drv-stm32h7-qspi` and `drv-hf-api` crates depend on. but...that's a Lot...

drv/gimlet-hf-server/src/main.rs
labbott: I took the suggestion for `read_status`. Maybe it should apply elsehwere?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -381,8 +402,9 @@ impl ServerImpl {
     }
 
     fn set_and_check_write_enable(&self) -> Result<(), HfError> {
-        self.qspi.write_enable();
-        let status = self.qspi.read_status();
+        self.qspi.write_enable().map_err(|x| qspi_to_hf(x))?;
+        // FIXME
```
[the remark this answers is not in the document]
labbott: Removed

drv/auxflash-server/src/main.rs
```diff
@@ -46,13 +46,23 @@ impl<'a> TlvcRead for SlotReader<'a> {
         dest: &mut [u8],
     ) -> Result<(), TlvcReadError<Self::Error>> {
         let addr: u32 = self.base + u32::try_from(offset).unwrap_lite();
-        self.qspi.read_memory(addr, dest);
+        // The QSPI driver was updated to return errors but this wants to
+        // be infallible
+        self.qspi.read_memory(addr, dest).unwrap_lite();
```
mkeeter, line 51: Out of curiosity, did you try making `Self::Error` into `AuxFlashError` here (instead of `Infallible`)?

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -186,8 +186,19 @@ impl Qspi {
         // off the front.
         let mut data = data;
         while !data.is_empty() {
+            let sr = self.reg.sr.read();
+
+            if sr.tof().bit_is_set() {
+                self.reg.fcr.modify(|_, w| w.ctof().set_bit());
```
mkeeter, line 192: Do you need to clear interrupt flags here (and below) before returning the error?

drv/auxflash-server/src/main.rs
labbott: I did, it still seemed to want an `Infallible` but I also didn't dig into it too closely.

drv/stm32h7-qspi/src/lib.rs
labbott: I think I lost that in one of my changes but I also see a few more errors I may need to fix :upside_down_face:

drv/auxflash-server/src/main.rs
labbott: I was clearly not thinking when I tried this before and fixed it

mkeeter (approved):

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -168,8 +200,26 @@ impl Qspi {
...
+                self.disable_all_interrupts();
+                return Err(QspiError::Timeout);
+            }
+            if sr.tef().bit_is_set() {
+                self.reg.fcr.modify(|_, w| w.ctef().set_bit());
+                self.disable_all_interrupts();
+                return Err(QspiError::TransferError);
+            }
```
hawkw, line 214: It occurs to me that, rather than having all the error paths have to remember to explicitly disable IRQs, we could maybe wrap this in an inner function that we call and then match the `Result` from, so that there's one place where we clear the IRQs regardless of how we exited from the function. Not sure if this is worth the effort unless there's a possibility of adding additional error paths here in the future.

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -258,8 +316,24 @@ impl Qspi {
...
+            let sr = self.reg.sr.read();
+
+            if sr.tof().bit_is_set() {
+                self.reg.fcr.modify(|_, w| w.ctof().set_bit());
+                self.disable_all_interrupts();
+                return Err(QspiError::Timeout);
+            }
+            if sr.tef().bit_is_set() {
+                self.reg.fcr.modify(|_, w| w.ctef().set_bit());
+                self.disable_all_interrupts();
```
hawkw, line 328: similarly, is it worth refactoring this so that the error path disables interrupts in one place instead of multiple?

labbott: I'll give this a shot (and I deeply appreciate you having an eye for things like this!)

labbott:
I haven't forgotten about this but I'm doing some other work on HF/qspi and I want to avoid churning the code too much so I'm holding off a bit

PR #2073, opened 2025-05-20
Clear LM5066 faults on `power` startup

mkeeter:
The LM5066 has strong opinions on what error flags should be set on startup:

- `STATUS_BYTE` says the VIN fault flag is set by default, no comment about autoclearing; flags should be cleared with `CLEAR_FAULTS`
- `STATUS_WORD` says VIN fault flag is set by default, and is cleared when input voltage > UVLO threshold
- `STATUS_INPUT` says the VIN _warning_ flag is set by default (the VIN fault flag is 0 by default), and is cleared when input voltage > UVLO threshold

In practice, we see the fault flag set in all three of these registers, as well as `!PowerGood`.

This PR adds a `CLEAR_FAULTS` call to the `power` task startup, to persuade the power-good bit to actually reflect reality.  We don't expect the `power` task to restart, so this should be a one-off occurrence (and genuine faults later on will not be affected).

### DIFF
--- a/task/power/src/bsp/cosmo_a.rs
+++ b/task/power/src/bsp/cosmo_a.rs
@@ -8,7 +8,7 @@ use crate::{
     SensorId,
 };
 
-use drv_i2c_devices::max5970::*;
+use drv_i2c_devices::{lm5066::*, max5970::*};
 use ringbuf::*;
 use userlib::units::*;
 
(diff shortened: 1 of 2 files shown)


task/power/src/bsp/cosmo_a.rs
```diff
@@ -135,6 +135,12 @@ enum Trace {
         err: drv_i2c_api::ResponseCode,
     },
 
+    /// Calling `CLEAR_FAULTS` on the LM5066 failed
+    Lm5066ClearFaultsFailed {
+        index: usize,
+        err: drv_i2c_api::ResponseCode,
+    },
+
```
labbott, line 143: Should this be LM5066i (and a couple of other places)

mkeeter: That's done by the next PR in the stack (#2076); this one doesn't know about the alternate chip.

labbott: Ahh somehow I mixed up the order

PR #2083, opened 2025-06-05
Drop `sleep_for`

labbott:
It turns out the cost of sleeping for 1 tick is higher than just busy waiting. Just do a nop wait for now.

### DIFF
--- a/drv/stm32h7-hash/src/lib.rs
+++ b/drv/stm32h7-hash/src/lib.rs
@@ -110,9 +110,7 @@ impl Hash {
         self.block.iter_mut().for_each(|m| *m = 0);
         self.idx = 0;
         if self.is_busy() {
-            while self.is_busy() {
-                hl::sleep_for(1);
-            }
+            while self.is_busy() {}
         }
         unsafe {
             self.reg.cr.modify(|_, w| {
@@ -155,7 +153,6 @@ impl Hash {
             // XXX do i need to check DINIS? || !is_dinis_set() {
             while self.is_busy() {
                 // || !is_dinis_set() {
-                hl::sleep_for(1);
             }
         }
 
@@ -296,9 +293,7 @@ impl Hash {
             .modify(|_, w| w.dcie().set_bit().dinie().clear_bit());
 
         if self.is_busy() {
-            while self.is_busy() {
-                hl::sleep_for(1);
-            }
+            while self.is_busy() {}
         }
         sys_irq_control(self.interrupt, true);
         self.reg.str.modify(|_, w| w.dcal().set_bit());
@@ -306,16 +301,13 @@ impl Hash {
         // wait for calculation to finalize and interrupt
         loop {
             if self.is_busy() {
-                while self.is_busy() {
-                    hl::sleep_for(1);
-                }
+                while self.is_busy() {}
             }
             sys_recv_notification(self.interrupt);
             if self.reg.sr.read().dcis().bit() {
                 break;
             }
             sys_irq_control(self.interrupt, true); // XXX need this again?
-            hl::sleep_for(1);
         }
         // DCAL is supposedly not clearable by SW.
         // self.reg.str.modify(|_, w| w.dcal().clear_bit());


aapoalas (commented):
**suggestion:** You might want to consider using the [spin_loop](https://doc.rust-lang.org/beta/std/hint/fn.spin_loop.html) hint; it might not do anything on Cortex M, but maybe it does? Not sure; just a thought that popped to mind since the hint does seem appropriate to this use case.

labbott:
> **suggestion:** You might want to consider using the [spin_loop](https://doc.rust-lang.org/beta/core/hint/fn.spin_loop.html) hint; it might not do anything on Cortex M, but maybe it does? Not sure; just a thought that popped to mind since the hint does seem appropriate to this use case.

https://doc.rust-lang.org/stable/src/core/hint.rs.html#268-303 Doesn't look like it would do anything

aapoalas:
> https://doc.rust-lang.org/stable/src/core/hint.rs.html#268-303 Doesn't look like it would do anything

Would the `#[cfg(all(target_arch = "arm", target_feature = "v6"))]` not hit armv6 STM32 targets? Though perhaps those don't need the hash task?

But yeah, it's probably entirely meaningless in either case. I tried digging into the specs a bit since [someone](https://stackoverflow.com/a/42180091) suggested that using NOP is unsuitable for timing, but the results are pretty inconclusive: NOPs are often [[1]](https://developer.arm.com/documentation/ddi0333/h/programmer-s-model/additional-instructions/nop-compatible-hints?lang=en)[[2]](https://developer.arm.com/documentation/ddi0597/2025-03/Base-Instructions/NOP--No-Operation-?lang=en)[[3]](https://developer.arm.com/documentation/ddi0602/2025-03/Base-Instructions/NOP--No-operation-?lang=en) mentioned to not have any timing guarantees, but the [`__nop` intrinsic](https://developer.arm.com/documentation/dui0491/i/Compiler-specific-Features/--nop-intrinsic?lang=en) that `cortex_m::asm::nop` [seems to compile into](https://docs.rs/cortex-m/latest/src/cortex_m/asm.rs.html#34-36) _does_ seem to suggest it can be used for timing purposes inside loops. Mildly confusing.

In conclusion; it probably makes sense to not use any `spin_loop` hinting. My apologies for the noise.

labbott:
> > https://doc.rust-lang.org/stable/src/core/hint.rs.html#268-303 Doesn't look like it would do anything
> 
> Would the `#[cfg(all(target_arch = "arm", target_feature = "v6"))]` not hit armv6 STM32 targets? Though perhaps those don't need the hash task?
> 
> But yeah, it's probably entirely meaningless in either case. I tried digging into the specs a bit since [someone](https://stackoverflow.com/a/42180091) suggested that using NOP is unsuitable for timing, but the results are pretty inconclusive: NOPs are often [[1]](https://developer.arm.com/documentation/ddi0333/h/programmer-s-model/additional-instructions/nop-compatible-hints?lang=en)[[2]](https://developer.arm.com/documentation/ddi0597/2025-03/Base-Instructions/NOP--No-Operation-?lang=en)[[3]](https://developer.arm.com/documentation/ddi0602/2025-03/Base-Instructions/NOP--No-operation-?lang=en) mentioned to not have any timing guarantees, but the [`__nop` intrinsic](https://developer.arm.com/documentation/dui0491/i/Compiler-specific-Features/--nop-intrinsic?lang=en) that `cortex_m::asm::nop` [seems to compile into](https://docs.rs/cortex-m/latest/src/cortex_m/asm.rs.html#34-36) _does_ seem to suggest it can be used for timing purposes inside loops. Mildly confusing.
> 
> In conclusion; it probably makes sense to not use any `spin_loop` hinting. My apologies for the noise.

It's possible that the `v6` feature is additive so e.g. v8 would cover v6 + v7 etc. The `nop` also isn't actually used for timing, it's more a hint to the compiler that we are doing something and to not try to optimize the loop away (yes extra cautious about aggressive compilers!). If the CPU decides it doesn't get executed it's actually fine because we're still waiting for the block to finish up whatever it was doing independent of the CPU. I'm also not convinced of the value of trying to do "something else" anyway since we're not particularly concerned right now about going to low power states (common thing for yields to do).

labbott:
> FWIW, I think you can remove the `nop`. We tested compiler empty and infinite loop behavior on the last ~30 Rust releases during RustWeek, and the old behavior with empty loops appears to be long gone.
> 
> But leaving it in is also not unreasonable.
> 
> I'm not concerned about v6 support since this code is stm32h7-specific.

I went ahead and removed the `nop`s , good to hear we have more confidence in compilers!

PR #2084, opened 2025-06-05
Use hash directly

labbott:


cbiffle:
Hmmm. So I guess the reason _not_ to do this is if we think we're going to need the SHA block for anything else. (That's the reason it was a separate task to begin with.)

I certainly agree that we're not currently _using_ it for anything else, though. At least not in Hubris (I think endoscope uses it?).

I take it you also can't think of any use cases other than host flash?

labbott:
> Hmmm. So I guess the reason _not_ to do this is if we think we're going to need the SHA block for anything else. (That's the reason it was a separate task to begin with.)
> 
> I certainly agree that we're not currently _using_ it for anything else, though. At least not in Hubris (I think endoscope uses it?).
> 

Endoscope needs to use `sha3` instead of `sha2` (https://github.com/oxidecomputer/hubris/blob/master/lib/endoscope/src/main.rs#L75) 

> I take it you also can't think of any use cases other than host flash?

I guess as another way of phrasing the question: what features do we still need that would need the speed of the hash block vs just using a software hash? I strongly suspect the QSPI block is going to be the largest thing we're going to want to hash. It's tricky to optimize unknown features and I think we can make that a problem for future us (and you can even bring up this exact comment later!)

labbott:
It is also worth considering that getting rid of the hash task is a relatively small improvement (~1.5-2 seconds) and doing the change to use Quad SPI commands is going to give us a much bigger speed up.

labbott:
We decided against doing this for now. While this is technically faster, the relative speedup is not worth the cost.

PR #2085, opened 2025-06-05
Use faster QSPI commands

labbott:


drv/qspi-api/src/lib.rs
```diff
@@ -36,6 +37,12 @@ pub enum Command {
 
     BulkErase = 0xC7,
     SectorErase = 0xDC,
+    // 4-BYTE QUAD OUTPUT FAST READ
+    QuadRead = 0x6C,
+    // 4-BYTE DTR QUAD INPUT/OUTPUT FAST READ
+    // Does not work at the moment
```
mkeeter, line 43: Is this comment out of date now that we're at 33 MHz?

labbott: I never got it to work on the QSPI-let board so I never tried on an actual gimlet. I'll give it a shot but I'm not super optimistic.

labbott: Well I was wrong: this seems to work on the `gimlet-c` in paris. I do think going with DDR is even riskier than just regular quad read but I'll give this a shot a few more places.

labbott: Nope seems to be no good on a gimlet-e

mkeeter: Okay, I was just wondering about the comment – sounds like it's still correct!

mkeeter (approved):

PR #2086, opened 2025-06-09
Add Ignition flash driver to `cosmo` images

mkeeter:
This is a minimal flash driver, suitable for testing with `humility hiffy -c IgnitionFlash....`

It has somewhat been tested by Nathanael on a reworked Cosmo, but should be tested again to check that `Command::ReleaseFromDeepPowerDown` is working.

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -170,7 +170,6 @@ notifications = ["timer", "vcore"]
 [tasks.ignition_flash]
 name = "drv-ignition-flash"
 priority = 7
-max-sizes = {flash = 65536, ram = 8192 }
 stacksize = 1200
 start = true
 task-slots = ["sys", {spi_front = "spi3_driver"}, "spartan7_loader"]
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -167,6 +167,15 @@ task-slots = ["sys", "i2c_driver", {spi_front = "spi3_driver"}, "jefe", "packrat
 uses = ["mmio_sequencer", "mmio_info"]
 notifications = ["timer", "vcore"]
 
+[tasks.ignition_flash]
+name = "drv-ignition-flash"
+priority = 7
+max-sizes = {flash = 65536, ram = 8192 }
+stacksize = 1200
+start = true
+task-slots = ["sys", {spi_front = "spi3_driver"}, "spartan7_loader"]
+uses = ["mmio_sequencer"] # TODO make the mux a separate peripheral
+
 [tasks.spartan7_loader]
 name = "drv-spartan7-loader"
 features = ["h753"]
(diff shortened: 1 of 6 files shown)


app/cosmo/base.toml
```diff
@@ -167,6 +167,15 @@ task-slots = ["sys", "i2c_driver", {spi_front = "spi3_driver"}, "jefe", "packrat
 uses = ["mmio_sequencer", "mmio_info"]
 notifications = ["timer", "vcore"]
 
+[tasks.ignition_flash]
+name = "drv-ignition-flash"
+priority = 7
+max-sizes = {flash = 65536, ram = 8192 }
```
labbott: That's a lot of flash

drv/ignition-flash/src/main.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+#[export_name = "main"]
+fn main() -> ! {
+    let spi_front = drv_spi_api::Spi::from(SPI_FRONT.get_task_id());
+    let dev = spi_front.device(drv_spi_api::devices::MUX);
+    let (ping, pong) = mutable_statics::mutable_statics! {
+        static mut PING: [u8; 512] = [Default::default; _];
+        static mut PONG: [u8; 512] = [Default::default; _];
+    };
```
labbott, line 75: Why do we need the mutable statics here?

app/cosmo/base.toml
mkeeter: So it is!  I'm just going to let autosizing handle it (which brings us to 8192 bytes of flash, 4096 bytes of RAM)

drv/ignition-flash/src/main.rs
mkeeter: We need buffers to stage data for reading / writing; I'm using `mutable_statics` so that they're not on the stack, since they're kinda large.

labbott: 512*2 bytes doesn't seem that large to me but mutable statics also seem fine to me

nathanaelhuffman:
Power up command works great. This is good to go from my end!
```
➜  hubris git:(mkeeter/drv-ignition-flash) ✗ humility hiffy -c IgnitionFlash.select
IgnitionFlash.select() => ()
➜  hubris git:(mkeeter/drv-ignition-flash) ✗ humility hiffy -c IgnitionFlash.read_id
IgnitionFlash.read_id() => IgnitionFlash_read_id_REPLY { value: [ 0x10, 0x44, 0x0, 0x84, 0x2d, 0x9, 0x0, 0xa, 0xe3, 0xff, 0x22, 0x0, 0x15, 0x10, 0xf7, 0xc8,
 0xf4, 0x0, 0x0, 0x0 ] }
```

nathanaelhuffman (approved):
:shipit:

PR #2089, opened 2025-06-11
Async hash and read

labbott:


drv/auxflash-server/src/main.rs
```diff
@@ -197,20 +211,26 @@ impl ServerImpl {
...
+                .map_err(qspi_to_auxflash)?;
 
             // If we're at the start of a sector, erase it before we start
             // writing the copy.
             if write_addr % SECTOR_SIZE_BYTES == 0 {
                 self.set_and_check_write_enable()?;
-                self.qspi.sector_erase(write_addr as u32);
-                self.poll_for_write_complete(Some(1));
+                self.qspi
+                    .sector_erase(write_addr as u32)
```
aapoalas: **thought:** I wonder if it might be beneficial to set up some helper functions like

```rust
fn sector_erase(&(mut)self, write_addr: size) -> Result<(), AuxFlashError> {
  self.qspi.sector_erase(write_addr as u32).map_err(qspi_to_auxflash)
}
```
just to cut down on the repetition.

drv/cosmo-hf/src/hf.rs
```diff
@@ -30,8 +32,35 @@ const BONUS_SIZE_BYTES: u32 = 1024 * 1024 * 64;
...
+    Hashing {
+        dev: HfDevSelect,
+        addr: usize,
+        end: usize,
+    },
+    Uninitialized,
 }
 
+enum SlotHash {
+    // Nothing has requested a hash of this slot yet
```
aapoalas: **suggestion:** Doc comment maybe?

drv/cosmo-hf/src/hf.rs
```diff
@@ -238,6 +271,105 @@ impl ServerImpl {
             }
         }
     }
+
+    // This assumes `begin` and `end` have been bounds checked
```
aapoalas: **suggestion:** This sounds like it definitely should be a doc comment; should the method be `unsafe` as well since it has a precondition?

drv/cosmo-hf/src/hf.rs
```diff
@@ -238,6 +271,105 @@ impl ServerImpl {
...
+        match self.hash_state {
+            HashState::Hashing { dev, addr, end } => {
+                let step_size = BLOCK_STEP_SIZE;
+
+                let prev = self.dev;
+                self.set_dev(dev).unwrap();
+                // The only way we should get an error from this is if
+                // we somehow call update before we've initialized or
+                // after we've finished the hash.
+                self.hash_range_update(addr, addr + step_size).unwrap();
```
aapoalas: **question:** Should these be using `unwrap_lite()`?

Also, is it possible for `addr + step_size` to overflow with a range at the end of the memory range that's not quite aligned to block sizes? Presumably that's not possible, in which case this could perhaps be a unchecked add. Otherwise, a checked add might be in order.

Can also reuse the `addr + step_size` result below just to make sure the compiler doesn't accidentally miss the deduplication chance.

drv/mock-gimlet-hf-server/src/main.rs
```diff
@@ -189,6 +189,22 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
+    ) -> Result<(), RequestError<HfError>> {
+        Ok(())
+    }
+
+    fn get_cached_hash(
+        &mut self,
+        _: &RecvMessage,
+        _slot: HfDevSelect,
+    ) -> Result<[u8; SHA256_SZ], RequestError<HfError>> {
+        Ok(*b"mockmockmockmockmockmockmockmock")
```
aapoalas, line 219: **praise:** :awsm:

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -258,8 +340,22 @@ impl Qspi {
         // perform transfers.
         let mut out = out;
         while !out.is_empty() {
+            let sr = self.reg.sr.read();
```
aapoalas: **suggestion:** This timeout checking and error enabling ("teie-toie" sounds nice; musical!) repeats twice; would it make sense to deduplicate into a helper function similarly to `disable_all_interrupts`?

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -318,9 +414,17 @@ impl Qspi {
         }
 
         // Clean up by disabling our interrupt sources.
-        self.reg
-            .cr
-            .modify(|_, w| w.ftie().clear_bit().tcie().clear_bit());
+        self.reg.cr.modify(|_, w| {
```
aapoalas: **issue:** Should presumably be `self.disable_all_interrupts()`

aapoalas (commented):
Peanut gallery passing by again after a spell.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -655,26 +775,81 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
-                return Err(HfError::HashError.into());
-            }
+        self.hash_range_update(begin, end)?;
+        self.hash
+            .finalize_sha256()
+            .map_err(|_| HfError::HashError.into())
+    }
+
+    // This does a sha256 on the entire range _except_ sector0
+    // which is treted as `0xff`
```
aapoalas: **typo:**
```suggestion
    // which is treated as `0xff`
```

drv/gimlet-hf-server/src/main.rs
```diff
@@ -655,26 +775,81 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
+        Ok(())
+    }
+
+    fn get_cached_hash(
+        &mut self,
+        _: &RecvMessage,
+        dev: HfDevSelect,
+    ) -> Result<[u8; SHA256_SZ], RequestError<HfError>> {
+        match dev {
+            HfDevSelect::Flash0 => match self.cached_hash0 {
```
aapoalas: **suggestion:** This `SlotHash` => `Result<[u8; SHA256_SZ, RequestError<HfError>>` repeats 4 times over; would a `From` impl or such perhaps be warranted to deduplicate the code?

Cargo.toml
```diff
@@ -143,7 +143,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2", "deflate
 # Oxide forks and repos
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.4.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", branch = "read_host_flash", default-features = false, features = ["smoltcp"] }
```
cbiffle: Just to make it less likely that we forget this, this should get fixed up before merge.

drv/cosmo-hf/src/hf.rs
cbiffle: It doesn't look like misusing this can violate Rust safety, so I don't think `unsafe` is necessary.

As a reader it would be helpful to say _how_ these are bounds checked, like, are these sector numbers or byte indices, and are they bounds checked against the current flash chip or a portion of it, etc.

drv/cosmo-hf/src/hf.rs
```diff
@@ -238,6 +271,105 @@ impl ServerImpl {
...
+                .flash_read(
+                    // This unwrap is safe because we already checked the range
+                    // when building the initial `begin` address
+                    FlashAddr::new(addr as u32).unwrap_lite(),
+                    &mut &mut buf[..size],
+                )
+                .unwrap_lite();
+            if let Err(e) = self.hash.update(size as u32, &buf[..size]) {
+                ringbuf_entry!(Trace::HashUpdateError(e));
+                return Err(HfError::HashError);
```
cbiffle, line 273: (You don't necessarily need to change this in this PR, but I've been trying to notice cases where we're putting useful information in a ringbuf and not returning it, so that I can point out: we can return things in errors now, and that pattern is generally more useful when we're debugging over the network without full ringbuf access.)

drv/cosmo-hf/src/hf.rs
cbiffle: This calculation could technically overflow if the addr value is chosen to be very large, and the end value is chosen to be within 256 of u32::MAX (if I'm doing the math right). If this happens, we'll panic.

If we want to address this, we really need to maintain some invariants over the initial value of addr and step_size, I'm not sure altering the addition operation is the right approach.

I think this is a relatively low risk though.

cbiffle: (But, yes, these should probably be unwrap_lite)

drv/gimlet-hf-server/src/main.rs
```diff
@@ -655,26 +775,81 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
+                .update(self.block.len() as u32, &self.block)
+                .map_err(|_| RequestError::Runtime(HfError::HashError))?;
+        }
+
+        self.hash_state = HashState::Hashing {
+            dev,
+            addr: SECTOR_SIZE_BYTES,
+            end: self.capacity,
+        };
+        sys_set_timer(Some(sys_get_timer().now + 1), notifications::TIMER_MASK);
```
cbiffle: NB: `sys_set_timer_relative` provides a simpler API for this that _also_ bakes in avoiding a bounds check and panic (which can't happen in reality but the compiler doesn't know that). In case that's useful.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -432,6 +465,88 @@ impl ServerImpl {
...
+        Ok(())
+    }
+
+    fn step_hash(&mut self) {
+        match self.hash_state {
+            HashState::Hashing { dev, addr, end } => {
+                let step_size = BLOCK_STEP_SIZE;
+
+                let prev = self.dev_state;
+                self.set_dev(dev).unwrap();
```
cbiffle: I've gone and looked at other cases where we're unwrapping `set_dev`, and in each case it's asserted to be safe because we've checked that the flash is muxed to the SP. They could make this assertion because the code was straight-line and didn't potentially handle other IPCs asking to e.g. switch the mux back.

I don't think this property holds in the async case, so I think this is effectively an IPC/network controlled panic, unless I missed the interlock. We could do two different things to prevent this.

1. We could check for the Err result here and stop the hashing process because the mux has been switched. I'm a bit nervous about that since, _in theory,_ you could switch the mux back and forth between two calls to `step_hash` without us noticing. (Yeah, you'd have to do that in under a millisecond, but, still.)
2. We could block requests to switch the mux while a hash is in progress, by adding code to the `set_mux` path (and anywhere else that flips the mux). This appeals to me, since it ensures that the hash is hashing consistent stuff without interlopers.

Thoughts?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -432,6 +465,88 @@ impl ServerImpl {
...
+                            HfDevSelect::Flash0 => {
+                                self.cached_hash0 = SlotHash::Hash(v);
+                            }
+                            HfDevSelect::Flash1 => {
+                                self.cached_hash1 = SlotHash::Hash(v);
+                            }
+                        },
+                        Err(_) => (),
+                    },
+                    HashState::Hashing { .. } => sys_set_timer(
```
cbiffle: (consider `sys_set_timer_relative` here, it's shorter and cheaper)

drv/gimlet-hf-server/src/main.rs
```diff
@@ -432,6 +465,88 @@ impl ServerImpl {
...
+            HashState::Hashing { dev, addr, end } => {
+                let step_size = BLOCK_STEP_SIZE;
+
+                let prev = self.dev_state;
+                self.set_dev(dev).unwrap();
+                // The only way we should get an error from this is if
+                // we somehow call update before we've initialized or
+                // after we've finished the hash.
+                self.hash_range_update(addr, addr + step_size).unwrap();
+                self.set_dev(prev).unwrap(); // infallible if the earlier set_dev worked
```
cbiffle: Conceptually: why would we allow the dev to be switched to host while we're hashing? Wouldn't that potentially raise TOCTOU issues?

So do we need to be setting/restoring the mux setting each time we go through this, or should we maintain the invariant that if we're in a hashing state, we have set the mux to SP, and we'll restore it at the very end if at all?

drv/cosmo-hf/src/hf.rs
```diff
@@ -238,6 +271,105 @@ impl ServerImpl {
...
+        Ok(())
+    }
+
+    fn step_hash(&mut self) {
+        match self.hash_state {
+            HashState::Hashing { dev, addr, end } => {
+                let step_size = BLOCK_STEP_SIZE;
+
+                let prev = self.dev;
+                self.set_dev(dev).unwrap();
```
cbiffle, line 286: (See my comments on gimlet-hf's similar line, I think we need some more protection on this logic in an async world.)

drv/gimlet-hf-server/src/main.rs
cbiffle: My initial analysis is partially wrong. Spoke with @labbott about this directly and the problem is smaller than I'd anticipated, because the IPC handlers try to invalidate the cache where appropriate, which in turn prevents step_hash from doing anything useful (and panicking).

However, there's at least one path where this protection is incomplete rn: you can call `page_program_dev` with an invalid argument and arrange for it to call `set_dev`, then delegate to `page_program`, which rejects the argument and returns an error -- leaving the mux flipped.

We're poking at the code to see if there's a more central place to put cache invalidation to prevent this class of error.

drv/gimlet-hf-server/src/main.rs
cbiffle: (Laura's clarified the design intent here, which is that you can switch the mux all you want, but the hash should be invalidated if you do so. I think that means this save/restore code and its unwraps can be removed.)

cbiffle:
...Laura has also pointed out that I've confused `set_dev` and `set_mux` in several places, so, I may be entirely off base here. Wheeeeee

Re-checking.

mkeeter:
Should we put the hash data into a common place?  Right now, it looks like there's a decent amount of copypasta between `cosmo-hf` and `gimlet-hf`.

Something like
```rust
struct HashData {
    hash: drv_hash_api::Hash,
    cached_hash0: SlotHash,
    cached_hash1: SlotHash,
    state: HashState,
}

enum HashState {
    Done,
    Hashing {
        dev: HfDevSelect,
        addr: usize,
        end: usize,
    },
    Uninitialized,
}

enum SlotHash {
    // Nothing has requested a hash of this slot yet
    Uncalculated,
    // Something has modified this slot
    Recalculate,
    // In progress calculation
    HashInProgress,
    // A valid hash
    Hash([u8; SHA256_SZ]),
}
```

(then move more common code into methods on the new `HashData`)

labbott:
> Should we put the hash data into a common place? Right now, it looks like there's a decent amount of copypasta between `cosmo-hf` and `gimlet-hf`.
> 

I think that's reasonable. This started out as gimlet only then I realized I needed cosmo too. It does look like we can use the same approach in both places.

cbiffle:
Alright, so, there is an actual async state machine issue here, but it's not exactly what I described, as I'm paging this code back in.

The issue is that you can freely switch the _mux_ while a hash computation is in progress. (The mux is not the dev.) This is probably bad, since it implies potentially giving the host write access, or just corrupting the hash computation.

I currently feel like the best compromise might be:
- If the mux is flipped, cancel any _ongoing_ hash computation.
- But retain any previously computed result, since it represents a useful capture of the flash state at the point in time when it was computed.

This also suggests that we might want to reconsider aggressively invalidating the computed hash on other mutation operations -- since it's still an accurate description of the flash contents, even if (say) a page erase has happened since. I'm less convinced of this though, and think consistent invalidation is the safest option.

drv/cosmo-hf/src/hf.rs
```diff
@@ -30,8 +30,11 @@ const BONUS_SIZE_BYTES: u32 = 1024 * 1024 * 64;
 pub struct ServerImpl {
     pub drv: FlashDriver,
     pub dev: HfDevSelect,
+    hash: HashData,
 }
 
+const BLOCK_STEP_SIZE: usize = drv_hf_api::SECTOR_SIZE_BYTES;
```
mkeeter, line 39: Can you add a doc comment explaining what this is used for?

drv/cosmo-hf/src/hf.rs
```diff
@@ -238,6 +247,105 @@ impl ServerImpl {
             }
         }
     }
+
+    // This assumes `begin` and `end` have been bounds checked for overflow
+    // and against the flash clip bounds.
```
mkeeter:
```suggestion
    // and against the flash chip bounds.
```

drv/cosmo-hf/src/hf.rs
```diff
@@ -238,6 +247,105 @@ impl ServerImpl {
...
+                                self.hash.cached_hash1 = SlotHash::Hash(v);
+                            }
+                        },
+                        Err(_) => (),
+                    },
+                    HashState::Hashing { .. } => {
+                        set_timer_relative(1, notifications::TIMER_MASK);
+                    }
+                    HashState::Uninitialized => unreachable!(),
+                }
```
mkeeter: Vibes: it's weird to set the hash state, then immediately match on it.  Maybe move the finalization and `set_timer_relative` calls into the conditional above?

drv/cosmo-hf/src/hf.rs
```diff
@@ -238,6 +247,105 @@ impl ServerImpl {
...
+                    HashState::Done => match self.hash.task.finalize_sha256() {
+                        Ok(v) => match dev {
+                            HfDevSelect::Flash0 => {
+                                self.hash.cached_hash0 = SlotHash::Hash(v);
+                            }
+                            HfDevSelect::Flash1 => {
+                                self.hash.cached_hash1 = SlotHash::Hash(v);
+                            }
+                        },
+                        Err(_) => (),
```
mkeeter: Log the error in a ringbuf?

drv/cosmo-hf/src/hf.rs
```diff
@@ -503,6 +608,76 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         }
     }
 
+    // This does a sha256 on the entire range _except_ sector0
+    // which is treted as `0xff`. We don't hash sector0 because
```
mkeeter:
```suggestion
    // which is treated as `0xff`. We don't hash sector0 because
```

task/control-plane-agent/src/update/host_flash.rs
```diff
@@ -30,6 +45,35 @@ impl HostFlashUpdate {
...
+    }
+
+    pub(crate) fn read_page(
+        &self,
+        slot: u16,
+        addr: u32,
+        dest: &mut [u8],
+    ) -> Result<(), SpError> {
+        // It's tempting to disable access if an update is in progress but
+        // this is a debug feature so we need it handy at all times`
```
mkeeter:
```suggestion
        // this is a debug feature so we need it handy at all times
```

drv/cosmo-hf/src/hf.rs
```diff
@@ -503,6 +608,76 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         }
     }
 
+    // This does a sha256 on the entire range _except_ sector0
```
mkeeter: Indicate that this _starts_ a hash operation, rather then doing it immediately?

(also, is there a reason this is a regular comment versus a docstring?)

drv/gimlet-hf-server/src/main.rs
```diff
@@ -157,12 +165,17 @@ fn main() -> ! {
     let mut buffer = [0; idl::INCOMING_SIZE];
     let mut server = ServerImpl {
         qspi,
-        block: [0; 256],
         capacity: 1 << log2_capacity,
         mux_state: HfMuxState::SP,
         dev_state: HfDevSelect::Flash0,
         mux_select_pin: cfg.sp_host_mux_select,
         dev_select_pin: cfg.flash_dev_select,
+        hash: HashData {
```
mkeeter: Nit: this could be extracted into `HashData::new(HASH.get_task_id())`, since it's also initialized in `cosmo-hf`.

drv/cosmo-hf/src/hf.rs
```diff
@@ -554,11 +729,13 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
 
 impl NotificationHandler for ServerImpl {
     fn current_notification_mask(&self) -> u32 {
-        0
+        notifications::TIMER_MASK
```
mkeeter, line 728: Should we only request this notification if the hash is in progress?  I guess it's harmless to always have the bit active...

drv/gimlet-hf-server/src/main.rs
```diff
@@ -432,6 +454,94 @@ impl ServerImpl {
...
+        &mut self,
+        begin: usize,
+        end: usize,
+    ) -> Result<(), HfError> {
+        flash_block!(block);
+        for addr in (begin..end).step_by(block.len()) {
+            let size = if block.len() < (end - addr) {
+                block.len()
+            } else {
+                end - addr
```
mkeeter:
```suggestion
            let size = if block.len().min(end - addr);
```
?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -432,6 +454,94 @@ impl ServerImpl {
...
+                    HashState::Done
+                } else {
+                    HashState::Hashing {
+                        dev,
+                        addr: addr + step_size,
+                        end,
+                    }
+                };
+
+                match self.hash.state {
```
mkeeter: Same vibes here about moving this into the conditional above.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -432,6 +454,94 @@ impl ServerImpl {
...
+    fn invalidate_hash(&mut self) {
+        match self.dev_state {
+            HfDevSelect::Flash0 => {
+                self.hash.cached_hash0 = SlotHash::Recalculate;
+            }
+            HfDevSelect::Flash1 => {
+                self.hash.cached_hash1 = SlotHash::Recalculate;
+            }
+        }
+        self.hash.state = HashState::Uninitialized;
```
mkeeter: Vague question: is it weird to have one of the `cached_hashX` values be valid, but the overall hash state be `Uninitialized`?  It seems like that's possible if both hash slots are valid when we call `invalidate_hash`.

drv/gimlet-hf-server/src/main.rs
```diff
@@ -599,6 +725,10 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         }
 
         self.mux_state = state;
+        // This is probably slight over kill but caching is best effort
+        // and if we're swapping our mux back the host as probably
```
mkeeter:
```suggestion
        // and if we're swapping our mux back the host has probably
```

drv/cosmo-hf/src/hf.rs
```diff
@@ -503,6 +608,76 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
+        for _ in (0..SECTOR_SIZE_BYTES).step_by(buf.len()) {
+            self.hash
+                .task
+                .update(buf.len() as u32, &buf)
+                .map_err(|_| RequestError::Runtime(HfError::HashError))?;
+        }
+
+        self.hash.state = HashState::Hashing {
+            dev,
+            addr: drv_hf_api::SECTOR_SIZE_BYTES,
```
mkeeter, line 655: I think there's a mixup here between dev-relative addressing and absolute addressing.

Here, we always set the hash range as `SECTOR_SIZE_BYTES..SLOT_SIZE_BYTES`, and include a `dev`.  These are dev-relative addresses, not absolute addresses.

However, in `step_hash`, we call `hash_range_update` with these same address.  `hash_range_update` naively translates them to `FlashAddr` (i.e. a global address) with `FlashAddr::new`.  This seems wrong, because addresses with `HfDevSelect::Flash1` will incorrectly read from the flash0 address range.

(This is the whole point of `FlashAddr`; perhaps `FlashAddr::new` should be called `FlashAddr::i_promise_this_is_an_absolute_address`)

We should instead be using `Self::flash_addr_for(addr, dev)` to go from dev-relative to absolute addresses.

drv/cosmo-hf/src/hf.rs
mkeeter: (also true for `gimlet-hf` below)

drv/gimlet-hf-server/src/main.rs
labbott: Maybe `Uninitialized` would be better described as `NotRunning`?

drv/cosmo-hf/src/hf.rs
labbott: We don't have a ringbuf on `gimlet-hf` at the moment

drv/cosmo-hf/src/hf.rs
labbott: This seems to match what we do throughout hubris

mkeeter: Okay!

mkeeter (approved):

PR #2094, opened 2025-06-12
Add Cosmo SPD driver

mkeeter:
The SPD driver on Cosmo doesn't need to be an I2C target, but we still want to read data from the SPDs into Packrat – both to prove that it works, and (eventually) to report them to the host as inventory devices.  We also need SPD communication to read temperatures for the thermal loop.

To make life harder, SPD size changed between DDR4 and DDR5 (512 → 1024 bytes).  This PR modifies our existing Packrat implementation to make it generic.  There's a new `struct SpdData<const NUM_DIMMS: usize, const DATA_SIZE: usize>`, which is used for both Gimlet and Cosmo (with different values for `DATA_SIZE`).

Remaining tasks:
- [x] Test on Cosmo
- [x] Make sure this didn't break stuff on Gimlet
- [x] Add temperature sensing


mkeeter:
Testing on Gimlet looks reasonable (commit 7b2b45287ec0 versus `master`).

There are no task crashes, the Gimlet boots fine, and running `/usr/lib/fm/fmd/fmtopo -V` produces identical output, with the exception of temperature readings and specific `instance` numbers in the NVMe drive `io` group:

```
hc://:product-id=913-0000019:product-sn=BRM23230022:server-id=BRM23230022:chassis-id=BRM23230022:serial=060000
00000000000014EE81000BC2B1/chassis=0/bay=6/nvme=0/disk=0
  group: protocol                       version: 1   stability: Private/Private
    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM23230022:server-id=BRM23230022:chas
sis-id=BRM23230022:serial=06000000000000000014EE81000BC2B1/chassis=0/bay=6/nvme=0/disk=0
    FRU               fmri      hc://:product-id=913-0000019:product-sn=BRM23230022:server-id=BRM23230022:chassis-id=BRM23230022:serial=A079DFDD:part=WUS4C6432DSP3X3/chassis=0/bay=6/nvme=0
  group: authority                      version: 1   stability: Private/Private
    product-id        string    913-0000019
    product-sn        string    BRM23230022
    chassis-id        string    BRM23230022
    server-id         string    BRM23230022
  group: storage                        version: 1   stability: Private/Private
    logical-block-size uint64    0x200
    manufacturer      string    NVMe
    serial-number     string    A079DFDD
    firmware-revision string    R2210000
    model             string    WUS4C6432DSP3X3
    logical-disk      string    c2t0014EE81000BC2B1d0
  group: io                             version: 1   stability: Private/Private
    instance          uint32    0x4 <-- ⚠️ this changes!
    driver            string    blkdev
    module            fmri      mod:///mod-name=blkdev/mod-id=93
    devfs-path        string    /pci@0,0/pci1de,fff9@3,2/pci1b96,0@0/blkdev@w0014EE81000BC2B1,0
    phys-path         string[]  [ "/pci@0,0/pci1de,fff9@3,2/pci1b96,0@0/blkdev@w0014EE81000BC2B1,0" ]
```

drv/gimlet-seq-server/src/main.rs
```diff
@@ -1231,7 +1231,7 @@ fn read_spd_data_and_load_packrat(
                 spd.read_into(&mut tmp[chunk..])
             })?;
 
-            packrat.set_spd_eeprom(ndx, true, 0, &tmp);
+            packrat.set_spd_eeprom(ndx, spd::PAGE_SIZE, &tmp);
```
labbott, line 1234: This and the other change in the file look weird, or at least asymmetric. I think this is because the `bool` before switched which page it was and instead this is just an offset?

app/cosmo/base.toml
```diff
@@ -929,6 +939,7 @@ sensors = { temperature = 1 }
 removable = true
 refdes = "J46"
 
+
```
labbott: extra line

idl/cosmo-spd.idl
```diff
@@ -0,0 +1,14 @@
+// Cosmo SPD task API
+
+Interface(
```
labbott, line 5: I'm not quite sure what the purpose of this interface is, I think it's supposed to be a placeholder?

labbott (approved):
Light request to consider renaming `spd` -> `gimlet-spd` for consistency but otherwise LGTM with just a couple of questions to make sure I understand the architecture.

idl/cosmo-spd.idl
mkeeter: It's mostly for convenience, so that we can use the idol runtime for notifications / dispatch instead of writing that by hand.  I've added comments to that effect.

app/cosmo/base.toml
mkeeter: Yup, fixed.

drv/gimlet-seq-server/src/main.rs
mkeeter: Your interpretation is correct.  Previously, the `bool` selected either page 0 or page 1 (which were the only possible 256-byte pages).  Now, DDR5 DIMMs have _more_ pages, so we instead provide a global offset (which also subsumes the offset within the page).

mkeeter:
I'm going to push temperature sensing to a subsequent PR and bring this in, because everything else is working.

PR #2095, opened 2025-06-13
chore: Minor QSPI code deduplication

aapoalas:
Since this caught my eye in #2089 and was merged in #2071, I thought I'd put my hands where my mouth is and do the few deduplications that I thought might be considered useful or "pretty".

### Adjacent code error?

In [`drv/auxflash-api/src/lib.rs:191`](https://github.com/oxidecomputer/hubris/blob/master/drv/auxflash-api/src/lib.rs#L191) the call to `self.get_blob_by_tag(tag)` seems to be invalid; that function takes two parameters, `(slot, tag)`. I wonder if this is dead code or what's going on with that?

### DIFF
--- a/drv/stm32h7-qspi/src/lib.rs
+++ b/drv/stm32h7-qspi/src/lib.rs
@@ -289,7 +289,7 @@ impl Qspi {
             let fl = self.get_fifo_level()?;
 
             // How much space is in the FIFO?
-            let ffree = FIFO_SIZE - fl;
+            let ffree = FIFO_SIZE.saturating_sub(fl);
             if ffree >= FIFO_THRESH.min(data.len()) {
                 // Calculate the write size. Note that this may be bigger than
                 // the threshold used above above. We'll opportunistically


drv/stm32h7-qspi/src/lib.rs
```diff
@@ -268,15 +300,8 @@ impl Qspi {
...
-            .cr
-            .modify(|_, w| w.ftie().clear_bit().tcie().set_bit());
-        while self.transfer_not_complete() {
-            // Unmask our interrupt.
-            sys_irq_control(self.interrupt, true);
-            // And wait for it to arrive.
-            sys_recv_notification(self.interrupt);
-        }
+        self.wait_for_transfer_complete();
+
```
aapoalas, line 330: **note:** Here the following disabling of interrupts is only `FTIE` away from being equal to `disable_all_interrupts`, and `FTIE` is already disabled in `wait_for_transfer_coplete`. I'm tempted to reuse `disable_all_interrupts` below as well, even though it'd clear the `FTIE` bit unnecessary.

Any opinions on whether that'd be a good choice or not?

drv/auxflash-server/src/main.rs
```diff
@@ -148,7 +148,7 @@ impl ServerImpl {
         sleep: Option<u64>,
     ) -> Result<(), AuxFlashError> {
         loop {
-            let status = self.qspi.read_status().map_err(qspi_to_auxflash)?;
+            let status = self.read_qspi_status()?;
```
labbott: I admit I don't love this pattern of having an inner helper function that much because I find it harder to follow the code to find the location where it's actually doing things.

labbott (commented):
The diff stat on this is `+136 −93` i.e. a net increase of code. That's a hard sell for code deduplication.

aapoalas:
> The diff stat on this is `+136 −93` i.e. a net increase of code. That's a hard sell for code deduplication.

A good bit of that is coming from the added and/or copied comments, but that's absolutely true.

My thinking was that getting rid of superfluous/"not functionality wise critical" error conversions would make the code a bit less "noisy", and perhaps also improve generated code quality / icache usage. I didn't check the output though; I'll take a look at that to see if this actually has other upsides than just personal code esthetics :)

The changes in the STM32 QSPI driver code I _do_ think pulls its own weight, regardless.

drv/stm32h7-qspi/src/lib.rs
aapoalas: Oh, I just realised that below the other `wait_for_transfer_complete` ends precisely with a `disable_all_interrupts`. ie. Yes, it is okay and probably makes sense to use it here as well.

drv/auxflash-server/src/main.rs
aapoalas: Alright, after some assembly analysis it turned out these sorts of helper functions don't really compile well at all: I was hoping that the compiler would either choose to inline them or would choose to inline the `self.qspi.read_status` code into the `qspi_read_status` function for an in-total beneficial tradeoff.

This is not what happened: the helpers were not being inlined and the concrete calls were not being inlined into the helpers, so that's extra calls right there and extra code. Furthermore, the helpers were somehow really expensive? They were something like 30 or 60 bytes a pop, which seems ridiculous but that's what it was. Even further, for some reason the result of the original calls was being handled with the `uxtb` instruction but the result of the helpers was ending up using `uxtb.w` for an extra 2 bytes per instruction. So it was all around a bad job. Adding `#[inline(always)]` got it all back to the original state of course, but that's hardly a consolation since the only "benefit" at that point is an aesthetic one, which you admit to not loving.

(There's actually one singular `read_qspi_status` helper method of this sort in `gimlet-hf-server/src/main.rs`: should I remove it while here to align the overall coding style?)

aapoalas:
After deeper analysis, the helper methods were a total dud. The STM32 QSPI driver changes are less clear but a possible dud as well. Here're the numbers up front:

* `write_impl` changes from 306 bytes to 250 bytes. (measured on cosmo-a auxflash and gimlet-f hf)
* `read_impl` changes from 302 bytes to 246 bytes. (same sources; auxflash had a slightly smaller size both before and after, probably +/-0 difference)
* `get_fifo_level` is an added 60 bytes.
* `wait_for_transfer_complete` is an added 72 bytes.

In total, that's a 20 byte increase in assembly size. Binary size of cosmo-a auxflash increases by a whopping 2868 bytes. This is a loss in total, but perhaps it is acceptable given the actually negative number of code lines? Applying `#[inline(always)]` can of course be done to (mostly) return to status quo ante bellum (though there's still a binary size increase of 276 bytes for cosmo-a auxflash after this).

Final side note: `disable_all_interrupts` is actually being done unconditionally no matter how `read/write_impl_inner` ends; the source code could optionally be simplified to reflect this as well.

aapoalas:
@labbott Would you be willing to re-review this?

I removed the `.map_err(qspi_to_auxflash)` helper functions that I added as they did not improve code generation (using `#[inline(always)]` fixed this), and as per your previous review you didn't consider them helpful from a readability standpoint. I also tried to remove the singular helper function like this that already exists: this is the `fn read_qspi_status` in `gimlet-hf-server`'s `ServerImpl`. Removing that increased code size. (I assume adding such a helper to `auxflash-server`'s `ServerImpl` would reduce code size a little, since it has very much the same code structure as `gimlet-hf-server`.)

EDIT: I checked; adding the `read_qspi_status` helper to `auxflash` results in a 28 byte code size reduction.

The `get_fifo_level` and `wait_for_transfer_complete` helper functions that I added previously are now basically the only thing that this PR creates, and those are also set `#[inline(always)]` as without that, the diff resulted in a code size increase. I think they're still code structure wise worthwhile.

I additionally moved the `disable_all_interrupts` calls to the `read_impl` / `write_impl` functions; the Ok-path for the inner functions would end with a `disable_all_interrupts` call (or equivalent), and the outer function would check if the result is Err and call `disable_all_interrupts`; ie. both paths always end up disabling all interrupts. The compiler understood this, so this doesn't cause changes in the final executable with one exception: the `read_impl` grew by 4 bytes. I believe that this is because the code previously skipped disabling one interrupt due to knowing it's statically already, and I replaced that with the generic `disable_all_interrupts` to make the code "simpler" and match `write_impl_inner`'s Ok-path end.

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -293,18 +326,8 @@ impl Qspi {
...
-            .cr
-            .modify(|_, w| w.ftie().clear_bit().tcie().set_bit());
-        while self.transfer_not_complete() {
-            // Unmask our interrupt.
-            sys_irq_control(self.interrupt, true);
-            // And wait for it to arrive.
-            sys_recv_notification(self.interrupt);
-        }
-        self.reg.cr.modify(|_, w| {
-            w.tcie().clear_bit().teie().clear_bit().toie().clear_bit()
```
aapoalas, line 306: **note:** So here the code previously cleared just TCIE, TEIE, and TOIE. Now the Ok-path goes into `disable_all_interrupts`, clearing one more register which then causes a 4 byte increase in code size.

I would've actually thought the compiler could combine the error and success paths here but that doesn't seem to be the case; the 4 byte increase suggests that the Ok-path is independent of the Err-path, even though they both do the same thing.

PR #2096, opened 2025-06-13
Add cosmo dimms to host_sp_comms inventory

mkeeter:
(staged on top of #2094)

This adds DDR5 DIMMS to the `host_sp_comms` inventory reported over IPCC.

I'd like to do some refactoring to reduce stack allocation, because the `enum InventoryData` is now over 1 KiB, but this is reasonable for initial tests.

### DIFF
--- a/task/host-sp-comms/src/bsp/cosmo_a.rs
+++ b/task/host-sp-comms/src/bsp/cosmo_a.rs
@@ -43,7 +43,7 @@ impl ServerImpl {
     /// This function should only return an error if the index is invalid;
     /// in that case, our caller is responsible for encoding the error as
     /// ```
-    /// SpToHost::InventoryData{
+    /// SpToHost::InventoryData {
     ///     result: err
     ///     name: [0; u32],
     /// }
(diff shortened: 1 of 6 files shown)


task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -574,6 +577,93 @@ impl ServerImpl {
...
+                name[2] = b'0' + (i / 10);
+                name[3] = b'0' + (i % 10);
+            } else {
+                name[2] = b'0';
+                name[3] = b'0' + i;
+            }
+            name
+        };
+
+        const DIMM_SENSORS: [[SensorId; 2]; 12] = [
```
rmustacc: Looking at this it appears that you'll tell me what the sensor ID will always be; however, it may or may not actually be present and it's up to me to figure that out, right? At first I kind of expected parsing of byte 0xd4 of the SPD.

There will be other things like multiple PMICs and sensors from that in the future. I'm fine with that reality, but I'm not sure how confusing that'll be for other consumers.

task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -574,6 +577,93 @@ impl ServerImpl {
...
+            [
+                other_sensors::DIMM_A_FRONT_TEMPERATURE_SENSOR,
+                other_sensors::DIMM_A_BACK_TEMPERATURE_SENSOR,
```
rmustacc: Where do the specific sides come from here? My read of the DDR5 RDIMM base spec is that these aren't necessarily split like this. §5.1 Component Types and Placement says: "RCD and Temperature Sensors are placed on primary side of RDIMM and LRDIMM form factors except
the RDIMM 1Rx8 which has Temperature Sensors on the secondary side."

I think I would probably consider using the SPD nomenclature of TS0/TS1. Or at least, I would do that generally for anything we want to communicate upstack in hubris and allow something else to encode device-specific knowledge of where the sensors physically are.

task/host-sp-comms/src/bsp/cosmo_a.rs
mkeeter: If the device is absent, then we'll return `InventoryDataResult::DeviceAbsent` instead of an `InventoryData` result, same as Gimlet.

This table is of the **Hubris** sensor ID, i.e. how you'd associate a particular DIMM with temperature sensor data in MGS.  I've renamed it to `DIMM_TEMPERATURE_SENSORS` for clarity.  If the host wants to parse byte 0xd4, it also gets the full SPD contents, so it can do that itself 😄.  Still, we can definitely extend the `DimmDdr5Spd` with more data when we want to report other devices / sensors.

task/host-sp-comms/src/bsp/cosmo_a.rs
mkeeter: I think I got the front / back nomenclature from a Micron datasheet, but TS0 / TS1 makes sense to me.  I've fixed this back in #2094, which is where the sensors were first introduced.

task/host-sp-comms/src/main.rs
```diff
@@ -306,6 +312,12 @@ impl ServerImpl {
                 last_panic: [0; MAX_HOST_FAIL_MESSAGE_LEN],
                 etc_system: [0; MAX_ETC_SYSTEM_LEN],
                 dtrace_conf: [0; MAX_DTRACE_CONF_LEN],
+
+                // Default value for InventoryData
+                scratch: host_sp_messages::InventoryData::DimmSpd {
+                    id: [0u8; 512],
+                    temp_sensor: 0u32,
+                },
```
labbott, line 320: Would there ever be a case where we would see this value in a dump and confuse ourselves? Would it make more sense to add an official "default" value to the inventory?

mkeeter: For context, we need to allocate a global static `InventoryData` to (somewhat) reduce stack usage, since it's a gigantic struct.

I'm leaning towards *not* having a default value, but could be persuaded otherwise.

We will see the last value written to `scratch`  in a dump (since it's global), but I hope we won't confuse ourselves with it.  If we added a default value, the host would also need to be aware of it, and we'd have to reset to our default value after reading each inventory item (clearing the ~1 KiB of space).

Issue #1379, opened 2023-05-25
caboose_server does not initialize the returned buffer


lzrd:
`caboose_server` does not initialize the returned buffer which may return some permanents of previous queries.

```
slab:~/Oxide/src/hubris,lpc55xpresso$ for name in GITC BORD NAME VERS ; do echo $name: ; hum hiffy -c Caboose.get_key_by_tag -a name=$name -n 64; done 2>/dev/null
GITC:
Caboose.get_key_by_tag() => 0x2e
             \/  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
0x00000000 | 63 37 37 64 64 37 65 32 65 37 62 36 35 32 62 66 | c77dd7e2e7b652bf
0x00000010 | 61 30 65 38 31 63 37 35 37 33 64 32 33 65 31 38 | a0e81c7573d23e18
0x00000020 | 30 33 34 36 61 34 62 35 2d 64 69 72 74 79 00 00 | 0346a4b5-dirty..
0x00000030 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................
BORD:
Caboose.get_key_by_tag() => 0xf
             \/  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
0x00000000 | 6c 70 63 78 70 72 65 73 73 6f 35 35 73 36 39 66 | lpcxpresso55s69f
0x00000010 | 61 30 65 38 31 63 37 35 37 33 64 32 33 65 31 38 | a0e81c7573d23e18
0x00000020 | 30 33 34 36 61 34 62 35 2d 64 69 72 74 79 00 00 | 0346a4b5-dirty..
0x00000030 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................
NAME:
Caboose.get_key_by_tag() => 0xc
             \/  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
0x00000000 | 6c 70 63 35 35 78 70 72 65 73 73 6f 73 36 39 66 | lpc55xpressos69f
0x00000010 | 61 30 65 38 31 63 37 35 37 33 64 32 33 65 31 38 | a0e81c7573d23e18
0x00000020 | 30 33 34 36 61 34 62 35 2d 64 69 72 74 79 00 00 | 0346a4b5-dirty..
0x00000030 | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................
VERS:
[...]

PR #2099, opened 2025-06-17
Delete `caboose_reader`

labbott:
This was an early implementation of reading caboose that is only available on a few boards. It's known to be 'quirky' with humility so just delete it.


labbott:
https://github.com/oxidecomputer/hubris/issues/1379 inspiration for deletion

mkeeter (approved):

Issue #2075, opened 2025-05-20
LM5066 Current Scaling Incorrect


ericaasen:
On the LM5066 hot swap controllers on Cosmo, the current scaling is incorrect.  I suspect the values are being calculated based on a 5 milliohm current sense resistor, but this circuit uses a 7 milliohm sense resistor, so the readings are high by ~1.4x.

The "Reported Current" below is as reported by `humility sensors`

| Load Current | Reported Current |
|--------------|------------------|
| 1       | 1.4    |
| 2       | 2.82  |
| 3       | 4.24  |
| 3.5    | 4.94  |

There is also an ability to have the device average measurements for more accuracy/less noise:

![Image](https://github.com/user-attachments/assets/cc4d3182-803d-479d-a18a-cecbafdf95c1)

There is a convenient way of getting all of the averaged telemetry information at once using a block read command as well:

![Image](https://github.com/user-attachments/assets/058c2efb-f5f2-4cdb-802d-a3842b5185d6)

I would set the averaging to no more than 256, which gives us an update rate of ~4Hz, which should be plenty fast for our purposes.

ericaasen:
full pmbus dump for one HSC with a 1A load:

```
λ humility pmbus -r V54P5_FAN_CENTRAL
humility: attached via ST-Link V3
0x01 OPERATION                 0x80
0x19 CAPABILITY                0xb0
0x43 VOUT_UV_WARN_LIMIT        0x0000 = 0.523V
0x4f OT_FAULT_LIMIT            0x0fff = 255.938°C
0x51 OT_WARN_LIMIT             0x0fff = 255.938°C
0x57 VIN_OV_WARN_LIMIT         0x0fff = 89.536V
0x58 VIN_UV_WARN_LIMIT         0x0000 = 0.262V
0x5d IIN_OC_WARN_LIMIT         0x0fff
0x78 STATUS_BYTE               0x03
0x79 STATUS_WORD               0x0203
0x7a STATUS_VOUT               0x00
0x7c STATUS_INPUT              0x00
0x7d STATUS_TEMPERATURE        0x00
0x7e STATUS_CML                0x82
0x7f STATUS_OTHER              0x20
0x80 STATUS_MFR_SPECIFIC       0x80
0x86 READ_EIN                  0x2f 0x4f 0xc4 0xb1 0x9f 0x3a           |/O...:
0x88 READ_VIN                  0x0795 = 42.577V
0x89 READ_IIN                  0x0429
[...]

PR #2100, opened 2025-06-17
Enable LM5066I averaging

mkeeter:
Closes out #2075


task/power/src/bsp/cosmo_a.rs
```diff
@@ -363,6 +371,12 @@ impl State {
                     err: err.into(),
                 });
             }
+            if let Err(err) = p.enable_averaging(8) {
```
labbott, line 377: Was the choice of 8 just vibes?

mkeeter: Somewhat vibes, but chosen to be significantly faster than our actual sampling rate.  I've added a comment to that effect.

PR #2101, opened 2025-06-17
Fix sp-console UART corruption via FPGA change

nathanaelhuffman:
FPGA was too sloppy when generating baud rates. 
Also make front FPGA release spi output when not selected.


Aaron-Hartwig (approved):
@rmustacc FYI this will also pickup the post code counter fix (https://github.com/oxidecomputer/quartz/commit/27346c64c3c471e6f50168c8379b8fa3c1ebaa8b)

PR #2102, opened 2025-06-20
Add SPD temperature sensing

mkeeter:
Needs validation on actual hardware!

### DIFF
--- a/task/cosmo-spd/src/main.rs
+++ b/task/cosmo-spd/src/main.rs
@@ -15,8 +15,8 @@ use task_jefe_api::Jefe;
 use task_packrat_api::Packrat;
 use task_sensor_api::{config::other_sensors, NoData, Sensor, SensorId};
 use userlib::{
-    hl::sleep_for, sys_get_timer, sys_recv_notification, sys_set_timer,
-    task_slot, FromPrimitive, RecvMessage,
+    hl::sleep_for, set_timer_relative, sys_recv_notification, task_slot,
+    FromPrimitive, RecvMessage,
 };
 use zerocopy::IntoBytes;
 
@@ -146,12 +146,11 @@ fn main() -> ! {
 
     let sensor = Sensor::from(SENSOR.get_task_id());
     let mut server = ServerImpl {
-        deadline: 0u64,
         dimms,
         sensor,
         present,
     };
-    sys_set_timer(Some(0), notifications::TIMER_MASK);
+    set_timer_relative(0, notifications::TIMER_MASK);
     let mut buffer = [0; idl::INCOMING_SIZE];
 
     loop {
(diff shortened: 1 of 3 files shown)


task/cosmo-spd/src/main.rs
```diff
@@ -172,13 +180,159 @@ impl idl::InOrderCosmoSpdImpl for ServerImpl {
...
+                    let t = (raw_temp << 3) as i16;
+                    let temp_c = f32::from(t) * 0.0078125f32;
+
+                    // Send the value to the sensors task
+                    self.sensor.post_now(DIMM_SENSORS[index][pos], temp_c);
+                }
+            }
+            self.deadline = now + TIMER_INTERVAL;
+        }
+        sys_set_timer(Some(self.deadline), notifications::TIMER_MASK);
```
labbott: `set_timer_relative` ?

task/cosmo-spd/src/main.rs
```diff
@@ -172,13 +180,159 @@ impl idl::InOrderCosmoSpdImpl for ServerImpl {
...
+                }
+                if timed_out {
+                    None
+                } else {
+                    Some(self.dimms.$data.data())
+                }
+            }};
+        }
+        let now = sys_get_timer().now;
+        if now >= self.deadline {
```
labbott: It seems odd to check the deadline here in our notification handler if the timer is our only notification but it's been a while since I thought about hubris timer details

task/cosmo-spd/src/main.rs
mkeeter: 👍🏻

task/cosmo-spd/src/main.rs
mkeeter: I did some reading and believe that we don't need this conditional (or to track the deadline at all), so this has been simplified.

task/cosmo-spd/src/main.rs
```diff
@@ -171,13 +177,156 @@ impl idl::InOrderCosmoSpdImpl for ServerImpl {
...
+        // The FPGA register generation produces different types for bus0 and
+        // bus1, but they're the same shape, so we'll use a small macro for
+        // codegen.
+        macro_rules! dimm_read_temperature {
+            ($addr:ident, $cmd:ident, $count:ident, $data:ident) => {{
+                self.dimms.$cmd.modify(|b| {
+                    b.set_bus_addr($addr);
+                    b.set_len(2);
+                    b.set_reg_addr(0x31); // current sensed temperature
+                    b.set_op(2); // RANDOM_READ
```
nathanaelhuffman, line 246: I fixed the bug here. Needs to use the RANDOM_READ operation to actually write the address out to the chip before the read, with this change, everything works.

PR #2105, opened 2025-06-23
Introduce `hf-fail-server`

labbott:
Based on what was added for cosmo, use similar logic in gimlet to use an idol server that always returns an error code if we can't read the chip ID.


labbott:
Fixes issue #994

labbott:
This attempt doesn't actually work (messes up humility) and should probably just be fixed upstream in `idolatry` anyway

Issue #1267, opened 2023-04-05
Gimlet sequencer doesn't broadcast A1 point of A2->A0 transition


cbiffle:
This may or may not be a problem, but I noticed it when following a thread from manufacturing today.

I notice that [this code](https://github.com/oxidecomputer/hubris/blob/c7e40ed9f2eac0ea075349cdcbe643ae44d3ce40/drv/gimlet-seq-server/src/main.rs#L513) during the A2->A0 transition appears to be able to enter a partially powered on A1 state, and then just _hang out_ in the loop waiting to fully enter A0. If something interferes with the A0 PG coming up, this loop will spin indefinitely.

This is problematic for a couple of reasons.

1. It's rude -- this is another case where we probably want to transition into a "Failed" state that powers off and can be restarted.
2. It leaves A1 turned on without telling anyone about it: we don't do `set_state(PowerState::A2)` as part of this process.

In particular, the thermal task isn't informed that A1 is on, which would allow it to start reassuring the fan controller and prevent the watchdog from firing. I've gotten one report from the field that suggests that a machine in this state will spin its fans up uncontrollably. I think we could fix this by

- Detecting the entry to A1 separately from the A0 entry and broadcasting it (instead of being clever and doing both in the same loop)
- Updating the thermal task code to handle A1 properly, if required (currently it never sees A1, and its code is set up to handle it like A0, and I'm not confident that's correct).

cbiffle:
Bryan noted that #1303 may have affected this; I still think it's important to broadcast A1 on the way to A0 (or we should remove it from the enum so tasks stop trying to use it)

labbott:
Fixed by removing `A1` in #2107

PR #2107, opened 2025-06-23
Stop mentioing `A1` state outside the sequencer

labbott:
`A1` is supposed to be a transitory state from `A2` onto `A0`. Issue #1267 flagged that we were never broadcasting this information but also made the (modest) suggestion that we could also just remove it.

It turns out, nobody is attempting to do anything significant when the system is in A1. Remove this from the public API to make everything a little easier on the clients. The sequencer still tracks A1 internally.


labbott:
See https://github.com/oxidecomputer/hubris/issues/1267

Aaron-Hartwig (approved):
I agree this seems good to remove! ([RFD 81 reference for posterity](https://rfd.shared.oxide.computer/rfd/0081#_the_a1_sojourning_state))

PR #2109, opened 2025-06-24
Bump hiffy text size

mkeeter:
Cosmo has Too Many Sensors for `humility` dashboard:
```
c:\oxide\hubris (master -> origin)
λ humility dashboard
humility dashboard failed: HIF program cannot be serialized: program length (1067 operations) cannot fit in target program text (2048 bytes); failed after serializing 1024 operations
```

We *could* update Humility to split the program into multiple calls, but it's easier to bump the stack size here.

### DIFF
--- a/app/demo-stm32h7-nucleo/app-h743.toml
+++ b/app/demo-stm32h7-nucleo/app-h743.toml
@@ -2,6 +2,7 @@ name = "demo-stm32h743-nucleo"
 target = "thumbv7em-none-eabihf"
 board = "nucleo-h743zi2"
 chip = "../../chips/stm32h7"
+memory = "memory-large.toml"
 stacksize = 896
 
 [kernel]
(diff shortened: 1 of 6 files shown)


app/demo-stm32h7-nucleo/app-h743.toml
```diff
@@ -2,6 +2,7 @@ name = "demo-stm32h743-nucleo"
 target = "thumbv7em-none-eabihf"
 board = "nucleo-h743zi2"
 chip = "../../chips/stm32h7"
+memory = "memory-large.toml"
```
labbott, line 5: Is this accurate for the h743? I can't remember what the difference actually was...

mkeeter: Good question!

According to the H743 datasheet (DS12110 Rev 10), AXI-SRAM is only 384 Kbytes, versus the expected 512K.

However, the reference manual for the whole product line (RM0433) lists AXI SRAM as 512 Kbytes (0x24000000 - 0x2407FFFF).

I'm not sure who to believe!

mkeeter: Ah, the value of 384 KBytes in DS12110 is actually for H742 chips, which we aren't using, so I think we're fine here.

labbott (approved):
If this turns out to be wrong we can adjust the devboards

PR #2111, opened 2025-06-25
Use idol-generated `FailServer` for cosmo-hf

mkeeter:
https://github.com/oxidecomputer/idolatry/pull/60 automates the previously-manual process of generating a `FailServer` (which replies to every message with an error).


mkeeter:
Whoops, see https://github.com/oxidecomputer/idolatry/pull/61

labbott (approved):

PR #2104, opened 2025-06-20
gimlet-seq-server: use generated state machine status types in ringbuf

Aaron-Hartwig:
This PR makes use of the generated state machine status types we get from the FPGA register map in the ringbuf, making it a bit more human-friendly to tell what is going on, particularly where we got stuck when failing to sequence. Further enhancements could be made if we implemented something like https://github.com/oxidecomputer/hubris/issues/2103 to easily give detailed information about the status of readbacks (power goods, etc). 

This pulls in some updated register map descriptions for pieces related to sequencing (see https://github.com/oxidecomputer/quartz/commit/f9b6e3ebeacf4ad2087db0d4aecaff8ac0a1f2b1).

All of this is quality-of-life improvements for triaging Gimlet sequencing problems.

PR #2114, opened 2025-06-26
Fix sequencing wait condition

mkeeter:
https://github.com/oxidecomputer/hubris/pull/2104 accidentally waited for `Done` instead of `GroupcPg`, so the sequencer would never come up.


nathanaelhuffman (approved):
:shipit:

PR #2117, opened 2025-06-26
Log which sensor causes a thermal event

### LINKED ISSUE #1784: `thermal` should log which `sensor` gave the measurement that causes it to go `Uncontrollable`
Per Matt:
>The only code path that takes it into Uncontrollable without an error is an instantaneous reading that violates our high input limit

Right now if this happens, the ringbuf just looks like
```
27  982        4     4328 ControlPwm(0x0)
28  897        4        1 AutoState(Uncontrollable)
```
I think the desire here is to get an entry between these two that says which sensor it got the problematic reading from, and what the reading was. It is probably also worth logging what the high input limit is set to.

mkeeter:
Fixes #1784, at the cost of a truly cursed iterator


mkeeter:
Tested on Sidecar, and there are no crashes (though I can't easily trigger the actual failure mode).

task/thermal/src/control.rs
```diff
@@ -927,22 +927,27 @@ impl<'a> ThermalControl<'a> {
...
-        (static_inputs, dynamic_inputs): (
-            &'b [InputChannel],
-            &'b [Option<DynamicInputChannel>],
-        ),
-    ) -> impl Iterator<Item = (&'b T, ThermalProperties)> {
+        values: &'b [T; TEMPERATURE_ARRAY_SIZE],
+        static_inputs: &'b [InputChannel; bsp::NUM_TEMPERATURE_INPUTS],
+        dynamic_inputs: &'b [Option<DynamicInputChannel>;
+                bsp::NUM_DYNAMIC_TEMPERATURE_INPUTS],
+        dynamic_sensors: &'b [SensorId; bsp::NUM_DYNAMIC_TEMPERATURE_INPUTS],
```
labbott: Is there a better way to describe these than just `dynamic_inputs` and `dynamic_sensors`? When I was looking at the call site below these are both `dynamic_inputs` which confused me slightly

```
                for (sensor_id, v, model) in Self::zip_temperatures(
                    values,
                    self.bsp.inputs,
                    &self.dynamic_inputs,
                    self.bsp.dynamic_inputs,
                )
```

task/thermal/src/bsp/sidecar_bcd.rs
```diff
@@ -57,11 +57,12 @@ bitflags::bitflags! {
...
 pub(crate) struct Bsp {
-    pub inputs: &'static [InputChannel],
-    pub dynamic_inputs: &'static [SensorId],
+    /// Controlled sensors
+    pub inputs: &'static [InputChannel; NUM_TEMPERATURE_INPUTS],
+    pub dynamic_inputs: &'static [SensorId; NUM_DYNAMIC_TEMPERATURE_INPUTS],
 
     /// Monitored sensors
-    pub misc_sensors: &'static [TemperatureSensor],
+    pub misc_sensors: &'static [TemperatureSensor; NUM_TEMPERATURE_SENSORS],
```
labbott, line 65: I'm guessing this was for consistency even though we don't do anything with these sensors?

labbott (approved):

task/thermal/src/bsp/sidecar_bcd.rs
mkeeter: Yeah, I wanted to switch to strongly-sized arrays throughout.

task/thermal/src/control.rs
mkeeter: I've switched to passing in the `Bsp` itself, and renamed `dynamic_inputs` to `dynamic_channels`.  The iterator is still messy, but is slightly easier to read.

task/thermal/src/control.rs
```diff
@@ -918,31 +918,36 @@ impl<'a> ThermalControl<'a> {
...
-            &'b [InputChannel],
-            &'b [Option<DynamicInputChannel>],
-        ),
-    ) -> impl Iterator<Item = (&'b T, ThermalProperties)> {
-        assert_eq!(values.len(), static_inputs.len() + dynamic_inputs.len());
-        values
+        bsp: &'b Bsp,
+        values: &'b [T; TEMPERATURE_ARRAY_SIZE],
+        dynamic_channels: &'b [Option<DynamicInputChannel>;
+                bsp::NUM_DYNAMIC_TEMPERATURE_INPUTS],
```
hawkw: nit, take it or leave it: maybe consider a 
```rust
type DynamicChannelsArray = [
    Option<DynamicInputChannel>;
    bsp::NUM_DYNAMIC_TEMPERATURE_INPUTS
];
```

or something? Normally, I probably wouldn't suggest this, but the weird rustfmt wrapping here is honestly kind of hard to interpret at first glance...

hawkw (approved):
One little suggestion, but otherwise, LGTM

task/thermal/src/control.rs
mkeeter: Good idea, done

PR #2119, opened 2025-06-27
Add winbond parts to inventory

mkeeter:
I booted up these bits on a Cosmo, and it seems to work fine (other than not knowing how to decode the new type):

```
BRM13250003 # /usr/platform/oxide/bin/ipcc inventory 0x47
metadata:
    version: 0x0
    entries: 0x48
U21 (71) -- Result: 0 [Success]
Type 19, Payload: 0xb bytes
     0  1  2  3  4  5  6  7   8  9  a  b  c  d  e  f   0123456789abcdef
0:  ef 40 19 db 61 90 a4 e7  0f 43 35                | .@..a....C5
```


lib/host-sp-messages/src/lib.rs
```diff
@@ -507,6 +507,14 @@ pub enum InventoryData {
...
         temp_sensors: [SensorIndex; 2],
     },
+
+    /// W25Q256JVEIQ flash chip
+    W25q256jveqi {
+        mfr_id: u8,
+        memory_type: u8,
+        capacity: u8,
+        unique_id: [u8; 8],
+    }
```
hawkw: rustfmt would like a trailing comma here: https://github.com/oxidecomputer/hubris/actions/runs/15933362254/job/44947541537?pr=2119#step:4:25
```suggestion
    },
```

drv/auxflash-server/src/main.rs
```diff
@@ -254,7 +258,21 @@ impl idl::InOrderAuxFlashImpl for ServerImpl {
...
-        Ok(AuxFlashId(idbuf))
+        let mfr_id = idbuf[0];
+        let memory_type = idbuf[1];
+        let capacity = idbuf[2];
+
+        let mut unique_id = [0; 12];
+        self.qspi
+            .read_unique_id(&mut unique_id)
+            .map_err(qspi_to_auxflash)?;
+        let unique_id: [u8; 8] = unique_id[4..].try_into().unwrap_lite();
```
hawkw: are the first 4 bytes always zero or something? could be worth a comment...

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -97,6 +97,14 @@ impl Qspi {
         self.read_impl(Command::ReadId, None, buf)
     }
 
+    /// Reads the Device unique ID buffer
+    ///
+    /// This is implemented in terms of the Winbond part; other chips may
+    /// require different commands.
```
hawkw: I kinda wonder if this ought to be called "read_winbond_unique_id` for that reason? not a big deal.

Also, since it looks like elsewhere, we ignore the first 4 bytes, it might be nice if this comment documented the meaning of the 12 bytes read into the buffer...

drv/auxflash-server/src/main.rs
hawkw: so, it occurs to me that we could save 12 bytes of stack if we changed this to reuse `idbuf`, like
```suggestion
        let mut unique_id: &mut [u8; 12] = &mut idbuf[..12].try_into().unwrap_lite();
        self.qspi
            .read_unique_id(unique_id)
            .map_err(qspi_to_auxflash)?;
        let unique_id: [u8; 8] = unique_id[4..].try_into().unwrap_lite();
```
this comes at the cost of an additional panic site, though; idk if it's worth it... (it's probably not)

hawkw (approved):
overall this looks good, modulo some minor suggestions!

drv/auxflash-server/src/main.rs
mkeeter: I think this isn't worth it; the stack for this function is already huge (due to a Big Variant and lack of placement new).

drv/stm32h7-qspi/src/lib.rs
jgallagher: Is the "ignore first 4 bytes" thing specific to Winbond? If so, maybe this method should take a `[u8; 8]` and do the trimming itself?

mkeeter: I've come around to the even-more-controversial opinion of just returning the `[u8; 8]`.

I was passing in the buffer to match existing style (e.g. `qspi.read_id(..)` above), but just returning the 8-byte value seems reasonable (and it will probably be inlined anyways).

PR #2120, opened 2025-06-30
CPU present/coretypes bits

labbott:


labbott:
```
  10  461        5        1 CPUPresent(true)
  11  475        5        1 Coretype { coretype0: true, coretype1: false, coretype2: true, sp5r1: true, sp5r2: false, sp5r3: false, sp5r4: false }
```
From `cosmo-hubris`

labbott:
see https://github.com/oxidecomputer/hardware-cosmo/issues/658

labbott:
Thanks to @rmustacc for helping to test a CPU we don't support (type 0) and checking the no CPU case

drv/cosmo-seq-server/src/main.rs
```diff
@@ -414,6 +456,38 @@ impl ServerImpl {
...
+                    sp5r2,
+                    sp5r3,
+                    sp5r4
+                });
+
+                match (coretype0, coretype1, coretype2) {
+                    // These correspond to Type-2 and Type-3
+                    (true, false, true) | (true, false, false) => (),
+                    // Reject all other combos
+                    _ => return Err(CpuSeqError::UnrecognizedCPU),
```
mkeeter: Should we return to A2 here (and below)?  It's weird to be in A0 with the HF mux still attached to the SP.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -414,6 +456,38 @@ impl ServerImpl {
...
+                    coretype1,
+                    coretype2,
+                    sp5r1,
+                    sp5r2,
+                    sp5r3,
+                    sp5r4
+                });
+
+                match (coretype0, coretype1, coretype2) {
+                    // These correspond to Type-2 and Type-3
```
mkeeter, line 479: Can you add an AMD reference for these values (and the `sp5rX` combo below)?

labbott: I guess the reference manual name should be okay even if the file itself is confidential

drv/cosmo-seq-server/src/main.rs
labbott: You're right, I had somehow convinced myself we had error handling to do this already but we don't

mkeeter (approved):

drv/cosmo-seq-server/src/main.rs
```diff
@@ -69,7 +69,15 @@ enum Trace {
...
-
+    Coretype {
+        coretype0: bool,
+        coretype1: bool,
+        coretype2: bool,
+        sp5r1: bool,
+        sp5r2: bool,
+        sp5r3: bool,
+        sp5r4: bool,
+    },
```
hawkw, line 80: i briefly wondered whether there was a more space-efficient way to represent this, but it's only 7 bytes, and isn't the longest entry type in this ringbuf, so it doesn't actually matter. carry on!

drv/cosmo-seq-server/src/main.rs
```diff
@@ -414,6 +456,46 @@ impl ServerImpl {
...
+                    _ => {
+                        self.seq.power_ctrl.modify(|m| m.set_a0_en(false));
+                        return Err(CpuSeqError::UnrecognizedCPU);
+                    }
+                };
+
+                // From sp5-mobo-guide-56870_1.1.pdf table 73
+                match (sp5r1, sp5r2, sp5r3, sp5r4) {
+                    // There is only combo we accept here
+                    (true, false, false, false) => (),
```
hawkw, line 491: some part of me wants to suggest putting this in a struct, so that we match them by name rather than position, making it harder to mess up. but, we only ever do this once, and since we reject every other possible set of values, there's not that many chances to get it wrong. so... :woman_shrugging:

hawkw (approved):

PR #2121, opened 2025-06-30
stm32xx-i2c: remove I2cControl function table

cbiffle:
It turns out that, in practice, we only ever provide a single function for each of `wfi` and `enable` here. This makes the table itself an unnecessary level of indirection, which has some deleterious effects:

- It opens the possibility for _different_ I2cControl structs containing _different_ function pointers to be passed to each call, complicating analysis. (Searching for this phenomenon was what led me to notice that we only ever use one implementation.)

- It makes it a lot harder to recognize use of syscalls (or misuse of syscalls) in the I2C layer.

- It routes important parts of the I2C stack through indirect function calls, defeating stack analysis.

- It adds a bunch of code that isn't really necessary, as it turns out.

This commit removes the dispatch table, inlining the existing functions into their callsites (and introducing a utility function for the wfi behavior).

### DIFF
--- a/drv/stm32xx-i2c/src/lib.rs
+++ b/drv/stm32xx-i2c/src/lib.rs
@@ -76,17 +76,6 @@ pub enum I2cControlResult {
     TimedOut,
 }
 
-///
-/// A structure that defines interrupt control flow functions that will be
-/// used to pass control flow into the kernel to either enable or wait for
-/// interrupts.  Note that this is deliberately a struct and not a trait,
-/// allowing the [`I2cMuxDriver`] trait to itself be a trait object.
-///
-pub struct I2cControl {
-    pub enable: fn(u32),
-    pub wfi: fn(u32, I2cTimeout) -> I2cControlResult,
-}
-
 pub struct I2cTargetControl {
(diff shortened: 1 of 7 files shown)


drv/stm32xx-i2c/src/lib.rs
```diff
@@ -1190,3 +1178,35 @@ impl I2cController<'_> {
...
     }
 }
+
+/// Waits for `notification` or `timeout`, whichever comes first. This is
+/// factored out of `wfi` above for clarity.
+///
+/// This function, like `userlib::hl`, assumes that notification bit 31 is safe
+/// to use for the timer. If `notification` also has bit 31 set, weird things
+/// will happen, don't do that.
+fn wfi_raw(event_mask: u32, timeout: I2cTimeout) -> I2cControlResult {
```
mkeeter, line 1189: The docstring is now outdated (refers to `notification` rather than `event_mask`)

mkeeter:
I've rebased this onto `master` and update the docstring.  I think it should be good to go, but will do some hardware testing tomorrow to be certain.

PR #2122, opened 2025-06-30
Add Cosmo host flash to inventory

mkeeter:
- Add a new `HfChipId` type as a more structured chip ID, then use it in the various host flash drivers
- Add an `InventoryData::W25q01jvzeiq` variant and use it on Cosmo

This needs testing, so I'm opening as a draft for now.

### DIFF
--- a/drv/cosmo-hf/src/main.rs
+++ b/drv/cosmo-hf/src/main.rs
@@ -55,9 +55,16 @@ fn main() -> ! {
     };
     drv.flash_set_quad_enable();
 
-    // Check the flash chip's ID against Table 7.3.1 in the datasheet
+    // Check the flash chip's ID against Table 7.3.1 in the W25Q01JV datasheet.
     let id = drv.flash_read_id();
-    if id.mfr_id != 0xef || id.memory_type != 0x40 || id.capacity != 0x21 {
+    const WINBOND_MFR_ID: u8 = 0xef;
+    const EXPECTED_TYPE: u8 = 0x40;
+    const EXPECTED_CAPACITY: u8 = 0x21;
+
+    if id.mfr_id != WINBOND_MFR_ID
+        || id.memory_type != EXPECTED_TYPE
+        || id.capacity != EXPECTED_CAPACITY
+    {
         fail(drv_hf_api::HfError::BadChipId);
     }
 
@@ -112,7 +119,8 @@ mod instr {
 
 impl FlashDriver {
     fn flash_read_id(&mut self) -> drv_hf_api::HfChipId {
-        // Make sure die 0 is selected with a dummy read
+        // Make sure die 0 is selected with a dummy read, because the
+        // READ_UNIQUE_ID command is die-specific.
         let mut buf = [0u8; 4];
         self.flash_read(FlashAddr(0), &mut buf.as_mut_slice())
             .unwrap_lite(); // infallible when given a slice
--- a/drv/cosmo-hf/src/main.rs
+++ b/drv/cosmo-hf/src/main.rs
@@ -28,7 +28,6 @@ enum Trace {
     FpgaBusy,
     SectorEraseBusy,
     WriteBusy,
-    DummyBytes(u32),
 
     HashInitError(drv_hash_api::HashError),
     HashUpdateError(drv_hash_api::HashError),
@@ -118,22 +117,6 @@ impl FlashDriver {
         self.flash_read(FlashAddr(0), &mut buf.as_mut_slice())
             .unwrap_lite();
 
-        // Read the 12-byte unique ID (4 bytes of which are junk)
-        self.drv.data_bytes.set_count(12);
-        self.drv.addr.set_addr(0);
-        self.drv.dummy_cycles.set_count(0);
-        self.drv.instr.set_opcode(instr::READ_UNIQUE_ID);
-        self.wait_fpga_busy();
-        let v = self.drv.rx_fifo_rdata.fifo_data(); // dummy bytes
-        ringbuf_entry!(Trace::DummyBytes(v));
-        let mut unique_id = [0u8; 17];
-        for i in 0..2 {
-            let v = self.drv.rx_fifo_rdata.fifo_data();
-            for (j, byte) in v.to_le_bytes().iter().enumerate() {
-                unique_id[i * 4 + j] = *byte;
-            }
-        }
-
         self.clear_fifos();
         self.drv.data_bytes.set_count(3);
         self.drv.addr.set_addr(0);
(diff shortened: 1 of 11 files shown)


drv/cosmo-hf/src/main.rs
```diff
@@ -50,14 +50,14 @@ fn main() -> ! {
...
-    let drv = FlashDriver {
+    let mut drv = FlashDriver {
         drv: fmc_periph::SpiNor::new(seq.get_token()),
     };
     drv.flash_set_quad_enable();
 
     // Check the flash chip's ID against Table 7.3.1 in the datasheet
     let id = drv.flash_read_id();
-    if id[0..3] != [0xef, 0x40, 0x21] {
+    if id.mfr_id != 0xef || id.memory_type != 0x40 || id.capacity != 0x21 {
```
labbott: `const` the magic numbers?

drv/cosmo-hf/src/main.rs
```diff
@@ -103,28 +103,55 @@ mod instr {
...
     pub const QUAD_INPUT_PAGE_PROGRAM_4B: u8 = 0x34;
 }
 
 impl FlashDriver {
-    fn flash_read_id(&self) -> [u8; 20] {
+    fn flash_read_id(&mut self) -> drv_hf_api::HfChipId {
+        // Make sure die 0 is selected with a dummy read
+        let mut buf = [0u8; 4];
+        self.flash_read(FlashAddr(0), &mut buf.as_mut_slice())
+            .unwrap_lite(); // infallible when given a slice
```
labbott: Change in behavior or just being extra cautious?

drv/cosmo-hf/src/main.rs
mkeeter: Sure, done in 6fe01d6966b3

drv/cosmo-hf/src/main.rs
mkeeter: This is deliberate: reading the unique ID is die-specific, so we want to always read die 0 as our canonical ID.  I updated the comment to explain _why_ we're selecting die 0 here.

labbott: I kept forgetting that we changed this to the _unique_ id instead of just ID so the comment is helpful

drv/cosmo-hf/src/main.rs
```diff
@@ -50,14 +50,21 @@ fn main() -> ! {
...
         drv_spartan7_loader_api::Spartan7Loader::from(LOADER.get_task_id());
 
-    let drv = FlashDriver {
+    let mut drv = FlashDriver {
         drv: fmc_periph::SpiNor::new(seq.get_token()),
     };
     drv.flash_set_quad_enable();
 
-    // Check the flash chip's ID against Table 7.3.1 in the datasheet
+    // Check the flash chip's ID against Table 7.3.1 in the W25Q01JV datasheet.
```
rmustacc: I would include which revision of the datasheet this is. Especially as my copy of Rev E (March 04, 2024) has this as section 6.3.1 even thought the ToC is 7.3.1. I would also note that these IDs are specific to the JEDEC ID command as the table points out other values that are different from other commands.

lib/host-sp-messages/src/lib.rs
```diff
@@ -515,6 +515,18 @@ pub enum InventoryData {
         capacity: u8,
         unique_id: [u8; 8],
     },
+
+    /// Cosmo host flash
+    W25q01jvzeiq {
+        mfr_id: u8,
+        memory_type: u8,
+        capacity: u8,
```
rmustacc: I don't think there's any point to these three fields a they're implied by the constrained type. If we were going to make this a more generic type to cover more of the `W25Q` family, then it might make sense, but otherwise these are all implied by the actual type name.

More generally, a single byte mfg id isn't helpful because it's not enough to uniquely identify SPI NOR devices. JEDEC relies on two byte IDs to identify a vendor, a bank byte and an index in the byte. Winbond and Micron are in the first bank, which is why they often elide these continuations. However, Macronix for example, is not, but also doesn't have a 16-bit id there to tell you what bank it's in because that's how the rest of these go. To add insult to injury, the JEDEC ID for Winbond Electronic according to JEP106 is 0xDA not 0xEF that these devices contain.

The type and capacity values which the Winbond datasheets don't actually translate the meanings for (unlike say the Micron MT25Q series) similarly can't actually be interpreted to be generic, though we could make some guesses that 0x40 and 0x60 are 3.3V and 1.8V for the type and that the capacity bytes are following Micron MT25Q bit though this is only from spot checking a few.

I guess all this is to say that I probably wouldn't bother including these fields.

lib/host-sp-messages/src/lib.rs
```diff
@@ -515,6 +515,18 @@ pub enum InventoryData {
...
+
+        /// 64-bit unique ID for die 0
+        ///
+        /// Note that each 512M-bit die has a separate ID!
+        unique_id: [u8; 8],
```
rmustacc: I would probably include both die's `unique_id` values. Given that this consists of two dies, I would probably model it that way in topology. It's hard to say how practical it is to have a single die failure, but given that we are choosing which die we're addressing and the issues there, that gives us something.

Reviewing what we have from them, it's not clear if we'd actually ever get failure/recall information that is die specific from the manufacturer or if we'd ultimately just get date code information. It's probably just the latter to be honest, which we have no means of electronically identifying it appears.

lib/host-sp-messages/src/lib.rs
mkeeter: > I don't think there's any point to these three fields a they're implied by the constrained type. If we were going to make this a more generic type to cover more of the `W25Q` family, then it might make sense, but otherwise these are all implied by the actual type name.

Good point!  Should we also remove those fields from the auxflash variant above (`W25q256jveqi`), since I don't think it's been baked into host software yet?

rmustacc: I probably would.

lib/host-sp-messages/src/lib.rs
mkeeter: Done in 21caaea

lib/host-sp-messages/src/lib.rs
mkeeter: Done in 21caaea

drv/cosmo-hf/src/main.rs
mkeeter: Done in 21caaea

drv/cosmo-hf/src/main.rs
```diff
@@ -137,6 +132,35 @@ impl FlashDriver {
...
+        self.flash_read(FlashAddr(0), &mut buf.as_mut_slice())
+            .unwrap_lite(); // infallible when given a slice
+        let die0_id = self.read_unique_id();
+
+        // Then read the top die's unique ID
+        self.flash_read(FlashAddr(0x04000000), &mut buf.as_mut_slice())
+            .unwrap_lite(); // infallible when given a slice
+        let die1_id = self.read_unique_id();
+
+        let mut unique_id = [0u8; 17];
```
rmustacc, line 147: Why a 17 byte array when we're only storing 16 bytes?

rmustacc:
The updated inventory interfaces make sense to me. Thanks @mkeeter.

drv/cosmo-hf/src/main.rs
mkeeter: The `unique_id` array in `HfChipId` has to support multiple host flash types, so the _maximum_ unique ID length is 17 bytes (from the Gimlet's host flash).

rmustacc: Gotcha. I guess the assumption that this and the packing format are clear to other readers? I don't think I would have made the connection to that size and this is where I would use some constant as a way to know that every callers has to change this and bump it rather than open-coding it everywhere. But maybe this is more straightforward due to a compile time error.

mkeeter:
Tested on hardware and this works fine:
```console
BRM13250001 # /usr/platform/oxide/bin/ipcc inventory 0x48
metadata:
    version: 0x0
    entries: 0x49
U28 (72) -- Result: 0 [Success]
Type 20, Payload: 0x10 bytes
      0  1  2  3  4  5  6  7   8  9  a  b  c  d  e  f   0123456789abcdef
00:  ea 62 94 22 47 2b 1d 2e  ea 62 94 22 47 13 0b 35 | .b."G+...b."G..5
```

drv/cosmo-hf/src/main.rs
mkeeter: I think if we add a _third_ flash type, then it would make sense to use enum variants, but right now, it's not too hard to make sure that the sides are in agreement about how we're using the byte array.

PR #2124, opened 2025-07-01
Introduce `compute-sleds` release target

labbott:
With the addition of `cosmo`, we've switched to the term `compute-sled` to describe gimlet and cosmo together. Add this as a release target. We can keep the old `gimlet` target for backwards compatibility. If we end up really needing a `cosmo` only release we can add that as a target too.

### DIFF
--- a/.github/workflows/build-boards.yml
+++ b/.github/workflows/build-boards.yml
@@ -118,7 +118,7 @@ jobs:
       os: ${{ inputs.os }}
 
   build-gimlet:
-    if: ${{ inputs.board-set == 'gimlet' || inputs.board-set == 'all' }}
+    if: ${{ inputs.board-set == 'gimlet' || inputs.board-set == 'all' || inputs.board-set == 'compute-sleds' }}
     name: build-gimlet
     strategy:
       matrix:
@@ -193,7 +193,7 @@ jobs:
       os: ${{ inputs.os }}
 
   build-cosmo:
-    if: ${{ inputs.board-set == 'cosmo' || inputs.board-set == 'all' }}
+    if: ${{ inputs.board-set == 'cosmo' || inputs.board-set == 'all' || inputs.board-set == 'compute-sleds' }}
     name: build-cosmo
     strategy:
       matrix:
--- a/.github/workflows/release-builds.yml
+++ b/.github/workflows/release-builds.yml
@@ -7,6 +7,7 @@ on:
       - "psc-*"
       - "sidecar-*"
       - "devboards-*"
+      - "compute-sleds-*"
 
 jobs:
   release-rot:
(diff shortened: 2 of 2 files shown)


.github/workflows/build-boards.yml
```diff
@@ -118,7 +118,7 @@ jobs:
       os: ${{ inputs.os }}
 
   build-gimlet:
-    if: ${{ inputs.board-set == 'gimlet' || inputs.board-set == 'all' }}
+    if: ${{ inputs.board-set == 'gimlet' || inputs.board-set == 'all' || inputs.board-set == 'compute-sleds' }}
```
labbott: Everyone should give an opinion if this should be `compute-sled` or `compute-sleds`

labbott: <strike>computes-sleds</strike>

.github/workflows/release-builds.yml
```diff
@@ -7,6 +7,7 @@ on:
       - "psc-*"
       - "sidecar-*"
       - "devboards-*"
+      - "compute-sleds-*"
```
mkeeter: Are we expecting to push a `compute-sleds-vX` tag, or should this be `cosmo-*` instead?

mkeeter: Okay, I think this is intentional, but it seems unintuitive to me – I would expect separate tags for `gimlet` and `cosmo` releases.

labbott: We went back and forth and I _thought_ we had decided to have them be a single unit for now? Or did I completely misremember our decision that was not recorded in an RFD :upside_down_face:

.github/workflows/release-builds.yml
```diff
@@ -39,4 +41,16 @@
     with:
       board-set: devboards
 
+  release-cosmo:
+    if: startsWith(github.ref, 'refs/tags/cosmo')
+    uses: ./.github/workflows/release.yml
+    with:
+      board-set: cosmo
+
+  release-all-sp:
```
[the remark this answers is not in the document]
sudomateo: Setting `permissions: {}` should stop the code scanner from complaining here. This job calls out to `.github/workflows/release.yml` which will create a GitHub release. I don't see anything that modifies pull requests.

Give that, the permissions needed by this job should be.

```yaml
permissions:
  contents: write
```

.github/workflows/release-builds.yml
```diff
@@ -39,4 +41,16 @@
...
+    if: startsWith(github.ref, 'refs/tags/all-sp')
+    uses: ./.github/workflows/release.yml
+    with:
+      board-set: all-sp
```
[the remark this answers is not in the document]
sudomateo: Same thing as noted here: https://github.com/oxidecomputer/hubris/pull/2124#discussion_r2190934633

.github/workflows/release-builds.yml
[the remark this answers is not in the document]
labbott: I decided to try what copilot eventually suggested. It was insufficient and @sudomateo was corrected we at least needed `contents: write` in addition to other parts for attestation

.github/workflows/release-builds.yml
labbott: Ended up taking a different approach to build all SPs (with a separate cosmo tag for backup)

PR #2126, opened 2025-07-03
packrat ereport storage and snitch implementation

hawkw:
Things fail.

Not finding out about it sucks.

This branch implements the Hubris side of the ereport ingestion system,
as described [RFD 545]. Work on this was started by @cbiffle in #2002,
which implemented the core ring-buffer data structure used to store
ereports. Meanwhile, oxidecomputer/management-gateway-service#370, 
oxidecomputer/omicron#7803, and oxidecomputer/omicron#8296 added
the MGS and Omicron components of this system.

This branch picks up where Cliff left off, and "draws the rest of the
owl" by implementing the aggregation of ereports in the `packrat` task
using this data structure, and adding a new `snitch` task, which acts as
a proxy to allow ereports stored by `packrat` to be read over the
management network.

## Architecture

Ereports are stored by `packrat` because we would like as many tasks as
possible to be able to report errors by making IPC call to the task
responsible for ereport storage. This means that the task aggregating
ereports must be a high-priority task, so that as many other tasks as
possible may be its clients. Additionally, we would like to include the
system's VPD identity as metadata for ereports, and this data is already
stored by packrat. Finally, we would like to minimize the likelihood of
the task that stores ereports crashing, as this would result in data
loss, and packrat already is expected not to crash.

On the other hand, the task that actually evacuates these ereports over
the management network must run at a priority lower than that of the
`net` task, of which it is a client. Thus the separation of
responsibilities between `packrat` and the `snitch`. The snitch task is
fairly simple. It receives packets sent to the ereport socket,
interprets the request message, and forwards the request to packrat. Any
ereports sent back by packrat are sent in response to the request. The
snitch ends up being a pretty dumb, stateless proxy: as the response
packet is encoded by packrat; all we end up doing is taking the bytes
received from packrat and stuffing them into the socket's send queue.
The real purpose of this thing is just to serve as a trampoline between
the high priority level of packrat and a priority level lower than that
of the net task.

## `snitch-core` Fixes

While testing behavior when the ereport buffer is full, I found a
potential panic in the existing `snitch-core` code. Previously, every
time ereports are read from the buffer while it is in the `Losing` state
(i.e., ereports have been discarded because the buffer was full),
`snitch-core` attempts to insert a new loss record at the end of the
buffer (calling `recover_if_needed()`). This ensures that the data loss
is reported to the reader ASAP. The problem is that this code assumed
that there would always be space for an additional loss record, and
panicked if it didn't fit. I added a test reproducing this panic in
ff93754b694fbbc6c9f60c0b85bab4b5470e3f98, and fixed it in
22044d10948b7e68fe98917b0b3ef35bbbe998d4 by changing the calculation of
whether recovery is possible.

When `recover_if_needed` is called while in the `Losing` state, we call
the `free_space()` method to determine whether we can recover. In the
`Losing` state, [this method would calculate the free space by
subtracting the space required for the loss record][1] that must be
encoded to transition out of the `Losing` state. However, in the case
where `recover_if_required()` is called with `required_space: None`
(which indicates that we're not trying to recover because we want to
insert a new record, but just because we want to report ongoing data
loss to the caller), [we check that the free space is greater than or
equal to 0][2]. This means that we would still try to insert a loss
record even if the free space was 0, resulting in a panic. I've fixed
this by moving the check that there's space for a loss record out of the
calculation of `free_space()` and into the _required_ space, in addition
to the requested value (which is 0 in the "we are inserting the loss
record to report loss" case). This way, we only insert the loss record
if it fits, which is the correct behavior.

I've also changed the assignment of ENAs in `snitch-core` to start at 1,
rather than 0, since ENA 0 is reserved in the wire protocol to indicate
"no ENA". In the "committed ENA" request field this means "don't flush
any ereports", and in the "start ENA" response field, ENA 0 means "no
ereports in this packet". Thus, the ereport store must start assigning
ENAs at ENA 1 for the initial loss record.

## Testing

Currently, no tasks actually produce ereports. To test that everything
works correctly, it was necessary to add a source of ereports, so I've
added [a little task][3] that just generates test ereports when asked
via `hiffy`. I've included some of that in [this comment][4]. This was
also used for testing the data-loss behavior discussed above.

[RFD 545]: https://rfd.shared.oxide.computer/rfd/0545
[1]:
    https://github.com/oxidecomputer/hubris/blob/e846b9d2481b13cf2b18a2a073bb49eef5f654de/lib/snitch-core/src/lib.rs#L110-L121
[2]:
    https://github.com/oxidecomputer/hubris/blob/e846b9d2481b13cf2b18a2a073bb49eef5f654de/lib/snitch-core/src/lib.rs#L297-L300
[3]: https://github.com/oxidecomputer/hubris/blob/864fa57a7c34a6225deddcffa0c7d54c3063eab6/task/ereportulator/src/main.rs


task/packrat/src/ereport.rs
```diff
@@ -0,0 +1,385 @@
...
+        let mut encoder = minicbor::Encoder::new(c);
+        // TODO(eliza): presently, this code bails out if the metadata map gets
+        // longer than our buffer. It would be nice to have a way to keep the
+        // encoded metadata up to the last complete key-value pair...
+        encoder
+            .str("hubris_archive_id")
+            .map_err(|_| Error::VpdMetadataTooLong)?
+            .bytes(&self.image_id[..])
+            .map_err(|_| Error::VpdMetadataTooLong)?;
+        if let Some(vpd) = vpd {
```
hawkw: this is a little bit sketch: previously, i was using an empty metadata map as an indication that MGS should request metadata again (by requesting restart ID 0). but, now that we're adding the image ID, we have one piece of metadata that's always knowable, and another few bits that are only knowable once the VPD has been sent to packrat. So, if MGS ingests ereports prior to us getting the VPD, we will presently send it a meta map with just the image ID, and it may then not send subsequent metadata refresh requests because we will still have the same restart ID.

I'm not totally sure what the best way to handle this is. I can see the following ways forward:

- **don't send any metadata until we have all the metadata.** this is maybe the cleanest, but if the VPD loading is busted, it means we'll never know the Hubris image ID. I dunno if this matters...
- **don't send any _ereports_ until VPD has been loaded.** i don't like this because it means a failure to read the VPD EEPROM means we'd never send ereports.
- **regenerate the restart ID when VPD is loaded.** i don't like this because it essentially generates a "fake restart"
- **somehow track whether we have sent off full metadata on a client-by-client basis**: bad, requires packrat to track individual clients; i don't want to be doing that as it seems really failure prone
- **something i wasn't smart enough to think of**; suggestions are welcome

drv/stm32h7-rng/src/main.rs
```diff
@@ -228,11 +233,42 @@ impl NotificationHandler for Stm32h7RngServer {
...
+    ringbuf_entry!(Trace::SetRestartId(result));
+}
+
 #[export_name = "main"]
 fn main() -> ! {
     let mut rng = Stm32h7Rng::new();
     rng.init();
 
+    #[cfg(feature = "ereport")]
+    generate_restart_id(&mut rng);
```
hawkw, line 270: only doing this once when the RNG driver starts up means that if packrat crashes, it will never get a new restart ID. do we care about this? IIUC, other things that cache data in packrat don't have any mechanism for recovering from packrat faulting. if we wanted to have this recover from packrat death, we could have a notification to the RNG driver to say "im dead, please send me a new restart ID."

task/snitch/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+                    .copied()
+                    .unwrap_or(gateway_ereport_messages::Ena::NONE),
+                &mut tx_buf[..],
+            ) {
+                Ok(size) => size,
+                Err(e) => {
+                    // Packrat's mad. Reject the request.
+                    count!(Event::ReadError(e));
+                    continue;
+                }
```
hawkw: in practice this means "we don't have a restart ID yet"

task/packrat/src/ereport.rs
hawkw: i decided to just go with the first option, which felt the least potentially error prone to me: https://github.com/oxidecomputer/hubris/pull/2126/commits/32244eee2d85c774ae9c379c57d2fffd0529c158

task/snitch/src/main.rs
```diff
@@ -0,0 +1,167 @@
...
+                &mut tx_buf[..],
+            ) {
+                Ok(size) => size,
+                Err(e) => {
+                    // Packrat's mad. Reject the request.
+                    //
+                    // Presently, the only time we'd see an error here is if we
+                    // have yet to generate a restart ID.
+                    count!(Event::ReadError(e));
+                    continue;
```
hawkw, line 136: It occurs to me that another possible way to handle this state is to have the snitch wait for a notification from packrat saying that the restart ID has been set, *before* it tries to handle any incoming packets. I'm not sure if there's any real advantage to letting the packets queue up rather than dropping them, though, because if we let them queue up, MGS will time them out and retry them anyway. So I think this ends up looking pretty much equivalent.

task/packrat/src/ereport.rs
```diff
@@ -0,0 +1,405 @@
...
+                encoder
+                    .str("baseboard_serial_number")
+                    .map_err(|_| MetadataError::TooLong)?
+                    .str(serial_number)
+                    .map_err(|_| MetadataError::TooLong)?;
+            }
+            Err(_) => ringbuf_entry!(Trace::MetadataError(
+                MetadataError::SerialNumberNotUtf8
+            )),
+        }
```
hawkw: The names "baseboard_serial_number" and "baseboard_part_number" are currently [special to Nexus](https://github.com/oxidecomputer/omicron/blob/77d097ac8c1f61fb4085a4e38444c14ea276995d/nexus/src/app/background/tasks/ereport_ingester.rs#L249-L262) as they are used to populate database columns. These keys are, admittedly, kind of long: I included the "baseboard_" prefix to avoid potential collisions with other ereport keys named "serial_number" or "part_number" that might refer to some other thing, because MGS will currently [just jam all the metadata keys directly into the ereport body](https://github.com/oxidecomputer/management-gateway-service/blob/77e316c812aa057b9714d0d99c4a7bdd36d45be2/gateway-sp-comms/src/ereport.rs#L504-L509), and I didn't want to clobber another field named "serial number".

Some things we might consider doing instead include: 
- Using "well-known" concise names for these that are shared by Hubris and MGS in the `gateway_ereport_messages` crate, so that MGS will just turn anything named "bord_sn" into "baseboard_serial_number" and so on. That saves us a handful of bytes at the cost of adding more things that are special-cased in MGS, complicating versioning.
- Have MGS insert the metadata map as a nested object in the ereport JSON, instead of flattening it, like this:
  ```json
  {
     "meta": {
       "serial": "BRM690420",
       "part_num": "123-000045",
       "hubris_img": "asdfff",
    },
    "class": "thingy_busted",
    "serial": "12345",
  }
  ```
  so that they don't clobber other things named "serial" or "part_number"
- Just change the Nexus code to care about shorter key names, so that _it_ turns "bord_sn" into "baseboard_serial_number" etc.

However, I don't really think solving for getting rid of a few bytes is all that important here. Two instances of the string "baseboard_" is only 20 bytes of our 1024-byte UDP packet, and it only gets sent in the initial metadata packets rather than adding overhead to _every_ ereport packet. So, maybe leaving things as they are is fine for now, and we can make this more complicated later if it turns out we really need to squeeze a few more bytes out of it?

task/packrat/src/main.rs
```diff
@@ -408,6 +476,82 @@ impl idl::InOrderPackratImpl for ServerImpl {
...
+    }
+
+    #[cfg(not(feature = "ereport"))]
+    fn deliver_ereport(
+        &mut self,
+        _: &RecvMessage,
+        _: LenLimit<Leased<idol_runtime::R, [u8]>, 1024usize>,
+    ) -> Result<(), RequestError<Infallible>> {
+        // go away, we don't know how to do that
+        Err(idol_runtime::ClientError::UnknownOperation.fail())
```
hawkw, line 507: Perhaps this actually ought to just be a no-op if the feature is disabled, so that if a task tries to send an ereport on a system that doesn't support it, it just does nothing instead of getting faulted? On the other hand, we probably _do_ want any other tasks that report ereports to feature flag doing so for space-reduction reasons (and to avoid spending time CBOR-ing up a bunch of stuff that just goes to /dev/null)...

hawkw:
<details>

<summary>manual testing on gimletlet</summary>

```console
# check that `net` has come up
eliza@theseus ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- net ip
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.21s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- net ip`
MAC address:  0e:1d:de:f0:d9:21
IPv6 address: fe80::0c1d:deff:fef0:d921

# create some ereports using the ereportulator (VPD is unset)
eliza@theseus ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=1
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.15s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=1`
Ereportulator.fake_ereport() => ()

eliza@theseus ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=2
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.16s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=2`
Ereportulator.fake_ereport() => ()

eliza@theseus ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=3
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.16s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=3`
Ereportulator.fake_ereport() => ()

# read ereports via faux-mgs, without sending the restart ID
eliza@theseus ~/Code/oxide/management-gateway-service $ cargo run --bin faux-mgs -- --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/faux-mgs --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports`
Jul 09 18:31:42.842 INFO creating SP handle on interface enp193s0f3u2, component: faux-mgs
Jul 09 18:31:42.844 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d921%2]:11111, interface: enp193s0f3u2, socket: control-plane-agent, component: faux-mgs
restart ID: f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2
count: 3

ereports:
0x1: {
    "hubris_task_gen": Number(0),
    "lost": Null,
}

0x2: {
    "badness": Number(1),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(128651),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
}

0x3: {
    "badness": Number(2),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(132301),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
}

# flush up to ENA 0x2 using the received restart ID:
eliza@theseus ~/Code/oxide/management-gateway-service $ cargo run --bin faux-mgs -- --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/faux-mgs --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports`
Jul 09 18:31:42.842 INFO creating SP handle on interface enp193s0f3u2, component: faux-mgs
Jul 09 18:31:42.844 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d921%2]:11111, interface: enp193s0f3u2, socket: control-plane-agent, component: faux-mgs
restart ID: f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2
count: 3

ereports:
0x1: {
    "hubris_task_gen": Number(0),
    "lost": Null,
}

0x2: {
    "badness": Number(1),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(128651),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
}

0x3: {
    "badness": Number(2),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(132301),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
}

eliza@theseus ~/Code/oxide/management-gateway-service $ cargo run --bin faux-mgs -- --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports --restart-id f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2 --start-ena 0x2 --committed-ena 0x2
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/faux-mgs --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports --restart-id f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2 --start-ena 0x2 --committed-ena 0x2`
Jul 09 18:32:36.292 INFO creating SP handle on interface enp193s0f3u2, component: faux-mgs
Jul 09 18:32:36.293 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d921%2]:11111, interface: enp193s0f3u2, socket: control-plane-agent, component: faux-mgs
restart ID: f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2
count: 1

ereports:
0x3: {
    "badness": Number(2),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(132301),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
}

# make another ereport, set fake VPD metadata, and make another ereport
eliza@theseus ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=3
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.16s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=3`
Ereportulator.fake_ereport() => ()

eliza@theseus ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.set_fake_vpd
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.16s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.set_fake_vpd`
Ereportulator.set_fake_vpd() => ()

eliza@theseus ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=5
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.18s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.fake_ereport -a n=5`
Ereportulator.fake_ereport() => ()

# request more ereports, flushing 0x3 --- note that we now see VPD metadata
eliza@theseus ~/Code/oxide/management-gateway-service $ cargo run --bin faux-mgs -- --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports --restart-id f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2 --start-ena 0x3 --committed-ena 0x3
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/faux-mgs --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports --restart-id f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2 --start-ena 0x3 --committed-ena 0x3`
Jul 09 18:33:28.229 INFO creating SP handle on interface enp193s0f3u2, component: faux-mgs
Jul 09 18:33:28.231 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d921%2]:11111, interface: enp193s0f3u2, socket: control-plane-agent, component: faux-mgs
restart ID: f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2
count: 2

ereports:
0x4: {
    "badness": Number(3),
    "baseboard_part_number": String("LOLNO000000"),
    "baseboard_serial_number": String("69426661337"),
    "hubris_archive_id": String("BELgNXi52l4"),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(319934),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
    "rev": Number(42),
}

0x5: {
    "badness": Number(5),
    "baseboard_part_number": String("LOLNO000000"),
    "baseboard_serial_number": String("69426661337"),
    "hubris_archive_id": String("BELgNXi52l4"),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(344051),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
    "rev": Number(42),
}

# try to read with the ENA of a flushed ereport, nothing happens:
eliza@theseus ~/Code/oxide/management-gateway-service $ cargo run --bin faux-mgs -- --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports --restart-id f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2 --start-ena 0x1
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/faux-mgs --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports --restart-id f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2 --start-ena 0x1`
Jul 09 18:39:23.118 INFO creating SP handle on interface enp193s0f3u2, component: faux-mgs
Jul 09 18:39:23.119 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d921%2]:11111, interface: enp193s0f3u2, socket: control-plane-agent, component: faux-mgs
restart ID: f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2
count: 2

ereports:
0x4: {
    "badness": Number(3),
    "baseboard_part_number": String("LOLNO000000"),
    "baseboard_serial_number": String("69426661337"),
    "hubris_archive_id": String("BELgNXi52l4"),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(319934),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
    "rev": Number(42),
}

0x5: {
    "badness": Number(5),
    "baseboard_part_number": String("LOLNO000000"),
    "baseboard_serial_number": String("69426661337"),
    "hubris_archive_id": String("BELgNXi52l4"),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(344051),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
    "rev": Number(42),
}

# requests with non-matching restart IDs don't flush, and start at
# the earliest known ENA
eliza@theseus ~/Code/oxide/management-gateway-service $ cargo run --bin faux-mgs -- --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports --restart-id $(uuidgen -r) --start-ena 0x5 --committed-ena 0x5
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/faux-mgs --interface enp193s0f3u2 --discovery-addr '[fe80::0c1d:deff:fef0:d921]:11111' ereports --restart-id 56c1eefc-2f0f-4358-b28c-3b867bd7e886 --start-ena 0x5 --committed-ena 0x5`
Jul 09 18:40:56.495 INFO creating SP handle on interface enp193s0f3u2, component: faux-mgs
Jul 09 18:40:56.496 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d921%2]:11111, interface: enp193s0f3u2, socket: control-plane-agent, component: faux-mgs
restart ID: f5a3acec-ee5d-daa4-fb77-93b2f5f8ddb2
restart IDs did not match (requested 56c1eefc-2f0f-4358-b28c-3b867bd7e886)
count: 2

ereports:
0x4: {
    "badness": Number(3),
    "baseboard_part_number": String("LOLNO000000"),
    "baseboard_serial_number": String("69426661337"),
    "hubris_archive_id": String("BELgNXi52l4"),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(319934),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
    "rev": Number(42),
}

0x5: {
    "badness": Number(5),
    "baseboard_part_number": String("LOLNO000000"),
    "baseboard_serial_number": String("69426661337"),
    "hubris_archive_id": String("BELgNXi52l4"),
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(344051),
    "k": String("test.ereport.please.ignore"),
    "msg": String("im dead"),
    "rev": Number(42),
}

 # taking a quick peek at Packrat's ringbuf:
eliza@theseus ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- ringbuf packrat
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.16s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- ringbuf packrat`
   0  252        1        1 RestartIdSet(Set(0xf5a3acecee5ddaa4fb7793b2f5f8ddb2))
   1  252        1        1 VpdIdentitySet(Set(VpdIdentity { part_number: [ 0x4c, 0x4f, 0x4c, 0x4e, 0x4f, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30 ], revision: 0x2a, serial: [ 0x36, 0x39, 0x34, 0x32, 0x36, 0x36, 0x36, 0x31, 0x33, 0x33, 0x37 ] }))
humility: ring buffer task_packrat::ereport::__RINGBUF in packrat:
   TOTAL VARIANT
      14 ReadRequest
      10 RestartIdMismatch
       7 MetadataEncoded
       7 Reported
       4 EreportReceived(Inserted)
       3 Flushed
  11  184        2        1 RestartIdMismatch { current_restart_id: 0xf5a3acecee5ddaa4fb7793b2f5f8ddb2 }
  12  206        2        1 MetadataEncoded { len: 0x67 }
  13  300        2        1 Reported { start_ena: 0x4, reports: 0x2, limit: 0xff }
  14  149        2        1 ReadRequest { restart_id: 0x0 }
  15  184        2        1 RestartIdMismatch { current_restart_id: 0xf5a3acecee5ddaa4fb7793b2f5f8ddb2 }
   0  206        3        1 MetadataEncoded { len: 0x67 }
   1  149        3        1 ReadRequest { restart_id: 0x56c1eefc2f0f4358b28c3b867bd7e886 }
   2  184        3        1 RestartIdMismatch { current_restart_id: 0xf5a3acecee5ddaa4fb7793b2f5f8ddb2 }
   3  206        3        1 MetadataEncoded { len: 0x67 }
   4  300        3        1 Reported { start_ena: 0x4, reports: 0x2, limit: 0xff }
   5  149        3        1 ReadRequest { restart_id: 0x0 }
   6  184        3        1 RestartIdMismatch { current_restart_id: 0xf5a3acecee5ddaa4fb7793b2f5f8ddb2 }
   7  206        3        1 MetadataEncoded { len: 0x67 }
   8  149        3        1 ReadRequest { restart_id: 0xf5a3acecee5ddaa4fb7793b2f5f8ddb2 }
   9  177        3        1 Flushed { committed_ena: 0x3, flushed: 0x0 }
  10  300        3        1 Reported { start_ena: 0x4, reports: 0x2, limit: 0xff }
```

</details>

task/packrat/src/main.rs
mkeeter: The `net` task has a similar-ish challenge, where anyone uses `task-net-api` must agree with the `net` task itself about whether VLANs are enabled.  We could do the same here: add an `ereport` feature to `task-packrat-api`, then in the API build script confirm that it agrees with the `packrat` task itself.  That wouldn't _exactly_ prevent tasks from calling the ereport functions, but it would provide some friction.

hawkw: Oh, that's a good idea --- I'd much rather be able to check this at build time.

app/cosmo/base.toml
```diff
@@ -1524,6 +1545,15 @@ port = 11113
 tx = { packets = 3, bytes = 1024 }
 rx = { packets = 3, bytes = 1024 }
 
+[config.net.sockets.ereport]
+kind = "udp"
+owner = {name = "snitch", notification = "socket"}
+port = 57005
```
mkeeter, line 1483: This is waiting for https://github.com/oxidecomputer/oana/pull/25

app/gimlet/base.toml
```diff
@@ -110,7 +110,7 @@ notifications = ["i2c2-irq", "i2c3-irq", "i2c4-irq"]
 [tasks.spd]
 name = "task-gimlet-spd"
 features = ["h753"]
-priority = 2
+priority = 4
```
mkeeter: Is this intentional?  `task-gimlet-spd` has traditionally had a high priority because it needs to reply to real-time events (I2C requests from the host).

task/packrat/src/main.rs
hawkw: Ah, hm, it's actually a bit fraught: we can add a feature flag to `task-packrat-api`, but AFAICT that feature flag can't actually control whether the idol-generated functions for the ereport APIs exist or not, so it can't actually do anything useful; you can still call those functions. This isn't a problem for the net API since the feature flag doesn't actually control the presence of specific IPC interfaces.

It seems like we could go and add support to the idol codegen for only generating an IPC's client stub if a feature flag is enabled, but I'm not immediately sure how tricky that would turn out to be...

app/grapefruit/app-dev.toml
```diff
@@ -6,3 +6,43 @@ inherit = "base.toml"
...
 interrupts = {"uart8.irq" = "usart-irq"}
+
+# Ereport stuff
+[tasks.packrat]
+stacksize = 1040
+features = ["ereport"]
+
+[tasks.snitch]
+name = "task-snitch"
+priority = 6
```
mkeeter: It looks like `net` is priority 3 in the Grapefruit image, does it matter that `snitch` isn't right above it?

mkeeter: (also, `packrat` has priority 3 here, while it's priority 1 in other manifests)

app/gimletlet/app-ereportlet.toml
```diff
@@ -0,0 +1,31 @@
+# Gimletlet Ereport test application
```
mkeeter, line 1: I don't feel strongly about this, but I'd be fine having this be part of the basic `gimletlet` image, since it's additive.

drv/stm32h7-rng/src/main.rs
mkeeter: A terrible option for resiliency would be for the RNG driver to call `generate_restart_id` once every _N_ seconds!

task/packrat/src/main.rs
hawkw: Opened https://github.com/oxidecomputer/idolatry/issues/62 for the idea of feature-flagged IPCs in Idol codegen.

app/gimletlet/app-ereportlet.toml
hawkw: Yeah. I was thinking we might not always want it to save on flash, but that seems like a problem that could be solved when it actually becomes a problem!

app/gimlet/base.toml
hawkw: agh, nope, that was definitely not intentional.

app/grapefruit/app-dev.toml
hawkw: ah, i think i just copied this from some other app.toml, it should probably be bumped up. good catch.

`packrat` had priority 3 on gimletlet, as well, and i bumped it up, i should do that on grapefruit too --- i forgor

app/psc/base.toml
```diff
@@ -111,10 +119,11 @@ notifications = ["i2c2-irq", "i2c3-irq"]
 [tasks.packrat]
 name = "task-packrat"
 priority = 3
```
hawkw: Should Packrat's priority be bumped up to 1 here?

hawkw: I guess I did everywhere else.

task/packrat-api/build.rs
```diff
@@ -2,10 +2,8 @@
...
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    idol::client::build_client_stub(
-        "../../idl/packrat.idol",
-        "client_stub.rs",
-    )?;
+fn main() -> anyhow::Result<()> {
+    idol::client::build_client_stub("../../idl/packrat.idol", "client_stub.rs")
+        .map_err(|e| anyhow::anyhow!("{e}"))?;
```
mkeeter, line 7: Is this for "correctly print error chains" purposes?  If so, I usually prefer `.map_err(anyhow::Result::from)`

hawkw: it is indeed for "correctly print error chain purposes"; IIRC the `From` conversion didn't work because the idol codegen's return type was missing `Sized` or something, lemme double check

task/packrat/src/ereport.rs
```diff
@@ -0,0 +1,405 @@
...
+            ringbuf_entry!(Trace::RestartIdMismatch {
+                current_restart_id: current_restart_id.into()
+            });
+
+            // If we don't have our VPD identity yet, don't send any metadata.
+            //
+            // We *could* include the Hubris image ID here even if our VPD
+            // identity hasn't been set, but sending an empty metadata map
+            // ensures that MGS will continue asking for metadata on subsequent
+            // requests.
```
mkeeter, line 205: Protocol question: what ensures that MGS keeps asking for metadata?

MGS always gets the new restart ID, so subsequent requests from MGS won't take this conditional (unless MGS deliberately avoids updating its restart ID if there's no metadata?)

task/packrat/src/ereport.rs
```diff
@@ -0,0 +1,405 @@
...
+        let current_restart_id =
+            self.restart_id.ok_or(EreportReadError::RestartIdNotSet)?;
+        // Skip over a header-sized initial chunk.
+        let first_data_byte = size_of::<ereport_messages::ResponseHeader>();
+
+        let mut position = first_data_byte;
+        let mut first_written_ena = None;
+
+        // Begin metadata map.
+        data.write_at(position, 0xbf)
```
mkeeter: Should we also define this as a constant (up near `CBOR_BREAK`)?

task/packrat-api/build.rs
hawkw: yeah:

```
   Compiling task-net v0.1.0 (/home/eliza/Code/oxide/hubris/task/net)
error[E0277]: the size for values of type `dyn std::error::Error + Send + Sync` cannot be known at compilation time
 --> task/packrat-api/build.rs:6:5
  |
6 | /     idol::client::build_client_stub("../../idl/packrat.idol", "client_stub.rs")
7 | |         .map_err(anyhow::Error::from)?;
  | |_____________________________________^ doesn't have a size known at compile-time
  |
  = help: the trait `Sized` is not implemented for `dyn std::error::Error + Send + Sync`, which is required by `anyhow::Error: std::convert::From<Box<dyn std::error::Error + Send + Sync>>`
  = help: the trait `std::convert::From<E>` is implemented for `anyhow::Error`
  = note: required for `Box<dyn std::error::Error + Send + Sync>` to implement `std::error::Error`
  = note: required for `anyhow::Error` to implement `std::convert::From<Box<dyn std::error::Error + Send + Sync>>`

error[E0277]: the size for values of type `dyn std::error::Error + Send + Sync` cannot be known at compilation time
 --> task/packrat-api/build.rs:7:18
  |
7 |         .map_err(anyhow::Error::from)?;
  |                  ^^^^^^^^^^^^^ doesn't have a size known at compile-time
  |
  = help: the trait `Sized` is not implemented for `dyn std::error::Error + Send + Sync`, which is required by `anyhow::Error: std::convert::From<Box<dyn std::error::Error + Send + Sync>>`
  = help: the trait `std::convert::From<E>` is implemented for `anyhow::Error`
  = note: required for `Box<dyn std::error::Error + Send + Sync>` to implement `std::error::Error`
  = note: required for `anyhow::Error` to implement `std::convert::From<Box<dyn std::error::Error + Send + Sync>>`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `task-packrat-api` (build script) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
Error: failed to build net
```

We should really improve the error types that the idol codegen stuff returns....

task/packrat/src/ereport.rs
```diff
@@ -0,0 +1,405 @@
...
+            let generation = tid.generation();
+
+            let entry = (
+                task_name,
+                u8::from(generation),
+                r.timestamp,
+                ByteGather(r.slices.0, r.slices.1),
+            );
+            let mut c =
+                minicbor::encode::write::Cursor::new(&mut self.recv[..]);
```
mkeeter, line 276: Should we implement `minicbor::write::Write` on a (wrapper type around a) lease, to avoid having to copy data twice?  It would be a decent amount of plumbing, so maybe not required in this PR...

task/packrat/src/ereport.rs
```diff
@@ -0,0 +1,405 @@
...
+            match minicbor::encode(&entry, &mut c) {
+                Ok(()) => {
+                    let size = c.position();
+                    // If there's no room left for this one in the lease, we're
+                    // done here.
```
mkeeter: I was initially concerned whether this leaves room for the final `CBOR_BREAK`.  I now _think_ it does, but adding a comment to make that explicit would be helpful.

task/packrat/src/ereport.rs
hawkw: When MGS receives a request to read ereports from a SP, if its metadata map for that SP is empty, it will send a "metadata refresh request" with restart ID 0 and a limit of 0 ereports, before it forwards the request it received:
https://github.com/oxidecomputer/management-gateway-service/blob/77e316c812aa057b9714d0d99c4a7bdd36d45be2/gateway-sp-comms/src/ereport.rs#L109-L150

MGS doesn't actually store the restart ID, that's stored by whoever is making requests to MGS. It only updates its metadata map if it receives a non-empty metadata map from the SP, or if the request's restart ID doesn't match the SP's:

https://github.com/oxidecomputer/management-gateway-service/blob/77e316c812aa057b9714d0d99c4a7bdd36d45be2/gateway-sp-comms/src/ereport.rs#L351-L360

Looking at the code again, I think the `Option` around the metadata map complicates things and it really should be checking if the actual map is empty...I think there _might_ be a bug where we don't re-request metadata if the SP restarts and then sends empty meta. I'll go fix that.

task/packrat/src/ereport.rs
```diff
@@ -0,0 +1,405 @@
...
+        vpd: &VpdIdentity,
+    ) -> Result<&[u8], MetadataError> {
+        let c = minicbor::encode::write::Cursor::new(&mut self.recv[..]);
+        let mut encoder = minicbor::Encoder::new(c);
+        // TODO(eliza): presently, this code bails out if the metadata map gets
+        // longer than our buffer. It would be nice to have a way to keep the
+        // encoded metadata up to the last complete key-value pair...
+        encoder
+            .str("hubris_archive_id")
+            .map_err(|_| MetadataError::TooLong)?
```
mkeeter: Should we just do a `From` implementation, to get rid of these `map_err` calls?
```rust
impl From<minicbor::encode::write::EndOfSlice> for MetadataError {
    fn from(_: minicbor::encode::write::EndOfSlice) -> MetadataError {
        MetadataError::TooLong
    }
}
```

task/packrat/src/ereport.rs
hawkw: that's a good idea. we should probably do that in a lib crate, because tasks that want to send ereports to packrat will probably also want to write into a lease. 

however, there is another reason for the double-buffering in this instance, which is that we're currently using it to determine how long the ereport is before copying it into the lease, so that we can put it in the lease only if it fits. since what we're doing here is "write as many ereports as fit", we don't want to error out if we run out of space when encoding. we could probably handle that by hanging onto the previous position we had written to, and not advancing it unless we wrote the whole thing.

task/packrat/src/ereport.rs
```diff
@@ -0,0 +1,405 @@
...
+
+struct ByteGather<'a, 'b>(&'a [u8], &'b [u8]);
+
+impl<C> minicbor::Encode<C> for ByteGather<'_, '_> {
+    fn encode<W: minicbor::encode::Write>(
+        &self,
+        e: &mut minicbor::Encoder<W>,
+        _ctx: &mut C,
+    ) -> Result<(), minicbor::encode::Error<W::Error>> {
+        e.bytes_len((self.0.len() + self.1.len()) as u64)?;
```
mkeeter: nit: `wrapping_add` to remove a potential panic site?

mkeeter: on the other hand, if someone manages to give us two buffers that overflow a `usize`, maybe we want to know about it!

task/packrat/src/ereport.rs
mkeeter: > we could probably handle that by hanging onto the previous position we had written to, and not advancing it unless we wrote the whole thing.

Yeah, I was thinking along those lines.  We'd need a wrapper type and a new error type which distinguishes between "lease write failed" and "ran out of space", so I'm fine punting for now.

task/packrat/src/main.rs
mkeeter: > Ah, hm, it's actually a bit fraught: we can add a feature flag to `task-packrat-api`, but AFAICT that feature flag can't actually control whether the idol-generated functions for the ereport APIs exist or not, so it can't actually do anything useful; you can still call those functions.

Yeah, that's true; I was thinking that adding / checking the flag would force people to actually think about it, but it's definitely not bulletproof.

We could always use the tried-and-true strategy of
```toml
[features]
i-have-considered-ereports-when-using-packrat-api = []
```

task/packrat/src/ereport.rs
hawkw: good idea!

task/snitch/src/main.rs
mkeeter: It would take some priority shuffling, but one way to recover from `packrat` restarts is to have `snitch` poke the RNG driver to deliver a new restart ID to packrat here!

This would require `snitch -> rng driver -> packrat` in terms of priorities, so someone could theoretically starve snitch by... spamming the RNG driver with requests (seems unlikely).

task/packrat/src/ereport.rs
hawkw: yeah, i started on this, but it's *just* complicated enough i'd kinda prefer to do it in a follow-up branch

task/packrat-api/build.rs
hawkw: opened https://github.com/oxidecomputer/idolatry/issues/63 for improvements to idol's errors which i would like to get around to Some Day

task/packrat/src/ereport.rs
hawkw: we use the `CBOR_BREAK` byte twice, but we only use the "begin map" and "begin list" bytes once, so it felt less important...i'll change them for consistency anyway, since i agree it reads nicer.

lib/snitch-core/src/lib.rs
```diff
@@ -63,6 +63,13 @@ pub struct Store<const N: usize> {
     stored_record_count: usize,
 }
 
+#[derive(Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "counters", derive(counters::Count))]
```
mkeeter, line 67: I don't see an actual `counters` features in `snitch-core/Cargo.toml`, does this actually work?  (also, I thought `rustc` had started complaining if you used features that didn't exist)

hawkw: `counters` is an optional dependency:
https://github.com/oxidecomputer/hubris/blob/89519098481e332c56a28b9aab9a53609c7d01dc/lib/snitch-core/Cargo.toml#L9
an optional dependency creates an implicit feature with the same name as the dependency

lib/snitch-core/src/lib.rs
```diff
@@ -227,6 +243,7 @@ impl<const N: usize> Store<N> {
         // full. If we are able to recover here, but do _not_ have enough bytes
         // free for the next incoming record, then we'd just kick back into loss
         // state.... necessitating another loss record. And so forth.
+        index as usize
```
mkeeter: The docstring says that it returns the number of records that were discarded, but has discarded records from `0..=index`, i.e. `index + 1` records.

Also, do we actually need this return value?  It looks like it was added in the PR, but is unused.

hawkw: Whoops, you're right about the off-by-one.

It's not totally unused: I added the return value so that we could put it in the ringbuf, here:
https://github.com/oxidecomputer/hubris/blob/89519098481e332c56a28b9aab9a53609c7d01dc/task/packrat/src/ereport.rs#L188-L192

I agree that it's not strictly necessary and we could remove it, I just wanted it for debugging.

lib/snitch-core/src/lib.rs
```diff
@@ -135,7 +136,18 @@ impl<const N: usize> Store<N> {
...
+    ///
+    /// - [`InsertResult::Inserted`] if the record was successfully inserted.
+    /// - [`InsertResult::Lost`] if the record was lost due to insufficient
+    ///   space.
+    pub fn insert(
+        &mut self,
+        sender: u16,
+        timestamp: u64,
+        data: &[u8],
+    ) -> InsertResult {
```
mkeeter, line 150: Do we use this return value anywhere (or log it to a ringbuf)?

hawkw: Yes, I added it specifically to go in the ringbuf, here:

https://github.com/oxidecomputer/hubris/blob/89519098481e332c56a28b9aab9a53609c7d01dc/task/packrat/src/ereport.rs#L126-L135

I have also mused about a potential future enhancement where we inform the task that submitted an ereport whether it was able to fit in the buffer or not, so that it can attempt it again if it was dropped. But, I would be very cautious about using that outside of unrecoverable errors like sequencing failure, where the sequencer task just goes into a "do nothing forever" loop anyway. I could imagine it periodically retrying an ereport if it's in the "i am dead now" loop.

lib/snitch-core/src/lib.rs
mkeeter: TIL!

lib/snitch-core/src/lib.rs
mkeeter: Okay, I'm fine keeping it if the off-by-one is fixed!

lib/snitch-core/src/lib.rs
mkeeter: 👍🏻 I'm fine with the current situation for now

mkeeter (approved):

task/packrat/src/ereport.rs
hawkw: Did this in #2164, there was a bit of jank due to the `minicbor` API but I think it was worth it.

Issue #1735, opened 2024-04-06
"chill out time" for tasks that crash a lot


hawkw:
If a task crashes a whole lot of times in a loop, it can potentially degrade the rest of the system, especially if we take dumps from that task. This can starve other tasks, increasing latency for other operations. 

This is a potential behavior an attacker that can force a task to crash from outside the system might exploit (I'm calling this "CrashDoS" because it sounds like HashDoS). It can also happen if a buggy task does something that panics every time it does it. 

The supervisor knows how many times it's restarted a task. Maybe a supervisor should wait a bit before restarting a task that's crashed a whole lot of times?

hawkw:
Ah, looks like @mkeeter previously opened #616 a while ago.

mkeeter:
Done in #2129

Issue #616, opened 2022-06-17
Smarter task restarts in jefe


mkeeter:
In #615, I found that a crash-looping task can permanently DDOS lower-priority tasks (in this case, `hiffy` was stuck in `_start`).

We should consider a smarter policy in the scheduler to make this more robust – whether exponential backoff or something else.

mkeeter:
Done in #2129

PR #2129, opened 2025-07-07
Add delays to `jefe` restarts

mkeeter:
In the past, we have seen rapidly-crashing tasks perform a denial-of-service attack on lower-priority tasks (#616, see also #1735).

This PR adds a `const MIN_RUN_TIME` to the `jefe` supervisor.  If a task crashes before that amount of time has elapsed, then its restart is delayed to meet that minimum.  I picked 50 ms arbitrarily, which limits a task to 20 restarts per second.

Tracking task start times adds 8 bytes of additional state per task, requiring more RAM for the supervisor (and in some cases bumping it to the next size threshold).

The PR is mean to be a starting point for design discussions, so I've marked it as a draft.  For example, iwe've talked about exponential backoff; I could also imagine wanting this policy driven by a feature flag, instead of hard-coded into the supervisor.


mkeeter:
Summarizing design discussions [from Matrix](https://matrix.oxide.computer/#/!jxnlDLnJsVaIRstnrt:oxide.computer/$mbOC2-7S8-m5ZF7b4rcJsS8O7UHG-aG_JFA2STeWSdE?via=oxide.computer&via=unix.house&via=matrix.org):

@cbiffle notes that this delay affects any task which is blocked on the restarting task.  Blocked tasks would previously receive a dead code right when the restarting task died; now, that will be delayed by up to `MIN_RUN_TIME`.

One potential fix is to tweak kernel IPCs so that unblocking callers could be done separately from starting the task.

Right now, this is half possible: we can call `kipc::restart_task(index, false)` to restart the task (notifying its callers) but holding it in `SchecdState::Stopped`.  However, there's no KIPC to start a task that has been halted in this manner.  We could add a KIPC which does this, checking to see if the task is halted and calling
```rust
    tasks[index].set_healthy_state(SchedState::Runnable);
```

We could also punt on this: most tasks will simply retry the message with the new generation number, so it doesn't really matter if they're blocked waiting for the old task to die versus waiting for the new task to start up.

task/jefe/src/main.rs
```diff
@@ -54,6 +54,12 @@ pub enum Disposition {
 // notification, but can otherwise be arbitrary.
 const TIMER_INTERVAL: u32 = 100;
 
+/// Minimum amount of time a task must run before being restarted
+///
+/// If a task runs for *less* than this amount of time before crashing, its
+/// restart is delayed to hit this value.
+const MIN_RUN_TIME: u64 = 50;
```
labbott: Can you add the (really good) math about calculating restarts per second here as well so if we think we need to tune this we have an idea about how to do so?

mkeeter: done!

PR #2130, opened 2025-07-07
Remove tabs from idl files

mkeeter:


mkeeter:
I don't see any pretty-printers / autoformatters for RON, unfortunately.

(KDL fixes this)

PR #2132, opened 2025-07-08
Check cosmo `CORETYPE` in the sequencer

labbott:
We have a bug preventing `CPU_PRESENT` from working at the moment


labbott:
see https://github.com/oxidecomputer/hardware-cosmo/issues/658

PR #2135, opened 2025-07-09
add more T6 debugging and properly stagger cld_rst_l and perst_l

nathanaelhuffman:
This helps resolve T6 not coming up reliably on cosmo.


citrus-it:
In testing, this resolves https://github.com/oxidecomputer/stlouis/issues/758

PR #2136, opened 2025-07-09
Dead codes returned by `kipc::restart_task` are of the old generation

mkeeter:
When a task is restarted by `jefe`, it calls `kipc::restart_task`.

Within the kernel, any tasks blocked on the target task receive a dead code.

Dead codes typically include the _new_ task generation, so that callers can update their task ID.  However, in this case, we return the older generation.  This adds an extra round-trip to idempotent calls:

- Task A calls into task B with generation 0
- Task B dies (RIP)
- `jefe` restarts the task, sending a dead code with generation 0 to task A (**this is the bug**)
- Task A retries with generation 0 again
- Due to generation mismatch, the call fails, and Task A gets a dead code with generation 1
- Task A retries with generation 1
- The call finally succeeds

This PR does three things:

- Updates the docs
  - Call the `kipc` operation `restart_task` instead of `reinit_task`, to match the actual code
  - Explicitly declare that the dead code from restarted tasks is of the new generation
- Fix the kernel bug
- Add a test to the kernel test suite to check this behavior
  - This required updating the test supervisor to optionally auto-restart failing tasks

### DIFF
--- a/doc/kipc.adoc
+++ b/doc/kipc.adoc
@@ -141,7 +141,8 @@ zero otherwise) and the stack erased.
 5. Any _other_ tasks that were blocked in IPC with the targeted task (either
 waiting to deliver a message, waiting for a reply to a delivered message, or
 waiting to receive) are interrupted and given a <<death,dead code>> to indicate
-that the IPC will never complete.
+that the IPC will never complete.  The dead code's generation is the *new*
+generation number for the faulted task.
 
 ==== Request
 
--- a/test/test-suite/src/main.rs
+++ b/test/test-suite/src/main.rs
@@ -106,6 +106,7 @@ test_cases! {
     test_panic,
     test_restart,
     test_restart_taskgen,
+    test_panic_taskgen,
     test_borrow_info,
     test_borrow_read,
     test_borrow_write,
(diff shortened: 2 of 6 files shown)


doc/kipc.adoc
```diff
@@ -118,14 +118,14 @@ pub enum SchedState {
 }
 ----
 
-=== `reinit_task` (2)
+=== `restart_task` (2)
```
hawkw: Turbo nitpicky, feel free to ignore: I feel like this name was chosen intentionally to avoid using the word "start", as the task may or may not actually be _started_ depending on whether `start` is `true` or `false`.

test/test-suite/src/main.rs
```diff
@@ -921,6 +922,54 @@ fn test_restart() {
...
+    let (rc, len) = userlib::sys_send(
+        assist,
+        AssistOp::FastPanic as u16,
+        &0u32.to_le_bytes(),
+        response.as_mut_bytes(),
+        &[],
+    );
+
+    // Clean up by disabling auto-restart before making our assertions
+    set_autorestart(false);
```
hawkw, line 963: I think this is _probably_ fine, though I wondered about what happens if we get faulted in the `sys_send` call and never have the chance to unset auto-restart. I guess in that case, the test suite just fails, because the main test task never gets autorestarted, and that's fine.

hawkw: It occurs to me that if we wanted to be more fastidious about this, we _could_ instead make the auto-restart behavior be "auto-restart exactly one task per `RunnerOp::AutoRestart` request", and have `set_autorestart(false)` "consume" that if it was not used. However, given the above, i really don't think this matters, so the current thing is probably fine.

doc/kipc.adoc
mkeeter: I don't feel strongly about the name, but want the documentation to match the actual function in `kipc` (which is `restart_task`).  I'd be happy to switch both to `reinit_task` if that's the consensus!

hawkw: oh, haha, i totally missed that this didn't reflect what the actual function was called. carry on!

cbiffle (approved):
The logic fix here looks great, thank you for catching this.

I'm less sold on changing the name of the syscall -- reinit is the name used inside the kernel, and it was intended to express that the _start_ part of the operation is optional. (Though perhaps not usefully so!) But I could be persuaded pretty easily if folks prefer the new name.

mkeeter:
I've now switched the name in **code** to `reinit_task` (rather than changing the docs to `restart_task`).

PR #2139, opened 2025-07-09
Make RAM non-executable on STM32Fx parts

mkeeter:
This fixes a failing test (`execdata`) in the kernel test suite on the STM32F3 Discovery Board.


hawkw (approved):
It's a bummer we don't currently have a way to run hardware-in-the-loop tests automatically in CI...

PR #2147, opened 2025-07-14
.gitignore: additional editor ignores

dancrossnyc:
`.gitignore` ignores most VSCode customizations,
but we could also ignore .zed/ and .helix/.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ itm.txt
 !.vscode/launch.json
 !.vscode/tasks.json
 !.vscode/extensions.json
+
+.zed/
+.helix/


dancrossnyc:
It turns out, there's a much better way to do this....  Via @jgallagher and @sudomateo, it turns out you can put user-specific ignores in config files that apply to all projects. So e.g., put a blanket rule ignoring per-project editor configs in $XDG_CONFIG_HOME/.git/ignore and there's no need to modify the project `.gitignore`.

dancrossnyc:
Another option is to add a `excludesfile = ...` line to the `[core]` section of one's git config to do more or less the same thing. E.g.,

```
[core]
    excludesfile = ~/.gitexclude
```

PR #2148, opened 2025-07-14
Enable averaging for MAX5970

### LINKED ISSUE #2145: Implement averaging of ring buffer values for Sharkfin HSC
Due to noise on the measurement for the 12V current on the Sharkfin HSC, please implement an averaging function using the contents of the ring buffer of voltage and current values found in the HSC.

As an added bonus, once the ring buffer data can be pulled, it can also be used to look at voltage and current waveforms right before and after a fault (see the relevant section of the MAX5970 datasheet for details on how that works)

See https://github.com/oxidecomputer/hardware-cosmo-sharkfin/issues/55 for context around why I am requesting this.

mkeeter:
Fixes #2145 by performing current readings that average samples from the MAX5970's internal ringbufs.  This is optional; I only enable it for the 12V rails on Cosmo Sharkfins, and only for current readings.  The chip samples at 1 KHz, so even taking 50 samples is far from our 1 Hz polling rate.

Tested on `alfred:cosmo-hubris-sp`.

## Before
Screenshot 2025-07-14 at 1 59 25 PM

## After
Screenshot 2025-07-14 at 1 45 16 PM

### DIFF
--- a/drv/i2c-devices/src/max5970.rs
+++ b/drv/i2c-devices/src/max5970.rs
@@ -482,12 +482,14 @@ impl Max5970 {
         let mut buf = [0u8; 100];
         self.device.read_reg_into(reg as u8, &mut buf)?;
 
-        let mut sum = 0f32;
+        // 50x 10-bit values can't overflow a u16
+        let mut sum = 0u16;
         for c in buf.chunks_exact(2) {
-            sum += self.convert_current(status, c[0], c[1])?.0;
+            sum += ((c[0] as u16) << 2) | (c[1] as u16);
         }
-        let mean = sum / (buf.len() / 2) as f32;
-        Ok(Amperes(mean))
+        let mean = sum / 50;
+        let [msb, lsb] = mean.to_be_bytes();
+        self.convert_current(status, msb, lsb)
     }
 }
 


task/power/src/bsp/cosmo_a.rs
```diff
@@ -345,7 +345,7 @@ impl State {
         .enumerate()
         {
             let (dev, rail) = (builder)(i2c_task);
-            let m = Max5970::new(&dev, rail, Ohms(0.005));
+            let m = Max5970::new(&dev, rail, Ohms(0.005), true);
```
labbott, line 348: It feels really weird to set something like this in two places but I think it's just the way the tweak is working

labbott (approved):

drv/i2c-devices/src/max5970.rs
```diff
@@ -277,14 +284,27 @@ pub struct Max5970 {
...
+    /// If enabled, then current readings return an averaged value
+    avg: bool,
+
+    /// Indicates that 10-bit mode is enabled for averaging
+    ///
+    /// When the chip turns on, its internal ringbuf stores temperatures as
+    /// single-byte values.  We want to get the full 10-bit (2-byte) values,
+    /// which requires changing a register.  This flag tells us whether we've
+    /// made that change; it's a `Cell` because current reading takes `&self`.
+    avg_mode_enabled: core::cell::Cell<bool>,
```
hawkw: the relationship between these is a little unclear from the comments --- can we rephrase it to make it a little clearer that `avg` means the driver was _configured_ to enable averaging mode, and `avg_mode_enabled` means that we actually wrote to the register to turn it on?

drv/i2c-devices/src/max5970.rs
```diff
@@ -416,11 +436,59 @@ impl Max5970 {
...
+            Register::cbuf_ba_ch1_i
+        };
+        let mut buf = [0u8; 100];
+        self.device.read_reg_into(reg as u8, &mut buf)?;
+
+        let mut sum = 0f32;
+        for c in buf.chunks_exact(2) {
+            sum += self.convert_current(status, c[0], c[1])?.0;
+        }
+        let mean = sum / (buf.len() / 2) as f32;
```
hawkw: I note we are calculating the mean as an integer and then turning it to a float — presumably that's to avoid costly floating point ops on v7m?

hawkw: also, super ultra nitpicky and unimportant: i wonder if `buf.len() / 2` gets const-folded here? i would kind of guess it does, but...

drv/i2c-devices/src/max5970.rs
mkeeter: Good call, I've switched to `read_averaged_current` and `avg_config_done`.

drv/i2c-devices/src/max5970.rs
mkeeter: We are actually calculating the mean as a floating-point value!  It may be wiser to compute it as an integer, though...

hawkw: Oh. Operator precedence, yeah. I thought the cast applied to the whole expression because I can't read.

mkeeter: I switched to integer math in bb65a651b030, and stuff still looks fine:
Screenshot 2025-07-15 at 4 28 10 PM

PR #2151, opened 2025-07-15
xtask: Override default caboose version from env

hawkw:
When flashing an engineering Hubris image on an SP that will participate
in a control-plane driven software update, it is necessary to populate
the `VERS` key in the caboose must be present so that the update
machienry doesn't get confused. If it is desirable for the engineering
image to not be overwritten during an update, the verson must also be
set to the version expected by the system. Being able to do this is
useful for testing Hubris changes in integration with the rest of the
system.

Historically, such images have been created by hacking up the `xtask
dist` code to write the desired version string when creating the default
caboose. I thought that this seemed like a common enough thing to want
to do that there should just be a way to ask for it without having to
change the build system. Therefore, this branch adds a
`HUBRIS_CABOOSE_VERS` env var that will override the `VERS` key in the
caboose if building an image that has a default caboose.

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -328,8 +328,29 @@ pub fn package(
     app_toml: &Path,
     tasks_to_build: Option<Vec<String>>,
     dirty_ok: bool,
+    caboose_args: super::CabooseArgs,
 ) -> Result<BTreeMap<String, AllocationMap>> {
     let cfg = PackageConfig::new(app_toml, verbose, edges)?;
+
+    // If an overridden caboose version was given, it will only be used if the
+    // config requests a default caboose.
+    if let Some(ref vers) = caboose_args.version_override {
+        let Some(ref caboose) = &cfg.toml.caboose else {
+            bail!(
+                "an overridden default caboose version was requested \
+                 (HUBRIS_CABOOSE_VERS={vers:?}), but {} does not specify a \
+                 caboose",
+                app_toml.display(),
+            );
+        };
+        anyhow::ensure!(
+            caboose.default,
+            "the default caboose version is overriden \
+             (HUBRIS_CABOOSE_VERS={vers:?}), but {} does not specify a
+             default caboose",
+            app_toml.display(),
+        );
+    }
 
     // Verify that our dump configuration is correct (or absent)
     check_dump_config(&cfg.toml)?;
@@ -630,15 +651,23 @@ pub fn package(
         let archive_name = build_archive(&cfg, image_name, raw_image)?;
 
         // Post-build modifications: populate a default caboose if requested
-        if let Some(caboose) = &cfg.toml.caboose {
+        if let Some(ref caboose) = cfg.toml.caboose {
             if caboose.default {
                 let mut archive =
                     hubtools::RawHubrisArchive::load(&archive_name)
                         .context("loading archive with hubtools")?;
+                if let Some(ref vers) = caboose_args.version_override {
+                    println!(
+                        "note: asked to override default caboose `VERS` to \
+                         {vers:?}"
+                    );
+                }
                 // The Git hash is included in the default caboose under the key
                 // `GITC`, so we don't include it in the pseudo-version.
                 archive
-                    .write_default_caboose(None)
+                    .write_default_caboose(
+                        caboose_args.version_override.as_ref(),
+                    )
                     .context("writing caboose into archive")?;
                 archive.overwrite().context("overwriting archive")?;
             }
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -332,26 +332,6 @@ pub fn package(
 ) -> Result<BTreeMap<String, AllocationMap>> {
     let cfg = PackageConfig::new(app_toml, verbose, edges)?;
 
-    // If an overridden caboose version was given, it will only be used if the
-    // config requests a default caboose.
-    if let Some(ref vers) = caboose_args.version_override {
-        let Some(ref caboose) = &cfg.toml.caboose else {
-            bail!(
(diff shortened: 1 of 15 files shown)


build/xtask/src/main.rs
```diff
@@ -224,6 +238,20 @@ pub struct HumilityArgs {
...
+    /// written.
+    ///
+    /// This is intended to be used when an engineering image must be
+    /// flashed in an environment that expects a particular caboose version.
+    ///
+    /// Overriding the default caboose version string is only permitted if
+    /// the app.toml specifies the default caboose.
+    #[clap(env = "HUBRIS_CABOOSE_VERS")]
+    version_override: Option<String>,
+}
```
hawkw: I did this as a separate struct with `#[clap(flatten)]` not because I anticipated adding more arguments to it later, but just because I felt that this was nicer than copy-pasting the doc comment into each subcommand --- this way, we can change the help text in one place.

This kinda made me want to factor out some of the other common arguments that are shared across several tasks in a similar way, but I thought that was best left to a separate commit.

build/xtask/src/dist.rs
```diff
@@ -328,9 +328,30 @@ pub fn package(
...
+            );
+        };
+        anyhow::ensure!(
+            caboose.default,
+            "the default caboose version is overriden \
+             (HUBRIS_CABOOSE_VERS={vers:?}), but {} does not specify a
+             default caboose",
+            app_toml.display(),
+        );
+    }
```
hawkw: This is checked early on, rather than when we actually write the caboose, so that we can avoid building all the tasks and stuff only to throw it away because we didn't like the arguments. This does mean we have to `if let` on the presence of the caboose and such in two places, but it felt worth it to not spend a bunch of time building an image before telling you that you asked for something we don't know how to do.

build/xtask/src/dist.rs
```diff
@@ -630,15 +651,23 @@ pub fn package(
         let archive_name = build_archive(&cfg, image_name, raw_image)?;
 
         // Post-build modifications: populate a default caboose if requested
-        if let Some(caboose) = &cfg.toml.caboose {
+        if let Some(ref caboose) = cfg.toml.caboose {
```
hawkw: Perhaps we could also make the presence of the argument _create_ a default caboose if there isn't any caboose section in the config, but I didn't need that, so I didn't do it. :woman_shrugging:

build/xtask/src/dist.rs
```diff
@@ -328,9 +328,30 @@ pub fn package(
...
+                app_toml.display(),
+            );
+        };
+        anyhow::ensure!(
+            caboose.default,
+            "the default caboose version is overriden \
+             (HUBRIS_CABOOSE_VERS={vers:?}), but {} does not specify a
+             default caboose",
+            app_toml.display(),
+        );
```
hawkw: It probably would also be possible to make the env var override a version key in a non-default caboose, but (a) I didn't need that and therefore I didn't do it, and (b) it seemed safer to bail in that case --- I didn't want to create the opportunity for a stray `export HUBRIS_CABOOSE_VERS` to produce a screwed up release image.

build/xtask/src/main.rs
```diff
@@ -302,7 +350,14 @@ fn run(xtask: Xtask) -> Result<()> {
...
             dirty,
         } => {
-            let allocs = dist::package(verbose >= 2, false, &cfg, None, dirty)?;
+            let allocs = dist::package(
+                verbose >= 2,
+                false,
+                &cfg,
+                None,
+                dirty,
+                CabooseArgs::default(),
```
mkeeter: I think this should take `caboose_args`, because it's a superset of `dist` (producing a usable archive).

hawkw: Ah, yeah, you're right, the caboose version might change the caboose size, and you might also want to run sizes to get an image _and_ print its size. Will change that.

build/xtask/src/main.rs
labbott: I had originally envisioned this as being independent of any argument but I think this approach is actually the right one.

I think you've also stumbled on some historical detritus. I'm not sure of the value in having a caboose specified but not giving it a default value, especially since we now have his everywhere except 2(?) places This feels like some hold over before we had figured out our release process.  If we're going to do some refactoring around this area I think making the caboose required would also simplify a lot of things

mkeeter: > I think you've also stumbled on some historical detritus. I'm not sure of the value in having a caboose specified but not giving it a default value, especially since we now have his everywhere except 2(?) places

For historical context, I was trying to decouple the existence of the caboose (as a reserved chunk of flash in the Hubris build system) from its contents (generated post-build by unknown actors).

Quoth [the docs](https://hubris.oxide.computer/reference/#caboose):

> Other than reserving the two words above, the Hubris build system is _deliberately agnostic_ to the contents of the caboose; it is left to a separate release engineering process to decide what to store there. 

In practice, no one is using it besides us, and we never do anything except the default caboose, so I'd be open to some kind of simplification.

hawkw: I'm happy to make additional simplifications here, or in another branch. If we're always setting a default caboose, it sounds like the best thing to do is to just always have a default value for the caboose and remove the `caboose.default = (true|false)` TOML key? Given that we only have two app.tomls that don't include a default caboose and they're both dev boards, it seems like omitting it was probably an oversight.

Any objections to doing that in this PR, or should I do it separately?

hawkw: Opened  #2156 for just always doing the default caboose. I'm happy to merge it separately, or merge it into this branch.

hawkw: At @labbott's urging I just went and merged that into this PR.

labbott (approved):

PR #2153, opened 2025-07-15
Check for host flash hash running before initing hash block

labbott:
If the control plane sends multiple `StartHostFlashHash` messages it is possible to re-init the hash block while it is in progress.


drv/cosmo-hf/src/hf.rs
```diff
@@ -615,9 +615,6 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         dev: HfDevSelect,
     ) -> Result<(), RequestError<HfError>> {
         self.drv.check_flash_mux_state()?;
-        if self.hash.task.init_sha256().is_err() {
-            return Err(HfError::HashError.into());
-        }
 
         match self.hash.state {
```
hawkw, line 621: perhaps worth a comment on why we do this in that order?

hawkw (approved):

PR #2154, opened 2025-07-16
Use 16-channel virtual mux on Cosmo

mkeeter:
Right now, the Cosmo FPGAs (main and front) both pretend to contain a set of PCA9545 muxes.  This PR switches to a **single** virtual 16-channel mux, which reduces writes (since we no longer need to deactivate virtual mux _N_ before activating virtual mux _M_).

Having a single mux also makes it more logically consistent to have a single reset pin, see https://github.com/oxidecomputer/hardware-cosmo/issues/619

Finally, it gives us a place to put the delay required by the physical switches' timing that _doesn't_ mess with our existing driver.

~So far, this PR only has Hubris changes; FPGA updates to come.~


app/cosmo/base.toml
```diff
@@ -634,8 +584,8 @@ removable = true
 # SMBUS_FPGA2_TO_CEMD
 [[config.i2c.devices]]
 bus = "front"
-mux = 2
-segment = 1
+mux = 1
+segment = 4
```
labbott, line 588: Why does this start at segment 4?

mkeeter: I don't understand the question; U.2s `A` through `C` U.2 are on segment 1-3 (above), and `D` is on segment 4 (here)

labbott: Question withdrawn once I hit the expand button to see more (thanks github)

PR #2156, opened 2025-07-16
build: Always include default caboose

hawkw:
As discussed [here][1] and in subsequent comments, we include the default metadata values in every image with a caboose, so we may as well just not make it optional. Doing so lets us remove the `caboose.default = (true|false)` TOML key. This simplifies some things in the build system. I've also updated the docs to reflect this.

[1]: https://github.com/oxidecomputer/hubris/pull/2151#discussion_r2210530056

Depends on #2151


doc/guide/caboose.html
```diff
@@ -0,0 +1,646 @@
+<!DOCTYPE html>
```
labbott: Did you intend to check this in? I was briefly shocked by the diffstat

labbott (commented):
I'd vote for adding this to the other PR fwiw but having it separate is fine too

doc/guide/caboose.html
hawkw: agh, whoops --- i staged the wrong file by mistake. will fix!

PR #2158, opened 2025-07-17
Flash chassis LED on Cosmo

mkeeter:
To help debug https://github.com/oxidecomputer/hubris/issues/2157 , we'd like to have some indication that the Hubris kernel is running.

PR #2159, opened 2025-07-17
Bump Cosmo `control-plane-agent` stack size

mkeeter:
This isn't detected by the stack size check because `serde` has recursive calls.

```
humility: attached to dump
system time = 123546
20 control_plane_agent          0   8 FAULT: stack overflow; sp=0x2402ffd8 (was: ready)
   could not read registers: read of 32 bytes from invalid address: 0x2402ffd8
   guessing at stack trace using saved frame pointer
   |
   +--->  0x24031870 0x08054b98 <hubpack::de::SeqAccess as serde::de::SeqAccess>::next_element_seed
                     @ /crates.io/hubpack-0.1.2/src/de.rs:280
```

Marking it as a draft, pending testing on hardware (as part of #2158)

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -268,7 +268,7 @@ notifications = ["socket"]
 [tasks.control_plane_agent]
 name = "task-control-plane-agent"
 priority = 8
-stacksize = 6256
+stacksize = 7000
 start = true
 uses = ["usart1"]
 task-slots = [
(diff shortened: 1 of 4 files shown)


mkeeter:
> Is it possible that we need to make a similar change on Gimlet? control-plane-agent code is basically the same on both compute sleds...

yeah, I'll do so, since this is particularly rough failure mode

PR #2161, opened 2025-07-18
Decouple front-io functionality from sidecar-seq

Aaron-Hartwig:
All the logic about how to interact with the front I/O board (power control, FPGA programming/communication, I2C comms) was very tightly coupled with Sidecar's business logic. The goal of this PR is to separate that stuff out into a dedicated front-io-server which can then handle all of the relevant details while exposing an API for other tasks. The big forcing factor here is working on the Medusa design (finally). Switching the medusa app to utilizing the front-io-server will be a separate PR.


Aaron-Hartwig:
This commit is deployed to `SW1` on `london`. I was able to ignition cycle the sidecar and bounce its scrimlet and everything came back up happily.

PR #2164, opened 2025-07-18
`minicbor-lease` adapter for encoding straight into a lease

hawkw:
Based on [a suggestion][1] from @mkeeter, this branch adds an adapter that implements `minicbor::encode::write::Write` for a cursor into a writable lease. This lets us avoid double-buffering in the `read_ereports` path, where we would previously encode each ereport into the receive buffer, and then copy the contents of that into the lease if there is space for it. The new approach avoids the memcpy, and also doesn't encode the ereport in the case where there isn't space left for it.

I implemented the lease-writer adapter thing in a separate `minicbor-lease` crate, as I was hoping it would be useful elsewhere...but I'm not actually sure if it will be, now that I think of it: the IPC for delivering an ereport to packrat leases the data _from_ the caller, so callers will just encode into a normal buffer they own. Ah well.

[1]: https://github.com/oxidecomputer/hubris/pull/2126#discussion_r2198529251

### DIFF
--- a/lib/minicbor-lease/Cargo.toml
+++ b/lib/minicbor-lease/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "minicbor-lease"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+minicbor = { workspace = true }
+idol-runtime = { workspace = true }
+
+[lints]
+workspace = true
--- a/lib/minicbor-lease/src/lib.rs
+++ b/lib/minicbor-lease/src/lib.rs
@@ -9,9 +9,10 @@
 
 /// An adapter implementing [`minicbor::encode::write::Write`] for
 /// [`idol_runtime::Leased`] byte buffers.
-pub struct LeasedWriter<A> {
-    lease: idol_runtime::Leased<A, [u8]>,
+pub struct LeasedWriter<'lease, A> {
+    lease: &'lease mut idol_runtime::Leased<A, [u8]>,
     pos: usize,
+    ran_out_of_space: bool,
 }
 
 #[derive(Copy, Clone, PartialEq)]
@@ -20,7 +21,7 @@ pub enum Error {
     EndOfLease,
 }
 
-impl minicbor::encode::write::Write for LeasedWriter<A>
+impl minicbor::encode::write::Write for LeasedWriter<'_, A>
 where
     A: idol_runtime::AttributeWrite,
 {
@@ -29,7 +30,8 @@ where
     fn write_all(&mut self, buf: &[u8]) -> Result<(), Self::Error> {
         let end = self.pos + buf.len();
         if end >= self.lease.len() {
-            return Error::EndOfLease;
+            self.ran_out_of_space = true;
+            return Err(Error::EndOfLease);
         }
         self.lease
             .write_range(self.pos..end, buf)
@@ -41,12 +43,12 @@ where
     }
 }
 
-impl<A> LeasedWriter<A>
+impl<A> LeasedWriter<'_, A>
 where
     A: idol_runtime::AttributeWrite,
 {
     /// Returns a new `LeasedWriter` starting at byte 0 of the lease.
-    pub fn new(lease: idol_runtime::Leased<A, [u8]>) -> Self {
+    pub fn new(lease: &mut idol_runtime::Leased<A, [u8]>) -> Self {
         Self { lease, pos: 0 }
     }
 
@@ -55,7 +57,7 @@ where
     ///
     /// This is intended for cases where some data has already been written to
     /// the lease.
-    pub fn starting_at(lease: idol_runtime::Leased<A, [u8]>) -> Self {
+    pub fn starting_at(lease: &mut idol_runtime::Leased<A, [u8]>) -> Self {
         Self { lease, pos: 0 }
     }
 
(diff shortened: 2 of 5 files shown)


hawkw:
@mkeeter Not sure whether you'd prefer to review this as its own PR or if I should just go ahead and merge it straight into #2126 --- whatever you'd prefer works for me!

mkeeter:
I'd vote to keep it separate!  Review to come (either this afternoon or Monday)

lib/minicbor-lease/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "minicbor-lease"
+version = "0.1.0"
+edition = "2021"
```
mkeeter: Should we switch to 2024?  It's ✨the future✨

lib/minicbor-lease/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// The other side of the lease has gone away.
+    WentAway,
+    /// Data could not be written as there was no room left in the lease.
+    EndOfLease,
+}
+
+/// Errors returned by the [`minicbor::encode::write::Write`] implementation for
+/// [`LeasedWriter`].
+#[derive(Copy, Clone, PartialEq)]
+pub struct WriteError(());
```
mkeeter: Any reason not to do
```suggestion
pub struct WriteError;
```
?

lib/minicbor-lease/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+        self.lease
+    }
+
+    /// Returns `true` if the last `write_all` call to return an error failed
+    /// due to running out of space.
+    ///
+    /// This is an unfortunate workaround for a limitation of the `minicbor`
+    /// API: the errors our `Write` implementation can return are wrapped in an
+    /// `encode::Error`, and we can't actually get our errors *out* of that
+    /// wrapper, so there's no way to tell whether the error was becasue we ran
```
mkeeter:
```suggestion
    /// wrapper, so there's no way to tell whether the error was because we ran
```

lib/minicbor-lease/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+pub struct WriteError(());
+
+impl<A> minicbor::encode::write::Write for LeasedWriter<'_, A>
+where
+    A: idol_runtime::AttributeWrite,
+{
+    type Error = WriteError;
+
+    fn write_all(&mut self, buf: &[u8]) -> Result<(), Self::Error> {
+        let end = self.pos + buf.len();
```
mkeeter: Are we sensitive to panic sites here?  I don't feel strongly about it, but you could use checked addition (since if you overflow a `usize`, you're definitely out of space)

lib/minicbor-lease/Cargo.toml
hawkw: I believe the compiler pinned in `rust-toolchain.toml` is too old for edition 2024, and will have to be updated. But lemme double check.

hawkw: Yup: 
```console
eliza@theseus ~/Code/oxide/hubris $ cargo xtask build app/gimletlet/app-ereportlet.toml packrat
error: failed to load manifest for workspace member `/home/eliza/Code/oxide/hubris/lib/minicbor-lease`
referenced by workspace at `/home/eliza/Code/oxide/hubris/Cargo.toml`

Caused by:
  failed to parse manifest at `/home/eliza/Code/oxide/hubris/lib/minicbor-lease/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.83.0-nightly (a9a418d1a 2024-09-15)).
  Consider adding `cargo-features = ["edition2024"]` to the top of Cargo.toml (above the [package] table) to tell Cargo you are opting in to use this unstable feature.
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

It would be nice to update to a newer nightly and switch to the 2024 edition, but I would rather not do that as part of _this_ PR.

mkeeter: Agreed, carry on!

lib/minicbor-lease/src/lib.rs
hawkw: Yeah, that's not a bad idea!

hawkw:
@mkeeter opened an upstream issue to make the minicbor API less annoying: https://github.com/twittner/minicbor/issues/36

Cargo.toml
```diff
@@ -163,6 +163,10 @@ tlvc-text = { git = "https://github.com/oxidecomputer/tlvc", default-features =
 transceiver-messages = { git = "https://github.com/oxidecomputer/transceiver-control/", default-features = false }
 vsc7448-pac = { git = "https://github.com/oxidecomputer/vsc7448", default-features = false }
 
+[patch.crates-io]
+# See https://gitlab.com/robigalia/ssmarshal/-/merge_requests/2
+ssmarshal = { git = "https://gitlab.com/de-vri-es/ssmarshal", rev = "10e90cbe389c8f07c52815857551a051946881ab" }
```
hawkw, line 168: This is necessary due to `ssmarshal` no longer compiling with Rust 2024, as when the edition is Rust 2024, `serde` will require `core::error::Error` implementations for error types, and `ssmarshal` doesn't have those.

Since the last commit to ssmarshal's `main` branch was 8 years ago, I'm not convinced this PR gets merged any time soon. It would probably be worth getting rid of our ssmarshal deps at some point, since Hubpack is basically the same thing but more featureful...

lib/minicbor-lease/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+
+/// An adapter implementing [`minicbor::encode::write::Write`] for
+/// [`idol_runtime::Leased`] byte buffers.
+pub struct LeasedWriter<'lease, A>
+where
+    A: idol_runtime::AttributeWrite,
+{
+    lease: &'lease mut idol_runtime::Leased<A, [u8]>,
+    pos: usize,
+    ran_out_of_space: bool,
```
mkeeter: Is this still needed, now that we can get the inner error?  I see it written, but no one ever actually checks it.

task/packrat/src/ereport.rs
```diff
@@ -143,41 +142,42 @@ impl EreportStore {
...
+            first_data_byte,
+            &mut data,
+        ));
+
+        fn handle_encode_err(
+            err: minicbor::encode::Error<minicbor_lease::Error>,
+        ) -> RequestError<EreportReadError> {
+            // These should always be write errors; everything we write should
+            // always encode successfully.
+            match err.into_write().unwrap_lite() {
```
mkeeter: This makes me nervous: if for some reason we try to send a message which _doesn't_ encode properly, it kills `packrat` instead of the caller.  I'd vote to instead return `ClientError::BadMessageContents` (with a comment saying it shouldn't ever happen), since `packrat` is supposed to remain alive under all circumstances.

hawkw: Hmm, that's fair, I'll make it do that.

mkeeter (approved):

PR #2174, opened 2025-07-29
Check for DIMM failures in cosmo

labbott:
check the various DIMM PCAMP signals


task/cosmo-spd/src/main.rs
```diff
@@ -31,6 +31,7 @@ enum Trace {
     Ready,
     Present { index: usize, present: bool },
     TemperatureReadTimeout { index: usize, pos: usize },
+    DimmFailure { index: usize },
```
hawkw, line 34: What do you think about adding `#[derive(Count)]` to the `Trace` enum, and changing the `ringbuf!` to a `counted_ringbuf!`? It seems it would be nice to record a count of DIMM failures so that we can still see the total number that occurred over the entire lifetime of the SPD task, even if they get pushed off the end of the ringbuf by other stuff.

task/cosmo-spd/src/main.rs
```diff
@@ -263,6 +264,15 @@ impl idol_runtime::NotificationHandler for ServerImpl {
             }};
         }
 
+        macro_rules! check_dimm_failure {
+            ($dev:ident, $index:expr) => {
+                let $dev = self.dimms.dimm_pcamp.$dev();
+                if !$dev {
+                    ringbuf_entry!(Trace::DimmFailure { index: $index });
```
hawkw: Eventually, we'll definitely want to record an ereport here as well --- I opened an issue for that: #2173

task/cosmo-spd/src/main.rs
```diff
@@ -263,6 +264,15 @@ impl idol_runtime::NotificationHandler for ServerImpl {
...
+                }
+            };
+        }
```
hawkw: It's a bit odd to me that the macro also generates the `if` block that records the ringbuf event, since the ringbuf event just includes the `index` variable that we already have as a variable in the outer scope. There's nothing about that code that actually differs based on the index, but we're generating a separate `if` block for each one. Perhaps the compiler is smart enough to figure out that it's the same code and not generate a whole bunch of branches, but I'm not sure if I trust it --- if it doesn't, this is a lot more code than we need to generate.

What if we simplified the macro to just return the result of the call to `dimm_pcamp.$dev()`, so we could have one `if` block outside the macro code?
```suggestion
        macro_rules! check_dimm_failure {
            ($dev:ident) => {
                self.dimms.dimm_pcamp.$dev()
            };
        }
```

Then, we'd change the `match` like this:
```rust
                let dev = match index {
                    0 => {
                        check_dimm_failure!(pcamp_a)
                    }
                    1 => {
                        check_dimm_failure!(pcamp_b)
                    }
                    2 => {
                        check_dimm_failure!(pcamp_c)
                    }
                    3 => {
                        check_dimm_failure!(pcamp_d)
                    }
                    4 => {
                        check_dimm_failure!(pcamp_e)
                    }
                    5 => {
                        check_dimm_failure!(pcamp_f)
                    }
                    6 => {
                        check_dimm_failure!(pcamp_g)
                    }
                    7 => {
                        check_dimm_failure!(pcamp_h)
                    }
                    8 => {
                        check_dimm_failure!(pcamp_i)
                    }
                    9 => {
                        check_dimm_failure!(pcamp_j)
                    }
                    10 => {
                        check_dimm_failure!(pcamp_k)
                    }
                    11 => {
                        check_dimm_failure!(pcamp_l)
                    }
                    _ => unreachable!(),
                };
                if !dev {
                    rustingbuf_entry!(Trace::DimmFailure { index });
                }
```

labbott: It turns out with this approach it's just as easy to not have the macro

hawkw: ...oh, yeah, of course it is, haha. Can't believe I didn't realize that. New code looks great!

rmustacc:
Is there an assumption that something else has already taken corrective action on the box here? If we lose a DIMM regulator power good, which I realize is only one part of the PCAMP situation, but is the default behavior based on the PMIC situation, then that implies that we need to MAPO, capture information, and transition to A2. Is this merely logging here because the FPGA has issued the MAPO? Running with a failed DIMM here is not helpful to anyone and the host is going to panic in mere moment after this if we actually have this failure.

Related, if we do, is the normal hot-swap logic that we have in A2 going to actually handle clearing faults and related?

labbott:
Keeping this as a draft until we finalize how we want to handle our response/reporting

PR #2176, opened 2025-07-30
Add sequencer irq pin and fix nic state reporting

nathanaelhuffman:
This wires out the sequencer irq stuff to a pin that goes to the SP and fixes the A0HP reporting before the nic was out of reset.


labbott (approved):
LGTM

PR #2181, opened 2025-08-04
Introduce cosmo sequencer interrupts

labbott:
Instead of polling for our state changes, rely on the sequencer to deliver interrupts to us

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -74,6 +74,12 @@ port = "F"
 pin = 2
 owner = {name = "sprot", notification = "rot_irq"}
 
+[tasks.sys.config.gpio-irqs.seq_irq]
+# FPGA1_TO_SP_IRQ1_L
+port = "F"
+pin = 5
+owner = {name = "cosmo_seq", notification = "seq_irq"}
+
 [tasks.spi2_driver]
 name = "drv-stm32h7-spi-server"
 priority = 3
@@ -175,7 +181,7 @@ stacksize = 2600
 start = true
 task-slots = ["sys", "i2c_driver", {spi_front = "spi3_driver"}, "jefe", "packrat", "auxflash", "spartan7_loader", "hf"]
 uses = ["mmio_sequencer", "mmio_info"]
-notifications = ["timer", "vcore"]
+notifications = ["timer", "vcore", "seq-irq"]
 
 [tasks.ignition_flash]
 name = "drv-ignition-flash"
--- a/drv/cosmo-seq-server/src/main.rs
+++ b/drv/cosmo-seq-server/src/main.rs
@@ -70,6 +70,11 @@ enum Trace {
         our_state: PowerState,
         seq_state: Result<fmc_periph::A0Sm, u8>,
     },
+    SequencerInterrupt {
+        our_state: PowerState,
+        seq_state: Result<fmc_periph::A0Sm, u8>,
+        ifr: fmc_periph::IfrView,
+    },
     PowerDownError(drv_cpu_seq_api::SeqError),
     Coretype {
         coretype0: bool,
@@ -85,6 +90,9 @@ enum Trace {
         pwrokn: u8,
     },
     Thermtrip,
+    A0MapoInterrupt,
+    SmerrInterrupt,
+    UnexpectedInterrupt,
 }
 counted_ringbuf!(Trace, 128, Trace::None);
 
@@ -132,6 +140,8 @@ const SP5R1_PULL: sys_api::Pull = sys_api::Pull::None;
 const SP5R2_PULL: sys_api::Pull = sys_api::Pull::None;
 const SP5R3_PULL: sys_api::Pull = sys_api::Pull::None;
 const SP5R4_PULL: sys_api::Pull = sys_api::Pull::None;
+
+use gpio_irq_pins::SEQ_IRQ;
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -222,6 +232,10 @@ fn init() -> Result<ServerImpl, SeqError> {
     sys.gpio_configure_input(SP5_TO_SP_SP5R2, SP5R2_PULL);
     sys.gpio_configure_input(SP5_TO_SP_SP5R3, SP5R3_PULL);
     sys.gpio_configure_input(SP5_TO_SP_SP5R4, SP5R4_PULL);
+
+    // Sequencer interrupt
+    sys.gpio_configure_input(SEQ_IRQ, sys_api::Pull::None);
+    sys.gpio_irq_configure(notifications::SEQ_IRQ_MASK, sys_api::Edge::Falling);
 
     let spi_front = drv_spi_api::Spi::from(SPI_FRONT.get_task_id());
     let aux = drv_auxflash_api::AuxFlash::from(AUXFLASH.get_task_id());
@@ -390,6 +404,7 @@ impl ServerImpl {
     fn log_state_registers(&self) -> StateMachineStates {
         let seq_api_status = (&self.seq.seq_api_status).into();
         let nic_api_status = (&self.seq.nic_api_status).into();
+
         ringbuf_entry!(Trace::RegStateValues {
             seq_api_status,
             seq_raw_status: (&self.seq.seq_raw_status).into(),
@@ -513,6 +528,7 @@ impl ServerImpl {
                     }
                 };
 
+                self.enable_sequencer_interrupts();
                 // Flip the host flash mux so the CPU can read from it
                 // (this is secretly infallible on Cosmo, so we can unwrap it)
                 self.hf.set_mux(drv_hf_api::HfMuxState::HostCPU).unwrap();
@@ -521,6 +537,8 @@ impl ServerImpl {
             | (PowerState::A0PlusHP, PowerState::A2)
             | (PowerState::A0Thermtrip, PowerState::A2)
             | (PowerState::A0Reset, PowerState::A2) => {
+                // Disable our interrupts before we shutdown
+                self.disable_sequencer_interrupts();
                 self.seq.power_ctrl.modify(|m| m.set_a0_en(false));
                 let mut okay = false;
                 for _ in 0..200 {
@@ -543,7 +561,6 @@ impl ServerImpl {
                     // We can't do much else here, since we already cleared the
                     // a0_en flag to disable the sequencer.
                 }
-
                 // Flip the host flash mux so the SP can read from it
                 // (this is secretly infallible on Cosmo, so we can unwrap it)
                 self.hf.set_mux(drv_hf_api::HfMuxState::SP).unwrap();
@@ -571,13 +588,14 @@ impl ServerImpl {
 
     /// Returns the current timer interval, in milliseconds
     ///
-    /// If we are in `A0`, then we are waiting for the NIC to come up; if we are
-    /// in `A0PlusHP`, we're polling for a thermtrip or for someone disabling
-    /// the NIC.  In other states, there's no need to poll.
+    /// If we are in `A0`, then we are waiting for the NIC to come up;
+    /// Once we are in `A0PlusHP` we rely on sequencer interrupts for
+    /// all our state transitions. We still want to catch an unexpected
+    /// case of sequencer failure so poll for that case specifically.
     fn poll_interval(&self) -> Option<u32> {
         match self.state {
             PowerState::A0 => Some(10),
-            PowerState::A0PlusHP => Some(100),
+            PowerState::A0PlusHP => Some(1000),
             _ => None,
         }
     }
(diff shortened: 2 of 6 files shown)


labbott:
Marking as draft because I expect we'll want to discuss this some

drv/cosmo-seq-server/src/main.rs
```diff
@@ -662,45 +760,6 @@ impl NotificationHandler for ServerImpl {
...
-                });
-                self.log_pg_registers();
-
-                // Power down to A2, updating our internal state.  We can't
-                // handle errors here, so log them and continue.
-                if let Err(e) = self
-                    .set_state_impl(PowerState::A2, StateChangeReason::Other)
-                {
-                    ringbuf_entry!(Trace::PowerDownError(e))
-                }
```
labbott: I like the idea of dropping the polling in the sequencer entirely but I think we _might_ lose this case if we don't get a specific interrupt

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +597,89 @@ impl ServerImpl {
...
+        ringbuf_entry!(Trace::SequencerInterrupt {
+            our_state: self.state,
+            seq_state: state.seq,
+        });
+
+        if ifr.thermtrip {
+            self.seq.ifr.modify(|h| h.set_thermtrip(false));
+            ringbuf_entry!(Trace::Thermtrip);
+            self.set_state_internal(PowerState::A0Thermtrip)
+            // this is a terminal state (for now)
```
aapoalas: **question:** Is this line unreachable? Or does control return here and then fall through to the other if's?

If not, I'd suggest an `unreachable_unchecked` here for both reader's and compiler's benefit. If yes, should there be an early return here instead?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +597,89 @@ impl ServerImpl {
...
+        }
+        if ifr.amd_pwrok_fedge || ifr.amd_rstn_fedge {
+            let rstn = self.seq.amd_reset_fedges.counts();
+            let pwrokn = self.seq.amd_pwrok_fedges.counts();
+            ringbuf_entry!(Trace::ResetCounts { rstn, pwrokn });
+            // counters are cleared in the A2 -> A0 transition
+            self.set_state_internal(PowerState::A0Reset);
+            // host_sp_comms will be notified of this change and will
+            // call back into this task to reboot the system (going to
+            // A2 then back into A0)
```
aapoalas, line 698: **question:** Is this likewise a terminal state like above, and if so then the above suggestions would go for here as well.

aapoalas (commented):
Few questions from the peanut gallery.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +597,89 @@ impl ServerImpl {
...
+            seq_state: state.seq,
+        });
+
+        if ifr.thermtrip {
+            self.seq.ifr.modify(|h| h.set_thermtrip(false));
+            ringbuf_entry!(Trace::Thermtrip);
+            self.set_state_internal(PowerState::A0Thermtrip)
+            // this is a terminal state (for now)
+        }
+        if ifr.amd_pwrok_fedge || ifr.amd_rstn_fedge {
```
mkeeter, line 691: Previously, this was an `else if`, to establish priorities between different flags.  Was the change intentional?  I worry that if multiple flags are set, the behavior could be confusing – we could log a `Thermtrip`, but end up in `A0Reset` (and be rebooted by `host_sp_comms`).

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +597,89 @@ impl ServerImpl {
...
+            let rstn = self.seq.amd_reset_fedges.counts();
+            let pwrokn = self.seq.amd_pwrok_fedges.counts();
+            ringbuf_entry!(Trace::ResetCounts { rstn, pwrokn });
+            // counters are cleared in the A2 -> A0 transition
```
mkeeter: Previously, we waited for the A2 -> A0 transition to clear both counters and the interrupt flags themselves.  Now, I think we need to clear the flags here; otherwise the FPGA will keep asserting the interrupt line.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +597,89 @@ impl ServerImpl {
...
+            self.set_state_internal(PowerState::A0Reset);
+            // host_sp_comms will be notified of this change and will
+            // call back into this task to reboot the system (going to
+            // A2 then back into A0)
+        }
+
+        if ifr.smerr_assert {
+            // Give up?
+            self.seq.ifr.modify(|h| h.set_smerr_assert(false));
+            self.emergency_a2(StateChangeReason::SmerrAssert);
```
mkeeter: Both this and the case below will get us permanently stuck in A2, because `host-sp-comms` will not initiate a power cycle (it only checks for `A0Reset`).  I'm not sure what our policy should be here, any thoughts?

drv/cosmo-seq-server/src/main.rs
mkeeter: Yeah, I agree that we'd lose this case if the FPGA doesn't notify us.

app/cosmo/base.toml
```diff
@@ -74,6 +74,12 @@ port = "F"
 pin = 2
 owner = {name = "sprot", notification = "rot_irq"}
 
+[tasks.sys.config.gpio-irqs.seq_irq]
+# FPGA1_TO_SP_IRQA1_L
```
mkeeter:
```suggestion
# FPGA1_TO_SP_IRQ1_L
```

drv/cosmo-seq-server/src/main.rs
labbott: This was not (exactly) intentional. I was perhaps naive in assuming that we'd only ever get one pin at a time and mostly copied over the logic. I think what we want is logging the full interrupt state but only taking a single action based on priority.

drv/cosmo-seq-server/src/main.rs
labbott: I may have worked around this by changing where we enable/disable interrupt when going from A2 -> A0 but this is getting a rework anyway

drv/cosmo-seq-server/src/main.rs
labbott: I think `smerr_assert` and `a0mapo` are similar to `thermtrip` which looks to also be terminal and won't get restarted (assuming I can read this correctly), I expect this is where we'll want our ereporting to come in to deliver useful information elsewhere.

labbott: I thought about this a little more (and corrected my understanding of how `A0::ThermTrip` works) and I think treating these like an unexpected sequencer state and go to A2 makes sense for now. Mapo means power off so having that happen as soon as possible makes sense. `smerr` happens with a 'critical system management error'. There's some argument that a reset might be appropriate but staying in A2 is probably fine.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -571,13 +580,13 @@ impl ServerImpl {
...
-    /// in `A0PlusHP`, we're polling for a thermtrip or for someone disabling
-    /// the NIC.  In other states, there's no need to poll.
+    /// If we are in `A0`, then we are waiting for the NIC to come up;
+    /// Once we are in `A0PlusHP` we rely on interrupts from the sequencer
+    /// to notify us of any changes
     fn poll_interval(&self) -> Option<u32> {
         match self.state {
             PowerState::A0 => Some(10),
-            PowerState::A0PlusHP => Some(100),
+            PowerState::A0PlusHP => None,
```
hawkw: turbo nit, not important: couldn't we just delete this line, since it falls into the `_` case below?
```suggestion
```

drv/cosmo-seq-server/src/main.rs
hawkw: This is definitely not unreachable, I think the comment here is just saying that when THERMTRIP fires, we don't transition into any other state until explicitly powercycled by the control plane. We don't transition back to the A2 state here.

hawkw: Regarding the early return, see https://github.com/oxidecomputer/hubris/pull/2181/files#r2254552184

drv/cosmo-seq-server/src/main.rs
hawkw: Yeah, I agree that it's probably nice to record ringbuf entries for any bits that were set. But, we should definitely change this so we don't clobber the "thermtrip is a terminal state" with an A2 transition from some other bit, on the off chance that one's set.

drv/cosmo-seq-server/src/main.rs
hawkw: The comment I left there applies here too, it's terminal with regards to the sequencer state machine, absent of external inputs through other mechanisms, but it doesn't mean it's terminal for _this task_.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -641,15 +733,21 @@ impl idl::InOrderSequencerImpl for ServerImpl {
...
 impl NotificationHandler for ServerImpl {
     fn current_notification_mask(&self) -> u32 {
-        notifications::TIMER_MASK
+        notifications::TIMER_MASK | notifications::SEQ_IRQ_MASK
     }
 
     fn handle_notification(&mut self, bits: u32) {
+        if (bits & notifications::SEQ_IRQ_MASK) != 0 {
+            self.handle_sequencer_interrupt();
+            return;
```
hawkw: hm, do we not want to fall through and _also_ do everything else if the timer notification _also_ fired?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +597,89 @@ impl ServerImpl {
...
+        );
+    }
+
+    fn handle_sequencer_interrupt(&mut self) {
+        let ifr = self.seq.ifr.view();
+
+        let state = self.log_state_registers();
+        ringbuf_entry!(Trace::SequencerInterrupt {
+            our_state: self.state,
+            seq_state: state.seq,
```
hawkw, line 669: might we also want to include the IFR bits here?

drv/cosmo-seq-server/src/main.rs
labbott: Hmmmm, my thought had been that once we have our interrupts firing we shouldn't need the timer state. I think you're right that there is a case before we've gotten to `A0PlusHP` where we could be getting faults.

drv/cosmo-seq-server/src/main.rs
labbott: haha yeah this was me experimenting with getting rid of the timer interrup

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +593,127 @@ impl ServerImpl {
...
+            // host_sp_comms will be notified of this change and will
+            // call back into this task to reboot the system (going to
+            // A2 then back into A0)
+        }
+
+        if ifr.thermtrip {
+            self.seq.ifr.modify(|h| h.set_thermtrip(false));
+            ringbuf_entry!(Trace::Thermtrip);
+            action = InternalAction::ThermTrip;
+            // Reat place for an ereport?
```
hawkw: typo:
```suggestion
            // Great place for an ereport?
```

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +593,127 @@ impl ServerImpl {
...
+            m.set_smerr_assert(true);
+            m.set_a0mapo(true);
+            m.set_nicmapo(true);
+            m.set_amd_pwrok_fedge(true);
+            m.set_amd_rstn_fedge(true);
+        });
+        let _ = self.sys.gpio_irq_control(
+            notifications::SEQ_IRQ_MASK,
+            sys_api::IrqControl::Enable,
+        );
```
hawkw: Hm, it occurs to me that perhaps we should enable the EXTI IRQ *prior* to unmasking the bits in the FPGA's IER, just to be extra certain that we won't miss an interrupt because we haven't enabled the EXTI IRQ yet?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +593,127 @@ impl ServerImpl {
...
+            // Reat place for an ereport?
+        }
+
+        if ifr.a0mapo {
+            self.log_pg_registers();
+            self.seq.ifr.modify(|h| h.set_a0mapo(false));
+            ringbuf_entry!(Trace::A0MapoInterrupt);
+            action = InternalAction::Mapo;
+            // Great place for an ereport?
+        }
```
hawkw, line 716: turbo nit, sorry: the previous `if` block has a blank line after it, which this one doesn't:
```suggestion
        }

```

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +593,127 @@ impl ServerImpl {
...
+        enum InternalAction {
+            Reset,
+            ThermTrip,
+            Smerr,
+            Mapo,
+        }
+
+        // We check these in lowest to highest priority. The expectation is
+        // we probably(?) won't see multiple of these set at a time but
+        // it's important to account for that case;
```
hawkw: Not a huge deal, but it might be nice to also write down _why_ we prioritize these alerts in that order?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +593,127 @@ impl ServerImpl {
...
+        }
+        // Fan Fault is unconnected
+        // NIC MAPO is unconnected
+
+        match action {
+            InternalAction::Reset => {
+                self.set_state_internal(PowerState::A0Reset);
+            }
+            InternalAction::ThermTrip => {
+                self.set_state_internal(PowerState::A0Thermtrip);
```
mkeeter, line 737:
```suggestion
                self.set_state_internal(PowerState::A0Thermtrip);
                // host_sp_comms will be notified of this change and will
                // call back into this task to reboot the system (going to
                // A2 then back into A0)
```

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +593,127 @@ impl ServerImpl {
...
+            ringbuf_entry!(Trace::SmerrInterrupt);
+            action = InternalAction::Smerr;
+            // Great place for an ereport?
+        }
+        // Fan Fault is unconnected
+        // NIC MAPO is unconnected
+
+        match action {
+            InternalAction::Reset => {
+                self.set_state_internal(PowerState::A0Reset);
```
mkeeter, line 732: Note that this is a terminal state (for now)?

(also relevant for `Mapo` and `Smerr`)

drv/cosmo-seq-server/src/main.rs
labbott: I decided to bring this case back by polling but on a lower interval

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +605,134 @@ impl ServerImpl {
...
+
+        // We check these in lowest to highest priority. We start with
+        // reset since we expect the CPU to handle that nicely.
+        // Thermal trip is a terminal state in that we log it but don't
+        // actually make any changes to the sequencer.
+        // SMERR is treated as a higher priority than MAPO arbitrarily.
+        // we probably(?) won't see multiple of these set at a time but
+        // it's important to account for that case;
+
+        let mut action = InternalAction::Reset;
```
mkeeter: Robustness nit: how about initializing this to `None`, then log a ringbuf message if none of the flags are set?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -588,6 +605,134 @@ impl ServerImpl {
...
+    fn emergency_a2(&mut self, reason: StateChangeReason) {
+        // Power down to A2, updating our internal state.  We can't
+        // handle errors here, so log them and continue.
+        if let Err(e) = self.set_state_impl(PowerState::A2, reason) {
+            ringbuf_entry!(Trace::PowerDownError(e))
+        }
+    }
+
+    fn enable_sequencer_interrupts(&mut self) {
+        let _ = self.sys.gpio_irq_control(
```
mkeeter, line 632: Should we clear`seq.ifr` flags here, so we don't instantly get an interrupt that accumulated while they were disabled?

labbott: My thought was that we shouldn't get any interrupts because we disable the interrupts before we go A0 -> A2 and nothing should be powered in A2. I think you're right that we could still get an interrupt during that transition. If it's an actual error (thermtrip, mapo, smerr) would clearing it be the correct behavior or would we actually want it to fire as soon as possible?

mkeeter: I'm not sure about FPGA behavior here – @nathanaelhuffman does `ier` control flags being set in `ifr`, or does it just control when the FPGA pulls the IRQ line to the SP?

nathanaelhuffman: ifr flags can and do set regardless of the interrupt enable state. This allows one to poll even non-enabled interrupts if one desires, or use them as flags that an event happened.  The actual irq pin to the SP is generated by a bitwise AND of the enables and the flags and then OR-reduced such that any set flag with a corresponding enable will trigger the IRQ pin to the SP.

mkeeter: Thanks! @labbott, I think that means we want to clear the flags before re-enabling interrupts, to avoid spurious flags that accumulated while we had them disabled – does that make sense to you?  Even if we a thermtrip / mapo / smerr fired simultaneously with going to A2, we weren't going to do anything about it...

labbott: yep, makes sense. Thanks for the extra eyes!

mkeeter (approved):

Issue #2185, opened 2025-08-06
SP does not properly translate version in VersionedRotBootInfo message


lzrd:
# SP does not properly translate version in VersionedRotBootInfo message

## Problem

The VersionedRotBootInfo message retrieves unattested boot and image state from the RoT using a versioned protocol for backward/forward compatibility. MGS uses one-based versions while RoT uses zero-based versions, requiring translation by the SP.

**The SP has two related issues:**
1. **No version translation**: Passes MGS version requests directly to RoT without converting one-based to zero-based
2. **No version limiting**: Doesn't clamp requests to versions the SP can handle, allowing RoT to return response variants the SP cannot deserialize

This causes no production issues currently since MGS only requests `HIGHEST_KNOWN_VERSION`. However, adding a new RotBootInfo variant will cause deserialization failures during mixed-version deployments unless both issues are fixed.

## Impact

Mixed-version deployments (master SP + new RoT) will experience failures:
- `HIGHEST_KNOWN_VALUE` requests return deserialization errors  
- Firmware updates and health checks affected

The proposed SP fix (*fix-sp-rbi-translation*) addresses both issues by properly translating and limiting version requests.

## Testing

Note that in these tests, versions above and below the implemented versions are
used to test the corner conditions. Only the responses to requesting the
`RotBootInfo::HIGHEST_KNOWN_VERSION` would impact a production environment.
In the case of the current MGS main branch, this value is '3'. When a new
variant is introduced, it will be incremented to '4'.

The tables below are from running tests between different versions.
  - *main* branch of `management-gateway-service` (current production deployment)
  - *master* branch of Hubris (current production SP/RoT)
  - *fix* is a Hubris branch that implements the proposed fix in the SP image (*fix-sp-rbi-translation*)
  - *bdl* is an MGS or Hubris branch that has both the SP *fix* and a new `RotBootInfo` variant implemented (*boot-decision-log*)

NOTE: In these tables, only the MGS versioning is used. The actual RoT version is one less than the MGS version.

## Current Implementation

| MGS | SP | RoT | MGS Request Version | RoT Response/Error |
| --- | -- | --- | ----------- | ----------------------- |
| main | master | master | HIGHEST_KNOWN_VALUE | V3 |
| main | master | master | 0 | Error response from SP: update: RoT boot info version is not supported |
| main | master | master | 1 | V1 |
| main | master | master | 2 | V2 |
| main | master | master | 3 | V3 |
[...]

PR #2186, opened 2025-08-06
Fix version translation for VersionedRotBootInfo

lzrd:
Translate MGS 1-based versions to RoT 0-based versions for the SP/RoT `versioned_rot_boot_info()` call and clamp the version to be the lesser of the MGS's and SP's `RotBootInfo::HIGHEST_KNOWN_VERSION`.

The version argument of the VersionedRotBootInfo message was being used directly for the sprot `versioned_rot_boot_info` call. These are actually two different name spaces. This change helps when an upcoming PR introduces new MGS and RoT varients to support the `BootDecisionLog`.

Without this change, a new MGS and RoT running against an old SP will have some cases where an MGS request is understood by the RoT but the RoT response cannot be deserialized by the SP and results in sending an error to the MGS.

Issue #2185

### DIFF
--- a/task/control-plane-agent/src/mgs_common.rs
+++ b/task/control-plane-agent/src/mgs_common.rs
@@ -27,12 +27,15 @@ use drv_sprot_api::{
 use drv_stm32h7_update_api::Update;
 use gateway_messages::{
     CfpaPage, DiscoverResponse, DumpSegment, DumpTask, Fwid as GwFwid,
-    ImageError as GwImageError, PowerState, RotBootInfo as GwRotBootInfo,
-    RotError, RotRequest, RotResponse, RotSlotId as GwRotSlotId,
-    RotStateV2 as GwRotStateV2, RotStateV3 as GwRotStateV3, SensorReading,
-    SensorRequest, SensorRequestKind, SensorResponse, SpComponent,
-    SpError as GwSpError, SpPort as GwSpPort, SpStateV2 as GwSpStateV2,
-    UpdateStatus, VpdError as GwVpdError, WatchdogError,
+    ImageError as GwImageError, ImageVersion as GwImageVersion, PowerState,
+    RotBootInfo as GwRotBootInfo, RotBootState as GwRotBootState, RotError,
+    RotImageDetails as GwRotImageDetails, RotRequest, RotResponse,
+    RotSlotId as GwRotSlotId, RotState as GwRotState,
+    RotStateV2 as GwRotStateV2, RotStateV3 as GwRotStateV3,
+    RotUpdateDetails as GwRotUpdateDetails, SensorReading, SensorRequest,
+    SensorRequestKind, SensorResponse, SpComponent, SpError as GwSpError,
+    SpPort as GwSpPort, SpStateV2 as GwSpStateV2, UpdateStatus,
+    VpdError as GwVpdError, WatchdogError,
 };
 use ringbuf::ringbuf_entry_root as ringbuf_entry;
 use static_assertions::const_assert;
@@ -608,6 +611,25 @@ impl MgsCommon {
         Ok(())
     }
 
+    /// Returns the `GwRotBootInfo` version requested by MGS, or falls back to
+    /// the highest supported version that does not exceed the request.
+    ///
+    /// This "best-effort" behavior simplifies the client's logic. Later
+    /// response versions are improvements and the control plane benefits from
+    /// the most complete information for update planning. We provide the best
+    /// available data directly rather than forcing the client to handle errors
+    /// and search for a supported version. Callers that require strict
+    /// versioning can treat version mismatches as an error.
+    ///
+    /// The supported versions are constrained by this SP and the RoT firmware.
+    /// Some version skew between the control plane (MGS), SP, and RoT is
+    /// expected.
+    /// Temporary version skew is **unavoidable** during normal system updates
+    /// but can also result from update failures or they can occur during
+    /// development.
+    /// The fallback logic is designed to handle these mismatches gracefully.
+    /// Introducing a new `GwRotBootInfo` variant requires care, as an
+    /// unhandled mismatch can break the update process.
     pub(crate) fn versioned_rot_boot_info(
         &mut self,
         version: u8,
@@ -616,27 +638,63 @@ impl MgsCommon {
             version
         }));
 
-        // Force update of this code if new variants are introduced
+        // For full context on the version mapping, see the RoT's
+        // `versioned_rot_boot_info` function in `drv/lpc55-update-server`
+        // and the corresponding MGS data structures.
+
+        // Ensure that the version mapping logic below is
+        // intentionally reviewed whenever a new `GwRotBootInfo` version
+        // or SpVersionedRotBootInfo variant is added.
+        // If these assertions fail, both the `match version` mapping and the
+        // `match self.sprot...` response construction must be updated.
+
+        // Force update of this code if new MGS variants are introduced
+        const _HIGHEST_KNOWN_MGS_VERSION: u8 = 3;
         static_assertions::const_assert_eq!(
             GwRotBootInfo::HIGHEST_KNOWN_VERSION,
-            3u8
+            _HIGHEST_KNOWN_MGS_VERSION
         );
+
+        // Force update of this code if new ROT variants are introduced
+        const HIGHEST_KNOWN_ROT_VERSION: u8 = 2;
+        static_assertions::const_assert_eq!(
+            SpVersionedRotBootInfo::HIGHEST_KNOWN_VERSION,
+            HIGHEST_KNOWN_ROT_VERSION
+        );
+
         // Map the MGS RotBootInfo 1-based versions to RoT 0-based versions.
         let rot_version = match version {
+            // There is no version -1 in the RoT version number scheme.
             0 => return Err(GwSpError::RequestUnsupportedForComponent),
-            1 => 0,
+
+            // See Issue #2193: To maintain compatibility with older systems,
+            // we intentionally preserve a legacy bug where MGS V1 was passed
+            // unaltered to the RoT instead of being mapped to RoT V0.
+            // 1 => 0, // This is the correct mapping.
+            1 => 1, // This is the bug-compatible mapping
+
+            // The remaining mappings are consistent with the design.
             2 => 1,
             3 => 2,
             // Clamp the version to the highest known by the SP.
-            _ => 2,
+            _ => HIGHEST_KNOWN_ROT_VERSION,
         };
 
         match self.sprot.versioned_rot_boot_info(rot_version)? {
-            // The RoT V1 response corresponds to the MGS V2
             SpVersionedRotBootInfo::V1(rot_v1) => {
-                Ok(GwRotBootInfo::V2(MgsRotStateV2::from(rot_v1).0))
+                if version == 1 {
+                    // The MGS request is for a deprecated RoT response.
+                    // Maintain bug compatibility until issue #2193 is resolved.
+                    // Construct an MGS V1 from the more recent RoT V1 struct.
+                    Ok(GwRotBootInfo::V1(MgsRotState::from(rot_v1).0))
+                } else {
+                    // The RoT V1 response corresponds to the MGS V2 or is the
+                    // highest that the RoT could provide.
+                    Ok(GwRotBootInfo::V2(MgsRotStateV2::from(rot_v1).0))
+                }
             }
-            // The RoT V2 response corresponds to the MGS V3
+            // The RoT V2 response corresponds to the MGS V3 or is the highest
+            // version the RoT could provide.
             SpVersionedRotBootInfo::V2(rot_v2) => {
                 Ok(GwRotBootInfo::V3(MgsRotStateV3::from(rot_v2).0))
             }
@@ -698,6 +756,39 @@ impl From<SpSlotId> for MgsRotSlotId {
         MgsRotSlotId(match id {
             SpSlotId::A => GwRotSlotId::A,
             SpSlotId::B => GwRotSlotId::B,
+        })
+    }
+}
+
+// Note: this struct and impl go away when issue #2193 is resolved.
+struct MgsRotState(GwRotState);
+
+impl From<SpRotBootInfo> for MgsRotState {
+    fn from(v1: SpRotBootInfo) -> MgsRotState {
+        MgsRotState(GwRotState {
+            rot_updates: GwRotUpdateDetails {
+                boot_state: GwRotBootState {
+                    active: MgsRotSlotId::from(v1.active).0,
+                    slot_a: v1.slot_a_sha3_256_digest.map(|digest| {
+                        GwRotImageDetails {
+                            version: GwImageVersion {
+                                version: 0,
+                                epoch: 0,
+                            },
+                            digest,
+                        }
+                    }),
+                    slot_b: v1.slot_b_sha3_256_digest.map(|digest| {
+                        GwRotImageDetails {
+                            version: GwImageVersion {
+                                version: 0,
+                                epoch: 0,
+                            },
+                            digest,
+                        }
+                    }),
+                },
+            },
         })
     }
 }
--- a/drv/lpc55-update-api/src/lib.rs
+++ b/drv/lpc55-update-api/src/lib.rs
@@ -161,6 +161,9 @@ pub enum VersionedRotBootInfo {
     V1(RotBootInfo),
     V2(RotBootInfoV2),
 }
+impl VersionedRotBootInfo {
+    pub const HIGHEST_KNOWN_VERSION: u8 = 2;
+}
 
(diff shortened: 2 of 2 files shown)


task/control-plane-agent/src/mgs_common.rs
```diff
@@ -620,17 +616,28 @@ impl MgsCommon {
...
+        // variants.
+        // Clamp the version to one known by the SP.
+        // Note that the MGS RotBootInfo versions are 1-based while
+        // RoT versions are zero-based. So, also adjust the index.
+        let rot_version = if version > GwRotBootInfo::HIGHEST_KNOWN_VERSION {
+            GwRotBootInfo::HIGHEST_KNOWN_VERSION
+        } else {
+            version
+        }
+        .saturating_sub(1u8);
```
labbott: This feels off. There was an RoT v0 but we still have version = 1 => RotBootInfoV1 so saying RoT versions are zero based doesn't seem correct (https://github.com/oxidecomputer/hubris/blob/master/drv/lpc55-update-server/src/main.rs#L337-L365). This also seems like it would break accessing versions 1 and 2 because 1 - 1 = 0 which returns an error instead of the V1 message and 2 = 1 which would now return V1 support instead of V2. We probably don't actually care about those cases but it seems strange to do so in a change to provide backwards compatibility.

This also seems to defeat the point of `HIGHEST_KNOWN_VERSION`. https://github.com/oxidecomputer/management-gateway-service/blob/main/gateway-messages/src/sp_to_mgs.rs#L645-L647 The request for the version comes from the control plane. This would imply that we've managed to update the RoT and the control plane to support the higher version but not the SP which feels like it breaks a lot of our implicit assumptions about how update is supposed to work.

The comment related to `HIGHEST_KNOWN_VERSION` seems to imply that MGS should handle the differences in versions. We already did the work in MGS to handle `RotStateV3` vs `RotStateV2` so I think it makes more sense to try and handle it up there than here in the SP.

mkeeter: I agree that this is a change in behavior.

In the new code, `version = 1` is unsupported, and `version = 2` constructs an `MgsRotStateV2` from a `RotBootInfo` (rather than from a `RotBootInfoV2`); I'm not confident this conversion propagates the same information.

If we want to avoid returning deserialization errors, then the SP could clamp the version it requests to `HIGHEST_KNOWN_VERSION`, but the `saturating_sub` seems incorrect.  I'll put some more notes in #2185

(I also agree that handling this at the MGS layer seems reasonable, to push complexity to the Big Computer)

lzrd: The SP is translating SP/RoT messages to MGS messages. If we look at the source code of the structs we can see that in the [versioned RotBootInfo](https://github.com/oxidecomputer/management-gateway-service/blob/91f7b0901988f9ad4c5aa10b44cc9108fa9cf9a7/gateway-messages/src/sp_to_mgs.rs#L632-L655) enum: [MGS V2](https://github.com/oxidecomputer/management-gateway-service/blob/91f7b0901988f9ad4c5aa10b44cc9108fa9cf9a7/gateway-messages/src/sp_to_mgs.rs#L440-L462) corresponds to [RoT V1](https://github.com/oxidecomputer/hubris/blob/437c053649220611cdffc8a8d8d4033c0ef4b89c/drv/lpc55-update-api/src/lib.rs#L98-L122) and that [MGS V3](https://github.com/oxidecomputer/management-gateway-service/blob/91f7b0901988f9ad4c5aa10b44cc9108fa9cf9a7/gateway-messages/src/sp_to_mgs.rs#L528-L563) corresponds to [RoT V2](https://github.com/oxidecomputer/hubris/blob/437c053649220611cdffc8a8d8d4033c0ef4b89c/drv/lpc55-update-api/src/lib.rs#L124-L157).
The MGS analog to the MGS [RotBotInfo::V1(RotState)](https://github.com/oxidecomputer/management-gateway-service/blob/91f7b0901988f9ad4c5aa10b44cc9108fa9cf9a7/gateway-messages/src/sp_to_mgs.rs#L410-L415) is the unsupported RoT "V0" case.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -620,17 +616,29 @@ impl MgsCommon {
...
+        } - 1;
+
+        match self.sprot.versioned_rot_boot_info(rot_version)? {
+            // The RoT V1 response corresponds to the MGS V2
+            SpVersionedRotBootInfo::V1(rot_v1) => {
+                Ok(GwRotBootInfo::V2(MgsRotStateV2::from(rot_v1).0))
+            }
+            // The RoT V2 response corresponds to the MGS V3
+            SpVersionedRotBootInfo::V2(rot_v2) => {
+                Ok(GwRotBootInfo::V3(MgsRotStateV3::from(rot_v2).0))
```
labbott: I find this significantly harder to reason about

labbott (commented):
I really think this is fundamental issue with `HIGHEST_KNOWN_VERSION` across the control plane + hubris. We already had to have some backwards compatibility when we added the V3 state https://github.com/oxidecomputer/omicron/blob/main/gateway/src/http_entrypoints.rs#L1179-L1204 for the bootloader. I would much rather see the control plane do something like e.g. https://github.com/oxidecomputer/omicron/blob/main/wicketd/src/update_tracker.rs#L1873-L1900 and then try one version lower vs handling this in the SP.

task/control-plane-agent/src/mgs_common.rs
lzrd: I'm only considering the simple delivery of the requested information from the RoT to the SP. The SP is only repackaging the equivalent data structures from the RoT to SP. It seems to me that this is independent from anything the control plane needs to do with the data. 

The SP isn't implementing any policy here. It just doesn't have the code to deserialize the RoT response for versions it hasn't seen.

With current code, MGS is requesting V3, and the RoT doesn't have a V3, it has a RoT-V2, the RoT's own highest version, which is what is returned but not what was request. The SP transcribes that to an MGS-V3 which happens to match what it wanted, but the process by which it got it was not correct.

In  .../update_trackers.rs, only the V3 version is useful, the others are error cases.

lzrd: I could do this instead. This addresses mapping, but not clamping vs returning an error.
If an error is returned for 3+ in this case (a new behavior that requires MGS changes), then the MGS needs to start guessing what version to request. Granted, that is likely only to be one more message exchange, but a new Error type that had the SP's highest known version could be helpful. The SP cannot speak for the RoT, so there could conceivably be another round trip if the RoT cannot satisfy the request as well. 
```
        // Force update of this code if new variants are introduced
        const _: () = assert!(GwRotBootInfo::HIGHEST_KNOWN_VERSION == 3);
        // Map MGS versions to RoT versions
        let rot_version = match version {
            0 => return Err(GwSpError::RequestUnsupportedForComponent),
            1 => 0,
            2 => 1,
            3 => 2,
            _ => 2,
        };
```

mkeeter: I like the explicit `match`; you should use `static_assertions::const_assert_eq!` here to make the assertion checked at compile time.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -620,17 +616,30 @@ impl MgsCommon {
...
+        );
+        // Map the MGS RotBootInfo 1-based versions to RoT 0-based versions.
+        let rot_version = match version {
+            0 => return Err(GwSpError::RequestUnsupportedForComponent),
+            1 => 0,
+            2 => 1,
+            3 => 2,
+            // Clamp the version to the highest known by the SP.
+            _ => 2,
+        };
```
labbott: This is much easier to review and reason about, thanks for doing this

mkeeter:
The new code is better, but I have two remaining issues:

- The comments do not explain what's going on to a casual observer, so it's hard to tell if the logic is correct
- If MGS requests `version = 1`, it will now fail.  This is a behavior change from before; I think the goal of this PR is to fix bugs, so we shouldn't change behavior.

Here is my suggested implementation:

```rust
    pub(crate) fn versioned_rot_boot_info(
        &mut self,
        version: u8,
    ) -> Result<GwRotBootInfo, GwSpError> {
        ringbuf_entry!(Log::MgsMessage(MgsMessage::VersionedRotBootInfo {
            version
        }));

        // Force update of this code if new variants are introduced
        static_assertions::const_assert_eq!(
            GwRotBootInfo::HIGHEST_KNOWN_VERSION,
            3u8
        );

        // The goal of this code is to return the `GwRotBootInfo` version
        // requested by MGS.  If this is not possible, then we return the
        // highest version that we can construct.  We are limited by both the SP
        // firmware (which you're reading right now) and the RoT (the function
        // `versioned_rot_boot_info` in `lpc55-update-server`).  The SP and RoT
        // may not be in sync with each other!

        // MGS and SP-RoT versions are independent of each other.  Based on the
        // MGS version requested, we ask the RoT for the oldest boot info
        // variant that has all of the information needed to build that MGS
        // variant.  If the RoT does not know about this version, then -- like
        // the SP -- it will return the highest version that it knows about.
        // We must not assume that it returns the requested version!
        let rot_version = match version {
            0 => return Err(GwSpError::RequestUnsupportedForComponent),
            1 => 1,
            2 => 1,
            3 => 2,
            // Clamp the version to the highest known by the SP.
            _ => 2,
        };

        match self.sprot.versioned_rot_boot_info(rot_version)? {
            SpVersionedRotBootInfo::V1(rot_v1) => {
                if version == 1 {
                    // MGS requested V1, so that's what we return
                    Ok(GwRotBootInfo::V1(MgsRotState::from(rot_v1).0))
                } else {
                    // MGS either requested V2, or requested something newer
                    // that the RoT couldn't provide; V2 is what we can
                    // construct from the RoT info, so it's what we return.
                    Ok(GwRotBootInfo::V2(MgsRotStateV2::from(rot_v1).0))
                }
            }
            SpVersionedRotBootInfo::V2(rot_v2) => {
                // MGS either requested V3 or something newer; V3 is what we can
                // construct.
                Ok(GwRotBootInfo::V3(MgsRotStateV3::from(rot_v2).0))
            }
        }
    }
```

Note that this requires restoring code which was deleted in the PR, to preserve `version = 1` functionality.  Full patch below the fold...

<details>

```diff
Commit ID: e7e8880c9dc84c2f6a57c108cf79b948aca2635f
Change ID: mprskounnytotqmzzlzxypyvoxssprmw
Author   : Matt Keeter <matt@oxide.computer> (2025-08-12 09:24:02)
Committer: Matt Keeter <matt@oxide.computer> (2025-08-12 09:25:47)

    (no description set)

diff --git a/task/control-plane-agent/src/mgs_common.rs b/task/control-plane-agent/src/mgs_common.rs
index c4dbc9e624..b7e470f5b6 100644
--- a/task/control-plane-agent/src/mgs_common.rs
+++ b/task/control-plane-agent/src/mgs_common.rs
@@ -620,17 +620,51 @@
             version
         }));
 
-        match self.sprot.versioned_rot_boot_info(version)? {
-            SpVersionedRotBootInfo::V1(v1) => {
-                Ok(GwRotBootInfo::V1(MgsRotState::from(v1).0))
-            }
-            SpVersionedRotBootInfo::V2(v2) => match version {
-                2 => Ok(GwRotBootInfo::V2(MgsRotStateV2::from(v2).0)),
-                // RoT's V2 is MGS V3 and the highest version that we can offer today.
-                _ => Ok(GwRotBootInfo::V3(MgsRotStateV3::from(v2).0)),
-            },
-            // New variants that this code doesn't know about yet will
-            // result in a deserialization error.
+        // Force update of this code if new variants are introduced
+        static_assertions::const_assert_eq!(
+            GwRotBootInfo::HIGHEST_KNOWN_VERSION,
+            3u8
+        );
+
+        // The goal of this code is to return the `GwRotBootInfo` version
+        // requested by MGS.  If this is not possible, then we return the
+        // highest version that we can construct.  We are limited by both the SP
+        // firmware (which you're reading right now) and the RoT (the function
+        // `versioned_rot_boot_info` in `lpc55-update-server`).  The SP and RoT
+        // may not be in sync with each other!
+
+        // MGS and SP-RoT versions are independent of each other.  Based on the
+        // MGS version requested, we ask the RoT for the oldest boot info
+        // variant that has all of the information needed to build that MGS
+        // variant.  If the RoT does not know about this version, then -- like
+        // the SP -- it will return the highest version that it knows about.
+        // We must not assume that it returns the requested version!
+        let rot_version = match version {
+            0 => return Err(GwSpError::RequestUnsupportedForComponent),
+            1 => 1,
+            2 => 1,
+            3 => 2,
+            // Clamp the version to the highest known by the SP.
+            _ => 2,
+        };
+
+        match self.sprot.versioned_rot_boot_info(rot_version)? {
+            SpVersionedRotBootInfo::V1(rot_v1) => {
+                if version == 1 {
+                    // MGS requested V1, so that's what we return
+                    Ok(GwRotBootInfo::V1(MgsRotState::from(rot_v1).0))
+                } else {
+                    // MGS either requested V2, or requested something newer
+                    // that the RoT couldn't provide; V2 is what we can
+                    // construct from the RoT info, so it's what we return.
+                    Ok(GwRotBootInfo::V2(MgsRotStateV2::from(rot_v1).0))
+                }
+            }
+            SpVersionedRotBootInfo::V2(rot_v2) => {
+                // MGS either requested V3 or something newer; V3 is what we can
+                // construct.
+                Ok(GwRotBootInfo::V3(MgsRotStateV3::from(rot_v2).0))
+            }
         }
     }
 ```
</details>

File: task/control-plane-agent/src/mgs_common.rs [hunk 4]
```diff
@@ -620,17 +616,30 @@ impl MgsCommon {
             version
         }));
 
-        match self.sprot.versioned_rot_boot_info(version)? {
-            SpVersionedRotBootInfo::V1(v1) => {
-                Ok(GwRotBootInfo::V1(MgsRotState::from(v1).0))
+        // Force update of this code if new variants are introduced
+        static_assertions::const_assert_eq!(
+            GwRotBootInfo::HIGHEST_KNOWN_VERSION,
+            3u8
```
labbott: Can you put the constant `3` somewhere to give a better idea what it's tracking (`sprot-api` perhaps?)

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -620,17 +616,30 @@ impl MgsCommon {
             version
         }));
 
-        match self.sprot.versioned_rot_boot_info(version)? {
-            SpVersionedRotBootInfo::V1(v1) => {
-                Ok(GwRotBootInfo::V1(MgsRotState::from(v1).0))
+        // Force update of this code if new variants are introduced
```
labbott: Without any other knowledge, it's not immediately clear what this assertion is telling us about what needs to be updated since we simply clamp the version anyway below. Can you be more specific in this comment about what you expect to be updated?

labbott: Matt's example comment would be a great starting point about what to mention should be updated

task/control-plane-agent/src/mgs_common.rs
mkeeter: I don't think it should go in `sprot-api`, since it's related to the SP-MGS communication; maybe just declare it as
```rust
const EXPECTED_HIGHEST_KNOWN_MGS_VERSION: u8 = 3u8;
```
?

labbott: That's a better approach.

Having this assertion still makes me think something is missing somewhere and I wish we could force the SP to handle this via changes to https://github.com/oxidecomputer/management-gateway-service/blob/main/gateway-messages/src/sp_impl.rs#L112

task/control-plane-agent/src/mgs_common.rs
lzrd: I have updated comments and created issue 2193 to document the bugs intentionally not fixed.
Rebased on latest updates to the master branch and run the full set of tests.
The code is now backward-bug compatible except that it fixes the case where the RoT introduces a new RotBootInfo variant.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -620,17 +638,66 @@ impl MgsCommon {
...
+        // or SpVersionedRotBootInfo variant is added.
+        // If these assertions fail, both the `match version` mapping and the
+        // `match self.sprot...` response construction must be updated.
+
+        // Force update of this code if new MGS variants are introduced
+        const _HIGHEST_KNOWN_MGS_VERSION: u8 = 3;
+        static_assertions::const_assert_eq!(
+            GwRotBootInfo::HIGHEST_KNOWN_VERSION,
+            _HIGHEST_KNOWN_MGS_VERSION
+        );
```
mkeeter, line 656: Nit:
```suggestion
        const HIGHEST_KNOWN_MGS_VERSION: u8 = 3;
        static_assertions::const_assert_eq!(
            GwRotBootInfo::HIGHEST_KNOWN_VERSION,
            HIGHEST_KNOWN_MGS_VERSION
        );
```

PR #2169, opened 2025-07-22
Update to nightly-2025-07-20

hawkw:
Currently, Hubris' Rust toolchain is pinned to `nightly-2024-09-17`,
which is from 10 months ago. This is kind of unfortunate, especially
because that nightly is too old to support the Rust 2024 edition, which
means we cannot update our dependencies on any crates where the latest
version is built with Rust 2024.

Beyond just updating the toolchain, there were some noteworthy changes:

- Naked functions are now stable (yay!), but the attribute changed to
  `#[unsafe(naked)]`. Inline assembly in naked functions must now use
  `core::arch::naked_asm!` rather than normal `asm!`. As far as I can
  tell, the only difference between this and regular `asm!` is that it
  does not support `options(noreturn)`, as I believe the `naked_asm!`
  block kind of implies at least some of the codegen differences for
  `noreturn`.
- New warnings on creation of temporary shared references to mutable
  statics showed up in `stm32h7-update-server`, where we were using
  `[T]::as_ptr()` on zero-sized slices in mutable statics that were used
  to get linker-generated addresses. `[T]::as_ptr()` takes an `&self`
  receiver, so this was generating a temporary shared reference to the
  mutable static. I changed this to use `&raw const`, which takes the
  address of the static without creating a shared reference.
- There was a substantial regression in flash and RAM usage on the new
  toolchain due to [a change in the behavior of the `#[used]`
  attribute][1] which revealed [an underlying issue where ringbufs were
  not zero initialized][2]. These issues were resolved separately in
  #2167, #2168, #2170, and oxidecomputer/idolatry#65.

In addition, there were a variety of unremarkable linting changes,
including slightly better dead code detection (which detected some new
dead code), and some annoying clippy nonsense.

Note that this branch requires oxidecomputer/idolatry#65, which updates
`idol` to generate code that doesn't emit warnings with the new
toolchain, and fixes some of the flash/RAM size regression in code
generated by `idol`.

Fixes #2165

[1]: https://github.com/oxidecomputer/hubris/issues/2165#issuecomment-3098980540
[2]: https://github.com/oxidecomputer/hubris/issues/2165#issuecomment-3104520732

PR #2194, opened 2025-08-13
Update to enable stable toolchain

### LINKED ISSUE #1927: We're still on the nightly toolchain and that's annoying
Hubris still requires the nightly toolchain to build, and that's annoying. It makes toolchain updates hard for several reasons:

- We're on nightly because we use unstable features, which (as the term suggests) are unstable. These can be broken at any time. This is not a hypothetical risk, either; it happens on most updates.
- Dependencies detect that we're on nightly and _they, too,_ have an unstable feature free-for-all. So the set of features that can break us is larger than the set we actually use. There's no commonly supported way to stop dependencies from doing this crap.
- If a new Rust fix or feature drops, we can't just bump to the released toolchain, we need to figure out which nightly version _roughly corresponds_ to the version we want. Because features get backported, it's entirely possible that _no_ nightly toolchain directly corresponds to the released version.

As of this writing, the things keeping us on nightly are:

- `feature(naked_functions)`: used in the kernel (though I'm attempting to fix this, see #1926), in `userlib`, and in a couple lpc55-specific low level libraries. I suspect all uses of `naked_functions` can be replaced by `global_asm!`, which is good, because the way `naked_functions` works has changed in fundamental ways after about 1.83.
- `feature(used_with_arg)`: I...have no idea why we're using this. But we are. It only appears in the test suite?
- `-Z emit-stack-sizes`: yes, our recent stack size monitoring requires nightly. This is annoying, but, we could make it optional.

...and as far as I can tell, _that's it._

thenewwazoo:
# PR content

This commit makes a small number of updates necessary to adopt the stable toolchain:
* Sets the stable toolchain as the default by config. Invoking nightly will require using `+nightly-2025-07-20` per #2169.
* Removes references to nightly requirements in the FAQ, and adds a note about enabling stack size checking being the only thing that requires nightly.
* Removes code that passes the `-Z` flag to cargo to enable nightly features.
* Updates `#[naked]` and `asm!` usage to stable convention, which is `#[unsafe(naked)]` and `naked_asm!`. Any affected assembly sections are expected to diverge (which was previously handled by `options(noreturn)`) or end with a trap (in the case of `_start`).
* Replaces `used(linker)` with `used`, which is [equivalent].
* Adds a code path in the `dist` build task to build endoscope if the task being built has a name `drv_lpc55_swd`, upon which it depends. This is because [artifact dependencies] are not stable, but `lcp55-swd` depends on having the endoscope binary available to its build script.
* Gates stack size emission and check behind a `rustc_version` channel check. Passing `+nightly-2025-07-20` to `cargo` (when invoking xtask) will enable the check.

[equivalent]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/attrs/data_structures/enum.UsedBy.html
[artifact dependencies]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#artifact-dependencies

Fixes #1927 

# Open questions

* [ ] This PR implicitly declares a MSRV of 1.89.0, but no MSRV policy, which requires the input of the maintainers.
* [ ] Excepting the FAQ, the preferred nightly version is not actually encoded anywhere. You just have to kind of already know that `2025-07-20` is the blessed nightly version.
* [ ] I have not run this on production hardware.
* [ ] Checking task stack sizes is something that should be done in CI, but that requires tooling changes.

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -14,10 +14,12 @@ use std::process::{Command, Stdio};
 
 use anyhow::{anyhow, bail, Context, Result};
 use atty::Stream;
+use cargo_metadata::Message;
 use indexmap::IndexMap;
 use lpc55_rom_data::FLASH_PAGE_SIZE as LPC55_FLASH_PAGE_SIZE;
 use multimap::MultiMap;
 use path_slash::{PathBufExt, PathExt};
+use rustc_version::{version_meta, Channel};
 use sha3::{Digest, Sha3_256};
 use zerocopy::IntoBytes;
 
@@ -373,6 +375,13 @@ pub fn package(
     // return value, because we're going to link them regardless of whether the
     // build changed.
     for name in cfg.toml.tasks.keys() {
+        // Magic for lpc55-swd to be able to include endoscope, until artifact
+        // dependencies are available on stable. See build_endoscope's commend for
+        // more.
+        if cfg.toml.tasks[name.as_str()].name == "drv-lpc55-swd" {
+            build_endoscope()?;
+        }
+
         if tasks_to_build.contains(name.as_str()) {
             build_task(&cfg, name)?;
         }
@@ -493,22 +502,24 @@ pub fn package(
             })
             .collect::<Result<_, _>>()?;
 
-        // Check stack sizes and resolve task slots in our linked files
-        let mut possible_stack_overflow = vec![];
-        for task_name in cfg.toml.tasks.keys() {
-            if tasks_to_build.contains(task_name.as_str()) {
-                if task_can_overflow(&cfg.toml, task_name, verbose)? {
-                    possible_stack_overflow.push(task_name);
-                }
+        if version_meta().unwrap().channel == Channel::Nightly {
+            // Check stack sizes and resolve task slots in our linked files
+            let mut possible_stack_overflow = vec![];
+            for task_name in cfg.toml.tasks.keys() {
+                if tasks_to_build.contains(task_name.as_str()) {
+                    if task_can_overflow(&cfg.toml, task_name, verbose)? {
+                        possible_stack_overflow.push(task_name);
+                    }
 
-                resolve_task_slots(&cfg, task_name, image_name)?;
+                    resolve_task_slots(&cfg, task_name, image_name)?;
+                }
             }
-        }
-        if !possible_stack_overflow.is_empty() {
-            bail!(
-                "tasks may overflow: {possible_stack_overflow:?}; \
-                 see logs above"
-            );
+            if !possible_stack_overflow.is_empty() {
+                bail!(
+                    "tasks may overflow: {possible_stack_overflow:?}; \
+        see logs above"
+                );
+            }
         }
 
         // Add an empty output section for the caboose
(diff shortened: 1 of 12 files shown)


build/xtask/src/dist.rs
```diff
@@ -493,22 +502,24 @@ pub fn package(
...
 
-                resolve_task_slots(&cfg, task_name, image_name)?;
+                    resolve_task_slots(&cfg, task_name, image_name)?;
+                }
+            }
+            if !possible_stack_overflow.is_empty() {
+                bail!(
+                    "tasks may overflow: {possible_stack_overflow:?}; \
+        see logs above"
+                );
```
labbott, line 521: Ooooh this is really a deal breaker right now I think :/

We really need the stack size checking feature as stackoverflows are one of the biggest hubris pain points

thenewwazoo: That's okay - the work is done, and going from "here is a _list_ of nightly features we use" to "here is the _one thing_ we use nightly for" is an improvement. and my goal was to knock out an easy task to familiarise myself with hubris, so: mission accomplished.

lmk if I should carve this change out slash refactor this PR as getting 90% there, or if there's more internal discussion to be had.

dancrossnyc: I confess I haven't looked too closely at the mechanism or how it is implemented in practice, but I wonder if we need it in production? Or could we make it part of a CI pipeline with `RUSTC_BOOTSTRAP=1`?

thenewwazoo: There are two aspects to the stack size check vis-a-vis compilation:
* The [build method in dist.rs](https://github.com/thenewwazoo/hubris/blob/276a9e7cccbec5744cea55704ea174fc00b81025/build/xtask/src/dist.rs#L2075) uses [RUSTFLAGS](https://github.com/thenewwazoo/hubris/blob/276a9e7cccbec5744cea55704ea174fc00b81025/build/xtask/src/dist.rs#L2109-L2111) to cause [the `rustc` it invokes](https://github.com/thenewwazoo/hubris/blob/276a9e7cccbec5744cea55704ea174fc00b81025/build/xtask/src/dist.rs#L2083) to emit stack size info. This requires that `rustc` binary to be a nightly version.
* The [package method in dist.rs](https://github.com/thenewwazoo/hubris/blob/276a9e7cccbec5744cea55704ea174fc00b81025/build/xtask/src/dist.rs#L507-L516) loops over every task and checks it for overflow. The [`get_max_stack` function](https://github.com/thenewwazoo/hubris/blob/276a9e7cccbec5744cea55704ea174fc00b81025/build/xtask/src/dist.rs#L1261) does not actually require nightly, but will explode the packaging process if the stack info is not found.

The first aspect is the tricker one. Right now, my current design expects `cargo` to be invoked with `+nightly-...` at the topmost level (i.e. `cargo +nightly-... xtask ...`) in order to implicitly select the nightly compiler. The `rustc` façade _will_ take a toolchain spec, so the toolchain selection could potentially be pushed into the dist task by conditionally adding a `+nightly-...` argument to `rustc`. The `BuildConfig` used to locate `rustc` appears to support [using a façade](https://github.com/thenewwazoo/hubris/blob/b35abbaa8097c2c1ee6dd3db5b3eafc3a90fd9d7/build/xtask/src/config.rs#L776-L780), though its default construction [specifies the sysroot](https://github.com/thenewwazoo/hubris/blob/b35abbaa8097c2c1ee6dd3db5b3eafc3a90fd9d7/build/xtask/src/dist.rs#L1110), which I think implies deeper plumbing would be necessary. The net result of this would be a stable compiler building a dist build task which invokes a nightly compiler to compile firmware tasks.

The second aspect is easier. If we shouldn't expect to find stack size info, just eat the error and move on.

thenewwazoo: Thinking about this a bit more, the stack checking is not something that ends up _in_ the production image, if that makes sense. You could use nightly to build a binary, emitting stack size info as a side-effect, and then throw that binary away in favor of a stable-toolchain-built binary, _assuming_ all else is equal (... a _big_ assumption, maybe).

dancrossnyc: Yeah, that was kind of what I was going for with my earlier comment about only doing the stack checks with in CI, and not production. I apologize for being terribly vague there, however.

But a two-step process, where one builds a binary with the stack size data via setting `RUSTC_BOOTSTRAP=1` and setting the magic `-Z` flag in `RUSTFLAGS`, and then building another binary without setting `RUSTC_BOOTSTRAP`, was what I meant.

I presume we can use the same compiler binary for both invocations, and that we don't need an actual nightly compiler binary for either, as long as we set `RUSTC_BOOTSTRAP=1` for the case where we need the stack data.

mkeeter:
I had forgotten about this PR, and ended up implementing the switch to stable-ish* Rust across https://github.com/oxidecomputer/hubris/pull/2537 + https://github.com/oxidecomputer/hubris/pull/2544.

Regardless, thank you for looking into this!  If you would like some Oxide swag, send me an email (matt at oxide dot computer) and I can see about shipping you a t-shirt 😄 

*using `RUSTC_BOOTSTRAP` to keep the stack size checks, because those catch our most common failure mode

PR #2195, opened 2025-08-14
Rename task's name member to bin-crate

### LINKED ISSUE #196: Change name of task's "name" attribute in app.toml to something less confusing
The task's name is set by the key in the tasks map, not the name attribute of the task entry.  The name attribute is the name of the bin crate to be built.

thenewwazoo:
This change renames the `name` member of the `Task` struct to `bin-crate` in order to make it more clear to new users that the value refers to an actual crate, and is not the name of the task (which is the key in the map).

(This is admittedly a very silly PR fixing an issue dating back to _2021_ but it did actually trip me up.)

Fixes #196

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -19,7 +19,7 @@ region = "flash"
 size = 256
 
 [tasks.jefe]
-name = "task-jefe"
+bin-crate = "task-jefe"
 priority = 0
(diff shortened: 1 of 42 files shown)


thenewwazoo:
Corresponding Humility PR [here](https://github.com/oxidecomputer/humility/pull/562)

PR #2199, opened 2025-08-18
Use refdes-based component IDs in control-plane-agent

### LINKED ISSUE #2197: need a way to get an I2C device's control-plane-agent component ID in ereports
When an ereport is reported about a particular device, the control plane will wish to poll the SP for additional data about that device. Since ereports are intended as edge-triggered notifications about an event that may not include all relevant data, they must include an identifier that can be used to index any components involved with the device when requesting additional data about the component from the SP.  In the `control-plane-agent` API, devices are indexed by _component IDs_. These component IDs are [also included in the Oximeter metrics recorded by MGS' SP sensor polling](https://github.com/oxidecomputer/omicron/blob/61ad056cee3d8cfa509799ce2d4f898bcafb54e4/oximeter/oximeter/schema/hardware-component.toml#L64-L68), so they are also necessary if a control plane diagnosis engine wishes to pull past sensor data recorded for a device when an ereport is received. Therefore, we would like to include component IDs in ereports.

Presently, these component IDs are generated by control-plane-agent based on the device's index in the `VALIDATE_DEVICES` array in `task_validate_api` or by an index in an `OUR_DEVICES` array known to `control-plane-agent`, in the [`Inventory::device_description`](https://github.com/oxidecomputer/hubris/blob/d601eaf303c5bff67d574e85a6a2af634444321a/task/control-plane-agent/src/inventory.rs#L84-L137) method. Unfortunately, this means there isn't a way to go from an `I2cDevice` in a  task to a component ID string for that device, which could be included when a task generates an ereport from that task. 

Note that #2189 adds code for including a device's reference designator in the schematic (its "refdes") in `I2cDevice` codegen, so that the refdes may be included in the ereport. However, this is not presently useful for the control plane when requesting additional details on the device, as the `control-plane-agent` API doesn't know about refdes. Devices cannot be looked up by refdes there, and the control plane cannot turn a refdes into a component ID using its own understanding of the SP's inventory, since refdes is not currently included in inventory collection.

I can think of several very different solutions to this problem, but would welcome any suggestions of things I've not thought of:

1. **Rework component IDs into something produced by codegen**: Rather than just having the `control-plane-agent` synthesize a device's component ID from its array index, could have the `build-i2c` codegen also generate the component ID. When generating the array literal, the build script could also track the index in the array and generate a string constant for the component ID, which can optionally be included in the device handle. This might be a bit annoying to plumb through everything, but seems doable.
2. **Have tasks ask the `control-plane-agent`/`validate` tasks for devices' component IDs**: We could treat `control-plane-agent` as the authoritative source for component IDs, and have tasks that want to generate an ereport for a device do an IPC to `control-plane-agent` to ask "hey, what is the component ID of this thing?". I don't like this idea, but I'm including it for completeness: it means we have lots of new dependencies on `control-plane-agent`, which may not always be possible due to priorities. Also, it _should_ be possible to know the component ID statically, so resolving it at runtime seems sad to me. 
3. **Include refdes in inventory**: We could turn the whole thing on its head and *not* use component IDs in ereports at all. Instead, we would add the device's refdes to the `control-plane-agent` inventory device description messages. This way, the control plane can include the refdes in the database records for devices along with the component ID. When the control plane receives an ereport for a particular device, it can look up the component ID for the corresponding refdes in its own inventory table if it needs to request additional information about the device by ID.
4. **Index devices by refdes**: We could also extend the previous idea by allowing devices to be looked up by refdes in `control-plane-agent`. This is potentially a bit tricky as the current approach lets us just parse an integer and turn it into an array index, rather than having to maintain a mapping of refdes strings to array indices, which is painful to do in a no-allocation, constrained-memory embedded system...

hawkw:
In the `gateway-sp-comms` protocol, hardware devices known to the SP are uniquely identified by _component IDs_. These IDs are generated by the SP and are given to the control plane when reporting inventory. The control plane may then request details about a particular hardware component using the ID received from the inventory.

Presently, these IDs use what I've been referring to as the "dev-NN scheme".  In this scheme, each component is identified by a string consisting of `"dev-` followed by a two digit number, which is generated by the `control-plane-agent` task from that device's position in an array of devices generated at build time. While @jgallagher  will claim that this is just a bad design driven entirely by expedience, it does have some redeeming qualities: the SP can look up devices very efficiently by parsing the `NN` part as an integer and using it to index the array, and all the identifiers are a fixed, short number of bytes. However, it also has some disadvantages:

- The identifiers are not semantically meaningful
- A given device's identifier may change across firmware versions if new devices are added to the `app.toml` or the ordering of devices changes
- The identifiers are known only to the `control-plane-agent` task, and cannot easily be accessed by other tasks.

That last point is particularly problematic for the fault management subsystem: we would like a variety of tasks to be able to generate ereports relating to particular hardware components. When producing such an ereport, we would like to be able to include an identifier for that device that the control plane may then use to request additional information about that device from the SP. These identifiers should also be able to be used by the control plane to look up that device in its own cached understanding of the SP's inventory, and for querying metrics collected from the SP about that component. Since the `dev-nn` component IDs are known only to the `control-plane-agent` task, other Hubris tasks cannot easily include them in ereports. This is discussed at length in #2197.

In that issue, we [decided][1] to solve this problem by changing the `control-plane-agent` task from identifying components using the `dev-NN` scheme to a new scheme based on reference designators (refdes). Since a device's refdes is set in the `app.toml` for the board, it can be accessed by any number of tasks if we include it in generated code for the I<sup>2</sup>C (and, eventually, SPI) APIs. We started on this in #2189. Using refdes as the identifiers for hardware components has a number of other advantages:

- In `host-sp-comms`, we report inventory to the host OS using refdes as the unique ID for devices, so this makes the MGS API consistent with the IPCC API.
- Refdes is a fixed property of the _board_, rather than a property of the particular layout of items in a TOML file. This means that it should not change across Hubris versions (barring a bug or incorrect refdes in an app config).
- The refdes indicates semantically meaningful information; it can be used to find a given component in the board's schematic.
- Refdes are also pretty short strings

This PR changes `control-plane-agent` to use the device's refdes as the component ID, rather than the `dev-nn` identifier. When looking up a device, we use an array mapping refdes strings to indices in the devices array, which is generated at build time. We perform a binary search over this array to find the index and then access the device by index.

Certain devices (typically, those with updateable firmware) use hard-coded component ID strings. On Gimlets, for example, these are `sp`, `sp3-host-cpu`, `host-boot-flash`, and `system-led`.  I left these alone and didn't mess with them.

In a handful of cases, multiple devices have the same refdes in `app.toml`. Typically, this is when a removable subassembly is on the "other side" of a connector --- for example, a sharkfin has a VPD EEPROM, a MAX5970 hotswap controller,  and the U.2 device's NVMe Basic Management Command, and these all have refdes like `J206`, for the CEM slot that the sharkfin occupies. In this case, the `host-sp-comms` task uses a scheme like `{connector_refdes}/{refdes_on_subassembly}`, so (for example) sharkfin A's VPD EEPROM is `J206/U7` and its HSC is `J206/U8`. Currently, this is all just hardcoded in `host-sp-comms`. In order to use this scheme in `control-plane-agent` as well, I've changed the toml format to accept refdes either as a string _or_ an array of strings, which are joined together at codegen-time with `/` when forming the device IDs, or `_`s when generating Rust identifiers. There were some preexisting cases of refdes strings like `Jxxx_Uxx` in some app.tomls; I've changed these to use the new format.

Finally, there were a handful of boards which lacked refdes for some or all of their devices. After this change, these boards no longer build, so I've gone and updated them.

Fixes #2197

[1]: https://github.com/oxidecomputer/hubris/issues/2197#issuecomment-3192600805

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -519,7 +519,7 @@ description = "U.2 Sharkfin A hot swap controller"
 power = { rails = [ "V12_U2A_A0", "V3P3_U2A_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_a_hsc"
-refdes = ["J200", "U1A"]
+refdes = ["J200", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -556,7 +556,7 @@ description = "U.2 Sharkfin B hot swap controller"
 power = { rails = [ "V12_U2B_A0", "V3P3_U2B_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_b_hsc"
-refdes = ["J201", "U1A"]
+refdes = ["J201", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -593,7 +593,7 @@ description = "U.2 Sharkfin C hot swap controller"
 power = { rails = [ "V12_U2C_A0", "V3P3_U2C_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_c_hsc"
-refdes = ["J202", "U1A"]
+refdes = ["J202", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -630,7 +630,7 @@ description = "U.2 Sharkfin D hot swap controller"
 power = { rails = [ "V12_U2D_A0", "V3P3_U2D_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_d_hsc"
-refdes = ["J203", "U1A"]
+refdes = ["J203", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -667,7 +667,7 @@ description = "U.2 Sharkfin E hot swap controller"
 power = { rails = [ "V12_U2E_A0", "V3P3_U2E_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_e_hsc"
-refdes = ["J204", "U1A"]
+refdes = ["J204", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -704,7 +704,7 @@ description = "U.2 Sharkfin F hot swap controller"
 power = { rails = [ "V12_U2F_A0", "V3P3_U2F_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_f_hsc"
-refdes = ["J205", "U1A"]
+refdes = ["J205", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -744,7 +744,7 @@ description = "U.2 Sharkfin G hot swap controller"
 power = { rails = [ "V12_U2G_A0", "V3P3_U2G_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_g_hsc"
-refdes = ["J206", "U1A"]
+refdes = ["J206", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -783,7 +783,7 @@ description = "U.2 Sharkfin H hot swap controller"
 power = { rails = [ "V12_U2H_A0", "V3P3_U2H_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_h_hsc"
-refdes = ["J207", "U1A"]
+refdes = ["J207", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -820,7 +820,7 @@ description = "U.2 Sharkfin I hot swap controller"
 power = { rails = [ "V12_U2I_A0", "V3P3_U2I_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_i_hsc"
-refdes = ["J208", "U1A"]
+refdes = ["J208", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -857,7 +857,7 @@ description = "U.2 Sharkfin J hot swap controller"
 power = { rails = [ "V12_U2J_A0", "V3P3_U2J_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_j_hsc"
-refdes = ["J209", "U1A"]
+refdes = ["J209", "U1"]
 removable = true
 
 [[config.i2c.devices]]
@@ -957,7 +957,7 @@ address = 0x3c
 device = "sbrmi"
 name = "RMI"
 description = "CPU via SB-RMI"
-refdes = ["U1U", "SBRMI"]
+refdes = ["U1", "SBRMI"]
 
 [[config.i2c.devices]]
 bus = "main"
@@ -968,7 +968,7 @@ device = "sbtsi"
 name = "CPU"
 description = "CPU temperature sensor"
 sensors = { temperature = 1 }
-refdes = ["U1U", "SBTSI"]
+refdes = ["U1", "SBTSI"]
 
 # Segment 5 is SEC_SP5_TO_FPGA1
 # Segment 6 is unused
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -500,8 +500,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin A VPD"
 name = "sharkfin_a_vpd"
-refdes = "J200"
-refdes-suffix = "U2"
+refdes = ["J200", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -514,8 +513,7 @@ description = "U.2 Sharkfin A hot swap controller"
 power = { rails = [ "V12_U2A_A0", "V3P3_U2A_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_a_hsc"
-refdes = "J200"
-refdes-suffix = "U1A"
+refdes = ["J200", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -527,8 +525,7 @@ device = "nvme_bmc"
 description = "U.2 A NVMe Basic Management Command"
 sensors = { temperature = 1 }
 name = "U2_N0"
-refdes = "J200"
-refdes-suffix = "J2"
+refdes = ["J200", "J2"]
 removable = true
 
 # SMBUS_FPGA2_TO_CEMB
@@ -540,8 +537,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin B VPD"
 name = "sharkfin_b_vpd"
-refdes = "J201"
-refdes-suffix = "U2"
+refdes = ["J201", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -554,8 +550,7 @@ description = "U.2 Sharkfin B hot swap controller"
 power = { rails = [ "V12_U2B_A0", "V3P3_U2B_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_b_hsc"
-refdes = "J201"
-refdes-suffix = "U1A"
+refdes = ["J201", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -567,8 +562,7 @@ device = "nvme_bmc"
 description = "U.2 B NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N1"
-refdes = "J201"
-refdes-suffix = "J2"
+refdes = ["J201", "J2"]
 removable = true
 
 # SMBUS_FPGA2_TO_CEMC
@@ -580,8 +574,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin C VPD"
 name = "sharkfin_c_vpd"
-refdes = "J202"
-refdes-suffix = "U2"
+refdes = ["J202", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -594,8 +587,7 @@ description = "U.2 Sharkfin C hot swap controller"
 power = { rails = [ "V12_U2C_A0", "V3P3_U2C_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_c_hsc"
-refdes = "J202"
-refdes-suffix = "U1A"
+refdes = ["J202", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -607,8 +599,7 @@ device = "nvme_bmc"
 description = "U.2 C NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N2"
-refdes = "J202"
-refdes-suffix = "J2"
+refdes = ["J202", "J2"]
 removable = true
 
 # SMBUS_FPGA2_TO_CEMD
@@ -620,8 +611,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin D VPD"
 name = "sharkfin_d_vpd"
-refdes = "J203"
-refdes-suffix = "U2"
+refdes = ["J203", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -634,8 +624,7 @@ description = "U.2 Sharkfin D hot swap controller"
 power = { rails = [ "V12_U2D_A0", "V3P3_U2D_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_d_hsc"
-refdes = "J203"
-refdes-suffix = "U1A"
+refdes = ["J203", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -647,8 +636,7 @@ device = "nvme_bmc"
 description = "U.2 D NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N3"
-refdes = "J203"
-refdes-suffix = "J2"
+refdes = ["J203", "J2"]
 removable = true
 
 # SMBUS_FPGA2_TO_CEME
@@ -660,8 +648,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin E VPD"
 name = "sharkfin_e_vpd"
-refdes = "J204"
-refdes-suffix = "U2"
+refdes = ["J204", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -674,8 +661,7 @@ description = "U.2 Sharkfin E hot swap controller"
 power = { rails = [ "V12_U2E_A0", "V3P3_U2E_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_e_hsc"
-refdes = "J204"
-refdes-suffix = "U1A"
+refdes = ["J204", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -687,8 +673,7 @@ device = "nvme_bmc"
 description = "U.2 E NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N4"
-refdes = "J204"
-refdes-suffix = "J2"
+refdes = ["J204", "J2"]
 removable = true
 
 # SMBUS_FPGA2_TO_CEMF
@@ -700,8 +685,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin F VPD"
 name = "sharkfin_f_vpd"
-refdes = "J205"
-refdes-suffix = "U2"
+refdes = ["J205", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -714,8 +698,7 @@ description = "U.2 Sharkfin F hot swap controller"
 power = { rails = [ "V12_U2F_A0", "V3P3_U2F_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_f_hsc"
-refdes = "J205"
-refdes-suffix = "U1A"
+refdes = ["J205", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -727,8 +710,7 @@ device = "nvme_bmc"
 description = "U.2 F NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N5"
-refdes = "J205"
-refdes-suffix = "J2"
+refdes = ["J205", "J2"]
 removable = true
 
 # Note that we skip segments 7 and 8, which are
@@ -743,8 +725,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin G VPD"
 name = "sharkfin_g_vpd"
-refdes = "J206"
-refdes-suffix = "U2"
+refdes = ["J206", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -757,8 +738,7 @@ description = "U.2 Sharkfin G hot swap controller"
 power = { rails = [ "V12_U2G_A0", "V3P3_U2G_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_g_hsc"
-refdes = "J206"
-refdes-suffix = "U1A"
+refdes = ["J206", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -770,8 +750,7 @@ device = "nvme_bmc"
 description = "U.2 G NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N6"
-refdes = "J206"
-refdes-suffix = "J2"
+refdes = ["J206", "J2"]
 removable = true
 
 
@@ -785,8 +764,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin H VPD"
 name = "sharkfin_h_vpd"
-refdes = "J207"
-refdes-suffix = "U2"
+refdes = ["J207", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -799,8 +777,7 @@ description = "U.2 Sharkfin H hot swap controller"
 power = { rails = [ "V12_U2H_A0", "V3P3_U2H_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_h_hsc"
-refdes = "J207"
-refdes-suffix = "U1A"
+refdes = ["J207", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -812,8 +789,7 @@ device = "nvme_bmc"
 description = "U.2 H NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N7"
-refdes = "J207"
-refdes-suffix = "J2"
+refdes = ["J207", "J2"]
 removable = true
 
 # SMBUS_FPGA2_TO_CEMI
@@ -825,8 +801,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin I VPD"
 name = "sharkfin_i_vpd"
-refdes = "J208"
-refdes-suffix = "U2"
+refdes = ["J208", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -839,8 +814,7 @@ description = "U.2 Sharkfin I hot swap controller"
 power = { rails = [ "V12_U2I_A0", "V3P3_U2I_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_i_hsc"
-refdes = "J208"
-refdes-suffix = "U1A"
+refdes = ["J208", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -852,8 +826,7 @@ device = "nvme_bmc"
 description = "U.2 I NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N8"
-refdes = "J208"
-refdes-suffix = "J2"
+refdes = ["J208", "J2"]
 removable = true
 
 # SMBUS_FPGA2_TO_CEMJ
@@ -865,8 +838,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "U.2 Sharkfin J VPD"
 name = "sharkfin_j_vpd"
-refdes = "J209"
-refdes-suffix = "U2"
+refdes = ["J209", "U2"]
 removable = true
 
 [[config.i2c.devices]]
@@ -879,8 +851,7 @@ description = "U.2 Sharkfin J hot swap controller"
 power = { rails = [ "V12_U2J_A0", "V3P3_U2J_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_j_hsc"
-refdes = "J209"
-refdes-suffix = "U1A"
+refdes = ["J209", "U1A"]
 removable = true
 
 [[config.i2c.devices]]
@@ -892,8 +863,7 @@ device = "nvme_bmc"
 description = "U.2 J NVMe Basic Management Control"
 sensors = { temperature = 1 }
 name = "U2_N9"
-refdes = "J209"
-refdes-suffix = "J2"
+refdes = ["J209", "J2"]
 removable = true
 # you are now leaving the sharkfin zone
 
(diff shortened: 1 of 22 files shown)


task/host-sp-comms/src/bsp/gimlet_bcde.rs
```diff
@@ -405,23 +404,21 @@ impl ServerImpl {
...
                 let dev = f(I2C.get_task_id());
 
-                // Convert the name from Jxxx (in the TOML file) -> Jxxx/U1
-                let mut name = *b"Jxxx/U1";
-                // All connector names should have length 4; that's checked by
-                // the type in the tuple assignment above.
-                name[..4].copy_from_slice(&connector_name);
+                // Convert the name from Jxxx_U1 (in the TOML file) -> Jxxx/U1
+                // TODO(eliza): let's just get this from the include_refdes i2c
+                // codegen...
```
hawkw: It would be nice to change most of this code in `host-sp-comms` to just use the component ID strings we generate (added in #2189). However, I felt like this was better done in a follow-up PR as I think it requires some additional codegen changes (I'd like to make the component ID not be optional, which breaks a lot of places where we construct I2C devices "manually").

task/host-sp-comms/src/bsp/gimlet_bcde.rs
```diff
@@ -515,20 +512,25 @@ impl ServerImpl {
...
+                    3 => by_refdes!(J209_U8, max5970),
+                    4 => by_refdes!(J210_U8, max5970),
+                    5 => by_refdes!(J211_U8, max5970),
+                    6 => by_refdes!(J212_U8, max5970),
+                    7 => by_refdes!(J213_U8, max5970),
+                    8 => by_refdes!(J214_U8, max5970),
+                    9 => by_refdes!(J215_U8, max5970),
+                    10 => by_refdes!(U275, max5970, 7),
                     _ => panic!(),
                 };
```
hawkw: Incidentally, this also happens to fix a preexisting bug where the SP would not report that the Sharkfin MAX5970s have voltage and current sensors:

```console
eliza@castle ~ $ rg MAX5970 fmtopo-brm42220082-before.txt
3967:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87
3969:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87
3981:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2A_A0HP:vout
3983:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2A_A0HP:vout
3997:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2B_A0HP:vout
3999:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2B_A0HP:vout
4013:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2A_A0HP:iout
4015:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2A_A0HP:iout
4029:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2B_A0HP:iout
4031:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2B_A0HP:iout
4107:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1
4109:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1
4169:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1
4171:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1
4231:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=2/board=0/ic=1
4233:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=2/board=0/ic=1
4293:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=3/board=0/ic=1
4295:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=3/board=0/ic=1
4355:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=4/board=0/ic=1
4357:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=4/board=0/ic=1
4417:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=5/board=0/ic=1
4419:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=5/board=0/ic=1
4479:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=6/board=0/ic=1
4481:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=6/board=0/ic=1
4541:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=7/board=0/ic=1
4543:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=7/board=0/ic=1
4603:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=8/board=0/ic=1
4605:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=8/board=0/ic=1
4665:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=9/board=0/ic=1
4667:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=9/board=0/ic=1

eliza@castle ~ $ rg MAX5970 fmtopo-brm42220082-after5.txt
3967:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87
3969:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87
3981:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2A_A0HP:vout
3983:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2A_A0HP:vout
3997:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2B_A0HP:vout
3999:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2B_A0HP:vout
4013:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2A_A0HP:iout
4015:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2A_A0HP:iout
4029:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2B_A0HP:iout
4031:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/ic=87?sensor=V3P3_M2B_A0HP:iout
4107:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1
4109:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1
4121:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1?sensor=V12:vout
4123:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1?sensor=V12:vout
4137:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1?sensor=V3P3:vout
4139:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1?sensor=V3P3:vout
4153:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1?sensor=V12:iout
4155:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1?sensor=V12:iout
4169:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1?sensor=V3P3:iout
4171:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=0/board=0/ic=1?sensor=V3P3:iout
4233:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1
4235:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1
4247:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1?sensor=V12:vout
4249:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1?sensor=V12:vout
4263:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1?sensor=V3P3:vout
4265:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1?sensor=V3P3:vout
4279:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1?sensor=V12:iout
4281:    resource          fmri      hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1?sensor=V12:iout
4295:hc://:product-id=913-0000019:product-sn=BRM42220082:server-id=BRM42220082:chassis-id=BRM42220082:part=MAX5970ETX/chassis=0/systemboard=0/slot=1/board=0/ic=1?sensor=V3P3:iout

# ... snipped out the rest as it's quite long now

```

app/cosmo/base.toml
```diff
@@ -519,7 +519,7 @@ description = "U.2 Sharkfin A hot swap controller"
 power = { rails = [ "V12_U2A_A0", "V3P3_U2A_A0" ], pmbus = false }
 sensors = { voltage = 2, current = 2 }
 name = "sharkfin_a_hsc"
-refdes = "J200"
+refdes = ["J200", "U1A"]
```
mkeeter: Nit: this should just be `U1` (also true for other HSCs).  On the schematic, the physical U1 chip is split into U1A, U1B, U1C symbols, but that's just for organization.

app/cosmo/base.toml
```diff
@@ -965,6 +968,7 @@ device = "sbtsi"
 name = "CPU"
 description = "CPU temperature sensor"
 sensors = { temperature = 1 }
+refdes = ["U1U", "SBTSI"]
```
mkeeter:
```suggestion
refdes = ["U1", "SBTSI"]
```
(same nit as HSCs; U1 is the designator)

app/cosmo/base.toml
```diff
@@ -1126,7 +1130,7 @@ name = "Northwest"
 description = "Northwest temperature sensor"
 sensors = { temperature = 1 }
 removable = true
-refdes = "J47"
+refdes = ["J47", "U1"]
```
mkeeter: Weak vibes: the temperature sensor board is basically just a temperature sensor on a stick, so I don't think it's necessary to have the chip listed separately.

app/donglet/app-g031.toml
```diff
@@ -96,11 +96,12 @@ address = 0x73
 [[config.i2c.devices]]
 controller = 1
 mux = 1
-segment = 1 
+segment = 1
 address = 0b1010_000
 device = "at24csw080"
 description = "Sharkfin VPD"
 removable = true
+refdes = ["J3", "J3"]
```
mkeeter: Where does the second `J3` come from?  I see the EEPROM as `U7` in the Sharkfin schematic

app/donglet/app-g031.toml
```diff
@@ -110,6 +111,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "Gimlet Fan VPD"
 removable = true
+refdes = ["J3", "J5"]
```
mkeeter: I think this should be U1, based on https://github.com/oxidecomputer/hardware-gimlet-fan-vpd/releases/tag/gimlet-fan-vpd_C_913-0000022_rev2_tapeout

app/donglet/app-g031.toml
```diff
@@ -119,6 +121,7 @@ address = 0b1010_000
 device = "at24csw080"
 description = "Sidecar Fan VPD"
 removable = true
+refdes = ["J3", "J4"]
```
mkeeter: Ah, I see the problem here – this should _also_ be U1, but they're different devices.  I still don't understand where `J5/J4/J3` came from, maybe a comment or dummy values (`SC_U1`)?

app/gimlet/base.toml
```diff
@@ -434,7 +434,7 @@ name = "Southwest"
 description = "Southwest temperature sensor"
 sensors = { temperature = 1 }
 removable = true
-refdes = "J194"
+refdes = ["J194", "U1"]
```
mkeeter: same vibes here as cosmo; I vaguely feel like the U1 isn't necessary

app/gimletlet/app.toml
```diff
@@ -284,6 +284,7 @@ address = 0x16
 description = "LM5066 evaluation board"
 power = { rails = [ "LM5066_EVL_VOUT" ] }
 sensors = { voltage = 1, current = 1, temperature = 1 }
+refdes = "LOL"
```
mkeeter, line 287: 🚀

app/medusa/base.toml
```diff
@@ -414,7 +434,8 @@ description = "Front IO V3P3_QSFP1_A0 rail"
 removable = true
 power = {rails = ["V3P3_QSFP1_A0"] }
 sensors = { temperature = 1, voltage = 1, current = 1 }
-refdes = "J61_U18" # on front IO board
+refdes = [
+        "J2", "U18"]
```
mkeeter: weird indentation?

build/i2c/src/lib.rs
```diff
@@ -1796,3 +1821,46 @@ where
...
+    ) -> String {
+        match self {
+            Self::Component(s) => {
+                let mut s = s.clone();
+                change_case(&mut s);
+                s
+            }
+            Self::Path(parts) => {
+                let len = parts.iter().map(String::len).sum::<usize>()
+                    + (parts.len() - 1);
```
mkeeter:
```suggestion
                    + (parts.len() - 1) * sep.len();
```
(heaven forbid we have to reallocate if someone specifies a multi-character separator)

task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -432,23 +431,22 @@ impl ServerImpl {
...
+                };
                 let dev = f(I2C.get_task_id());
 
                 // Convert the name from Jxxx (in the TOML file) -> Jxxx/U1
-                let mut name = *b"Jxx/U1";
                 // All connector names should have length 3; that's checked by
                 // the type in the tuple assignment above.
-                name[..3].copy_from_slice(&connector_name);
+                // TODO(eliza): get this from codegen...
+                name[4] = b'/';
```
mkeeter: Isn't the slash at index 3?
```suggestion
                name[3] = b'/';
```

task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -585,18 +588,18 @@ impl ServerImpl {
...
+            0 => by_refdes!(J200_U2, at24csw080),
+            1 => by_refdes!(J201_U2, at24csw080),
+            2 => by_refdes!(J202_U2, at24csw080),
+            3 => by_refdes!(J203_U2, at24csw080),
+            4 => by_refdes!(J204_U2, at24csw080),
+            5 => by_refdes!(J205_U2, at24csw080),
+            6 => by_refdes!(J206_U2, at24csw080),
+            7 => by_refdes!(J207_U2, at24csw080),
+            8 => by_refdes!(J208_U2, at24csw080),
+            9 => by_refdes!(J209_U2, at24csw080),
```
mkeeter, line 590: Two questions:

- Do we need to go from `_` to `/` here?
- Is the host going to be mad about the change here?

task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -102,17 +102,16 @@ impl ServerImpl {
             //
             // Sharkfin connectors start at J200 and are numbered sequentially
             4..=13 => {
-                let (designator, f): ([u8; 4], _) =
+                let (designator, f): ([u8; 7], _) =
                     Self::get_sharkfin_vpd(index as usize - 4);
                 let mut name = *b"____/U2/ID";
-                name[0..4].copy_from_slice(&designator);
+                name[0..4].copy_from_slice(&designator[0..4]);
```
mkeeter: Nit: this would be more clear if we overwrote the lower 7 bytes with the whole designator, e.g.

```rust
let mut name = *b"_______/ID";
name[0..7].copy_from_slice(&designator);
```

task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -102,17 +102,16 @@ impl ServerImpl {
...
                 self.read_eeprom_barcode(sequence, &name, f)
             }
             14..=23 => {
-                let (designator, f): ([u8; 4], _) =
+                let (mut name, f): ([u8; 7], _) =
                     Self::get_sharkfin_vpd(index as usize - 14);
-                let mut name = *b"____/U2";
-                name[0..4].copy_from_slice(&designator);
+                name[4] = b'/';
```
mkeeter: If you change `const fn byteify` to translate from `_` to `/`, you could make this a lot more foolproof!

mkeeter:
```rust
/// `const` function to convert a `&'static str` to a fixed-size byte array
///
/// This must be called a `const` parameter of `s.len()`
#[allow(dead_code)]
pub(crate) const fn byteify<const N: usize>(s: &'static [u8]) -> [u8; N] {
    let mut out = [0u8; N];
    let mut i = 0;
    while i < s.len() {
        if s[i] == b'_' {
            out[i] = b'/';
        } else {
            out[i] = s[i];
        }
        i += 1;
    }
    out
}
```

task/host-sp-comms/src/bsp/gimlet_bcde.rs
mkeeter: Out of curiosity, why did this change the host inventory behavior?

app/cosmo/base.toml
hawkw: yeah, i could go either way on these, i think there was one board (possibly one of the devboards?) where it was `Jx_U1` and i wanted to be consistent with that, but I can change them all to just the jack's refdes.

app/donglet/app-g031.toml
hawkw: I think on Donglet there's a separate PMOD adapter board that the vpd eeproms are plugged into, based on https://github.com/oxidecomputer/hardware-pmod_vpd_prog/blob/main/hardware-pmod_vpd_prog.pdf

Now that we can accept arbitrarily long refdes paths here, I think it could be `J3/J5/U1` and similar for the other eeproms on donglet.

app/gimletlet/app.toml
hawkw: i really didnt want to figure this out

task/host-sp-comms/src/bsp/cosmo_a.rs
hawkw: good call! i didn't mess with this code too much since i was hoping to change it to just use the IDs from codegen in a subsequent PR, but I think this is a good middle ground.

task/validate-api/build.rs
```diff
@@ -24,16 +24,59 @@ fn write_pub_device_descriptions(
...
+    // we'll just log them and keep going, so that we can tell the user about
+    // *all* the bad devices in the config file, rather than bailing out at the
+    // first one. At the end, we return an error if there were any bad devices.
+    // This way, you don't have to fix one issue and recompile in order to
+    // discover the next error.
+    //
+    let mut missing_ids = 0;
+    let mut duplicate_ids = 0;
+    let mut ids_too_long = 0;
+    let mut id2idx = std::collections::BTreeMap::new();
```
mkeeter, line 59: Add a comment here that we're using a `BTreeMap` to ensure that stuff is sorted?

task/validate-api/build.rs
```diff
@@ -53,7 +96,35 @@ fn write_pub_device_descriptions(
...
+        id2idx.len()
+    )?;
+    for (id, idx) in id2idx {
+        writeln!(file, "    ({id:?}, {idx}),")?;
+    }
+    writeln!(file, "];")?;
+    writeln!(
+        file,
+        "pub static DEVICE_INDICES_BY_ID: [([u8; MAX_ID_LENGTH], usize); \
+         DEVICE_INDICES_BY_ID_CONST.len()] = DEVICE_INDICES_BY_ID_CONST;"
```
mkeeter: Do we need both `const` and `static` versions here?

Also, naming them something like `DEVICE_INDICES_BY_SORTED_ID` would make the `binary_search` call more obvious.

task/host-sp-comms/src/bsp/cosmo_a.rs
hawkw: We do not need to go from `_` to `/` here because it was being done above when we called `get_sharkfin_vpd`. But also, your suggestion from https://github.com/oxidecomputer/hubris/pull/2199#discussion_r2291466702 completely obviates that.

hawkw: (And, no, the host is not mad about this, I tested using `fmtopo`)

task/host-sp-comms/src/bsp/cosmo_a.rs
hawkw: Also obviated by https://github.com/oxidecomputer/hubris/pull/2199#discussion_r2291466702

app/donglet/app-g031.toml
hawkw: same answer as https://github.com/oxidecomputer/hubris/pull/2199#discussion_r2291599648

app/donglet/app-g031.toml
hawkw: also https://github.com/oxidecomputer/hubris/pull/2199#discussion_r2291599648

task/validate-api/build.rs
hawkw: ah, i thought something was using the const but on further inspection you're right, i removed it. good call about the naming!

task/validate-api/build.rs
```diff
@@ -24,16 +24,60 @@ fn write_pub_device_descriptions(
...
+    // we'll just log them and keep going, so that we can tell the user about
+    // *all* the bad devices in the config file, rather than bailing out at the
+    // first one. At the end, we return an error if there were any bad devices.
+    // This way, you don't have to fix one issue and recompile in order to
+    // discover the next error.
+    //
+    let mut missing_ids = 0;
+    let mut duplicate_ids = 0;
+    let mut ids_too_long = 0;
+    // We use a BTreeMap here so that the list is ordered by device ID. When generating the array, we want to ensure it's
```
mkeeter: ...it's what?

hawkw: haha, i went to go change the name of the array before finishing the comment and then i forgot to finish it X__X

PR #2184, opened 2025-08-05
cpu-seq: Vcore PMBus alert ereports

hawkw:
This branch implements ereports for PMBus input fault and warning alerts in the Cosmo and Gimlet sequencers. There's kind of a lot going on here. Essentially, I've hacked up @bcantrill's `drv_gimlet_seq_server::vcore` module to produce an ereport when we detect a PMBus input alert. These ereports contain the value of PMBus status registers read from the VRM, as well as whether or not the VRM claims `POWER_GOOD` has been deasserted, and identity information of the device ID and power rail. On Cosmo, I've added similarish code on top of the sequencer interrupt handling code @labbott added in #2181. It's a bit more complex on Cosmo as there are two RAA22960A VRMs driving the `VDDCR_CPU0` and `VDDCR_CPU1` rails, rather than a single regulator as in SP3's single `VDD_VCORE` rail. 

This branch closes #2188, which has been merged into this PR.
Closes #2141

PR #2203, opened 2025-08-22
Eliza/device id codegen 2

hawkw:


hawkw:
[CI failure](https://github.com/oxidecomputer/hubris/actions/runs/17214455575/job/48834464469?pr=2203#step:3:28) for the Windows build looks like a (hopefully transient!) network issue installing the toolchain, I'm gonna restart it.

build/i2c/src/lib.rs
```diff
@@ -967,8 +971,8 @@ impl ConfigGenerator {
 {indent}    Controller::I2C{controller},
 {indent}    PortIndex({port}),
 {indent}    {segment},
-{indent}    {address:#x}
-{indent}){refdes_part}"##,
+{indent}    {address:#x},{component_id}
```
mkeeter, line 974:
```suggestion
{indent}    {address:#x},
{indent}    {component_id},
```
?

build/i2c/src/lib.rs
```diff
@@ -503,15 +503,19 @@ struct ConfigGenerator {
     /// hash of controllers to single port indices
     singletons: HashMap<u8, usize>,
 
-    /// if `true`, include refdes string in output
-    include_refdes: bool,
+    /// if `true`, include component ID string in output.
+    ///
+    /// this requires that the `"drv-i2c-api/component-id"` feature flag is
+    /// enabled. if that feature flag is enabled, then this MUST also be
```
mkeeter, line 509: Are we checking this anywhere (e.g. how [`net` checks for consistency in the build script](https://github.com/oxidecomputer/hubris/blob/0a78e9412efb512da1f9c9d882ccf4dcf51e2299/build/net/src/lib.rs#L88-L96))?

task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -74,23 +71,22 @@ impl ServerImpl {
...
+                let (dev, _sensors) = by_refdes!(U32, at24csw080);
+                self.read_at24csw080_id(sequence, dev)
             }
             2 => {
                 // J34/ID: Fan VPD barcode (not available in packrat)
                 self.read_fan_barcodes(
                     sequence,
                     b"J34/ID",
-                    i2c_config::devices::at24csw080_fan_vpd,
+                    i2c_config::devices::at24csw080_fan_vpd(I2C.get_task_id()),
```
mkeeter: Vibes: this is hard to scan, maybe switch to
```rust
let dev = i2c_config::devices::at24csw080_fan_vpd(I2C.get_task_id());
self.read_fan_barcodes(sequence, b"J34/ID", dev)
```
(also applies to a bunch of other places)

task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -203,46 +206,52 @@ impl ServerImpl {
...
-                        CommandCode::IC_DEVICE_ID as u8,
-                        ic_device_id,
-                    )?;
-                    dev.read_block(
-                        CommandCode::IC_DEVICE_REV as u8,
-                        ic_device_rev,
-                    )?;
-                    Ok(self.scratch)
-                })
+                self.tx_buf.try_encode_inventory(
```
mkeeter: Nit: there's a bunch of places where `dev.component_id().as_bytes()` adds another indentation level to `try_encode_inventory`.  I think getting the name first would prevent this, making the diff smaller and less indented:
```rust
let name = dev.component_id.as_bytes();
self.tx_buf.try_encode_inventory(sequence, name, || {
    // ... etc
```

task/host-sp-comms/src/bsp/cosmo_a.rs
```diff
@@ -102,16 +98,15 @@ impl ServerImpl {
...
             // Sharkfin connectors start at J200 and are numbered sequentially
             4..=13 => {
-                let (designator, f): ([u8; 7], _) =
-                    Self::get_sharkfin_vpd(index as usize - 4);
+                let dev = Self::get_sharkfin_vpd(index as usize - 4);
+                let dev_id = dev.component_id().as_bytes();
                 let mut name = *b"_______/ID";
-                name[0..7].copy_from_slice(&designator);
-                self.read_eeprom_barcode(sequence, &name, f)
+                name[0..7].copy_from_slice(&dev_id[0..7]);
```
mkeeter: This is less obviously correct (because `designator` is no longer a `[u8; 7]`), but not sure what we can do about it...

task/host-sp-comms/src/inventory/compute_sled_common.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// - A nested value `SASY`, which contains four more `BARC` values for each
+    ///   individual fan
+    ///
+    /// On success, packs the barcode into `self.tx_buf`; on failure, return an
+    /// error (`DeviceAbsent` if we saw `NoDevice`, or `DeviceFailed` on all
+    /// other errors).
+    pub(crate) fn read_fan_barcodes(
+        &mut self,
+        sequence: u64,
+        name: &[u8],
```
mkeeter: Should we remove `name` here and get it from `dev.component_id()` instead?

task/host-sp-comms/src/inventory/compute_sled_common.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    /// Reads the "BARC" value from a TLV-C blob in an AT24CSW080 EEPROM
+    ///
+    /// On success, packs the barcode into `self.tx_buf`; on failure, return an
+    /// error (`DeviceAbsent` if we saw `NoDevice`, or `DeviceFailed` on all
+    /// other errors).
+    pub(crate) fn read_eeprom_barcode(
+        &mut self,
+        sequence: u64,
+        name: &[u8],
```
mkeeter: Should we remove `name` here and get it from `dev.component_id()` instead?  We'd have to splice the trailing `/ID`, which is a little tricky because we don't know the length of `component_id`, but could allocate a generous buffer (16 bytes?)

task/host-sp-comms/src/inventory.rs
```diff
@@ -10,33 +10,12 @@
 /// Inventory API version (always 0 for now)
 pub(crate) const INVENTORY_API_VERSION: u32 = 0;
 
-/// `const` function to convert a `&'static str` to a fixed-size byte array
-///
-/// This must be called a `const` parameter of `s.len()`
-#[allow(dead_code)]
-pub(crate) const fn byteify<const N: usize>(s: &'static [u8]) -> [u8; N] {
```
mkeeter, line 17: RIP `byteify` 🫡

build/i2c/src/lib.rs
hawkw: If the component ID isn't present, the `component_id` string will be empty, so this would generate two `,`s in a row, resulting in a syntax error. I included the comma in `component_id` instead, so that we only emit it if the string is non-empty.

Similarly, including the indent and newline will generate that unconditionally, even when `component_id` is the empty string. That's not a syntax error but it looks kinda weird. Therefore, I also formatted the indent and newline into the string:

https://github.com/oxidecomputer/hubris/blob/ca76539b16570d4d3bfe91c5e76f136324cd4244/build/i2c/src/lib.rs#L955-L965

That results in _this_ format! macro looking kinda bad, but it's the way to get the best output IMO.

mkeeter: 👍🏻

task/host-sp-comms/src/inventory/compute_sled_common.rs
hawkw: Hm, so I had originally thought about doing that, but was reluctant to because of the buffer size and also because I thought it was a bit nicer to avoid changing up the existing code while moving things around.

Thinking about it now, though, it's occurred to me that we _do_ actually check at build time that the _max_ length of component ID strings is than `gateway_messages::SpComponent::MAX_ID_LENGTH`, but we presently do that in `task-validate-api`'s build script:
https://github.com/oxidecomputer/hubris/blob/0a78e9412efb512da1f9c9d882ccf4dcf51e2299/task/validate-api/build.rs#L65-L78

If we wanted to rely on that validation here, it might be a bit nicer to move it to `build-i2c`'s codegen. In addition to doing that, we could also be cute and generate a constant in `i2c_config.rs` for the *actual* maximum length of a component ID string, so the buffers used here would need only be that length + 3. That might be worth doing!

We _could_ also make the `build-i2c` codegen generate the component IDs as fixed-size byte arrays like in `gateway_messages::SpComponentId`, but I'd actually prefer not to, because generating them as string constants gets us a `len()` function and also avoids the extra null bytes.

mkeeter: I'm in favor of removing the `name`, and am basically fine with any of the options you proposed.  One particularly easy strategy would be sizing the buffer to `SpComponent::MAX_ID_LENGTH + 3`, instead of anything dynamic.

hawkw: yup, that's what i'm doing

task/host-sp-comms/src/bsp/gimlet_bcde.rs
```diff
@@ -104,16 +99,15 @@ impl ServerImpl {
...
-                name[0..7].copy_from_slice(&designator);
-                self.read_eeprom_barcode(sequence, &name, f)
+                name[0..7].copy_from_slice(&dev_id[0..7]);
+                self.read_eeprom_barcode(sequence, &name, dev)
             }
             30..=39 => {
-                let (designator, f): ([u8; 7], _) =
-                    Self::get_sharkfin_vpd(index as usize - 30);
-                self.read_at24csw080_id(sequence, &designator, f)
+                let dev = Self::get_sharkfin_vpd(index as usize - 14);
```
hawkw: so...this somehow changed from `- 30` to `- 14` and caused a panic, which...how the heck did that happen? anyway, will fix that

build/i2c/src/lib.rs
hawkw: So, we aren't presently, but while working on picking up these changes in #2184, I came up with what I think is a nicer way of doing it (see https://github.com/oxidecomputer/hubris/pull/2184/commits/b027106beb566016694e2828091bd9d6172af89a): we can have the `drv-i2c-api` crate enable a feature flag on the build script when its `component-id` feature is enabled, which keeps them in sync without having to worry about it. If you like that, I can pull it in to this branch?

task/host-sp-comms/src/bsp/cosmo_a.rs
hawkw: done in https://github.com/oxidecomputer/hubris/pull/2203/commits/a80c20db438b3cf82e47d0a8ca9dc54a9b1d3992!

task/host-sp-comms/src/bsp/cosmo_a.rs
hawkw: I think the new code from https://github.com/oxidecomputer/hubris/pull/2203/commits/29f657cbf55e2ad6c5eee925a9d3eaf6054982ca obviates _some_ of the unfortunate bits here.

build/i2c/src/lib.rs
mkeeter: I'm fine leaving it until #2184; you're the only person touching this stuff right now, so it's hard to mess up.

mkeeter (approved):

PR #2205, opened 2025-08-25
Fix missing jsons and update grapefruit with APOB stuff

nathanaelhuffman:
I snapped a new FPGA release since it's cleaner from a tracking perspective but there are no functional changes here vs what is already merged, just a different build.

This also updates the .json files and provides a new grapefruit image with updated grapefruit .json files also.


mkeeter:
While you're in here, can you delete [`spi_nor_reg_map.json`](https://github.com/oxidecomputer/hubris/blob/master/drv/spartan7-loader/cosmo-seq/spi_nor_reg_map.json), which I think is unused?

PR #2206, opened 2025-08-25
Write APOB address to the FPGA when switching mux to host

mkeeter:
This lets us actually use eMCR, and should be robust to changes in EFS directory organization (because we actively parse it in Hubris).

Here's an example on Grapefruit, showing a promising trace:
```
➜  hubris jj:(nwqq) humility hiffy -c HostFlash.set_mux -astate=HostCPU
HostFlash.set_mux() => ()
➜  hubris jj:(nwqq) humility ringbuf hf
humility: ring buffer drv_cosmo_hf::__RINGBUF in hf:
   TOTAL VARIANT
    8232 FpgaBusy
       1 ApobFound
   0  226        1     8232 FpgaBusy
   1  576        1        1 ApobFound(ApobLocation { start: 0x22c000, size: 0xd0000 })
   ```
   
@citrus-it can you test on real hardware?

### DIFF
--- a/drv/cosmo-hf/src/apob.rs
+++ b/drv/cosmo-hf/src/apob.rs
@@ -1,0 +1,120 @@
+//! Tools to extract the APOB location from an AMD ROM
+//!
+//! For details, see AMD document 57299; tables and sections in this code refer
+//! to Rev. 2.0 February 2025.
+
+use crate::hf::ServerImpl;
+use drv_hf_api::HfError;
+use userlib::UnwrapLite;
+use zerocopy::{FromBytes, Immutable, IntoBytes};
+
+/// Embedded firmware structure (Table 3)
+///
+/// Only relevant fields are included here.
+#[derive(FromBytes, Immutable, IntoBytes)]
+#[repr(C)]
+pub struct Efs {
+    signature: u32,
+    _padding1: [u8; 16],
+    psp_dir_offset: u32,
+    _padding2: [u8; 16],
+    bios_dir_offset: u32,
+}
+
+const EFS_SIGNATURE: u32 = 0x55aa55aa;
+const BHD_DIR_COOKIE: u32 = 0x44484224; // $BHD
+const APOB_NV_COPY: u8 = 0x63; // Table 29
+
+/// BIOS Directory Table Header (Table 17)
+#[derive(FromBytes, Immutable, IntoBytes)]
+#[repr(C)]
+pub struct BhdDir {
+    cookie: u32,
+    checksum: u32,
+    num: u32,
+    info: u32,
+}
+
+/// BIOS Directory Table Entry (Table 18)
+#[derive(FromBytes, Immutable, IntoBytes)]
+#[repr(C)]
+pub struct DirEntry {
+    entry_type: u8,
+    region_type: u8,
+    _unused1: u8,
+    _unused2: u8,
+    size: u32,
+    src_address: u64, // highest 2 bits are `addr_mode`
+    dst_address: u64,
+}
+
+impl ServerImpl {
+    /// Reads a typed value from bonus flash
+    fn amd_read<T: FromBytes + Immutable + IntoBytes>(
+        &mut self,
+        addr: u32,
+    ) -> Result<T, HfError> {
+        let mut out = T::new_zeroed();
+        self.drv
+            .flash_read(
+                Self::bonus_addr(addr, core::mem::size_of_val(&out) as u32)?,
+                &mut out.as_mut_bytes(),
+            )
+            .unwrap_lite(); // flash_read is infallible when using a slice
+        Ok(out)
+    }
+
+    /// Find the APOB location from the bonus flash region
+    pub fn find_apob(&mut self) -> Result<ApobLocation, ApobError> {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by citrus-it** [comment]
I've tested this on a Cosmo (BRM15250001) and it works perfectly. The FPGA registers are being set to the right values, and zeroed when the image has no eMCR support. The system is booting with eMCR when programmed with the right OS bits and with the APOB in bonus flash.

File: drv/cosmo-hf/src/apob.rs [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+        self.drv
+            .flash_read(
+                self.flash_addr(addr, core::mem::size_of_val(&out) as u32)?,
+                &mut out.as_mut_bytes(),
+            )
+            .unwrap_lite(); // flash_read is infallible when using a slice
+        Ok(out)
+    }
+
+    /// Find the APOB location from the bonus flash region
```
citrus-it: This comment still references bonus flash.

mkeeter: Good catch, fixed

citrus-it (approved):

drv/cosmo-hf/src/apob.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/// BIOS Directory Table Entry (Table 18)
+#[derive(FromBytes, Immutable, IntoBytes)]
+#[repr(C)]
+pub struct DirEntry {
+    entry_type: u8,
+    region_type: u8,
+    _unused1: u8,
+    _unused2: u8,
```
hawkw: nit, take it or leave it: just to be consistent with the `Efs` struct in this module, why not:
```suggestion
    _padding [u8; 2],
```

drv/cosmo-hf/src/apob.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct ApobLocation {
+    pub start: u32,
+    pub size: u32,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
```
hawkw, line 111: nit: why do these derive `Debug`? do they ever get debug printed? we could probably save a few static strings by getting rid of the debug.
```suggestion
#[derive(Copy, Clone, PartialEq)]
pub struct ApobLocation {
    pub start: u32,
    pub size: u32,
}

#[derive(Copy, Clone, PartialEq)]
```

drv/cosmo-hf/src/apob.rs
```diff
@@ -0,0 +1,124 @@
...
+//! to Rev. 2.0 February 2025.
+
+use crate::hf::ServerImpl;
+use drv_hf_api::HfError;
+use userlib::UnwrapLite;
+use zerocopy::{FromBytes, Immutable, IntoBytes};
+
+/// Embedded firmware structure (Table 3)
+///
+/// Only relevant fields are included here.
```
hawkw: i presume that means there may be further bytes after those defined in this structure, so its size cannot be used to offset past the EFS?

hawkw (commented):
I'm not too familiar with the details of the AMD firmware stuff, but I gave the code a quick review --- overall, looks good to me, with a few very tiny nits.

drv/cosmo-hf/src/apob.rs
mkeeter: This is because `RingbufEntry<T>` requires `T: Debug`.  I don't think that's actually necessary, but a quick experiment to remove it + these `Debug` annotations produced an image with the same hash, so the static strings are already being stripped.

drv/cosmo-hf/src/apob.rs
```diff
@@ -0,0 +1,124 @@
...
+use userlib::UnwrapLite;
+use zerocopy::{FromBytes, Immutable, IntoBytes};
+
+/// Embedded firmware structure (Table 3)
+///
+/// Only relevant fields are included here; the EFS extends beyond the size of
+/// this `struct`, but we don't care about any subsequent fields.
+#[derive(FromBytes, Immutable, IntoBytes)]
+#[repr(C)]
+pub struct Efs {
```
luqmana, line 21: Any reason didn't just use the definitions from [amd-efs](https://github.com/oxidecomputer/amd-efs)? It should be no_std as I believe @daym intended to make it usable from hubris.

luqmana: (Don't take as a blocker, just curious. Very excited for eMCR regardless!)

drv/cosmo-hf/src/apob.rs
mkeeter: Yup, added a comment to that effect

drv/cosmo-hf/src/apob.rs
mkeeter: Fixed!

drv/cosmo-hf/src/apob.rs
mkeeter: I didn't know about it!  Taking a quick look, I think integrating it would be non-trivial, so I'm going to punt on it for now.

PR #2207, opened 2025-08-26
Fix incorrect hash offset

mkeeter:
In the current code, both `hash` and `hash_range_update` apply the `flash_addr` correction (going from relative offset to absolute address).

This means that if `Flash1` is selected, `hash` will end up hashing the Bonus Flash range:
- `hash` applies an offset to get an absolute address in `Flash1`
- `hash_range_update` re-applies that offset to an already-absolute address, pushing the address into Bonus Flash

I fixed this by not applying the offset in `hash` (although I still call `flash_addr` to perform address checks).

### DIFF
--- a/drv/cosmo-hf/src/hf.rs
+++ b/drv/cosmo-hf/src/hf.rs
@@ -618,9 +618,15 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
             ringbuf_entry!(Trace::HashInitError(e));
             return Err(HfError::HashError.into());
         }
-        let begin = self.flash_addr(addr, len)?.get() as usize;
-        let end = begin + len as usize;
-        self.hash_range_update(self.dev, begin, end)?;
+
+        // Check that the hash range is valid.  We **do not** pass the resulting
+        // value to `hash_range_update`, which expects relative offsets!
+        let _check = self.flash_addr(addr, len)?;
+        self.hash_range_update(
+            self.dev,
+            addr as usize,
+            addr as usize + len as usize,
+        )?;
 
         match self.hash.task.finalize_sha256() {
             Ok(sum) => Ok(sum),
--- a/drv/cosmo-hf/src/main.rs
+++ b/drv/cosmo-hf/src/main.rs
@@ -88,9 +88,6 @@ impl FlashAddr {
             None
         }
     }
-    fn get(&self) -> u32 {
-        self.0
-    }
 }
 
 /// Driver for a QSPI NOR flash controlled by an FPGA over FMC


drv/cosmo-hf/src/hf.rs
```diff
@@ -618,9 +618,15 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
             return Err(HfError::HashError.into());
         }
-        let begin = self.flash_addr(addr, len)?.get() as usize;
-        let end = begin + len as usize;
-        self.hash_range_update(self.dev, begin, end)?;
+
+        // Check that the hash range is valid.  We **do not** pass the resulting
+        // value to `hash_range_update`, which expects relative offsets!
+        let _check = self.flash_addr(addr, len)?;
+        self.hash_range_update(
```
jgallagher, line 625: Is this fundamentally different from gimlet? I don't see a `flash_addr()` function there at all, so I assume so.

No objections to this change, but I'm wondering if we should do work (or open an issue to do work) to make similar errors less likely? `flash_addr()` returns `FlashAddr` which is a newtype specifically for absolute addresses; would it also make sense to have a newtype for relative addresses, then change all the methods that take a raw address to take one or the other?

mkeeter: Yeah, Gimlet has physically different devices controlled by a mux, versus Cosmo using a memory offset to split one physical device into two virtual flashes (plus Bonus Flash).

> No objections to this change, but I'm wondering if we should do work (or open an issue to do work) to make similar errors less likely?

The idea was that `FlashAddr` is always an absolute address, and `u32` is always a relative address.  I considered having a dedicated `FlashOffset` type for relative addresses, but we do a _ton_ of math on relative addresses, so we'd have to constantly be converting back and forth...

This PR does remove `.get()`, but people can still access the `.0` member 🤷🏻‍♂️

jgallagher: > > No objections to this change, but I'm wondering if we should do work (or open an issue to do work) to make similar errors less likely?
> 
> The idea was that `FlashAddr` is always an absolute address, and `u32` is always a relative address. I considered having a dedicated `FlashOffset` type for relative addresses, but we do a _ton_ of math on relative addresses, so we'd have to constantly be converting back and forth...
>
> This PR does remove `.get()`, but people can still access the `.0` member 🤷🏻‍♂️

Makes sense. (Although I guess we _could_ implement the math operators on these newtypes?) If it seems like a lot of busywork for not much value, no objection to not doing it!

Issue #2200, opened 2025-08-18
Support for Raspberry Pi Pico 2 aka RP235x


thenewwazoo:
I'm fairly far along porting Hubris to a Raspberry Pico 2 board (specifically a W, but that's "just" a Pico 2 with an SPI-attached radio), and want to check in to see if contributing such a port would be welcome. I understand that Hubris is not intended to be all things to all people, but the Pico 2 seems potentially relevant because:

* the RP235x chips are veeeerrrry popular by dint of being supported by the RPi foundation
* the Cortex-M33 has an MMU, which the RP2040 lacks
* a (now-outdated, draft) RP2040 port [exists](https://github.com/oxidecomputer/hubris/pull/538/files)
* there are interesting secure boot features that may be appealing to Hubris users
* it's possibly a good jumping-off point for RISC-V and multi-core support to already have a cheap board supported

There are a few questions/concerns that have arisen as I've worked:

* I've got most of a porting guide doc in my head - do you want one?
* It's basically necessary to build apps/tasks in a fork anyway; should ports stay in forks as a rule? (e.g. here's [hubris-flipper]). Being able to build apps and tasks out-of-tree would be cool.
* Peripheral abstractions can range from [vry smol] to [exceptionally exhaustive]. The e.g. [LPC55 abstractions] appear to be _fairly_ extensive; how much of the RP235x datasheet should be encoded in _this repository_ given that y'all don't use it?

[hubris-flipper]: https://github.com/pokeylope/hubris-flipper
[vry smol]: https://github.com/oxidecomputer/hubris/blob/3c6cb9588d0561f9c087f8e3ef5330c0b937db21/drv/rp2040-sys-api/src/lib.rs#L78-L88
[exceptionally exhaustive]: https://github.com/thenewwazoo/lpc177x_8x-hal/blob/75d7f5d686b6cc92c12945a4048445ce400bcf1f/src/gpio/mod.rs#L727
[LPC55 abstractions]: https://github.com/oxidecomputer/hubris/blob/6067a065de3373482613b3fa2e90a0e307b2eeb1/drv/lpc55-gpio-api/src/lib.rs#L147

aapoalas:
I expect this would be more welcome in https://github.com/cbiffle/exhubris/

I would personally be pretty interested in that as I have an unused Pico 2W (IIRC) gathering dust in my closet :)

thenewwazoo:
I've submitted PR #2210 as a starting point. 

[...]

PR #2210, opened 2025-08-29
Add support for RP2350 aka Raspberry Pi Pico 2

thenewwazoo:
This change adds very, very basic support for the Raspberry Pi Pico 2's RP2350 chip and its 4 MiB onboard flash. A new demo app is added with only jefe and idle tasks. This is mostly intended as a basis for discussion of #2200.

Nonetheless, it runs and passes tests!

```
$ cargo xtask humility app/demo-pi-pico-2/app.toml -- tasks
     Running `target/debug/xtask humility app/demo-pi-pico-2/app.toml -- tasks`
system time = 859044
 0 jefe                         0   0 recv, notif: fault timer(T+56)
 1 idle                         0   5 RUNNING
 ```
 
**NOTE** that this is intended to pair with an as-yet-unsubmitted [update to humility](https://github.com/thenewwazoo/humility/tree/bmatt/update-probe-rs), the status of which is likely [uncertain](https://github.com/oxidecomputer/humility/issues/530).

 Additional minor tweaks include:
 * some error reporting when files can't be found (emitting the sought filename)
 * emitting a firmware map in the dist directory

### DIFF
--- a/doc/guide/caboose.adoc
+++ b/doc/guide/caboose.adoc
@@ -37,8 +37,8 @@ mapped as an MPU region. Only tasks declared in `caboose.tasks` are allowed to
 read data from the caboose region of flash.  If other tasks attempt to read from
 this memory region, they will experience the typical memory fault.
 
-The Hubris build system will populate the caboose with start and end words
-(32-bit words) words, and a sequence of 
+The Hubris build system will populate the caboose with 32-bit start and end
+words, and a sequence of
 https://github.com/oxidecomputer/tlvc[TLV-C] key-value pairs containing image
 metadata:
 
@@ -48,12 +48,12 @@ metadata:
 | Type
 | Description
 
-| **Start** 
-| `u32` 
+| **Start**
+| `u32`
 | `abi::CABOOSE_MAGIC`
 
-| `GITC` tag 
-| TLV-C 
+| `GITC` tag
+| TLV-C
 | The current Git commit hash with an optional trailing "-dirty" if the
 repository contains uncommitted changes.
 
@@ -74,7 +74,7 @@ if it was set.
 | `u8`
 | _(filled with `0xFF`)_
 
-| **End**   
+| **End**
 | `u32`
 | Caboose size (little-endian `u32`)
 |===
@@ -111,8 +111,8 @@ let caboose: Option<&'static [u8]> = drv_caboose_pos::CABOOSE_POS.as_slice();
 can't know the caboose position until all tasks have been built)
 
 Besides the start and end words and the default metadata described above, the
-Hubris build system is agnostic to any further contents of the caboose. A 
+Hubris build system is agnostic to any further contents of the caboose. A
 separate release engineering process may decide to store any arbitrary data in
-the remaining space. The 
+the remaining space. The
 https://github.com/oxidecomputer/hubtools[`hubtools` repository] includes a
 library and CLI for modifying the caboose of a Hubris archive.

### CODE REVIEW

File: app/demo-pi-pico-2/src/main.rs [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+
+use cortex_m_rt::entry;
+
+/// Image definition to set up the chip for booting
+///
+/// See datasheet 5.1.4 Image Definitions for more. These specific values come from
+/// section 5.9.5. Minimum viable image metadata.
+#[link_section = ".image_def"]
+#[used]
+pub static RP235X_IMAGE_DEF_ARM_MIN: [u32; 5] = [
```
aapoalas, line 20: **suggestion:** This seems like it should rather be a `repr(C)` struct with named fields instead of an array with commented indexes.

**nitpick:** I'd personally go with `usize` types since these seem to be addresses, though the Hubris codebase does use `u32` for addresses very often. It makes sense in the APIs but (IMO) less so inside the services / tasks.

build/xtask/src/config.rs
```diff
@@ -144,7 +144,13 @@ impl Config {
         let mut peripherals: IndexMap<String, Peripheral> = {
             let chip_file =
                 cfg.parent().unwrap().join(&toml.chip).join("chip.toml");
-            let chip_contents = std::fs::read(chip_file)?;
+            let chip_contents =
```
aapoalas, line 147: **praise:** These seem like helpful additions!

build/xtask/src/dist.rs
```diff
@@ -2009,6 +2015,7 @@ fn build(
         format!(
             "-C link-arg=-z -C link-arg=common-page-size=0x20 \
              -C link-arg=-z -C link-arg=max-page-size=0x20 \
+             -C link-arg=-Map=target/firmware.map \
```
aapoalas: **question:** Do I get this right that this makes the linker generate an additional linker map file as output? That seems potentially useful, I wonder if there was a reason why it wasn't already being generated?

aapoalas (commented):
Left a couple of comments from an outside to an outsider.

IMO this is pretty damn cool, and I'll try take this for a spin if I get the time.

That being said: I'm not Lorax and I don't speak for the ~~trees~~Oxide so take this with a large grain of salt, but I would assume this would find a better home in the exhubris repository I linked in the issue you opened. The owner of said repository is one of the lead architects of Hubris and an employee of Oxide, so it's not like it's a far away and lonely place. Rather it's precisely the "home away from home" where the Hubris kernel is intended to find usage outside of the direct Oxide Computer goals. Assuming that Oxide doesn't have plans to take up RPi, having this code in the main tree would effectively be just a bit of extra code they have to checkout or (worse) keep from going stale. They'd presumably also then have to take PRs for the code, and review and merge them since they wouldn't be using or maintaining it themselves. If the code was in exhubris instead, it'd be that bit more usage of Hubris (good for Oxide) but outside of their direct purview (again good) and thus nothing they need to worry about dealing with on company time (excellent!).

But as said, I don't speak for them so don't read this as truth, I'm just guessing here. Great stuff getting Hubris working on RPi <3 !

app/demo-pi-pico-2/src/main.rs
thenewwazoo: I agree, this should eventually be a `repr(C)` struct since it encodes fields with values and flags (in 5.9.3.1). For now I've gone with the raw bytes described in the datasheet (5.9.5.1) since it's an MVP and I'm not even sure Oxide wants this in their repo 😆 If I get any sort of encouragement from them, I expect I'll expand on this and other aspects of the PR.

As for `u32` vs `usize`, in the ultimate structure, it'd be a mix of types, with `u8`/`u16`/`u32` for values and `usize` for the address, as noted. For now, it's just `u32` bytes straight out of the datasheet.

build/xtask/src/config.rs
thenewwazoo: Thanks - I should have called them out in the PR. They were diagnostics added in places I tripped over while working on the port.

build/xtask/src/dist.rs
thenewwazoo: It does - as implemented, it's not scoped to a particular target (so building another app clobbers the last build's firmware.map) but I was using it when working on laying out the image metadata, and I left it. I don't love where it lands. I'll take a note to either remove it, or do it right and figure out how to put it alongside the `dist` directory.

aapoalas: Not sure if you have the toml available here but this code is used elsewhere in this file:

```rust
let dist_dir = Path::new("target").join(&toml.name).join("dist");
```

thenewwazoo:
> Left a couple of comments from an outsider.
> 
> IMO this is pretty damn cool, and I'll try take this for a spin if I get the time.
> 
> Great stuff getting Hubris working on RPi <3 !

Thank you! Please note that I got this working alongside [an update to humility](https://github.com/thenewwazoo/humility/tree/bmatt/update-probe-rs) that is required for it to work, and which is its own ball of wax.

Your suggestion for exhubris is well-heard, and I'm sorry for not responding earlier. I was... focused... on getting hubris booting. I like the idea and goals of exhubris, but it is quite a different product with a different learning curve, and its future and relationship with hubris itself isn't entirely clear (authorship notwithstanding). I decided to just EAFP and see how it went.

build/xtask/src/dist.rs
thenewwazoo: Thanks, that put me on the path. It now lands in `target/${app_name}/dist/firmware.map`.

aapoalas:
> Your suggestion for exhubris is well-heard, and I'm sorry for not responding earlier. I was... focused... on getting hubris booting. I like the idea and goals of exhubris, but it is quite a different product with a different learning curve, and its future and relationship with hubris itself isn't entirely clear (authorship notwithstanding). I decided to just EAFP and see how it went.

Sorry for being vocal about it as well. I removed the section about it in the above comment (as you of course noticed) because after posting it I realised it's not really my place to try tell you what to do. Let's hope this gets some bandwidth from the Oxide folks! :)

PR #2211, opened 2025-09-02
Swap FMC base address

labbott:
The default attributes for the FMC region are normal and executable. This allows speculative accesses. This doesn't apply to most hubris tasks but it does apply to the kernel. Swap the FMC banks to use a different default memory map which marks it as device/execute never to avoid speculation.


mkeeter:
We talked about this in chat, and I think we want to switch to having `peek32` and `poke32` return a `Result` type with a clear error about the old address space.  The change will break existing scripts anyways, and it would be a better UX to break them in a way that leaves a bread crumb.

Alternatively, we could silently remap `0x60000000` to `0xC0000000`.  This... might not be as awful as it sounded in my head?  (network requests could do the same, obviating the need for the new error variant)

labbott:
> 
> Alternatively, we could silently remap `0x60000000` to `0xC0000000`. This... might not be as awful as it sounded in my head? (network requests could do the same, obviating the need for the new error variant)

The remap was much cleaner than I expected. Nobody should be poking anything outside the FMC region anyway.

mkeeter (approved):

Issue #2223, opened 2025-09-11
`cosmo-spd` should support going to A2


mkeeter:
Right now, `cosmo-spd` waits for the system to go to A0 before checking for DIMM presence and starting the thermal polling loop.

However, if we go back to A2, it doesn't stop polling; and on subsequent transitions to A0, it does not re-check presence.

PR #2224, opened 2025-09-12
Handle A0 -> A2 transition in `cosmo-spd`

mkeeter:
Previously, `cosmo-spd` waited for the system to enter A0 before starting up, but didn't do anything if we went back to A2.

This PR moves that initialization into an `activate` function – to be called whenever we see an A2 -> A0 transition – and adds a corresponding `deactivate` function for A0 -> A2 transitions.  The `activate` function checks presence and reloads SPD information into packrat each time it's called, in case someone swapped DIMMs while the system was in A2.

When `cosmo-spd` is active, it both polls temperature sensors and listens for `jefe` state change notifications; when inactive, it _only_ listens for `jefe` notifications.

Tested on the `cosmo-hubris` machine, and it seems to work fine.

(https://github.com/oxidecomputer/quartz/issues/400 remains an issue here)

Fixes https://github.com/oxidecomputer/hubris/issues/2223

### DIFF
--- a/task/cosmo-spd/src/main.rs
+++ b/task/cosmo-spd/src/main.rs
@@ -10,7 +10,7 @@
 use drv_cpu_seq_api::PowerState;
 use drv_spartan7_loader_api::Spartan7Loader;
 use idol_runtime::RequestError;
-use ringbuf::{ringbuf, ringbuf_entry};
+use ringbuf::{counted_ringbuf, ringbuf_entry};
 use task_jefe_api::Jefe;
 use task_packrat_api::Packrat;
 use task_sensor_api::{config::other_sensors, NoData, Sensor, SensorId};
@@ -25,7 +25,7 @@ task_slot!(PACKRAT, packrat);
 task_slot!(LOADER, spartan7_loader);
 task_slot!(SENSOR, sensor);
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(counters::Count, Copy, Clone, PartialEq)]
 enum Trace {
     None,
     Activate { time: u64 },
@@ -34,7 +34,7 @@ enum Trace {
     TemperatureReadTimeout { index: usize, pos: usize },
 }
 
-ringbuf!(Trace, 32, Trace::None);
+counted_ringbuf!(Trace, 32, Trace::None);
 
 #[export_name = "main"]
 fn main() -> ! {
@@ -57,7 +57,7 @@ fn main() -> ! {
     // Wait for entry to A0 before we enable our i2c controller.  Normally,
     // we'd be able to read SPDs in A2, but there's a hardware errata:
     // https://github.com/oxidecomputer/hardware-cosmo/issues/689
-    server.check_active();
+    server.update_state();
 
     set_timer_relative(0, notifications::TIMER_MASK);
     let mut buffer = [0; idl::INCOMING_SIZE];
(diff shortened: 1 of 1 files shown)


task/cosmo-spd/src/main.rs
```diff
@@ -164,75 +73,136 @@ const DIMM_COUNT: usize = 12;
...
+        });
+        self.active = false;
+
+        // Mark all sensors as off
+        for index in 0..DIMM_COUNT {
+            for pos in 0..2 {
+                self.sensor
+                    .nodata_now(DIMM_SENSORS[index][pos], NoData::DeviceOff);
+            }
+        }
```
hawkw: [Clippy dislikes this](https://github.com/oxidecomputer/hubris/actions/runs/17680732830/job/50254058493?pr=2224#step:9:603) and would much rather see you write:
```suggestion
        for dimm in DIMM_SENSORS {
            for sensor in dimm {
                self.sensor
                    .nodata_now(sensor, NoData::DeviceOff);
            }
        }
```

task/cosmo-spd/src/main.rs
```diff
@@ -164,75 +73,136 @@ const DIMM_COUNT: usize = 12;
...
 
-impl idl::InOrderCosmoSpdImpl for ServerImpl {
-    fn ping(
-        &mut self,
-        _mgs: &RecvMessage,
-    ) -> Result<(), RequestError<core::convert::Infallible>> {
-        Ok(())
+impl ServerImpl {
+    /// Get our current state from `jefe` and activate / deactivate as needed
+    fn check_active(&mut self) {
```
hawkw: naming nit, feel free to disregard: i'm not the biggest fan of "check" here, which to me kind of feels like something that would return an error if we're not in a particular state. not immediately sure what would be a better name; perhaps `sync_active` or something? i'm also not going to block merging over this...

task/cosmo-spd/src/main.rs
```diff
@@ -164,75 +73,136 @@ const DIMM_COUNT: usize = 12;
...
+                    10 => s.set_bus1_k(true),
+                    11 => s.set_bus1_l(true),
+                    DIMM_COUNT.. => unreachable!(),
+                }
+            });
+
+            // Read 4x256 bytes from the FPGA's buffer and copy to Packrat
+            self.dimms.spd_rd_ptr.set_addr(0);
+            for i in 0..4 {
+                // Limited by max lease size for Packrat
```
hawkw, line 175: nit, feel free to disregard: i don't suppose there's a constant from the Idol codegen that we could use to express that programmatically? not a big deal.

task/cosmo-spd/src/main.rs
```diff
@@ -28,7 +28,8 @@ task_slot!(SENSOR, sensor);
 #[derive(Copy, Clone, PartialEq)]
 enum Trace {
     None,
-    Ready,
+    Activate { time: u64 },
+    Deactivate { time: u64 },
```
hawkw, line 32: wonder if we ought to add ringbuf counters to track the total number of these events?

task/cosmo-spd/src/main.rs
mkeeter: Switched to `update_state`, which is a little better

task/cosmo-spd/src/main.rs
mkeeter: There is not, unfortunately!

task/cosmo-spd/src/main.rs
mkeeter: Sure!

task/cosmo-spd/src/main.rs
hawkw: Ah well, never mind!

PR #2214, opened 2025-09-03
psc_seq: rectifier ereports

hawkw:
This branch adds ereports to the `psc_seq` task for major PSU events. This is the first step in implementing the ["Rectifier Removal"][flow2] and ["Upstream Power Loss"][flow3] fault-management flows that @rmustacc described in RFD 589.

In particular, we generate ereports for the following events:

- `hw.insert.psu`: reported when a rectifier is inserted (a previously-deasserted presence pin is asserted)
- `hw.remove.psu`: reported when a previously present rectifier is removed (a previously-asserted presence pin is deasserted)
- `hw.pwr.pwr_good.bad`: reported when a rectifier deasserts `POWER_GOOD`
- `hw.pwr.pwr_good.good` : reported when a rectifier that has been restarted due to a fault comes back and reasserts `POWER_GOOD`

All these ereports contain a `fruid` key, which is an object containing the following keys:

- `mfr`: the manufacturer part number of the rectifier read over PMBus (in practice this is always the string "Murata-PS")
- `mpn`: the manufacturer part number of the rectifier read over PMBus (in practice, always the string "MWOCP68-3600-D-RM")
- `serial`: the manufacturer serial of the rectifier, read over PMBus
- `fw_rev`: the manufacturer firmware revision of the rectifier, read over PMbus

In order to also record FRUID when a rectifier is _removed_, the PSC sequencer has to hang onto it in memory, rather than reading it from the rectifier on-demand in when the ereport is recorded.

In addition, the `psu.fault` and `psu.fault_cleared` ereports also record the PMBus status registers of the rectifier, which the control plane can use to diagnose the fault.

With help from @ericaasen, I've tested this change by removing and replacing rectifiers, and by removing input power from a rectifier (and eventually bringing it back). All these events cause the expected sequence of ereports to be generated.

This branch does **not** implement a number of other things we'll need to do in order to fully implement the power shelf fault-management flows from RFD 589. In particular, things I didn't do here include:

- **Monitoring of PMBus `PMALERT_L` pins from the rectifiers.** This change only detects faults by reading the `POWER_GOOD` pins. In practice, this seems sufficient to detect PMBus faults, as the rectifier will deassert its `POWER_GOOD` when faulting. However, if we want to also receive PMBus _warnings_, where the rectifier does *not* deassert `POWER_GOOD`, we'll also need to watch `PMALERT`. Presently, there won't be any such warnings, as we haven't set thresholds for warnings. But, we might want that later.
[...]

PR #2225, opened 2025-09-15
ereport quality of life improvements

hawkw:
This branch makes several improvements to the packrat ereport APIs based on experience using them. It consists of three commits that make independent changes but which depend on each other. I'd like to rebase-merge this, and the three commits can be reviewed independently in series. The changes here are:

1. **packrat: return an error if an ereport was lost (183d1555ec81fc6f2c52d62c4d9a93a17289e43c)**
   
   This commit changes `task-packrat-api`'s `deliver_ereport` IPC to return
   an error if the ereport did not fit in the buffer, rather than being
   infallible. In the future, this may allow some tasks to attempt
   re-delivery periodically until there's space.
2. **packrat: Add helper API for `serde` ereports (49710b0bdd02653914219ea79d4cbac5d7623825)**

   In practice, so far, most ereport payloads have been defined as structs
   implementing `serde::Serialize` and encoded using the `minicbor-serde`
   adapter crate. This requires a bit of boilerplate code, so it seems
   worthwhile to provide a method that just does that. This commit adds one
   in `task-packrat-api`. It's feature-flagged so we can avoid compiling
   `minicbor`, `minicbor-serde`, and `serde` in tasks that don't submit
   ereports using this.
3. **cpu_seq: Encode ereports into static buffers (dc47a6042a192b5d4eb5c09e6ceb4aa5b53b1ac4)**

   In [this comment][1], @cbiffle reminded me that we can use
   `ClaimOnceCell` to use a `static` to create byte buffers for encoding
   ereports. This way, the buffer is never on the stack, reducing the risk
   of stack overflows when recording an ereport.

   This commit changes `drv_gimlet_seq_server` and `drv_cosmo_seq_server`
   to use that technique for their ereports.
    
[1]: https://github.com/oxidecomputer/hubris/pull/2214#discussion_r2337531694


drv/gimlet-seq-server/src/main.rs
```diff
@@ -201,9 +201,13 @@ struct ServerImpl<S: SpiServer> {
     hf: hf_api::HostFlash,
     vcore: vcore::VCore,
     deadline: u64,
+    // Buffer for encoding ereports. This is a static so that it's not on the
+    // stack when handling interrupts.
+    ereport_buf: &'static mut [u8],
```
mkeeter: Nit: this is a `&mut [u8]` here and a `&mut [u8; EREPORT_BUF_LEN]` in `cosmo-seq-server`.  Any reason for the inconsistency?

idl/packrat.idol
```diff
@@ -120,7 +120,11 @@ Interface(
             leases: {
                 "data": (type: "[u8]", read: true, max_len: Some(1024)),
             },
-            reply: Simple("()"),
+
```
mkeeter: Extra newline?
```suggestion
```

drv/gimlet-seq-server/src/main.rs
hawkw: Nope, I just wasn't thinking too hard about it. Fixing!

PR #2226, opened 2025-09-15
Add Sidecar image for Reverso

mkeeter:
"Reverso" is a board with the same form factor as a Gimlet which loops back the two backplane connectors.

In this configuration, packets leaving the VSC7448 destined for SP _i_ instead arrive at the peer Sidecar, appearing to have originated at SP _i_.

In the unlocked VLAN configuration, this is a problem: packets can arrive at one tech port and emerge from the other tech port, which creates a routing loop.  When this happens, our CM's network unceremoniously turns off the offending network port, and we have to manually get it re-enabled.

This PR adds a `VlanTargets` parameter to unlock-related code:

- `VlanTargets::EverySp` is our existing behavior, where the technician port can communicate with every cubby
- `VlanTargets::ScrimletOnly` limits technician ports to only communicate with Scrimlet cubbies, which are guaranteed to be populated

By preventing the technician port from talking to cubbies with Reverso installed, we can avoid network loops.

The switch zone in the scrimlet can still communicate with every cubby; indeed, this is how we expect to test connectivity with Reverso installed.

### DIFF
--- a/app/sidecar/rev-d-reverso.toml
+++ b/app/sidecar/rev-d-reverso.toml
@@ -1,0 +1,2 @@
+name = "sidecar-d-reverso"
+inherit = ["rev-d-dev.toml", "reverso.toml"]


Aaron-Hartwig:
I think this is looking good so far. We will also want a rev-d variant here as that is what forward production is. Right now our test rack at the CM has a C and D vintage Sidecar.

mkeeter:
@eanders01 can you let me know when the new butler is installed so I can test this out?

eanders01:
> @eanders01 can you let me know when the new butler is installed so I can test this out?

Yes! Awaiting shipment from @jclulow. Ready to receive and will let you know!

mkeeter:
This is no longer necessary – we're shipping lab machines with 4 NICs, so there's no intermediate switch to get mad about routing loops.

jclulow:
@mkeeter I fear we may actually need this after all: we have **silver** connected to a rack with ~10 reversos, and while it had connectivity for a little while, it was sporadic, and subsequently seems to have stopped working at all.  I can no longer seem to reach SPs or the switch zone, though I can still see broadcasts from them, and a lot of other possibly looped traffic.  We haven't been thrown off the network (a definite improvement!) but it's also not currently working the way we'd like.

Would it help to look at packet dumps from snoop on the interfaces connected to the techports in the rack with reversos?

hawkw:
Hello! I have rebased Matt's branch here onto the latest `master` and confirmed that the `sidecar-d-reverso` still builds cleanly on my machine. So that's good!

hawkw:
Er, I meant to click "reopen and comment" with that last comment, which, apparently, is the only way to re-open a PR?

hawkw:
Uh. Wacky. That button just doesn't do anything. I guess I will open a fresh PR.

PR #2228, opened 2025-09-16
Disable SWD pins when debugger is connected

### LINKED ISSUE #2222: RoT needs to release SWD lines and treat the SP as reset (and thus mutated) if a dongle is detected
Yesterday @hawkw were attempting to resuscitate an ailing Gimlet (spoiler: we made it worse). Among the stack of problems we had to address before we could get to the one at hand: we couldn't raise the SP over SWD by connecting a debug harness.

Manually forcing the RoT into reset using an external board fixed this.

Since the RoT drives the SP's SWD port for attestation and monitoring purposes, it's intended to allow debugger use by releasing those pins -- otherwise, it's fighting to drive the nets. It appears that we don't have any code to actually _do_ this.

There is a `dongle_detected` routine that is polled when an SP reset is observed (or, I think, on RoT startup?) and which will suppress setup of the SWD pins. https://github.com/oxidecomputer/hubris/blob/master/drv/lpc55-swd/src/main.rs#L1572-L1580

However, nothing will _reverse_ setup of the SWD pins and restore them to high-impedance state, as far as I can tell (@mkeeter was also unable to find this code).

This also has implications for attestation: the appearance of a dongle on the SP SWD interface suggests that the state in the SP may be arbitrarily modified, which means the RoT needs to treat it as potentially corrupt, refuse to attest to its integrity, and generally wait for it to reset into a known state.

Changes that we likely need to make here:
1. The dongle detect line needs to be proactively monitored with a pin change interrupt, not passively polled when other signals change.
2. Detection of a dongle needs to cause prompt release of any SWD lines the RoT is driving.
3. It also needs to cause invalidation of any measurements of the SP's contents and a return to the "waiting for reset" state in the state machine.

I may have missed something there, I'm paging this state back in during OxCon.

mkeeter:
Fixes #2222

This PR adds state to the RoT's SWD task, tracking whether it is "connected" (pins configured for SPI) or "disconnected" (pins configured as GPIOs).  When the `JTAG_DETECT` line fires, we switch to disconnected; when we attempt to communicate with the SP, we try to reconnect.

There are two new generated functions: `switch_io_connected` and `switch_io_disconnected`.  In addition, I extracted the repeated "build a helper function to set pins" logic in the build script.

I tested this on a grapefruit by resetting the SP with / without debugger attached, and confirmed that measurements were made when the debugger was detached.  In addition, I could see in the RoT ringbufs that it detected the presence of the debugger and switched IO pins to inputs.

Other small cleanups:
- Line wrapping comments
- Remove the unused `db_reset_sp` function


app/oxide-rot-1/app.toml
```diff
@@ -114,6 +114,18 @@ in_cfg = [
...
+    { pin = { port = 0, pin = 8 }, alt = 3 },
+    { pin = { port = 0, pin = 9 }, alt = 0, mode = "pulldown" },
+    { pin = { port = 0, pin = 7 }, alt = 3 },
+]
+# Connected = MOSI off, MISO off, SCK off
+disconnected_cfg = [
+    { pin = { port = 0, pin = 8 }, alt = 0, mode = "pulldown" },
+    { pin = { port = 0, pin = 9 }, alt = 0, mode = "pulldown" },
+    { pin = { port = 0, pin = 7 }, alt = 0, mode = "pulldown" },
+]
```
hawkw: Was about to ask whether this could be defined in the `base.toml` but then i realized...we don't have one for the RoT! Carry on!

app/oxide-rot-1/app-dev.toml
```diff
@@ -127,6 +127,18 @@ in_cfg = [
     { pin = { port = 0, pin = 9 }, alt = 3 },
     { pin = { port = 0, pin = 8 }, alt = 0, mode = "pulldown" },
 ]
+# Connected = MOSI on, MISO off, SCK on
+connected_cfg = [
+    { pin = { port = 0, pin = 8 }, alt = 3 },
+    { pin = { port = 0, pin = 9 }, alt = 0, mode = "pulldown" },
+    { pin = { port = 0, pin = 7 }, alt = 3 },
+]
+# Connected = MOSI off, MISO off, SCK off
```
hawkw: shouldn't this be:
```suggestion
# Disconnected = MOSI off, MISO off, SCK off
```

app/oxide-rot-1/app.toml
```diff
@@ -114,6 +114,18 @@ in_cfg = [
     { pin = { port = 0, pin = 9 }, alt = 3 },
     { pin = { port = 0, pin = 8 }, alt = 0, mode = "pulldown" },
 ]
+# Connected = MOSI on, MISO off, SCK on
+connected_cfg = [
+    { pin = { port = 0, pin = 8 }, alt = 3 },
+    { pin = { port = 0, pin = 9 }, alt = 0, mode = "pulldown" },
+    { pin = { port = 0, pin = 7 }, alt = 3 },
+]
+# Connected = MOSI off, MISO off, SCK off
```
hawkw: and this?
```suggestion
# Disconnected = MOSI off, MISO off, SCK off
```

drv/lpc55-swd/src/main.rs
```diff
@@ -715,44 +711,40 @@ impl NotificationHandler for ServerImpl {
...
             //  de-asserted so that the SP_RESET that also fired could be
             //  handled successfully.
-            //}
+            ringbuf_entry!(Trace::SpResetFired);
+            if matches!(self.state, SwdState::Connected { .. })
+                && !self.do_handle_sp_reset()
+            {
+                // If handling the SP reset failed, clear the attestation log
+                ringbuf_entry!(Trace::InvalidateSpMeasurement);
+                let _ = self.invalidate_sp_measurement();
```
hawkw: previously there was a comment here saying that `invalidate_sp_measurement()` logs to the ringbuf: https://github.com/oxidecomputer/hubris/pull/2228/files#diff-703eba1954e126d581c58dcb778473d2547077100facfcf04514b30ef3fc2c35L764-L768

IIUC that function hasn't changed here, is that still the case?

hawkw (commented):
It might be nice to have some high-level documentation around the state machine here in the top-level block comments in `drv/lpc55-swd`'s `main.rs`?

drv/lpc55-swd/src/main.rs
mkeeter: That comment was to explain why it was okay to discard the `Result`.  However, we _never_ use the returned value, so I've just changed `invalidate_sp_measurements` to not return anything.

app/oxide-rot-1/app.toml
```diff
@@ -114,6 +114,18 @@ in_cfg = [
     { pin = { port = 0, pin = 9 }, alt = 3 },
     { pin = { port = 0, pin = 8 }, alt = 0, mode = "pulldown" },
 ]
+# Connected = MOSI on, MISO off, SCK on
+connected_cfg = [
+    { pin = { port = 0, pin = 8 }, alt = 3 },
+    { pin = { port = 0, pin = 9 }, alt = 0, mode = "pulldown" },
+    { pin = { port = 0, pin = 7 }, alt = 3 },
+]
```
labbott, line 122: A separate connected_cfg feels a little redundant given we already have the `out_cfg` and `in_cfg`. Could we move configuration of `SCK` into the `in_cfg` and `out_cfg` and assume we will already need to program in vs. out and drop the `connected_cfg` ?

drv/lpc55-gpio/src/main.rs
```diff
@@ -176,79 +176,79 @@ impl idl::InOrderPinsImpl for ServerImpl<'_> {
...
+    ) -> Result<(), RequestError<core::convert::Infallible>> {
+        let mask = pint_slot.mask();
+        match flag {
+            PintFlag::Rising => {
+                self.pint.rise.write(|w| unsafe { w.bits(mask) })
+            }
+            PintFlag::Falling => {
+                self.pint.fall.write(|w| unsafe { w.bits(mask) })
+            }
+            PintFlag::Both => self.pint.ist.write(|w| unsafe { w.bits(mask) }),
```
labbott, line 199: Can you make a comment here to double check this if we do level checking? per Table 404.
`Write 1 (level-sensitive): Switch the active level for this pin (in the IENF register).`

mkeeter: Sure, done in 660eb243ddf6

app/oxide-rot-1/app.toml
mkeeter: I did consider that, but decided against it – switching from `in_cfg` to `out_cfg` (and vice versa) is timing-critical, since it happens during SWD transactions, so I didn't want to add additional operations.  Let me know if you feel strongly one way or the other here!

labbott: that's a good argument for the setup you have here so let's leave it.

mkeeter:
> It might be nice to have some high-level documentation around the state machine here in the top-level block comments in `drv/lpc55-swd`'s `main.rs`?

@hawkw  I'd like to punt on this for now – I'd like to do a larger refactoring into a more explicit state machine, but don't want that to hold up this PR.

drv/lpc55-swd/build.rs
```diff
@@ -36,12 +38,41 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
     let dest_path = out_dir.join("swd.rs");
     let mut file = std::fs::File::create(dest_path)?;
 
-    let out_cfg = &config.out_cfg;
-    let in_cfg = &config.in_cfg;
     let spi_periph = quote::format_ident!("Fc{}", config.spi_num);
     let flexcomm = quote::format_ident!("FLEXCOMM{}", config.spi_num);
     let spi_regs = quote::format_ident!("SPI{}", config.spi_num);
 
+    let gpio_config_fn = |pins: &[PinConfig], name| {
```
cbiffle: This doesn't look like it's closing over any other locals, in which case I'd suggest rewriting it as a `fn` rather than a closure. Among other things, debugger feedback is clearer (it'll have a name).

If I've missed the part where it's closing over state, ignore me.

drv/lpc55-swd/build.rs
```diff
@@ -36,12 +38,41 @@ fn generate_swd_functions(config: &TaskConfig) -> Result<()> {
...
+                    // though.
+                    unsafe {
+                        core::ptr::write_volatile(base as *mut u32, conf);
+                    }
+                )*
+            }
+        }
+    };
+
+    let switch_io_out = gpio_config_fn(&config.out_cfg, "switch_io_out");
```
cbiffle: (Thought: these could be a loop, e.g. over an array of `[(&config.out_cfg, "switch_io_out"), (&config_in.cfg, "switch_io_in"), ...]` -- then instead of factoring out a function above, it could just be the body of the loop)

drv/lpc55-swd/src/main.rs
```diff
@@ -651,28 +639,32 @@ impl NotificationHandler for ServerImpl {
...
-                    invalidate = true;
-                    self.next_use_must_setup_swd();
+                    self.invalidate_sp_measurement();
+
+                    // Cancel ongoing transactions
+                    self.transaction = None;
+                    self.state = SwdState::Disconnected;
+                    switch_io_disconnected();
                 }
+                let _ = gpio.pint_clear(SLOT, PintFlag::Both);
```
cbiffle: Hmmm. It looks like we're using pin change interrupts for the falling edge, and polling detection (below) for the high state. So why is this clearing `Both`?

drv/lpc55-swd/src/main.rs
```diff
@@ -651,28 +639,32 @@ impl NotificationHandler for ServerImpl {
...
-                unreachable!();
+                // Spurious notification, ignore it
             }
-            let _ = gpio.pint_op(SLOT, PintOp::Clear, PintCondition::Status);
             sys_irq_control(notifications::JTAG_DETECT_IRQ_MASK, true);
+        } else if matches!(self.state, SwdState::Disconnected)
+            && !self.swd_dongle_detected()
+        {
+            switch_io_connected();
+            self.state = SwdState::Connected { init: false };
```
cbiffle: So, reading over this change, I'm having a hard time convincing myself that every update to this state is accompanied by an actual reconfiguration of the pins. Should the maintenance of the state variable move into the various `switch_io_*` routines to simplify reasoning locally about it?

drv/lpc55-swd/src/main.rs
```diff
@@ -715,44 +707,34 @@ impl NotificationHandler for ServerImpl {
...
+            {
+                // If handling the SP reset failed, clear the attestation log
+                ringbuf_entry!(Trace::InvalidateSpMeasurement);
+                self.invalidate_sp_measurement();
+            }
+            self.next_use_must_setup_swd();
 
+            // Squelch the interrupts generated by this handler's toggling of
+            // SP_RESET.
             const SLOT: PintSlot = ROT_TO_SP_RESET_L_IN_PINT_SLOT;
```
cbiffle: I'd suggest inlining this constant since it appears to only be used in one place now.

drv/lpc55-swd/build.rs
mkeeter: It was closing over "I don't want to import `proc_macro2` to give an explicit type to `name`" 😛 

Moving the logic into a loop cleans this up, done in f7eb5e07dd77

drv/lpc55-swd/build.rs
mkeeter: Done in f7eb5e07dd77 (there's still a closure, but it's a `map` function on an array, which is unobjectionable)

drv/lpc55-swd/src/main.rs
mkeeter: Done in f10cdc98e671

drv/lpc55-swd/src/main.rs
mkeeter: I don't want to move that logic into `switch_io_*`, which are generated by the `build.rs` to toggle pin states, but I added `on_swd_connected` and `on_swd_disconnected` in 622b50cf6722

drv/lpc55-swd/src/main.rs
mkeeter: This is historical (previously code was `let _ = gpio.pint_op(SLOT, PintOp::Clear, PintCondition::Status);`)

Fixed in c7b11dfe8719

lzrd (approved):
Very good. Thanks!

PR #2230, opened 2025-09-16
Fix code that assumes timer notification => timer fired

cbiffle:
There are two main reasons why a timer notification bit might be set without the timer firing:

1. Subtle state machine bug in previous iteration (not super common but it has happened).

2. Another task, accidentally or deliberately, posting to the bit.

This changes the cases I could easily find to check that their timers have actually expired before acting.

### DIFF
--- a/drv/minibar-ignition-server/src/main.rs
+++ b/drv/minibar-ignition-server/src/main.rs
@@ -399,11 +399,12 @@ impl idol_runtime::NotificationHandler for ServerImpl {
     }
 
     fn handle_notification(&mut self, _bits: u32) {
-        if sys_get_timer().deadline.is_some() {
+        let timer = sys_get_timer();
+        if timer.deadline.is_some() {
             return;
         }
 
-        let start = sys_get_timer().now;
+        let start = timer.now;
 
         // Only poll the presence summary if the port count seems reasonable. A
         // count of 0xff may occur if the FPGA is running an incorrect


mkeeter (approved):

drv/gimlet-seq-server/src/main.rs
```diff
@@ -545,7 +545,9 @@ impl<S: SpiServer> NotificationHandler for ServerImpl<S> {
             self.vcore.handle_notification(self.ereport_buf);
         }
 
-        if (bits & notifications::TIMER_MASK) == 0 {
+        if (bits & notifications::TIMER_MASK) == 0
+            || sys_get_timer().deadline.is_some()
+        {
```
hawkw, line 550: i considered suggesting that this just check `deadline.is_some()` for simplicity but i guess if the notification bit is unset we can avoid the syscall to `sys_get_timer`...

drv/ignition-server/src/main.rs
```diff
@@ -421,6 +421,10 @@ impl idol_runtime::NotificationHandler for ServerImpl {
     }
 
     fn handle_notification(&mut self, _bits: u32) {
+        if sys_get_timer().deadline.is_some() {
+            return;
+        }
+
         let start = sys_get_timer().now;
```
hawkw: can we just do one `sys_get_timer` call here?
```suggestion
        let timer = sys_get_timer();
        if timer.deadline.is_some() {
            return;
        }

        let start = timer.now;
```

drv/ignition-server/src/main.rs
```diff
@@ -421,6 +421,10 @@ impl idol_runtime::NotificationHandler for ServerImpl {
     }
 
     fn handle_notification(&mut self, _bits: u32) {
+        if sys_get_timer().deadline.is_some() {
```
hawkw: should this check the notification bits first so we can avoid calling `sys_get_timer()` if the timer notification bit is unset?

drv/meanwell/src/main.rs
```diff
@@ -112,6 +112,10 @@ impl NotificationHandler for ServerImpl {
     }
 
     fn handle_notification(&mut self, _bits: u32) {
+        if userlib::sys_get_timer().deadline.is_some() {
```
hawkw, line 115: similarly, might we want to check that the bit is unset before we do a syscall?

drv/minibar-ignition-server/src/main.rs
```diff
@@ -399,6 +399,10 @@ impl idol_runtime::NotificationHandler for ServerImpl {
     }
 
     fn handle_notification(&mut self, _bits: u32) {
+        if sys_get_timer().deadline.is_some() {
+            return;
+        }
+
         let start = sys_get_timer().now;
```
hawkw: again we probably don't need to do the syscall twice in a row here?
```suggestion
        let timer = sys_get_timer();
        if timer.deadline.is_some() {
            return;
        }

        let start = timer.now;
```

drv/sidecar-seq-server/src/main.rs
```diff
@@ -775,6 +775,10 @@ impl NotificationHandler for ServerImpl {
     }
 
     fn handle_notification(&mut self, _bits: u32) {
+        if sys_get_timer().deadline.is_some() {
```
hawkw, line 778: could check the bit first?

drv/ignition-server/src/main.rs
cbiffle: There's an implicit assumption here (present before this change) that there's only one notification in use in this task, so if we get here, the bit is set.

PR #2232, opened 2025-09-17
Add `NotificationBits` type; use it in `sys_recv_notification`

mkeeter:
This is a building block to make it harder to mis-handle spurious notifications.  Eventually, I'd like to make `NotificationHandler::handle_notification` in Idol also use this type!

Staged on top of #2230


drv/psc-seq-server/src/main.rs
```diff
@@ -470,9 +470,7 @@ fn main() -> ! {
         // Wait for a pin change or timer.
         let n = sys_recv_notification(sleep_notifications);
         // If the timer bit is set _and the timer has actually fired_...
-        if n & notifications::TIMER_MASK != 0
-            && sys_get_timer().deadline.is_none()
-        {
+        if n.is_timer_set(notifications::TIMER_MASK) {
```
cbiffle: I think, from this use example, that this name is wrong. "Timer set" kind of implies that you set the timer, using `sys_set_timer`.

What about `has_timer_fired`

drv/stm32xx-i2c/src/lib.rs
```diff
@@ -1199,30 +1199,27 @@ fn wfi_raw(event_mask: u32, timeout: I2cTimeout) -> I2cControlResult {
     loop {
         let received = sys_recv_notification(event_mask | TIMER_NOTIFICATION);
 
-        // If the event arrived _and_ our timer went off, prioritize the event and
-        // ignore the timeout.
-        if received & event_mask != 0 {
+        // If the event arrived _and_ our timer went off, prioritize the event
+        // and ignore the timeout.
+        if received.is_set(event_mask, || true) {
```
cbiffle: It'd be nice to backport the utility function for this from your subsequent PR, or switch this to checking the raw bits. I'm comfortable either way.

sys/userlib/src/lib.rs
```diff
@@ -330,16 +330,41 @@ pub fn sys_recv(
     }
 }
 
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub struct NotificationBits(u32);
```
cbiffle, line 339: I'm trying to keep public things in userlib with doc comments, can we add one here?

cbiffle, line 339: This should be `repr(transparent)` and, tbh, it's not clear that it should implement `Eq` and `PartialEq` (since it's a set -- how often are we checking that a full set of bits matches?).

Anything that _does_ want to check the full set (e.g. the kernel test suite, I'm betting) can use the raw bits.

sys/userlib/src/lib.rs
```diff
@@ -330,16 +330,41 @@ pub fn sys_recv(
...
+impl NotificationBits {
+    /// Wraps a `u32`
+    #[inline]
+    pub fn new(bits: u32) -> Self {
+        Self(bits)
+    }
+
+    /// Checks whether the given mask is active, along with `cond` being true
+    #[inline]
+    pub fn is_set<F: Fn() -> bool>(&self, mask: u32, cond: F) -> bool {
```
cbiffle: Per conversation in chat, how about something like `check_condition` for this

sys/userlib/src/lib.rs
```diff
@@ -330,16 +330,41 @@ pub fn sys_recv(
...
+pub struct NotificationBits(u32);
+
+impl NotificationBits {
+    /// Wraps a `u32`
+    #[inline]
+    pub fn new(bits: u32) -> Self {
+        Self(bits)
+    }
+
+    /// Checks whether the given mask is active, along with `cond` being true
```
cbiffle: I'd like this to give more usage advice, what about something like

Checks if a condition signaled by notification bits holds. The `cond` function should verify the underlying condition, but as an optimization, it will only be called if any bits set in `mask` are also set in `self`. This double-check is important, because notification bits can potentially be set even if the underlying condition is not true, such as with extra pended interrupts or inter-task use of `sys_post`.

If you find yourself passing `|| true` here, see the (name of other function)

sys/userlib/src/lib.rs
```diff
@@ -330,16 +330,41 @@ pub fn sys_recv(
...
+    #[inline]
+    pub fn is_set<F: Fn() -> bool>(&self, mask: u32, cond: F) -> bool {
+        (self.0 & mask) != 0 && cond()
+    }
+
+    /// Checks whether a timer notification has expired
```
cbiffle: More precisely: `Checks if a timer appears to have gone off, because the timer_mask bit is set and the kernel timer is no longer set.`

drv/stm32xx-i2c/src/lib.rs
mkeeter: I backported `check_notification_mask` from the subsequent PR

drv/psc-seq-server/src/main.rs
mkeeter: I like it, done!

sys/userlib/src/lib.rs
mkeeter: I added `PartialEq, Eq` because `nucleo-user-button` logs the result of `sys_recv_notification` into a ringbuf (which requires `PartialEq` to compare new values).  With the addition of `get_raw_bits`, I'll just log the `u32` instead and remove the derives.

sys/userlib/src/lib.rs
mkeeter: I like it, done.

sys/userlib/src/lib.rs
mkeeter: Sure, done!

sys/userlib/src/lib.rs
mkeeter: Done!

sys/userlib/src/lib.rs
mkeeter: Sure, done!

cbiffle (approved):
Nice cleanup!

PR #2234, opened 2025-09-17
bump `toml` and `toml_edit`

mkeeter:
The pinned version of `toml_edit` silently discards heterogeneous (mixed inline / block) tables.

This is unfortunate.


cbiffle (approved):

cbiffle:
toml-rs's `preserve_order` continues to be the only non-additive Cargo feature that has ever actually caused a bug in any software I support. Sigh.

PR #2238, opened 2025-09-19
stm32xx-sys: fix glitch on initial pin configuration

### LINKED ISSUE #1824: I2C is generating 600-750ns glitches, and it should feel bad about this.
This is another episode in the hunt for the Gimlet Disk Wumpus. See also #1821 #1822 #1823.

**tl;dr:** we're glitching the data lines on I2C again, and since this violates setup/hold, devices can respond to this in arbitrarily annoying ways.

Example trace:

2024-07-12-144459_3472x1547_scrot

This is just after we stop the bus reset process (which we started for no obvious reason, see #1823) and resume normal service. We are generating ~680ns negative glitches on both lines, which means that when we reconfigure the pins, we're taking them through low-state push-pull at least briefly on their way back to peripheral-controlled open-drain.

This event is not unique; I see a couple of glitches per minute on average in otherwise working Gimlet traces. (I _think_ they are all related to bus resets.)

These glitches are lengthy enough to bypass the I2C standard glitch filter (50ns), but short enough that they violate the I2C spec's setup and hold requirements. (They're also illegal in the protocol state machine, but, so are a lot of things.) This could potentially trigger metastability or misbehavior of devices on the I2C bus.

We had behavior like this up until February 2023 which was resulting in bus lockups, discussed in #1126, so I would assume that this could cause bus lockups.

cbiffle:
I wound up here while hunting #1824, which I had assumed to be I2C-specific. It isn't! It just isn't visible on most other interfaces for subtle reasons.

The GPIO setup code was writing registers in the following order:

1. MODER (selecting from: input, output, alternate, analog)
2. OTYPER (selecting push-pull vs open-drain)
3. OSPEEDR (selecting slew rate)
4. PUPDR (enabling optional pull resistors)
5. AFRx (choosing an alternate function

At reset, the vast majority of GPIOs are in analog mode. This write order meant that, when switching any pin to Alternate mode after reset, we would briefly expose AF0 before switching to the correct Alternate Function mux value. AF0 is not actually implemented on most pins, and in practice seems to be equivalent to a constant 0 output.

This meant that in cases where a pin has an external pullup resistor, the pin would show a very brief negative pulse before releasing. This was most obvious on I2C, where external pullup resistors are standard, and where a random negative glitch on a pin is Bad, Actually (per the standard).

This commit reorders the register writes and more clearly enumerates the conditions it's trying to uphold. When switching between two arbitrary peripherals, you need to go through something like Analog mode to avoid a visible glitch -- but afaict we never do this. When switching between a peripheral and Analog mode for tristating I2C pins, the new order ensures glitch-free operation, which I've verified on my logic analyzer.

Fixes #1824.

### DIFF
--- a/drv/stm32xx-gpio-common/src/server.rs
+++ b/drv/stm32xx-gpio-common/src/server.rs
@@ -120,12 +120,29 @@ impl<T: GpioPeriph> AnyGpioPeriph for T {
         let mask_4l = lsbs_4l.wrapping_mul(0b1111);
         let mask_4h = lsbs_4h.wrapping_mul(0b1111);
 
-        // MODER contains 16x 2-bit fields.
-        let moder_val = u32::from(atts & 0b11);
-        self.moder().modify(|r, w| unsafe {
-            // See comment re: wrapping_mul above.
-            w.bits((r.bits() & !mask_2) | moder_val.wrapping_mul(lsbs_2))
-        });
+        // Register updates -- we do this carefully in a specific order to try
+        // to minimize glitching. Specifically, we're trying to maintain two
+        // properties:
+        //
+        // 1. The initial configuration of a pin (out of the Analog mode chosen
+        //    at reset) should appear atomic without visible glitches, with the
+        //    exception of enabling a pull up/down resistor, which if requested
+        //    cannot be done atomically.
+        //
+        // 2. _Reconfiguration_ of a pin between (input, output, alternate,
+        //    analog) modes should be atomic as long as the drive type and pull
+        //    mode are not changed.
+        //
+        // We can do this by writing registers in sequence:
+        // - OTYPER (open-drain), OSPEEDR (slew rate) in any order
+        // - PUPDR (activation of pull up/down resistor) at some point
+        // - AFRx (alternate peripheral function select)
+        // - MODER last
+        //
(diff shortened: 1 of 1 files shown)


dancrossnyc (approved):

drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -120,12 +120,28 @@ impl<T: GpioPeriph> AnyGpioPeriph for T {
...
+        //    analog) modes should be atomic as long as the drive type and pull
+        //    mode are not changed.
+        //
+        // We can do this by writing registers in sequence:
+        // - OTYPER, OSPEEDR in any order
+        // - PUPDR at some point
+        // - AFRx
+        // - MODER last
+        //
+        // (Because the effect of PUPDR is inherently non-atomic and separately
```
hawkw, line 142: nit, take it or leave it: perhaps worth briefly saying _what_ "the effect of PUPDR" is here, so the reader doesn't have to go look it up?

cbiffle: Fixed!

hawkw: _oh_, `PUP` means "pullup", that makes sense --- i was wondering what the "puppy register" did...

PR #2239, opened 2025-09-19
jefe: fix comparison inversion in timeout handling

cbiffle:
We recently taught Jefe to impose a restart delay before standing back up a crashing task. However, the implementation contained a logic error that could manifest in certain corner cases: it would actually restart the task _the first time it considered it,_ but only if the deadline hadn't actually elapsed yet.

Once the deadline elapsed, Jefe would enter an infinite loop of setting its own timer into the past and responding to the notification, never yielding the CPU.

It turns out this is relatively easy to trigger by putting a task into a fast crashloop -- I hit this when forcing the I2C driver to crash while chasing an unrelated bug.

This commit flips the timestamp comparison logic to implement the restart timeout for reals, causing Jefe to tolerate fast crashloops without taking the system down.

### DIFF
--- a/task/jefe/src/main.rs
+++ b/task/jefe/src/main.rs
@@ -358,19 +358,26 @@ impl idol_runtime::NotificationHandler for ServerImpl<'_> {
         external::check(self.task_states, now);
 
         if bits.has_timer_fired(notifications::TIMER_MASK) {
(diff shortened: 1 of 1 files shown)


hawkw (approved):
Looks good to me, and I appreciate the added comments!

PR #2242, opened 2025-09-22
cpu-seq: Add more ereports

hawkw:
Depends on #2246

This branch adds new ereports to `drv_gimlet_seq_server` and
`drv_cosmo_seq_server` for failures going to A0 and for events that
result in unexpected power-offs, such as THERMTRIP, sequencer FPGA A0
MAPO, and SMERR assertions. This depends on the `microcbor` derives for
CBOR encoding I added in #2246, which make it much easier for a task to
record a variety of ereports represented by different Rust types,
without having to worry too much about the maximum length of the
encoding buffer.

This branch isn't ready to merge yet, as it depends on #2246. However,
I'd love to get reviews on this now, mostly to get some feedback on the
data included in these ereports and the class hierarchy/general message
schemas. I've kinda made these up; if there's more information that
would be useful to include in these, I'd love to get some advice from
@rmustacc and others.

### DIFF
--- a/drv/cosmo-seq-server/src/main.rs
+++ b/drv/cosmo-seq-server/src/main.rs
@@ -1045,8 +1045,6 @@ pub(crate) enum EreportClass {
     //
     #[cbor(rename = "hw.cpu.a0_fail.unknown")]
     UnrecognizedCPU,
-    #[cbor(rename = "hw.a0_fail")]
-    A0Failure,
 }
 
 pub(crate) type Ereport<T> = task_packrat_api::Ereport<EreportClass, T>;


drv/cosmo-seq-server/src/main.rs
```diff
@@ -959,6 +1026,80 @@ impl NotificationHandler for ServerImpl {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+#[derive(Eq, PartialEq, Copy, Clone, microcbor::Encode, counters::Count)]
+pub(crate) enum EreportClass {
+    //
+    // Interrupts
+    //
+    #[cbor(rename = "hw.cpu.thermtrip")]
```
rmustacc, line 1034: I've been going back and forth with myself on whether to phrase this as specific to amd or not. e.g. `hw.cpu.amd.thermtrip`, as we're referring to its thermal trip assertion pin. I think it's probably better to keep this as is.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -959,6 +1026,80 @@ impl NotificationHandler for ServerImpl {
...
+    Thermtrip,
+    #[cbor(rename = "hw.seq.smerr")]
```
rmustacc, line 1036: So, I don't think we should phrase this in terms of the sequencer as this is a fact about the CPU. So probably closer to `hw.cpu.smerr`. It's not clear to me again if we want to namespace this or if we'd try to make these similar. This almost feels like `hw.cpu.amd.smerr` (though I had to look up what smerr is).

drv/cosmo-seq-server/src/main.rs
```diff
@@ -959,6 +1026,80 @@ impl NotificationHandler for ServerImpl {
...
+    Smerr,
+    #[cbor(rename = "hw.seq.a0_mapo")]
```
rmustacc, line 1038: In this case I don't think we should phrase this as specific to the sequencer, this feels more related to power. So I would think of this as `hw.pwr.a0.mapo`. This kind of organization assumes we want to make a class of A0/A2-specific power events. I also think we should consider instead `hw.pwr.mapo` with a power domain present as an argument. A small part of me prefers this in case we detect A2 issues or we have a more complex situation with the Metro FPGA.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -959,6 +1026,80 @@ impl NotificationHandler for ServerImpl {
...
+    A0Mapo,
+    #[cbor(rename = "hw.pwr.pmbus.alert")]
+    PmbusAlert,
+
+    //
+    // Initialization failures
+    //
+    #[cbor(rename = "hw.cpu.a0_fail.unknown")]
```
rmustacc, line 1046: I get where you're coming from here, but I think I would basically keep this as `hw.cpu.unsup` or you know the old callback to `enotsup`. I think `unknown` here is a bit confusing and it's not clear to me that an explicit `a0_fail` subclass makes sense here.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -787,22 +822,54 @@ impl ServerImpl {
             self.seq.ifr.modify(|h| h.set_thermtrip(false));
             ringbuf_entry!(Trace::Thermtrip);
             action = InternalAction::ThermTrip;
-            // Great place for an ereport?
+            let ereport = Ereport {
+                class: EreportClass::Thermtrip,
+                version: 0,
+                report: &HOST_CPU_REFDES,
+                // TODO(eliza): eventually, it would be nice to include sequencer
```
rmustacc, line 829: What sequencer registers are you imagining for this?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -787,22 +822,54 @@ impl ServerImpl {
...
             self.seq.ifr.modify(|h| h.set_a0mapo(false));
             ringbuf_entry!(Trace::A0MapoInterrupt);
             action = InternalAction::Mapo;
-            // Great place for an ereport?
+
+            let ereport = Ereport {
+                class: EreportClass::A0Mapo,
+                version: 0,
+                report: &SEQ_FPGA_REFDES,
+                // TODO(eliza): eventually, it would be nice to include sequencer
```
rmustacc, line 848: I guess some of the sequencer registers may make sense, but I think the things we'll want to be following up with are the actual power failures that occurred.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -787,22 +822,54 @@ impl ServerImpl {
...
             self.seq.ifr.modify(|h| h.set_smerr_assert(false));
             ringbuf_entry!(Trace::SmerrInterrupt);
             action = InternalAction::Smerr;
-            // Great place for an ereport?
+
+            let ereport = Ereport {
+                class: EreportClass::Smerr,
+                version: 0,
+                report: &SEQ_FPGA_REFDES,
+                // TODO(eliza): eventually, it would be nice to include sequencer
```
rmustacc, line 866: What sequencer state is relevant here?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -959,6 +1026,80 @@ impl NotificationHandler for ServerImpl {
...
+pub(crate) struct UnrecognizedCPU {
+    #[cbor(flatten)]
+    refdes: &'static HostCpuRefdes,
+    coretype0: bool,
+    coretype1: bool,
+    coretype2: bool,
+    sp5r1: bool,
+    sp5r2: bool,
+    sp5r3: bool,
+    sp5r4: bool,
```
rmustacc, line 1062: So this is now scoped to a socket. I think we would to think about how the payloads change across families and how to name fields that way. SP3 for example only had a single coretype and two revision registers. I would speculate that SP7 will be closer to SP5, but it won't be called `sp5`.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -787,22 +822,54 @@ impl ServerImpl {
...
 
         if ifr.a0mapo {
             self.log_pg_registers();
             self.seq.ifr.modify(|h| h.set_a0mapo(false));
             ringbuf_entry!(Trace::A0MapoInterrupt);
             action = InternalAction::Mapo;
-            // Great place for an ereport?
+
+            let ereport = Ereport {
+                class: EreportClass::A0Mapo,
```
rmustacc, line 845: We should consider adding the SP's notion of when the state transition to A0 begun and the tick we detected the event in here as additional information to get a sense of relative uptime. The tick detection may not be relevant if that's part of the standard payload. This applies to most of the sequencer loop related actions. I wouldn't do this on an unsupported CPU.

drv/cosmo-seq-server/src/vcore.rs
```diff
@@ -377,26 +378,18 @@ impl VCore {
...
 
         let ereport = Ereport {
-            k: "hw.pwr.pmbus.alert",
-            v: 0,
-            rail,
-            refdes: device.i2c_device().component_id(),
-            time: now,
-            pmbus_status,
-            pwr_good: power_good,
+            class: crate::EreportClass::PmbusAlert,
```
rmustacc, line 381: Just confirming my read of this that the actual generated payload hasn't changed.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -202,42 +210,41 @@ struct ServerImpl<S: SpiServer> {
...
-    word: Option<u16>,
-    input: Option<u8>,
-    iout: Option<u8>,
-    vout: Option<u8>,
-    temp: Option<u8>,
-    cml: Option<u8>,
-    mfr: Option<u8>,
+    #[cbor(rename = "hw.cpu.a0_fail.unknown")]
+    UnrecognizedCPU,
+    #[cbor(rename = "hw.cpu.a0_fail.no_cpu")]
```
rmustacc, line 237: `hw.cpu.missing`? I think this is another place to not have `no_cpu`. Is there a reason this is in Gimlet and not Cosmo?

drv/gimlet-seq-server/src/main.rs
```diff
@@ -202,42 +210,41 @@ struct ServerImpl<S: SpiServer> {
...
+    NoCPUPresent,
+
+    #[cbor(rename = "hw.a0_fail.timeout.a1")]
```
rmustacc, line 240: For this set, this is less of an immediate, here's how this should be, but rather, I think we should consider the following:

* Here we're differentiating sequencing failures into subclasses, but we're not for Cosmo.
* I think in all of these I don't know that we want it to be A0 specific here, because if a Tofino or X2 fail to sequence they'll fail for other reasons.
* I wonder if we want a general `hw.cpu.seq.fail` that has payload specifics for us to figure out what makes sense across platforms. This would tentatively be paired with a `hw.cpu.seq.done` or maybe `hw.cpu.seq.begin`.

This makes me think that I'm thinking about this solely as CPU sequencing, when other stuff is happening in here. I dunno. I think I'll need to spend some time in the sequencer state here and I think we should talk with @nathanaelhuffman as to where we want the "public" / "committed" information to be.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -202,42 +210,41 @@ struct ServerImpl<S: SpiServer> {
...
+    Ereport<A0FailureEreport>,
+    Ereport<&'static HostCpuRefdes>,
+];
 
-#[derive(microcbor::Encode)]
+type Ereport<T> = task_packrat_api::Ereport<EreportClass, T>;
+
+#[derive(Copy, Clone, Eq, PartialEq, microcbor::Encode, counters::Count)]
 pub enum EreportClass {
+    #[cbor(rename = "hw.cpu.thermtrip")]
```
rmustacc, line 230: Seeing these in both Gimlet and Cosmo makes me wonder if there should be shared definitions here even if it has to get compiled into both tasks. That way we're consistent about the data that each contains.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -202,42 +210,41 @@ struct ServerImpl<S: SpiServer> {
...
+    #[cbor(rename = "hw.cpu.a0_fail.no_cpu")]
+    NoCPUPresent,
+
+    #[cbor(rename = "hw.a0_fail.timeout.a1")]
+    A1Timeout,
+    #[cbor(rename = "hw.a0_fail.timeout.a0")]
+    A0Timeout,
+    #[cbor(rename = "hw.a0_fail.timeout.groupc")]
+    A0TimeoutGroupC,
+    #[cbor(rename = "hw.pwr.pmbus.a0_fail.i2c_err")]
```
rmustacc, line 246: The class on here is weird to me, but I think we want to figure out the above first. It's not really a PMBus-specific happening here right? It's actually that we couldn't communicate over I2C based on my memory, though that might be off.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -1004,26 +1019,27 @@ impl<S: SpiServer> ServerImpl<S> {
...
         //
         ringbuf_entry!(Trace::A0Failed(err));
-        record_reg(Addr::IFR);
-        record_reg(Addr::DBG_MAX_A0SMSTATUS);
-        record_reg(Addr::MAX_GROUPB_PG);
-        record_reg(Addr::MAX_GROUPC_PG);
-        record_reg(Addr::FLT_A0_SMSTATUS);
-        record_reg(Addr::FLT_GROUPB_PG);
-        record_reg(Addr::FLT_GROUPC_PG);
+        let seq_status = SeqStatus {
```
rmustacc, line 1034: So this is interesting and suggests that the sequencing stuff may need to be board specific. This almost makes me want to think that this is `hw.gimlet.a0.seq.fail` or something. This needs more thought, but it's clear that we can't use the same payload between both. Or that we need some way for consumers to know that the board will change the payload. The reason I have the board here is that it's board-specific.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -1341,6 +1448,95 @@ fn read_spd_data_and_load_packrat(
...
+    UnrecognizedCPU(Coretype),
+    NoCPUPresent,
+    A1Timeout,
+    A0Timeout,
+    A0TimeoutGroupC,
+    I2cFault {
+        refdes: &'static str,
+        rail: &'static str,
+        #[count(children)]
+        err: i2c::ResponseCode,
```
rmustacc, line 1462: This feels like something that we need to think carefully about. I think there's a desire to change this around a bit right, but this is starting to bake it into a more public API.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -1341,6 +1448,95 @@ fn read_spd_data_and_load_packrat(
...
+    seq_status: SeqStatus,
+}
+
+#[derive(microcbor::EncodeFields)]
+enum A0FailureDetails {
+    UnrecognizedCPU {
+        #[cbor(flatten)]
+        refdes: &'static HostCpuRefdes,
+        #[cbor(flatten)]
+        coretype: Coretype,
```
rmustacc, line 1479: Why does this have coretype but not SP3r1/SP3r2?

drv/cosmo-seq-server/src/main.rs
hawkw: yeah, i like `unsup` --- i thought about `unknown_cpu` but that felt like a few too many characters. I agree that `unknown` feels like it could mean "unknown error", which is not what this is supposed to mean. if we dropped `a0_fail`, `unknown_cpu` seems not _too_ bad...

drv/cosmo-seq-server/src/vcore.rs
hawkw: correct, yeah

drv/gimlet-seq-server/src/main.rs
hawkw: `Coretype` here is a Rust struct that actually includes `coretype`/`sp3r1`/`sp3r2` fields; it's `flatten`ed so we would actually generate the CBOR:

```
"coretype": <bool>,
"sp3r1": <bool>,
"sp3r2": <bool>,
```

It's a separate struct because it gets passed around in a few places here. Naming it `Coretype` may have been a mistake.

PR #2245, opened 2025-09-23
stm32xx-gpio-common: fix another glitch

### LINKED ISSUE #1824: I2C is generating 600-750ns glitches, and it should feel bad about this.

cbiffle:
My original fix only made a difference when switching _into_ Alternate mode. In the relatively common case of switching _out of_ Alternate into e.g. Analog or Input (to tristate the pin), code usually uses a utility wrapper that passes AF0 to gpio_configure_raw... and even with the fix in place, we'd update AFRx to AF0 before changing MODER.

Which is the right thing to do if you're going _to_ Alternate, but not vice versa.

So, we got brief glitches to AF0 (which in practice on most pins means pulled to ground) when attempting to tristate pins.

It's possible that we need to revisit this API, which is arguably overly clever. But for now, this commit changes the logic so we don't update AFRx unless relevant.

Fixes #1824 (again).

### DIFF
--- a/drv/stm32xx-gpio-common/src/server.rs
+++ b/drv/stm32xx-gpio-common/src/server.rs
@@ -162,19 +162,33 @@ impl<T: GpioPeriph> AnyGpioPeriph for T {
             // See comment re: wrapping_mul above.
             w.bits((r.bits() & !mask_2) | pupdr_val.wrapping_mul(lsbs_2))
         });
-        // AFRx contains 8x 4-bit fields.
-        let af_val = u32::from((atts >> 7) & 0b1111);
-        self.afrl().modify(|r, w| unsafe {
-            // See comment re: wrapping_mul above.
-            w.bits((r.bits() & !mask_4l) | af_val.wrapping_mul(lsbs_4l))
-        });
-        self.afrh().modify(|r, w| unsafe {
-            // See comment re: wrapping_mul above.
-            w.bits((r.bits() & !mask_4h) | af_val.wrapping_mul(lsbs_4h))
-        });
+
+        let moder_val = u32::from(atts & 0b11);
+
+        // Only update the alternate fields mux if the mode will be Alternate
+        // when we're done with it. This avoids a glitch in resetting the mux if
+        // switching from Alternate _to_ something else. For the other two
+        // cases:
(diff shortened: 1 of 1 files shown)


drv/stm32xx-gpio-common/src/server.rs
```diff
@@ -162,19 +162,33 @@ impl<T: GpioPeriph> AnyGpioPeriph for T {
...
+        // Only update the alternate fields mux if the mode will be Alternate
+        // when we're done with it. This avoids a glitch in resetting the mux if
+        // switching from Alternate _to_ something else. For the other two
+        // cases:
+        // - Switching between Alternate settings is as glitch-free as is
+        //   feasible in the hardware (though you might want to consider going
+        //   through Analog anyway).
+        //
+        // - Switching _to_ Alternate from a different mode updates the mux
+        //   before MODER, so the update timing is immaterial.
```
hawkw, line 177: thanks for including this!

cbiffle: Every time I have to re-fix this bug, the comments will get longer. For better or worse.

PR #2246, opened 2025-09-24
microcbor: derivable CBOR encoding with statically-known max length

### LINKED ISSUE #2241: want some kind of build-time validation that ereports fit in encoding buffers
When a task delivers an ereport to packrat, the CBOR is encoded into a buffer in the reporting task which is then leased to Packrat. Since Hubris doesn't use dynamic allocation, these buffers must be fixed-size. This means that there is a possibility for an ereport to be too big to ever fit in the buffer size that was chosen by the programmer. That would be bad.

Presently, I've been dealing with this by making code that produces an ereport record whether or not it fit in the buffer in a ringbuf, along with the encoded length. For example:
https://github.com/oxidecomputer/hubris/blob/43e13b955df9c2bf959074aa5e75f634a435a0de/drv/gimlet-seq-server/src/vcore.rs#L278-L292
Then, I'll trigger the condition that produces the ereport and check that it was recorded successfully, and adjust the size of the buffer if it's too big. This "sort of works" for now, but it's not an approach that will scale as we add more ereports to things, and as we add more data to existing ereports. And, it only works when the condition that generates an ereport is something we can reliably and safely reproduce in the lab, which may not always be the case.[^1] It would be much better to have a way to ensure that these buffers are right-sized at build time.

For hubpack-encoded messages, this would be easy. Since Hubpack doesn't support variable-length fields, it has a `#[derive(SerializedSize)]` that can statically determine the maximum size of a message, and then the buffer could just be sized to whatever the max size of all the messages it will be used for. This doesn't work for ereports, which are encoded in CBOR,[^1] a format that permits variable-size fields. 

Looking at the current ereport messages, though, the only variable-size fields are `&'static str`s, for the ereport class string and stuff like a component's refdes or a power rail. For example:
https://github.com/oxidecomputer/hubris/blob/43e13b955df9c2bf959074aa5e75f634a435a0de/drv/gimlet-seq-server/src/vcore.rs#L269-L277
Generally there is a relatively small set of such values that we actually use, rather than dynamically formatted arbitrary strings, but we don't currently have a way of showing that at compile-time. I don't think some kind of proc-macro for determining this at compile-time will be easy, though, at least without some way to say "this is actually coming from a fixed set of strings" (perhaps an enum with some more attributes, like `serde(rename = "...")`?).

Alternatively, we could do this using unit tests that just construct the ereports and assert their serialized form fits in the buffer. This would be the easy, pragmatic approach...except for one wrinkle: today, the structs defining ereport messages are defined in task code. We can't easily include tests that are intended to run on a development machine in a task crate, since the task code usually doesn't compile at all for Big Computer targets like `x86_64-unknown-linux-gnu` or `aarch64-apple-darwin`. We could work around that by having separate crate(s) for the ereport messages that can compile for `#![no_std]` MCU targets but also run tests on dev machine targets, but that's kind of a pain. Perhaps there's some build system hack that could make this easier...

I would love other suggestions if anyone can think of something I'm overlooking here!

[^1]: E.g. if we want a "computer is on fire" ereport, we might have to set the computer on fire just to check that it fits in the buffer. :)
[^2]: At present, always using `minicbor-serde`.

hawkw:
When encoding ereports in Hubris, the CBOR messages are generally
simple and consist of fixed-size data. However, if buffer sizes for 
encoding those messages are just chosen arbitrarily by the programmer,
it is possible that subsequent changes to the ereport messages (such as
adding a new field...) will increase the encoded size beyond the chosen
buffer size, leading to encoding failures and data loss.

Therefore, this branch adds a new crate that provides traits and derive
macros for generating CBOR encoding implementations that determine the
maximum needed buffer size at compile time, using `minicbor`. While
this was primarily intended for use in the ereport system, it's fairly
general purpose, so I've named the new crate `microcbor` --- since
it's implemented using `minicbor`, and is designed for use on
microcontrollers. 

I've also added a crate for representing strings as fixed-size arrays.
This is necessary since all field values in types deriving
`microcbor::Encode` must have a known maximum length at compile time,
which isn't the case for strings --- even `&'static str` constants.
This looks superficially similar to `heapless::String`, but it's
actually quite different: the primary purpose of this thing is to be
able to be constructable from a `&'static str` in a `const fn`,
failing if it's too long to fit in the max size buffer.
`heapless::String`, on the other hand, is designed to behave more like
the `alloc::string::String` type and allow the string to be
mutated/appended to at runtime. But, only _empty_ `heapless::String`s
can be `const fn` constructed. Also, my thing is backed by an array,
so we could potentially use it elsewhere, such as in `hubpack` or
`zerocopy`-encoded messages. Since this seemed generally usefulish
outside of the CBOR stuff, I stuck it in its own `fixedstr` crate.

The new libraries have a bunch of documentation explaining how they
work and how to use them, so I won't duplicate it all in the commit
message. Additionally, as a sort of trial balloon for the new stuff,
I've modified the ereports from `drv_gimlet_seq_server` to use it.
I'll update other tasks subsequently, but I'd like to make a
follow-up change to have `build-i2c` just generate `FixedStr`s for
component IDs instead, and I figured it was better to keep this diff
smaller for now.

Fixes #2241

### DIFF
--- a/task/packrat-api/Cargo.toml
+++ b/task/packrat-api/Cargo.toml
@@ -4,8 +4,29 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
+#
+# The `serde` and `microcbor` feature flags enable the presence of helper
+# APIs for reporting ereports encoded using either `serde` or `microcbor`'s
+# serialization traits.
+#
+# Both of these APIs will ultimately encode the ereport using `minicbor`, but
+# using different derived encoding implementations on top of `minicbor`.
+# The `microcbor`-based API is recommended for use in new code, as it allows
+# the automatic calculation of encoding buffer sizes. See the `microcbor`
+# crate's documentation for more details.
+#
+
 # Enable Serde support for the ereport API.
+#
+# When deriving `serde::Serialize` for ereport message types, they may be
+# delivered to packrat using the `Packrat::serialize_ereport` method, if this
+# `serde` feature flag is enabled.
 serde = ["dep:serde", "dep:minicbor", "dep:minicbor-serde"]
+# Enable microcbor support for the ereport API.
+#
+# When deriving `microcbor::Encode` for ereport message types, they may be
+# delivered to packrat using the `Packrat::encode_ereport` method, if this
+# feature flag is enabled.
 microcbor = ["dep:microcbor", "dep:static_assertions", "dep:minicbor"]
 
 [dependencies]


hawkw:
Okay, so the current approach as of e2e510621801ec7635026a3a915d57fad020288a seems to be more or less functional. But, there's one big flaw: it turns out there isn't really a way to take something like the current trait:
https://github.com/oxidecomputer/hubris/blob/13de2cf7564527fcbb37e404be1874a0679eb668/lib/ereport/src/lib.rs#L12-L18
and write a function like
```rust
pub fn encode_ereport<'buf, E: EreportData>(
    ereport: &E,
    buf: &'buf mut [u8; E::MAX_CBOR_LEN],
) -> &'buf [u8] {
    let mut encoder = Encoder::new(buf);
    todo!()
}
```
because this runs afoul of 
```
error: generic parameters may not be used in const operations
  --> lib/ereport/src/lib.rs:22:25
   |
22 |     buf: &'buf mut [u8; E::MAX_CBOR_LEN],
   |                         ^^^^^^^^^^^^^^^ cannot perform const operation using `E`
   |
   = note: type parameters may not be used in const expressions
   = help: add `#![feature(generic_const_exprs)]` to allow generic const expressions
```

This feels pretty unfortunate, as a function like this is basically the UX I was hoping to be able to have. Currently, the user can get the type to tell it what the right length for the buffer is, and use it to construct a buffer, but we can't easily have an API that ensures you _use_ the correct-length buffer when encoding, which is too bad.

We could just enable the nightly-only `#![feature(generic_const_exprs)]` and have it work, but I think we're trying to avoid adding new nightly features.

Another option I'm thinking about is changing the trait to something like:

```rust
pub trait EreportData: Encode<()> {
    /// The maximum length of the CBOR-encoded representation of this value.
    ///
    /// The value is free to encode fewer than this many bytes, but may not
    /// encode more.
    const MAX_CBOR_LEN: usize;
    type NeededBuf: AsRef<[u8]> + AsMut<[u8]>;
}
```
and having the derive macro generate a `NeededBuf` associated type that's `[u8; MAX_CBOR_LEN]`. We could then have an `encode` function that's like

```rust
pub fn encode_ereport<'buf, E: EreportData>(
    ereport: &E,
    buf: &'buf mut E::NeededBuf,
) -> &'buf [u8] {
   // ...
}
```
and you wouldn't be able to call it with a too-short array. But, this might be overthinking it...

hawkw:
**Update**: nope, the approach proposed in https://github.com/oxidecomputer/hubris/pull/2246#issuecomment-3340127700 _still_ ends up using a generic parameter in a const expression if the type is generic over another type implementing `EreportData`. So, we really just can't quite do that without `#![feature(generic_const_exprs)]`. I think the current state of this is still useful as it at least gives us a way to "manually" construct a correct-length buffer. You can choose to misuse the derived implementation with a wrong-length buffer, but...you could do this before, too. It would be nice to have a misuse-resistant API, but I can accept just having an API that does the Right Thing if you _don't_ misuse it!

hawkw:
Some notes from the field after starting to update #2242 to use this:

- It would be nice if an enum with unit variants could still `#[derive(microcbor::EncodeFields)]` and just skip the unit variants entirely (perhaps only if explicitly annotated). This way, if you wanted to add one of a set of additional fields, or nothing, to a parent type, you can have something like
  ```rust
  #[derive(microcbor::EncodeFields)]
  enum AdditionalFields {
      Foo { a: ..., b: ... },
      Bar { c: ..., d: ... },
      Nothing,
  }
  ```
  Alternatively there could just be an `EncodeFields` implementation for `Option<T: EncodeFields>` that just skips it if it's `None`.
- It would be nice to have something like serde's "internally tagged" enum repr, so that an enum can flatten its fields into a parent struct _and_ add a string tag field, like:
  ```rust
  #[derive(microcbor::EncodeFields)]
  #[cbor(tag = "k")]
  enum EreportSubclass {
      #[cbor(rename = "hw.foo.bar")]
      FooBar { a: usize, b: bool },
      #[cbor(rename = "hw.foo.baz")]
      FooBaz { c: u32, d: Option<u8> },
  }
  ```
  If the enum _also_ derived `Encode`, it would generate the same thing, but as a nested map of the tag and all the other stuff.

Not sure if I'm going to add these in this PR or in a follow-up.

hawkw:
Okay, 851f5d6be0485abdfa78576503494c6fb4e1e5c0 adds the "tagged" enum representation I mentioned in https://github.com/oxidecomputer/hubris/pull/2246#issuecomment-3357670399. And then a38bb49c7dca6973e38b9a0a3ca67ce8c2cb283b _immediately renames it_ to the much more unpleasant "variant ID", since "tag" means a different thing in CBOR.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -208,7 +208,38 @@ struct ServerImpl<S: SpiServer> {
 }
 
 const TIMER_INTERVAL: u32 = 10;
-const EREPORT_BUF_LEN: usize = 256;
+const EREPORT_BUF_LEN: usize = <task_packrat_api::Ereport<
+    EreportClass,
+    EreportKind,
+> as microcbor::StaticCborLen>::MAX_CBOR_LEN;
```
mkeeter:
```suggestion
const EREPORT_BUF_LEN: usize = microcbor::max_cbor_len_for!(
    task_packrat_api::Ereport<EreportClass, EreportKind>
);
```
?

drv/gimlet-seq-server/src/vcore.rs
```diff
@@ -266,27 +273,27 @@ impl VCore {
...
             Ok(len) => ringbuf_entry!(Trace::EreportSent(len)),
-            Err(task_packrat_api::EreportSerializeError::Packrat {
-                len,
-                err,
-            }) => {
+            Err(task_packrat_api::EreportEncodeError::Packrat { len, err }) => {
                 ringbuf_entry!(Trace::EreportLost(len, err))
             }
-            Err(task_packrat_api::EreportSerializeError::Serialize(_)) => {
+            Err(task_packrat_api::EreportEncodeError::Encoder(_)) => {
```
mkeeter, line 296: Is `EreportTooBig` still the right thing to log here?

lib/fixedstr/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        let mut idx = 0;
+        while idx < len {
+            buf[idx] = bytes[idx];
+            idx += 1;
+        }
+        Ok(Self { buf, len })
+    }
+
+    pub const fn from_str(s: &str) -> Self {
+        match Self::try_from_str(s) {
```
mkeeter, line 67: TIL that `unwrap` is a `const fn` for `Option` but not for `Result` 🙃

lib/fixedstr/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    pub const fn try_from_utf8(bytes: &[u8]) -> Result<Self, FromUtf8Error> {
+        let s = match core::str::from_utf8(bytes) {
+            Ok(s) => s,
+            Err(e) => return Err(FromUtf8Error::InvalidUtf8(e)),
+        };
+        match Self::try_from_str(s) {
+            Ok(s) => Ok(s),
+            Err(_) => Err(FromUtf8Error::TooLong),
```
mkeeter: I'd vote to be explicit here:
```suggestion
            Err(StringTooLong) => Err(FromUtf8Error::TooLong),
```

lib/fixedstr/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        struct ExpectedLen(usize);
+        impl serde::de::Expected for ExpectedLen {
+            fn fmt(
+                &self,
+                f: &mut core::fmt::Formatter<'_>,
+            ) -> core::fmt::Result {
+                write!(f, "a string of length {}", self.0)
```
mkeeter:
```suggestion
                write!(f, "a string of length <= {} bytes", self.0)
```

lib/fixedstr/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+        let s = <&'de str>::deserialize(deserializer)?;
+        Self::try_from_str(s).map_err(|_| {
```
mkeeter:
```suggestion
        Self::try_from_str(s).map_err(|_: StringTooLong| {
```
Obviously not necessary for type inference, but helpful to the reader (and would break if that error type became more complex in the future)

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+/// **Note**: We use the (somewhat unfortunate) terminology "variant ID" rather
+/// than "tag", because in CBOR, the term "tag" refers to a [completely
+/// different thing][cbor-tags].
+///
+/// If the `#[cbor(variant_id = "id_field_name")]` attribute is present, any
+/// variant of the enum will additionally encode a key-value pair where the
+/// key is the provided variant ID field name, and the value is the variant's
+/// name (or the value of a `#[cbor(rename = "...")]` attribute on that
+/// variant if one is present).
+///
```
mkeeter, line 63: Explain what happens if it's not present?  I guess fields within the variant are just plopped into the CBOR object with indication of variant?

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+///   representation](#enum-variant-id-encoding) with the specified field name
+///   name when encoding this enum. Note that this attribute may *not* be used
+///   on enums which have tuple-like (unnamed fields) variants.
+///
+/// ## Field Attributes
+///
+/// The following `#[cbor(..)]` attributes are supported on fields of structs
+/// and enum variants:
+///
+/// - `#[cbor(skip)]`: Completely skip ignoring this field. If a field is
```
mkeeter: Delete "ignoring" here?

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+///   skipped, it will not be included in the encoded CBOR output.
+///
+/// - `#[cbor(skip_if_nil)]`: Skip encoding this field if it would encode a
+///   CBOR `nil` value.
+///
+///   This attribute will cause the generated `Encode` implementation to call
+///   the value's `Encode::is_nil` method to determine if the field would emit
+///   a `nil` value. If it returns `true`, the field will no tbe encoded at
```
mkeeter:
```suggestion
///   a `nil` value. If it returns `true`, the field will not be encoded at
```

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+///   This attribute may only be placed on fields which are of types that
+///   implement the [`EncodeFields`] trait. [`EncodeFields`] may be derived
+///   for any struct or enum type which has named fields.
+///
+///   Only structs and enum variants whose fields are named may use the
+///   `#[cbor(flatten)]` attribute on their fields. Using `#[cbor(flatten)]`
+///   on fields of a tuple struct or tuple-like enum variant will result in a
+///   compile error. An enum type which has both struct-like and tuple-like
+///   variants *may* use `#[cbor(flatten)]`, but only within its struct-like
+///  variants.
```
mkeeter:
```suggestion
///   variants.
```

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+            input.attrs,
+            input.vis,
+            input.ident,
+            input.generics,
+            data.clone(),
+        )
+        .map(|tokens| tokens.to_token_stream().into()),
+        _ => Err(syn::Error::new_spanned(
+            input,
+            "`StaticCborLen` can only be derived for `struct` and `enum` \
```
mkeeter, line 294:
```suggestion
            "`Encode` can only be derived for `struct` and `enum` \
```

It looks like that it derives both, but someone typing `#[derive(Encode)]` would be confused to get an error message about `StaticCborLen`

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+    }
+}
+
+const HELPER_ATTR: &str = "cbor";
+const RENAME_ATTR: &str = "rename";
+const VARIANT_ID_ATTR: &str = "variant_id";
+
+fn gen_enum_encode_impl(
+    attrs: Vec<Attribute>,
+    _vis: Visibility,
```
mkeeter: Is there a reason to keep the dead arguments in this and other `gen_` functions?

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+                            __microcbor_renamed_encoder
+                                .map(1)?
+                                .str(#field_name)?
+                                .str(#name)?;
+                        }
+                    });
+                    variant_lens.push(quote! {
+                        #[allow(non_snake_case)]
+                        let #variant_name = {
+                            // this will encode exactly 1 field, so we use the
```
mkeeter: Maybe this comment should live in the `variant_patterns` section above?

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+                }
+            },
+            syn::Fields::Named(ref fields) => {
+                let mut field_gen =
+                    FieldGenerator::for_variant(FieldType::Named);
+                for field in &fields.named {
+                    field_gen.add_field(field)?;
+                }
+                if let Some(ref field_name) = variant_id_field_name {
+                    field_gen.add_variant_id_field(field_name, &name);
```
mkeeter: Do we detect variant IDs that collide with field names?

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+                                .str(#field_name)?
+                                .str(#name)?;
+                        }
+                    });
+                    variant_lens.push(quote! {
+                        #[allow(non_snake_case)]
+                        let #variant_name = {
+                            // this will encode exactly 1 field, so we use the
+                            // length-prefixed repr to save a byte.
+                            let mut len = ::microcbor::u64_cbor_len(1);
```
mkeeter: I would just write this as `1` with a comment saying it's a major type map + length 1, which is always encoded into a single byte (we're never actually encoding a `u64`).

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+                match self {
+                    #(#variant_patterns,)*
+                }
+                Ok(())
+            }
+        }
+    })
+}
+
+fn gen_encode_fields_impl(
```
mkeeter: Scoot this below `gen_encode_struct_impl`, to avoid breaking up the `gen_encode_impl` flow?

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+        &mut self,
+        field_name: &LitStr,
+        variant_id: &LitStr,
+    ) {
+        self.field_len_exprs.push(quote! {
+            len += ::microcbor::str_cbor_len(#field_name)
+        });
+        self.field_len_exprs.push(quote! {
+            len += ::microcbor::str_cbor_len(#variant_id)
+        });
```
mkeeter, line 885: Does this need to be two pushes?
```suggestion
        self.field_len_exprs.push(quote! {
            len += ::microcbor::str_cbor_len(#field_name)
            len += ::microcbor::str_cbor_len(#variant_id)
        });
```

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+            field_patterns: Vec::new(),
+            field_len_exprs: Vec::new(),
+            field_encode_exprs: Vec::new(),
+            where_bounds: Vec::new(),
+            any_skipped: false,
+            field_type,
+        }
+    }
+
+    fn for_variant(field_type: FieldType) -> Self {
```
mkeeter: This looks the same as `FieldGenerator::for_struct`, consolidate to `FieldGenerator::new`?

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+                }
+                (FieldType::Named, true) => {
+                    let field_ident = field.ident.as_ref().unwrap();
+                    self.field_patterns.push(quote! { #field_ident: _ });
+                    return Ok(());
+                }
+                (FieldType::Named, false) => {
+                    let field_ident = field.ident.as_ref().expect(
+                        "if we are generating named fields, there should \
+                             be an ident for each field",
```
mkeeter: Indentation:
```suggestion
                         be an ident for each field",
```

lib/microcbor-derive/src/lib.rs
```diff
@@ -0,0 +1,1030 @@
...
+                quote! {
+                    #encode_name
+                    ::microcbor::Encode::<()>::encode(
+                        #field_ident,
+                        __microcbor_renamed_encoder,
+                        __microcbor_renamed_ctx,
+                    )?;
+                }
+            });
+            self.where_bounds.push(quote! {
```
mkeeter: Scoot this to the top of the `} else {`, for symmetry with the first branch of the conditional

lib/microcbor/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+//! This crate provides the [`StaticCborLen`] trait for types that can be
+//! encoded as CBOR with a known maximum encoded length. In addition, it
+//! provides [`#[derive(Encode)`](macro@Encode) and
+//! [`#[derive(EncodeFields)`](macro@EncodeFields) derive attributes for
+//! deriving implementations of the [`Encode`] and [`StaticCborLen`] traits.
+//!
+//! ## Wait, Why Not `#[derive(serde::Serialize)]`?
+//!
+//! Well, the obvious one is that there's no way to know how many bytes a given
+//! type's`Serialize` implementation will produce at compile-time.
```
mkeeter:
```suggestion
//! type's `Serialize` implementation will produce at compile-time.
```

lib/microcbor/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    // https://docs.rs/minicbor/2.1.1/src/minicbor/encode.rs.html#539-546
+    u64 = 9,
+
+    // https://docs.rs/minicbor/2.1.1/src/minicbor/encode.rs.html#580
+    f32 = 5,
+
+    // https://docs.rs/minicbor/2.1.1/src/minicbor/encode.rs.html#586
+    f64 = 9,
+
+    //https://docs.rs/minicbor/2.1.1/src/minicbor/encode.rs.html#501
```
mkeeter:
```suggestion
    // https://docs.rs/minicbor/2.1.1/src/minicbor/encode.rs.html#501
```

lib/microcbor/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    fn encode_fields<W: Write>(
+        &self,
+        e: &mut Encoder<W>,
+        _: &mut C,
+    ) -> Result<(), encode::Error<W::Error>>;
+}
+
+#[cfg(feature = "fixedstr")]
+impl<const LEN: usize> StaticCborLen for fixedstr::FixedStr<LEN> {
```
mkeeter: Why put this in `microcbor`, instead of having an optional `microcbor` dependency in `fixedstr` (like it already has for `minicbor` and `serde`)?

drv/gimlet-seq-server/src/main.rs
hawkw: ah, yup, no reason not to use that here!

lib/fixedstr/src/lib.rs
hawkw: yeah, because unwrapping a `Result` will invoke the error type's `Debug` implementation, which is not `const fn`. isn't that cool?

drv/gimlet-seq-server/src/vcore.rs
hawkw: Yes, that's the only reason encoding would fail, since none of the `Encode` implementations will return an error under any other circumstances. While using the `StaticCborLen` trait/`max_cbor_len_for!` macro helps us to ensure the buffer is the right length, it's [still possible to see encoding fail because of a wrong-sized buffer](https://github.com/oxidecomputer/hubris/pull/2246#issuecomment-3348170342) if someone e.g. forgets to add an encoded type to the `max_cbor_len_for!` invocation, so the error is still possible. That would be a programmer error, so arguably, the task should panic, but I felt weird about making that take down the whole sequencer.

lib/microcbor-derive/src/lib.rs
hawkw: yeah, thanks

lib/microcbor-derive/src/lib.rs
hawkw: No, but perhaps we should; I don't know if CBOR itself permits multiple map entries with the same key, but I'm sure whoever's on the receiving end may get confused when decoding something like that.

I will note that we can't easily check that for the `EncodeFields` version of this, though, you might flatten something with a variant ID field into an enclosing map that also has a field with the same name. There isn't really a nice way for the macro to detect that case, sadly.

lib/microcbor-derive/src/lib.rs
hawkw: Yes, because when we expand the length expressions, we do `#(#field_len_exprs;)`, so the `;` is coming from the expansion. I don't think there's any cases where we don't want a `;`, though, because it's always a `+=` expression, whose type is already unit. So I think this could be easily changed to just put the `;` in the fragment and do one push here.

lib/microcbor/src/lib.rs
hawkw: haha, this is purely a historical artifact; originally `FixedStr` was defined in this crate, and then I decided it ought to be its own crate. So, the impl was here for that reason and not because this felt like the best factoring. I think putting the impl in `fixedstr` instead seems like a better factoring, will change it

lib/microcbor-derive/src/lib.rs
hawkw: yeah, that's also a historical artifact, i think --- at one point this constructed it with different values. will fix.

lib/microcbor-derive/src/lib.rs
hawkw: changed it in 38767cdb0b25415aa2448f40ac991eec484ca435!

hawkw:
@mkeeter i think i've addressed all your feedback! hoping to get a look from @cbiffle as well before merging.

lib/microcbor-derive/src/lib.rs
hawkw: done in https://github.com/oxidecomputer/hubris/pull/2246/commits/cef85a4d5fb117410b6e5f3532ebd57678a29cf7

drv/gimlet-seq-server/src/main.rs
```diff
@@ -208,7 +208,37 @@ struct ServerImpl<S: SpiServer> {
...
+pub enum EreportClass {
+    #[cbor(rename = "hw.pwr.pmbus.alert")]
+    PmbusAlert,
+}
+
+#[derive(microcbor::EncodeFields)]
+pub(crate) enum EreportKind {
+    PmbusAlert {
+        refdes: fixedstr::FixedStr<{ crate::i2c_config::MAX_COMPONENT_ID_LEN }>,
+        rail: &'static fixedstr::FixedStr<10>,
```
labbott: where does the `10` come from?

lib/microcbor/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+//! deriving implementations of the [`Encode`] and [`StaticCborLen`] traits.
+//!
+//! ## Wait, Why Not `#[derive(serde::Serialize)]`?
+//!
+//! Well, the obvious one is that there's no way to know how many bytes a given
+//! type's `Serialize` implementation will produce at compile-time.
+//!
+//! Another limitation, though, is that `serde`'s `#[serde(flatten)]` attribute
+//! requires `liballoc`, as the flattened fields are temporarily stored on the
+//! heap while encoding. This means that `#[serde(flatten)]` cannot be used to
```
labbott, line 33: TIL this fun fact about `serde(flatten)`

task/packrat-api/Cargo.toml
```diff
@@ -6,13 +6,16 @@ edition = "2021"
 [features]
 # Enable Serde support for the ereport API.
 serde = ["dep:serde", "dep:minicbor", "dep:minicbor-serde"]
+microcbor = ["dep:microcbor", "dep:static_assertions", "dep:minicbor"]
+default = ["microcbor"]
```
labbott: The correct set of features to select here is non-obvious to me . we have some tasks that have `features = ["serde"]` which presumably want ereports but we're also automatically selecting the `microcbor`. It's not clear when I would/would not select these features.

drv/gimlet-seq-server/src/main.rs
hawkw: 10 comes from manually counting the characters in the rail names currently used in this task. I probably should comment on that.

task/packrat-api/Cargo.toml
hawkw: I changed this so that the `microcbor` feature is no longer on by default. The features control the presence of APIs for delivering ereports using either `serde` or `microcbor`'s traits, as appropriate. I added some comments describing this in dbcc218e7556c2965d5b9252922fe621df32c3b7.

drv/gimlet-seq-server/src/main.rs
hawkw: Documented this in 0e7c48c6d4697f74511e493dd2e5c46f250158a3. Also, it turns out it only needs to be 9 bytes currently, whoops!

lib/microcbor/src/lib.rs
hawkw: Yeah, I was unpleasantly surprised to discover that...

PR #2248, opened 2025-09-24
Update sparse registry hash for new toolchain

mkeeter:
When we bumped the toolchain, this changed as well.

Keeping the hash correct reduces the length of panic strings.


hawkw:
It could be useful to write down someplace --- maybe in the `rust-toolchain.toml` file? --- that this has to be done when updating the toolchain. We can do that separately though.

mkeeter:
Good idea, pushed.

PR #2250, opened 2025-09-25
ADM1273 support

### LINKED ISSUE #2247: ADM1273 transition
The ADM1273 is a drop-in replacement from Analog for the ADM1272 (which has been EOL'd). This largely entails a renaming of ADM1272 things into something like ADM127x. Additionally, `Validate` for the `Adm127x` device driver should handle both models. The [pmbus crate](https://github.com/oxidecomputer/pmbus/blob/master/src/adm1272.ron) should be updated as well.

Aaron-Hartwig:
This change is largely a renaming of ADM1272 to ADM127x since the ADM1273 is a drop-in replacement and has the same behavior. The only thing of substance that was modified was how we Validate this device. The current behavior is the read the MFG_MODEL register and expect it to be `ADM1272-2A`. I think this is a bit fragile as the die revision is tacked on there (`-2A`) and could change without us knowing (the orderable PN here is actually still `ADM1272-1A` even though `ADM1272-2A` is being delivered). Then there is the additional complication that with the ADM1273 we would see something like `ADM1273-$A`. Thus, I've changed the behavior of the `adm127x::validate` to read the MFG_ID register, which should have a result of `ADI` for both parts.

fixes #2247


Aaron-Hartwig:
This PR relies on https://github.com/oxidecomputer/pmbus/pull/24 landing first.

mkeeter:
I'd vote to keep reading the model register, but do a `starts_with` check for either `ADM1272` or `ADM1273`.

Aaron-Hartwig:
Tested on my Sidecar B.

Original:
```
aaron@astoria ~ $ pfexec humility validate | grep adm
humility: attached to 483:3754 via ST-Link V3
 0 validated    1 B1 -   0x10 adm1272       Fan 1 hot swap controller
 5 validated    1 B2 -   0x13 adm1272       Fan 0 hot swap controller
 8 validated    3 C  -   0x16 adm1272       54V hot swap controller
14 validated    3 H  -   0x13 adm1272       Fan 2 hot swap controller
15 validated    3 H  -   0x10 adm1272       Fan 3 hot swap controller
```

This branch:
```
aaron@astoria ~ $ pfexec humility validate | grep adm
humility: attached to 483:3754 via ST-Link V3
 0 validated    1 B1 -   0x10 adm127x       Fan 1 hot swap controller
 5 validated    1 B2 -   0x13 adm127x       Fan 0 hot swap controller
 8 validated    3 C  -   0x16 adm127x       54V hot swap controller
14 validated    3 H  -   0x13 adm127x       Fan 2 hot swap controller
15 validated    3 H  -   0x10 adm127x       Fan 3 hot swap controller
```

mkeeter (approved):

PR #2253, opened 2025-09-25
host-sp-messages: don't copy that floppy^H^H^H^H^H^HInventoryData

hawkw:
The `InventoryData` enum in `host-sp-messages` is *real* big: at least
>512B for the `DimmSpd` message, or perhaps more if one of the other
variants whose fields I didn't want to add up is more than that. Currently, it `#[derive(Copy)]`s, which feels like a big stack overflow footgun: if an `InventoryData` value is accidentally moved by value, it gets copied on the stack. This kills the Hubris.

Thus, this commit just removes the `#[derive(Copy)]`, which should take the bullets out of the footgun. Other `host-sp-messages` are left alone, and everything still builds fine. While the `SpToHost` and `HostToSp` messages also derive `Copy`, they aren't as big, and the `Copy` impl is necessary because they get stuck in ringbufs.


mkeeter (approved):
Maybe add a comment saying "Copy is deliberately not implemented on this type because it's very large, and incidental copies can blow up the stack"?

hawkw:
> Maybe add a comment saying "Copy is deliberately not implemented on this type because it's very large, and incidental copies can blow up the stack"?

i'll do that once #2254 is merged and it's possible to switch branches in my editor again :)

Issue #2255, opened 2025-09-29
Issues reading revision ID resistors on Gimlet


mkeeter:
We have seen several Gimlet sleds experiencing confusion when reading their `SEQ_REV_ID` pins.

This typically manifests as a failure to boot Hubris, because the kernel checks for the expected pin values early in boot:
```
rm@cadbury ~ $ pfexec humility -t c71 tasks
humility: attached to 0483:3754:001100184741500820383733 via ST-Link V3
humility tasks failed: kernel has panicked on boot: "panicked at app/gimlet/src/main.rs:128:5:\nassertion `left == right` failed\n  left: 6\n right: 2"
```

Hubris reads these pins by configuring pull-up resistors on PG0-2, waiting in a busy loop (giving the pins time to charge), then reading their values:
https://github.com/oxidecomputer/hubris/blob/560c66f4aeaca7f850a7439f9945f889260dcf75/app/gimlet/src/main.rs#L72-L128

We've seen at least three failures:

- 0XV1:9130000019:006:BRM42220023: https://github.com/oxidecomputer/meta/issues/583#issuecomment-3287018886 (which also has other hardware issues)
- The Gimlet in Eric's basement is hardware rev 11, which should be a rev D board.  It reads either `0b011` or `0b111`, seemingly at random
- `c71` in the lab (above)

Once or twice could be a coincidence or bad hardware, but three times is starting to raise red flags.

I tried increasing the delay loop by a factor of 50, and still saw intermittent misreads on Eric's sled.  Looking at the decompiled kernel code, everything looks reasonable:
```
 r0 is 0x58021800, PORTG base address
 r2 is 0b010101

 80004cc:   f3bf 8f4f   dsb sy
 80004d0:   6801        ldr r1, [r0, #0]                // Load MODER
 80004d2:   f021 013f   bic.w   r1, r1, #63 ; 0x3f      // Clear lower 6 bits, setting input mode
 80004d6:   6001        str r1, [r0, #0]                // write back to MODER
 80004d8:   68c1        ldr r1, [r0, #12]               // load PUPDR
 80004da:   f362 0105   bfi r1, r2, #0, #6              // Enable pull-ups
 80004de:   60c1        str r1, [r0, #12]               // Write PUPDR
 80004e0:   f641 6147   movw    r1, #7751   ; 0x1e47    // delay (155 * 100) / 2 + 1
 80004e4:   3901        subs    r1, #1                  // delay decrement
[...]

PR #2258, opened 2025-09-30
Use FPGA's pull-up resistors for Gimlet revision ID pins

mkeeter:
Fixes https://github.com/oxidecomputer/hubris/issues/2255

Right now, the `SEQ_REV_ID{0,1,2}` pins have physical pull-downs for `0` bits, and use internal pull-up resistors in the SP so that unconnected pins go to `1`.  The pull-downs are 10KΩ in revision prior to rev F.

Unfortunately, the sequencer FPGA shares those revision ID lines, and has weak pull-ups when unconfigured.  Having both the SP and FPGA pulling up those pins can put their voltage at a borderline level.  There's a weirder problem, too: given precise timing, enabling the SP's internal pullups can cause the FPGA IO drivers to latch up to about 2.8V.

This PR fixes the issue by **only** relying on the FPGA's pull-up resistors.  It has a few moving pieces:

- The SP must wait for the FPGA's power good lines to go high before doing anything
- The SP must **deconfigure** the FPGA, to drop those pins into the unconfigured (weak pull-up) state
  - Our previous FPGA image configures those pins as inputs (with no pull-ups), so otherwise this would fail on a warm reset
  - In this PR, we also include an FPGA image which enables pull-ups for those pins, but that will only help on **subsequent** updates
- The SP **must not** enable its own pull-ups (to avoid pulling the voltage too high)


mkeeter:
@Aaron-Hartwig we should switch to a proper Quartz build and update the "built from X revision" link in the README

app/gimlet/src/main.rs
```diff
@@ -68,47 +68,78 @@ fn system_init() {
...
+    // Make PC6 (SEQ_REG_TO_SP_V3P3_PG) and PC7 (SEQ_REG_TO_SP_V1P2_PG) inputs,
+    // then wait for both of them to go high
+    #[rustfmt::skip]
+    p.GPIOC.moder.modify(|_, w| w
+        .moder6().input()
+        .moder7().input());
+    const SEQ_PG: u32 = 0b11 << 6;
+    while p.GPIOC.idr.read().bits() & SEQ_PG != SEQ_PG {
+        cortex_m::asm::nop();
+    }
```
labbott, line 112: Nightmare scenario `PG` never gets set, a forever looping hubris kind of sucks to diagnose. Maaaybe some ridiculous timeout and then panic?

cbiffle:
My testing on Gimlet C71 suggests that this will work.

app/gimlet/src/main.rs
mkeeter: Done, we now have a loop and explicitly panic after ~seconds.

app/gimlet/src/main.rs
```diff
@@ -68,47 +68,89 @@ fn system_init() {
...
+
+    // Make CRESET (PD5) an output (initially high), then toggle it low to reset
+    // the FPGA bitstream.  The minimum CRESET pulse is 200 ns, or 13 cycles,
+    // but there's a 1µF capacitor on that line.  Let's assume we're discharging
+    // the capacitor at 5 mA from 3V3; in that case, it will take 0.66 ms, or
+    // 42K cycles.  We'll be conservative and pad it to 100K cycles.
+    p.GPIOD.bsrr.write(|w| unsafe { w.bits(1 << 5) }); // set high
+    p.GPIOD.moder.modify(|_, w| w.moder5().output());
+    p.GPIOD.bsrr.write(|w| unsafe { w.bits(1 << (5 + 16)) }); // active low
```
labbott: This took me a minute to remember how `bsrr` worked. I don't know if using e.g. `bs5()` and `br5()` would be clearer (could go either way).

labbott (approved):

app/gimlet/src/main.rs
mkeeter: Good call, I was pulling this from our GPIO driver, but using the aliases is cleaner to set / clear a single bit.

app/gimlet/src/main.rs
```diff
@@ -68,47 +68,89 @@ fn system_init() {
     // you've flashed the wrong one, only.
     //
     // The revision is on pins PG[2:0], with PG2 as the MSB.
+    //
+    // In earlier firmware, we used the SP's internal pull-up resistors.
+    // However, the combination of both FPGA and SP pull-up resistors was too
+    // strong on certain hardware revisions, leading to marginal voltage
+    // readings.
```
hawkw: turbo nitpicky: might want to link to the Github issue for this here, in case a future reader wants to read over the backstory?

app/gimlet/src/main.rs
```diff
@@ -68,47 +68,89 @@ fn system_init() {
...
+    // Instead, we reset the FPGA, which puts its pins into their default
+    // configuration (with weak pull-ups enabled).  Then, we wait for the pins
+    // to charge before reading the values.
 
-    // Un-gate the clock to GPIO bank G.
+    // Un-gate the clock to GPIO bank G (for revision ID), D (for FPGA reset),
+    // and C (for sequencer PG)
     p.RCC.ahb4enr.modify(|_, w| w.gpiogen().set_bit());
+    p.RCC.ahb4enr.modify(|_, w| w.gpioden().set_bit());
+    p.RCC.ahb4enr.modify(|_, w| w.gpiocen().set_bit());
```
hawkw: Out of curiosity: is it important that this be done in three separate writes? Or, could we do
```suggestion
    p.RCC.ahb4enr.modify(|_, w| {
        w.gpiogen().set_bit());
        w.gpioden().set_bit());
        w.gpiocen().set_bit());
    });
```
if having the writes be separate is important, it could be nice to write that down so someone doesn't come along and "refactor" it.

app/gimlet/src/main.rs
```diff
@@ -68,47 +68,89 @@ fn system_init() {
...
     cortex_m::asm::dsb();
-    // PG2:0 are already inputs after reset, but without any pull resistors.
+
+    // Make PC6 (SEQ_REG_TO_SP_V3P3_PG) and PC7 (SEQ_REG_TO_SP_V1P2_PG) inputs,
+    // then wait for both of them to go high.  We time out after 1M iterations
+    // (with 100 cycles each), which is roughly 1.5s.
+    #[rustfmt::skip]
+    p.GPIOC.moder.modify(|_, w| w
+        .moder6().input()
+        .moder7().input());
```
hawkw: what was so bad about rustfmt's intended formatting? is it that the name of the bit and the `.input()` went on separate lines?

could we instead do a separate expression for each bit, so we could have the `rustfmt`-approved formatting of:
```suggestion
    p.GPIOC.moder.modify(|_, w| {
        w.moder6().input();
        w.moder7().input();
    });
```

hawkw: oh, i see that there are other `#[rustfmt::skip]`ped blocks elsewhere in this file. meh. whatever. i still would prefer to have rustfmt-approved formatting, but :woman_shrugging:

app/gimlet/src/main.rs
mkeeter: Sure, done

app/gimlet/src/main.rs
mkeeter: I was initially unsure whether they were all on `ahb4enr`; consolidated!

app/gimlet/src/main.rs
mkeeter: This was matching historical style, but I agree that multiple calls is cleaner (although we can't have the trailing semicolon, because the function has to return the `w` reference 🫠).  Fixed!

PR #2259, opened 2025-09-30
stm32h7-startup: reorder AXI erratum and cache setup

cbiffle:
@mkeeter noticed nondeterminism in the delay loop for measurement handoff, which caused us to realize that it was being performed _before_ the caches were set up ... which means small alignment changes in flash can have huge impact on the execution time of a tight loop.

While considering fixes, I noticed that the fix for the AXI SRAM read erratum had gradually drifted below other code. This was safe when I initially implemented it, because we weren't using AXI SRAM for kernel stack etc. at the time. We are now!

This commit moves the AXI SRAM erratum fix into pre_init where it runs before most RAM accesses, and moves cache setup earlier in system_init_custom to ensure that caches are on for the rest of the process, including measurement handoff.

### DIFF
--- a/drv/stm32h7-startup/src/lib.rs
+++ b/drv/stm32h7-startup/src/lib.rs
@@ -52,6 +52,15 @@ unsafe fn system_pre_init() {
 
     // Okay, yay, we can use some RAMs now.
 
+    #[cfg(any(feature = "h743", feature = "h753"))]
+    {
+        // Workaround for erratum 2.2.9 "Reading from AXI SRAM may lead to data
+        // read corruption" - limits AXI SRAM read concurrency.
+        let axi = &*device::AXI::ptr();
+        axi.targ7_fn_mod
+            .modify(|_, w| w.read_iss_override().set_bit());
+    }
+
     // We'll do the rest in system_init.
 }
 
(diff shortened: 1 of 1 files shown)


drv/stm32h7-startup/src/lib.rs
```diff
@@ -52,6 +52,15 @@ unsafe fn system_pre_init() {
 
     // Okay, yay, we can use some RAMs now.
 
+    #[cfg(any(feature = "h743", feature = "h753"))]
+    {
+        // Workaround for erratum 2.2.9 "Reading from AXI SRAM may lead to data
+        // read corruption" - limits AXI SRAM read concurrency.
+        let axi = &*device::AXI::ptr();
+        axi.targ7_fn_mod
+            .modify(|_, w| w.read_iss_override().set_bit());
```
hawkw, line 61: ought this to have some kind of big "DON'T DO X, Y, OR Z BEFORE THIS PART" comment to prevent it from being reordered by accident again?

cbiffle: I'm open to suggestions, I thought it was pretty clear before. Moving it into preinit should make that more likely, though preinit also should maybe get some screamy comments?

hawkw: 🤷‍♀️ I dunno. Mainly I was just thinking that if this got reordered once before, we should make sure it doesn't happen again. If you think the existing comments are sufficient for that, then that's fine by me.

cbiffle: Gonna tackle this comment in a followup.

PR #2260, opened 2025-09-30
Update cortex-m-rt, cortex-m, PACs; fix things

cbiffle:
This started as a followup to #2259, trying to ensure that no startup code used RAM before RAM was all the way turned on. It has ... grown, since then.

The scope so far turns out to be:

1. Rewrite our `pre_init` routine in assembler for more precise control over the use of stack.
2. Update cortex-m-rt to 0.7.x, where they made the same fix.
3. Update all the crates that have firm dependencies on cortex-m-rt, which is primarily the PACs.
4. Fix problems introduced by arbitrary API churn in the PACs. (Right now, every PAC version bump is treated as a major version, so lots of public API gets renamed. Every time.)
5. Fix problems _revealed by_ fixing PAC API churn, like a case where we were expecting divisor enum variants to have similar names without mentioning it, and a case where LPC55 code depended on the STM32H7 SPI driver.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -659,9 +659,9 @@ checksum = "f9236877021b66ad90f833d8a73a7acb702b985b64c5986682d9f1f1a184f0fb"
 
 [[package]]
 name = "cortex-m"
-version = "0.7.5"
+version = "0.7.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cd20d4ac4aa86f4f75f239d59e542ef67de87cce2c282818dc6e84155d3ea126"
+checksum = "8ec610d8f49840a5b376c69663b6369e71f4b34484b9b2eb29fb918d92516cb9"
 dependencies = [
  "bare-metal 0.2.5",
  "bitfield 0.13.2",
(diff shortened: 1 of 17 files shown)


drv/stm32h7-startup/src/lib.rs
```diff
@@ -4,64 +4,84 @@
...
+    "1:  ldr r1, [r0, #{PWR_CSR1_offset}]",
+    "    tst r1, #(1 << {PWR_CSR1_ACTVOSRDY_bit})",
+    "    beq 1b",
+
+    // Turn on all of the smaller non-TCM non-AXI SRAMs, in case the program
+    // puts data there.
+    "    movw r0, :lower16:{RCC_addr}",
+    "    movt r0, :upper16:{RCC_addr}",
+    "    ldr r1, [r0, #{RCC_AHB2ENR_offset}]",
+    "    orrs r1, #((1 << {RCC_AHB2ENR_SRAM1EN_bit}) \
```
aapoalas, line 50: **question:** Is `orr` available in Thumb ISA? Since you don't care about the condition flags here and below. I couldn't find a definite answer from ARMs documentation (the usually excellent docs seem to have a bit of a blindspot to instruction listing on Thumb? Or they seem to suggest that rather `ORRS` is not a thing?). Not that it really matters any.

drv/stm32h7-startup/src/lib.rs
```diff
@@ -4,64 +4,84 @@
...
+    "    ldr r1, [r0]",
+    "    orrs r1, #(1 << {AXI_TARG7_FN_MOD_READ_ISS_OVERRIDE_bit})",
+    "    str r1, [r0]",
+
+    // Aaaaand we're done.
+    "    bx lr",
+    ".cfi_endproc",
+    ".size __pre_init, . - __pre_init",
+
+    PWR_addr = const 0x5802_4800, //device::PWR::ptr(),
```
aapoalas, line 67: **note:** Support for const pointers seems to be pretty close to working on nightly: https://github.com/rust-lang/rust/pull/138618

drv/spi-api/build.rs
```diff
@@ -17,21 +17,17 @@ fn main() -> Result<()> {
...
-            writeln!(
-                &mut file,
-                "    // {periph} ({} devices)",
-                p.devices.len()
-            )?;
-            for (i, name) in p.devices.keys().enumerate() {
-                let name = name.to_uppercase();
-                writeln!(&mut file, "    pub const {name}: u8 = {i};")?;
-            }
+    let global_config = build_util::config::<SpiGlobalConfig>()?;
```
aapoalas, line 20: **praise:** Wow, this would've probably been a really confusing bug: build succeeds but devices are somehow missing from the output?

PR #2262, opened 2025-10-01
Revert "stm32xx-i2c: remove soft timeout, LostInterrupt"

cbiffle:
This reverts commit e199470557c39d3accc39e346e907d7a5ef7f905, plus a small change to adapt to the NotificationBits API, which changed since this commit was originally applied.

This adds a known-flaky system load detector that crashes the i2c driver. We're doing this because otherwise the i2c driver still contains bugs that result in unpredictable hangs, and hangs are currently believed to be worse than crashes.

I'm really hoping to de-revert this in the very near future.


labbott (approved):

drv/stm32xx-i2c/src/lib.rs
```diff
@@ -1104,3 +1189,50 @@ impl I2cController<'_> {
...
+
+        // If the event arrived _and_ our timer went off, prioritize the event and
+        // ignore the timeout.
+        if received.check_notification_mask(event_mask) {
+            // Attempt to clear our timer. Note that this does not protect against
+            // the race where the kernel decided to wake us up, but the timer went
+            // off before we got to this point.
+            sys_set_timer(None, TIMER_NOTIFICATION);
+            break I2cControlResult::Interrupted;
+        } else {
```
mkeeter: Nit: we could use the new `received.has_timer_fired(TIMER_NOTIFICATION)` here

drv/stm32xx-i2c/src/lib.rs
```diff
@@ -431,13 +456,73 @@ impl I2cController<'_> {
...
+        });
+
+        panic!();
+    }
+
+    ///
+    /// A common routine to wait for interrupts, panicking the driver if the
+    /// interrupt doesn't arrive in an arbitrarily chosen period of time.
+    ///
+    fn wfi(&self) -> Result<(), drv_i2c_api::ResponseCode> {
```
mkeeter: Nit: this doesn't actually need to return a result, since it either returns `Ok(())` or panics.  Fixing this would minimize the diff, since it would let us remove a bunch of `?` below.

Issue #2236, opened 2025-09-18
Stack traces are broken after compiler bump


mkeeter:
After b35abbaa8097c2c1ee6dd3db5b3eafc3a90fd9d7, Humility can no longer get stack traces:

```
$ humility -d hubris.core.0 tasks -sl hf
humility: attached to dump
system time = 119804
ID TASK                       GEN PRI STATE
24 hf                           0   5 recv, notif: timer
   stack unwind failed: Do not have unwind info for the given address.
```

This works fine on the previous commit (86d3a0aa4), and the relevant commit is a compiler bump, which is very suspicious!

Here's an example (zipped) dump that reproduces the issue: [hubris.core.0.zip](https://github.com/user-attachments/files/22409683/hubris.core.0.zip)

Unclear if this is a Hubris or Humility issue, more to come!

dancrossnyc:
Oops, I mentioned this a couple of weeks ago in chat, tangentially related to another issue, but forgot to file an issue for it. Mea culpa!

cbiffle:
Mentioned this in chat, but -- the stack unwinding is failing for any task parked in the kernel, which means the issue is almost certainly with Humility's heuristics for starting a stack trace in a syscall stub.

Because our syscall stubs are in assembly, they don't have the DWARF info our normal functions have, _and_ they rearrange the registers so the "r7 is frame pointer" assumption fails. Humility contains some heuristics for detangling this but they're inherently fragile... something may have changed that broke them.

In exhubris I worked out the (awful, noisy) pattern of explicit assembly annotations for adding the DWARF info to the stubs. We could try porting that over and seeing if it fixes the problem.

dancrossnyc:
I'd like to double back with you regarding backporting some exhubris bits more generally, Cliff, if you're still game?

cbiffle:
> I'd like to double back with you regarding backporting some exhubris bits more generally, Cliff, if you're still game?

Yeah, we should set up some kind of a sync meeting on that. I'd love to make some progress there.

cbiffle:
Alright, Humility's syscall disassembly heuristics -- while fragile! -- are not to blame here.

The issue is that our toolchain (currently `nightly-2025-07-20`) is not producing line number information for any `naked_asm` block, as far as I can tell. The syscall stubs are affected, but so is `_start`, and you can see this with `addr2line`:

```
[...]

PR #2264, opened 2025-10-02
Add `grapefruit/app-mini.toml`

mkeeter:
While debugging #2236, I wanted a tiny image that produced compact system dumps – they're faster to load in a debug build of Humility, and we'll eventually want to commit one to the Humility test suite.

This image produces a 421KiB dump, versus 7 MiB for a full Cosmo image.  It's just enough to blink an LED so that you know it's been flashed and is running.

### DIFF
--- a/app/grapefruit/app-mini.toml
+++ b/app/grapefruit/app-mini.toml
@@ -1,0 +1,51 @@
+# Tiny Grapefruit image, mostly for testing Humility
+board = "grapefruit"
+name = "grapefruit-mini"
+target = "thumbv7em-none-eabihf"
+chip = "../../chips/stm32h7"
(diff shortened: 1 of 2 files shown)


hawkw (commented):
If we're checking this image into Git, it seems like we might also want to add a CI job that builds it, to make sure nobody breaks it inadvertently?

labbott:
> If we're checking this image into Git, it seems like we might also want to add a CI job that builds it, to make sure nobody breaks it inadvertently?

I think this should be picked up by https://github.com/oxidecomputer/hubris/pull/2263 (thanks @pietroalbini !)

hawkw (approved):

PR #2270, opened 2025-10-15
Tag releases with Brussels

emilyalbini:
The first phase of the SP and RoT release process is to determine the version number, tag the latest commit with that version number, and push it to kick off the build. To start fully automating the SP release process, this PR replaces this phase with a single GitHub Actions workflow that can be started through the UI:

image
<br/><br/>

This uses [Brussels](https://github.com/oxidecomputer/brussels), the new SP and RoT release tooling I'm working on, and in particular the changes I introduced in oxidecomputer/brussels#3. Brussels is fetched as a pre-built binary from its GitHub Release to ensure it doesn't delay the release process by trying to compile it.

The `brussels init` command will be executed in its own job as the first build step. It will look at all the existing tags to determine what is the next version number, calculate the tag name, and expose all of that information into `GITHUB_OUTPUT` to let the rest of the workflow know the version to use. It will also write all the metadata it gathered in a `brussels-manifest.json` file, so that future Brussels steps can reference that metadata.

The tag will not be created and pushed by `brussels init`. Instead, we will let the "create GitHub Release" step at the end of the workflow create the tag on its own, as GitHub will do so when creating a release without an underlying tag.

While I was there, I also changed the build process to inject the version number directly during the build, instead of waiting for permslip to add it as part of the signature process.

This is just the first improvement to the SP release process that I'm going to do. The end goal is for SP releases to be fully automated (either by requiring a single button click or by putting the whole release process in a cronjob).

Note that by default the workflow will use the latest release of Brussels. An optional field when starting the workflow run lets you download Brussels from a CI job in the Brussels repository. This mostly exists to help me test changes without publishing endless releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,7 @@ jobs:
     name: Initialize the release
     runs-on: ubuntu-24.04
     outputs:
+      version: ${{ steps.init.outputs.version }}
       tag_name: ${{ steps.init.outputs.tag_name }}
       hubris_app_dirs: ${{ steps.init.outputs.hubris_app_dirs }}
     steps:
@@ -62,6 +63,7 @@ jobs:
       os: oxide-colo-builder-hubris
       filter-directories: ${{ needs.init.outputs.hubris_app_dirs }}
       upload-artifacts: true
+      caboose-version: ${{ needs.init.outputs.version }}
       attest: true
 
   release:
@@ -84,7 +86,7 @@ jobs:
         tag_name: ${{ needs.init.outputs.tag_name }}
         name: "${{ github.event.inputs.group }} release"
         fail_on_unmatched_files: true
-        body: "These are UNVERSIONED hubris artifacts"
+        body: "These are versioned but UNSIGNED hubris ${{ github.event.inputs.group }} artifacts"
         files: |
           ${{ steps.grab.outputs.download-path }}/*.zip
           ${{ steps.grab.outputs.download-path }}/brussels-manifest.json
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,48 +1,92 @@
...
       with:
         path: out
-    - name: prep
-      run: |
-        OUT=${{ steps.grab.outputs.download-path }}
-        for build in `ls $OUT`; do
-           for f in `ls $OUT/$build`; do
-              mv $OUT/$build/$f $OUT/$f
-           done
-        done
```
labbott, line 39: Is this backwards compatible? It's not strictly necessary to keep it compatible but I just want to know if other tooling needs to be updated (honestly I forget what most of this was supposed to be doing anyway)

.github/workflows/release.yml
```diff
@@ -1,48 +1,92 @@
...
-    - name: cut release
+    - name: Publish the release
       uses: softprops/action-gh-release@v1
       with:
-        name: "${{ inputs.board-set }} release"
+        tag_name: ${{ needs.init.outputs.tag_name }}
+        name: "${{ github.event.inputs.group }} release"
         fail_on_unmatched_files: true
-        body: "These are UNVERSIONED hubris artifacts"
+        body: "These are versioned but UNSIGNED hubris ${{ github.event.inputs.group }} artifacts"
```
labbott, line 89: I'm slightly wary about releasing anything versioned because it is a change in our assumptions about what is released/trusted. That said, we're re-thinking this entire release process anyway so I think this is fine and clarifies what these artifacts are.

.github/workflows/release.yml
emilyalbini: Yes, the behavior of `merge-multiple: true` is equivalent to that snippet of code, and does what you'd expect when downloading multiple artifacts (all of their file end up in the target directory, instead of a subdir per artifact).

PR #2275, opened 2025-10-27
add minibar-a and minibar-b boards

Aaron-Hartwig:
These boards can both use the same FPGA image. Due to a hardware bug, the SP could not talk to the management network switch on minibar-a, but that was fixed on minibar-b so we can add the `net` task there.

### DIFF
--- a/app/minibar/src/main.rs
+++ b/app/minibar/src/main.rs
@@ -85,6 +85,7 @@ fn system_init() {
 
     // Build the full ID
     let rev = p.GPIOK.idr.read().bits();
+    // The pins are listed in descending order as that how they got wired up.
     let rev = [7, 6, 5]
         .iter()
         .enumerate()


app/minibar/src/main.rs
```diff
@@ -85,15 +85,17 @@ fn system_init() {
 
     // Build the full ID
     let rev = p.GPIOK.idr.read().bits();
-    let rev = [5, 6, 7]
+    let rev = [7, 6, 5]
```
labbott, line 89: This is a little non obvious, would you mind adding a short comment explaining this sequence?

labbott (commented):
LGTM otherwise

app/minibar/src/main.rs
Aaron-Hartwig: Agreed. Will do!

PR #2284, opened 2025-11-11
psu_seq: refactor ereports to use `microcbor`

hawkw:
PR #2246 introduced the `microcbor` crate for derivable CBOR serialization with a fixed max length. That branch changed the ereports in `drv_gimlet_seq_server` to use the new derives. The intent at the time was to update other existing ereports, such as those in `drv_psc_seq_server`, separately. This branch does that.


drv/psc-seq-server/src/main.rs
```diff
@@ -1172,37 +1191,30 @@ include!(concat!(env!("OUT_DIR"), "/i2c_config.rs"));
...
-        pub(super) refdes: &'static str,
-        #[serde(serialize_with = "serialize_psu_rail")]
-        pub(super) rail: Slot,
-        #[serde(serialize_with = "serialize_psu_slot")]
-        pub(super) slot: Slot,
+    #[derive(microcbor::EncodeFields)]
+    pub(super) struct EreportFields {
+        pub(super) refdes: FixedStr<20>, // Component ID max length
+        pub(super) rail: FixedStr<8>,    // "V54_PSUx"
+        pub(super) slot: u8,
```
mkeeter, line 1212: Could this be a `PsuSlot` instead?  That would save some round-tripping through `u8` / `usize`.

hawkw: Currently `microcbor` doesn't have a way to tell it to `PsuSlot` to serialize as an integer. I should add that, but was going to separately.

mkeeter: Okay, sounds reasonable!

PR #2285, opened 2025-11-12
Add online tech port unlock key to authorized_keys

plotnick:
Public key from [OSS#207](https://github.com/oxidecomputer/online-signing-service/issues/207) extracted via:
```
permslip --url=https://signer-us-west.corp.oxide.computer public-key "Tech Port Unlock Production 1" > techport.key && ssh-keygen -i -m PKCS8 -f techport.key
```

Also update `ssh-key` crate to pull in [SSH#289](https://github.com/RustCrypto/SSH/pull/289), so that we can use the key name (which includes spaces) as a comment. If this is controversial, happy to roll it back, but Matt did the work to fix this, we may as well use it!

Builds with `cargo xtask dist app/sidecar/rev-d.toml`, but have I not run any other tests or attempted to run an image on hardware.

### DIFF
--- a/drv/cosmo-hf/src/main.rs
+++ b/drv/cosmo-hf/src/main.rs
@@ -216,7 +216,11 @@ impl FlashDriver {
         // desired status transition occurs in less than 1ms, avoiding a 1-2ms
         // sleep and round-trip through the scheduler. status transitions
         // quickly.
-        const MAX_BUSY_POLLS: u32 = 32;
+        //
+        // Initial tests show that waiting on the FPGA takes ~50 polls when
+        // page programming. Waiting even 1 tick is still more costly than
+        // just waiting for the poll to finish.
+        const MAX_BUSY_POLLS: u32 = 100;
 
         let mut busy_polls = 0;
         while !poll(self) {


support/support_tokens/README.md
```diff
@@ -1,9 +1,13 @@
-This folder contains the public keys for our support PIV tokens.
+This folder contains the public keys for our support PIV tokens and online
+tech port unlock.
 
-`authorized_keys` is from [`oxidecomputer/support-piv-tokens@36abeb82`](https://github.com/oxidecomputer/support-piv-tokens/blob/36abeb82aacb3640be2f6c988e49844356cf65a8/authorized_keys).
+Most of `authorized_keys` is from [`oxidecomputer/support-piv-tokens@36abeb82`](https://github.com/oxidecomputer/support-piv-tokens/blob/36abeb82aacb3640be2f6c988e49844356cf65a8/authorized_keys),
+edited to remove spaces from the comment field; see [`rustcrypto/SSH#289`](https://github.com/RustCrypto/SSH/pull/289).
```
mkeeter: Should we un-edit this, now that we've bumped `SSH` to support comments with spaces?

mkeeter:
I have run the same `permslip` command and gotten the same key (`ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBF+RgEfwv33ejIgatKLqdmJiAMsodKrbRlXus4rQFFZWbXRDhwcvIDrXZyRJaNh0GeDx9tT8E2l+z6lRN80SxS8=`).

support/support_tokens/README.md
plotnick: Good question; I thought about that. I didn't know if those comments are recorded or used anywhere else, so didn't want to edit the existing entries. But yes, I think we should unless there's a good reason not to.

flihp:
LGTM as well: is the same public key we approved yesterday

support/support_tokens/README.md
plotnick: All right, I think I spent more time thinking about it than doing it. Done in 09d52599; if you would, please verify that the keys have remain unchanged, only the comments have changed and now match [the ones in support-piv-tokens](https://github.com/oxidecomputer/support-piv-tokens/blob/main/authorized_keys).

plotnick: So, having edited the comments, they will no longer match those in Omicron. Should we now re-mirror?

(I knew this was going to be a small rabbit hole.)

mkeeter: The file in [omicron](https://github.com/oxidecomputer/omicron/blob/main/smf/switch_zone_setup/support_authorized_keys) already has spaces, so it looks like they now match!

mkeeter:
Thinking about testing this: we could temporarily edit this `match` to always return an ECDSA challenge (regardless of RoT lifecycle state) https://github.com/oxidecomputer/hubris/blob/ec18e4f11aaa14600c61f67335c32b250ef38269/task/control-plane-agent/src/mgs_sidecar.rs#L675-L695

Then we could test on an (unlocked) lab Sidecar, with both a Yubikey holder and someone that can use the new online unlock.

PR #2286, opened 2025-11-13
Tweak polling timeout in `cosmo-hf`

labbott:
The existing polling number will result in us always hitting the sleep case when waiting for the FPGA to go idle when programming a page. It turns out sleeping for even one tick will take up much more time than waiting for the poll to complete. Up our timeout to something sufficiently large we are unlikely to hit it on common cases.

`faux-mgs update host-flash` before

```
real	6m46.185s
user	0m4.559s
sys	0m3.476s
```

after

```
real	3m31.962s
user	0m4.325s
sys	0m3.188s
```


mkeeter (approved):

labbott:
I think there's probably more work to be done here but this is a definite improvement so I'm going to go ahead and merge

PR #2293, opened 2025-11-20
Cosmo Host flash performance improvements

labbott:


drv/cosmo-hf/src/hf.rs
```diff
@@ -421,11 +421,13 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
         self.drv.check_flash_mux_state()?;
         let addr = self.flash_addr(addr, data.len() as u32)?;
         self.invalidate_write();
+        let mut block: [u8; PAGE_SIZE_BYTES] = [0; PAGE_SIZE_BYTES];
```
mkeeter: We already have a page-size buffer that's **not** on the stack in `self.apob_bufs.page`.

How about renaming `ApobBufs` to just `HfBufs` (renaming its `persistent_data` member to `apob_persistent_data`), moving it from `apob.rs` into this file, then using those buffers instead of stack arrays?

drv/cosmo-hf/src/main.rs
```diff
@@ -281,6 +284,34 @@ impl FlashDriver {
         self.wait_flash_busy(Trace::SectorEraseBusy);
     }
 
+    /// Returns true if the full 64KB sector is erased
+    /// (all bits are set to `1`)
+    fn flash_is_sector_erased(&mut self, addr: FlashAddr) -> bool {
+        let cnt = SECTOR_SIZE_BYTES / (PAGE_SIZE_BYTES as u32);
+        for i in 0..cnt {
+            let addr = addr.0 + i * (PAGE_SIZE_BYTES as u32);
+            self.clear_fifos();
```
mkeeter, line 293: I think you could move everything except `self.drv.addr.set_addr(addr)` out of this loop, dunno how much of a performance impact that would have.

drv/cosmo-hf/src/main.rs
```diff
@@ -314,44 +345,38 @@ impl FlashDriver {
...
-                        };
-                        *byte = d;
-                    }
-                }
-                let v = u32::from_le_bytes(v);
-                self.drv.tx_fifo_wdata.set_fifo_data(v);
-            }
-            self.drv.instr.set_opcode(instr::QUAD_INPUT_PAGE_PROGRAM_4B);
-            self.wait_fpga_busy();
+    /// This function will only return an error if the slice len is not a full
```
mkeeter: It makes sense that this now takes a slice, but I'm not sure why this PR changes it to only accept a single page.  Was the overhead of breaking the slice into page chunks too much?  That seems implausible, but if that's the case, the signature should change to take `&[u8; PAGE_SIZE]` to enforce correctness on the callers.

labbott: Hmmm, now that you mentioned it I don't think this is correct for the persistent case since that's less than a page. `page_program` is limited to a page size and I don't think the current loop correctly implements the address for more than a page https://github.com/oxidecomputer/hubris/blob/master/drv/cosmo-hf/src/main.rs#L326-L354 (unless I've misread something)

mkeeter: I agree that the lack of address updates seems wrong in the existing code 😅 

Since persistent data writes < 1 page, how about fixing that in the existing (arbitrary-length) write?

aapoalas:
The [tlvc PR](https://github.com/oxidecomputer/tlvc/pull/9) I've had open for a while might be worth returning to for Cosmo perf work: it's probably not a large effect on performance one way or the other, but it does probably remove panics so that's always nice.

I also have done some work on the Cosmo host flash APIs in a branch to remove panics here; I think it would require a little bit of more work to make it worth a PR, but it did also remove some panics so that's nice. (I think I also found one bug IIRC.) If you'd be willing to review it, I can take a look at freshening up that work.

drv/cosmo-hf/src/main.rs
labbott: I think I miss what "that" is referring to here but I was going to relax the requirement to be limited to a single page write. I think most of the code elsewhere is already breaking things down in to page sizes so it seems redundant to do that here too.

labbott:
> I also have done some work on the Cosmo host flash APIs in a branch to remove panics here; I think it would require a little bit of more work to make it worth a PR, but it did also remove some panics so that's nice. (I think I also found one bug IIRC.) If you'd be willing to review it, I can take a look at freshening up that work.

It's a little hard to say without actually seeing the code. I'm not opposed to removing panics (I've sometimes had a war against `unwrap`) but it can also be a trade off for having the code be harder to follow.

aapoalas:
> It's a little hard to say without actually seeing the code. I'm not opposed to removing panics (I've sometimes had a war against `unwrap`) but it can also be a trade off for having the code be harder to follow.

The commits is [here](https://github.com/oxidecomputer/hubris/commit/b1183332d2215638491574e826abb6b3ca6161f7) but it's currently full of wrapping arithmetic to remove arithmetic panics that the compiler couldn't figure out, together with comments to explain why there's no chance of an overflow/wrap there: this is "correct" but not actually **correct**. Freshening this up would consist of replacing the wrapping arithmetic with such code transformations, together with checked or unchecked arithmetic as necessary, to properly explain the intended semantics of the code to the compiler, to enable it to remove the panics on its own and unsafely assert them when it can't.

drv/cosmo-hf/src/main.rs
labbott: What actually triggers the start of the transaction? is it the addr or the OpCode? I'm not a huge fan of breaking this up, especially since there doesn't seem to be a notable increase in performance.

hawkw: Yeah, if we're going to do this, it seems worth consulting the datasheet to be sure that the other configurations aren't cleared by the transaction completing?

drv/cosmo-hf/src/main.rs
```diff
@@ -270,6 +270,9 @@ impl FlashDriver {
 
     /// Erases the 64KiB flash sector containing the given address
     fn flash_sector_erase(&mut self, addr: FlashAddr) {
+        if self.flash_is_sector_erased(addr) {
+            return;
+        }
```
hawkw, line 275: A (potentially sketchy) thought: might we consider caching erased sectors internally, so that once the driver has either erased or written to a sector, it writes down whether or not the sector is erased, and we can avoid reading all 64KB to check if it's erased? We can always fall back to doing that for sectors the driver isn't sure about.

drv/cosmo-hf/src/main.rs
```diff
@@ -281,6 +284,34 @@ impl FlashDriver {
...
+        for i in 0..cnt {
+            let addr = addr.0 + i * (PAGE_SIZE_BYTES as u32);
+            self.clear_fifos();
+            self.drv.data_bytes.set_count(PAGE_SIZE_BYTES as u16);
+            self.drv.addr.set_addr(addr);
+            self.drv.dummy_cycles.set_count(8);
+            self.drv.instr.set_opcode(instr::FAST_READ_QUAD_OUTPUT_4B);
+            let mut erased = true;
+            // We need to complete the read of the full PAGE_SIZE
+            // even if we find a non-erased byte
```
hawkw: Is that to ensure that the transaction has completed before trying to do something else?

drv/cosmo-hf/src/main.rs
```diff
@@ -314,44 +345,38 @@ impl FlashDriver {
...
+
+        self.flash_write_enable();
+        self.drv.data_bytes.set_count(len as u16);
+        self.drv.addr.set_addr(addr.0);
+        self.drv.dummy_cycles.set_count(0);
 
-            // Wait for the busy flag to be unset
-            self.wait_flash_busy(Trace::WriteBusy);
+        for chunk in data.chunks(4) {
+            let v = u32::from_le_bytes(chunk.try_into().unwrap());
```
hawkw: nit: should this be an `unwrap_lite()` to avoid having a bonus string?

drv/cosmo-hf/src/main.rs
labbott: I thought about that but decided against it for now simply because it's more work/bookkeeping . I'm hoping the performance here is "good enough" (will mupdate without being painfully slow)

drv/cosmo-hf/src/main.rs
labbott: Yes, I was seeing what looked like incorrect behavior (mismatch of written vs expected data) without completing the full `PAGE_SIZE` read.

drv/cosmo-hf/src/main.rs
labbott: oh yeah probably

drv/cosmo-hf/src/main.rs
mkeeter: IIRC writing the `OpCode` is what triggers the start of the transaction.  The other fields are internal to the FPGA, so we're the ones that define their behavior (e.g. whether they clear), but I don't feel strongly about saving a few writes.

drv/cosmo-hf/src/main.rs
mkeeter: My suggestion is to allow `flash_write` to write arbitrary-length buffers (i.e. the previous implementation), but fix the bug that you noticed where `addr` isn't incremented between pages.

This has the advantage that the function would become infallible, so callers wouldn't need to think about error handling.  If callers are already splitting stuff up into page-size transactions, this would add a tiny bit of overhead, but save an `unwrap` / `unwrap_lite`.

drv/cosmo-hf/src/main.rs
hawkw: maybe worth noting that a little more explicitly in this comment?

nathanaelhuffman: is that possibly because you just need to bleed down the FIFOs? The hardware here will stuff the remainign bytes into the FIFOs.  If you have waited for the transaction to finish, you could optionally use the fifo clear bits to nuke them?  This is basically a "you asked for PAGE_SIZE bytes, so you get PAGE_SIZE bytes into the response FIFO" I think

labbott: so the loop had `self.clear_fifos()` each time which I thought should be clearing everything. I'm also now questioning if this was actually incorrect given I also spent too long dealing with https://github.com/oxidecomputer/humility/pull/585 . Perhaps I will check again...

aapoalas:
I opened https://github.com/oxidecomputer/hubris/pull/2296 from the perf work I had done earlier, fixed up to not use wrapping arithmetic all over the place.

With those changes in place and the tlvc PR in use as well, there are only a handful of panics or unwraps in the cosmo-b-dev/dist/auxflash and at least none in any direct hot loops that I can see. Maybe some of that stuff is useful here?

drv/cosmo-hf/src/main.rs
labbott: I tried this again, I think I was missing another `wait_for_fpga` for the FPGA to actually be finished processing after draining the FIFO. With that in we don't actually see any difference in exiting out early vs just completing all the reads, probably because it still takes just as much time to drain. I'll tweak the comment.

nathanaelhuffman (approved):
ty for the work here!

drv/cosmo-hf/src/main.rs
hawkw: Yup, I think that's fine, especially since this change already improves things substantially.

drv/cosmo-hf/src/main.rs
```diff
@@ -281,6 +284,37 @@ impl FlashDriver {
...
+            self.drv.data_bytes.set_count(PAGE_SIZE_BYTES as u16);
+            self.drv.addr.set_addr(addr);
+            self.drv.dummy_cycles.set_count(8);
+            self.drv.instr.set_opcode(instr::FAST_READ_QUAD_OUTPUT_4B);
+            let mut erased = true;
+            // Technicaly we could terminiate this loop early when
+            // we find the first non-erased byte but we still have
+            // to drain the FIFOs and wait for the FPGA which means
+            // there's no performance gain vs just reading everything
+            // in a loop here.
```
hawkw: thanks for adding this, this is a lot clearer. i noticed a couple typos and grammar nits:
```suggestion
            // Technically, we could terminate this loop early when
            // we find the first non-erased byte, but we still have
            // to drain the FIFOs and wait for the FPGA, which means
            // there's no performance gain vs just reading everything
            // in a loop here.
```

drv/cosmo-hf/src/main.rs
```diff
@@ -314,45 +348,41 @@ impl FlashDriver {
...
-                for (j, byte) in v.iter_mut().enumerate() {
-                    let k = i * 4 + j;
-                    if k < len {
-                        let Some(d) = data.read() else {
-                            return Err(());
-                        };
-                        *byte = d;
-                    }
+                for (i, b) in chunk.iter().enumerate() {
+                    v[i] = *b;
```
hawkw: suuuuper nitpick, please feel free to ignore me: i think that `iter().enumerate()` over `chunk` elides bounds-checking accesses to `chunk`, but not to `v`; i _think_ if we did
```suggestion
                for (b, v_i) in chunk.iter().zip(v.iter_mut()) {
                    *v_i = *b;
```
we might elide both bounds checks?

drv/cosmo-hf/src/main.rs
```diff
@@ -281,6 +284,37 @@ impl FlashDriver {
...
+        let cnt = SECTOR_SIZE_BYTES / (PAGE_SIZE_BYTES as u32);
+        for i in 0..cnt {
+            let addr = addr.0 + i * (PAGE_SIZE_BYTES as u32);
+            self.clear_fifos();
+            self.drv.data_bytes.set_count(PAGE_SIZE_BYTES as u16);
+            self.drv.addr.set_addr(addr);
+            self.drv.dummy_cycles.set_count(8);
+            self.drv.instr.set_opcode(instr::FAST_READ_QUAD_OUTPUT_4B);
+            let mut erased = true;
+            // Technicaly we could terminiate this loop early when
```
mkeeter: nit:
```suggestion
            // Technically we could terminate this loop early when
```

drv/cosmo-hf/src/main.rs
```diff
@@ -314,45 +348,41 @@ impl FlashDriver {
...
-        addr: FlashAddr,
-        data: &mut dyn idol_runtime::BufReader<'_>,
-    ) -> Result<(), ()> {
-        loop {
-            let len = data.remaining_size().min(PAGE_SIZE_BYTES);
-            if len == 0 {
-                break;
-            }
+    /// This function will only return an error if the slice len is greater than
+    /// a page
```
mkeeter:
```suggestion
```

drv/cosmo-hf/src/main.rs
hawkw: i had to go godbolt this because i have autism or something, and yeah, the zippy one is nicer: https://godbolt.org/z/3avMbshn5

mkeeter (approved):

hawkw (approved):

PR #2296, opened 2025-11-20
feat(auxflash): Slot, SlotOffset wrappers

aapoalas:


drv/auxflash-api/src/lib.rs
```diff
@@ -158,18 +158,19 @@ where
...
                     if inner_chunk.header().tag == tag {
-                        // At this point, the inner reader is positioned *after*
-                        // our target chunk.  We back off by the full length of
-                        // the chunk (including the header), then offset by the
-                        // header size to get to the beginning of the blob data.
-                        let (_, inner_offset, _) = inner_reader.into_inner();
-                        let pos = inner_offset
-                            - inner_chunk.header().total_len_in_bytes() as u64
-                            + core::mem::size_of::<tlvc::ChunkHeader>() as u64;
+                        let pos = u32::try_from(inner_chunk.body_position())
```
aapoalas, line 161: **note:** I was looking at the original code here and became convinced that this just wants the `body_position` from `inner_chunk`. There's still a chance I'm wrong, I didn't drag a comb through the code to really, truly convince myself of it but definitely the `inner_offset` is the position of the next chunk, and offsetting that by the total size - header size should give us the previous chunk's body position.

It seems prudent to just add an API to expose that location to tlvc; I have added that API to the PR that I have open in [tlvc](https://github.com/oxidecomputer/tlvc/pull/9).

drv/auxflash-server/src/main.rs
```diff
@@ -202,27 +213,48 @@ impl ServerImpl {
             self.active_slot.ok_or(AuxFlashError::NoActiveSlot)?;
 
         let spare_slot = active_slot ^ 1;
+        const {
```
aapoalas: **note:** I've added these const-assertions here and there to try and help the compiler understand (locally) that checking `active_slot < SLOT_COUNT` also asserts that `active_slot * SLOT_SIZE + SLOT_SIZE` cannot overflow etc. Sometimes it works, but not every time.

drv/auxflash-server/src/main.rs
```diff
@@ -202,27 +213,48 @@ impl ServerImpl {
...
+            assert!(SLOT_COUNT.checked_mul(SLOT_SIZE as u32).is_some());
+            assert!(SLOT_COUNT > 0 && SLOT_COUNT.is_multiple_of(2));
+        }
+        // SAFETY: active_slot and spare_slot are valid slot numbers. The
+        // active slot is chosen using scan_for_active_slot and never
+        // reassigned, while spare_slot is its dual. Slots are always created
+        // in pairs as per above check, so flipping the 0th bit keeps slot
+        // validity.
+        unsafe {
+            assert_unchecked(active_slot < SLOT_COUNT);
```
aapoalas: **thought:** Having a type for the slot number would seem pretty useful.

drv/auxflash-server/src/main.rs
```diff
@@ -202,27 +213,48 @@ impl ServerImpl {
...
+            base: active_slot_base,
         };
         let mut reader = TlvcReader::begin(handle)
             .map_err(|_| AuxFlashError::TlvcReaderBeginFailed)?;
         while let Ok(Some(..)) = reader.next() {
             // Nothing to do here
         }
-        let data_size = SLOT_SIZE - reader.remaining() as usize;
+        // SAFETY: this cannot wrap, as our reader's max size is SLOT_SIZE.
+        let data_size =
```
aapoalas: **note:** Nope, the compiler couldn't figure this one out :(

drv/auxflash-server/src/main.rs
```diff
@@ -248,8 +280,12 @@ impl ServerImpl {
                 .map_err(qspi_to_auxflash)?;
             self.poll_for_write_complete(None)?;
 
-            read_addr += amount;
-            write_addr += amount;
+            // SAFETY: these cannot overflow as amount is at most
```
aapoalas: **note:** Nope, compiler is unable to figure this one out for itself :(

drv/auxflash-server/src/main.rs
```diff
@@ -372,36 +423,53 @@ impl idl::InOrderAuxFlashImpl for ServerImpl {
...
-        let mut read = 0;
-        while addr < mem_end {
-            let amount = (mem_end - addr).min(buf.len());
-            data.read_range(read..(read + amount), &mut buf[..amount])
+        let mut mem_offset = 0usize;
+        while mem_offset < data_len {
+            let amount = (data_len - mem_offset).min(buf.len());
+            // SAFETY: amount takes us to data_len in at most buf sized
+            // chunks. The next offset is always at most data_len.
+            let next_mem_offset = unsafe { mem_offset.unchecked_add(amount) };
```
aapoalas: **note:** Nope, not even this one :(

drv/stm32h7-qspi/src/lib.rs
```diff
@@ -256,7 +256,7 @@ impl Qspi {
 
             // How much space is in the FIFO?
             let fl = usize::from(sr.flevel().bits());
-            let ffree = FIFO_SIZE - fl;
+            let ffree = FIFO_SIZE.saturating_sub(fl);
```
aapoalas, line 259: **note:** Easy panic path here to remove; saturating arithmetic is what we know this comes out to, so it's an easy choice. Equally well it would be fine to make this unchecked.

PR #2301, opened 2025-11-25
Merge PR workflows into a single one

emilyalbini:
This PR merges the four workflows that run in every PR into a single one, and adds a "CI finished" job that executes when all workflows finish. This has a few consequences:

* Once we gate PRs on the new "CI finished" job instead of individual jobs, all jobs will be required for a PR to be merged. Adding new required jobs or renaming existing ones won't need admin access to change branch protection anymore.

* Cancelling CI for a PR will require cancelling only one workflow rather than four.

* The "Actions" tab will be cleaner, with only one execution per PR and less workflows to scan through on the sidebar.

I remember talking with folks a few weeks/months ago about this, and there was a concern about having to wait more time before being able to click the merge button if we make all of the workflows required. IMO this is less of a concern now that we can mark a PR as "merge when all required checks pass", does that alleviate the concerns?


labbott:
> I remember talking with folks a few weeks/months ago about this, and there was a concern about having to wait more time before being able to click the merge button if we make all of the workflows required. IMO this is less of a concern now that we can mark a PR as "merge when all required checks pass", does that alleviate the concerns?

I think it does. I think when we were originally having problems the "merge when all checks pass" feature was not available or didn't do what we want.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+    steps:
+      - name: Checkout the source code
+        uses: actions/checkout@v6
+
+      - name: cargo fmt 
+        run: cargo fmt --all --check
+
+  finish:
+    name: CI finished
+    runs-on: ubuntu-slim
```
labbott, line 52: out of curiousity, what's the difference between `ubuntu-latest` and `ubuntu-slim`

emilyalbini: `ubuntu-slim` is a very recent 1 vCPU runner that's supposed to start faster and have less overhead.

https://github.blog/changelog/2025-10-28-1-vcpu-linux-runner-now-available-in-github-actions-in-public-preview/

PR #2302, opened 2025-11-26
More CI improvements

emilyalbini:
Followup to #2301, this makes a few improvements to PR CI:

* Old jobs will now be cancelled automatically if you push a new commit to the same PR. This will hopefully avoid waste.
* The documentation workflow will now be part of the main workflow, and split into two jobs: building the documentation site will happen every time (even in PRs), and actually deploying that to GitHub Pages will happen in a separate job (only on the default branch). This will make sure we test generating the documentation in the PR itself.
* Renamed jobs to be consistent, now that we don't have to gate on individual branch names in branch protection.
* Tightened up permissions for the workflow, and removed a redundant third party action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,14 +17,18 @@ permissions: {}
 
 jobs:
   dist:
-    name: dist
+    name: ${{ matrix.name }}
     permissions:
       contents: read
       id-token: write
       attestations: write
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            name: Linux
+          - os: windows-latest
+            name: Windows
     uses: ./.github/workflows/build-boards.yml
     with:
       os: ${{ matrix.os }}
@@ -43,7 +47,7 @@ jobs:
         uses: apache/skywalking-eyes/header@501a28d2fb4a9b962661987e50cf0219631b32ff
 
   tests:
-    name: tests
+    name: Run tests
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/build-boards.yml
```diff
@@ -106,20 +106,17 @@ jobs:
...
-        with:
-          repo: "oxidecomputer/humility"
-          version: "59047694"
-          file: "humility"
-          target: "target/release/humility"
+        run: |
+          gh release download --repo oxidecomputer/humility --pattern humility --dir target/release nightly
+          chmod +x target/release/humility
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
labbott: Oh this is much nicer. If you didn't go digging, we're doing this weird nightly hack so we could pick up humility fixes immediately (this was a bigger concern when there was more churn between humility and hubris)

.github/workflows/build-boards.yml
labbott, line 61: Can we drop the permissions here now since they are being defined in `ci.yml`?

emilyalbini: Not really, permissions need to be defined in both places: in `ci.yml` the permissions the whole reusable workflow can use are determined, and then within each job in the reusable workflow we need to define the permissions that job uses.

I tried to make it smarter and only get id-token and attestations permissions when requesting an attested build, but GitHub didn't budge :(

labbott: Sigh Github

labbott (approved):

emilyalbini:
Don't merge this yet.

emilyalbini:
Ok it turns out `gh` is not installed in the colo runners, switched that invocation with a direct `curl` to download the file.

PR #2305, opened 2025-11-26
stm32h7-eth: mask MMC interrupts

### LINKED ISSUE #2304: STM32H7 Ethernet driver stops yielding CPU after many packets
The current version of our STM32H7 Ethernet MAC driver contains a bug that causes it, after considerable uptime and many packets exchanged, to start starving lower priority tasks of CPU. Because the driver runs in the `net` task, and all network services are _inherently_ lower priority than `net`, this has the visible effect of causing the management interface to stop responding to UDP (ICMP continues to work).

This issue has been present since the very first version of the driver, and I think explains some rare/intermittent network issues we've seen in test systems. It's been rare in practice because it takes some months to manifest, and we've been restarting machines for update more often than that. Unfortunately, the evidence that helped us finally nail it was collected after it manifested at a customer site.

# Symptoms

Sometimes, when we least expected it, an SP would drop off the management network. This often went unnoticed, because all non-management traffic kept operating fine. When we noticed it early on, it often manifested as "sled won't take an update." As the control plane has become more chatty as features have been built out, this has become somewhat more impactful.

The exact symptoms are:

- Control plane agent stops responding to control plane traffic.
- Other task-based UDP services, such as udpecho, broadcast, and debug dump access (if enabled), also stop responding.
- ICMP and NNDP continue working fine. The machine can be detected and pinged.
- On a compute sled, host serial access to its own SP also stops working, though this is generally much harder to notice (the host pretty much just tolerates it).
- The uptime of the SP at the time of failure is typically measured in months, but is not consistent.

# One cause: hardware management counter interrupts

While I'm not convinced this is the _only_ cause, we've identified one culprit: the STM32H7 Ethernet management counters.

The Ethernet block in the 'H7 provides a suite of management counters as part of its MMC subsystem. These count things like "number of unicast packets received," "CRC errors," etc. We were ignoring them because they just quietly count and eventually rollover. If you're not using them, you needn't configure them.

...or so we thought.

After this happened to a group of machines _nearly simultaneously,_ the team pored over datasheets and code to try to find anything we'd missed. It turns out, we had in fact missed something. The chapter of the reference manual on the Ethernet MAC is very poorly written and edited, and makes a number of internally contradictory statements about (among other things) how interrupts are enabled/disabled. It turns out, the mechanism is inconsistent. The key is this diagram, which is _not_ in the interrupts chapter, but instead sandwiched inexplicably between two unrelated register definitions at the end of the chapter:

Image

The important thing is the MMCTXIS/MMCRXIS signals, which -- despite having the same visual shape as the other interrupt sources, such that you could easily overlook them in a scan of the diagram -- differ in a critical way: they are OR'd, without any interrupt enable signal. In other words, the MMC interrupt sources are enabled by default. They can be _masked_ at the MMC, but you have to write a register to achieve this. These are the only interrupt sources in the Ethernet block with this behavior.

Part of the reason we missed this is that the chapter cries wolf: it describes literally every interrupt source, separately, as "enabled by default," when none of them are... except this one.

If you fail to mask these interrupt sources, but enable the Ethernet interrupt, you will eventually start taking interrupts when one or more of the counters reach their halfway mark (`2**31 - 1`). You must explicitly clear these interrupt sources that you didn't enable, or the interrupt will endlessly recur. Our driver was not expecting this, and did not contain code to clear the interrupt sources.

This sort of interrupt storm manifests, in a Hubris application, as the task responsible for the interrupt failing to yield. It takes the steps it _believes_ are necessary to resolve the interrupt source, and then asks the kernel to re-enable the interrupt...which will just cause it to receive another interrupt notification when it next tries to sleep. So (in the case of an Oxide compute sled) a lot of the system kept working, but lower-priority tasks were starved of CPU, and in particular all network clients were out of commission.

The fix: while initializing the Ethernet MAC, be sure to write to the `MMC_RX_INTERRUPT_MASK` and `MMC_TX_INTERRUPT_MASK` registers, putting 1 bits in the appropriate positions to disable all these interrupt sources. Or, make sure that your interrupt service routine clears the MMC interrupt sources -- this is somewhat involved:
1. Read the MACISR register to determine that MMCTXIS or MMCRXIS is set.
2. Then read `ETH_MMC_TX_INTERRUPT` (resp. `ETH_MMC_RX_INTERRUPT`) to see which counter is complaining.
3. Then issue a read to the appropriate counter, which clears the interrupt source as a side effect.

Or, more likely, just read every counter. It's probably faster anyway.

If anyone's trying to debug this in their own application, note that the `ETH_MMC_CONTROL` register contains a `CNTPRST` (counter preset) bit that will _force_ the counters to their interrupt thresholds, making the issue very easy to reproduce once you know where to look.

cbiffle:
The MMC (management counters) turns out to have some interrupt sources that are enabled by default, unlike every other part of the block. We haven't noticed this because they only trigger after 2**31 events have happened. When triggered, the fact that we don't clear the interrupt source would cause the net task to fail to yield CPU to any lower priority task.

See the issue for more details.

Fixes #2304.

### DIFF
--- a/drv/stm32h7-eth/src/lib.rs
+++ b/drv/stm32h7-eth/src/lib.rs
@@ -209,6 +209,41 @@ impl Ethernet {
                 .set_bit()
         });
 
+        // Disable potential interrupt sources coming from the MMC (management
+        // counters) block. These sources are not otherwise gated, and are on by
+        // default until masked. Failing to mask these interrupt sources will
+        // start producing interrupts much later, as counters hit their halfway
+        // point (generally about 2**31). The interrupt code below is not
+        // prepared to handle these counter interrupts.
+        mac.mmc_rx_interrupt_mask.write(|w| {
+            // Safety: The stm32h7 0.14 crate doesn't model RXLPITRCIM (bit 27),
+            // but it's defined in the reference manual and required to disable
+            // an interrupt, so, no safety implications.
+            unsafe {
+                w.bits(1 << 27);
+            }
+
+            w.rxcrcerpim().set_bit();
+            w.rxalgnerpim().set_bit();
+            w.rxucgpim().set_bit();
+            w.rxlpiuscim().set_bit();
+            w
+        });
+        mac.mmc_tx_interrupt_mask.write(|w| {
+            // Safety: The stm32h7 0.14 crate doesn't model TXLPITRCIM (bit 27)
+            // but it's defined in the reference manual and required to disable
+            // an interrupt, so, no safety implications.
+            unsafe {
+                w.bits(1 << 27);
+            }
+
+            w.txscolgpim().set_bit();
+            w.txmcolgpim().set_bit();
+            w.txgpktim().set_bit();
+            w.txlpiuscim().set_bit();
+            w
+        });
+
         #[cfg(feature = "vlan")]
         {
             // If we're in VLAN mode, we _only_ support VLAN operation.


drv/stm32h7-eth/src/lib.rs
```diff
@@ -209,6 +209,41 @@ impl Ethernet {
...
+        // Disable potential interrupt sources coming from the MMC (management
+        // counters) block. These sources are not otherwise gated, and are on by
+        // default until masked. Failing to mask these interrupt sources will
+        // start producing interrupts much later, as counters hit their halfway
+        // point (generally about 2**31). The interrupt code below is not
+        // prepared to handle these counter interrupts.
+        mac.mmc_rx_interrupt_mask.write(|w| {
+            // Safety: The stm32h7 0.14 crate doesn't model RXLPITRCIM (bit 27),
+            // but it's defined in the reference manual and required to disable
+            // an interrupt, so, no safety implications.
```
hawkw, line 221: Cool. :)

drv/stm32h7-eth/src/lib.rs
```diff
@@ -209,6 +209,41 @@ impl Ethernet {
                 .set_bit()
         });
 
+        // Disable potential interrupt sources coming from the MMC (management
+        // counters) block. These sources are not otherwise gated, and are on by
+        // default until masked. Failing to mask these interrupt sources will
+        // start producing interrupts much later, as counters hit their halfway
+        // point (generally about 2**31). The interrupt code below is not
+        // prepared to handle these counter interrupts.
```
hawkw, line 217: Might we also want to acknowledge and otherwise ignore those interrupts, as a defensive programming thing?

drv/stm32h7-eth/src/lib.rs
```diff
@@ -209,6 +209,41 @@ impl Ethernet {
                 .set_bit()
         });
 
+        // Disable potential interrupt sources coming from the MMC (management
+        // counters) block. These sources are not otherwise gated, and are on by
+        // default until masked. Failing to mask these interrupt sources will
+        // start producing interrupts much later, as counters hit their halfway
+        // point (generally about 2**31). The interrupt code below is not
```
hawkw, line 216: Do we know what happens when these counters overflow? Does masking the IRQs also protect us from whatever happens if they do?

hawkw (approved):
This looks good to me, great catch! I left some non-blocking questions.

drv/stm32h7-eth/src/lib.rs
cbiffle: I don't think so -- I would be opening to noticing them and panicking, maybe. The case we'd be defending against would be someone removing this code, basically.

drv/stm32h7-eth/src/lib.rs
cbiffle: They just wrap, unless you set the bit asking them to saturate instead.

PR #2312, opened 2025-12-03
Support Cosmo CPUs in SBRMI validation

mkeeter:
This should fix racktest failures of the form
```
component U1/SBRMI (CPU via SB-RMI) is "failed", not "present"
```

(reported in https://github.com/oxidecomputer/mfg-troubleshooting/issues/171#issuecomment-3606985546)

Tested on `cosmo-os`; after this change, we are currently clean on ~opsec~ `humility validate`

### DIFF
--- a/drv/i2c-devices/src/sbrmi.rs
+++ b/drv/i2c-devices/src/sbrmi.rs
@@ -343,7 +343,8 @@ impl Validate<Error> for Sbrmi {
         let sbrmi = Sbrmi::new(device);
         let rev = sbrmi.read_reg(Register::Revision)?;
 
-        // Support either Gimlet or Cosmo CPUs
+        // Support either Gimlet or Cosmo CPUs, documented in the AMD PPR docs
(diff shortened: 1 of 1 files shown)


drv/i2c-devices/src/sbrmi.rs
```diff
@@ -343,6 +343,7 @@ impl Validate<Error> for Sbrmi {
         let sbrmi = Sbrmi::new(device);
         let rev = sbrmi.read_reg(Register::Revision)?;
 
-        Ok(rev == 0x10)
+        // Support either Gimlet or Cosmo CPUs
+        Ok(rev == 0x10 || rev == 0x21)
```
labbott: Is there a document number we can cite here for future us?

labbott (approved):

drv/i2c-devices/src/sbrmi.rs
mkeeter: Sure, done

Issue #2309, opened 2025-12-01
ereport: hubris task panicked/faulted


hawkw:
In production environments, it is currently somewhat difficult to become aware of Hubris task panics and faults. While MGS can ask the SP to list task dumps as part of the API for reading dumps, this requires that the control plane (or faux-mgs user) proactively ask the SP whether it has any record of panicked tasks, rather than recording panics as they occur. In particular, if the SP resets, the whole system is removed or loses power, or the SP is no longer reachable over the management network, no record exists that a panic occurred. We should probably generate an ereport when a task panics to increase the likelihood that there is some record of panicking tasks.

This is somewhat more complex to implement than other ereports, primarily because panicked tasks must be reported by the supervisor (`jefe`), which runs at a higher priority than `packrat`, which is responsible for recording ereports. Thus, we must have a scheme where `jefe` exposes an API to collect ereports, and sends a notification to `packrat` when it would like its ereports collected. See #2127 for details on that.

This scheme implies that ereports from the supervisor must be collected _asynchronously_, which is different from ereports from tasks running at lower priorities than packrat. Generally, a task which wishes to record an ereport performs a synchronous IPC to `packrat`, which either succeeds or fails, and then the reporting task goes along with its day. For supervisor ereports, though, `jefe` must send `packrat` a notification saying "hey, I have some ereports", and then, when `packrat` is scheduled, it *eventually* comes around and collect them. During the time between when `jefe` notifies `packrat` and when `packrat` eventually manages to come snarf up the ereports, additional tasks may have panicked. This means that `jefe` must have some capacity to buffer multiple ereports or otherwise be capable of generating more than one ereport if multiple tasks have panicked before `packrat` comes asking.

Some things we would probably want these ereports to include are:

- Name of the task that panicked/faulted (obvious, but worth mentioning)
- Generation of the task when it panicked/faulted
- [The type of fault (and potentially fault details)](https://github.com/oxidecomputer/hubris/blob/c75dfa8a062ae51d0986d4d4834d55863073e243/sys/abi/src/lib.rs#L309-L356)
- If this was a panic, he panic message. @cbiffle points out to me that [the Hubris userlib will only record the first 128 bytes of a UTF-8 panic message](https://github.com/oxidecomputer/hubris/blob/c75dfa8a062ae51d0986d4d4834d55863073e243/sys/userlib/src/lib.rs#L1416-L1431), which is very helpful for right-sizing the ereport buffer.
[...]

PR #2313, opened 2025-12-03
Add `read_panic_message` kipc

### LINKED ISSUE #2311: want some kinda way for Jefe to actually know things about panic messages
when a task is compiled with the `"panic-messages"` feature flag enabled on the `userlib` crate, the first 128 bytes of the panic message are formatted into a buffer in `userlib`'s panic handler, here:
https://github.com/oxidecomputer/hubris/blob/12519139d5bad10e8d18dfe4e2e0918a685d42a4/sys/userlib/src/lib.rs#L1521-L1554

this is passed to `sys_panic`, which turns the `&[u8]` into a pointer and length pair and passes them to the panic syscall stub:
https://github.com/oxidecomputer/hubris/blob/12519139d5bad10e8d18dfe4e2e0918a685d42a4/sys/userlib/src/lib.rs#L1096-L1098

...which just leaves them lying on the floor where it found them and does nothing! hooray!

https://github.com/oxidecomputer/hubris/blob/12519139d5bad10e8d18dfe4e2e0918a685d42a4/sys/userlib/src/lib.rs#L1103-L1152

currently, when a task panics, Jefe dumps its memory, and the panic message is present in the dump. `humility` just happens to be aware of which registers the panic message base/length go in, and special-cases reading it out of the buffer: https://github.com/oxidecomputer/humility/blob/098ca7e6bfe6c2044eea18a250d139f093355d86/cmd/tasks/src/lib.rs#L761-L775
so, when you use humility to inspect a dump, you get to know about the panic message instead of having to hex dump the whole thing and look for some UTF-8 bytes that look vaguely panic-message-y. this part is good.

however, neither Jefe nor the kernel nor the dump-agent are actually aware of this thing your people call a "panic message". what is that? all the dump agent sees is some memory regions which have some bytes in them. in particular, the kernel-side handler for the panic syscall (`explicit_panic()`), does not actually read any caller-saved registers at all, which makes this "the easiest syscall!":
https://github.com/oxidecomputer/hubris/blob/12519139d5bad10e8d18dfe4e2e0918a685d42a4/sys/kern/src/syscalls.rs#L795-L801

this is fine because it's Humility that actually understands  that some of those bytes might happen to be a message of some form which is related to the panic in some kind of a way. at least, this was fine until #2309 came along and wanted to include the panic message in an ereport. that requires Jefe to be able to programmatically find a "panic message", turn it into a string, and stuff it into a Different Buffer that gets given to packrat. currently there is no way to do this, and i would like to rectify that.

as i see it, there are basically a couple obvious ways to do this:

1. make the dump-agent somehow search for a panic message on the stack whilst its recording a task dump, which seems hard and scary, or, 
2. make the kernel actually read the panic message buffer base and length in the `explicit_panic` syscall handler and stuff them in the `FaultInfo`, and then have Jefe get that back via `kipc::read_task_status` and pass it to the dump-agent, so that it can remember to also put the panic message someplace where we can actually find it later.

the second one seems like the better option giving that those args are already on the stack anyway and scanning the whole dump for a panic message is both hard and flaky, a metaphor which definitely makes sense and is normal. okay, great. so let's do that.

hawkw:
Currently, there is no way to programmatically access the panic message of a task which has faulted due to a Rust panic fron within the Hubris userspace. This branch adds a new `read_panic_message` kipc that copies the contents of a panicked task's panic message buffer into the caller. If the requested task has not panicked, this kipc returns an error indicating this. This is intended by use by supervisor implementations or other tasks which wish to report panic messages from userspace.

I've also added a test case that exercises this functionality.

Fixes #2311

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -2,7 +2,6 @@ target = "thumbv7em-none-eabihf"
 chip = "../../chips/stm32h7"
 memory = "memory-large.toml"
 stacksize = 896
-fwid = true
 
 [kernel]
 name = "cosmo"
@@ -181,7 +180,7 @@ max-sizes = {flash = 131072, ram = 16384 }
 stacksize = 2600
 start = true
 task-slots = ["sys", "i2c_driver", {spi_front = "spi3_driver"}, "jefe", "packrat", "auxflash", "spartan7_loader", "hf"]
-uses = ["mmio_sequencer", "mmio_info", "mmio_espi"]
+uses = ["mmio_sequencer", "mmio_info", "mmio_espi", "mmio_debug_ctrl"]
 notifications = ["timer", "vcore", "seq-irq"]
 
 [tasks.ignition_flash]
--- a/app/gemini-bu/app.toml
+++ b/app/gemini-bu/app.toml
@@ -3,7 +3,6 @@ target = "thumbv7em-none-eabihf"
 board = "gemini-bu-1"
 chip = "../../chips/stm32h7"
 stacksize = 896
-fwid = true
 
 [kernel]
 name = "gemini-bu"
--- a/app/gimlet/base.toml
+++ b/app/gimlet/base.toml
@@ -2,7 +2,6 @@ target = "thumbv7em-none-eabihf"
 chip = "../../chips/stm32h7"
 memory = "memory-large.toml"
 stacksize = 896
-fwid = true
 
 [kernel]
 name = "gimlet"
--- a/app/gimletlet/app-mgmt.toml
+++ b/app/gimletlet/app-mgmt.toml
@@ -3,7 +3,6 @@ target = "thumbv7em-none-eabihf"
 board = "gimletlet-1"
 chip = "../../chips/stm32h7"
 stacksize = 1024
-fwid = true
 
 [kernel]
 name = "gimletlet"
--- a/app/gimletlet/base-gimletlet2.toml
+++ b/app/gimletlet/base-gimletlet2.toml
@@ -5,7 +5,6 @@ memory = "memory-large.toml"
 stacksize = 896
 epoch = 0
 version = 0
-fwid = true
 
 [kernel]
 name = "gimletlet"
--- a/app/grapefruit/base.toml
+++ b/app/grapefruit/base.toml
@@ -5,7 +5,6 @@ memory = "memory-large.toml"
 stacksize = 896
 epoch = 0
 version = 0
-fwid = true
 
 [mmio]
 peripheral-region = "fmc_nor_psram_bank_1"
--- a/app/lpc55xpresso/app.toml
+++ b/app/lpc55xpresso/app.toml
@@ -4,7 +4,6 @@ board = "lpcxpresso55s69"
 chip = "../../chips/lpc55"
 stacksize = 1024
 image-names = ["a", "b"]
-fwid = true
 
 [kernel]
 name = "lpc55xpresso"
--- a/app/medusa/base.toml
+++ b/app/medusa/base.toml
@@ -2,7 +2,6 @@ target = "thumbv7em-none-eabihf"
 chip = "../../chips/stm32h7"
 memory = "memory-large.toml"
 stacksize = 896
-fwid = true
 
 [kernel]
 name = "medusa"
--- a/app/minibar/base.toml
+++ b/app/minibar/base.toml
@@ -2,7 +2,6 @@ target = "thumbv7em-none-eabihf"
 chip = "../../chips/stm32h7"
 stacksize = 896
 memory = "memory-large.toml"
-fwid = true
 
 [kernel]
 name = "minibar"
--- a/app/oxide-rot-1/app-dev.toml
+++ b/app/oxide-rot-1/app-dev.toml
@@ -10,7 +10,6 @@ stacksize = 1024
 image-names = ["a", "b"]
 epoch = 0
 version = 0
-fwid = true
 
 [kernel]
 name = "oxide-rot-1"
--- a/app/oxide-rot-1/app.toml
+++ b/app/oxide-rot-1/app.toml
@@ -7,7 +7,6 @@ stacksize = 1024
 image-names = ["a", "b"]
 epoch = 0
 version = 0
-fwid = true
 
 [kernel]
 name = "oxide-rot-1"
--- a/app/psc/base.toml
+++ b/app/psc/base.toml
@@ -2,7 +2,6 @@ target = "thumbv7em-none-eabihf"
 chip = "../../chips/stm32h7"
 memory = "memory-large.toml"
 stacksize = 896
-fwid = true
 
 [kernel]
 name = "psc"
--- a/app/rot-carrier/app.toml
+++ b/app/rot-carrier/app.toml
@@ -6,7 +6,6 @@ stacksize = 1024
 image-names = ["a", "b"]
 epoch = 0
 version = 0
-fwid = true
 
 [kernel]
 name = "rot-carrier"
--- a/app/sidecar/base.toml
+++ b/app/sidecar/base.toml
@@ -2,7 +2,6 @@ target = "thumbv7em-none-eabihf"
 chip = "../../chips/stm32h7"
 stacksize = 896
 memory = "memory-large.toml"
-fwid = true
 
 [kernel]
 name = "sidecar"
--- a/build/xtask/src/config.rs
+++ b/build/xtask/src/config.rs
@@ -28,8 +28,6 @@ struct RawConfig {
     epoch: u32,
     #[serde(default)]
     version: u32,
-    #[serde(default)]
-    fwid: bool,
     memory: Option<String>,
     #[serde(default)]
     image_names: Vec<String>,
@@ -67,7 +65,6 @@ pub struct Config {
     pub epoch: u32,
     pub mmio: Option<MmioData>,
     pub version: u32,
-    pub fwid: bool,
     pub image_names: Vec<String>,
     pub signing: Option<RoTMfgSettings>,
     pub stacksize: Option<u32>,
@@ -262,7 +259,6 @@ impl Config {
             mmio,
             epoch: toml.epoch,
             version: toml.version,
-            fwid: toml.fwid,
             signing: toml.signing,
             stacksize: toml.stacksize,
             kernel: toml.kernel,
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -3,7 +3,6 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use std::collections::{BTreeMap, BTreeSet, HashMap, VecDeque};
-use std::ffi::OsStr;
 use std::fmt::Write as _;
 use std::fs::{self, File};
 use std::hash::{Hash, Hasher};
@@ -15,10 +14,8 @@ use std::process::{Command, Stdio};
 use anyhow::{anyhow, bail, Context, Result};
 use atty::Stream;
 use indexmap::IndexMap;
-use lpc55_rom_data::FLASH_PAGE_SIZE as LPC55_FLASH_PAGE_SIZE;
 use multimap::MultiMap;
 use path_slash::{PathBufExt, PathExt};
-use sha3::{Digest, Sha3_256};
 use zerocopy::IntoBytes;
 
 use crate::{
@@ -720,10 +717,6 @@ pub fn package(
             archive.overwrite()?;
         }
 
-        if cfg.toml.fwid {
-            write_fwid(&cfg, image_name, &flash, &archive_name)?;
-        }
-
         // Unzip the signed + caboose'd images into our build directory
         let archive = hubtools::RawHubrisArchive::load(&archive_name)
             .context("loading archive with hubtools")?;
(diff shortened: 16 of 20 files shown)


doc/kipc.adoc
```diff
@@ -332,7 +332,7 @@ This interface is primarily intended for testing interrupt handling, and can
 only be called by a task running as the supervisor. When this KIPC is called,
 an actual machine interrupt is triggered for the relevant hardware interrupt
 source. The kernel then responds to the interrupt using its normal
-interrupt-handling infrastructure, which dispatches a notification to the
```
hawkw, line 335: sorry, my editor decided to do this...

sys/kern/src/kipc.rs
```diff
@@ -507,3 +510,61 @@ fn find_faulted_task(
...
+        ..
+    } = tasks[index].state()
+    {
+        // Okay, good
+    } else {
+        return Err(UserError::Recoverable(
+            abi::ReadPanicMessageError::TaskNotPanicked as u32,
+            NextTask::Same,
+        ));
+    }
```
mkeeter: nit: if you want to use `let ...` statements consistently, you could write
```suggestion
    let TaskState::Faulted {
        fault: FaultInfo::Panic,
        ..
    } = tasks[index].state()
    else {
        return Err(UserError::Recoverable(
            abi::ReadPanicMessageError::TaskNotPanicked as u32,
            NextTask::Same,
        ));
    };
```

sys/kern/src/kipc.rs
```diff
@@ -507,3 +510,61 @@ fn find_faulted_task(
...
+    message: USlice<u8>,
+    response: USlice<u8>,
+) -> Result<NextTask, UserError> {
+    let index: u32 = deserialize_message(&tasks[caller], message)?;
+    let index = index as usize;
+    if index >= tasks.len() {
+        return Err(UserError::Unrecoverable(FaultInfo::SyscallUsage(
+            UsageError::TaskOutOfRange,
+        )));
+    }
```
mkeeter: Similarly, something like
```suggestion
    let Some(task) = tasks.get(index) else {
        return Err(UserError::Unrecoverable(FaultInfo::SyscallUsage(
            UsageError::TaskOutOfRange,
        )));
    };
```
(then use `task` below instead of indexing repeatedly)

sys/userlib/src/kipc.rs
```diff
@@ -162,3 +162,37 @@ pub fn software_irq(task: usize, mask: u32) {
...
+///    from `buf`, and contains the task's panic message as a sequence of
+///   UTF-8 bytes. Note that the slice may be empty, if the task has panicked
+///   but was compiled without panic messages enabled.
+/// - [`Err`]`(`[`ReadPanicMessageError::TaskNotPanicked`]`)` if the task is
+///   not currently faulted due to a panic.
+/// - [`Err`]`(`[`ReadPanicMessageError::BadPanicMessage`]`)` if the task has
+///   panicked but the panic message buffer is invalid to read from.
+pub fn read_panic_message(
+    task: usize,
+    buf: &mut [u8; 128],
```
mkeeter: Promote `const BUFSIZE: usize = 128;` out of the panic handler into a global constant (`PANIC_BUFSIZE`), then use it here?  It could even be public so others can size their buffers appropriately.

hawkw: good call, will do!

sys/kern/src/kipc.rs
```diff
@@ -507,3 +510,61 @@ fn find_faulted_task(
...
+    } else {
+        return Err(UserError::Recoverable(
+            abi::ReadPanicMessageError::TaskNotPanicked as u32,
+            NextTask::Same,
+        ));
+    }
+
+    let Ok(message) = tasks[index].save().as_panic_args().message else {
+        // XXX(eliza): should we have a whole error code for this, or just
+        // return length 0?
```
labbott: I like error codes if it means less thinking. Would returning length 0 conflict with the empty slice when panics are disabled?

hawkw: Yeah, the question here was really whether it was worth differentiating between "no panic message because the task was compiled it disabled" and "no panic message because the alleged buffer was not a valid ptr+len pair in normal memory". I think it is, since one of them is a bug, and the other isn't. Removed the comment.

doc/kipc.adoc
```diff
@@ -402,6 +402,47 @@ while let Some(fault) = kipc::find_faulted_task(next_task) {
...
+==== Preconditions
+
+`task_index` must be a valid task index for this system.
+
+==== Response
+
+[source,rust]
+----
+Result<&str, abi::ReadPanicMessageError>
+----
```
cbiffle, line 428:
```suggestion
----
Result<&[u8], abi::ReadPanicMessageError>
----
```

doc/kipc.adoc
```diff
@@ -402,6 +402,47 @@ while let Some(fault) = kipc::find_faulted_task(next_task) {
...
+message is invalid, this KIPC returns the
+`abi::ReadPanicMessageError::BadPanicMessage` response code. If the target task
+uses the panic handler provided by the `userlib` crate, this should not be
+possible. However, it may occur if a task calls the `sys_panic` syscall
+directly with invalid arguments.
+
+The panic message is truncated to the length of the response buffer. Since
+Hubris only stores the first 128 bytes of panic messages, a 128-byte response
+buffer will never result in further truncation.
+
```
cbiffle, line 446:
```suggestion

The panic message is normally UTF-8. However, because of this byte-based truncation, and because the panic message is coming from a failing task in the first place, be careful not to *assume* that the panic message is valid UTF-8. See the `<[u8]>::utf8_chunks` function for one possible method of handling the panic message safely.
```

sys/kern/src/kipc.rs
```diff
@@ -507,3 +510,68 @@ fn find_faulted_task(
...
+            UsageError::TaskOutOfRange,
+        )));
+    };
+
+    // Make sure the task is actually panicked.
+    let TaskState::Faulted {
+        fault: FaultInfo::Panic,
+        ..
+    } = task.state()
+    else {
```
cbiffle, line 533: huuuh, I've never seen a `let`/`else` used in a way that doesn't bind any names like this. I assume you prefer this over `if matches!`?

sys/kern/src/kipc.rs
```diff
@@ -507,3 +510,68 @@ fn find_faulted_task(
...
+        return Err(UserError::Recoverable(
```
cbiffle, line 534: Do you have any code that uses this API yet? I suspect this error is going to be `unwrap()`'d, in which case we might want to eliminate it and have the kernel fault the caller if they got the task state wrong. (Which is to say, switch this to `Unrecoverable`.)

FWIW all the other errors in this implementation look right to me.

sys/kern/src/kipc.rs
```diff
@@ -507,3 +510,68 @@ fn find_faulted_task(
...
+            NextTask::Same,
+        ));
+    };
+
+    // Note that if the panic was recorded by `userlib`'s panic handler, it will
+    // never exceed 128 bytes in length, and if the caller requested this kipc
+    // using the `userlib::ipc::read_panic_message()` wrapper, then the caller's
+    // buffer will always be exactly 128 bytes long. However, we can't rely on
+    // that here, as either task *could* be an arbitrary binary that wasn't
+    // compiled with the Hubris userlib, so we need to be safe regardless.
```
cbiffle, line 559: (agreed)

doc/kipc.adoc
```diff
@@ -402,6 +402,47 @@ while let Some(fault) = kipc::find_faulted_task(next_task) {
...
+If the requested task has panicked, but the buffer containing its panic
+message is invalid, this KIPC returns the
+`abi::ReadPanicMessageError::BadPanicMessage` response code. If the target task
+uses the panic handler provided by the `userlib` crate, this should not be
+possible. However, it may occur if a task calls the `sys_panic` syscall
+directly with invalid arguments.
+
+The panic message is truncated to the length of the response buffer. Since
+Hubris only stores the first 128 bytes of panic messages, a 128-byte response
+buffer will never result in further truncation.
```
cbiffle:
```suggestion
buffer will normally not result in further truncation.
```

(since tasks _can_ panic with large buffers, technically, it's just not easy)

cbiffle: It's slightly unfortunate that, in this API design, there is no way for a caller to _detect_ that the failed task's panic message was too long to fit in their response buffer and has been truncated.

I'm honestly not sure such a thing would be _useful,_ but I wanted to point it out.

doc/kipc.adoc
```diff
@@ -402,6 +402,47 @@ while let Some(fault) = kipc::find_faulted_task(next_task) {
...
+
+If the requested task is not currently in the faulted state due to a panic,
+this KIPC returns the `abi::ReadPanicMessageError::TaskNotPanicked` response code.
+
+If the requested task has panicked, but the buffer containing its panic
+message is invalid, this KIPC returns the
+`abi::ReadPanicMessageError::BadPanicMessage` response code. If the target task
+uses the panic handler provided by the `userlib` crate, this should not be
+possible. However, it may occur if a task calls the `sys_panic` syscall
+directly with invalid arguments.
```
cbiffle:
```suggestion
directly with invalid arguments, such as a slice pointing outside the task's memory.
```

sys/userlib/src/kipc.rs
```diff
@@ -162,3 +162,41 @@ pub fn software_irq(task: usize, mask: u32) {
         &[],
     );
 }
+
+/// Reads a task's panic message into the provided `buf`, if the task is
+/// panicked.
+///
+/// Note that Hubris only preserves the first [`PANIC_MESSAGE_MAX_LEN`] bytes of
```
cbiffle:
```suggestion
/// Note that Hubris normally only preserves the first [`PANIC_MESSAGE_MAX_LEN`] bytes of
```

sys/userlib/src/kipc.rs
```diff
@@ -162,3 +162,41 @@ pub fn software_irq(task: usize, mask: u32) {
...
+/// a task's panic message, and panic messages greater than that length are
+/// truncated. Thus, this function accepts a buffer of that length.
```
cbiffle: Just wanted to note that we could most likely loosen this later, if required, by changing this to a `&mut [u8]` and not breaking callers.

sys/userlib/src/kipc.rs
```diff
@@ -162,3 +162,41 @@ pub fn software_irq(task: usize, mask: u32) {
...
+///
+/// # Returns
+///
+/// - [`Ok`]`(&[u8])` if the task is panicked. The returned slice is borrowed
+///   from `buf`, and contains the task's panic message as a sequence of
+///   UTF-8 bytes. Note that the slice may be empty, if the task has panicked
```
cbiffle: Technically, nothing ensures that the contents of that slice are UTF-8, so I wouldn't advertise that. The panic message is _normally expected to be_ UTF-8 but it may be truncated in the middle of a multi-byte sequence, and the task is panicking so strictly speaking it could have stomped all over its RAM, producing garbage.

This is another case where seeing a caller might clarify the API design -- it might make sense to call `utf8_chunks` on the buffer _and return the iterator that results_ so that the caller can easily step over valid and invalid sections if desired. Or, callers may just byte-copy the result into CBOR and not parse it at all (in which case it's quite important that it not hit CBOR as UTF-8!).

test/test-suite/src/main.rs
```diff
@@ -1553,6 +1554,38 @@ fn test_irq_status() {
     assert_eq!(status, expected_status);
 }
 
+/// Tests that when a task panics, its panic message can be read via the `read_panic_message` kipc.
+fn test_read_panic_message() {
+    set_autorestart(false);
+
+    let mut buf = [0u8; 128];
```
cbiffle:
```suggestion
    let mut buf = [0u8; userlib::PANIC_MESSAGE_MAX_LEN];
```

test/test-suite/src/main.rs
```diff
@@ -1553,6 +1554,38 @@ fn test_irq_status() {
...
+        response.as_mut_bytes(),
+        &[],
+    );
+
+    let msg =
+        kipc::read_panic_message(ASSIST.get_task_index().into(), &mut buf)
+            .unwrap();
+    // it should look kinda like a panic message (but since the line number may
+    // change, don't make assertions about the entire contents of the string...
+    assert!(core::str::from_utf8(msg)
```
cbiffle: I'm not sure the `expect` here is right -- assuming the assistant panics with an ASCII-only message (which it likely does) you could still get this test to fail by running it in a deep path containing non-ASCII characters.

Since in the end you're only checking the prefix of the string, you can get the valid UTF-8 prefix using

`msg.utf8_chunks().next().unwrap().valid()`

(the `unwrap()` there is an unfortunate detail of the `utf8_chunks` implementation, which is defined as returning an iterator yielding at least one chunk... but the type system can't see that.)

doc/kipc.adoc
hawkw: good call, thanks!

sys/kern/src/kipc.rs
```diff
@@ -507,3 +510,68 @@ fn find_faulted_task(
...
+    } = task.state()
+    else {
+        return Err(UserError::Recoverable(
+            abi::ReadPanicMessageError::TaskNotPanicked as u32,
+            NextTask::Same,
+        ));
+    };
+
+    let Ok(message) = task.save().as_panic_args().message else {
+        return Err(UserError::Recoverable(
```
cbiffle, line 548: (In case anyone's curious, I was musing on how this could fail in practice. Because we're getting a `USlice<u8>`, it can't be misaligned. In practice, the only "invalid" combination that will trigger this specific check is: a slice that spans the end of the address space, such that `base + len` would overflow.)

hawkw: I should probably write that down here, huh?

test/test-suite/src/main.rs
hawkw: Ah, that's a good point, thanks!

sys/userlib/src/kipc.rs
hawkw: Yeah. So, I was planning to probably just encode the valid regions as a CBOR string in the caller which I haven't written yet. But, I had a vague sense that the KIPC API ought not make too many decisions about what the caller wants; I had kind of been considering having a higher-level `read_panic_message_utf8` or something that just gives you the valid UTF-8 regions. The sense I got from the other KIPC APIs is that they don't tend to do much beyond what's necessary to actually send and receive the IPC from the kernel, so I wanted to be consistent with that.

On the other hand, I suspect basically all callers are going to want to get a string out and not want anything that isn't UTF-8, so maybe this API _should_ just do the higher-level behavior.  I think you're starting to sell me on that.

sys/userlib/src/kipc.rs
hawkw: Yup, that was my thinking as well.

sys/kern/src/kipc.rs
hawkw: Hmm, I was imagining we might see this error in the event that a task responsible for collecting panic messages (i.e., `packrat`) gets pinged about a panic, but doesn't actually get to run before a timer in `jefe` elapses and the task gets restarted. In that case, the task would no longer have a panic message to share, and `packrat` (or whomever wants the panic message) has just missed its chance to read the panic. I would definitely _not_ want to panic packrat in that case, so I felt like this error ought to be handle-able without panicking.

If we _don't_ end up implementing that behavior in the supervisor, and instead make it wait to be informed that a panic message has been collected before restarting the faulted task, it might make more sense to make this case a fault for the caller. But, the approach we discussed in https://github.com/oxidecomputer/hubris/issues/2309#issuecomment-3607491107 made me feel inclined to go down the path of treating the restart cooldown in `jefe` as a timeout for collecting panic messages, so I was planning to do that (so that `packrat` or similar can't block other tasks from restarting indefinitely). And, even if _we_ did decide to make `jefe` wait for panic messages to be recorded before restarting a task, making this a handleable error allows more flexibility for other (theoretical) supervisor implementations to do other things.

sys/kern/src/kipc.rs
hawkw: Yeah, it just felt a little bit better than 
```rust
if !matches!(task.state(), TaskState::Faulted { fault: FaultInfo::Panic, .. })) {
    return ...;
}
```
for some vague aesthetic reason. But it should be equivalent.

sys/kern/src/kipc.rs
cbiffle: > Hmm, I was imagining we might see this error in the event that a task responsible for collecting panic messages (i.e., `packrat`) gets pinged about a panic, but doesn't actually get to run before a timer in `jefe` elapses and the task gets restarted. In that case, the task would no longer have a panic message to share, and `packrat` (or whomever wants the panic message) has just missed its chance to read the panic. I would definitely _not_ want to panic packrat in that case, so I felt like this error ought to be handle-able without panicking.

I find this convincing, I hadn't considered which task was likely to be calling this API.

Issue #2280, opened 2025-11-05
Cosmo might need a little thermal loop tuning


ericaasen:
When firing up a CPU synthetic load on a Group G (500W) processor on a Cosmo Rev B (with the good heatsink), the temperature on the processor overshoots a decent bit and then settles back down so we might need to be more aggressive in how we ramp the fans or what temperature we let the CPU idle at.

<img width="2045" height="1518" alt="Image" src="https://github.com/user-attachments/assets/e5cfe648-ce9c-485b-8ab2-6217eed26f7a" />

data from `dashboard`: 

[cosmo_b_500W_6.4T_drives_first_all_loadgen_then_cpu_loadgen_only.csv](https://github.com/user-attachments/files/23377107/cosmo_b_500W_6.4T_drives_first_all_loadgen_then_cpu_loadgen_only.csv)

As the filename implies, the first spike in CPU temperatures is from running

`svcadm enable *loadgen*`

The second spike is from running

`svcadm enable cpu-loadgen`

This is on the Cosmo Rev B behind `niles` which has a K.2 and an SP SWD.  The SWD is setup with an alias of `cosmo-b` and you can login through the K.2 using `ssh root@fe80::eaea:6aff:fe09:cafb%axf9`  The only reason it has a K.2 is because it doesn't seem to want to advertise its host interface through the T6 currently

ericaasen:
If you do want to use the cosmo, please check with @tomsteininger to make sure he's not running anything on there as he was looking into some `lmar` stuff

refugeesus:
So we saw the effects of this in a different scenario in the emc chamber this evening. Essentially, the CPU reaches some threshold temp and the fans seem to bounce between 50% and 100% rate in order to handle the spikes in temp. We will probably want to adjust the tuning here for our new processor. More details to follow in the EMC doc here: https://docs.google.com/document/d/1wRXduLGGjQZ6PgJ2-XIcGMdOv8yr13qGgDbq3osQ7zk/edit?tab=t.0#heading=h.f2o8c0fjvxpa

mkeeter:
There are a few things going on here.  The big issue is that the control loop is undertuned for the processor; in an ideal world, we should never enter the `Overheat` state

However, there are also things that could improve in our overheat handling.

[...]

PR #2320, opened 2025-12-10
Tweak cosmo thermal tuning

mkeeter:
This aims to address significant overshoot seen in #2280 

Here's the baseline performance when I load the system `svcadm enable cpu-loadgen` (testing on `niles`'s `cosmo-b`):

Screenshot 2025-12-10 at 4 51 48 PM

With the new parameters, the overshoot is much more limited:

Screenshot 2025-12-10 at 4 51 33 PM

There's probably more room for improvement, but this should prevent the worst of it.

### DIFF
--- a/task/thermal/src/bsp/cosmo_ab.rs
+++ b/task/thermal/src/bsp/cosmo_ab.rs
@@ -135,9 +135,9 @@ impl Bsp {
             // Based on experimental tuning!
             pid_config: PidConfig {
                 zero: 35.0,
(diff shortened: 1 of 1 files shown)


mkeeter:
A few more tests, checking stability as we toggle `cpu-loadgen` at 5 and 30s intervals

Screenshot 2025-12-11 at 9 00 52 AM

Screenshot 2025-12-11 at 9 05 16 AM

Issue #2321, opened 2025-12-11
Firmware mitigation options for BMR491 voltage droop defect


cbiffle:
Flex has proposed a firmware mitigation for existing BMR491 IBCs that have the intermittent voltage droop issue. I'm creating this tracking issue so we can document its implementation in the firmware (or our decision not to implement it, should we decide it's not right!).

The suggestion is, concretely,

- To disable the input undervoltage check by writing `VIN_OFF` to 0.  (All register names in this list are in the PMBus interface.)
- Send `VOUT_COMMAND` of 0x0060 (sets the ranges for the subsequent commands)
- Override `MAX_DUTY` to 0.95 / 0xF8EA
- Set `VOUT_UV_FAULT_LIMIT` to 0x0058
- Set `VOUT_UV_FAULT_RESPONSE` to 0x80 (disable, do not retry)

They then suggest using `STORE_USER_ALL` to make this persist.

My interpretation of their doc against the PMBus spec suggests that the fix is plausible. It should effectively disable the input undervoltage monitor, which makes some sense -- it's the input stage for that circuit that has the out-of-tolerance components on the BMR491. To compensate it enables VOUT undervoltage monitoring, which _should_ be less prone to false detection.

We don't appear to do much programming of the BMR491 online right now, so I think we'd need to do this in manufacturing on any defective BMR491s. For systems already in the field, we'd probably want to distribute a firmware update that applies the mitigation at startup. This raises the question of whether we can reliably detect a defective (or potentially defective) '491 in-system, which we need to double check.

@hawkw @ericaasen - did I miss things?

cbiffle:
I was staring at the "disable + do not retry" part of this recommendation, and I _think_ it's right for our context?

If we applied this mitigation, a droop in the 12V output would trigger an undervoltage shutdown of the BMR491, which Ignition _should_ see as a fault. Cycling at the upstream hotswap controller should still work, so it should be recoverable without manual intervention, unless there's an erratum in this area I've missed.

Thanks to @rmustacc for reminding me of some circuit details here.

cbiffle:
I have been assured that we ought to be able to detect a potentially-defective unit through PMBus by checking the revision code. So that's lovely.

In that case, the sequencer task is probably the right place to apply the mitigation on the compute sleds. We should be able to verify if it's been applied already or not, since the affected registers are read/write.

hawkw:
[...]

PR #2322, opened 2025-12-11
BMR491 flaw mitigation

cbiffle:
This applies the mitigation for the flaw in the BMR491 R1C parts, which can cause intermittent voltage droop that makes the server angry. The original fix was described in #2321, and then expanded on in RFD 630 (which is unfortunately not public, but I plan to update 2321 with the deets -- or you can read the code here).


drv/i2c-devices/src/bmr491.rs
```diff
@@ -73,6 +73,104 @@ impl Bmr491 {
...
+        // is supposed to be fixed in R1D, and in the event that an R1E is
+        // released, it will _probably_ remain fixed.
+        //
+        // But since we don't know that for sure, currently we're treating only
+        // the R1D revision as fixed. Applying the mitigation to fixed future
+        // IBCs should not be destructive, so we can fix the firmware when that
+        // day comes.
+        if rev.starts_with(b"R1D ") {
+            // Assume the mitigation is unnecessary.
+            return Ok(());
```
hawkw: It might be nice if we put something in the ringbuf indicating whether we've decided we have a revision that needs the mitigation or not, just so that we can have a record of what the driver decided to do someplace?

drv/i2c-devices/src/bmr491.rs
```diff
@@ -73,6 +73,104 @@ impl Bmr491 {
...
+        let current_vout_uv_fault_limit =
+            pmbus_read!(self.device, VOUT_UV_FAULT_LIMIT)?;
+        let current_vout_uv_fault_response =
+            pmbus_read!(self.device, VOUT_UV_FAULT_RESPONSE)?;
+
+        if current_vin_off.0 == 0
+            && current_vout_command.0 == 0x0060
+            && current_max_duty.0 == 0xF8EA
+            && current_vout_uv_fault_limit.0 == 0x0058
+            && current_vout_uv_fault_response.0 == 0x80
```
hawkw: nitpicky, probably unnecessary, but i wonder if we ought to make the mitigation `VIN_OFF`/`VOUT_COMMAND`/`MAX_DUTY`/`VOUT_UV_FAULT_LIMIT`/`VOUT_UV_FAULT_RESPONSE` values constants, and use them both in this check and when setting the registers later on.

that way, if we ever decide to change these values, we don't run the risk of accidentally changing the value we set but not the one we check for.

drv/i2c-devices/src/bmr491.rs
```diff
@@ -73,6 +73,104 @@ impl Bmr491 {
...
+        {
+            // The device configuration already reflects the mitigation.
+            return Ok(());
```
hawkw: if we ringbuf the driver's decisions as per my earlier comment, we should probably also say something when we believe the mitigation has already been applied.

drv/i2c-devices/src/bmr491.rs
```diff
@@ -73,6 +73,104 @@ impl Bmr491 {
...
+        // retry options are all wrong for our use case, they can cause it to
+        // power itself on and off repeatedly before stopping, or just stop; we
+        // choose the latter.)
+        pmbus_write!(
+            self.device,
+            VOUT_UV_FAULT_RESPONSE,
+            VOUT_UV_FAULT_RESPONSE::CommandData(0x80)
+        )?;
+
+        pmbus_write!(self.device, STORE_USER_ALL)?;
```
hawkw: I thought RFD 630 [determined we were *not* going to do this](https://rfd.shared.oxide.computer/rfd/0630#det-persist)? Or has this PR not yet been updated to match?

drv/i2c-devices/src/bmr491.rs
cbiffle: Done in newest version.

drv/i2c-devices/src/bmr491.rs
cbiffle: Ringbuf added.

drv/i2c-devices/src/bmr491.rs
cbiffle: PR updated to reflect RFD determinations.

drv/i2c-devices/src/bmr491.rs
cbiffle: Fixed in the latest revision.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -477,6 +479,42 @@ impl<S: SpiServer + Clone> ServerImpl<S> {
...
+        {
+            use drv_i2c_devices::bmr491::{
+                Bmr491, ExternalInputVoltageProtection,
+            };
+
+            let dev = i2c_config::devices::bmr491_u431(I2C.get_task_id());
+            let driver = Bmr491::new(&dev, 0);
+
+            // Gimlet provides external undervoltage protection that is better
+            // than what we'd get from the 491, so we rely on it.
```
hawkw: somewhat nitpicky, sorry: this comment feels a bit unclear about what "it" is that we're relying on. perhaps:
```suggestion
            // than what we'd get from the 491, so we rely on that instead.
```

drv/cosmo-seq-server/src/main.rs
```diff
@@ -166,6 +168,41 @@ fn main() -> ! {
     let packrat = Packrat::from(PACKRAT.get_task_id());
     read_vpd_and_load_packrat(&packrat, I2C.get_task_id());
 
+    //
+    // Apply the configuration mitigation on the BMR491, if required. This
+    // is an external device access and may fail. We'll attempt it thrice
+    // and then allow boot to continue.
```
hawkw, line 195: 'twould be nice if this comment referenced/linked to RFD 630, so that the reader is not left asking what "the mitigation" is. i realize that the Flex document itself is confidential, but we can at least say the name of the RFD so that Oxide employees can know where to look for more info.

drv/sidecar-seq-server/src/main.rs
```diff
@@ -832,6 +834,38 @@ fn main() -> ! {
...
+                Err(e) => {
+                    ringbuf_entry!(Trace::Bmr491MitigationFailed(e));
+                    if i == RETRIES - 1 {
+                        // TODO squawk here
+                    } else {
+                        continue;
+                    }
+                }
+            }
+        }
```
hawkw: take it or leave it: since we repeat identical retry code everywhere that we apply the mitigation...perhaps this should just be part of the "apply mitigation" function? though, perhaps not, due to the ringbuf stuff...we could give the driver its own ringbuf, but i'm not sure if that's a great idea.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -102,6 +102,8 @@ enum Trace {
     },
     UnexpectedInterrupt,
     CPUPresent(bool),
+    Bmr491MitigationFailed(drv_i2c_devices::bmr491::Error),
+    Bmr491MitigationOk(drv_i2c_devices::bmr491::MitigationResult),
```
hawkw: we may wish to add `#[count(children)]` to one or both of these --- especially the `MitigationResult` so that it shows up in the counters even if it falls off the end of the ringbuf. adding counters to the I2C error would make the counters table a lot bigger, but the `MitigationResult` should be fine since it's only 3 variants.
```suggestion
    Bmr491MitigationFailed(#[count(children)] drv_i2c_devices::bmr491::Error),
    Bmr491MitigationOk(#[count(children)] drv_i2c_devices::bmr491::MitigationResult),
```

drv/sidecar-seq-server/src/main.rs
```diff
@@ -57,6 +57,8 @@ enum Trace {
     MainboardControllerSha(u32),
     InvalidMainboardControllerId(u32),
     ExpectedMainboardControllerChecksum(u32),
+    Bmr491MitigationFailed(drv_i2c_devices::bmr491::Error),
+    Bmr491MitigationOk(drv_i2c_devices::bmr491::MitigationResult),
```
hawkw: this ringbuf doesn't have counters currently, but it feels like it could be a good idea to add them, at least so that these variants get counted --- since they occur pretty early in initialization, i think we want to have a counter for it so that we can see if it's ever happened, even if it's fallen off the end of the ringbuf.

drv/i2c-devices/src/bmr491.rs
```diff
@@ -73,6 +99,143 @@ impl Bmr491 {
...
+    ///
+    /// - To retain some degree of brownout protection, they recommend instead
+    ///   enabling the _output_ undervoltage monitor.
+    ///
+    /// This function will check to see if the mitigation is already applied, or
+    /// if the device reports an unaffected revision. In either case, it will
+    /// leave the settings untouched.
+    ///
+    /// The output undervoltage monitor will be configured _only_ on request
+    /// (when the `include_vout_detect` parameter is `true`). This is because
```
hawkw: i think this was previously a bool but you've changed it to an enum, we should update the comment here?

drv/i2c-devices/src/bmr491.rs
```diff
@@ -73,6 +99,143 @@ impl Bmr491 {
...
+                )?;
+                // And configure it to shut off the IBC without retry on fault.
+                // (The retry options are all wrong for our use case, they can
+                // cause it to power itself on and off repeatedly before
+                // stopping, or just stop; we choose the latter.)
+                pmbus_write!(
+                    self.device,
+                    VOUT_UV_FAULT_RESPONSE,
+                    VOUT_UV_FAULT_RESPONSE::CommandData(
+                        EXPECTED_UV_FAULT_RESPONSE
```
hawkw, line 309: goofy turbo nit: the "expected" naming feels a bit weird here --- it's "expected" when we're checking for the presence of the mitigation above, but here, it's _desired_...perhaps "MITIGATION_...`? or you could just ignore me and i won't be upset. :)

drv/cosmo-seq-server/src/main.rs
cbiffle: I was kinda trying to embed the information required into the comments, since RFDs are not public, so referencing them from firmware isn't super helpful for customers etc. But I can mention it.

drv/sidecar-seq-server/src/main.rs
cbiffle: I deliberately did not give the driver a ringbuf, because I'm not convinced it will be properly linker-GC'd in tasks using i2c but not the bmr491.

But, factoring the retry loop out seems reasonable.

drv/cosmo-seq-server/src/main.rs
cbiffle: Please take a look at the revised failed/result types in the newest version, I'd welcome advice on how best to counters them given that they're now more complex.

drv/sidecar-seq-server/src/main.rs
cbiffle: Factored out in the newest version.

drv/i2c-devices/src/bmr491.rs
cbiffle: This and the other comment you noted are fixed in the latest commit.

drv/cosmo-seq-server/src/main.rs
hawkw: :woman_shrugging: I was thinking more for future Oxide employees who might be reading this and not otherwise know where to look for the background knowledge. But, it's not a big deal.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -166,7 +171,57 @@ fn main() -> ! {
...
+                    (r.failures, r.last_failure, true)
+                }
+                Err(e) => {
+                    ringbuf_entry!(Trace::Bmr491MitigationFailed(e));
+                    (e.retries, Some(e.last_cause), false)
+                }
+            };
+
+        if failures > 0 {
+            // Report the failure even if we eventually succeeded.
```
hawkw, line 215: good call!

drv/cosmo-seq-server/src/main.rs
```diff
@@ -166,7 +171,57 @@ fn main() -> ! {
...
+            try_send_ereport(
+                &packrat,
+                &mut ereport_buf[..],
+                EreportClass::Bmr491MitigationFailure,
+                EreportKind::Bmr491MitigationFailure {
+                    failures,
+                    last_cause: last_cause.unwrap(),
```
hawkw: nit, take it or leave it: seems to me you could avoid the `unwrap()` by making the `if` condition be "`last_cause` is `Some`"?
```suggestion
        if let Some(last_cause) = last_cause {
            // Report the failure even if we eventually succeeded.
            try_send_ereport(
                &packrat,
                &mut ereport_buf[..],
                EreportClass::Bmr491MitigationFailure,
                EreportKind::Bmr491MitigationFailure {
                    failures,
                    last_cause,
```
since i can't seem to see a path in which that would be `Some` and `failures == 0`?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -387,10 +445,26 @@ struct ServerImpl {
 
 const EREPORT_BUF_LEN: usize = 256;
 
+#[derive(microcbor::Encode)]
+pub enum EreportClass {
+    #[cbor(rename = "hw.pwr.bmr491.mitfail")]
```
hawkw, line 454: at one point, @rmustacc requested that we try to follow a scheme in which all `hw.*` ereports include a refdes field, and all `hw.pwr.*` ereports would include a rail, could we add those to the body of the ereport here?

drv/cosmo-seq-server/src/main.rs
```diff
@@ -387,10 +445,26 @@ struct ServerImpl {
 
 const EREPORT_BUF_LEN: usize = 256;
```
hawkw: now that we are using `microcbor` for the BMR491 ereport, it would be nice if this was calculated from the max length of the ereports we actually send here rather than hardcoded at 256B, though i will admit this is my fault for neglecting to update this task to use microcbor for the V_core_ ereports...

drv/gimlet-seq-server/src/main.rs
```diff
@@ -155,6 +158,9 @@ enum Trace {
         retries_remaining: u8,
     },
     StartFailed(#[count(children)] SeqError),
+    EreportSent(usize),
+    EreportLost(usize, packrat_api::EreportWriteError),
+    EreportTooBig,
```
hawkw, line 161: can we include the `EreportClass` enum in (ideally) all of these, or at least the `EreportLost` and `EreportTooBig` entries? I am now remembering that the main reason I ended up representing classes as a separate enum rather than making the whole ereport an enum with the class as the `variant_id` was so that we could track _which_ ereport class was too big for the buffer in these ringbuf entries.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -229,6 +237,11 @@ pub(crate) enum EreportKind {
         pwr_good: Option<bool>,
         pmbus_status: PmbusStatus,
     },
+    Bmr491MitigationFailure {
```
hawkw, line 238: the same comment i left in the Cosmo sequencer applies here as well: Robert asks that we include refdes and rail strings in these ereports.

hawkw: also, perhaps we ought to consider factoring out the `Bmr491MitigationFailure` message body into something that can live with the shared code, and make this just be
```suggestion
    Bmr491MitigationFailure(bmr491::MitigationFailureEreport)
```
since we intend to always use the same fields for the ereport regardless of whether it comes from a Gimlet, Cosmo, or Sidecar?

drv/sidecar-seq-server/src/main.rs
```diff
@@ -120,6 +125,25 @@ const TIMER_INTERVAL: u64 = 1000;
...
+    Bmr491MitigationFailure,
+}
+
+#[derive(microcbor::EncodeFields)]
+pub(crate) enum EreportKind {
+    Bmr491MitigationFailure {
+        failures: u32,
+        last_cause: drv_i2c_devices::bmr491::MitigationFailureKind,
+        succeeded: bool,
+    },
```
hawkw, line 144: same notes as everywhere else

drv/cosmo-seq-server/src/main.rs
hawkw: i'm happy to come back and change the `vcore` module to use `microcbor` in a separate PR, though.

drv/i2c-devices/src/bmr491.rs
```diff
@@ -65,6 +65,60 @@ impl From<Error> for ResponseCode {
...
+///
+/// This throws away the specific PMBus error codes because they've already
+/// thrown away most of the useful information, due to the I2C API's flattening
+/// of errors. This should still be useful.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, microcbor::Encode)]
+pub enum MitigationFailureKind {
+    RevisionReadFailed,
+    StateReadFailed,
+    UpdateFailed,
+}
```
hawkw, line 127: these are gonna get `microcbor::Encode`d as some fairly longish strings...i wonder if we could shave off a few bytes by e.g. dropping the "...Failed" from their serialized representations, since, ya'know, the value is going to be a field named `last_failure` anyway?

drv/sidecar-seq-server/src/main.rs
```diff
@@ -57,6 +57,8 @@ enum Trace {
     MainboardControllerSha(u32),
     InvalidMainboardControllerId(u32),
     ExpectedMainboardControllerChecksum(u32),
+    Bmr491MitigationFailed(drv_i2c_devices::bmr491::MitigationFailure),
+    Bmr491MitigationOk(drv_i2c_devices::bmr491::MitigationResult),
```
hawkw: i still think the ringbuf should be counted, so that we can see if these were _ever_ recorded even if a bunch of other stuff has happened since...

drv/i2c-devices/src/lib.rs
hawkw, line 1: OKAY RIGHT YES THE PMBUS MACROS. I do think I agree that the new code makes a _lot_ more sense than the old `match foo(match bar {` situation which was really quite mind-bending.

drv/cosmo-seq-server/src/main.rs
labbott: If you decide to keep the constant, can you move it closer to the erport above? Took me a minute to figure out where the constant was

drv/cosmo-seq-server/src/main.rs
cbiffle: Thanks! It introduced kind of a lot of complexity.

drv/cosmo-seq-server/src/main.rs
cbiffle: Hmm, I like it -- both are kind of implicit but this one has one fewer unwrap.

I initially returned them together as an Option<(NonZero, Cause)> but that caused other issues in the end.

drv/gimlet-seq-server/src/main.rs
cbiffle: I was going to pose the opposite question -- I copied these ringbuf values from Cosmo, and I was wondering if we actually want them. Given that we don't really expect ereports to fill up in practice, and packrat tracks this on its side if they do, do we need to duplicate this across all ereport generators in a format that we won't proactively notice (ringbufs)?

drv/sidecar-seq-server/src/main.rs
cbiffle: I'm game, but I would like advice on the specifics, as I have never used this before.

drv/cosmo-seq-server/src/main.rs
cbiffle: I also don't understand why this one server has a hardcoded constant and the others compute theirs, but I think it's because it relies on the legacy serde support? I treated fixing that as out of scope for this change.

Happy to move it though.

drv/cosmo-seq-server/src/main.rs
cbiffle: I don't fully understand the suggestion, can you point me to an existing example? Refdes is easy enough to include.

It's not clear what rail this applies to -- the BMR491 is responsible for many rails, and the mitigation affects all of them directly or indirectly.

drv/cosmo-seq-server/src/main.rs
cbiffle: So I guess we could, like, `max` 256 with the computed value, to take the serde reports into account?

drv/gimlet-seq-server/src/main.rs
cbiffle: Do we have any examples of factoring out an ereport like this that I could refer to? I hesitate to include this in i2c-devices because it seems very Oxide-compute-server-specific.

drv/i2c-devices/src/bmr491.rs
cbiffle: I can shorten the names. Do you think I should also shorten `last_failure`? Maybe something like `cause`? I included the word `last` because I was trying to make it clear that this is just the _final_ one (since including a variable length list of failures seemed like overengineering), but it could probably go?

drv/cosmo-seq-server/src/main.rs
cbiffle: @hawkw and I spoke and I'm going to ensure that all these reports include a refdes, because we can do that unambiguously, but we don't think there's a relevant rail name.

drv/sidecar-seq-server/src/main.rs
cbiffle: Alright, @hawkw and I spoke, and to count these I need to disassemble them -- both of these types are structs that include several metadata fields, including counts, that don't make sense to count.

I have rejiggered the ringbuf approach entirely in the latest version.

drv/i2c-devices/src/bmr491.rs
```diff
@@ -12,14 +12,24 @@ use crate::{
 };
 use drv_i2c_api::*;
 use pmbus::commands::*;
+use ringbuf::{ringbuf, ringbuf_entry};
 use userlib::units::*;
 
+#[derive(Copy, Clone, PartialEq, counters::Count)]
+pub enum Trace {
+    None,
```
hawkw, line 20: if we're gonna keep counters on this ringbuf, let's put a 
```suggestion
    #[count(skip)]
    None,
```
on this to save 8 bytes?

drv/i2c-devices/src/bmr491.rs
```diff
@@ -65,6 +75,62 @@ impl From<Error> for ResponseCode {
...
+/// This throws away the specific PMBus error codes because they've already
+/// thrown away most of the useful information, due to the I2C API's flattening
+/// of errors. This should still be useful.
+#[derive(
+    Copy, Clone, Debug, Eq, PartialEq, microcbor::Encode, counters::Count,
+)]
+pub enum MitigationFailureKind {
+    RevisionReadFailed,
+    StateReadFailed,
+    UpdateFailed,
```
hawkw, line 126: take it or leave it: i still think we should probably
```suggestion
    #[cbor(rename = "RevRead")]
    RevisionReadFailed,
    #[cbor(rename = "StateRead")]
    StateReadFailed,
    #[cbor(rename = "Update")]
    UpdateFailed,
```

to try and claw back a couple bytes of ereport buffer...

drv/i2c-devices/src/bmr491.rs
```diff
@@ -12,14 +12,24 @@ use crate::{
...
 use userlib::units::*;
 
+#[derive(Copy, Clone, PartialEq, counters::Count)]
+pub enum Trace {
+    None,
+    MitigationFailed(#[count(children)] MitigationFailureKind),
+    MitigationApplied(#[count(children)] MitigationAction),
+}
+
+ringbuf!(Trace, 8, Trace::None);
```
hawkw, line 25: nit: why 8? it kinda seems like this will only ever have a maximum of `RETRIES` entries in it...

actually, i wonder if we can use the constant for that here, or if the macro requires it to be a literal...

hawkw, line 25: in order to have counters, we have to declare the ringbuf with `counted_ringbuf!`. but, i actually think we don't need to, since we've moved this to its own ringbuf and we're not gonna keep stuffing more entries into it for other events --- it really should only ever have between 1-3 entries depending on retries, so the value of the counters for ensuring that we have some memory that something happened even if it's fallen off the end of the ringbuf isn't really important here.

we could probably just remove all the counters derives now that the ringbuf lives in this crate...

drv/i2c-devices/src/bmr491.rs
```diff
@@ -65,6 +75,62 @@ impl From<Error> for ResponseCode {
...
+
+/// Information about a failure to apply a mitigation.
+///
+/// This throws away the specific PMBus error codes because they've already
+/// thrown away most of the useful information, due to the I2C API's flattening
+/// of errors. This should still be useful.
+#[derive(
+    Copy, Clone, Debug, Eq, PartialEq, microcbor::Encode, counters::Count,
+)]
+pub enum MitigationFailureKind {
```
hawkw, line 120: naming nitpick, feel free to ignore me:
```suggestion
pub enum MitigationError {
```
feels a lot less of a mouthful than `FailureKind`...

drv/i2c-devices/src/bmr491.rs
```diff
@@ -73,6 +139,188 @@ impl Bmr491 {
...
+        }
+
+        // Read out the affected registers to see if we've already applied the
+        // mitigation.
+        let current_vin_off = pmbus_read!(self.device, VIN_OFF)
+            .map_err(|_| MitigationFailureKind::StateReadFailed)?;
+        let basic_mitigation_ok = current_vin_off.0 == 0;
+
+        // This is 95%.
+        const EXPECTED_MAX_DUTY: u16 = 0xEAF8;
```
hawkw, line 243: should we perhaps note that this is _not_ the value for `MAX_DUTY` suggested by Flex, which is both (a) confidential and (b) incorrect? :upside_down_face:

drv/gimlet-seq-server/src/main.rs
hawkw: we do not presently. part of why i haven't yet is because i haven't had a good crate to put them in --- i think what i may end up doing is putting together some kinda "sequencer ereports" crate since there are starting to be a bunch of ereport types shared across the various `drv-$WHATEVER-seq` tasks. but i'm happy to do that myself later when i do some other ereport refactoring.

drv/gimlet-seq-server/src/main.rs
hawkw: packrat does track when an ereport doesn't fit in _its_ buffer (the `EreportLost` case) but it doesn't tell you what class the ereport was. and, it doesn't track the case of "the ereport did not fit in _our_ encoding buffer", which is a programmer error. i'm open to not recording these ringbuf entries, as it would certainly simplify the code somewhat, but...idk. getting rid of them would definitely let us not have the `EreportClass` enum as a separate thing which would make a bunch of stuff nicer. i think we're better off leaving that as something for me to mess with later, though, rather than refactoring everything as part of the BMR491 mitigation change.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -160,13 +164,70 @@ struct StateMachineStates {
...
+const EREPORT_BUF_LEN: usize = {
+    let n = microcbor::max_cbor_len_for!(
+        task_packrat_api::Ereport<EreportClass, EreportKind>,
+    );
+    // someday, we will have const max.
+    if n < 256 {
+        256
+    } else {
+        n
+    }
```
hawkw, line 176: this is also (as you are aware) a temporary workaround for the fact that the `drv_cosmo_seq::vcore` module still records its ereports using `serde` due to an apparent oversight or screwed up rebase on my part (sorry). when i come around and do some ereport refactoring after this PR merges, we can just get rid of this...

drv/i2c-devices/src/bmr491.rs
cbiffle: So, I considered that, but I know people often insert new records in ringbufs and I wanted to make sure this didn't start losing mitigation info. It's conservative of me.

drv/i2c-devices/src/bmr491.rs
cbiffle: Hmm, I feel like that leaves us with the ambiguous pair `MitigationFailure` and `MitigationError`. I could switch it to `MitigationError` and `MitigationErrorKind` if you'd prefer. (The `ErrorKind` is by analogy to `io::ErrorKind`.)

drv/i2c-devices/src/bmr491.rs
cbiffle: Agreed, I simply forgot to do this with all the other churn.

drv/i2c-devices/src/bmr491.rs
cbiffle: Counters removed.

hawkw:
[ci failure](https://github.com/oxidecomputer/hubris/actions/runs/21047502477/job/60525591167?pr=2322) appears to be a network burp.

drv/i2c-devices/src/bmr491.rs
hawkw: :woman_shrugging: it's fine as is, i don't actually care about this.

jordanhendricks:
@cbiffle: @Lydia-Sylla from the mfg sw team [asked](https://github.com/oxidecomputer/mfg-quality/issues/140#issuecomment-4112852136) about whether it would make sense to remove the test that checks for the IBC droop in mfg, which this PR has mitigated. I wrote my [thoughts](https://github.com/oxidecomputer/mfg-quality/issues/140#issuecomment-4129183680) in the oxidecomputer/mfg-quality#140 issue, and there I asserted that it's worthwhile to keep a mfg test as a regression test for the hubris mitigation. I've realized after thinking about it a bit more, that I don't have the technical context to confidently make the assertion that a regression test is useful. What do you think?

PR #2332, opened 2025-12-30
Allow reading persistent component slot via mgs

### LINKED ISSUE #2331: Need means of getting persistent versus current active slots
The host phase 1 image has a notion of which slot should be selected on cold power up which is stored in a sector of the host SPI flash rom. The faux-mags `ComponentSetActiveSlot` allows this to be changed persistently or transiently.  However, the `ComponentGetActiveSlot` does not have the ability to select between the persistent and non-persistent option. Current implementations of the host logic only return the current in-memory active slot which may have diverged from the persistent one. It's not clear to me what the cleanest way to get at this is while honoring the MGS compatibility needs.

luqmana:
Along with https://github.com/oxidecomputer/management-gateway-service/pull/476 fixes #2331.

### DIFF
--- a/task/control-plane-agent/src/update/host_flash.rs
+++ b/task/control-plane-agent/src/update/host_flash.rs
@@ -6,8 +6,8 @@ use super::{common::CurrentUpdate, ComponentUpdater};
 use crate::mgs_handler::{BorrowedUpdateBuffer, UpdateBuffer};
 use core::ops::Range;
 use drv_hf_api::{
-    HfDevSelect, HfError, HfPersistentData, HfProtectMode, HostFlash,
-    PAGE_SIZE_BYTES, SECTOR_SIZE_BYTES,
+    HfDevSelect, HfError, HfProtectMode, HostFlash, PAGE_SIZE_BYTES,
+    SECTOR_SIZE_BYTES,
 };
 use gateway_messages::{
     ComponentUpdatePrepare, HfError as GwHfError, SpComponent, SpError,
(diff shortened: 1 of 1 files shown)


task/control-plane-agent/src/update/host_flash.rs
```diff
@@ -75,13 +75,26 @@ impl HostFlashUpdate {
...
+            .map(Self::dev_to_slot)
+            .map_err(|err| SpError::ComponentOperationFailed(err as u32))
+    }
+
+    pub(crate) fn persistent_slot(&self) -> Result<u16, SpError> {
+        match self.task.get_persistent_data() {
+            Ok(HfPersistentData { dev_select }) => {
+                Ok(Self::dev_to_slot(dev_select))
+            }
+            Err(HfError::NoPersistentData) => Ok(0),
```
labbott: Maybe a comment explaining why it's okay to fall back to slot 0?

luqmana: Added a comment but maybe it's worth just passing through the error so we don't mask failures to set a persistent slot?

labbott: returning the error would be my preference, I had assumed this was needed for some automatic testing or edge case where `0` made sense

luqmana: Ok, updated!

labbott (approved):
LGTM thanks for doing this work

PR #2341, opened 2026-01-05
ereport: Task faulted/panicked

### LINKED ISSUE #2309: ereport: hubris task panicked/faulted

hawkw:
Depends on #2313, #2350, #2358
Fixes #2309 

It's currently somewhat difficult to become aware of Hubris task panics and other task faults in a production environment. While MGS can ask the SP to list task dumps as part of the API for reading dumps, this requires that the control plane (or faux-mgs user) proactively ask the SP whether it has any record of panicked tasks, rather than recording panics as they occur. Therefore, we should have a proactive notification from the SP indicating that task faults have occurred. 

This commit adds code to `packrat` for producing an ereport when a task has faulted. This could eventually be used by the control plane to trigger dump collection and produce a service bundle. In addition, it will provide a more permanent record that a task faulted at a particular time, even if the SP that contains the faulted task is later reset or replaced with an entirely different SP. This works using an approach similar to the one described by @cbiffle in [this comment][1]. There's a detailed description of how this works [in the module-level RustDoc for `ereport.rs` in Packrat][2].

The ereports that come out of this thing look like this:

```console
eliza@hekate ~/Code/oxide/hubris $ faux-mgs --interface eno1np0 --discovery-addr '[fe80::0c1d:deff:fef0:d922]:11111' ereports
Jan 15 10:27:41.370 INFO creating SP handle on interface eno1np0, component: faux-mgs
Jan 15 10:27:41.372 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d922%2]:11111, interface: eno1np0, socket: control-plane-agent, component: faux-mgs
restart ID: 4e54b7f1-e13a-d9bb-709a-c7e863d64a64
count: 4

ereports:
0x1: {
    "hubris_task_gen": Number(0),
    "lost": Null,
}

0x2: {
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(378010),
    "k": String("hubris.fault.panic"),
    "msg": String("panicked at task/ereportulator/src/main.rs:158:9:\nim dead lol"),
    "v": Number(0),
}

0x3: {
    "by": Object {
        "gen": Number(0),
        "task": String("jefe"),
    },
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("user_leds"),
    "hubris_uptime_ms": Number(382914),
    "k": String("hubris.fault.injected"),
    "v": Number(0),
}

0x4: {
    "by": Object {
        "gen": Number(0),
        "task": String("jefe"),
    },
    "hubris_task_gen": Number(1),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(388215),
    "k": String("hubris.fault.injected"),
    "v": Number(0),
}
```

[1]: https://github.com/oxidecomputer/hubris/issues/2309#issuecomment-3607491107
[2]: https://github.com/oxidecomputer/hubris/blob/1e2b121fcd165dcdbcbdfdc69c36d35520dfa954/task/packrat/src/ereport.rs#L15-L95


task/packrat/src/ereport.rs
```diff
@@ -341,13 +368,205 @@ impl EreportStore {
...
+                encoder.str("k")?.str("hubris.fault.panic")?;
+                encoder.str("msg")?;
+                match kipc::read_panic_message(
+                    usize::from(task.0),
+                    self.panic_buf,
+                ) {
+                    Ok(msg_chunks) => {
+                        encoder.begin_str()?;
+                        for chunk in msg_chunks {
+                            let chunk = chunk.valid();
```
cbiffle: This currently has the effect of _skipping_ any invalid bytes with no indication that they were ever there. Consider concatenating a unicode replacement character as a placeholder for any invalid section, or something.

hawkw: good call, fixed: https://github.com/oxidecomputer/hubris/pull/2341/commits/f4e57131308fc6b32f55b0591a92caf0fa84c3d8

hawkw:
Thinking about things a bit more, there's some more changes I think I want to make here before it's really ready to land. In particular:

- Currently, we've changed the fault cooldown behavior in Jefe to just _always_ wait 50 ms between restarts, rather than only doing so when a task has not been running for at least that long between two subsequent faults (see https://github.com/oxidecomputer/hubris/blob/b31df4359ffc4f06136474fee952e32f9466b34e/task/jefe/src/main.rs). This means that there's now always 50 ms of latency for all task restarts. This is to give Packrat time to generate an ereport, but it feels a bit not great.

  I think we should be doing a somewhat more complex thing here. We should probably implement the approach that @cbiffle described in https://github.com/oxidecomputer/hubris/issues/2309#issuecomment-3607491107, and add a way for Packrat to let Jefe know it has finished generating an ereport for a fault. That way, we can possibly reduce the latency for restarts a bit by saying "we will always give Packrat *up to* 50ms to produce a fault report, but if it finishes before then _and_ the task has already been running for a while, we will restart it sooner".
- Currently, if all faulted tasks have already been restarted by the time Packrat actually processes the "task faulted" notification, Packrat just does nothing.[^1] We should maybe fix this by having packrat do some kind of "some task probably faulted but I couldn't figure out which one" ereport, so that it's not totally lost.

Personally, I think we should definitely do the second point here (some kind of "task faults may have occurred" ereport) before merging this PR. I'm on the fence about whether the first point (reducing restart latency) is worth doing now or not. It's a bit more complexity in Jefe...

@cbiffle, any thoughts?

[^1]: Well, it ringbufs about it, but in production, that's equivalent to "doing nothing".

task/packrat/src/ereport.rs
```diff
@@ -395,25 +424,294 @@ impl EreportStore {
...
+            //
+            // 1. The fault notification was spurious (in practice, this
+            //    probably means someone is dorking around with Hiffy and sent
+            //    a fake notification just to mess with us...)
+            // 2. Tasks faulted, but we were not scheduled for at least 50ms
+            //    after the faults occurred, and Jefe has already restarted
+            //    them by the time we were permtited to run.
+            //
+            // We should probably record some kind of ereport about this.
+        }
```
hawkw: i still wanna figure out what i need to put in the ereport in this case --- what class should it be, etc. `hubris.fault.maybe_faults` or something weird like that.

It's also a bit strange because the function for recording an ereport in the ereport ringbuffer _requires_ a task ID as part of the `insert` function. For all the other ereports, I've used the ID of the task that faulted for that field, rather than the ID of Packrat (who is actually generating the ereport) or Jefe (who is spiritually sort of responsible for reporting it in some vibes-based way); this felt like the right thing in general. However, when the ereport just says "some task _may_ have faulted", I'm not totally sure what ID I want to put in here, since I don't want to incorrectly suggest that Jefe or Packrat has faulted...hmm...

cbiffle: I think taskID of Packrat and a distinguishing class would be fine.

cbiffle:
So I think the "max of 50ms" interplay between Jefe and Packrat sounds promising, and doesn't seem like _too much_ additional supervisor complexity -- particularly while crashdumps are still in Jefe. If we want to reduce complexity, I'd start there.

I agree that having a "whoops" ereport if packrat finds no faulted tasks would be useful. As one possible alternative... and I'm not sure if this is a good idea or not... Jefe could buffer the faulted taskIDs and provide packrat with a way to collect them... we could then say "this specific task fell over but the system was too loaded for me to say why exactly". TaskIDs are a lot smaller than the full Fault record.

That said, packrat is by nature typically just one priority level under Jefe, so it should be able to respond in a timely fashion in most cases. The thing most likely to starve it is ... crash dumps.

hawkw:
> I agree that having a "whoops" ereport if packrat finds no faulted tasks would be useful. As one possible alternative... and I'm not sure if this is a good idea or not... Jefe could buffer the faulted taskIDs and provide packrat with a way to collect them... we could then say "this specific task fell over but the system was too loaded for me to say why exactly". TaskIDs are a lot smaller than the full Fault record.

Yeah, I've also wondered about doing that; it might be a good idea. We could also do a fixed-size array of `hubris_num_tasks::NUM_TASKS` counters or some such.

hawkw:
> > I agree that having a "whoops" ereport if packrat finds no faulted tasks would be useful. As one possible alternative... and I'm not sure if this is a good idea or not... Jefe could buffer the faulted taskIDs and provide packrat with a way to collect them... we could then say "this specific task fell over but the system was too loaded for me to say why exactly". TaskIDs are a lot smaller than the full Fault record.
> 
> Yeah, I've also wondered about doing that; it might be a good idea. We could also do a fixed-size array of `hubris_num_tasks::NUM_TASKS` counters or some such.

Actually, upon thinking about this a bit more, there _is_ actually a scheme where we don't need to add a new IPC to Jefe at all. Instead, we could just do something where Packrat stores an array of the last seen generation number of each task index. When Packrat is notified of faults, it can scan each task's current generation and compare it to the last one it saw to check if the task has faulted.

Here's my attempt at doing that, which is both conceptually quite elegant and implementationally somewhat disgusting: https://github.com/oxidecomputer/hubris/compare/eliza/fault-ereport...eliza/fault-counts#diff-48cf874f5ac8432941e2ba390792b33a94f9aea18dd933bbdb105cd23b93c9ee

cbiffle:
> Actually, upon thinking about this a bit more, there _is_ actually a scheme where we don't need to add a new IPC to Jefe at all. Instead, we could just do something where Packrat stores an array of the last seen generation number of each task index. When Packrat is notified of faults, it can scan each task's current generation and compare it to the last one it saw to check if the task has faulted.

I _almost_ suggested that, actually. My concern is mostly theoretical -- that it can't guarantee that it's a _fault_ that restarted the task. Yeah, currently, tasks mostly restart due to faults, but that's not necessarily inherent.

But for now it's basically equivalent I think?

hawkw:
> > Actually, upon thinking about this a bit more, there _is_ actually a scheme where we don't need to add a new IPC to Jefe at all. Instead, we could just do something where Packrat stores an array of the last seen generation number of each task index. When Packrat is notified of faults, it can scan each task's current generation and compare it to the last one it saw to check if the task has faulted.
> 
> I _almost_ suggested that, actually. My concern is mostly theoretical -- that it can't guarantee that it's a _fault_ that restarted the task. Yeah, currently, tasks mostly restart due to faults, but that's not necessarily inherent.
> 
> But for now it's basically equivalent I think?

After a bit more thinking, I'm thinking about going back to an approach where we ask Jefe to send us a list of fault counters explicitly, rather than looking at generations. This is mostly for the reason @cbiffle points out: a task _can_ also explicitly ask to be restarted without faulting (though I'm not sure if anything in our production images actually uses this capability). It has a couple other advantages, though: it's a bit quicker for Packrat to do (one IPC to Jefe rather than `NUM_TASKS` syscalls), and it lets us use a bigger counter than the `u8` generation number, which reduces the likelihood that the counter will wrap around and end up at the same value it was last time Packrat checked, missing the fault.

On the other hand, this would mean that we can no longer uphold the property that "Packrat never makes IPC requests to other tasks", which is documented in a few places. I think an infallible IPC to the supervisor is probably safe, but I'm not sure if we're comfortable violating that property for any reason...

hawkw:
> > > Actually, upon thinking about this a bit more, there _is_ actually a scheme where we don't need to add a new IPC to Jefe at all. Instead, we could just do something where Packrat stores an array of the last seen generation number of each task index. When Packrat is notified of faults, it can scan each task's current generation and compare it to the last one it saw to check if the task has faulted.
> > 
> > 
> > I _almost_ suggested that, actually. My concern is mostly theoretical -- that it can't guarantee that it's a _fault_ that restarted the task. Yeah, currently, tasks mostly restart due to faults, but that's not necessarily inherent.
> > But for now it's basically equivalent I think?
> 
> After a bit more thinking, I'm thinking about going back to an approach where we ask Jefe to send us a list of fault counters explicitly, rather than looking at generations. This is mostly for the reason @cbiffle points out: a task _can_ also explicitly ask to be restarted without faulting (though I'm not sure if anything in our production images actually uses this capability). It has a couple other advantages, though: it's a bit quicker for Packrat to do (one IPC to Jefe rather than `NUM_TASKS` syscalls), and it lets us use a bigger counter than the `u8` generation number, which reduces the likelihood that the counter will wrap around and end up at the same value it was last time Packrat checked, missing the fault.
> 
> On the other hand, this would mean that we can no longer uphold the property that "Packrat never makes IPC requests to other tasks", which is documented in a few places. I think an infallible IPC to the supervisor is probably safe, but I'm not sure if we're comfortable violating that property for any reason...

OKAY NEVERMIND IT TURNS OUT I WAS ACTUALLY SUPER WRONG ABOUT THIS AND WE ACTUALLY DO NEED A DEDICATED NOTION OF A "FAULT COUNTER". Using a "restart count" (whether generations from refresh_task_id or the full-size 32-bit generation from the kernel) won't work, because the restart count/generation is incremented _when the task is restarted_. If Packrat responds to a Jefe notification in a timely manner and tries to handle the fault _before_ the task has been fully restarted, it actually will appear to have _not faulted_, because the generation count hasn't been incremented yet...and that's the entire time window for which we can read an entire panic message and fault details. So that wrecks it. Never mind. It also doesn't work in situations such as one where the task is held by Hiffy and doesn't get restarted.

I'm going to go back to the Jefe fault counter IPC, since that would actually do the right thing, and I don't feel like making Packrat a client of the supervisor is the end of the world.

hawkw:
current thing works:

```console
eliza@hekate ~/Code/oxide/hubris $ faux-mgs --interface eno1np0 --discovery-addr '[fe80::0c1d:deff:fef0:d922]:11111' ereports
Jan 14 10:00:29.534 INFO creating SP handle on interface eno1np0, component: faux-mgs
Jan 14 10:00:29.535 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d922%2]:11111, interface: eno1np0, socket: control-plane-agent, component: faux-mgs
restart ID: 83369fba-36bc-4759-7fb9-0d58d92d014a
count: 1

ereports:
0x1: {
    "hubris_task_gen": Number(0),
    "lost": Null,
}


eliza@hekate ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.panicme
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.11s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Ereportulator.panicme`
Ereportulator.panicme() => Err(<server died; its new ID is 1>)

eliza@hekate ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- jefe -f user_leds
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.11s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- jefe -f user_leds`
humility: successfully changed disposition for user_leds

eliza@hekate ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- jefe -f ereportulator
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.12s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- jefe -f ereportulator`
humility: successfully changed disposition for ereportulator

eliza@hekate ~/Code/oxide/hubris $ faux-mgs --interface eno1np0 --discovery-addr '[fe80::0c1d:deff:fef0:d922]:11111' ereports
Jan 14 10:00:55.075 INFO creating SP handle on interface eno1np0, component: faux-mgs
Jan 14 10:00:55.076 INFO initial discovery complete, addr: [fe80::c1d:deff:fef0:d922%2]:11111, interface: eno1np0, socket: control-plane-agent, component: faux-mgs
restart ID: 83369fba-36bc-4759-7fb9-0d58d92d014a
count: 4

ereports:
0x1: {
    "hubris_task_gen": Number(0),
    "lost": Null,
}

0x2: {
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(46439),
    "k": String("hubris.fault.panic"),
    "msg": String("panicked at task/ereportulator/src/main.rs:158:9:\nim dead lol"),
    "v": Number(0),
}

0x3: {
    "by": Object {
        "gen": Number(0),
        "task": String("jefe"),
    },
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("user_leds"),
    "hubris_uptime_ms": Number(52888),
    "k": String("hubris.fault.injected"),
    "v": Number(0),
}

0x4: {
    "by": Object {
        "gen": Number(0),
        "task": String("jefe"),
    },
    "hubris_task_gen": Number(1),
    "hubris_task_name": String("ereportulator"),
    "hubris_uptime_ms": Number(62089),
    "k": String("hubris.fault.injected"),
    "v": Number(0),
}
```

task/jefe/src/main.rs
```diff
@@ -456,7 +442,7 @@ impl idol_runtime::NotificationHandler for ServerImpl<'_> {
 
                 // If we're aware that this task is in a fault state (or waiting
                 // in timeout), don't bother making a syscall to enquire.
-                let TaskState::Running { started_at } = &status.state else {
+                let TaskState::Running = &status.state else {
```
cbiffle: So at this point I would _strongly_ encourage using `if` and `!=`, just because this formulation is now pretty weird.

task/jefe/src/main.rs
```diff
@@ -485,30 +471,33 @@ impl idol_runtime::NotificationHandler for ServerImpl<'_> {
...
-                        status.state = TaskState::Timeout { restart_at };
-                        self.deadline = self.deadline.min(restart_at);
-                        self.any_tasks_in_timeout = true;
-                    } else {
-                        // Stand it back up immediately
-                        kipc::reinit_task(fault_index, true);
-                        status.state = TaskState::Running { started_at: now };
-                    }
+                    // Put it into timeout.
+                    let restart_at = now.wrapping_add(MIN_RUN_TIME);
```
cbiffle: I think you've changed the semantics of `MIN_RUN_TIME` here to actually be `MIN_RESTART_DELAY` -- was that deliberate?

task/packrat/src/ereport.rs
```diff
@@ -395,25 +474,433 @@ impl EreportStore {
...
+        // made some more room in our buffer and could potentially record a
+        // previously held fault.
+        //
+        // In either case, begin by asking Jefe for the current task fault
+        // counters.
+        self.jefe.read_fault_counts(&mut self.fault_count_buf);
+        let mut nfaulted: usize = 0;
+        let mut nreported: usize = 0;
+
+        // Who farted^H^H^H^H^H^Hfaulted?
```
cbiffle: You can save some keystrokes there with `^W`

task/jefe/src/main.rs
hawkw: Yeah, the name probably should have changed along with it. It's now intended to give packrat enough time to make a "good" ereport (i.e. reading fault details and (possibly) the panic message out of the faulted task) before it's restarted, as well as the debouncing of rapidly-restarting tasks

task/packrat/src/ereport.rs
```diff
@@ -395,25 +573,433 @@ impl EreportStore {
...
+        let mut encoder = minicbor::Encoder::new(cursor);
+        encoder.begin_map()?;
+        // Ereport version.
+        encoder.str("v")?.u32(0)?;
+
+        // If the task has faulted multiple times since the last ereport we
+        // generated for it, record the count.
+        if nfaults > 1 {
+            encoder.str("nfaults")?.u32(nfaults as u32)?;
+        }
```
hawkw, line 810: perhaps we really should produce a separate ereport for held faults and the most recent fault, so that we're not giving the implication that they're multiple instances of the same fault kind when we are able to read the cause of the most recent one...

task/packrat/src/ereport.rs
```diff
@@ -395,25 +573,440 @@ impl EreportStore {
...
+            // faults have occurred.
+            //
+            // This bit looks a little weird, but it'll make sense if you think
+            // about it...trust me!
+            let (nfaults, timestamp) =
+                match (new_count.cmp(&*count), *last_unrecorded_fault_time) {
+                    // If the new fault count is less than the current count,
+                    // then Jefe's counter has wrapped around. The number of
+                    // times the task has faulted is the difference between the
+                    // prior count and `u32::MAX`, plus the new fault count.
```
cbiffle:
```suggestion
                    // prior count and `usize::MAX`, plus the new fault count.
```

PR #2342, opened 2026-01-06
Add job checking for reproducibility issues

### LINKED ISSUE #2299: Add CI job(s) testing for reproducibility
As discussed in the product security sync last week, we have a customer who cares about reproducing the SP and RoT build. Before we fully commit to them that our build system is reproducible, we'll need to add a CI job building Hubris in a different enough way to give us confidence.

What to include in the build is still up in the air, but it should at least:

* Test with a non-deterministic filesystem (disorderfs).
* Test with different source code location.
* Test with different build time.
* Test with a different system toolchain (linker, potentially C compiler?) to the extend that we are willing to guarantee.

emilyalbini:
This PR adds a set of CI jobs to try and prevent nondeterminism in the build process. While they are not a guarantee that Hubris will be reproducible, it will hopefully catch most cases where we accidentally introduce nondeterminism.

There are two new jobs, **reproducible-a** and **reproducible-b**, building the same board (I arbitrarily picked cosmo-b, it can be anything). The first job uses the standard Ubuntu runner with little to no modification, while the second job tries to mutate the build environment as much as possible to introduce sources of nondeterminism. The code of the second job is heavily commented to describe each source and how we introduce it.

Inspiration for which variations to introduce was inspired by both past personal experience and [the documentation provided by the reproducible-builds](https://reproducible-builds.org/docs/env-variations/) project. There is a source I wanted to include (different UIDs) but didn't, and that's because running things as different users inside of GitHub Actions was breaking too many things to be worth it.

There is then a final new job, **reproducible-check**, that gets the artifacts from both builders and compares them with [diffoscope](https://diffoscope.org/), another tool by the reproducible-builds project. In case there are differences it will upload the diffoscope report as an artifact for analysis. This is [a run where I intentionally induced nondeterminism](https://github.com/oxidecomputer/hubris/actions/runs/20787386877), if you want to look at how things would be in case of a failure (there is a link to the diffoscope report in there).

If there are other sources of nondeterminism we want to try and inject let me know and I'll see how feasible it is to add them.

Fixes #2299

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,6 +120,186 @@ jobs:
         uses: actions/deploy-pages@v4
         id: deployment
 
+  # Hubris builds have to be reproducible, and we want to test that in CI. This job does a build of
+  # a few boards with the standard Ubuntu image and no interference, as the baseline to compire to.
+  reproducible-a:
+    name: Reproducibility (A, ${{ matrix.name }})
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    strategy:
+      matrix: &reproducibility-matrix
+        include:
+          - name: cosmo-b
+            image: default
+            app_toml: app/cosmo/rev-b.toml
+
+          - name: oxide-rot-1
+            image: a
+            app_toml: app/oxide-rot-1/app.toml
+    steps:
+      - name: Checkout the source code
+        uses: actions/checkout@v6
+
+      # We check explicitly to ensure the other job has a different one.
+      - name: Check that GCC is the system C toolchain
+        run: cc --version | grep -q "Free Software Foundation"
+
+      - name: Build a Hubris board
+        run: |
+          umask 0007 # We set the umask explicitly here to ensure the other job has a different one.
+          cargo xtask dist "$app_toml"
+        env:
+          app_toml: ${{ matrix.app_toml }}
+
+      - name: Upload the artifact to be later checked
+        uses: actions/upload-artifact@v6
+        with:
+          name: reproducible-${{ matrix.name }}-a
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -120,6 +120,165 @@ jobs:
         uses: actions/deploy-pages@v4
         id: deployment
 
+  # Hubris builds have to be reproducible, and we want to test that in CI. This job does a build of
+  # an arbitrary board (specifically cosmo-b, but it can be changed to any board) with the standard
+  # Ubuntu image and no interference, as the baseline to compire to.
```
labbott: Any chance we could also test an RoT board? There's enough disjoint code between SP and RoT and I could unfortunately see us accidentally sneaking in something non-reproducible to the RoT. Arguably there is some amount of disjoint code between e.g. cosmo and sidecar but that amount is smaller than SP vs RoT.

.github/workflows/ci.yml
```diff
@@ -120,6 +120,165 @@ jobs:
...
+      - name: Checkout the source code in the standard GitHub Actions directory
+        uses: actions/checkout@v6
+
+      # We run the Hubris build in a different directory, to ensure that paths are not hardcoded. We
+      # also use a very long path, as a reproducibility issue Emily found in the wild in the past
+      # was a rust-lang/rust test failing when built in a path that was too long.
+      #
+      # We also use disorderfs to randomize the ordering of listing directories, to catch code
+      # assuming directory entries are always returned in the same order.
+      - name: Prepare a custom build root directory with disorderfs
```
labbott, line 207: TIL disorderfs. This is lightly terrifying and also useful!

.github/workflows/ci.yml
emilyalbini: Done! Added both the RoT and Cosmo. It should also be trivial to add more in the future by just adding extra boards to the matrix.

PR #2343, opened 2026-01-06
packrat: excise archive ID from ereports

hawkw:
Currently, ereport metadata contains a `hubris_archive_id` field with the 8-byte image ID. Upstack software does not currently use this field for anything, at time of writing. @cbiffle would like to ensure that upstack software continues to not use it for anything in the future. Therefore, this commit removes it.

I'll replace it with a better identifier, such as the image name and Git SHA, in a subsequent commit.


cbiffle (approved):
<3

PR #2349, opened 2026-01-12
sidecar_seq: add A0 interlock on power up

Aaron-Hartwig:
We currently have an arbitrary sleep here which allows the Tofino's I2C interface to be working properly before we try to use it. I'm replacing that with an interlock with the FPGA which keeps us from proceeding until the FPGA sees Tpor2pcie as satisfied. While also effectively getting us to a place where the I2C interface will be operational, this also helps ensure that later when we release PERST it is safe to do so.

### DIFF
--- a/drv/sidecar-seq-server/src/tofino.rs
+++ b/drv/sidecar-seq-server/src/tofino.rs
@@ -134,7 +134,9 @@ impl Tofino {
                                             SeqError::SequencerTimeoutNotInA0
                                         )
                                     );
-                                    return Err(SeqError::SequencerTimeoutNotInA0);
+                                    return Err(
+                                        SeqError::SequencerTimeoutNotInA0,
+                                    );
                                 }
                                 tries += 1;
                                 ringbuf_entry!(Trace::TofinoNotInA0);


drv/sidecar-seq-server/src/tofino.rs
```diff
@@ -110,11 +110,31 @@ impl Tofino {
...
+                while !in_a0 {
+                    match self.sequencer.state() {
+                        Ok(state) => {
+                            in_a0 = state == TofinoSeqState::A0;
+                            if !in_a0 {
+                                ringbuf_entry!(Trace::TofinoNotInA0);
+                                hl::sleep_for(25);
+                            }
+                        }
+                        Err(_) => {
```
labbott, line 146: Are we going to be sad about losing the error here?

drv/sidecar-seq-server/src/tofino.rs
```diff
@@ -110,11 +110,31 @@ impl Tofino {
...
-                // ACK properly.
-                hl::sleep_for(50);
+                // The Tofino E&M Specification (TF2-DS2-003EA.pdf) specifies a
+                // need for at least 200ms between POR release and PERST
+                // release to allow for HW initialization and VDD to settle
+                // at the VID value. The FPGA tracks the timing here, so wait
+                // until it signifies this time has elapsed by officially
+                // entering the A0 power state.
+                let mut in_a0 = false;
+                while !in_a0 {
```
labbott, line 126: This leaves us spinning forever if we can't go into A0. Is this going to be debuggable enough if that happens?

drv/sidecar-seq-server/src/tofino.rs
Aaron-Hartwig: I don't think so. In general `FpgaError` is a lower-level problem than what normal application could would care about (it is more interesting during initial device configuration). To get to this point in sequencing the FPGA is in a happy state and we'd not typically expect any problems, and probably not anything we could act on (e.g. "something happened with SPI"). In practice the sequencer generally maps `FpgaErrors` to `SeqError::FpgaError` to indicate this sort of thing happened.

drv/sidecar-seq-server/src/tofino.rs
Aaron-Hartwig: 🤦🏻, I'll adjust this! Definitely don't want an infinite loop.

Aaron-Hartwig:
Updated with feedback and did several reboot loops to make sure this works as expected, and it does. We are in A0 after 9 entries of `TofinoNotInA0` each time (the new code times out after 12 tries), so 225ms. Given the prior VID fiddling and the FPGA's 250 ms total PoR to PCIe delay, that all feels right.

drv/sidecar-seq-server/src/tofino.rs
```diff
@@ -110,11 +110,47 @@ impl Tofino {
...
+                        Ok(state) => {
+                            in_a0 = state == TofinoSeqState::A0;
+                            if !in_a0 {
+                                if tries > MAX_TRIES {
+                                    ringbuf_entry_root!(
+                                        Trace::TofinoSequencerError(
+                                            SeqError::SequencerTimeout
+                                        )
+                                    );
+                                    return Err(SeqError::SequencerTimeout);
```
labbott: Just thinking out loud about how easy this is going to be to diagnose from looking at the ringbuf. I think it should be easy because the ringbuf entry should be unique and even if we've somehow looped the ringbuf the counter entries should still show up and give a nice pointer of THERE IS YOUR PROBLEM.

Aaron-Hartwig: I think it is pretty easy as it stands today, but could of course be made even more straight forward. There are two paths to getting to the `SequencerTimeout` error in this `power_up` routine.

The first is that we don't get the VID successfully, resulting in a ringbuf that looks something like:
```
1 TofinoPowerUp
1 TofinoVidAttempt(0x0)
1 TofinoNoVid
...
1 TofinoVidAttempt(0x7)
1 TofinoNoVid
1 SequencerTimeout
```

The second is what I introduce above and would look like:
```
1  TofinoPowerUp
1  TofinoVidAttempt(0x0)
1  SetVddCoreVout(Volts(0.893))
1  TofinoVidAck
12 TofinoNotInA0
1  SequencerTimeout
```

So I think the 12 `TofinoNotInA0` would give us a way to decipher this as-is, but I can add a more distinguishable `SequencerTimeout`-like entries and errors.

PR #2350, opened 2026-01-12
jefe: Add `read_fault_counts` IPC

hawkw:
This commit adds a new fault-counting capability to Jefe to track the total number of times a given task has faulted. This is requisite for #2341 for reasons I discussed in detail in [this comment][1] and friends. To wit, we cannot easily use the task's generation from its task ID (or the corresponding 32-bit generation count from the kernel) for detecting faults, as those count the number of times the task has *restarted*, and we hope that in the common case, Packrat will generate ereports for faults *before* Jefe has actually restarted the faulted task, so that the panic message can be read out of the dead task's corpse before it's clobbered and so forth. Also, a task may fault and *not* be restarted, such as if someone is "influencing Jefe externally" via the use of `humility jefe -H`. And, finally, tasks may explicitly ask to be restarted without faulting. Thus, we must ask Jefe for an actual fault counter, rather than attempting to use the generation number as an imitation fault counter.

The new IPC takes a leased array of `u32` counters that's `hubris_num_tasks::NUM_TASKS` long, since the intended use case is to read _all_ fault counts in Packrat and perform a scan for tasks whose counts have changed. This felt better than doing a separate IPC for each task. To use a leased fixed-size array effectively here required an `idol-runtime` change (oxidecomputer/idolatry#71) in order to allow `jefe` to write to the array piece-by-piece, without having to construct an array on its stack and write the whole thing into the lease. This branch also updates the `idol-runtime` dependency to include this change.

I 've feature flagged this thing so that we can save several bytes of Jefe on the little boards.

[1]: https://github.com/oxidecomputer/hubris/pull/2341#issuecomment-3740829030


idl/jefe.idol
```diff
@@ -106,5 +106,13 @@ Interface(
...
             idempotent: true,
         ),
+
+        "read_fault_counts": (
+            description: "Reads the total number of times each task has faulted since the system started.",
+            args: {},
+            reply: Simple("TaskFaultCounts"),
+            idempotent: true,
+            encoding: Zerocopy,
+        ),
```
hawkw, line 118: possibly this API is actually bad, since it forces a `NUM_TASKS * 4`-byte array onto the caller's stack --- on Gimletlet, which has 27 tasks (!), this is a 108-byte array. perhaps  it should either take the counters array as a lease, or be done per-task instead of all tasks at once...

task/jefe/src/main.rs
```diff
@@ -74,6 +74,15 @@ fn main() -> ! {
...
     external::set_ready();
 
+    #[cfg(feature = "fault-counters")]
+    let fault_counts = {
+        use static_cell::ClaimOnceCell;
+
+        static COUNTS: ClaimOnceCell<task_jefe_api::TaskFaultCounts> =
+            ClaimOnceCell::new([0usize; NUM_TASKS]);
+        COUNTS.claim()
+    };
```
hawkw, line 84: this saves some stack on Jefe's behalf though. we could also consider giving the task status array the same treatment, which might be nice to do in a separate PR...

hawkw:
It does, at least, appear to work:

```console
eliza@hekate ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Jefe.read_fault_counts
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.12s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Jefe.read_fault_counts`
Jefe.read_fault_counts() => TaskFaultCounts([ 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 ])

eliza@hekate ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- jefe -f ereportulator
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.11s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- jefe -f ereportulator`
humility: successfully changed disposition for ereportulator

eliza@hekate ~/Code/oxide/hubris $ HUMILITY_TARGET=gimletlet cargo xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Jefe.read_fault_counts
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.11s
     Running `target/debug/xtask humility app/gimletlet/app-ereportlet.toml -- hiffy -c Jefe.read_fault_counts`
Jefe.read_fault_counts() => TaskFaultCounts([ 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1 ])
```

idl/jefe.idol
cbiffle: Counters array as a lease seems right to me, fwiw.

We could also alter the Idol codegen to not require the return value to be on the stack... the kernel knows nothing of stacks, that's just the codegen's choice. But the API would suffer, I think.

hawkw: Yeah, I think it should probably be a lease --- I wanted to be able to check that it worked using Hiffy first, though. I'll go and change it now

hawkw: Oh, actually, there is one biggish downside to making it a lease, which is that Idol codegen currently expects the `max_len` field for leases to be an `Option<usize>`, and (critically) not an `Option<ArbitraryRustTokensThatAreWedgedIntoTheGeneratedCode>`. So you can't write something like this:
```ron
        "read_fault_counts": (
            description: "Reads the total number of times each task has faulted since the system started.",
            args: {},
            leases: {
                "counts": (type: "[u32]", write: true, max_len: Some(hubris_num_tasks::NUM_TASKS)),
            },
            reply: Simple("()"),
            idempotent: true,
            encoding: Zerocopy,
        ),
    },
```

hawkw: i had really wanted to use `NUM_TASKS` to ensure that both sides agree on how long the array is and not have to handle too big/too small arrays, but...maybe it's more important to avoid the big stack array? I dunno.

hawkw: we could also go and hack up the codegen to accept a non-integer thing here, i guess

hawkw: Figured out that actually, all we had to do was make a little tweak to `idol_runtime` (idolatry#71) to make passing a leased fixed-size array useable...

PR #2351, opened 2026-01-13
jefe: move a biggish bit of Jefe off the stack

hawkw:
Currently, Jefe represents task states in an array that's initialized in the `main()` function and passed as a `&mut [TaskStatus; NUM_TASKS]` slice to the `ServerImpl` initializer. This is a kind of biggish  array, since each `TaskStatus` is in excess of 64 bytes (it's a timestamp plus a couple enum discriminants), multiplied by `NUM_TASKS`, which can be in the vicinity of 20-30 tasks. Because the array is initialized as a local variable in the `main()` function, the array is on the stack all of the time. It doesn't have to be on the stack, and it could also be not on the stack. 

This commit makes it not be on the stack by putting it in a static instead. Also, I made the static be initialized by code that runs in `const fn`, rather than having to loop over the held tasks at runtime and update them to be held by mutating the array. This just seemed a lot nicer, despite making the code look a bit less nicer. Oh well.

This makes Jefe's stack usage a lot smaller; contrast the Gimletlet image on `master`:
```
jefe: 2048 bytes (limit is 2200)
     [+0] _start
   [+840] main
   [+184] task_jefe::dump::dump_task
   [+776] task_jefe::dump::dump_task_run
    [+56] humpty::DumpAreaHeader::read_and_check
   [+136] task_jefe::dump::dump_task_run::{{closure}}
    [+40] <static_cell::StaticCell<ringbuf::Ringbuf<T,u16,_>> as ringbuf::RecordEntry<T>>::record_entry
    [+16] <humpty::DumpError<T> as core::cmp::PartialEq>::eq
```

...and after this change:

```
jefe: 1424 bytes (limit is 2200)
     [+0] _start
   [+216] main
   [+184] task_jefe::dump::dump_task
   [+776] task_jefe::dump::dump_task_run
    [+56] humpty::DumpAreaHeader::read_and_check
   [+136] task_jefe::dump::dump_task_run::{{closure}}
    [+40] <static_cell::StaticCell<ringbuf::Ringbuf<T,u16,_>> as ringbuf::RecordEntry<T>>::record_entry
    [+16] <humpty::DumpError<T> as core::cmp::PartialEq>::eq
```

So that's some 600B of `main` gone away, which seems cool.

We could probably go through and adjust all the Jefe task sizes to be smaller now that they can be smaller.


task/jefe/src/main.rs
```diff
@@ -64,10 +64,23 @@ const MIN_RUN_TIME: u64 = 50;
...
+            let mut tasks = [INITIAL_STATE; NUM_TASKS];
+            let mut i = 0;
+            while i < generated::HELD_TASKS.len() {
+                let held_task = generated::HELD_TASKS[i] as usize;
+                tasks[held_task].disposition = Disposition::Hold;
+                i += 1;
+            }
+            ClaimOnceCell::new(tasks)
+        };
+        STATES.claim()
```
hawkw, line 82: oh, ugh. one major issue with this change is that this use of `ClaimOnceCell` breaks the build with the `no-panic` userlib feature, since the `claim()` method panics if the cell is already claimed (which never happens, but can't be statically eliminated). that's sad. maybe this change is a non-starter for now.

hawkw: (this is why we can't currently build this branch for `donglet` and any other app that builds jefe with `no-panic` enabled: https://github.com/oxidecomputer/hubris/actions/runs/20967849704/job/60263104877?pr=2351)

PR #2358, opened 2026-01-14
jefe: Add mechanism for notifying another task on task faults

hawkw:
In order for Packrat to produce an ereport when another task faults (see #2341), we must have a way to notify it when such a fault has occurred. 

Furthermore, our scheme for reporting such faults requires that some data be recorded *before* the task has been restarted. In particular, if a task faults due to a panic, the panic message must be read out of its address space before it is restarted, via the `read_panic_message` KIPC I added in #2313. Similarly, the `read_task_status` KIPC can only provide details on the fault while the task remains in the `Faulted` state; when it is restarted, that data is cleared. Thus, we must delay the task's restart for some period of time so that Packrat can gather this information and stuff it into an ereport.

This commit solves both of these problems. First, I've add a new config field to Jefe to allow providing a list of other tasks to which a notification should be posted when a fault occurs (similarly to the existing state-change notification). Second, I've changed the timeout behavior a bit so that, if Jefe is notifying other tasks on a fault, it will _always_ wait at least 5 ms before restarting a faulted task, even if the task has been running for longer than `MIN_RUN_TIME` since its last fault. This way, we always allow some time for a fault report to be produced if someone cares about the task having faulted. The 5 ms minimum delay is elided if there are no tasks to notify on faults.

### DIFF
--- a/task/jefe/src/main.rs
+++ b/task/jefe/src/main.rs
@@ -507,6 +507,12 @@ impl idol_runtime::NotificationHandler for ServerImpl<'_> {
                     // always wait at least `MIN_RESTART_DELAY` before
                     // restarting the task, even if it has run for longer than
                     // `MIN_RUN_TIME` since its last restart.
+                    //
+                    // Clippy doesn't like this `if` condition because it
+                    // doesn't know about codegen; this expression *should*
+                    // always evaluate either to true or false depending on
+                    // Jefe's config.
+                    #[allow(clippy::const_is_empty)]
                     if !generated::FAULT_MAILING_LIST.is_empty() {
                         extra_delay = extra_delay.max(Some(MIN_RESTART_DELAY));
                     }


cbiffle (approved):
Nit: should this include the change to the read_task_status kipc that allows non-supervisor access? I think it's basically required for this to be useful, but I could imagine leaving it in the change that uses it.

sys/num-tasks/build.rs
```diff
@@ -23,6 +23,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     if build_util::has_feature("task-enum") {
         writeln!(task_file, "#[allow(non_camel_case_types)]").unwrap();
+        writeln!(task_file, "#[derive(Copy, Clone)]").unwrap();
```
hawkw, line 26: these...should probably have always been there tbh.

Issue #2361, opened 2026-01-19
cosmo-b fan ramp not quick enough when baseline temperature is high


leftwo:
We saw an issue where a cosmo-b sled in london racklette was not able to ramp the fans fast enough to prevent the CPU from hitting overtemp.
The ambient temperature in the fridge (machine room) that contained the cosmo-b was measured around 80F.
The ambient had come down from mid 90s earlier in the day, so components were probably still pretty warm.

We started up a heavy write IO workload (raw zvol init) on 10 disks at the same time.  This caused the CPU temp to get too hot and the SP powered off the cosmo.

```
  11 1112     3610        1 CriticalDueTo { sensor_id: SensorId(0x37), temperature: Celsius(90.142) }
  12 1120     3610        1 AutoState(Overheated)
```

The full `ringbuf thermal` output (thanks to angela because I can't get the syntax correct):
```
angela@castle ~ $ humility -i fe80::aa40:25ff:fe05:700%london_sw1tp0 -a /staff/alan/image-for-dublin/extracted-ls6/repo/targets/9ddee3e4d85411efaffec562e20452ad309498b237cf06f3ec60986c9a7ad40c.switch_sp-sidecar-b-1.0.56.tar.gz ringbuf thermal
humility: connecting to fe80::aa40:25ff:fe05:700%15
humility: ring buffer drv_i2c_devices::emc2305::__RINGBUF in thermal:
humility: ring buffer drv_i2c_devices::max31790::__RINGBUF in thermal:
humility: ring buffer task_thermal::__RINGBUF in thermal:
   TOTAL VARIANT
    4450 ControlPwm
      29 AutoState(Boot)
       7 AutoState(Running)
       1 AutoState(Overheated)
       1 AutoState(Uncontrollable)
      25 AddedDynamicInput
       8 FanAdded
       5 RemovedDynamicInput
       3 PowerModeChanged
       2 FanControllerInitialized
       1 Start
       1 ThermalMode(Auto)
       1 CriticalDueTo
       1 PowerDownAt
       1 SetFanWatchdogOk
 NDX LINE      GEN    COUNT PAYLOAD
  30 1065        5        1 AutoState(Running)
  31 1206        5        1 ControlPwm(0x33)
   0 1206        6        1 ControlPwm(0x34)
   1 1206        6        1 ControlPwm(0x33)
   2 1206        6        6 ControlPwm(0x34)
   3 1206        6        6 ControlPwm(0x35)
[...]

PR #2362, opened 2026-01-21
thermal: it's PARTY TIME!!!

hawkw:
## The Problem

We have recently been seeing a number of Cosmo-Bs being shut down by the thermal control loop under load, and we we would really prefer they didn't do this. This has generally occurred due to the machine suddenly starting an intensive CPU and NVMe workload (initializing a large number of raw zvols) while inlet air temperature is uncomfortable but within our specified range of 35º F (2º C) and 95º F (35º C): ambient temperatures in "the fridge" in Oxide EMY have been between 75º F and 90º F when thermal shutdowns have occurred. In all observed cases of these thermal shutdowns, we have not seen CPU _T<sub>ctl</sub>_ reach the thresholds of 95 Magic AMD Thermal Sadness Units (at which point the CPU throttles itself) or 115 (at which point `THERMTRIP_L` is asserted). Instead, these systems have shut down due to a timeout in the thermal control loop, which limits the amount of time the system spends in the `Overheated` state to 60 seconds. See #2361 for details on the observed pathology.

## Background

To elaborate a bit here, the Hubris thermal control loop operates in two primary control regimes: the "normal" regime (represented by [`ThermalControlState::Running`], in which fan speeds are governed by PID control with temperature sensor readings as the process variable); and the "overheating" regime (represented by [`ThermalControlState::Overheated`], in which we operate fans at 100% of their PWM duty cycle until we return to the normal control regime. 

For each device whose temperature measurements influence the control loop (on Cosmo, the CPU _T<sub>ctl</sub>_, the M.2 and U.2 SSDs, and T6 NIC), [we define three temperature thresholds][thresholds]: `target_temperature`, `critical_temperature`, and `power_down_temperature`. These impact the behavior of the control loop as follows:

- If any device ever reports a temperature which is over its `power_down_temperature`, the thermal control loop powers the system off immediately,
- If we are in the normal control regime and any device reports a temperature which is over its `critical_temperature` threshold, we transition to the "overheating" control regime,
- Once we are in the "overheating" control regime, we remain there (running at 100% duty cycle) until _every_ device's temperature measurement is less than or equal to it's `target_temperature` (i.e., it has returned to nominal).

There's one additional factor, which will become important shortly: when we enter the "overheating" control regime, we start a timeout for `overheat_timeout_ms`, which is currently always [set to 60 seconds][overheat_timeout_ms]. If 60 seconds have elapsed without returning to nominal, the thermal control loop powers the system off. In effect, if the thermal loop feels that things have not gotten better fast enough, it decides to kill itself immediately.

The present behavior was introduced in PR #2317. Prior to that change, we would exit the "overheating" control regime as soon as all temperatures were at least one degree[^1] below that device's critical threshold, and return to normal control. The effect of this change is that we now remain in the "overheating" regime for _longer_. Perhaps a bit unintuitively, though, the intent (and, I believe, the effect!) of this change was to _reduce_ the likelihood of thermal shutdowns: when we return to normal PID control after dropping 1 degree under critical, the fan speed will go down and the system will immediately start overheating again. In this rapid oscillation between control regimes, we didn't spend enough time at 100% PWM to cool down sufficiently, and something would hit its `power_down_temperature` threshold. 

However, the other impact of #2317 is that the system spends a lot more time in the control regime in which it considers itself to be "overheating", and does not stop the "guess I should kill myself" timeout until temperatures have decreased by a much greater margin than was required prior. Matt[^2] explicitly calls this out in PR:

> Note that this change means the system will stay in `Overheated` _longer_, because it has to get to the normal temperature (instead of critical - 1°).  I didn't change `overheat_timeout_ms` from its current value of 60 seconds; we _really_ should be able to get from critical → target temperature in 60 seconds.

That last sentence --- "we _really_ should be able to get from critical → target temperature in 60 seconds" --- is demonstrably not the case in situations when the inlet air temperature is as hot as it's been in the fridge for the last couple days. Looking at data collected from an overheating Cosmo this morning, we see that once we transition to the "overheating" control regime, the CPU _T<sub>ctl</sub>_ values are steadily decreasing towards the threshold value...just not _fast enough_ to make it there in less than a minute. See [this comment][metrics] for details.

Okay, that was a lot. But now we should have enough context to discuss...

## The (Possible) Solution

This change is based on the following observations:

1. If anything goes over a critical threshold, we want to remain at the max PWM duty cycle until everything returns to nominal (this was what #2317 did, and we want to preserve that property)
2. We want to shut the system down safely if running the fans at full blast fails to improve the thermal situation (this is what the 60-second `overheat_timeout_ms` is supposed to be for, and we don't want to get rid of that behavior)
3. However, the rate at which we can reject heat decreases as inlet temperature increases (I think? I barely remember high-school physics class), so if it's warmer than a typical air-conditioned data center cold isle, it will take us longer to reach a nominal temperature and transition back to the normal control regime. This may take longer than 60 seconds, which is the problem.
4. Just increasing the timeout feels gross. All timeouts are arbitrary. I don't really have a convincing argument for why 120 seconds is meaningfully different than 60 seconds except "it would have stopped that specific Cosmo from shutting down while Alan was running his raw zvol test this morning".
5. Instead, what we really want is to  distinguish between "something got too hot, but things are actively getting better and so we shouldn't kill ourselves" and "no matter how much we spin our fans, we can't seem to make the computer cool off and we should therefore probably just call it".
6. BONUS OBSERVATION: I don't really know *that* much about control theory and just kinda randomly twiddling the PID gains feels scary.

Based on this, this branch introduces a new state to the control loop: TURBO MODE!!!!!.

TURBO MODE!!!!! is an extension of the "overheating" control regime in which we continue to run the fans at full blast, but we _don't_ continue counting down for sixty seconds and then kill ourselves. When a temperature reading goes over the device's critical threshold, we still enter the `Overheated` state, as we did previously. If we remain in `Overheated` for 60 seconds, we shut down the system; again, as we did previously. And, if any temperature hits the device's `power_down_temperature` threshold, we still power down the system immediately, as we did previously. But, we add a new transition out of `ThermalControlState::Overheated`: if we see that _all_ monitored sensors have gone below their critical thresholds, we transition from the overheated state to TURBO MODE!!!!!!

In TURBO MODE!!!, we continue to run the fans at 100% duty cycle, but no longer track the timeout. If all temperatures return to nominal, we exit TURBO MODE!!!!!!!!! and return to normal control. If anything goes back above its critical threshold, we transition from TURBO MODE!!!! back to the "overheating" state, restarting the timeout. This way, if things are below critical but not yet nominal, we keep running at full fan speed for as long as it takes to get back to nominal, rather than giving up if it takes slightly longer than 60 seconds.

[`ThermalControlState::Running`]: https://github.com/oxidecomputer/hubris/blob/ddfb8782336a86cf5c04e48eedc05d6a0ee25756/task/thermal/src/control.rs#L623-L627
[`ThermalControlState::Overheated`]: https://github.com/oxidecomputer/hubris/blob/ddfb8782336a86cf5c04e48eedc05d6a0ee25756/task/thermal/src/control.rs#L629-L636
[thresholds]: https://github.com/oxidecomputer/hubris/blob/ddfb8782336a86cf5c04e48eedc05d6a0ee25756/task/thermal/src/bsp/cosmo_ab.rs#L154-L199
[overheat_timeout_ms]: https://github.com/oxidecomputer/hubris/blob/ddfb8782336a86cf5c04e48eedc05d6a0ee25756/task/thermal/src/control.rs#L711
[metrics]: https://github.com/oxidecomputer/hubris/issues/2361#issuecomment-3780770788

[^1]: Or one Happy Fun AMD _T<sub>ctl</sub>_ Badness Amount Unit.
[^2]: Who I'm not going to tag in this as he's on leave.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,9 +217,30 @@ jobs:
       # duration estimates are based on time reported by the runner???
       - name: Move forward in time to ensure a different build date 
         run: |
+          echo "Before time change: $(date)"
+
           sudo timedatectl set-ntp false
-          sudo timedatectl set-time "$(date -d '1 day ago 11 hours ago' "+%Y-%m-%d %H:%M:%S")"
-          date
+
+          # Changing the system time only works if NTP is disabled. We disable NTP in the command
+          # above, but the command exits as soon as systemd *begins* to stop the NTP daemon. We saw
+          # CI builds where shutting down NTP took longer than expected, causing set-time command to
+          # fail with this message:
+          #
+          #     Failed to set time: Previous request is not finished, refusing.
+          #
+          # To prevent spurious failures, we repeatedly try to set the time with a backoff. The
+          # error message was added to systemd in https://github.com/systemd/systemd/pull/11424.
+          set_time() {
+            for backoff in 1 2 4 8 16; do
+              sudo timedatectl set-time "$1" && return 0
+              echo "retrying in $backoff seconds"
+              sleep $backoff
+            done
+            return 1
+          }
+          set_time "$(date -d '1 day ago 11 hours ago' "+%Y-%m-%d %H:%M:%S")"
+
+          echo "After time change: $(date)"
 
       - name: Build a Hubris board
         run: |


hawkw:
@labbott:

> I feel like I want a state machine doc here showing The ThermalControlState transitions though. 

 here's a Mermaid diagram that's kinda gross-looking:

```mermaid
stateDiagram-v2
    [*] --> Running
    Running --> Overheat: any temp over critical
    Overheat --> Running: all temps nominal
    Overheat --> Turbo: all temps under critical
    Turbo --> Overheat: any temp over critical
    Turbo --> Running: all temps nominal
    Running --> Uncontrollable: any temp over power_down
    Overheat --> Uncontrollable: any temp over power_down
    Overheat --> Uncontrollable: 60 seconds elapse
    Turbo --> Uncontrollable: any temp over power_down
    Uncontrollable --> [*]
```

I may try to turn this into an ASCII graphic that can go in a source code comment.

hawkw:
> It does seem like we could potentially bounce between `Overheated` and `Turbo` indefinitely. Is there a point where we should lie down/try not to cry/cry a lot/give up if it seems like we will never reach nominal?

Yeah, I've wondered about this. Since the way we ended up here was kind of a case of "timeouts, timeouts, always wrong, some too short and some too long", I was a little reluctant to add a second layer of timeouts to the state that was added specifically to work around the overly aggressive timeout.

Ideally, I think maybe we would have a timeout that tracks how long we spent over critical without reaching nominal that isn't reset when we transition from `Overheat` -> `Turbo` but that only counts time spent in `Overheat`. The timeout would only be reset if we return to normal control, but is paused while below critical thresholds. Implementing that is just a little bit more annoying because it requires tracking a sum of time spent in `Overheat` as well as the current start time, but we could do that.

I did also think about fancier things like trying to look at the rate of change, but that felt a little too complex. I think it's worth keeping the behavior in the "overheated" control regime simple, and I didn't wanna have to store a bunch of past measurements to track change over time for RAM reasons.

task/thermal/src/control.rs
```diff
@@ -1197,6 +1227,77 @@ impl<'a> ThermalControl<'a> {
...
+                        if model.should_power_down(temperature) {
+                            any_power_down = Some((sensor_id, temperature));
+                        }
+                        if model.is_critical(temperature) {
+                            any_critical = Some((sensor_id, temperature));
+                        }
+                        worst_margin =
+                            worst_margin.min(model.margin(temperature).0);
+                    }
+                }
```
hawkw, line 1324: part of me wondered a bit about trying to refactor this so that more code is shared between the scans over temp readings in the `Turbo` and `Overheated` states, but...the transition criteria for each state are different enough that it's slightly too annoying.

labbott:
> Ideally, I think maybe we would have a timeout that tracks how long we spent over critical without reaching nominal that isn't reset when we transition from `Overheat` -> `Turbo` but that only counts time spent in `Overheat`. The timeout would only be reset if we return to normal control, but is paused while below critical thresholds. Implementing that is just a little bit more annoying because it requires tracking a sum of time spent in `Overheat` as well as the current start time, but we could do that.

Is there any way we could gather this data to have it available but not use it for control decisions? I guess technically if we're not over critical we're supposed to be "fine" but I'm trying to think of data to gather to make it easier to refine this loop. Maybe this is already logged in the counted ringbuf with the number of times we hit turbo?

hawkw:
> > Ideally, I think maybe we would have a timeout that tracks how long we spent over critical without reaching nominal that isn't reset when we transition from `Overheat` -> `Turbo` but that only counts time spent in `Overheat`. The timeout would only be reset if we return to normal control, but is paused while below critical thresholds. Implementing that is just a little bit more annoying because it requires tracking a sum of time spent in `Overheat` as well as the current start time, but we could do that.
> 
> Is there any way we could gather this data to have it available but not use it for control decisions? I guess technically if we're not over critical we're supposed to be "fine" but I'm trying to think of data to gather to make it easier to refine this loop. Maybe this is already logged in the counted ringbuf with the number of times we hit turbo?

Ah, whoops, https://github.com/oxidecomputer/hubris/pull/2362/commits/2923d0e9509fa083632dfaaa6de8e74ae3837618 just makes the change to make it count against the timeout across transitions between `Critical` and `Turbo`; I can change that back to just log the data if you'd prefer.

task/thermal/src/control.rs
hawkw: https://github.com/oxidecomputer/hubris/pull/2362/commits/ab50bf31ba85f8dd9f5f12180b3e2cfdf03ffc96 is (IMO) about the best pass at reducing repetition in the state machine I could come up with.

task/thermal/src/control.rs
```diff
@@ -632,9 +636,26 @@ enum ThermalControlState {
...
+    /// the `Overheated` state and enter TURBO MODE!!!!, a special state that's
+    /// kind of halfway between `Overheated` and normal operation. In TURBO
+    /// MODE, we continue to run the fans at their max duty cycle, but we don't
+    /// track the overheated timeout. If anything goes above critical while in
+    /// TURBO MODE!!!!!, we return to `Overheated`.
+    ///
+    /// This gives us an opportunity to recover from overheating by running the
+    /// fans aggressively without also deciding to give up and kill ourselves
+    /// while things are improving but not fast enough.
+    Turbo {
```
hawkw: considered _extremely_ unimportant, but: Y'know...I chose this name because I thought it was funny, but it occurs to me now that perhaps the existing colloquialism "fan party" (meaning: "when one or more compute sleds in the office goes MAX RPM") deserves to be enshrined in the code.

task/thermal/src/control.rs
```diff
@@ -615,6 +615,45 @@ type DynamicChannelsArray =
 ///
 /// Note that the canonical temperatures are stored in the `sensors` task; we
 /// copy them into these arrays for local operations.
+///
+/// The thermal control loop transitions between these states:
```
labbott: thanks this makes things much clearer for me

task/thermal/src/control.rs
labbott: `SorryForFanPartying`

task/thermal/src/control.rs
hawkw: thank @rmustacc for the lovely hand crafted ascii art! i'm going to also add a textual description of the transitions here as well.

hawkw: @labbott FYI, commits b7c9b5ab57b2c6c4564e97b6f8035708b9a67719 and e82029346b24e7d6956789caeeb73f9cd7c31a42 added a bunch more documentation and renamed some things, to try and make the states and their transitions a little clearer.

task/thermal/src/control.rs
```diff
@@ -611,7 +615,110 @@ type DynamicChannelsArray =
...
+/// cannot reduce temperatures below a critical threshold.  Therefore, the
+/// timeout is only applied while any component temperature(s) are at or above
+/// critical thresholds.  If running the fans at full speed is effectively
+/// reducing the system temperature, but we have not yet returned to normal
+/// control, the timeout is not applied.  Therefore, we separate the overheated
+/// control regime into two substates:
+///
+/// - `Overheat`, in which at least one component is critical and the timeout
+///    is being tracked, and
+/// - `Turbo`, in which all temperatures are below critical, and we will run
```
labbott: Need to update the docs with `FanParty` here if that's what we're using

task/thermal/src/control.rs
```diff
@@ -1165,32 +1288,72 @@ impl<'a> ThermalControl<'a> {
...
+                    let values = *values;
+                    self.record_leaving_critical(now_ms);
+                    self.state = ThermalControlState::FanParty { values };
                     ringbuf_entry!(Trace::AutoState(self.get_state()));
 
-                    ControlResult::Pwm(PWMDuty(pwm as u8))
-                } else if now_ms > *start_time + self.overheat_timeout_ms {
+                    ControlResult::Pwm(PWMDuty(
+                        self.pid_config.max_output as u8,
+                    ))
```
labbott: Worth it to pull this out to a separte function for consistency?

hawkw: I went back and forth on it, since the other `transition_to_whatever` functions were factored out because they got called in multiple places, and this one wasn't. But, I agree that it's more consistent, so why not.

hawkw:
@labbott This comment https://github.com/oxidecomputer/hubris/issues/2361#issuecomment-3795379774 from Alan includes ringbufs from the latest commit on this branch (7eed7e438414f1c907997c866e300be70add8e1d)  during a thermal excursion triggered by raw zvol initialization (the same trigger as the past thermal excursions that led to shutdown).

The key observations are:

1. We survived the thermal excursion without a shutdown and returned to normal PID control afterwards:
   ```
   17 1352       35        1 ControlPwm(0x44)                                                            
   18 1401       35        1 CriticalDueTo { sensor_id: SensorId(0x37), temperature: Celsius(90.0295) }  
   19 1409       35        1 AutoState(Overheated)                                                       
   20 1352       35        3 ControlPwm(0x64)                                                            
   21 1430       35        1 AutoState(FanParty)                                                         
   22 1352       35       47 ControlPwm(0x64)                                                            
   23 1478       35        1 OverheatedFor(0xc353)                                                       
   24 1481       35        1 CriticalFor(0xb87)                                                          
   25 1388       35        1 AutoState(Running)                                                          
   26 1352       35        2 ControlPwm(0x21)                                                            
   27 1352       35        1 ControlPwm(0x25)
   ```
2. Looking at the ringbuf, we can see that we went over a critical threshold _once_ at the beginning of the overheat excursion, and after a couple of loop iterations in overheated, we went under critical, and remained under critical for the rest of the time we spent in the max-PWM overheat control regime. We can see there is a single `AutoState(Overheated)` entry followed by three `ControlPwm(0x64)`s and then an `AutoState(FanParty)`. We remain in `FanParty` (i.e., not crossing back over a critical threshold) until we transition back to normal control.
3. Based on the overheat timings we added, we see that:
   - We were in the overheated state for a total of 50003 ms (0xc353)
   - Of that, 2951 ms (0xb87) were spent with a component over critical.

So, it looks like this is behaving as intended. At least under the conditions we observed here (71F ambient in the fridge according to Alan, actual inlet temps for this sled may be somewhat higher depending on which test rack it's in?), we don't spend a lot of time ping-ponging between the `Critical` and `FanParty` states. That may not be the case with worse inlet temperatures, but it's what we saw here.

task/thermal/src/control.rs
```diff
@@ -611,7 +615,111 @@ type DynamicChannelsArray =
 ///
 /// Note that the canonical temperatures are stored in the `sensors` task; we
 /// copy them into these arrays for local operations.
+///
+/// ## Theory of Operation
```
labbott, line 619: This is super helpful!

hawkw:
> Thanks for doing the extra testing

Thank @leftwo for that!

leftwo:
How do we make sure this gets rolled into Omicron 18?

labbott:
> How do we make sure this gets rolled into Omicron 18?

This is going to be tagged in the next hubris release. There needs to be one infra fix that @emilyalbini is working on before we can do any release.

PR #2363, opened 2026-01-21
minibar: add control-plane-agent task to minibar

adamlouis:
In manufacturing contexts, we'd like to like to read Ignition port state to address https://github.com/oxidecomputer/facade/issues/298.

We'd like to do this over the management network with `faux-mgs`.

For the minibar to to respond to `faux-mgs` messages, it needs the the `control-place-agent` task.

This PR is an attempt at the bare minimum needed to add `control-plane-agent` to minibar. This is primarily copy+paste / compiler error message driven. cc @Aaron-Hartwig for comments on if there's something more / different for what we want here.

I've tested this PR on the `argon` manufacturing station in Emeryville - a minibar-lite with a rev B board.

This change is required for the ignition test in the additional facade tests: https://github.com/oxidecomputer/facade/pull/339

### DIFF
--- a/.github/workflows/build-boards.yml
+++ b/.github/workflows/build-boards.yml
@@ -108,8 +108,14 @@ jobs:
       - name: Fetch Humility
         if: runner.os == 'Linux'
         run: |
-          curl -fLo target/release/humility https://github.com/oxidecomputer/humility/releases/download/nightly/humility
+          gh release download -R oxidecomputer/humility nightly -p humility -O target/release/humility
           chmod +x target/release/humility
+        env:
+          # The oxidecomputer/humility repository is public, so *in theory* we don't need a token to
+          # download releases from it. In practice we occasionally saw downloading Humility failing
+          # with a 429 (too many requests) status code. To avoid that we use the `gh` CLI with the
+          # job's GitHub token to download, granting us higher rate limits.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Test Humility manifest
         if: runner.os == 'Linux'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,9 +217,30 @@ jobs:
       # duration estimates are based on time reported by the runner???
       - name: Move forward in time to ensure a different build date 
         run: |
+          echo "Before time change: $(date)"
+
           sudo timedatectl set-ntp false
-          sudo timedatectl set-time "$(date -d '1 day ago 11 hours ago' "+%Y-%m-%d %H:%M:%S")"
-          date
+
+          # Changing the system time only works if NTP is disabled. We disable NTP in the command
+          # above, but the command exits as soon as systemd *begins* to stop the NTP daemon. We saw
+          # CI builds where shutting down NTP took longer than expected, causing set-time command to
+          # fail with this message:
+          #
+          #     Failed to set time: Previous request is not finished, refusing.
+          #
+          # To prevent spurious failures, we repeatedly try to set the time with a backoff. The
+          # error message was added to systemd in https://github.com/systemd/systemd/pull/11424.
+          set_time() {
+            for backoff in 1 2 4 8 16; do
+              sudo timedatectl set-time "$1" && return 0
+              echo "retrying in $backoff seconds"
+              sleep $backoff
+            done
+            return 1
+          }
+          set_time "$(date -d '1 day ago 11 hours ago' "+%Y-%m-%d %H:%M:%S")"
+
+          echo "After time change: $(date)"
 
       - name: Build a Hubris board
         run: |
--- a/task/thermal-api/src/lib.rs
+++ b/task/thermal-api/src/lib.rs
@@ -75,8 +75,9 @@ pub enum ThermalMode {
 pub enum ThermalAutoState {
     Boot,
     Running,
-    Overheated,
+    Critical,
     Uncontrollable,
+    FanParty,
 }
 
 /// Properties for a particular part in the system
--- a/task/thermal/src/control.rs
+++ b/task/thermal/src/control.rs
@@ -484,6 +484,10 @@ pub(crate) struct ThermalControl<'a> {
 
     /// Has the fan watchdog been configured yet?
     fan_watchdog_configured: bool,
+
+    /// Tracks the total duration of excursions into the overheated control
+    /// regime.
+    overheat_timer: Option<OverheatTimer>,
 }
 
 /// Represents the state of a temperature sensor, which either has a valid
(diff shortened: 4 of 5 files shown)


hawkw:
Your CI build is unhappy because [`rustfmt` needs to be run](https://github.com/oxidecomputer/hubris/actions/runs/21230101668/job/61086485723?pr=2363) (sorry!)

app/minibar/rev-b.toml
```diff
@@ -50,6 +50,26 @@ extern-regions = [ "sram1", "sram2", "sram3", "sram4" ]
 notifications = ["socket"]
 features = ["net", "vlan"]
 
+[tasks.control_plane_agent]
```
hawkw, line 56: Is there a reason that we only include a `control-plane-agent` in the rev B image? Does networking not work on the rev A board or something? Could we maybe have a comment explaining why this is the case?

task/control-plane-agent/src/mgs_minibar.rs
```diff
@@ -0,0 +1,648 @@
...
+pub(crate) type UpdateBuffer =
+    update_buffer::UpdateBuffer<SpComponent, UPDATE_BUFFER_SIZE>;
+pub(crate) type BorrowedUpdateBuffer = update_buffer::BorrowedUpdateBuffer<
+    'static,
+    SpComponent,
+    UPDATE_BUFFER_SIZE,
+>;
+static UPDATE_MEMORY: UpdateBuffer = UpdateBuffer::new();
+
+#[path = "mgs_minibar/ignition.rs"]
```
hawkw: nitpick: why not just name the file `ignition_handler.rs` and not have to do this?

task/control-plane-agent/src/mgs_minibar.rs
```diff
@@ -0,0 +1,648 @@
...
+};
+use host_sp_messages::HostStartupOptions;
+use idol_runtime::{Leased, RequestError};
+use ringbuf::ringbuf_entry_root;
+use task_control_plane_agent_api::{ControlPlaneAgentError, OxideIdentity};
+use task_net_api::{MacAddress, UdpMetadata, VLanId};
+use userlib::sys_get_timer;
+
+// Minimal update buffer types required for compilation (not actually used)
+const UPDATE_BUFFER_SIZE: usize = 1024;
```
hawkw: if the update buffer is not actually used, does it need to be 1KiB? Is there something asserting that it's at least 1KiB someplace?

task/control-plane-agent/src/mgs_minibar.rs
```diff
@@ -0,0 +1,648 @@
...
+    ) -> Result<usize, RequestError<ControlPlaneAgentError>> {
+        Err(ControlPlaneAgentError::DataUnavailable.into())
+    }
+
+    pub(crate) fn startup_options_impl(
+        &self,
+    ) -> Result<HostStartupOptions, RequestError<ControlPlaneAgentError>> {
+        // We don't have a host to give startup options; no one should be
+        // calling this method.
+        Err(ControlPlaneAgentError::InvalidStartupOptions.into())
```
hawkw: perhaps these errors all ought to be [`ControlPlaneAgentError::OperationUnsupported`](https://github.com/oxidecomputer/hubris/blob/ddfb8782336a86cf5c04e48eedc05d6a0ee25756/task/control-plane-agent-api/src/lib.rs#L26)? I think that more accurately communicates "I am not a board that knows how to do that"

hawkw: ...oh, i guess this is what the PSC and Sidecar implementations return for these methods. That feels a little weird to me, but I guess it's better to do what we do elsewhere, maybe? It could be worth checking if that error variant even existed when it was decided that PSC/Sidecar should return these...

app/minibar/rev-b.toml
Aaron-Hartwig: Yeah the rev A boards had a hardware bug that kept the minibar's SP from being reachable over the management network.

task/control-plane-agent/src/main.rs
```diff
@@ -259,12 +260,19 @@ impl ServerImpl {
...
+        #[cfg(not(feature = "minibar"))]
+        let mask = notifications::SOCKET_MASK
             | notifications::USART_IRQ_MASK
-            | notifications::TIMER_MASK
+            | notifications::TIMER_MASK;
+
+        #[cfg(feature = "minibar")]
+        let mask = notifications::SOCKET_MASK | notifications::TIMER_MASK;
+
+        mask
```
Aaron-Hartwig, line 276: Probably worth adding a comment in here about why this bifurcation exists. That is, why minibar has a different set of notifications than all other users.

task/control-plane-agent/src/mgs_minibar.rs
Aaron-Hartwig: I don't spend much time in this code, but I agree that `OperationUnsupported` feels like a much better error here (and probably those other places too!). This would also apply to `set_startup_options_impl` below.

app/minibar/rev-b.toml
labbott: Are we likely to get a rev-c? We could make figuring out the split a problem for Future Us if needed.

task/control-plane-agent/src/mgs_minibar.rs
```diff
@@ -0,0 +1,659 @@
...
+    ) -> Result<(), SpError> {
+        ringbuf_entry_root!(Log::MgsMessage(MgsMessage::UpdatePrepare {
+            length: update.total_size,
+            component: update.component,
+            id: update.id,
+            slot: update.slot,
+        }));
+
+        Err(SpError::RequestUnsupportedForComponent)
+    }
```
labbott, line 268: Would we ever want to update minibar over the network? that seems like it could be useful.

app/minibar/rev-b.toml
hawkw: Okay, thanks Aaron. Would love a comment stating that here.

adamlouis: Comment added!

There's no plan I am aware of for rev C - seems like something ok for future us?

task/control-plane-agent/src/mgs_minibar.rs
adamlouis: Yes, definitely!

Right now in manufacturing, we _do_ have programming headers on all the minibar SPs.

If ok, would be helpful to defer this until I have few more cycles to test

hawkw: Right now, there's a ton of stuff related to SP updates that are dead code on `minibar`. I think that, in the interest of urgency, I'm just going to add `#[cfg(not(feature = "minibar"))]`, but will add commentary to the effect that we can always un-cfg-out it if we want to update minibars over the network later.

hawkw: Update: we discussed this on matrix and decided that updates and other common functionality that we get "for free" from `MgsCommon` may as well just be enabled, since (a) figuring out which stuff needs to be `cfg`'d out to get Minibar to build successfully without dead-code warnings was kinda annoying, and (b) it seems good to include them in the image even if we're not planning to use them _now_, since it could be a secret tool that will help us later.

task/control-plane-agent/src/mgs_minibar.rs
hawkw: I'm a little inclined to leave this the way it is on PSC and Sidecar in this PR, and then change all of them separately.

task/control-plane-agent/src/ignition_controller.rs
```diff
@@ -0,0 +1,207 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
labbott, line 1: This entire file seems like a lot of copy/paste from `task/control-plane-agent/src/mgs_sidecar/ignition.rs` can we see if there's anything we can keep in common?

hawkw: aa83c6992bb7e91633e992611f3a616e077411ea factors this out so that the majority of the ignition controller code is defined once and shared by Sidecar and Minibar.

I note that minibar lacks code for actually sending ignition commands, which is present on sidecar. Perhaps we ought to put that in both systems as well? It seems like it could be useful to be able to actually send ignition commands to the attached sled...

hawkw: okay, I just went ahead and enabled all Ignition functionality on both Minibar and Sidecar, because why not? 73df690c8ed467ecf46e7966e78076275b99450c

labbott:
LGTM but I think we should give this a quick test on a Real Sidecar too to make sure none of the refactoring broke anything (unlikely! but you never know!)

hawkw:
> LGTM but I think we should give this a quick test on a Real Sidecar too to make sure none of the refactoring broke anything (unlikely! but you never know!)

This morning I updated both sidecars on `madrid` to an image built from this branch. I tested using Ignition to power-cycle a sled from Wicket and it appears to have worked. If there are other Ignition commands you think ought to be tested, let me know, I'm happy to do other stuff.

I do note that SW0 on `madrid` seems to be happy while SW1's ignition controller seems sad.

<details>
<summary> wicket "screenshots"</summary>

SW0:
```
 ╭──────────────────╮╭──────────────────────────────────────────────────────────────────────────────╮
 │OVERVIEW          ││OXIDE RACK / SWITCH 0                                                         │
 │UPDATE            │╰──────────────────────────────────────────────────────────────────────────────╯
 │RACK SETUP        │╭──────────────────────────────────────────────────────────────────────────────╮
 │                  ││ Ignition:                                                                    │
 │                  ││   • Identity: Sidecar                                                        │
 │                  ││   • Power: Yes                                                               │
 │                  ││   • Faults: A3                                                               │
 │                  ││   • Controller 0 detected: Yes                                               │
 │                  ││   • Controller 1 detected: No                                                │
 │                  ││                                                                              │
 │                  ││ Service Processor:                                                           │
 │                  ││   • Serial number: BRM44220005                                               │
 │                  ││   • Model: 913-0000006                                                       │
 │                  ││   • Revision: 4                                                              │
 │                  ││   • Base MAC Address: a8:40:25:05:01:00                                      │
 │                  ││   • Power State: A0                                                          │
 │                  ││   • Active Slot:                                                             │
 │                  ││       • Hubris Archive ID: 55164ea8f8d90eb9                                  │
 │                  ││       • Git Commit: 4e8ca46698be46d002bbabe18d97c641dc3cf1a3                 │
 │                  ││       • Board: sidecar-b                                                     │
 │                  ││       • Name: sidecar-b
```

SW1:

```
 ╭──────────────────╮╭──────────────────────────────────────────────────────────────────────────────╮
 │OVERVIEW          ││OXIDE RACK / SWITCH 1                                                         │
 │UPDATE            │╰──────────────────────────────────────────────────────────────────────────────╯
 │RACK SETUP        │╭──────────────────────────────────────────────────────────────────────────────╮
 │                  ││ Ignition: Not available                                                      │
 │                  ││                                                                              │
 │                  ││ Service Processor:                                                           │
 │                  ││   • Serial number: BRM44220011                                               │
 │                  ││   • Model: 913-0000006                                                       │
 │                  ││   • Revision: 4                                                              │
 │                  ││   • Base MAC Address: a8:40:25:05:06:00                                      │
 │                  ││   • Power State: A0                                                          │
 │                  ││   • Active Slot:                                                             │
 │                  ││       • Hubris Archive ID: 55164ea8f8d90eb9                                  │
 │                  ││       • Git Commit: 4e8ca46698be46d002bbabe18d97c641dc3cf1a3                 │
 │                  ││       • Board: sidecar-b                                                     │
 │                  ││       • Name: sidecar-b
```

</details>

But, all the racklette environments look similar to this, so I think that's not my fault.

PR #2366, opened 2026-01-26
Fix spurious failure when changing the time

emilyalbini:
Eliza and Laura pointed out [a spurious failure](https://github.com/oxidecomputer/hubris/actions/runs/21369343067/job/61509812427) caused by the new reproducibility check jobs I added a bit ago, and this PR fixes it. Details on what caused the failure are in code comments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,9 +217,30 @@ jobs:
       # duration estimates are based on time reported by the runner???
       - name: Move forward in time to ensure a different build date 
         run: |
+          echo "Before time change: $(date)"
+
           sudo timedatectl set-ntp false
-          sudo timedatectl set-time "$(date -d '1 day ago 11 hours ago' "+%Y-%m-%d %H:%M:%S")"
-          date
+
+          # Changing the system time only works if NTP is disabled. We disable NTP in the command
+          # above, but the command exits as soon as systemd *begins* to stop the NTP daemon. We saw
+          # CI builds where shutting down NTP took longer than expected, causing set-time command to
+          # fail with this message:
+          #
+          #     Failed to set time: Previous request is not finished, refusing.
+          #
+          # To prevent spurious failures, we repeatedly try to set the time with a backoff. The
+          # error message was added to systemd in https://github.com/systemd/systemd/pull/11424.
+          set_time() {
+            for backoff in 1 2 4 8 16; do
+              sudo timedatectl set-time "$1" && return 0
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -217,9 +217,30 @@ jobs:
...
+
+          # Changing the system time only works if NTP is disabled. We disable NTP in the command
+          # above, but the command exits as soon as systemd *begins* to stop the NTP daemon. We saw
+          # CI builds where shutting down NTP took longer than expected, causing set-time command to
+          # fail with this message:
+          #
+          #     Failed to set time: Previous request is not finished, refusing.
+          #
+          # To prevent spurious failures, we repeatedly try to set the time with a backoff. The
+          # error message was added to systemd in https://github.com/systemd/systemd/pull/11424.
```
hawkw, line 232: this is a super-duper ultra nitpick but can we wrap this at the 80th column?

.github/workflows/ci.yml
```diff
@@ -217,9 +217,30 @@ jobs:
...
+          set_time() {
+            for backoff in 1 2 4 8 16; do
+              sudo timedatectl set-time "$1" && return 0
+              echo "retrying in $backoff seconds"
+              sleep $backoff
+            done
+            return 1
+          }
+          set_time "$(date -d '1 day ago 11 hours ago' "+%Y-%m-%d %H:%M:%S")"
```
hawkw, line 241: will the "run" step fail if `set_time` returns 1? i don't know if Github Actions will implicitly `set -e` or if we need to do that manually?

.github/workflows/ci.yml
emilyalbini: Hmm, every other comment in the file is wrapped at 100 (whoops, forgot to switch my text editor's settings when working on this file), should I rewrap the whole file?

.github/workflows/ci.yml
emilyalbini: Yeah, the step will fail if `set_time` returns 1. [The default shell on Linux and macOS is `bash -e`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell).

hawkw: awesome, carry on

.github/workflows/ci.yml
hawkw: meh, let's do that in a separate PR to keep the git history reasonabler, if we do it at all...

PR #2370, opened 2026-01-28
minibar: PERST fixes, PCIe power control changes

Aaron-Hartwig:
This picks up a new FPGA image which no longer enables the PCIe power rails by default. Additionally, it changes a few things about PERST. The relevant SP software bits are that we now provide a way for the PCIe power rails to be controlled and have their status checked.


Aaron-Hartwig:
Sorry for all the noise in the JSON here, we recently changed our register map output to output a flattened map.

adamlouis:
I've tested on `argon`, `bromine`, & others - works!

Issue #2355, opened 2026-01-13
thermal: want a mechanism to disable components instead of the main system


Aaron-Hartwig:
As the thermal management subsystem works today, when any sensor reaches its `model.should_power_down` value, which gives us a `ControlResult::PowerDown`, with the ultimate result of thermal task calling the `power_down()` method of its BSP. I think what we want here is a way for a model to specify _what_ should be powered down. For example: if a transceiver module were to to have reported a temperature above the power down threshold in our model for it, _the entire switch would power down_. This behavior is not acceptable given the failure mode. The behavior we'd want is for the thermal model to optionally specify a device to shut down instead of the whole system.

hawkw:
When we do this, I'd like to also make sure that the disabledness makes it all the way into the control-plane-agent inventory...which it _might_, if it gets disabled in a way that is visible to `validate`, but it also might not --- I _think_ that validate will broadly only return `Unavailable` and `DeviceOff` if it doesn't get an I2C ACK but I'd need to check whether that's actually true?

https://github.com/oxidecomputer/hubris/blob/81a9b54dae5284842bfadee883d20aa70c50e4c5/task/control-plane-agent/src/inventory.rs#L110-L125

hawkw:
The most obvious naive way of doing this for transceivers, where the `thermal` task calls into the `transceivers` task to say "hey, turn off this transceiver", is not gonna work with Hubris' IPC model, since `transceivers` is an IPC client of `thermal` (and thus, lower priority), in order to add/remove dynamic inputs for transceivers. That's a shame. I can think of a couple other approaches, but I'll have to spend more time looking at the code to determine which make sense.

One possibility is to have the `thermal` task track a list of individual components that need to be powered down, and notify `transceivers` (and eventually, other tasks, if we want to do similar things for other components), and then `transceivers` comes and asks `thermal` which components need to be shut down. This is not my facorite thing, especially because it means that the eventual thermal shutdown of the transceivers happens "eventually", rather than within that thermal loop iteration.

Another option is for `thermal` to know how to shut down a transceiver in the Sidecar BSP, and have the thermal loop just go and do that by doing whatever I2C stuff is necessary to turn it off. Whether or not this makes sense depends a bit on the complexity of shutting down a transceiver --- I don't really want to put a large amount of transceiver specific knowledge in `thermal` that's duplicated with `transceivers`...

hawkw:
[...]

PR #2374, opened 2026-02-02
[WIP] Don't power down the whole sidecar due to one hot xcvr

hawkw:
See #2355


hawkw:
good news: it does work! after hardcoding port 7 to return a temperature of 99.0C, I see that...

...the switch remains in A0:

image

...but the allegedly hot transceiver is powered off:

image

Presently there's nothing that will attempt to re-enable that module, though, so the only way to clear the fault is to reset the SP. That will need to be figured out before this is ready for prime-time.

PR #2376, opened 2026-02-05
power: add the RAA229620A to some filter lists

cbiffle:
This is required for getting the power task to treat it like a Renesas power supply, basically.

### DIFF
--- a/task/power/src/main.rs
+++ b/task/power/src/main.rs
@@ -185,6 +185,7 @@ impl Device {
     ) -> Result<Amperes, ResponseCode> {
         match &self {
             Device::Raa229618(dev) => Ok(dev.read_phase_current(phase)?),
+            Device::Raa229620A(dev) => Ok(dev.read_phase_current(phase)?),
             Device::Isl68224(dev) => Ok(dev.read_phase_current(phase)?),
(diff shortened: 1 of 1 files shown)


labbott (approved):

task/power/src/main.rs
```diff
@@ -1127,7 +1128,12 @@ impl idl::InOrderPowerImpl for ServerImpl {
             .iter()
             .find(|d| {
                 d.i2c_device().address == addr
-                    && matches!(d, Device::Raa229618(..) | Device::Isl68224(..))
+                    && matches!(
+                        d,
+                        Device::Raa229618(..)
+                            | Device::Raa229620A(..)
+                            | Device::Isl68224(..)
```
hawkw: nit, take it or leave it: as this list has grown a bit more now and is repeated in `rendmp_dma_read` and `rendmp_dma_write`, should we maybe just have a `fn is_rendmp_device()` or something that both functions can just write
```rust
self.devices
    .iter()
    .find(|d| d.i2c_device().address == addr && is_rendmp_device(d))
```
or something?

cbiffle: Done!

PR #2383, opened 2026-02-10
[WIP] `thermal` task ereports

hawkw:


jamesmunns:
@hawkw should we close this now that #2645 has landed and #2639 is active?

Issue #2385, opened 2026-02-12
Thermal loop doesn't distinguish between "real" and "fake" temperatures


nathanaelhuffman:
```
➜  ~ humility -d 1001-sled26_hubris.core.0 ringbuf thermal
humility: attached to dump
humility: ring buffer drv_i2c_devices::emc2305::__RINGBUF in thermal:
humility: ring buffer drv_i2c_devices::max31790::__RINGBUF in thermal:
humility: ring buffer task_thermal::__RINGBUF in thermal:
   TOTAL VARIANT
    7545 ControlPwm
      57 SensorReadFailed
       3 AutoState(Boot)
       2 AutoState(Running)
       1 AutoState(Overheated)
       1 AutoState(Uncontrollable)
       6 FanAdded
       3 MiscReadFailed
       2 PowerModeChanged
       1 Start
       1 ThermalMode(Auto)
       1 PowerDownDueTo
       1 CriticalDueTo
       1 FanControllerInitialized
       1 PowerDownAt
       1 SetFanWatchdogOk
 NDX LINE      GEN    COUNT PAYLOAD
  30  895        4        1 SensorReadFailed(SensorId(0x38), I2cError(NoDevice))
  31 1206        4        1 ControlPwm(0x57)
   0  895        5        1 SensorReadFailed(SensorId(0x38), I2cError(NoDevice))
   1 1112        5        1 CriticalDueTo { sensor_id: SensorId(0x38), temperature: Celsius(80.3285) }
   2 1120        5        1 AutoState(Overheated)
   3 1206        5        1 ControlPwm(0x64)
   4  895        5        1 SensorReadFailed(SensorId(0x38), I2cError(NoDevice))
   5 1206        5        1 ControlPwm(0x64)
   6  895        5        1 SensorReadFailed(SensorId(0x38), I2cError(NoDevice))
   7 1206        5        1 ControlPwm(0x64)
   8  895        5        1 SensorReadFailed(SensorId(0x38), I2cError(NoDevice))
   9 1206        5        1 ControlPwm(0x64)
  10  895        5        1 SensorReadFailed(SensorId(0x38), I2cError(NoDevice))
  11 1206        5        1 ControlPwm(0x64)
  12  895        5        1 SensorReadFailed(SensorId(0x38), I2cError(NoDevice))
  13 1206        5        1 ControlPwm(0x64)
  14  895        5        1 SensorReadFailed(SensorId(0x38), I2cError(NoDevice))
[...]

Issue #2384, opened 2026-02-12
On Cosmo, hubris misses NIC_MAPO leading to a fake "Overheat"


nathanaelhuffman:
Here's a snip from the cosmo_seq that shows the issue, I've included from the inital power on just to set the baseline.

```
   TOTAL VARIANT
   13942 RegStateValues
     534 ContinueBitstreamLoad
     112 CPUPresent
       3 SetState(InitialPowerOn)
       1 SetState(Overheat)
       1 FpgaInit
       1 WaitForDone
       1 Programmed
       1 Startup
       1 SequencerInterrupt
       1 Coretype
       1 PmbusAlert
 NDX LINE      GEN    COUNT PAYLOAD
...
 62  468        6        1 SetState { prev: Some(A0), next: A0PlusHP, why: InitialPowerOn, now: 0x231fa }
  63  442        6       97 RegStateValues { seq_api_status: SeqApiStatusView { a0_sm: Ok(Done) }, seq_raw_status: SeqRawStatusView { hw_sm: 0xe }, nic_api_status: NicApiStatusView { nic_sm: Ok(Done) }, nic_raw_status: NicRawStatusView { hw_sm: 0x6 } }
  64  442        6        1 RegStateValues { seq_api_status: SeqApiStatusView { a0_sm: Ok(Done) }, seq_raw_status: SeqRawStatusView { hw_sm: 0xe }, nic_api_status: NicApiStatusView { nic_sm: Ok(NicReset) }, nic_raw_status: NicRawStatusView { hw_sm: 0x6 } }
  65  442        6      171 RegStateValues { seq_api_status: SeqApiStatusView { a0_sm: Ok(Done) }, seq_raw_status: SeqRawStatusView { hw_sm: 0xe }, nic_api_status: NicApiStatusView { nic_sm: Ok(Done) }, nic_raw_status: NicRawStatusView { hw_sm: 0x6 } }
[...]

PR #2386, opened 2026-02-12
thermal: distinguish between real and made up temperatures 

hawkw:
Presently, the `thermal` task's ringbuf attempts to gaslight the reader
by presenting totally fake temperatures as though they were real
measurements from a sensor. This is because the thermal loop contains
logic which attempts to extrapolate a worst-case temperature projection
based on the last temperature reading received from a device and the
time which has elapsed since that reading:
https://github.com/oxidecomputer/hubris/blob/ddfb8782336a86cf5c04e48eedc05d6a0ee25756/task/thermal/src/control.rs#L508-L526

This logic is not _wrong_, per se. Extrapolating a worst-case
temperature is the correct behavior in the absence of sensor data, even
when it leads to a thermal shutdown which is not actually necessary: the
thermal loop is correct to assume the worst in the absence of real data.
However, the way we presently present these fake temperatures in the
ringbuf is *very* confusing to the reader, as they appear to be real
life temperature measurements that actually came from the device.
Recently (see https://github.com/oxidecomputer/hubris/issues/2385), this reporting of fake temperatures as though they
were real lead @nathanaelhuffman and I to hypothesize about a
manufacturing defect with T6 heatsinks when the actual issue was just
that we couldn't read the NIC's temperature sensor as it had MAPO-ed a
while ago (see https://github.com/oxidecomputer/hubris/issues/2384).

This commit is my attempt to make the ringbuf entries less misleading.
Now, in addition to recording the worst case estimate that lead to a
thermal control state transition (to the `Critical` or `Uncontrollable`
states), we *also* record the real temperature reading from the device
alongside it. The names of the ringbuf fields now attempt to indicate
which is a real temperature and which is a worst-case estimate.

Note that the fake temperatures are only presented to a reader as though
they were real in the thermal loop's ringbuf; they are _not_ sent to the
`sensors` task and will not be reported externally through the
`control-plane-agent`'s `component_details` or `read_sensor_value` APIs.
Instead, we will just report `nodata_now` to the `sensors` task when
encountering an error reading a sensor measurement --- which is correct.
See:
https://github.com/oxidecomputer/hubris/blob/ddfb8782336a86cf5c04e48eedc05d6a0ee25756/task/thermal/src/control.rs#L881-L901

So, only the ringbuf must be changed to make this less misleading.

Fixes https://github.com/oxidecomputer/hubris/issues/2385


task/thermal/src/control.rs
```diff
@@ -508,6 +508,24 @@ struct TimestampedTemperatureReading {
...
 }
 
+/// Represents a worst-case temperature reading from the thermal model,
+/// including the estimated temperature and the time since the last actual
+/// sensor reading (lag).
+#[derive(Copy, Clone, PartialEq)]
+pub(crate) struct WorstCaseTemperature {
+    /// The worst-case temperature estimate from the thermal model, projected
+    /// from the `last_reading`.
+    worst_case_temp: Celsius,
```
hawkw, line 518: https://www.youtube.com/watch?v=GM-e46xdcUo

task/thermal/src/main.rs
```diff
@@ -82,11 +82,49 @@ enum Trace {
...
+    #[count(skip)]
+    LastRealTemperature {
+        sensor_id: SensorId,
+        /// The most recent real life (not fake) temperature measurement from
+        /// the sensor.
         temperature: units::Celsius,
+        /// The (approximate) time, in seconds, since the real life temperature
+        /// measurement was received.
+        age_s: f32,
     },
```
hawkw, line 128: I split this into its own ringbuf entries because adding two more 32-bit fields to `CriticalDoTo` and `PowerDownDueTo` would have probably made them the largest ringbuf entry and made each ringbuf slot a word bigger, which I wanted to avoid.

labbott:
It's pronounced "jaslighting" and "kelcius"

task/thermal/src/main.rs
cbiffle: I was going to ask this exact question, good anticipation! :-)

PR #2387, opened 2026-02-12
Publish SP releases with Brussels

emilyalbini:
This PR changes the Hubris release process to rely on Brussels to create the GitHub release. **Using the sprot-release tool is still needed for now**, as I haven't integrated Brussels with the Omicron releng tooling (that's my next task). We can still start using the Brussels logic for creating the release.

Most of the changes are actually in oxidecomputer/brussels#8, this just hooks things up in Hubris. I tested the release workflow on my test repo yesterday and it seemed to work, so hopefully this won't cause much breakage.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,7 @@ jobs:
         name: Download Brussels
         run: |
           if [[ "${BRUSSELS_RUN_ID}" == "" ]]; then
-            gh release download -R oxidecomputer/brussels -p brussels v0.2.0
+            gh release download -R oxidecomputer/brussels -p brussels v0.3.0
           else
             gh run download "${BRUSSELS_RUN_ID}" -R oxidecomputer/brussels -n prebuilt-binary
           fi


.github/workflows/release.yml
```diff
@@ -93,7 +119,7 @@ jobs:
         merge-multiple: true
 
     - name: Publish the release
-      uses: softprops/action-gh-release@v1
+      uses: softprops/action-gh-release@v2
```
labbott, line 122: What's the difference from v1 to v2?

build/xtask/src/gha_prepare_artifacts.rs
```diff
@@ -0,0 +1,189 @@
...
+                );
+            }
+            Ok(hashes)
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct Bundle {
```
labbott, line 164: Guessing none of these are in a crate we can easily use?

.github/workflows/release.yml
emilyalbini: It makes it work with immutable releases.

In v1, the action would create the release in a published state and then add artifacts to it, which breaks once immutable releases are enabled. v2 uses the correct behavior of creating the release in a draft state, adding artifacts to it, and then marking the release as published (turning it immutable).

build/xtask/src/gha_prepare_artifacts.rs
emilyalbini: All of the crates defining the types I'm aware of (including my sigstore-verify crate) also bring in a bunch of crypto dependencies to do the actual sigstore validation, which would needlessly slow down xtask :frowning_face:

Issue #2398, opened 2026-02-24
Add fan hotswap PG's to sequencer ring buf


nathanaelhuffman:
We don't allow sequencing from A2 if we don't have PGs from the 3 fan controllers.  We don't currently report these pin states in a sequence timeout so we just timeout in IDLE since the FPGA didn't move. We should report more info here.

This plays into https://github.com/oxidecomputer/hubris/issues/2372 as well

hawkw:
These should probably also be included in an ereport when we emit one on sequencing timeouts...

nathanaelhuffman:
Yep definitely, and I think if they're not happy we could just not start sequencing and report that rather than waiting for the timeout since we know that apriori

labbott:
Should be fixed by #2436

Issue #2394, opened 2026-02-19
Cosmo sequencer PMBus monitoring ought to include more rails


hawkw:
The other day we saw a Cosmo fail to sequence due to a timeout while waiting for the Group A rails. Looking at this sled's ringbuf we did not see `POWER_GOOD` on any of `V1P5_RTC`, `V3P3_SP5_A1`, or `V1P8_SP5_A1` rails.

Issue #2372, opened 2026-01-29
hubris should have more knowledge of sequencing failures when trying to go to A0


nathanaelhuffman:
Rather than just report all of the rail status bits and hardware state machine states, it should also provide information about which rails we were expecting given the current state to make diagnosis easier.  This is straight forward and all of the information required is available in fpga registers if hubris knew which ones were expected in each state

nathanaelhuffman:
You can find the cosmo sequence here for reference: https://github.com/oxidecomputer/quartz/blob/b9428476b5903b87953d8e82755706a53ae88b92/hdl/projects/cosmo_seq/sequencer/a1_a0_seq.vhd#L243

we pause at GROUP_A_PG_AND_WAIT for the following rails:
`v1p5_rtc.pg, 3p3_sp5_a1.pg and v1p8_sp5_a1.pg`

we pause at SLP_CHECKPOINT for two bits of handshake from the AMD CPU and the DDR bulk HSCs power good:
`abcdef_hsc.pg  power_group.ghijkl_hsc.pg` and 
```
slp_s3_l = '1' from AMD CPU
slp_s5_l = '1' from AMD CPU
```

we pause at GROUP_B_PG_AND_WAIT for the following rails:
`v1p1_sp5.pg`

we pause at GROUP_C_PG_AND_WAIT for the following rails:
```
power_group.vddio_sp5_a0.pg and 
power_group.vddcr_cpu1.pg and
power_group.vddcr_cpu0.pg and
power_group.vddcr_soc.pg
```
we pause at WAIT_PWROK for the PWROK from AMD CPU

we pause at WAIT_RESET_L_RELEASE for the RESET_L release from the AMD CPU.

We should have live visibility to all of these  rails and bits in FPGA registers, so when we timeout in one of these states we should report those that were not as expected to make diagnosis

mkeeter:
See also https://github.com/oxidecomputer/quartz/blob/ndh/a0-fault-tree/hdl/projects/cosmo_seq/sequencer/docs/a0_sequencing_fault_tree.md

labbott:
Should be addressed by https://github.com/oxidecomputer/hubris/pull/2436

Issue #2364, opened 2026-01-22
unfortunately, a way to turn a SensorId into a refdes is probably required


hawkw:
since the `thermal` task deals (exclusively?) in `SensorId`, and i would really like to be able to have a refdes to use when making an ereport to say which thing it was that was over it's critical temp. putting a `SensorId` in an ereport is of course unacceptable since the control plane doesn't really  know what they are, and we should not have two different ways to identify devices upstack any more than we already do. also they can change if you reorder stuff in the TOML which would be sad. 

therefore we will need to have a thingy that takes a sensor ID and gives a refdes back. codegen will undoubtedly be involved.

hawkw:
actually, hum, perhaps not; we just have to plumb it through from the BSP as a constant, or get it from the `I2cDevice`. That's not as bad.

hawkw:
> actually, hum, perhaps not; we just have to plumb it through from the BSP as a constant, or get it from the `I2cDevice`. That's not as bad.

**UPDATE**: I was wrong about this, it _is_ necessary in the case of dynamic inputs, which the `thermal` task knows about only as `SensorId`s and not as `I2cDevice`s. So, since the current `thermal` code kind of abstracts over the possibility that some things will be dynamic and other things will be `I2cDevice`s that it owns by just dealing in sensor IDs, it's probably better to have a sensor ID to refdes mapping. So we do need this.

hawkw:
#2383 includes a codegen change to support this. I have to make up my mind a bit about how I want to do it and whether it's worth generating the LUT only for non-I2C sensors to save RAM...

PR #2399, opened 2026-02-24
Add shared ereport types crate, more sequencer ereports

### LINKED ISSUE #2142: ereport: sequencing failure
There are conditions under which the Gimlet (and Cosmo?) sequencers will fail permanently. This should generate an ereport.

hawkw:
Presently, the compute sled sequencer tasks (`cosmo_seq` and
`gimlet_seq`) duplicate a large number of type definitions for ereport
messages. This is due to our present practice of defining a task's
ereports as one big enum of all the ereports it may send, which is
almost always unique to that task. This is unfortunate, as it means that
we must duplicate these messages when we would like them to be
consistent between multiple tasks.

Furthermore, the sequencer tasks currently only emit ereports for a
small subset of events that we care about --- mostly V<sub>core</sub>
regulator PMBus alerts. There are several other events that should
definitely be reported. I had previously attempted to add some new
ereports to the sequencer tasks in #2242, but at the time, neither
@rmustacc nor I were particularly satisfied with the content and class
hierarchy of those ereports, and I was unhappy with the fact that these
ereports added additional duplicated code between the two sequencer
tasks.

This PR is a second attempt to add some of the sequencer ereports I
would like, using the new API added in #2397 to define those ereport
messages in a shared crate, rather than duplicating them between the two
sequencer tasks. In particular, I've done the following:

1. Introduce a new `lib/ereports` crate that contains shared ereport
   message type definitions, using the `#[ereport(...)]` attribute from
   #2397 to define the ereports as separate types.
2. Refactor the existing `gimlet_seq` and `cosmo_seq` ereports to move
   the message types for PMBus alerts and BMR491 mitigation failures to
   the `ereports` crate, and use them in both sequencer tasks.
3. Also refactor the `psc_seq` task to use the shared definition of
   `PmbusStatus` registers. I didn't change the actual ereport messages
   there, as they're pretty different from the CPU sequencer ones.
4. Added tracking of when the sequencer transitioned to the current
    power state, so that this can be included in some of the ereports, 
    as @rmustacc suggested in [this comment][1].
5. Add new ereports for the following events:
   + The CPU asserting its `THERMTRIP_L` pin
   + The `CPU_PRESENT` pin is not asserted
   + The CPU asserting its `SMERR_L` pin (on Cosmo, as I'm not sure if
     the Gimlet sequencer task can notice this)
   + Failure to sequence due to an unsupported CPU type (`coretype` and
     `sp3r{x}`/`sp5r{x}` pins)
6. Assorted refactoring of the ereport code in both sequencers

I did _not_ add ereports for MAPO or NIC MAPO events, as per a
conversation on Matrix with @rmustacc, I think we will want to report
MAPOs by adding better monitoring of `POWER_GOOD` status from various
power rails. Some of that is described in issues #2372, #2394, and
perhaps also #2398.

This does _not_ actually close #2142, as that describes sequencing
failures due to timeouts waiting for `POWER_GOOD` on various rails,
which I will want to implement subsequently.

Closes #2242.

[1]: https://github.com/oxidecomputer/hubris/pull/2242#discussion_r2418273766


drv/cosmo-seq-server/src/main.rs
```diff
@@ -219,17 +221,15 @@ fn main() -> ! {
...
-                EreportKind::Bmr491MitigationFailure {
-                    refdes: FixedStr::from_str(dev.component_id()),
-                    failures,
-                    last_cause,
-                    succeeded,
-                },
-            );
+            let ereport = ereports::pwr::Bmr491MitigationFailure {
+                refdes: FixedStr::<{ REFDES_LEN }>::from_str(
+                    dev.component_id(),
```
hawkw: It's a bit annoying that these are generic over `REFDES_LEN`, since it means that the downstream task must specify it both in the `max_cbor_len_for!` macro _and_ when actually constructing the `FixedStr`s (as the type is not inferred there, since it could be _any_ `FixedStr`). 

The alternative to this would be making the `ereports` crate depend on `build_i2c`, so that it can get the `MAX_COMPONENT_ID_LEN` value from there. But, that felt a bit gross for the lib that just defines a bunch of types to have to go build all the I2C stuff. Also, there might be some cases where a task needs to override the length in order to use a refdes that _isn't_ coming from the I2C config. For instance, when we wire up the "look up refdes for a sensor ID" stuff in #2364 (which i started on in #2383), a sensor which is not an I2C device may have a longer refdes than any of the I2C devices in the config (unlikely but who knows!). Hmm.

hawkw:
Testing out the `SMERR_L` and `THERMTRIP` IRQ ereports on Cosmo, by manually poking the sequencer FPGA Interrupt Generation Register (IGR):

```console
eliza@jeeves ~ $ pfexec humility hiffy -c FmcDemo.poke32 -a addr=0xc0000208,value=0x2
humility: WARNING: archive in environment variable overriding archive in environment file
humility: attached to 1fc9:0143:ISS1EG0OBG2HT via CMSIS-DAP
FmcDemo.poke32() => ()
eliza@jeeves ~ $ pfexec humility hiffy -c Sequencer.get_state
humility: WARNING: archive in environment variable overriding archive in environment file
humility: attached to 1fc9:0143:ISS1EG0OBG2HT via CMSIS-DAP
Sequencer.get_state() => A0Thermtrip
eliza@jeeves ~ $ pfexec humility hiffy -c Sequencer.set_state -a state=A2
humility: WARNING: archive in environment variable overriding archive in environment file
humility: attached to 1fc9:0143:ISS1EG0OBG2HT via CMSIS-DAP
Sequencer.set_state() => Changed
eliza@jeeves ~ $ pfexec humility hiffy -c Sequencer.set_state -a state=A0
humility: WARNING: archive in environment variable overriding archive in environment file
humility: attached to 1fc9:0143:ISS1EG0OBG2HT via CMSIS-DAP
Sequencer.set_state() => Changed
eliza@jeeves ~ $ pilot sp -i e1000g0 exec -e 'ereports' fe80::aa40:25ff:fe04:11c5
Feb 26 18:46:53.957 INFO creating SP handle on interface e1000g0, component: faux-mgs
Feb 26 18:46:53.988 INFO initial discovery complete, addr: [fe80::aa40:25ff:fe04:11c5%2]:11111, interface: e1000g0, socket: control-plane-agent, component: faux-mgs
restart ID: 7b2025d6-239d-9da0-5ddd-d6b13b6fb64b
count: 2

ereports:
0x1: {
    "baseboard_part_number": String("913-0000023"),
    "baseboard_rev": Number(2),
    "baseboard_serial_number": String("2J80WF9K\0\0\0"),
    "hubris_caboose": Object {
        "board": String("cosmo-b"),
        "commit": String("579b66a6f5c2a940073f4574228df91ca1f69306-dirty"),
        "version": Null,
    },
    "hubris_task_gen": Number(0),
    "lost": Null,
}

0x2: {
    "baseboard_part_number": String("913-0000023"),
    "baseboard_rev": Number(2),
    "baseboard_serial_number": String("2J80WF9K\0\0\0"),
    "dev_id": String("sp5-host-cpu"),
    "hubris_caboose": Object {
        "board": String("cosmo-b"),
        "commit": String("579b66a6f5c2a940073f4574228df91ca1f69306-dirty"),
        "version": Null,
    },
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("cosmo_seq"),
    "hubris_uptime_ms": Number(568146),
    "k": String("hw.cpu.amd.thermtrip"),
    "refdes": String("P0"),
    "state": Object {
        "cur": String("A0PlusHP"),
        "since": Number(314536),
    },
    "v": Number(0),
}

eliza@jeeves ~ $ pfexec humility hiffy -c Sequencer.get_state
humility: WARNING: archive in environment variable overriding archive in environment file
humility: attached to 1fc9:0143:ISS1EG0OBG2HT via CMSIS-DAP
Sequencer.get_state() => A0
eliza@jeeves ~ $ pfexec humility hiffy -c FmcDemo.poke32 -a addr=0xc0000208,value=0x4
humility: WARNING: archive in environment variable overriding archive in environment file
humility: attached to 1fc9:0143:ISS1EG0OBG2HT via CMSIS-DAP
FmcDemo.poke32() => ()
eliza@jeeves ~ $ pilot sp -i e1000g0 exec -e 'ereports' fe80::aa40:25ff:fe04:11c5
Feb 26 18:47:39.106 INFO creating SP handle on interface e1000g0, component: faux-mgs
Feb 26 18:47:39.137 INFO initial discovery complete, addr: [fe80::aa40:25ff:fe04:11c5%2]:11111, interface: e1000g0, socket: control-plane-agent, component: faux-mgs
restart ID: 7b2025d6-239d-9da0-5ddd-d6b13b6fb64b
count: 3

ereports:
0x1: {
    "baseboard_part_number": String("913-0000023"),
    "baseboard_rev": Number(2),
    "baseboard_serial_number": String("2J80WF9K\0\0\0"),
    "hubris_caboose": Object {
        "board": String("cosmo-b"),
        "commit": String("579b66a6f5c2a940073f4574228df91ca1f69306-dirty"),
        "version": Null,
    },
    "hubris_task_gen": Number(0),
    "lost": Null,
}

0x2: {
    "baseboard_part_number": String("913-0000023"),
    "baseboard_rev": Number(2),
    "baseboard_serial_number": String("2J80WF9K\0\0\0"),
    "dev_id": String("sp5-host-cpu"),
    "hubris_caboose": Object {
        "board": String("cosmo-b"),
        "commit": String("579b66a6f5c2a940073f4574228df91ca1f69306-dirty"),
        "version": Null,
    },
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("cosmo_seq"),
    "hubris_uptime_ms": Number(568146),
    "k": String("hw.cpu.amd.thermtrip"),
    "refdes": String("P0"),
    "state": Object {
        "cur": String("A0PlusHP"),
        "since": Number(314536),
    },
    "v": Number(0),
}

0x3: {
    "baseboard_part_number": String("913-0000023"),
    "baseboard_rev": Number(2),
    "baseboard_serial_number": String("2J80WF9K\0\0\0"),
    "dev_id": String("sp5-host-cpu"),
    "hubris_caboose": Object {
        "board": String("cosmo-b"),
        "commit": String("579b66a6f5c2a940073f4574228df91ca1f69306-dirty"),
        "version": Null,
    },
    "hubris_task_gen": Number(0),
    "hubris_task_name": String("cosmo_seq"),
    "hubris_uptime_ms": Number(714333),
    "k": String("hw.cpu.amd.smerr"),
    "refdes": String("P0"),
    "state": Object {
        "cur": String("A0"),
        "since": Number(617259),
    },
    "v": Number(0),
}

eliza@jeeves ~ $ pfexec humility hiffy -c Sequencer.get_state
humility: WARNING: archive in environment variable overriding archive in environment file
humility: attached to 1fc9:0143:ISS1EG0OBG2HT via CMSIS-DAP
Sequencer.get_state() => A2
eliza@jeeves ~ $
```

hawkw:
@rmustacc I've requested your review on this mostly just for any feedback on the new ereport messages and their contents, since the new ereports are based on your suggestions from #2242. Hopefully, now that we have a shared library with most of the ereport type definitions, it should be a bit easier to review just the ereport messages, without having to wade through a whole bunch of implementation details moving Rust around. In particular, you can see the complete list of all the ereport classes that each sequencer task will emit here (for Gimlet):

https://github.com/oxidecomputer/hubris/blob/092f4949a3e2a074f6bf4412edd45a070e71e399/drv/gimlet-seq-server/src/main.rs#L1601-L1606

and here (for Cosmo):

https://github.com/oxidecomputer/hubris/blob/092f4949a3e2a074f6bf4412edd45a070e71e399/drv/cosmo-seq-server/src/main.rs#L1215-L1221

and you should be able to find most of these Rust types defined in this library, except for a few whose content is specific to the Cosmo or Gimlet sequencer:

https://github.com/oxidecomputer/hubris/tree/092f4949a3e2a074f6bf4412edd45a070e71e399/lib/ereports/src

Please don't feel like you need to review anything beyond how we feel about the ereports, their contents, and the class hierarchy and naming --- thanks in advance for taking a look!

drv/cosmo-seq-server/src/main.rs
```diff
@@ -648,26 +607,33 @@ impl ServerImpl {
...
+                    self.seq.power_ctrl.modify(|m| m.set_a0_en(false));
+                    let ereport = ereports::cpu::UnsupportedCpu {
+                        cpu: &HOST_CPU_REFDES,
+                        cpu_type: CpuTypeBits {
+                            coretype: [coretype0, coretype1, coretype2],
+                            sp5rx: [sp5r1, sp5r2, sp5r3, sp5r4],
+                            coretype_ok,
+                            sp5rx_ok,
+                        },
+                    };
```
labbott, line 639: I was having a thought of "do we need something to compare against to avoid guess and check about what the bits should be" but based on what we've heard from manufacturing any error here is treated as "back to the inspection bin with ye" so I guess this is fine

drv/cosmo-seq-server/src/main.rs
```diff
@@ -962,7 +929,10 @@ impl ServerImpl {
             self.seq.ifr.modify(|h| h.set_thermtrip(true));
             ringbuf_entry!(Trace::Thermtrip);
             action = InternalAction::ThermTrip;
-            // Great place for an ereport?
+            self.ereporter.try_send_ereport(&ereports::cpu::Thermtrip {
+                cpu: &HOST_CPU_REFDES,
+                state: self.ereport_current_state(),
+            });
```
labbott, line 942: Is the expectation we'll get other ereports independently? If we're tripping the CPU thermtrip I'm also expecting us to get reports from the thermal task too?

drv/cosmo-seq-server/src/main.rs
labbott, line 950: Guessing this is still WIP?

lib/ereports/src/cpu.rs
```diff
@@ -0,0 +1,51 @@
...
+/// An ereport representing an AMD CPU's `THERMTRIP` assertion.
+#[derive(Clone, Encode)]
+#[ereport(class = "hw.cpu.amd.thermtrip", version = 0)]
+pub struct Thermtrip {
+    #[cbor(flatten)]
+    pub cpu: &'static HostCpuRefdes,
+    pub state: crate::pwr::CurrentState,
+}
+
+/// An ereport representing an AMD CPU's `SMERR_L` assertion.
```
labbott, line 19: Have you been able to test generating this? Last time I looked the AMD docs were very clear you should handle it but not how to trigger it.

drv/cosmo-seq-server/src/main.rs
hawkw: Yes, that's the eventual goal! This one is meant to _just_ tell us that we saw the `THERMTRIP` pin asserted by the host CPU, which is all we really know about here. The thermal task may or may not be independently aware of a THERMTRIP assertion (we really expect the thermal task to also independently try to power down the system when the host CPU is reaching its thermal limits), so if it did something, we might _not_ get a THERMTRIP ereport and may instead get some other thing. Eventually we will also expect the control plane to try and look at temperatures upon receiving this ereport.

lib/ereports/src/cpu.rs
hawkw: Soooo...thanks to @nathanaelhuffman, the FPGA now has an Interrupt Generation Register (IGR) with bits corresponding to those in the IFR and IER. We can poke bits in that register and cause the FPGA to raise the corresponding interrupt to the SP (setting the IFR bit and, if the IER flag enables the interrupt, raising the corresponding IRQ). This does _not_ cause the host CPU to actually assert that pin, and does not cause whatever else the FPGA does if the actual input pin that triggers the interrupt is asserted.

If you take a look at the testing I did in https://github.com/oxidecomputer/hubris/pull/2399#issuecomment-3968546786, you can see that I was able to poke the FPGA into simulating both the THERMTRIP and SMERR interrupts and that we got the expected ereports out of the sequencer task in both cases. Neither of those really simulate anything _else_ that may or may not have happened if the host CPU actually asserted those pins. In this case, that isn't a big deal, since (AFAIK) those pins are basically just muxed by the FPGA on our behalf and don't cause it to go and do anything else.

labbott: There's a part of me that still wants to see the CPU SMERR but this is really great testing and I'm glad to know it works

drv/cosmo-seq-server/src/main.rs
hawkw: Yeah, I mentioned in the PR description that per a discussion with rm, we probably want to communicate about MAPOs by monitoring the state of POWER_GOOD on various rails, and that it'll be a bit more complex than just sending a simple ereport when this IRQ (and the NIC MAPO one) fires. I didn't do that as part of this PR as I wanted to get both the simple ones and some of the refactoring in without also trying to figure all that stuff out.

lib/ereports/src/cpu.rs
hawkw: I would also like that! I would also like to know what it even means if the CPU does SMERR...

drv/cosmo-seq-server/src/main.rs
rmustacc: We have exactly what we want the bits to be at +612 for the coretype and only a single combination for the sp5rx bits at +620, so I don't think we are guessing and checking? Or at the least we do know from the AMD docs exactly what to look for.

But what we are missing is basically an ereport for CPU missing. That is, with a lot of these failures I don't think we're going to get to this point as we will ultimately also fail with `CpuNotPresent` so I think we may want to add a second ereport for that case ala `hw.cpu.miss`, perhaps something a bit more general for a missing component, or something. I'm not sure that needs to be part of this change.

lib/ereports/src/pwr.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub word: Option<u16>,
+    pub input: Option<u8>,
+    pub iout: Option<u8>,
+    pub vout: Option<u8>,
+    pub temp: Option<u8>,
+    pub cml: Option<u8>,
+    pub mfr: Option<u8>,
+}
+
+/// Represents the current power state, and how long we have been in that state.
```
rmustacc: I wonder if it's worth commenting to folks that the expectation is that this should be used in fields before we transition. That is, no one wants to see we've been in A2 for 2 ticks, but rather the long time that it was in A0 (hopefully) ahead of that.

lib/ereports/src/pwr.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Copy, Clone, Encode)]
+pub struct CurrentState {
+    /// The current CPU power state.
+    pub cur: drv_cpu_power_state::PowerState,
+    /// The Hubris tick (in milliseconds) at which the transition to this state
```
rmustacc: Are we going to promise that this is in ticks or that it is converted to milliseconds? While I don't think the `hubris` kernel will change here I think we want to phrase this in a unit that's independent of the tick rate so that way software can interpret it the same way but just know that if for some reason that ever changes for whatever reason (some new tick source that's higher-fidelity internal to an SoC for some reason that I can't imagine), that this doesn't change.

lib/ereports/src/cpu.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct Smerr {
+    #[cbor(flatten)]
+    pub cpu: &'static HostCpuRefdes,
+    pub state: crate::pwr::CurrentState,
+}
+
+/// An ereport representing an unsupported CPU.
+#[derive(Clone, Encode)]
+#[ereport(class = "hw.cpu.unsup", version = 0)]
+pub struct UnsupportedCpu<T: EncodeFields<()>> {
```
rmustacc: I think a thing for us to talk through with this is how do consumers end up normalizing this information? In particular, it seems like they'll need to look at the board type to understand what is expected so they can know what to compare against. I think that might be okay. But SP3 systems have only a single bit coretype and a 2 bit SPxRy field. Where as SP5 has 3 and 4 bits respectively. SP7/8 look more like SP5 than SP3.

I have convinced myself it's probably fine to end up in that state, but I want to make sure we all realize that consumers will have to deal with the fact that the contents of `cpu_type` will vary based on the CPU generation and will have to get external knowledge to figure out what is relevant there (perhaps using the `dev_id` or something else).

drv/gimlet-seq-server/src/vcore.rs
```diff
@@ -342,20 +329,16 @@ impl VCore {
...
-                pwr_good,
-                pmbus_status: status,
-            },
-        );
+        let ereport = PmbusAlert {
+            refdes: FixedStr::<{ crate::REFDES_LEN }>::from_str(
+                self.device.i2c_device().component_id(),
+            ),
+            rail: RAIL,
+            time: now,
```
rmustacc, line 337: Seeing `time` here vs. `since` makes me wonder if there is a general value to always just having us have to pass in a `time` that's distinct from the `hubris_uptime_ms` as I assume this is trying to represent detection time whereas the `hubris_uptime_ms` is representing the kernel uptime when the ereport is generated. This may be something worth punting and not saying we should redo the PMBus Alert, just seeing it here in this context made me think about it.

drv/cosmo-seq-server/src/main.rs
```diff
@@ -1230,27 +1210,54 @@ impl NotificationHandler for ServerImpl {
...
+        refdes: fixedstr::FixedString::from_str("P0"),
+        dev_id: fixedstr::FixedString::from_str("sp5-host-cpu"),
+    };
+
+#[derive(Clone, microcbor::EncodeFields)]
+struct CpuTypeBits {
+    coretype: [bool; 3],
+    sp5rx: [bool; 4],
+    coretype_ok: bool,
+    sp5rx_ok: bool,
```
rmustacc: I went back and forth a bit on whether or not we should include these items. I guess it's probably useful to include this, but I do wonder if we want to keep the the `sp5rx` stuff more generic and not force users to end up switching this on a per-type. e.g. just `coretype`, `rev`, `coretype_ok`, `rev_ok`.

In general I think we want to always have the same set of members for a given class in an ereport payload.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -1609,31 +1596,57 @@ cfg_if::cfg_if! {
...
+];
+
+static HOST_CPU_REFDES: ereports::cpu::HostCpuRefdes =
+    ereports::cpu::HostCpuRefdes {
+        refdes: fixedstr::FixedString::from_str("P0"),
+        dev_id: fixedstr::FixedString::from_str("sp3-host-cpu"),
+    };
+
+#[derive(Clone, microcbor::EncodeFields)]
+struct CpuTypeBits {
```
rmustacc: See notes on SP5 version about how we might want to think about field naming.

lib/ereports/src/cpu.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[derive(Clone, EncodeFields)]
+pub struct HostCpuRefdes {
+    /// On both Gimlet and Cosmo, the host CPU's refdes is `P0`.
+    pub refdes: FixedString<2>,
+    /// As the host CPU's `control-plane-agent` device ID is different from its
+    /// refdes, we must include both in the ereport.
+    ///
+    /// On Gimlet, this is `sp3-host-cpu` and on Cosmo, it is `sp5-host-cpu`.
```
rmustacc, line 59: If we were to exceed either length for some reason how does it fail? Does the length here represent a contract or would the upper level contract with the control plane be that these are basically guaranteed to be character arrays, but it should not assume the size.

lib/ereports/src/cpu.rs
rmustacc: > I would also like that! I would also like to know what it even means if the CPU does SMERR...

Can you clarify what you're looking to understand? We can do some research on error injection though.

lib/ereports/src/cpu.rs
hawkw: None of the lengths of the various fixed-length strings used here are visible to the control plane, which just receives these as CBOR strings of arbitrary length. In Hubris' definitions of these types, the max lengths have to be explicitly declared so that we can calculate the size of the encoding buffer automatically. 

In this case, if someone tries to use a longer string for either field (i.e. if we did a new board where the CPU's refdes was three characters, or something), that would be a compiler error, which would basically just tell us that we needed to increase the max length in this struct, ensuring that the calculated encoding buffer size is also increased.

drv/gimlet-seq-server/src/vcore.rs
hawkw: So, I'm not 100% sure if the `time` field on this ereport is really strictly necessary. You are correct that the `time` field represents the time at which we detected the IRQ pin being asserted, versus the `hubris_uptime_ms` at which the ereport is actually delivered to `packrat`. In this case I had added it to this specific ereport as, prior to actually delivering the ereport, we do a bunch of I2C reads to get the PMBus status registers, which could conceivably take over a millisecond (or more, if there are higher-priority tasks also using the I2C driver). So, I thought there was maybe some value in including the hubris timestamp at which the IRQ was asserted, since we already had it. But, I'm also not really sure if recording the more precise timestamp is actually something we will have a use for in practice...

I think we might want to think about a generalized/conventional way to say "here is the time at which the event occurred, distinct from the time at which the ereport was reported", but I agree that we should think about that separately from this change...and as I mentioned above, I'm not sure if precisely recording that timestamp when it only differs by one or two milliseconds is actually something we need that badly.

hawkw: Also, the `since` thing will still remain separate from a mechanism for saying "this is the more precise event timestamp", since it's intended to be the time at which the prior state transition occurred.

lib/ereports/src/cpu.rs
hawkw: So, this is mainly just curiosity on my part, but... all I could find about `SMERR_L` is that the _AMD Socket SP5 Processor Motherboard Design Guide_ (56870 Rev. 1.1) says[^1] that this signal should be connected to the motherboard component that's responsible for power sequencing, and that the motherboard must reset the CPU upon its assertion, but it doesn't really explain under what conditions it might be asserted besides that it is related to "system management". I'm a bit curious as to whether we know anything beyond that since it might indicate that there's additional information that we might want to report based on an understanding of what it means.

[^1]: I'm paraphrasing what it actually says a bit since it's not public documentation.

drv/cosmo-seq-server/src/main.rs
hawkw: Hm, thinking about this, I think we either want to move this towards the generic `coretype` and `rev` where the values of those fields are socket-dependent, or having separate `hw.cpu.unsup.amd.{sp3, sp5}` or something. I'm _kinda_ inclined to go with the simpler thing of just having coretype and rev and bits saying which was the one that was unexpected, and just including the actual values of those pins mostly for human debug purposes, since I don't really imagine upstack automated diagnosis having much to say beyond "there is something wrong with this sled and it should be RMA'd" upon receipt of this ereport...

lib/ereports/src/cpu.rs
hawkw: Yeah. We could also add an explicit field indicating the socket if you think that's worthwhile, although you're correct that the `dev_id` will currently also communicate this. I also considered introducing socket-specific subclasses, but that felt unnecessary (as I discussed in https://github.com/oxidecomputer/hubris/pull/2399#discussion_r2913591592), since I'm anticipating that most of the automated diagnosis of this will probably just be "RMA the sled".

lib/ereports/src/pwr.rs
hawkw: I made this explicitly milliseconds in 981c9bdb2d7b9d4bdfb4ce1b42fe7dab585b259f --- so any future Hubris image running on a MCU with higher precision timers will end up downconverting.

lib/ereports/src/pwr.rs
hawkw: Added a comment to that effect in 981c9bdb2d7b9d4bdfb4ce1b42fe7dab585b259f.

drv/cosmo-seq-server/src/main.rs
hawkw: How do you feel about https://github.com/oxidecomputer/hubris/pull/2399/changes/f1cd8e479f72b7a74d1dd5de9d832f87d7e5ba03? I've made the fields consistent between sockets so we now always have a `coretype.ok` and ` rev.ok`, and `coretype.bits`/`rev.bits` which are arrays whose length depends on the socket but which are always arrays with the same name. That should mean upstack software doesn't need to be aware of the socket unless it really cares about the individual type/rev pins.

hawkw: Per a discussion with rm, I think we're happy with these for now.

drv/gimlet-seq-server/src/main.rs
```diff
@@ -1117,12 +1103,16 @@ impl<S: SpiServer> ServerImpl<S> {
...
     // time-critical bits to assure that we don't melt!).
     //
-    fn check_thermtrip(&mut self, ifr: u8) {
+    fn check_thermtrip(&mut self, ifr: u8, now: u64) {
         let thermtrip = Reg::IFR::THERMTRIP;
 
         if ifr & thermtrip != 0 {
             self.seq.clear_bytes(Addr::IFR, &[thermtrip]).unwrap_lite();
-            self.update_state_internal(PowerState::A0Thermtrip);
+            self.ereporter.try_send_ereport(&ereports::cpu::Thermtrip {
```
mkeeter, line 1111: Nit: this won't use `now`, because it's not written until `update_state_internal` is called.

mkeeter: Maybe that's intentional, though, to reflect the before-thermtrip state?

hawkw: yeah, that's correct, that field is meant to indicate what state we were in *when the event occurred*, not any subsequent state we might be transitioning to. There's a comment on the type explaining that.

PR #2400, opened 2026-02-24
Control plane agent for Nucleo boards

lzrd:
Improve Nucleo support for Hubris and hardware testing

The Nucleo board is a useful off-the-shelf board for prototyping and for
testing "pure Hubris" code paths. Configuring the control plane agent
into the image makes the board more useful.

The Nucleo board can also be a convenient platform for developing I2C
device drivers.

While there is enforcement in xtask to ensure that the vlan feature is
enabled for production boards, the feature results in compilation errors
if those guards are commented out and the vlan feature is removed from
the `app.toml`. That is fixed so that the Nucleo boards can opt out of
the vlan feature.

All of the changes in the machine code output are explained by line
number changes (ringbuf, etc.), the caboose `GITC` value, a different
Hubris ImageId, and one layout change due to a line number constant value
being a different size: in `control_plane_agent` in `handle_mgs_request`
a `cmp.w r0, #728` became `move r1,#729; cmp r0, r1`

While people have prototyped a Nucleo + NXP LPC55S69 xPresso board to
simulate SP + RoT, there is no intention to officially support that
configuration.

Configuration changes to Nucleo boards:
  - add `control_plane_agent` (CPA)
  - add "Caboose" for Hubris image identification
  - omitting VLAN tags on Nucleo boards
  - `update_server` to manage firmware updates
  - `sprot` is assigned the appropriate pins for connection to a non-existent RoT

`control_plane_agent` references `i2c_driver` unconditionally via
`task_slot!(I2C, i2c_driver)`, so add it to CPA's task-slots in
`app-h743.toml` (`app-h753.toml` already had it).

### DIFF
--- a/app/demo-stm32h7-nucleo/README.md
+++ b/app/demo-stm32h7-nucleo/README.md
@@ -1,1 +1,129 @@
-# STM32H7 demo application
+# STM32H7 Nucleo-144 Demo Application
+
+Hubris application configurations for the STM32H7 Nucleo-144
+development boards:
+
+- `app-h743.toml` -- NUCLEO-H743ZI (order code NUCLEO-H743ZI2)
+- `app-h753.toml` -- NUCLEO-H753ZI
+
+The Nucleo boards are readily available development boards for general
+Hubris development. Although they are not part of any Oxide Computer
+product, they are configured to work with the Management Gateway Service
+to allow over-the-network update and testing.
+
+## Board differences
+
+Both boards use the same Nucleo-144 form factor and share identical
+pin assignments. The STM32H753 includes a hardware hash accelerator
+supported by Hubris that the STM32H743 lacks.
+
+## VLAN
+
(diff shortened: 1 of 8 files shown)


app/demo-stm32h7-nucleo/README.md
```diff
@@ -1 +1,129 @@
...
+
+Hubris application configurations for the STM32H7 Nucleo-144
+development boards:
+
+- `app-h743.toml` -- NUCLEO-H743ZI (order code NUCLEO-H743ZI2)
+- `app-h753.toml` -- NUCLEO-H753ZI
+
+The Nucleo boards are readily available development boards for general
+Hubris development. Although they are not part of any Oxide Computer
+product, they are configured to work with the Management Gateway Service
```
jamesmunns, line 11: Might be worth mentioning that the "Management Gateway Service" task is "Control Plane Agent (CPA)", as it is referred to below.

jamesmunns (approved):
Overall looks reasonable, one docs note

PR #2415, opened 2026-03-03
Remap toolchain dir and check for reproducibility

mkeeter:
@hawkw noticed home directory paths leaking into a Hubris binary.

I saw the same thing:
```
➜  dist jj:(qlx) strings kernel | grep mjk
/Users/mjk/.rustup/toolchains/nightly-2025-07-20-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/bit.rs
/Users/mjk/.rustup/toolchains/nightly-2025-07-20-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/adapters/enumerate.rs
/Users/mjk/.rustup/toolchains/nightly-2025-07-20-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/fmt
```

This PR adds another path to the array of `--remap-path-prefix`, and also adds a post-build check that the user's home directory has not snuck into the task binaries.

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -334,14 +334,27 @@ mod checked_types {
 // Republish these types for widespread availability
 pub use checked_types::{ContiguousRanges, OrderedVecDeque};
 
+/// Container of named flags used when calling [`package`]
+#[derive(Copy, Clone, Debug)]
+pub struct PackageFlags {
+    pub verbose: bool,
+    pub edges: bool,
+    pub dirty_ok: bool,
+    pub skip_path_check: bool,
+}
+
 pub fn package(
-    verbose: bool,
-    edges: bool,
     app_toml: &Path,
+    flags: PackageFlags,
     tasks_to_build: Option<Vec<String>>,
-    dirty_ok: bool,
     caboose_args: super::CabooseArgs,
 ) -> Result<BTreeMap<String, AllocationMap>> {
+    let PackageFlags {
+        verbose,
+        edges,
+        dirty_ok,
+        skip_path_check,
+    } = flags;
     let cfg = PackageConfig::new(app_toml, verbose, edges)?;
 
     // Verify that our dump configuration is correct (or absent)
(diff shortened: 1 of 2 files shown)


build/xtask/src/dist.rs
```diff
@@ -497,6 +506,12 @@ pub fn package(
                 if task_can_overflow(&cfg.toml, task_name, verbose)? {
                     possible_stack_overflow.push(task_name);
                 }
+                if task_contains_home(&cfg.toml, task_name)? {
+                    bail!(
+                        "task {task_name} contains your home directory; \
+                         the build is probably not reproducible!"
+                    );
```
hawkw: Hm, I feel a bit on the fence as to whether or not we should `bail!` here. If we have non-reproduceable paths, we can still continue onwards and produce a useable image, and it seems a bit sad to say "you CANNOT build hubris if you are e.g. testing a new compiler version that introduced another place where a path might sneak in". I feel like maybe this should be something where we print a warning, which looks like a Cargo warning, and then maybe return a nonzero exit status, but still go on and make the image instead of giving up immediately.

build/xtask/src/dist.rs
```diff
@@ -1045,6 +1060,25 @@ fn build_task(cfg: &PackageConfig, name: &str) -> Result<()> {
         .context(format!("failed to build {name}"))
 }
 
+/// Returns `Ok(true)` if the given task contains the byte sequence `/home/`
```
hawkw: Seems like it doesn't actually do this --- it looks like it's actually looking specifically for *the current user's actual home directory path as determined based on the operating system from which the build is being performed".

mkeeter: Yup, fixed!

build/xtask/src/dist.rs
mkeeter: I'm also on the fence, but leaning towards being strict: if someone is updating the compiler, we don't want them to be able to skip fixing this, and I don't think it will misfire under common circumstances.

hawkw: Hmm...maybe there could be an explicit `--emergency-skip-path-check` or something for use in situations where it is necessary to build an image and defer fixing this? 

On the other hand, since the xtask is right here in the repo, I suppose the BREAK IN CASE OF EMERGENCY solution could be to comment out the path check. But, thinking about it for a bit, that seems _worse_ to encourage, as it will also pass CI and it's possible for reviewers to miss that it has been commented out, and for the committer to forget. I feel like this is especially easy to accidentally if you are in the midst of a toolchain update which, from past experience, often results in a diff that is both Huge and Mostly Boring, so the author has been working on it for a long time and forgot that they commented out some check, while the reveiewer's eyes may have reasonably glazed over by the time they've scrolled to that point. So, having the way you skip the check be passing a CLI option that CI does _not_ pass actually feels a lot less scary, once I've thought through it a bit.

mkeeter: I added a `--dangerously-skip-path-check` to `xtask dist` and `xtask flash` in 411aa7535a05, since they're the most common ways to produce an archive.

In this commit, other subcommands which produce an archive (`gdb`, `sizes`, etc) require the path check to pass; subcommands which do not (`build`) do not.  I figure – even in our hypothetical situations – no one is iterating with `cargo xtask gdb`.

PR #2421, opened 2026-03-03
Change conditions for VSC8504 to be listed as down

mkeeter:
In https://github.com/oxidecomputer/facade/issues/386 , we have seen the MAC side of the VSC8504 (i.e. the side connected to the VSC7448 over QSGMII) report its status as **down**.  This counts as a racktest failure and requires a Sidecar reset or power-cycle.

I think this is a spurious failure report.

I can reproduce a similar situation on the benchtop by exposing the `phy_vsc8504_init` function to hiffy, then hitting it: each time I hit it, a random subset of ports 40-43 (i.e. the 4x SGMII ports broken out by the VSC8504) show up as **down** in `humility monorail status`.

However, packets are still transiting the VSC8504 in both directions!  The attached PSC's `rx` counters are going up, as are the port 40 `rx` counters on the VSC7448.

Clearly, our decision about whether the port is **down** is not totally accurate.  How is this decision made?

There are a few status bits that we _could_ check here: in the MAC PCS status register, there are bits for both comma detect (`mac_pcs_sig_detect`) and interface link status (`mac_link_status`).

In our existing code, we think the port is down if `mac_pcs_sig_detect` is 0.  There's also a suspicious comment saying that we've seen packets making their way through the system even when `mac_link_status` is 0, so we don't check it:

https://github.com/oxidecomputer/hubris/blob/1427344de0ff884bbe55e085cb38c5b136be3dd5/task/monorail-server/src/server.rs#L465-L470

However, when I triggered this behavior, I saw the opposite: `mac_pcs_sig_detect` is 0, and `mac_link_status` is 1 (and packets continue to flow).

There are two possibilities here:

- Packets can flow if *either* `mac_link_status` or `mac_pcs_sig_detect` is 1
- Our previous code was wrong and we just need `mac_link_status` to be 1

This PR assumes the former: the port will now show up as **down** only if *both* the flags are zero.

We also had to adjust the QSGMII valid bits in `MAC_SERDES_STATUS` to accept QSGMII sync *or* MAC comma detect, instead of requiring both bits to be set – in my test setup, I saw just the QSGMII sync bit set.

With this change, in my benchtop testing, `humility monorail status` reports the link is **up** with just the QSGMII valid / link status bit set.  If I artificially break the link (by unconfiguring port 40), then the link is correctly report as **down**.

### DIFF
--- a/task/monorail-server/src/bsp/sidecar_bcd.rs
+++ b/task/monorail-server/src/bsp/sidecar_bcd.rs
@@ -11,7 +11,7 @@ use userlib::{hl::sleep_for, task_slot, UnwrapLite};
 use vsc7448::{
     config::Speed, miim_phy::Vsc7448MiimPhy, Vsc7448, Vsc7448Rw, VscError,
 };
-use vsc7448_pac::{DEVCPU_GCB, HSIO, VAUI0, VAUI1};
+use vsc7448_pac::{phy, DEVCPU_GCB, HSIO, VAUI0, VAUI1};
 use vsc85xx::{vsc8504::Vsc8504, vsc8562::Vsc8562Phy, PhyRw};
 
 task_slot!(SEQ, seq);
@@ -39,6 +39,11 @@ enum Trace {
     AutomaticLock,
     LockError(#[count(children)] VscError),
     UnlockError(#[count(children)] VscError),
+    Vsc8504SuspiciousMacStatus {
+        mac_serdes_pcs_status: u16,
+        mac_serdes_status: u16,
+    },
+    Vsc8504ReadError(VscError),
 }
 ringbuf!(Trace, 16, Trace::None);
 
@@ -561,6 +566,37 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
             self.link_down_at = None;
         }
 
+        // Do some logging of the VSC8504 state for debug
+        if let Err(e) = self.check_vsc8504() {
+            ringbuf_entry!(Trace::Vsc8504ReadError(e));
+        }
+
+        Ok(())
+    }
+
+    /// Check VSC8504 state, logging anything unusual
+    ///
+    /// Specifically, we log if the VSC8504 isn't reporting *both* MAC comma
+    /// detect and MAC interface link up.  In practice, we still see packets
+    /// transiting the system (in both directions) with either of these bits
+    /// cleared, but it's a weird edge case that we might want to see in a
+    /// system dump.  See `facade#386` for additional context.
+    fn check_vsc8504(&self) -> Result<(), VscError> {
+        let mut phy_rw = Vsc7448MiimPhy::new(self.vsc7448.rw, 0);
+        let phy_port = 4; // The VSC8504 uses ports 4-7
+        let phy = vsc85xx::Phy::new(phy_port, &mut phy_rw);
+        let mac_serdes_pcs_status =
+            phy.read(phy::EXTENDED_3::MAC_SERDES_PCS_STATUS())?;
+        let mac_serdes_status =
+            phy.read(phy::EXTENDED_3::MAC_SERDES_STATUS())?;
+        if mac_serdes_pcs_status.mac_pcs_sig_detect() == 0
+            || mac_serdes_pcs_status.mac_link_status() == 0
+        {
+            ringbuf_entry!(Trace::Vsc8504SuspiciousMacStatus {
+                mac_serdes_pcs_status: mac_serdes_pcs_status.into(),
+                mac_serdes_status: mac_serdes_status.into(),
+            });
+        }
         Ok(())
     }
 


task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -561,6 +566,32 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
+                    || mac_serdes_pcs_status.mac_link_status() == 0
+                {
+                    ringbuf_entry!(Trace::Vsc8504MacStatus {
+                        mac_serdes_pcs_status: mac_serdes_pcs_status.into(),
+                        mac_serdes_status: mac_serdes_status.into(),
+                    });
+                }
+                Ok(())
+            })
+            .expect("port 40 always has a PHY");
```
hawkw: is panicking here and immediately restarting the task actually the proper thing to do if this apparently unreachable condition _was_ true? and, if it is, would we want to use `unwrap_lite()` with a comment so as to avoid putting a string in the binary?

task/monorail-server/src/server.rs
```diff
@@ -462,24 +462,36 @@ impl<'a, R: Vsc7448Rw> idl::InOrderMonorailImpl for ServerImpl<'a, R> {
...
+                    // mode.
+                    //
+                    // Experimentally, packets get through when either
+                    // MAC_LINK_STATUS or MAC_PCS_SIG_DETECT are set in
+                    // MAC_SERDES_PCS_STATUS (or both, obviously).  These
+                    // correspond to "QSGMII sync status", and "MAC comma
+                    // detect" in the MAC_SERDES_STATUS register.
+                    //
+                    // Despite what "ENT-AN1175" says, I don't see anything in
+                    // register 24G.  As such, we'll be optimistic: if *eithe**
```
hawkw: typos:
```suggestion
                    // register 24G.  As such, we'll be optimistic: if **either**
```

task/monorail-server/src/bsp/sidecar_bcd.rs
mkeeter: This is truly unreachable, but I decided to inline the relevant part of `phy_fn` to remove any doubt.

task/monorail-server/src/server.rs
mkeeter: Fixed!

task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -561,6 +566,37 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
             self.link_down_at = None;
         }
 
+        // Do some logging of the VSC8504 state for debug
+        if let Err(e) = self.check_vsc8504() {
+            ringbuf_entry!(Trace::Vsc8504ReadError(e));
```
labbott, line 571: It looks like we would only expect this if the phy is really dead? (I don't think there is anything to change here just curious)

task/monorail-server/src/bsp/sidecar_bcd.rs
```diff
@@ -561,6 +566,37 @@ impl<'a, R: Vsc7448Rw> Bsp<'a, R> {
...
+        let mac_serdes_status =
+            phy.read(phy::EXTENDED_3::MAC_SERDES_STATUS())?;
+        if mac_serdes_pcs_status.mac_pcs_sig_detect() == 0
+            || mac_serdes_pcs_status.mac_link_status() == 0
+        {
+            ringbuf_entry!(Trace::Vsc8504MacStatus {
+                mac_serdes_pcs_status: mac_serdes_pcs_status.into(),
+                mac_serdes_status: mac_serdes_status.into(),
+            });
+        }
```
labbott, line 599: It's not immediately obvious if this ringbuf entry is an expected or unexpected state or how much of a red flag it should be if we see it

labbott: really  I think my question is if some poor engineer is looking at this ringbuf at 11pm and sees this entry is this immediately obvious/helpful?

task/monorail-server/src/bsp/sidecar_bcd.rs
mkeeter: Yeah, or if communication to the PHY fails in some way (e.g. I2C flakiness)

task/monorail-server/src/bsp/sidecar_bcd.rs
mkeeter: I agree that it's not immediately obvious / helpful, but I'm deliberately erring on the side of logging extra information, even though we're not sure how to interpret it (because it's unclear what the PHY flags actually indicate w.r.t. packets getting through).

labbott: I phrased this poorly but my question was is there a way we can make it clearer what's going on? It might raise questions why we're seeing something called `Vsc8504MacStatus` sometimes but not others.

mkeeter: Okay, renamed to `Vsc8504SuspiciousMacStatus`

PR #2423, opened 2026-03-05
host_sp_comms: add new BootStage IPCC message

citrus-it:
The intention of this message is currently only that it ends up in the host_sp_comms ringbuf to assist in diagnosing hung sled issues. At present we are sometimes not even sure if we've reached x86.

This will form part of a broader strategy to make boot progress more observable; described further in https://github.com/oxidecomputer/stlouis/issues/936

### DIFF
--- a/task/host-sp-comms/src/main.rs
+++ b/task/host-sp-comms/src/main.rs
@@ -834,11 +834,11 @@ impl ServerImpl {
             | HostToSp::GetIdentity
             | HostToSp::GetStatus
             | HostToSp::AckSpStart
-            | HostToSp::BootStage { .. }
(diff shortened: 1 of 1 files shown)


task/host-sp-comms/src/main.rs
```diff
@@ -834,6 +834,7 @@ impl ServerImpl {
             | HostToSp::GetIdentity
             | HostToSp::GetStatus
             | HostToSp::AckSpStart
+            | HostToSp::BootStage { .. }
```
mkeeter: Nit: put this be at L841 to match enum ordering?

PR #2424, opened 2026-03-05
Read POST code ringbuf

mkeeter:
This is the Hubris side of https://github.com/oxidecomputer/management-gateway-service/pull/480

We add a new `sp5-post-codes` component, which returns a dynamic number of `PostCode` component details based on the sequencer FPGA's ringbuf.

### DIFF
--- a/idl/cpu-seq.idol
+++ b/idl/cpu-seq.idol
@@ -68,6 +68,20 @@ Interface(
             reply: Simple("u32"),
             idempotent: true,
         ),
+        "post_code_buffer_len": (
+            doc: "Read the number of POST codes in the FPGA's ringbuffer (Cosmo only)",
+            args: {},
+            reply: Simple("u32"),
+            idempotent: true,
+        ),
+        "get_post_code": (
+            doc: "Reads a POST code from the FPGA's ringbuffer (Cosmo only)",
+            args: {
+                "index": "u32",
+            },
+            reply: Simple("u32"),
+            idempotent: true,
+        ),
         "gpio_edge_count": (
             doc: "Read the GPIO edge count from the sequencer FPGA (Gimlet only)",
             args: {},
--- a/task/control-plane-agent/src/inventory.rs
+++ b/task/control-plane-agent/src/inventory.rs
@@ -37,19 +37,17 @@ impl Inventory {
         OUR_DEVICES.len() + VALIDATE_DEVICES.len()
     }
 
-    pub(crate) fn num_component_details(
+    pub(crate) fn num_component_details<F>(
         &self,
         component: &SpComponent,
-    ) -> Result<u32, SpError> {
+        our_device_lookup: F,
+    ) -> Result<u32, SpError>
+    where
+        F: Fn(&SpComponent) -> u32,
+    {
         match Index::try_from(component)? {
             Index::OurDevice(d) => {
-                match OUR_DEVICES[d].component {
-                    // The SP5 CPU can report a POST code and GPIO cycle count
-                    SpComponent::SP5_HOST_CPU => Ok(2),
-                    // The SP3 CPU can report GPIO toggle counts
-                    SpComponent::SP3_HOST_CPU => Ok(1),
-                    _ => Ok(0),
-                }
+                Ok(our_device_lookup(&OUR_DEVICES[d].component))
             }
             Index::ValidateDevice(i) => {
                 Ok(VALIDATE_DEVICES[i].sensors.len() as u32)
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -150,7 +149,7 @@ zip = { version = "0.6", default-features = false, features = ["bzip2", "deflate
 apob = { git = "https://github.com/oxidecomputer/apob", default-features = false }
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.4.0" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1" }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"], branch = "mkeeter/sp5-post-code-buffer" }
```
mkeeter: FIX THIS BEFORE MERGING

idl/cpu-seq.idol
```diff
@@ -68,6 +68,20 @@ Interface(
             reply: Simple("u32"),
             idempotent: true,
         ),
+        "post_code_buffer_len": (
+            doc: "Read the number of POST codes in the FPGA's ringbuffer (Cosmo only)",
```
hawkw, line 72: the (cosmo only) makes this somewhat feel like it could be another place where @cbiffle's thoughts about being able to serve multiple discrete IPC interfaces from the same task could be nice to have. although that is certainly not a blocker for this PR!

idl/cpu-seq.idol
```diff
@@ -68,6 +68,20 @@ Interface(
             reply: Simple("u32"),
             idempotent: true,
         ),
```
hawkw, line 70: part of me wonders if we ought to remove this IPC and replace it with something in the API crate that just implements it by asking for the `post_code_buffer_len` and then requesting the last entry based on that. but, meh.

task/control-plane-agent/src/inventory.rs
```diff
@@ -257,6 +255,14 @@ mod devices_with_static_validation {
             capabilities: DeviceCapabilities::HAS_SERIAL_CONSOLE,
             presence: DevicePresence::Present, // TODO: ok to assume always present?
         },
+        #[cfg(feature = "cosmo")]
+        DeviceDescription {
+            component: SpComponent::SP5_POST_CODES,
+            device: SpComponent::SP5_POST_CODES.const_as_str(),
+            description: "Cosmo SP5 POST code buffer",
+            capabilities: DeviceCapabilities::empty(),
+            presence: DevicePresence::Present, // TODO: ok to assume always present?
```
hawkw: hm, i think that really, it ought to be present if (and only if) the FPGA is present, which, well...if the FPGA is not present, we are having a very bad time, I guess. i _almost_ said that it would be present if the host CPU is, but, well, that's not actually the case, if there's no host CPU, we still have a buffer of POST codes, it's just empty :)

hawkw: i will add that some part of me feels a bit uncomfy about how we are modeling this as a distinct "device" rather than as a property of the host CPU (or perhaps the FPGA), but...there's tons of existing precedent for stuff like that in the control-plane-agent API, like Gimlet's CPU GPIO toggle count, the CPU temperature sensor being a totally separate device from the CPU itself, and so on.

i do think that given the way the `component-details` and `inventory` APIs work, this _is_ the right thing to do (otherwise you would have a situation where if you wanted to read T_ctl_ from the CPU temperature sensor, you would also have to read 16384B of POST codes, which you couldn't opt out of!). but, it _feels_ conceptually uncomfortable the way we have a "flat" inventory where these are exposed to upstack software as totally separate things and that we sometimes talk about them in UIs as "separate devices". ideally they would all be grouped together as children of the same refdes. basically what i am saying is "man, i wish the `control-plane-agent` inventory API was shaped more like the IPCC inventory API and could better model topology _trees_, but...**none of this is your problem and definitely is not getting fixed in _this_ PR**!

idl/cpu-seq.idol
mkeeter: Eh, they are stored in different locations in the FPGA, so I figure it's reasonable to expose separate interfaces.

idl/cpu-seq.idol
mkeeter: Agreed!

task/control-plane-agent/src/inventory.rs
mkeeter: The TODO is also copy-pasted from the previous block, so I'm just going to delete it.

I agree that MGS `component-details` and `inventory` are... full of character (we serialized a fixed `enum` to untyped TLV blobs, then deserialize them on the other side based on the TLV codes??)

hawkw: Yeah. I definitely was not trying to say "we need to figure all this out right now". :)

Issue #2422, opened 2026-03-04
an xtask for building RustDoc for Hubris crates would be nice


hawkw:
One cannot easily build the RustDoc for a number of crates in the repo. For one thing, much of our code cannot be built without an explicitly set Cortex-M target, so one must pass `--target <pick your favorite supported cortex-m toolchain>` in order to document any crate which depends on `userlib`. However, that alone is not sufficient if a crate either has a build script which uses the Hubris `app.toml` config, or depends on a crate which does. For example:

```console
eliza@theseus ~/Code/oxide/hubris $ cargo doc --open -p ereports --target thumbv7em-none-eabihf --no-deps
   Compiling task-sensor-api v0.1.0 (/home/eliza/Code/oxide/hubris/task/sensor-api)
error: failed to run custom build command for `task-sensor-api v0.1.0 (/home/eliza/Code/oxide/hubris/task/sensor-api)`

Caused by:
  process didn't exit successfully: `/home/eliza/Code/oxide/hubris/target/debug/build/task-sensor-api-06a98560f4251f57/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=../../idl/sensor.idol
  cargo::rerun-if-env-changed=HUBRIS_APP_CONFIG

  --- stderr

  thread 'main' panicked at build/i2c/src/lib.rs:523:17:
  malformed config.i2c: app.toml missing global config section [config]
  stack backtrace:
     0: __rustc::rust_begin_unwind
               at /rustc/0d9592026226f5a667a0da60c13b955e0b486a07/library/std/src/panicking.rs:697:5
     1: core::panicking::panic_fmt
               at /rustc/0d9592026226f5a667a0da60c13b955e0b486a07/library/core/src/panicking.rs:75:14
     2: build_i2c::ConfigGenerator::new
               at /home/eliza/Code/oxide/hubris/build/i2c/src/lib.rs:523:17
     3: build_i2c::codegen
               at /home/eliza/Code/oxide/hubris/build/i2c/src/lib.rs:1720:17
     4: build_script_build::main
               at ./build.rs:37:5
     5: core::ops::function::FnOnce::call_once
               at /home/eliza/.rustup/toolchains/nightly-2025-07-20-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:253:5
  note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

[...]

PR #2428, opened 2026-03-06
Move thermal buffers off the stack

mkeeter:
I tried to make a change to `thermal` and the stack complained.  I then realized that `main`'s stack in `thermal` is huge:
```
thermal: 7000 bytes (limit is 8096)
     [+0] _start
  [+6008] main
   [+896] task_thermal::control::ThermalControl::transition_to_running
    [+56] task_thermal::control::ThermalControl::record_leaving_overheat
    [+40] <ringbuf::CountedRingbuf<T,C,_> as ringbuf::RecordEntry<T>>::record_entry
```

This PR shrinks it down to a more reasonable size by moving temperature data arrays into a static buffer:
```
thermal: 2704 bytes (limit is 8096)
     [+0] _start
  [+2320] main
   [+112] task_thermal::control::ThermalControl::set_pwm
    [+16] task_thermal::bsp::Bsp::fan_control
    [+64] task_thermal::control::Max31790State::initialize
    [+16] drv_i2c_devices::max31790::write_reg8
    [+88] drv_i2c_api::I2cDevice::write
    [+24] core::panicking::panic_fmt
    [+56] __rustc[b0c4d0eb67640fa4]
     [+8] core::fmt::Write::write_fmt
```

There's a little bit of trickery here: we store a single `[Option<TemperatureReading>; TEMPERATURE_ARRAY_SIZE]` array, then wrap it in a special type when it's guaranteed that all the values are `Some(..)`.

We also have to (unfortunately) add a `ThermalControlState::Empty`, because we need a valid temporary value to put into the `state` position while modifying + replacing it.

### DIFF
--- a/task/thermal/src/control.rs
+++ b/task/thermal/src/control.rs
@@ -27,7 +27,7 @@ use task_thermal_api::{SensorReadError, ThermalAutoState, ThermalProperties};
 use userlib::{
     sys_get_timer,
     units::{Celsius, PWMDuty, Rpm},
-    TaskId,
+    TaskId, UnwrapLite,
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -495,7 +495,7 @@ pub(crate) struct ThermalControl<'a> {
 /// Represents the state of a temperature sensor, which either has a valid
 /// reading or is marked as inactive (due to power state or being missing)
 #[derive(Copy, Clone, Debug)]
-pub enum TemperatureReading {
+enum TemperatureReading {
     /// Normal reading, timestamped using monotonic system time
     Valid(TimestampedTemperatureReading),
 
@@ -505,7 +505,7 @@ pub enum TemperatureReading {
 
 /// Represents a temperature reading at the time at which it was taken
 #[derive(Copy, Clone, Debug)]
-pub struct TimestampedTemperatureReading {
+struct TimestampedTemperatureReading {
     time_ms: u64,
     value: Celsius,
 }
@@ -796,24 +796,48 @@ enum ThermalControlState {
 }
 
 impl ThermalControlState {
-    fn data(&self) -> &'static RawTemperatureArray {
+    /// Sets all temperature readings to `None` and returns the array
+    fn reset_values(&mut self) -> OptionalTemperatureArray {
         match self {
             ThermalControlState::Boot { values }
-            | ThermalControlState::Uncontrollable(values) => values.data(),
+            | ThermalControlState::Uncontrollable(values) => {
+                values.reset_values()
+            }
             ThermalControlState::Running { values, .. }
             | ThermalControlState::Critical { values, .. }
-            | ThermalControlState::FanParty { values } => values.data(),
+            | ThermalControlState::FanParty { values } => values.reset_values(),
         }
     }
 }
 
+/// Abstractions over temperature array data
+///
+/// We have three main types:
+/// - A `RawTemperatureArray` is allocated in static memory using
+///   `mutable_statics!` and provides the underlying storage for other array
+///   types.  It stores optional temperature readings in `Cell`s, so that it can
+///   be passed around by shared reference.
+/// - A [`OptionalTemperatureArray`] is a thin wrapper around a
+///   `&'static RawTemperatureArray`, providing getters / setters / iterators.
+/// - A [`TemperatureArray`] is a thin wrapper around a
+///   `&'static RawTemperatureArray` which guarantees that all of the
+///   temperatures in the array are `Some(..)`.
+///
+/// This module exists to preserve the `TemperatureArray` invariants (or at
+/// least make it harder to mess with them; someone could get at the inner
+/// `&'static RawTemperatureArray and poke values directly, but let's just not
+/// do that).
 mod temperature_array {
     use super::{
         Bsp, Cell, DynamicChannelsArray, SensorId, TemperatureReading,
-        ThermalProperties, TEMPERATURE_ARRAY_SIZE,
+        ThermalProperties, UnwrapLite, TEMPERATURE_ARRAY_SIZE,
     };
 
     /// Array of optional temperature readings
+    ///
+    /// The array contains `Cell` objects so that it can be passed by shared
+    /// reference; otherwise, it becomes hard to transition between states in
+    /// the state machine because we can't move out a `&mut RawTemperatureArray`
     pub type RawTemperatureArray =
         [Cell<Option<TemperatureReading>>; TEMPERATURE_ARRAY_SIZE];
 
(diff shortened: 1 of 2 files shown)


task/thermal/src/control.rs
```diff
@@ -503,7 +503,7 @@ enum TemperatureReading {
 
 /// Represents a temperature reading at the time at which it was taken
 #[derive(Copy, Clone, Debug)]
-struct TimestampedTemperatureReading {
+pub struct TimestampedTemperatureReading {
```
hawkw: why did these become `pub`?

task/thermal/src/control.rs
```diff
@@ -785,14 +783,233 @@ enum ThermalControlState {
...
+        let prev = core::mem::replace(self, ThermalControlState::Empty);
+        let values = match prev {
+            ThermalControlState::Boot { values }
+            | ThermalControlState::Uncontrollable(values) => values.take(),
+            ThermalControlState::Running { values, .. }
+            | ThermalControlState::Critical { values, .. }
+            | ThermalControlState::FanParty { values, .. } => values.take(),
+            ThermalControlState::Empty => panic!(),
+        };
+        TemperatureArray::new(values).unwrap()
```
hawkw: turbo nit: should this perhaps be `unwrap_lite()` to save a couple bytes of text?

task/thermal/src/control.rs
```diff
@@ -785,14 +783,233 @@ enum ThermalControlState {
...
+
+    /// A `TemperatureArrayToken` attests to a valid temperature array
+    ///
+    /// This may either be a `TemperatureArray` object or a
+    /// `OptionalTemperatureArray` object where all values are `Some(..)`.
+    ///
+    /// It is used as a hint that APIs require a valid temperature array, to
+    /// make them harder to misuse (but not impossible; one could stash a token
+    /// away then reuse it when the array is no longer valid).
+    pub struct TemperatureArrayToken {
```
hawkw: Super unimportant nitpick, feel free to ignore me: I might consider naming this something like `NonEmptyToken` or something so that the name indicates what it symbolizes?

task/thermal/src/control.rs
```diff
@@ -785,14 +783,233 @@ enum ThermalControlState {
...
+    }
+
+    /// A `TemperatureArrayToken` attests to a valid temperature array
+    ///
+    /// This may either be a `TemperatureArray` object or a
+    /// `OptionalTemperatureArray` object where all values are `Some(..)`.
+    ///
+    /// It is used as a hint that APIs require a valid temperature array, to
+    /// make them harder to misuse (but not impossible; one could stash a token
+    /// away then reuse it when the array is no longer valid).
```
hawkw: it might be possible to prevent this if the token borrowed the array, but I presume this not gonna work as we want to be able to mutate the array?

task/thermal/src/control.rs
```diff
@@ -1247,7 +1442,8 @@ impl<'a> ThermalControl<'a> {
                 if let Some(due_to) = any_power_down {
                     self.transition_to_uncontrollable_due_to(due_to, now_ms)
                 } else if let Some(due_to) = any_critical {
-                    let values = *values;
+                    let token = values.token();
```
hawkw: hmm, really, maybe it's the _method_ that i'd like to be named something more descriptive than `token()`...not a big deal though.

task/thermal/src/control.rs
mkeeter: Something else started exposing them, so I got the  `type X is more private than the item Y` warning.  Moot point after reorganizing things, so I've reverted this change.

mkeeter:
@hawkw I switched to an array of `[Cell<Option<TemperatureReading>>; ..]`, which removes a bunch of the complexity - no need for `ThermalControlState::Empty` or the `TemperatureArrayToken`.

mkeeter:
I did a quick `stackmargin` audit (since I had edited margins by vibes) and they still seem reasonable:
```
19 thermal            0x2404c000       8096       6952       1144 # old sidecar
 7 thermal            0x24002000       6000       4968       1032 # old gimlet
 8 thermal            0x24002000       6000       2720       3280 # old cosmo
19 thermal            0x2404c000       4000       2688       1312 # new sidecar
 7 thermal            0x24002000       2000       1296        704 # new gimlet
 8 thermal            0x24002000       3000       1288       1712 # new cosmo
```

task/thermal/src/control.rs
```diff
@@ -785,15 +785,160 @@ enum ThermalControlState {
...
+            ThermalControlState::Boot { values }
+            | ThermalControlState::Uncontrollable(values) => values.data(),
+            ThermalControlState::Running { values, .. }
+            | ThermalControlState::Critical { values, .. }
+            | ThermalControlState::FanParty { values } => values.data(),
+        }
+    }
 }
 
+mod temperature_array {
```
hawkw, line 830: Nit, take it or leave it: I wonder somewhat if a higher level theory of operation for the relationship between the `RawTemperatureArray`, `OptionalTemperatureArray`, and `TemperatureArray` types could be included in the module-level docs for this?

task/thermal/src/control.rs
```diff
@@ -785,15 +785,160 @@ enum ThermalControlState {
...
+        /// # Panics
+        /// If `i` is out of range for the array
+        pub fn set(&self, i: usize, value: TemperatureReading) {
+            self.0[i].set(Some(value))
+        }
+    }
+
+    /// Temperature state iterator with `TemperatureReading` values
+    impl TemperatureArray {
+        pub fn zip_temperatures<'a>(
```
hawkw: I think the doc comment is meant to be on the method
```suggestion
    impl TemperatureArray {
    
        /// Temperature state iterator with `TemperatureReading` values
        pub fn zip_temperatures<'a>(
```

hawkw: also, why are there two separate `impl` blocks here?

task/thermal/src/control.rs
```diff
@@ -785,15 +785,160 @@ enum ThermalControlState {
...
+    ///
+    /// The `values` array contains static and dynamic values (in order);
+    /// this function will panic if sizes are mismatched.
+    ///
+    /// Every dynamic input is represented by an `Option<DynamicInputChannel>`.
+    /// If the input is not present right now, it will be `None`, but will
+    /// continue to take up space to preserve ordering.
+    ///
+    /// In cases where dynamic inputs are not present (i.e. they are `None` in
+    /// the array), the iterator will skip that entire tuple.
```
hawkw, line 937: i feel like some of this documentation could also be included on the public `OptionalTemperatureArray::zip_temperatures`, but 🤷‍♀️ --- we never build API docs for this stuff (see #2422), so putting it on private functions is probably fine.

task/thermal/src/control.rs
mkeeter: Yup, done!

task/thermal/src/control.rs
mkeeter: Stuff got shuffled around, both things are now fixed

task/thermal/src/control.rs
mkeeter: I didn't want to repeat this on both `OptionalTemperature::zip_temperatures` and `TemperatureArray::zip_temperatures`, so I figured this was the right place (since as you said we don't actually build docs) 🤷🏻‍♂️

PR #2430, opened 2026-03-09
Add PWM sensor to thermal task

mkeeter:
(staged on top of #2428)

This adds a new `fan_ctrl` sensor, which is updated by the thermal loop.  In combination with https://github.com/oxidecomputer/humility/pull/602, we can log both fan speed (which was already possible) and commanded PWM values with `humility dashboard -o out.csv`

Here's an example while running `stress-ng`:

Screenshot 2026-03-09 at 10 54 42 AM

The hysteresis from the fan controller is quite noticeable if we plot PWM vs RPM:

Screenshot 2026-03-09 at 10 55 01 AM

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -1352,6 +1352,12 @@ device = "dimm"
 description = "DIMM L, sensor 1"
 sensors.temperature = 1
 
+[[config.sensor.devices]]
+name = "fan_ctrl"
+device = "thermal_loop"
+description = "Thermal loop"
+sensors.pwm = 1
+
 ################################################################################
 [config.spi.spi2]
 controller = 2
--- a/app/sidecar/base.toml
+++ b/app/sidecar/base.toml
@@ -1064,6 +1064,12 @@ device = "qsfp"
 description = "QSFP transceiver 31"
 sensors.temperature = 1
 
+[[config.sensor.devices]]
+name = "fan_ctrl"
+device = "thermal_loop"
+description = "Thermal loop"
+sensors.pwm = 1
+
 [config.spi.spi1]
 controller = 1
 
--- a/task/thermal/src/control.rs
+++ b/task/thermal/src/control.rs
@@ -1495,7 +1495,7 @@ impl<'a> ThermalControl<'a> {
             ControlResult::Pwm(target_pwm) => {
                 // Send the new RPM to all of our fans
                 ringbuf_entry!(Trace::ControlPwm(target_pwm.0));
-                self.set_pwm(target_pwm)?;
+                self.set_pwm(Ok(target_pwm), now_ms)?;
             }
             ControlResult::PowerDown => {
                 ringbuf_entry!(Trace::PowerDownAt(sys_get_timer().now));
@@ -1504,7 +1504,7 @@ impl<'a> ThermalControl<'a> {
                 if let Err(e) = self.bsp.power_down() {
                     ringbuf_entry!(Trace::PowerDownFailed(e));
                 }
-                self.set_pwm(PWMDuty(0))?;
+                self.set_pwm(Err(task_sensor_api::NoData::DeviceOff), now_ms)?;
             }
         }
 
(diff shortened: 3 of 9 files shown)


app/cosmo/base.toml
```diff
@@ -1354,6 +1354,12 @@ device = "dimm"
 description = "DIMM L, sensor 1"
 sensors.temperature = 1
 
+[[config.sensor.devices]]
+name = "fan_ctrl"
+device = "thermal_loop"
+description = "Thermal loop PWM command"
+sensors.pwm = 1
```
hawkw, line 1359: Conceptually (or...aesthetically?), i kind of would prefer this to be an additional sensor on [the MAX31790 device](https://github.com/oxidecomputer/hubris/blob/7e0e7afbe1bb11abcf58fbdecb333df4273834d0/app/cosmo/base.toml#L1139-L1144)...but I don't think the sensors codegen will play nice with an I2C device that has both I2C sensors _and_ a sensor that is posted to the sensor API from a Hubris task. I dunno, maybe it's better modeled as a property of the Hubris thermal loop rather than of the PWM controller, but thinking about things from the perspective of the inventory representing a topology tree (which we mostly don't do), it feels like it _should_ be a child of the MAX31790 node in that tree to most accurately represent the system. But, whatever.

task/thermal/src/control.rs
```diff
@@ -1554,11 +1554,19 @@ impl<'a> ThermalControl<'a> {
...
         match control_result {
             ControlResult::Pwm(target_pwm) => {
                 // Send the new RPM to all of our fans
                 ringbuf_entry!(Trace::ControlPwm(target_pwm.0));
                 self.set_pwm(target_pwm)?;
+                self.sensor_api.post(
+                    OUTPUT_PWM_SENSOR,
+                    target_pwm.0 as f32,
+                    now_ms,
+                );
```
hawkw: I kinda wonder if maybe we ought to just make `self.set_pwm` also post the new duty cycle to the sensors task, so that it's impossible to set the duty cycle without also posting value to the sensors task?

task/thermal/src/control.rs
```diff
@@ -1554,11 +1554,19 @@ impl<'a> ThermalControl<'a> {
...
+        pub const OUTPUT_PWM_SENSOR: SensorId =
+            task_sensor_api::config::other_sensors::THERMAL_LOOP_FAN_CTRL_PWM_SENSOR;
         match control_result {
             ControlResult::Pwm(target_pwm) => {
                 // Send the new RPM to all of our fans
                 ringbuf_entry!(Trace::ControlPwm(target_pwm.0));
                 self.set_pwm(target_pwm)?;
+                self.sensor_api.post(
+                    OUTPUT_PWM_SENSOR,
+                    target_pwm.0 as f32,
```
hawkw: I kinda went back and forth on reporting this as the raw control value versus reporting it as a percentage (so that rather than 0x64 we would say 100%). Maybe the raw value is better and upstack software can figure out what percentage that represents? I dunno.

app/cosmo/base.toml
mkeeter: I thought about this a little, but the MAX31790 actually has **6 different PWM values** (one for each fan) – the fact that they're always the same value is a property of the Hubris thermal loop!

(and yes, `sensors` codegen would probably be mad as well)

task/thermal/src/control.rs
mkeeter: I think for the `sensors` data, we want the raw value.  However, when we expose this to MGS, I'm expecting to convert it into a `NormalizedPwm` (0.0-1.0) using the variant from https://github.com/oxidecomputer/management-gateway-service/pull/481

hawkw: oh okay sounds good!

app/cosmo/base.toml
hawkw: that's reasonable, carry on.

hawkw: (i still wish things were different, but that's not your fault)

task/thermal/src/control.rs
mkeeter: Done – it's a little awkward (because we need to represent the `NoData::DeviceOff` case), so I switched the input argument to a `Result<PWMDuty, NoData>`.

hawkw: Hmm, did we _have_ to do it like that? I was imagining that in the `DeviceOff` state we would just call `self.sensor_api.post` directly. Rather than _making `set_pwm` the only way we ever post a sensor reading_, I was just thinking that _any time we set the PWM duty cycle we also post a sensor reading_ (but this would not be the only place we posted a sensor reading). `set_pwm(Err(NoData::DeviceOff))` feels a bit awkward to me.

mkeeter: I think the `Result<..>` shenanigans are the simplest way to do it.  In our previous world where `set_pwm` just takes a `PwmDuty`, when we go uncontrollable, we'd still call `set_pwm(PwmDuty(0))` – and would therefore post 0 to the `sensors` task because we don't know any better.

The `Result<..>` argument lets us distinguish between "there's an error, so post the error but set the PWM to 0" versus "situation normal, post a pwm of 0 and set it".

hawkw: _Oh_ I see, yeah, okay. I don't hate it.

task/thermal/src/control.rs
```diff
@@ -1487,7 +1487,7 @@ impl<'a> ThermalControl<'a> {
                 if let Err(e) = self.bsp.power_down() {
                     ringbuf_entry!(Trace::PowerDownFailed(e));
                 }
-                self.set_pwm(PWMDuty(0))?;
+                self.set_pwm(Err(task_sensor_api::NoData::DeviceOff), now_ms)?;
```
labbott, line 1507: It feels strange to treat `DeviceOff` as an error. Under what circumstances would we hit PowerDown?

app/sidecar/base.toml
```diff
@@ -1064,6 +1064,12 @@ device = "qsfp"
 description = "QSFP transceiver 31"
 sensors.temperature = 1
 
+[[config.sensor.devices]]
+name = "fan_ctrl"
+device = "thermal_loop"
+description = "Thermal loop PWM command"
```
labbott: Nit: the command part of this description feels weird. Just drop it?

task/thermal/src/control.rs
mkeeter: This is because we want `set_pwm` to be solely responsible for logging values to the `sensors` task, but need a way to distinguish between "the PWM value is 0 because everything is fine" versus "there's a big problem, set the PWM value to 0"

(see discussions with Eliza above)

app/sidecar/base.toml
mkeeter: Thinking about it, I'm going to edit the device description to just "Thermal loop", since PWM is also specified as the sensor type.

PR #2441, opened 2026-03-19
Notion of a per-task working directory

mkeeter:
I did an experiment where Hubris builds were isolated by image + task, which made incremental rebuilds much faster (45 → 11 seconds) but take up much more space (1 GiB → 23 GiB).  This PR **doesn't** implement that change, but it **does** lay the groundwork for that (or similar) changes by abstracting away a per-task working directory.  Now, instead of being hard-coded to `target`, the build system has to ask the `PackageConfig` for a working directory (passing in the task name as an argument).  Right now, we always return `target`, so there should be no functional change.

In many cases, passing around the whole `&PackageConfig` as an argument makes other arguments unnecessary (because they were being extracted out of the `PackageConfig`.

### DIFF
--- a/build/xtask/src/dist.rs
+++ b/build/xtask/src/dist.rs
@@ -77,6 +77,9 @@ pub struct PackageConfig {
     /// `target/$NAME/dist`.
     dist_dir: PathBuf,
 
+    /// Directory where the build takes place
+    work_dir: PathBuf,
+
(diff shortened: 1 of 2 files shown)


labbott (approved):
I _think_ this looks okay but we may want another set of eyes on it in case I missed something subtle

PR #2444, opened 2026-03-19
Make task config only readable from task crate build script

mkeeter:
Right now, dispatching `cargo` once per task is a reason why incremental rebuilds are slow.  I would like to live in a world where we dispatch `cargo` once to build multiple (compatible†) tasks.

One issue with this plan: build scripts for library crates are free to read `HUBRIS_TASK_CONFIG` and react accordingly.  In a world where multiple tasks are built in a single `cargo` invocation, there's no single `HUBRIS_TASK_CONFIG` to read.

The PR eliminates that behavior: build scripts that read `HUBRIS_TASK_CONFIG` must have a `CARGO_PKG_NAME` that matches the task crate name.  In other words, only **a task's** build script may read the task config.  Otherwise, the `build_util::*task*config` function will panic.

This is useful because it also works in the opposite direction: when this property is true, we can use `CARGO_PKG_NAME` to unambiguously‡ pick which task config to use.

It turns out that we *almost* had this property already!  The one exception is the `drv-stm32h7-spi-server-core` crate, which had a build script that read the parent task configuration to pick which SPI peripheral to codegen.  In this PR, I make that codegen the parent crate's responsibility; `drv_stm32h7_spi_server_core::SpiServerCore::init` now takes a configuration from the caller, instead of baking it into a global variable.

The build script for `drv-stm32h7-spi-server-core` is moved to a separate `build-stm32h7-spi-server-core`, so that all of the parent crates can invoke it themselves.

† "compatible" means that they enable the exact same set of features in their dependencies, to avoid feature unification surprises
‡ We will still be unable to simultaneously build multiple tasks that use the same crate, e.g. `spi2` and `spi3`, but that's fine

### DIFF
--- a/build/util/src/lib.rs
+++ b/build/util/src/lib.rs
@@ -45,7 +45,7 @@ pub fn target_os() -> String {
 
 /// Reads the `HUBRIS_TASK_NAME` env var.
 pub fn task_name() -> String {
-    crate::env_var("HUBRIS_TASK_NAME").expect("missing HUBRIS_TASK_NAME")
+    env_var("HUBRIS_TASK_NAME").expect("missing HUBRIS_TASK_NAME")
 }
 
 /// Checks to see whether the given feature is enabled
@@ -84,7 +84,7 @@ pub fn expose_m_profile() -> Result<()> {
 
 /// Returns the `HUBRIS_BOARD` envvar, if set.
 pub fn target_board() -> Option<String> {
-    crate::env_var("HUBRIS_BOARD").ok()
+    env_var("HUBRIS_BOARD").ok()
 }
 
 /// Exposes the board type from the `HUBRIS_BOARD` envvar into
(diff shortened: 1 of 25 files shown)


build/util/src/lib.rs
```diff
@@ -167,20 +167,32 @@ pub fn task_config<T: DeserializeOwned>() -> Result<T> {
...
+) -> Result<Option<toml_task::Task<T>>> {
+    let out = toml_from_env::<toml_task::Task<T>>("HUBRIS_TASK_CONFIG")?;
+    if let Some(out) = &out {
+        let crate_name = env_var("CARGO_PKG_NAME").unwrap();
+        assert_eq!(
+            out.name, crate_name,
+            "can't read task config from non-task crate {crate_name}"
+        );
+    }
+    Ok(out)
```
labbott, line 195: having both `try_task_full_config` and `task_maybe_config` seems redundant, can we choose one or the other?

PR #2449, opened 2026-03-20
grapefruit: break out rev-a support, add rev-b

Aaron-Hartwig:


app/grapefruit/rev-b.toml
```diff
@@ -0,0 +1,82 @@
+name = "grapefruit-standalone"
+board = "grapefruit-b"
+inherit = "base.toml"
+
+[tasks.jefe]
```
mkeeter: Nit: these changes are duplicated between `rev-a` and `rev-b`.  They could instead be extracted into a `standalone.toml`, to be inherited by both `rev-a.toml` and `rev-b.toml` (but not `rev-a-ruby.toml`)

app/grapefruit/base.toml
```diff
@@ -395,17 +394,18 @@ rx = { packets = 3, bytes = 4096 }
...
 
 [config]
+
+################################################################################
+
 [[config.i2c.controllers]]
 controller = 1
 
 [config.i2c.controllers.ports.B]
-name = "fpga"
```
mkeeter, line 402: Just move this whole controller into the `rev-b.toml`, so no one can try talking to the FRUID on rev A?

app/grapefruit/rev-a.toml
```diff
@@ -0,0 +1,5 @@
+name = "grapefruit-a"
+board = "grapefruit-a"
+inherit = "standalone.toml"
+
```
mkeeter: Nit: delete these two lines?

mkeeter (approved):
Two small suggestions, but looks good otherwise

PR #2454, opened 2026-03-25
host_sp_comms: add low-level "debug" interface

dancrossnyc:
When debugging recent IPCC problems, it would have been very useful to have a low-level "debug" interface for sending traffic between host and SP so that we could probe the issues in question on a bench machine.

I hacked something for this by overloading an IPCC transaction that wasn't used on the machine I wanted to test on, which was enormously impactful, as it exposed a quartz problem in the FPGA that Nathanael was able to fix quickly.  But it was obviously a hack, and moreoever, since it was tied to IPCC (and IPCC was considered suspect at the time), using it felt like a bit of a gamble.

Having a dedicated debugging interface that is below the level of IPCC, and not complected with COBS encoding, Hubpack, and so on, would be generally useful.  This is the Hubris side of that: I've implemented a parallel protocol for supporting three simple debug messages:

1. A discard message: this simply throws away what is sent to it, with the side effect of emitting a message into a ring buffer.
2. An echo message: this echos the data sent to it back to the distant end, and produces a ring buffer entry.
3. A "chargen" message: this causes $n$ bytes from a repeating pattern of printable ASCII characters to be back to the host, limited by the number requested and size of the output buffer.

Since the IPCC header and magic number are fixed, debug messages are easily distinguishable from normal IPCC messages by looking for a prefix string in a received "frame" from the host: the two can never collide.

### DIFF
--- a/task/host-sp-comms/src/main.rs
+++ b/task/host-sp-comms/src/main.rs
@@ -107,12 +107,25 @@ const MAX_HOST_FAIL_MESSAGE_LEN: usize = 4096;
 // of that for us.
 const NUM_HOST_MAC_ADDRESSES: u16 = 3;
 
+// The same IO path can be used for both IPCC, and a lower-level debugging
+// interface, for testing the IO path itself.  We differentiate between the two
+// by detecting specific header types, and parsing the corresponding messages
+// into a typed enum: debug message handling is then short-circuited.
+enum DebugCmd<'a> {
+    Discard,
+    Echo(&'a [u8]),
+    CharGen(u16),
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, counters::Count)]
 enum Trace {
     #[count(skip)]
     None,
     UartRxOverrun,
     ParseError(#[count(children)] DecodeFailureReason),
+    DebugDiscard,
+    DebugEcho(u64),
+    DebugCharGen(u16),
     SetState {
         now: u64,
         #[count(children)]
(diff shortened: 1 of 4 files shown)


dancrossnyc:
I still have to test this properly. On the host side, we should probably just write a ktest for it.

@citrus-it, for your general information.

dancrossnyc:
It turns out, our existing espi ktest is sufficient to test ths.

task/host-sp-comms/src/main.rs
```diff
@@ -786,6 +799,57 @@ impl ServerImpl {
...
+                }
+                match cmd {
+                    DebugCmd::Discard => ringbuf_entry!(Trace::DebugDiscard),
+                    DebugCmd::Echo(data) => {
+                        ringbuf_entry!(Trace::DebugEcho(data.len() as u64));
+                        let _ = self.tx_buf.try_copy_raw_data(data);
+                    }
+                    DebugCmd::CharGen(count) => {
+                        ringbuf_entry!(Trace::DebugCharGen(count));
+                        // const CHARGEN: &'static [u8] =
```
mkeeter: Remove dead comment?

task/host-sp-comms/src/main.rs
```diff
@@ -800,10 +864,11 @@ impl ServerImpl {
     //
     // This method always (i.e., on success or failure) clears `rx_buf` before
     // returning to prepare for the next packet.
-    fn process_message(
+    fn process_ipcc_message(
         &mut self,
         reset_tx_buf: bool,
     ) -> Result<(), DecodeFailureReason> {
+        // If the message was Not a debug message, then parse it normally.
```
mkeeter: This comment looks out of place, delete (or move to `process_message` above)?
```suggestion
```

task/host-sp-comms/src/main.rs
```diff
@@ -1690,6 +1756,48 @@ fn parse_received_message(
...
+        return false;
+    }
+    msg[0..5] == *b"DEBUG"
+}
+
+// This is conceptually a method on `ServerImpl`, but it takes references to
+// several of its fields instead of `self` to avoid borrow checker issues.
+fn parse_debug_message(
+    msg: &[u8],
+) -> Result<DebugCmd<'_>, DecodeFailureReason> {
```
mkeeter, line 1788: Vibes: it could be cleaner to combine this function and `is_debug_message`, i.e.

```rust
fn parse_debug_message(
    msg: &[u8],
) -> Result<DebugCmd<'_>, DecodeFailureReason> {
    if msg.len() < 7 || msg[0..5] != *b"DEBUG" {
        return Err(DecodeFailureReason::Deserialize);
    }
    match &msg[5..7] {
        b"07" => Ok(DebugCmd::Echo(&msg[7..])),
        b"09" => Ok(DebugCmd::Discard),
        b"13" if msg.len() == 11 => {
            let nstr = str::from_utf8(&msg[7..11])
                .map_err(|_| DecodeFailureReason::Deserialize)?;
            let n = u16::from_str_radix(nstr, 16)
                .map_err(|_| DecodeFailureReason::Deserialize)?;
            Ok(DebugCmd::CharGen(n))
        }
        _ => Err(DecodeFailureReason::Deserialize),
    }
}
```

Then, `process_message` could be written as
```rust
    // Process a request message from the host.
    fn process_message(
        &mut self,
        reset_tx_buf: bool,
    ) -> Result<(), DecodeFailureReason> {
        // Debug messages have a distinct header that separates them from normal
        // IPCC messages.
        if let Some(m) = parse_debug_message(self.rx_buf) {
            self.process_debug_message(m, reset_tx_buf)
        } else {
            self.process_ipcc_message(reset_tx_buf)
        }
    }
```
...and `process_debug_message` takes a `DebugCmd` directly, instead of re-checking.

Now that I've written this all out, I suspect it runs afoul of borrow-checker issues (since `self.rx_buf` has to be borrowed by `parse_debug_message` and the `DebugCmd`); if you can figure out a way to make it work, it would be nice!

task/host-sp-comms/src/tx_buf.rs
```diff
@@ -292,6 +292,40 @@ impl TxBuf {
...
+        let mut idx = 0;
+        self.pkt[idx] = 0;
+        idx += 1;
+        for b in it.take(max) {
+            if b == 0 {
+                return Err(());
+            }
+            self.pkt[idx] = b;
+            idx += 1;
+        }
```
mkeeter, line 324: You could do this with a `zip` instead of an explicit `idx`, but I don't think it's worth the trouble.

task/host-sp-comms/src/main.rs
dancrossnyc: Done.

task/host-sp-comms/src/main.rs
dancrossnyc: Deleted. Less is more in this case, I think.

task/host-sp-comms/src/main.rs
dancrossnyc: Sadly, I had tried something very similar when I first wrote this, but yeah, the borrow checker was unhappy. :-(

task/host-sp-comms/src/tx_buf.rs
dancrossnyc: I thought about trying to do something like that, but it gets a little gnarly when you chain the trailing zero byte (I suppose one could chain the leading zero, as well)?

mkeeter (approved):

PR #2466, opened 2026-04-08
Make `hiffy` callable over the network

mkeeter:
For too long, a network-attached Humility has only been able to run a subset of `hiffy` programs (ones that exclusively use the `Send` operations, using the `udprpc` hack).  This PR adds a `net` feature to the `hiffy` task, restoring it to full usefulness when network-attached.

`hiffy` accepts three messages:

- Write program text
- Write data
- Kick (begin execution)

For simplicity, these messages **only** write to RAM.  Even the `HiffyKick` message doesn't actually start execution itself; it just sets the `HIFFY_KICK` variable, which is checked the main loop.  This mimics the behavior of the debugger.

Interestingly, we don't need any new infrastructure to **read** the results; we can use the existing RAM-reading infrastructure.

Unfortunately, Github is butchering the diff here.  The existing code (look for kick, run program) is unchanged but for indentation, because we have **two notifications** (timer and `net` activity).  If the new `net` notification fires, then we handle the packet in a new `net::State` object.

I'm opening this as a draft while I polish up the Humility side, but it's ready for review now.

### DIFF
--- a/task/hiffy/src/main.rs
+++ b/task/hiffy/src/main.rs
@@ -196,14 +196,12 @@ fn main() -> ! {
     let mut net_state = net::State::new();
 
     loop {
-        HIFFY_READY.store(1, Ordering::Relaxed);
-        hl::sleep_for(sleep_ms);
-        HIFFY_READY.store(0, Ordering::Relaxed);
-
         // Sleep until either the timer expires or we receive a notification
         // from the `net` task indicating that it's ready for us.
         let deadline = sys_get_timer().now.saturating_add(sleep_ms);
+        HIFFY_READY.store(1, Ordering::Relaxed);
         sys_set_timer(Some(deadline), notifications::TIMER_MASK);
+        HIFFY_READY.store(0, Ordering::Relaxed);
 
         #[cfg(feature = "net")]
         let bits = notifications::SOCKET_MASK | notifications::TIMER_MASK;
@@ -329,7 +327,9 @@ unsafe fn bind_lifetime_mut<'a, const N: usize>(
 
 #[cfg(feature = "net")]
 mod net {
-    use super::{HIFFY_DATA, HIFFY_KICK, HIFFY_TEXT, notifications};
+    use super::{
+        HIFFY_DATA, HIFFY_KICK, HIFFY_TEXT, bind_lifetime_mut, notifications,
+    };
     use core::sync::atomic::Ordering;
     use static_cell::ClaimOnceCell;
     use task_net_api::{
@@ -482,10 +482,18 @@ mod net {
             // Perform the actual operation
             match RpcOp::from_u16(header.operation.get()) {
                 Some(RpcOp::WriteHiffyText) => {
+                    // Dummy object to bind references to a non-static lifetime
+                    let lifetime = ();
                     let offset = header.arg.get() as usize;
-                    // TODO: using a `static mut` global in ways that are not
-                    // obviously sound (but no worse than elsewhere in the file)
-                    let text = unsafe { &mut HIFFY_TEXT };
+
+                    // SAFETY: we are constructing a slice with a bounded
+                    // lifetime, and are in single-threaded code.  We don't
+                    // expect a debugger to be editing our memory.  If someone
+                    // is simultaneously editing `HIFFY_TEXT` with a debugger
+                    // *and* over the network, they deserve whatever happens.
+                    let text = unsafe {
+                        bind_lifetime_mut(&lifetime, &raw mut HIFFY_TEXT)
+                    };
                     if let Some(chunk) = offset
                         .checked_add(rest.len())
                         .and_then(|e| text.get_mut(offset..e))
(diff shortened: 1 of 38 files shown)


mkeeter:
Marking this a ready-for-review now that https://github.com/oxidecomputer/humility/pull/616 is out

task/hiffy/src/main.rs
```diff
@@ -161,66 +161,280 @@ fn main() -> ! {
...
+            if HIFFY_KICK.load(Ordering::SeqCst) == 0 {
+                sleeps += 1;
 
-            // Exponentially backoff our sleep value, but no more than 250ms
-            if sleeps == 10 {
-                sleep_ms = core::cmp::min(sleep_ms * 10, 250);
-                sleeps = 0;
+                // Exponentially backoff our sleep value, but no more than 250ms
+                if sleeps == 10 {
+                    sleep_ms = core::cmp::min(sleep_ms * 10, 250);
```
lzrd, line 225: Same literal constant repeated. Make it symbolic.

task/hiffy/src/main.rs
```diff
@@ -161,66 +161,280 @@ fn main() -> ! {
...
+
+            let rv = execute::<_, NLABELS>(
+                text,
+                HIFFY_FUNCS,
+                data,
+                &mut stack,
+                rstack,
+                &mut *HIFFY_SCRATCH.borrow_mut(),
+                check,
+            );
```
lzrd: We're in an intentional insecure mode. But, do we at least get an integrity check on the network transport? There isn't one at this layer. This is debug/YOLO mode, but a light check to scope a data corruption error would be good.
Maybe just a comment on data integrity for now.

lzrd (commented):
I would fix the repeated literal constants.
It would be nice to see all of the loop timing and retry constants be together.

task/hiffy/src/main.rs
mkeeter: Can you suggest names for these constants?  The first 10 is "number of iterations to run between updates to `sleep_ms`", the second 10 is "scale factor between updates"; neither of them have a good pithy name.

(This is also unchanged from the previous code, just indented)

task/hiffy/src/main.rs
mkeeter: We use IPv6 UDP, which has a 16-bit checksum.  This is true for all Hubris network activity, so calling out Hiffy as uniquely bad doesn't make sense to me.

task/hiffy/src/main.rs
```diff
@@ -161,66 +161,282 @@ fn main() -> ! {
...
         HIFFY_READY.fetch_sub(1, Ordering::SeqCst);
 
-        if HIFFY_KICK.load(Ordering::SeqCst) == 0 {
-            sleeps += 1;
+        #[cfg(feature = "net")]
+        if notif.check_notification_mask(notifications::SOCKET_MASK) {
+            net_state.check_net();
+        }
+
+        if notif.has_timer_fired(notifications::TIMER_MASK) {
```
labbott, line 218: It feels odd to potentially check the net task here and then also have to check the timer. It's probably fine and I think  I'm having a strong reaction to this not being an idol server (hiffy probably pre-dates that?)

task/hiffy/src/main.rs
```diff
@@ -161,66 +161,282 @@ fn main() -> ! {
...
+            HIFFY_KICK.fetch_sub(1, Ordering::SeqCst);
+            sleep_ms = 1;
+            sleeps = 0;
+
+            // TODO without a safety comment explaining why these are safe, it is
+            // not clear if this is sound, do _not_ "fix" this by slapping on an
+            // addr_of_mut! without further analysis!
+            let text = unsafe { &HIFFY_TEXT };
+            let data = unsafe { &HIFFY_DATA };
+            let rstack = unsafe { &mut HIFFY_RSTACK[0..] };
```
labbott: I don't think we're less unsound but do we need further discussion either her or in another comment about interactions between hiffy over probe and hiffy over the network? I can handwave about soundness okay from one external input source (probe) but two (net and probe) makes things more complicated. Is using both net hiffy and probe hiffy in the category of "don't do that"?

mkeeter: How about we bring in https://github.com/oxidecomputer/hubris/pull/2475 first, then I update this PR to use those new primitives (e.g. lifetime-bounded slices)?

task/hiffy/src/main.rs
mkeeter: Can you say more about why it feels odd?  We're waiting on the combined `net` + timer notification bits, so it seems reasonable to check which one(s) fired and proceed accordingly.

I did consider making this an Idol server, but decided against it:

- `hiffy` doesn't expose any public IPCs
- We'd end up doing the same `notification` checks in the `impl NotificationHandler`, with more indirection / overhead from the Idol runtime

task/hiffy/src/main.rs
labbott: yes I think that would make things easier to reason about

task/hiffy/src/main.rs
labbott: I was initially concerned about TOCTOU with the net and timer aligning in weird ways but I don't think that's actually a concern on re-review

task/hiffy/src/main.rs
mkeeter: Done and rebased!

labbott (approved):

PR #2470, opened 2026-04-09
Don't always print stacks

mkeeter:
As much as I love the stack depth estimator, most of the time people don't care about it when running `cargo xtask sizes`; let's hide it behind a flag.


labbott (approved):
I'm kind of neutral about not always printing but I like the arg cleanup

PR #2475, opened 2026-04-10
Make `task-hiffy` less iffy

mkeeter:
`task-hiffy` has bad vibes: it uses `static mut` arrays, and accesses them without an obvious `SAFETY` comment.

The task is necessarily in a weird no man's land, because the debugger writes to those arrays to load HIF text and data; how can Rust know that they have changed?  A fully robust option would be to use `read_volatile` everywhere; unfortunately, downstream libraries (`hif` and eventually `postcard`) expect Rust slices, and we don't want to hack all of them (or double our memory footprint by copying data around).

This PR is an attempt to improve the vibes in three ways:

- Instead of accessing the globals directly, convert them to a pointer and then back into a slice **with a bounded lifetime**.  Previously, we were passing slices with `'static` lifetimes into HIF, which could theoretically stash them
- Add an assembly block which may invalidate any memory, so Rust must assume the arrays may have changed
- Add a SAFETY comment explaining the above

### DIFF
--- a/lib/armv6m-atomic-hack/src/lib.rs
+++ b/lib/armv6m-atomic-hack/src/lib.rs
@@ -35,6 +35,11 @@ pub trait AtomicU32Ext {
     fn swap(&self, val: u32, order: Ordering) -> u32;
     fn fetch_add(&self, val: u32, order: Ordering) -> u32;
     fn fetch_sub(&self, val: u32, order: Ordering) -> u32;
+    fn compare_exchange_acqrel(
+        &self,
+        current: u32,
+        new: u32,
+    ) -> Result<u32, u32>;
 }
 
 #[cfg(armv6m)]
@@ -62,6 +67,21 @@ impl AtomicU32Ext for AtomicU32 {
         self.store(rv.wrapping_sub(val), so);
         rv
     }
+
+    #[inline]
+    fn compare_exchange_acqrel(
+        &self,
+        current: u32,
+        new: u32,
+    ) -> Result<u32, u32> {
+        let rv = self.load(Ordering::Acquire);
+        if rv == current {
+            self.store(new, Ordering::Release);
+            Ok(rv)
+        } else {
+            Err(rv)
+        }
+    }
 }
 
 #[cfg(not(armv6m))]
@@ -79,6 +99,21 @@ impl AtomicU32Ext for AtomicU32 {
     #[inline]
     fn fetch_sub(&self, val: u32, order: Ordering) -> u32 {
         core::sync::atomic::AtomicU32::fetch_sub(self, val, order)
+    }
+
+    #[inline]
+    fn compare_exchange_acqrel(
+        &self,
+        current: u32,
+        new: u32,
+    ) -> Result<u32, u32> {
+        core::sync::atomic::AtomicU32::compare_exchange(
+            self,
+            current,
+            new,
+            Ordering::AcqRel,
+            Ordering::Acquire,
+        )
     }
 }
 
--- a/lib/armv6m-atomic-hack/src/lib.rs
+++ b/lib/armv6m-atomic-hack/src/lib.rs
@@ -35,11 +35,6 @@ pub trait AtomicU32Ext {
     fn swap(&self, val: u32, order: Ordering) -> u32;
     fn fetch_add(&self, val: u32, order: Ordering) -> u32;
     fn fetch_sub(&self, val: u32, order: Ordering) -> u32;
-    fn compare_exchange_acqrel(
-        &self,
-        current: u32,
-        new: u32,
-    ) -> Result<u32, u32>;
 }
 
 #[cfg(armv6m)]
@@ -67,21 +62,6 @@ impl AtomicU32Ext for AtomicU32 {
         self.store(rv.wrapping_sub(val), so);
         rv
     }
-
-    #[inline]
-    fn compare_exchange_acqrel(
-        &self,
-        current: u32,
-        new: u32,
-    ) -> Result<u32, u32> {
-        let rv = self.load(Ordering::Acquire);
-        if rv == current {
-            self.store(new, Ordering::Release);
-            Ok(rv)
-        } else {
-            Err(rv)
-        }
-    }
 }
 
 #[cfg(not(armv6m))]
@@ -99,21 +79,6 @@ impl AtomicU32Ext for AtomicU32 {
     #[inline]
     fn fetch_sub(&self, val: u32, order: Ordering) -> u32 {
         core::sync::atomic::AtomicU32::fetch_sub(self, val, order)
-    }
-
-    #[inline]
-    fn compare_exchange_acqrel(
-        &self,
-        current: u32,
-        new: u32,
-    ) -> Result<u32, u32> {
-        core::sync::atomic::AtomicU32::compare_exchange(
-            self,
-            current,
-            new,
-            Ordering::AcqRel,
-            Ordering::Acquire,
-        )
     }
 }
 
--- a/task/hiffy/src/main.rs
+++ b/task/hiffy/src/main.rs
@@ -14,14 +14,6 @@
 
 #![no_std]
 #![no_main]
-//
-// TODO: Hiffy is using unsafe and static mut in ways that are not obviously
-// sound. This became a warning in early 2024. In the interest of preventing
-// regressions in everything _else_ I'm suppressing the warning here so we can
-// turn Clippy back on. If you're reading this, this file is potentially unsound
-// and needs attention!
-//
-#![allow(static_mut_refs)]
 // This is necessary in order to use the `#[used(compiler)]` attribute on Hiffy
 // statics which are written to by Humility, and must not be optimized out.
 #![feature(used_with_arg)]
@@ -186,27 +178,42 @@ fn main() -> ! {
         sleep_ms = 1;
         sleeps = 0;
 
-        // TODO without a safety comment explaining why these are safe, it is
-        // not clear if this is sound, do _not_ "fix" this by slapping on an
-        // addr_of_mut! without further analysis!
-        let text = unsafe { &HIFFY_TEXT };
-        let data = unsafe { &HIFFY_DATA };
-        let rstack = unsafe { &mut HIFFY_RSTACK[0..] };
-
         let check = |offset: usize, op: &Op| -> Result<(), Failure> {
             trace_execute(offset, *op);
             Ok(())
         };
 
-        let rv = execute::<_, NLABELS>(
-            text,
-            HIFFY_FUNCS,
-            data,
-            &mut stack,
-            rstack,
-            &mut *HIFFY_SCRATCH.borrow_mut(),
-            check,
-        );
+        let rv = {
+            // Dummy object to bind references to a non-static lifetime
+            let lifetime = ();
+
+            // SAFETY: We construct references from our pointers with a limited
+            // (non-static) lifetime, so they can't escape this block.  We are
+            // in single-threaded code, so no one else can read or write to
+            // static memory.  While the HIF program is running, the debugger is
+            // only reading from `HIFFY_REQUESTS` and `HIFFY_ERRORS`; it is not
+            // writing to any locations in memory.
+            let (text, data, rstack) = unsafe {
+                // Use an inline assembly instruction without `nomem`, so the
+                // compiler must assume that any memory can be invalidated (in
+                // this case, by the debugger).
+                core::arch::asm!("", options(nostack, preserves_flags));
+                (
+                    bind_lifetime_ref(&lifetime, &raw const HIFFY_TEXT),
+                    bind_lifetime_ref(&lifetime, &raw const HIFFY_DATA),
+                    bind_lifetime_mut(&lifetime, &raw mut HIFFY_RSTACK),
+                )
+            };
+            execute::<_, NLABELS>(
+                text,
+                HIFFY_FUNCS,
+                data,
+                &mut stack,
+                rstack,
+                &mut *HIFFY_SCRATCH.borrow_mut(),
+                check,
+            )
+        };
 
         match rv {
             Ok(_) => {
@@ -223,4 +230,34 @@ fn main() -> ! {
             }
         }
     }
+}
+
+/// Converts an array pointer to a shared reference with a particular lifetime
+///
+/// # Safety
+/// `ptr` must point to a valid, aligned, initialized `[u8; N]`.
+/// The referent must not be mutated while the returned reference is live.
+#[expect(clippy::needless_lifetimes)] // gotta make it obvious
+unsafe fn bind_lifetime_ref<'a, const N: usize>(
+    _: &'a (),
+    array: *const [u8; N],
+) -> &'a [u8; N] {
+    // SAFETY: converting from pointer to reference is safe given the function's
+    // safety conditions (listed in docstring)
+    unsafe { array.as_ref().unwrap_lite() }
+}
+
+/// Converts an array pointer to a mutable reference with a particular lifetime
+///
+/// # Safety
+/// `ptr` must point to a valid, aligned, initialized `[u8; N]`.
+/// The referent must not be mutated while the returned reference is live.
+#[expect(clippy::needless_lifetimes, clippy::mut_from_ref)]
+unsafe fn bind_lifetime_mut<'a, const N: usize>(
+    _: &'a (),
+    array: *mut [u8; N],
+) -> &'a mut [u8; N] {
+    // SAFETY: converting from pointer to reference is safe given the function's
+    // safety conditions (listed in docstring)
+    unsafe { array.as_mut().unwrap_lite() }
 }
--- a/task/hiffy/src/main.rs
+++ b/task/hiffy/src/main.rs
@@ -18,9 +18,6 @@
 // statics which are written to by Humility, and must not be optimized out.
 #![feature(used_with_arg)]
 
-// This trait may not be needed, if compiling for a non-armv6m target.
-#[allow(unused_imports)]
-use armv6m_atomic_hack::AtomicU32Ext;
 use core::sync::atomic::{AtomicU32, Ordering};
 use hif::*;
 use static_cell::*;
@@ -155,7 +152,7 @@ fn main() -> ! {
         HIFFY_READY.store(0, Ordering::Relaxed);
 
         // Humility writes `1` to `HIFFY_KICK`
-        if HIFFY_KICK.compare_exchange_acqrel(1, 0).is_ok() {
+        if HIFFY_KICK.load(Ordering::Acquire) == 0 {
             sleeps += 1;
 
             // Exponentially backoff our sleep value, but no more than 250ms
@@ -171,6 +168,7 @@ fn main() -> ! {
         // Whenever we have been kicked, we adjust our timeout down to 1ms,
         // from which we will exponentially backoff
         //
+        HIFFY_KICK.store(0, Ordering::Release);
         sleep_ms = 1;
         sleeps = 0;
 
@@ -190,10 +188,6 @@ fn main() -> ! {
             // only reading from `HIFFY_REQUESTS` and `HIFFY_ERRORS`; it is not
             // writing to any locations in memory.
             let (text, data, rstack) = unsafe {
-                // Use an inline assembly instruction without `nomem`, so the
-                // compiler must assume that any memory can be invalidated (in
-                // this case, by the debugger).
-                core::arch::asm!("", options(nostack, preserves_flags));
                 (
                     bind_lifetime_ref(&lifetime, &raw const HIFFY_TEXT),
                     bind_lifetime_ref(&lifetime, &raw const HIFFY_DATA),
@@ -213,7 +207,8 @@ fn main() -> ! {
 
         match rv {
             Ok(_) => {
-                HIFFY_REQUESTS.fetch_add(1, Ordering::Release);
+                let prev = HIFFY_REQUESTS.load(Ordering::Relaxed);
+                HIFFY_REQUESTS.store(prev.wrapping_add(1), Ordering::Release);
                 trace_success();
             }
             Err(failure) => {
@@ -223,7 +218,8 @@ fn main() -> ! {
                 unsafe {
                     HIFFY_FAILURE = Some(failure);
                 }
-                HIFFY_ERRORS.fetch_add(1, Ordering::Release);
+                let prev = HIFFY_ERRORS.load(Ordering::Relaxed);
+                HIFFY_ERRORS.store(prev.wrapping_add(1), Ordering::Release);
                 trace_failure(failure);
             }
         }
--- a/task/hiffy/src/main.rs
+++ b/task/hiffy/src/main.rs
@@ -163,10 +163,7 @@ fn main() -> ! {
         HIFFY_READY.store(0, Ordering::Relaxed);
 
         // Humility writes `1` to `HIFFY_KICK`
-        if HIFFY_KICK
-            .compare_exchange_weak(1, 0, Ordering::Acquire, Ordering::Relaxed)
-            .is_ok()
-        {
+        if HIFFY_KICK.compare_exchange_acqrel(1, 0).is_ok() {
             sleeps += 1;
 
             // Exponentially backoff our sleep value, but no more than 250ms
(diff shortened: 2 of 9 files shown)


jamesmunns:
I'm not personally entirely convinced that inline-asm is useful here. I'll give my take, and then maybe propose what I would like to see. For reference, I had a [very similiar conversation with the opsem team](https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Persistent.20memory.20across.20reboots/with/576068283) for expressing "memory persistent across reboots" (e.g. rtc-mem, or just non-initialized sections with a warm reboot).

IMO, the main point here is that we are doing something spooky:

1. We have a block of memory, INSIDE the abstract machine (e.g. it is a variable known to the compiler as a static)
2. We have somewhat arbitrated access:
  1. The contents are "garbage" until we receive a request to do something. This is arbitrated by the `HIFFY_KICK` variable being non-zero
  2. When the variable IS zero, the contents are garbage
  3. When the variable IS NOT zero, we assume it ~will only change once~ will not change as long as it is non-zero.
  4. We therefore can model the contents of `HIFFY_TEXT`, `HIFFY_DATA`, and `HIFFY_RSTACK` as being contained within a mutex. When KICK is zero, it is locked by "someone else", and we cannot read or write the contents. When KICK is nonzero, the hiffy task has exclusive (ish, others might peek, but not in a way that is meaningful for us) ownership of these regions
3. One ambiguity I'm not sure about, technically we've made things a bit more complicated by immediately subtracting kick, so actually the mutex area becomes "released" when we either increment HIFFY_REQUESTS or HIFFY_ERRORS?
4. I think this PR's addition of non-`'static` lifetime bounds to the slice is probably a net-good: it prevents downstream callers from storing-off the static reference, stashing it somewhere.

In general to avoid the optimizer ruining our day, and to make it "aware" of spooky action at a distance, we need to make the optimizer realize that things can change outside of typical AM interactions. There are two ways to do this with a static:

* Make it a non-mut static `UnsafeCell`. `UnsafeCell`'s contract says that while you are not holding a reference derived from a call to `.get()`, the contents are allowed to change in any spooky way, but while you are holding that reference, it MUST NOT change in spooky ways, e.g. you must ONLY interact with that data through the reference you created. One papercut is that UnsafeCell is `!Sync` so you can't put it in a non-mut static, but that's why I wrote the [`grounded`](https://docs.rs/grounded/latest/grounded/) crate, which does the unsafe impl for you.
* Make it a mut static, AND ensure that is defined that the symbol of the static is "exported", e.g. "linker visible", therefore the compiler/optimizer must assume that usage of this region can "escape" the visibility here, e.g. external code has a chance to poke it. The three main ways to achieve this are:
  * give it an explicit `link_section` name
  * make the static `pub`
  * make the static `no_mangle`
    
Re: this comment:

> A fully robust option would be to use read_volatile everywhere

I somewhat disagree: I think this would likely *work*, but in general volatile is not intended for objects that are "inbounds" of the Rust Abstract Machine: we've given the static a name and a place, which means it is not just some MMIO or external linker section region outside the knowledge of the compiler. Even the [linux kernel](https://docs.kernel.org/process/volatile-considered-harmful.html) agrees volatile is not to be used for synchronization.

The "right" way to do this is generally with fences and ordering guarantees, and firming up our contract of what it means for the region to be "ours", and "not ours". This means likely continuing to use the atomics as we do (AtomicU32s are just `UnsafeCell<u32>`s!), and better define what our synchronization rules are, then treat it like a more boring manual mutex implementation. [Mara's book](https://mara.nl/atomics/memory-ordering.html) is an EXTREMELY good reference for exactly this kind of work.

My general suggestion for what we should be doing (not very far from what we do now):

1. Declare the regions in a way that makes them escaped, e.g. UnsafeCell/static mut+.
2. Optional: Collapse many of the atomic vars, AT LEAST any required for coordination, like READY/KICK, into a single atomic var that uses CAS operations (or critical sections) to guarantee atomicity
3. Determine when the region has data and can be "locked"
4. Do the locking, with appropriate "acquire" fence semantics, and create the scope-bounded mut refs necessary
5. Update any outputs, then drop the references, and mark the operation as complete with a "release" fence, ensuring that all writes are not reordered across the release.

There are also a couple things I don't *love* to see with atomics here:

1. The use of `SeqCst` everywhere, see the section "Myth: Sequentially consistent memory ordering is a great default and is always correct." in the link to Mara's book above
2. There are many separate atomic vars for the "coordination", e.g. both `HIFFY_KICK` and `HIFFY_REQUESTS`/`HIFFY_ERRORS` (or maybe `HIFFY_READY`?)
3. There are some non-CAS operations being done outside of a critical section on one variable, e.g. a separated `HIFFY_KICK.load` and `HIFFY_KICK.fetch_sub`, rather than a CAS loop that decrements atomically if non-zero.
4. Some uses of fetch_add and fetch_sub without checking the output at all, and in cases where `store(1)` or `store(0)` would be more appropriate. There's no actual guard against under/overflow here: atomic ops are always defined to wrap, which means we could underflow or get out of the "expected" band pretty easily.

I don't know how easy it would be to change some of the coordination rules, and update Humility at the same time. But I think just writing down EXACTLY what the rules are for mutual exclusion, we could nudge hiffy into a much sounder place, and then maybe consider if there are potential defects in how we do coordination with multiple atomic variables.

mkeeter:
Very helpful thoughts, thanks @jamesmunns !

My goal for this PR is to make changes which don't require a Humility bump, which limits us somewhat (can't change types of statics, can't mess with the atomic dance).

Here's a quick, somewhat-simplified diagram of how the pieces interact:

### Humility
```mermaid
flowchart LR
    Idle --READY == 1--> Kicked
    Kicked[Write KICK = 1]
    Kicked --REQUEST or ERRORS changed--> Done
    Done[Read results]
    Done --> Idle
```

### Hubris
```mermaid
flowchart LR
    Startup --> Ready
    Ready[Set READY = 1]
    Ready --Delay-->Unready
    Unready[Set READY = 0]
    Unready --KICK == 1--> Running[Clear KICK\nRun script]
    Unready --KICK == 0-->Ready
    Running --Success--> Success[Increment REQUESTS]
    Running --Failure--> Fail[Increment ERRORS]
    Success --> Ready
    Fail --> Ready
```

Data is owned by Hubris when `HIFFY_KICK == 0` and `HIFFY_REQUESTS/ERRORS` are unchanged.  Hubris _increments_ the request/error count when finished, so Humility has loaded the previous value and is checking for a changed value (again, kinda weird design, but let's not break compatibility now).

Follow-up changes based on feedback:

- Use `HIFFY_KICK.compare_exchange_weak(1, 0, Ordering::Acquire, Ordering::Relaxed).is_ok()` to check whether Humility has kicked us, instead of separate load + subtract operations.  In practice, Humility won't be adversarial, but let's do the right thing.
- Use `Ordering::Release` when updating `HIFFY_REQUESTS/ERRORS`
  - Move the `HIFFY_FAILURE` write to be **before** the write to `HIFFY_ERRORS`; this was wrong beforehand!
- Make `HIFFY_READY` writes `Relaxed`: it's a single atomic variable (so the writes are still ordered w.r.t. each other), and it isn't actually used as a memory barrier (just as a flag to Humility)
- Add `#[used]` annotations to the shared arrays (`TEXT` / `DATA` / `RSTACK`)
  - Do you have strong opinions on `#[no_mangle]` vs `#[used]` for making the symbol exported?  It looks like they both export the symbol, Humility hasn't had any issues finding the variables by name, and `#[used]` doesn't require `unsafe` 😅

mkeeter:
Uh oh, some of our CPUs don't actually support `compare_exchange` 🙃 

I'm going to polyfill it in [`AtomicU32Ext`](https://github.com/oxidecomputer/hubris/blob/mkeeter/hiffy-soundness/lib/armv6m-atomic-hack/src/lib.rs#L38) (only for the `AcqRel, Acquire` ordering, because I don't want to [copy this whole table over](https://github.com/rust-lang/rust/blob/14196dbfa3eb7c30195251eac092b1b86c8a2d84/library/core/src/sync/atomic.rs#L3978-L4025)).  This is a little sketchy, but no worse than what we're doing elsewhere in Hiffy, and properly conveys our intent.

jamesmunns:
@mkeeter imo, at least in the embedded rust world at large, there are two main ways to abstract over "atomic or not":

* Use the `portable-atomic` crate, ideally with the `critical-section` backend, to take critical sections to do CAS ops on non-CAS targets. This works best when there is a SINGLE atomic item, like an atomicbool/atomicu32 that controls things.
* Be generic over a [coordination trait](https://docs.rs/bbqueue/latest/bbqueue/traits/coordination/index.html), like I do in bbqueue, that exposes what actions should be done "atomically", and have a CAS and Critical Section impl for each, gated by cpu capabilities.

I actually don't know if we allow tasks to take "real" critical sections, and also I don't know how well this model works with something like a debugger which is... somewhat different than single/multi core modeling. If we don't let tasks disable atomics, there's always [kernel user helpers](https://www.kernel.org/doc/Documentation/arm/kernel_user_helpers.txt) like what linux offers for old archs.

> Do you have strong opinions on #[no_mangle] vs #[used] for making the symbol exported?

I'm not *certain* `#[used]` lifts the symbol to public. We should at least make it `pub`. IMO what we're doing is unsafe, we shouldn't necessarily steer away from `#[unsafe(no_mangle)]`. I'd have to check with the opsem folks about which attrs are guaranteed to "lift" a static's visibility.

Re: reviewing atomic ordering, I don't have my best thinking cap on right now, I'll see if my head is a little less cloudy tomorrow to see if all of your assertions wrt ordering make sense to me. I would also deeply trust @hawkw's opinions on the matter. If READY is part of the synchronization between the debugger and task (we can model the debugger's interactions as if it was a separate thread, running on the "debug" core), I'm a little sus at making it Relaxed.

hawkw:
I don't really think we _should_ be attempting to "abstract over atomic or not" here. Typically, it is only safe to allow something that must be atomic on multi-threaded platforms to be non-atomic on other platforms because the platforms where it is not atomic are inherently always single-threaded. That doesn't really apply here: *all* our targets are single-threaded, but in this code, there is always the possibility that we will be interrupted at any point by:
- an interrupt context switching us into the kernel, or,
- (perhaps more importantly), the debug probe stopping the world

I don't think it makes sense to think of the debugger as "another thread", exactly, since when the debugger is operating on us, we are halted. This makes it somewhat more similar to an interrupt than a concurrent thread of execution: it may halt us at any time, potentially in the midst of a non-atomic, multi-instruction operation, but we are not concurrently mutating or reading things when it has halted us.

Given that, I don't think we can safely rely on compare-and-swap operations in `hiffy` if they may not be atomic on some targets. We cannot, AFAIK, rely on a crate like `critical-section` as @bitshiftmask suggests, because `critical-section` will only prevent _hardware interrupts_ by temporarily disabling them for the duration of the critical section. But, disabling interrupts does not prevent a SWD probe from halting us (and I dont immediately know of a way we can refer a SWD halt, though perhaps there's some way to do so...). Instead, we might need to consider something like a sequence lock, which does not rely on `critical-section`, on systems which don't have CAS operations. 

I haven't had a chance to actually review the code here yet, but I intend to; this is mostly a response to James' last comment.

mkeeter:
> I'm not certain #[used] lifts the symbol to public. We should at least make it pub. IMO what we're doing is unsafe, we shouldn't necessarily steer away from #[unsafe(no_mangle)]. I'd have to check with the opsem folks about which attrs are guaranteed to "lift" a static's visibility.

This was my interpretation of the [Rust reference](https://doc.rust-lang.org/reference/abi.html#r-abi.used), which says the following about `#[used]`:

> The used attribute can only be applied to [static items](https://doc.rust-lang.org/reference/items/static-items.html). This [attribute](https://doc.rust-lang.org/reference/attributes.html) forces the compiler to keep the variable in the output object file (.o, .rlib, etc. excluding final binaries) even if the variable is not used, or referenced, by any other item in the crate. However, the linker is still free to remove such an item.

`#[no_mangle]` says the following:

> The no_mangle attribute may be used on any [item](https://doc.rust-lang.org/reference/items.html) to disable standard symbol name mangling. The symbol for the item will be the identifier of the item’s name.
> 
> Additionally, the item will be publicly exported from the produced library or object file, similar to the [used attribute](https://doc.rust-lang.org/reference/abi.html#the-used-attribute).

My reading of this is that `#[used]` does the part which we care about (public export), since `#[no_mangle]` defines this behavior in terms of `#[used]` (although it uses the phrase "similar to", which is annoyingly imprecise).

For the atomics wrangling, the module-level docs for `armv6m-atomic-hack` provide some additional context:

https://github.com/oxidecomputer/hubris/blob/2d57a804e6fca87af5fde674d7b146d74bf91627/lib/armv6m-atomic-hack/src/lib.rs#L5-L28

It's also worth noting that in our existing hiffy, the `fetch_add` and `fetch_sub` operations are _already_ implemented using this polyfill.

lib/armv6m-atomic-hack/src/lib.rs
```diff
@@ -35,6 +35,11 @@ pub trait AtomicU32Ext {
     fn swap(&self, val: u32, order: Ordering) -> u32;
     fn fetch_add(&self, val: u32, order: Ordering) -> u32;
     fn fetch_sub(&self, val: u32, order: Ordering) -> u32;
+    fn compare_exchange_acqrel(
+        &self,
+        current: u32,
+        new: u32,
+    ) -> Result<u32, u32>;
```
hawkw: it seems like, if we_are_ going to implement this in a non-atomic multi-instruction sequence, it wouldn't be _that_ much harder to just let the caller pass in orderings, instead of making it `_acqrel`...

task/hiffy/src/main.rs
```diff
@@ -162,11 +158,12 @@ fn main() -> ! {
     HIFFY_VERSION_PATCH.fetch_add(0, Ordering::SeqCst);
 
     loop {
-        HIFFY_READY.fetch_add(1, Ordering::SeqCst);
+        HIFFY_READY.store(1, Ordering::Relaxed);
         hl::sleep_for(sleep_ms);
-        HIFFY_READY.fetch_sub(1, Ordering::SeqCst);
+        HIFFY_READY.store(0, Ordering::Relaxed);
```
hawkw, line 198: I agree that these needn't be `SeqCst`. However, I think we _may_ actually be concerned about the second store to `HIFFY_READY` being reordered around the `compare_exchange`, which may be multiple instructions, to `HIFFY_KICK`.

task/hiffy/src/main.rs
```diff
@@ -162,11 +158,12 @@ fn main() -> ! {
...
 
-        if HIFFY_KICK.load(Ordering::SeqCst) == 0 {
+        // Humility writes `1` to `HIFFY_KICK`
+        if HIFFY_KICK.compare_exchange_acqrel(1, 0).is_ok() {
```
hawkw: I'm unconvinced by making this a CAS especially given that it may not actually be a CAS on some supported hardware. I think that in the previous code, _and_ in the case where we are on a non-CAS MCU, the debugger could observe the state where we have been kicked, but have not yet decremented `HIFFY_KICK`, but...I don't actually think that _matters_, since it's not going to kick us again, right? And, the tradeoff is that the `compare_exchange` makes it less obvious that this may or may not be possible by tricking the reader into believing it's actually a LDREX/STREX...

task/hiffy/src/main.rs
```diff
@@ -182,45 +179,91 @@ fn main() -> ! {
...
+            // (non-static) lifetime, so they can't escape this block.  We are
+            // in single-threaded code, so no one else can read or write to
+            // static memory.  While the HIF program is running, the debugger is
+            // only reading from `HIFFY_REQUESTS` and `HIFFY_ERRORS`; it is not
+            // writing to any locations in memory.
+            let (text, data, rstack) = unsafe {
+                // Use an inline assembly instruction without `nomem`, so the
+                // compiler must assume that any memory can be invalidated (in
+                // this case, by the debugger).
+                core::arch::asm!("", options(nostack, preserves_flags));
```
hawkw: Hmm, I think I trust you about this...

mkeeter: It's a little hand-wavey, but does seem to accomplish something:
https://godbolt.org/z/odEExhxYz

task/hiffy/src/main.rs
mkeeter: I'm fine either way – the polyfill expresses intent more clearly, but explicit load-stores are less misleading.  However, if we don't use the polyfill for CAS, we should also stop using it for `fetch_add` elsewhere in this code.  Deal?

jamesmunns:
@mkeeter:

> My reading of this is that #[used] does the part which we care about (public export), since #[no_mangle] defines this behavior in terms of #[used] (although it uses the phrase "similar to", which is annoyingly imprecise).

I'm less certain (but we should probably go ask T-Opsem), what I'm worried about is:

* `used` ensures the item is not optimized away, and therefore is visible to the program and debugger, BUT
* I am not sure if it makes the compiler aware that *other* **software components** can poke at it. I think we need the compiler to be convinced that other **software components** have a chance to peek and poke at the data, NOT just the debugger, as I'm not sure if the compiler really considers the role of a debugger in that aspect

I'm going off of [#t-opsem > Persistent memory across reboots @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Persistent.20memory.20across.20reboots/near/576070893), from Amanieu:

Screenshot 2026-04-14 at 3 19 47 PM

`used` was not explicitly mentioned there, so I'm less sure.

task/hiffy/src/main.rs
jamesmunns: I still think this is a roundabout way to achieve what `fence`/`compiler_fence` is designed for: https://godbolt.org/z/o9x1G47fd

mkeeter: Okay, I'm now coming around to the idea that the inline assembly is not necessary – but that a compiler fence *also* isn't necessary, because of the `Acquire` load of `HIFFY_KICK` above!

We know that Humility only writes `HIFFY_KICK` after writing values to the data arrays in RAM.  Doing an `Acquire` load of an atomic both forces the value to be loaded from RAM and adds a `dmb`, so it's already accomplishing the goals of `compiler_fence` / `fence`: https://godbolt.org/z/6ba7WhTP6

task/hiffy/src/main.rs
mkeeter: I've rewritten the `compare_exchange` as an explicit `load/store` pair, so I think we're now good on reordering, even with these marked as `Relaxed` (the load comes first and uses `Ordering::Acquire`).

task/hiffy/src/main.rs
mkeeter: I've replaced the polyfilled compare-and-swap with an explicit load/store pair, and done the same for the `fetch_add` operations below!

jamesmunns:
@mkeeter drafted some diagrams for you (let me know if you have notes/want the monodraw file):

```text
┌─────────────────────────────────────────────────────────────────────────────────┐
│                                                                                 │
│                                ┌────────────────┐                ┌────────┐     │
│          ┌────────┐ READY == 1 │ Read REQUEST   │   REQUEST or   │ Read   │     │
│        ┌▷│  Idle  │───────────▷│ Read ERRORS    │───────────────▷│ RESULT │     │
│        │ └────────┘            │ Write KICK = 1 │ ERRORS changed │        │     │
│        │                       └────────────────┘                └────────┘     │
│        │                                                              │         │
│        └──────────────────────────────────────────────────────────────┘         │
│ ┌──────────┐                                                                    │
└─┤ Humility ├────────────────────────────────────────────────────────────────────┘
  └──────────┘                                                                     
                                                                                   
┌─────────────────────────────────────────────────────────────────────────────────┐
│                                                                                 │
│                                        KICK == 0                                │
│                    ┌────────────────────────────────────────────────┐           │
│                    │                                                │           │
│                    │                                       ┌─────────────────┐  │
│  ┌─────────┐       │  ┌─────────────────┐     ┌───────┐    │ Write READY = 0 │  │
│  │ Startup │──┬────┴─▷│ Write READY = 1 │────▷│ Sleep │───▷│ Read KICK       │  │
│  └─────────┘  │       └─────────────────┘     └───────┘    │                 │  │
│               │                                            └─────────────────┘  │
│               │                                                     │           │
│               │  ┌───────────────────────────────┐                  ▽           │
│               │  │ Read REQUESTS                 │ Success ┌─────────────────┐  │
│               ├──│ Write REQUESTS = REQUESTS + 1 │◁────┐   │ Write KICK = 0  │  │
│               │  └───────────────────────────────┘     ├───│ Execute script  │  │
│               │  ┌───────────────────────────────┐     │   │                 │  │
│               │  │ Read ERRORS                   │     │   └─────────────────┘  │
│               └──│ Write ERRORS = ERRORS + 1     │◁────┘                        │
│                  └───────────────────────────────┘ Failure                      │
│ ┌────────────┐                                                                  │
└─┤ Hiffy Task ├──────────────────────────────────────────────────────────────────┘
  └────────────┘                                                                                                                                                      
```

task/hiffy/src/main.rs
```diff
@@ -116,8 +106,11 @@ cfg_if::cfg_if! {
 // - [`HIFFY_READY`]      => Variable that will be non-zero iff the HIF
 //                           execution engine is waiting to be kicked
 //
+#[used]
```
jamesmunns: @mkeeter somewhat insisting on no_mangle! I think `pub` + `used` might be enough, but I don't think there is any particular downside to picking `pub` + `no_mangle` instead, so let's be as "loud" as possible to the compiler about it. Also wrote up some justification/explanation here that I think is worth having for future reader's context.

The pub+no_mangle should apply to all three of these, as well as the four tracking variables and `HIFFY_FAILURE`.

```suggestion
// We are making the following items "no mangle" and "pub" to hint to the compiler
// that they are "exposed", and may be written (spookily) outside the scope
// of Rust itself. The aim is to prevent the optimizer from *assuming* the
// contents of these buffers will remain unchanged between accesses, as they
// will be written directly by the debugger.
//
// Below, we use atomic ordering (e.g. Acquire and Release) to inhibit
// compile- and run-time re-ordering around the explicit sequencing performed
// by the HIFFY_READY, HIFFY_KICK, HIFFY_REQUESTS, and HIFFY_ERRORS that are
// used to arbitrate shared access between the debugger and this software task.
#[unsafe(no_mangle)]
```

task/hiffy/src/main.rs
```diff
@@ -182,45 +168,90 @@ fn main() -> ! {
...
-            &mut stack,
-            rstack,
-            &mut *HIFFY_SCRATCH.borrow_mut(),
-            check,
-        );
+        let rv = {
+            // Dummy object to bind references to a non-static lifetime
+            let lifetime = ();
+
+            // SAFETY: We construct references from our pointers with a limited
```
jamesmunns, line 230: Could be worth mentioning the explanation + diagram above for this safety block

task/hiffy/src/main.rs
```diff
@@ -182,45 +168,90 @@ fn main() -> ! {
...
         match rv {
             Ok(_) => {
-                HIFFY_REQUESTS.fetch_add(1, Ordering::SeqCst);
+                let prev = HIFFY_REQUESTS.load(Ordering::Relaxed);
+                HIFFY_REQUESTS.store(prev.wrapping_add(1), Ordering::Release);
                 trace_success();
             }
             Err(failure) => {
-                HIFFY_ERRORS.fetch_add(1, Ordering::SeqCst);
+                // SAFETY: We are in single-threaded code and the debugger will
```
jamesmunns, line 262: Could be worth mentioning the explanation + diagram above for this safety block

jamesmunns (approved):
Overall happy with this, modulo the addition of the state diagrams, and the one main comment below.

jamesmunns:
Oh, and overall: I think this change is just spiffy.

mkeeter:
@jamesmunns I've added the diagrams, `no_mangle` block comment, and switched from `#[used]` to `#[unsafe(no_mangle)]` for the relevant variables (and made them `pub`).

Interestingly, I need to keep `#[used]` on the `HIFFY_VERSION_*` variables for them to be present in the output binary; they're not touched by Hubris code, so that's presumably fine.

I also switched the `HIFFY_FUNCTIONS` to `#[unsafe(no_mangle)]`; it was previously a mix of `#[unsafe(no_mangle)]` and `#[used(compiler)]`.  These aren't used in our code, but must be present so that Humility can extract types from the DWARF info 🫠

cbiffle (approved):

PR #2479, opened 2026-04-16
[xtask] Add support for `xtask doc` commands

### LINKED ISSUE #2422: an xtask for building RustDoc for Hubris crates would be nice

jamesmunns:
This PR factors out a `passthrough` helper, and uses it for `doc` and `clippy` invocations.

Closes #2422 

This somewhat punts on the flexibility items that @hawkw raised in #2422, as supporting both "before the `--`" and "after the `--`" args, in addition to the variable-length task list, was trickier than I wanted to get involved with in `clap`. That being said: it's not a feature anyone has needed yet, so I feel comfortable leaving that part as a todo.

### DIFF
--- a/build/xtask/src/main.rs
+++ b/build/xtask/src/main.rs
@@ -11,7 +11,6 @@ use crate::config::Config;
 
 mod auxflash;
 mod caboose_pos;
-mod clippy;
 mod config;
 mod dist;
 mod elf;
@@ -20,6 +19,7 @@ mod gha_prepare_artifacts;
 mod graph;
 mod humility;
 mod lsp;
+mod passthrough;
 mod print;
 mod sizes;
 mod task_slot;
@@ -154,6 +154,24 @@ enum Xtask {
         /// Configures the caboose for the generated archive.
         #[clap(flatten)]
         caboose_args: CabooseArgs,
+    },
+
+    /// Runs `cargo doc` on a specified task
+    Doc {
+        /// Request verbosity from tools we shell out to.
+        #[clap(short)]
+        verbose: bool,
+
+        /// Path to the image configuration file, in TOML.
+        cfg: PathBuf,
+
+        /// Name of task(s) to check.
+        tasks: Vec<String>,
+
+        /// Arguments passed directly to the `cargo doc` invocation,
+        /// e.g. `--open`
+        #[clap(last = true)]
+        doc_args: Vec<String>,
     },
 
     /// Runs `cargo clippy` on a specified task
@@ -489,7 +507,26 @@ fn run(xtask: Xtask) -> Result<()> {
             tasks,
             extra_options,
         } => {
-            clippy::run(verbose, cfg, &tasks, &extra_options)?;
+            // Clippy commands are passed AFTER the `--`, currently pre-options
+            // are not supported.
+            passthrough::run(
+                "clippy",
+                verbose,
+                cfg,
+                &tasks,
+                &[],
+                &extra_options,
+            )?;
+        }
+        Xtask::Doc {
+            verbose,
+            cfg,
+            tasks,
+            doc_args,
+        } => {
+            // Doc commands are passed BEFORE the `--`, currently post-options
+            // are not supported.
+            passthrough::run("doc", verbose, cfg, &tasks, &doc_args, &[])?;
         }
         Xtask::TaskSlots { task_bin } => {
             task_slot::dump_task_slot_table(&task_bin)?;
--- a/build/xtask/src/passthrough.rs
+++ b/build/xtask/src/passthrough.rs
@@ -8,11 +8,27 @@ use anyhow::{Result, bail};
 
 use crate::config::Config;
 
+/// A passthrough function for plumbing basic `cargo ...` commands in the
+/// context of a hubris build.
+///
+/// ## Arguments
+///
+/// * `cargo_cmd`: The cargo subcommand, e.g. `doc` or `clippy`
+/// * `verbose`: Toggles verbosity
+/// * `cfg`: Path to the `app.toml`
+/// * `tasks`: List of task names to run this command on, will run cargo
+///   subcommand on EACH specified task
+/// * `direct_opts`: Arguments to be passed BEFORE the `--` of the command
+///   invocation, e.g. `cargo subcommand $DIRECT_OPTS`
+/// * `post_opts`: Arguments to be passed AFTER the `--` of the command
+///   invocation, e.g. `cargo subcommand -- $POST_OPTS`
 pub fn run(
+    cargo_cmd: &str,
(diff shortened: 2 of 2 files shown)


jamesmunns:
Ah, I did just notice one limitation of my approach:

Like `xtask clippy`, it is bound to *task* invocations, not `package` limitations. This may not actually fulfull the original request by @hawkw.

For example, this *does* work:

```sh
cargo xtask doc app/cosmo/rev-b.toml packrat -- --open
```

But this *does not* work:

```sh
cargo xtask doc app/cosmo/rev-b.toml ereports -- --open
```

edit: This *does* work, but requires some knowledge of which crate is used in which app+task:

```sh
cargo xtask doc app/grapefruit/rev-b.toml ereportulator -- -p ereports --open
```

mkeeter:
(I haven't read the implementation yet, but here's some thoughts on CLI stuff)

On one hand, it only makes sense to document packages in the context of a specific `(manifest, task)` tuple – tasks may enable / disable features in downstream crates which change docs.

On the other hand, it would be nice if this Just Worked™ in a sensible way.  We do something similar in `cargo xtask lsp` by hard-coding a (increasingly out-of-date) set of manifests to check, then finding the first task in one of those manifests which uses the crate containing a source file.

I can think of a few ways to organize this at the CLI level:

- `cargo xtask doc app/cosmo/rev-b.toml [--task task] [target]`, where `target` is treated as either a crate or a task depending on whether `--task` is present (and if `--task` is present but `target` is absent, we target the task)
- `cargo xtask doc app/cosmo/rev-b.toml [--task task] [--crate crate]`, where we either document the task or the crate.  This could also use the "find the first task which enables that package" strategy if `--task` is missing

I lean slightly towards the former, because it matches our existing CLI (e.g. `cargo xtask build` and the current `cargo xtask clippy`), but could be persuaded otherwise.

mkeeter (approved):

jamesmunns:
We decided to merge this as-is, and to look into another docs solution in the near future.

PR #2481, opened 2026-04-17
[xtask] Add `aggro` subcommand for building aggregate docs

jamesmunns:
I've currently plumbed a good "base coat" of the `aggro` subcommand (name bikeshedding welcome, it was funny on a Friday morning), which generally:

* Takes a task config
* Figures out all the tasks in that config
* Splats all of the READMEs (that are present) into a single bundle, and renders them into a single HTML "report".

Currently, it looks like this:

Screenshot 2026-04-20 at 15-53-02 cosmo-b Aggregate Docs

[aggro.html](https://github.com/user-attachments/files/26898729/aggro.html)

From now, there are a couple of questions:

1. I could use some feedback on "matters of taste":
  * What information WOULD be useful to document on a per-app, or a per-task basis?
  * I included some simple tables, but these should be considered a proof of concept.
  * How do we want to shape this? One big MD/HTML/PDF file? Something more multi-page organized like `mdbook`?
2. I think we'll want to make a pass over our task docs:
  * Most tasks don't have any docs! But we can backfill this :)
  * We probably want to set out a bit of a "standard" of how drivers+tasks+apps are documented

I did try figuring out how to insert the task graph, sadly the `launch-rs` crate isn't as good at layouting as the `dot` bin is, and neither are *great* for making legible diagrams. We may want to break this up on a per-task basis just to reduce the number of entities per-graph.

Screenshot 2026-04-17 at 7 33 04 PM

### DIFF
--- a/build/xtask/src/aggro.rs
+++ b/build/xtask/src/aggro.rs
@@ -1,6 +1,8 @@
 use crate::config::Config;
 use anyhow::Result;
+use indexmap::IndexMap;
 use pulldown_cmark::{Event, HeadingLevel, Tag, TagEnd, html};
+use ordered_toml::Value;
 use std::{fmt::Write as _, fs, io::Write as _, path::Path};
 use toml_task::Task;
 
@@ -93,7 +95,7 @@ pub fn run(app_toml: &Path, output: Option<&Path>) -> Result<()> {
     write_app_info(&cfg, &mut html_buf)?;
 
     // STAGE 3: Task Header
-    write_task_header(&cfg, &mut html_buf)?;
+    write_task_header(&cfg, &cfg.tasks, &mut html_buf)?;
 
     // STAGE 4: Document each task
     for (_name, docpath, task) in task_docs {
@@ -121,6 +123,54 @@ fn write_app_header(cfg: &Config, buf: &mut String) -> Result<()> {
     writeln!(&mut mkdn, "# \"{}\" Application", cfg.name)?;
 
     // TODO: What else do we want here? Stuff about the app, not yet the docs?
+    struct IoWrite(String);
+    impl std::io::Write for IoWrite {
+        fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+            let Ok(s) = std::str::from_utf8(buf) else {
+                return Err(std::io::Error::other("not utf-8?"));
+            };
+            self.0.push_str(s);
+            Ok(buf.len())
+        }
+
+        fn flush(&mut self) -> std::io::Result<()> {
+            Ok(())
+        }
+    }
+    let mut dotout = IoWrite(String::new());
+    crate::graph::task_graph_inner(&cfg.app_toml_path, &mut dotout)?;
+
+    // TODO: the `layout` crate doesn't handle comments, filter these
+    let mut filtered = String::new();
+    for line in dotout.0.lines() {
+        if line.trim_start().starts_with("#") {
+            continue;
+        }
+        filtered.push_str(line);
+        filtered.push_str("\n");
+    }
+
+    let mut parser = layout::gv::DotParser::new(&filtered);
+    let graph = match parser.process() {
+        Ok(g) => g,
+        Err(e) => anyhow::bail!("Graphing error: '{e}'"),
+    };
+    let mut gb = layout::gv::GraphBuilder::new();
+    gb.visit_graph(&graph);
+    let mut vg = gb.get();
+    let mut svg = layout::backends::svg::SVGWriter::new();
+    vg.do_it(false, false, false, &mut svg);
+    let content = svg.finalize();
+
+    let mut file = std::fs::File::create("/tmp/layout.svg")?;
+    file.write_all(content.as_bytes())?;
+    file.flush()?;
+    drop(file);
+
+    writeln!(buf, "<svg>")?;
+    buf.push_str(&content);
+    writeln!(buf, "</svg>")?;
+
 
     // Write to HTML. We *don't* do touchup, because this is the top level
     let parser = pulldown_cmark::Parser::new_ext(&mkdn, PULLDOWN_OPTS);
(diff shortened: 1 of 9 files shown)


jamesmunns:
Thing to add: UDP sockets by task + port

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+#[derive(Default, Debug)]
+struct TaskMeta {
+    calls: HashSet<String>,
+    called_by: HashSet<String>,
+}
+
+pub fn run(app_toml: &Path, output: Option<&Path>) -> Result<()> {
+    let cfg = Config::from_file(app_toml)?;
+
+    println!("{}", std::env::current_dir().unwrap().display());
```
mkeeter, line 33: Remove debug print?

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+                    matches.push(path);
+                }
+
+                match matches.as_slice() {
+                    // No matches
+                    [] => None,
+                    // Exactly one match
+                    [found] => Some(found.clone()),
+                    _ => {
+                        panic!("too many readmes");
```
mkeeter, line 109: I've vote to log an error (including the task name) and continue instead of panicking, on the rare chance that a task has both `readme.md` and `readme.mkdn`.

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+    buf: &mut String,
+    meta: &HashMap<String, TaskMeta>,
+) -> Result<()> {
+    // Write this as markdown for laziness, then HTMLify it
+    let mut mkdn = String::new();
+    writeln!(&mut mkdn, "# \"{}\" Tasks", cfg.name)?;
+    writeln!(&mut mkdn)?;
+    writeln!(
+        &mut mkdn,
+        "| task: crate | priority | stack (bytes) | interrupts | client of | server for |"
```
mkeeter, line 194: I find "client of" / "server for" confusing, maybe "called by" and "calls into"?

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+        tasks.iter().map(|(a, b)| (a.clone(), b)).collect();
+    tasks.sort_unstable_by(|a, b| task_sort((&a.0, a.1), (&b.0, b.1)));
+
+    for (name, task) in tasks.iter() {
+        let prio = task.priority.to_string();
+
+        let stack = if let Some(amt) = task.stacksize {
+            amt.to_string()
+        } else {
+            "???".to_string()
```
mkeeter, line 210: Use `cfg.stacksize` as fallback before declaring defeat here.

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+        let mut base = readme.to_owned();
+        base.pop();
+        let stream = parser.map(|evt| touchup(evt, Some(&base)));
+        html::push_html(buf, stream);
+    } else {
+        // Placeholder for no docs!
+        //
+        // Write this as markdown for laziness, then HTMLify it
+        let mut mkdn = String::new();
+        writeln!(&mut mkdn, "# `{}` docs", task.name)?;
```
mkeeter, line 284: Let's remove this, since it just adds busyness to the page.

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+        writeln!(&mut mkdn)?;
+        writeln!(&mut mkdn, "(this page intentionally left blank)")?;
```
mkeeter, line 286:
```suggestion
        writeln!(&mut mkdn, "(no README available)")?;
```
to make it more clear what's going on

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+            match tag {
+                Tag::Link {
+                    link_type,
+                    dest_url,
+                    title,
+                    id,
+                } => {
+                    if !dest_url.starts_with("http") {
+                        // TODO: rewrite relative to `https://github.com/oxidecomputer/hubris/blob/master/`?
+                        // use _base to figure out relative paths, we might also need to
```
mkeeter, line 309: might also need to what??

build/xtask/src/graph.rs
```diff
@@ -15,12 +15,20 @@ use crate::config::Config;
 /// Generate a directed graph of task priorities and task_slot
 /// dependencies.
 pub fn task_graph(app_toml: &Path, path: &Path) -> Result<()> {
+    let dot = File::create(path)?;
```
mkeeter, line 18: Is this prep for including the graph in the docs?  I don't see it used anywhere.

build/xtask/src/config.rs
```diff
@@ -88,6 +89,7 @@ pub struct Config {
     pub app_config: String,
     pub auxflash: Option<AuxFlashData>,
     pub caboose: Option<CabooseConfig>,
+    pub docfile: Option<PathBuf>,
```
mkeeter, line 92: No one is using this yet, right?

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+                }
+                Tag::Image {
+                    link_type,
+                    dest_url,
+                    title,
+                    id,
+                } => {
+                    if !dest_url.starts_with("http") {
+                        // TODO: rewrite relative to `https://github.com/oxidecomputer/hubris/blob/master/`?
+                        // use _base to figure out relative paths, we might also need to
```
mkeeter, line 331: same here

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+                        println!("WARN: We should be rewriting {dest_url}!");
+                    }
+                    Event::Start(Tag::Image {
+                        link_type,
+                        dest_url,
+                        title,
+                        id,
+                    })
+                }
+                // Bump down headings one notch, to allow for top level docs
```
mkeeter, line 341: Weirdly, I still see `h1` headings for tasks in your sample file:

Image

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+}
+
+fn write_task_info(
+    name: &str,
+    task: &Task,
+    docs: Option<&Path>,
+    buf: &mut String,
+) -> Result<()> {
+    let mut mkdn = String::new();
+    writeln!(&mut mkdn, "# Task: \"{name}\" (`{}`)", task.name)?;
```
mkeeter, line 262: vibes:
```suggestion
    writeln!(&mut mkdn, "# `{name}` (`{}`)", task.name)?;
```
The task name is also used as an identifier in code, and I think it's overkill for the header to tell us it's a task.

build/xtask/src/config.rs
mkeeter: Ah, I see it's just Cosmo right now.

build/xtask/src/aggro.rs
```diff
@@ -0,0 +1,456 @@
...
+    writeln!(&mut mkdn, "# Application: \"{}\"", cfg.name)?;
+    writeln!(&mut mkdn)?;
+
+    // TODO: Application level docs?
+
+    // Write to HTML.
+    let parser = pulldown_cmark::Parser::new_ext(&mkdn, PULLDOWN_OPTS);
+    html::push_html(buf, parser);
+
+    if let Some(readme) = cfg.docfile.as_ref() {
```
mkeeter, line 156: Should we try to find a README if this is empty, based on Cargo metadata?

(I'm fine with this being a later PR)

PR #2483, opened 2026-04-21
auxflash-server: validate slot index to prevent out-of-bounds writes

evan-oxide:
Before, you could try to write (or read) a slot index higher than the number of slots. If there was extra unused space on the flash chip past the slots, this bug would presumably let you write to the unused space. The grapefruit I tested on doesn't have unused space, so it reported a `QspiTransferError`. Now with the bounds check added, it reports `InvalidSlot` instead.

This would also have been fixed by https://github.com/oxidecomputer/hubris/pull/2296


mkeeter (approved):
🎉

PR #2485, opened 2026-04-22
Fix multi-region reads in dumps

mkeeter:
While testing unrelated code, I noticed that reading a particular variable over the network failed.  I switched to `humility readmem`, and ran into the same issue:

```
➜  hubris jj:(zuuk) h readmem 0x24021f70
humility: connecting to fe80::c1d:8cff:fec0:e207%28
humility readmem failed: 0x24021f70 can't be read via the archive or over the network

Caused by:
    dump agent failed: invalid response: Err(BadSegmentAdd)
```

Suspiciously, the buffer in question spans multiple MPU regions:
```
0x080057a8 0x24021a00 - 0x24021bff     512 rw---- 5  packrat
0x080057bc 0x24021c00 - 0x24021fff    1KiB rw---- 5  packrat
0x080057d0 0x24022000 - 0x24023fff    8KiB rw---- 5  packrat
```

It turns out that this is the userland equivalent of #1674: `jefe` checks to see if a read was contained within a single MPU region, but would reject reads which span multiple regions (even if they're contiguous and owned by the same task).

In this PR, I rely on the fact that region descriptors are sorted by base address to incrementally check for overlaps.  I also add some documentation about `get_task_dump_region`'s behavior and guarantees: it's got special behavior for the 0th region, which may not be sorted*

*usually, kernel memory is below task memory, but now that we can put tasks in `dtcm`, that's not always true!

### DIFF
--- a/doc/kipc.adoc
+++ b/doc/kipc.adoc
@@ -260,6 +260,13 @@ the dump region index.  If the dump region index is equal to or greater
 than the number of dump regions for the specified task, `None` will
 be returned.
 
+Passing a dump region of 0 returns the task's descriptor in kernel memory, and
+will always return `Some(..)` (for a valid task).  Subsequent regions are
+guaranteed to be sorted by (ascending) base address.
+
+CAUTION: The descriptor region may not be sorted with respect to other regions;
+kernel memory may be either above or below task memory.
+
 === `read_task_dump_region` (7)
 
 For a given task and task dump region, this will read the specified region and
--- a/task/jefe/src/dump.rs
+++ b/task/jefe/src/dump.rs
@@ -261,7 +261,7 @@ pub fn dump_task(base: u32, task: usize) -> Result<u8, DumpAgentError> {
         // regions in a task) -- but could become so if these numbers become
         // larger.
         //
-        let Some(region) = kipc::get_task_dump_region(task, ndx) else {
+        let Some(region) = kipc::get_task_dump_region_raw(task, ndx) else {
             break;
         };
         if in_dump_area(region.base, region.size) {
@@ -323,9 +323,8 @@ pub fn dump_task_region(
     let mem = start..end;
     let mut okay = false;
 
-    // The kernel descriptor is always dump region 0, which is always valid to
-    // read (and can therefore be unwrapped).
-    let desc = kipc::get_task_dump_region(task, 0).unwrap_lite();
+    // Get the task descriptor region (in kernel memory)
+    let desc = kipc::get_task_desc_region(task);
     // Note: we implicitly trust kipc won't give us a region that wraps,
     // unlike untrusted user data from the request that we checked above.
     let desc_region = desc.base..desc.base + desc.size;
@@ -342,7 +341,8 @@ pub fn dump_task_region(
         // Note: we also implicitly trust that kipc gives us regions which are
         // in sorted order by base address.
         let mut mem = start..end;
-        for ndx in 1..=usize::MAX {
+        let mut started = false;
+        for ndx in 0..=usize::MAX {
             // This is Accidentally Quadratic; see the note in `dump_task`
             let Some(region) = kipc::get_task_dump_region(task, ndx) else {
                 break;
@@ -358,13 +358,17 @@ pub fn dump_task_region(
             let region = region.base..region.base + region.size;
             if region.contains(&mem.start) {
                 mem.start = region.end.min(mem.end);
+                started = true;
                 if mem.start == mem.end {
                     okay = true;
                     break;
                 }
-            } else if region.start > mem.start {
-                // If we are beyond the start of our `mem` region, then there
-                // are no more overlaps and we can bail out immediately.
+            } else if region.start > mem.start
+                || (started && region.start < mem.start)
+            {
+                // If we are beyond the start of our `mem` region (or have
+                // started overlapping but this region does not overlap), then
+                // there are no more overlaps and we can bail out immediately.
                 break;
             }
         }
--- a/task/jefe/src/dump.rs
+++ b/task/jefe/src/dump.rs
@@ -247,25 +247,11 @@ pub fn dump_task(base: u32, task: usize) -> Result<u8, DumpAgentError> {
 
     let area = dump_task_setup(base, DumpTaskContents::SingleTask)?;
 
-    for ndx in 0..=usize::MAX {
-        //
-        // We need to ask the kernel which regions we should dump for this
-        // task, which we do by asking for each dump region by index.  Note
-        // that get_task_dump_region is O(#regions) -- which makes this loop
-        // quadratic: O(#regions * #dumpable).  Fortunately, these numbers are
-        // very small: the number of regions is generally 3 or less (and -- as
-        // of this writing -- tops out at 7), and the number of dumpable
-        // regions is generally just one (two when including the task TCB, but
-        // that's constant time to extract).  So this isn't as bad as it
-        // potentially looks (and boils down to two iterations over all
-        // regions in a task) -- but could become so if these numbers become
-        // larger.
-        //
-        let Some(region) = kipc::get_task_dump_region_raw(task, ndx) else {
-            break;
-        };
+    // Helper function to add a region to the dump
+    let add_dump_region = |region: abi::TaskDumpRegion| {
+        // Skip regions which are in the space used for raw dump data
         if in_dump_area(region.base, region.size) {
-            continue;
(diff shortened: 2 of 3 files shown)


task/jefe/src/dump.rs
```diff
@@ -323,24 +323,50 @@ pub fn dump_task_region(
...
+        // `mem.start` as we find overlaps within the task dump regions. If
+        // `mem` becomes empty, then we know that it is valid (because the
+        // entire `mem` region has overlapped with task dump regions).
+        //
+        // Note: we also implicitly trust that kipc gives us regions which are
+        // in sorted order by base address.
+        let mut mem = start..end;
+        for ndx in 1..=usize::MAX {
+            // This is Accidentally Quadratic; see the note in `dump_task`
+            let Some(region) = kipc::get_task_dump_region(task, ndx) else {
```
jamesmunns, line 354: nit/out of scope refactor suggestion: It feels like we probably could stand to have a helper for "iterate over dump regions", to avoid manual calls like this. It would be neat to be able to do:

```rust
for region in kipc::iter_task_dump_regions(task).skip(1) {
  // ...
}
```

task/jefe/src/dump.rs
```diff
@@ -323,24 +323,50 @@ pub fn dump_task_region(
...
+                mem.start = region.end.min(mem.end);
+                if mem.start == mem.end {
+                    okay = true;
+                    break;
+                }
+            } else if region.start > mem.start {
+                // If we are beyond the start of our `mem` region, then there
+                // are no more overlaps and we can bail out immediately.
+                break;
+            }
```
jamesmunns, line 380: Misc perf nit: it might be worth tracking if we've STARTED matching on items, and bail here if `started && region.start < mem.start`, because then we've found non-continugous overlaps

Assuming:

* region(1): 0x1000..0x2000
* region(2): 0x3000..0x4000

And the user asked to dump `0x1000..0x3800`, currently I think they would never get `okay = true` because we'd take `0x1000..0x2000`, leaving `0x2000..0x3800`, but then `0x3000..0x4000` doesn't `contains` `0x2000`, but then we keep iterating over all remaining regions.

This could probably also be:

```rust
} else if (region.start != start) && region.start < mem.start {
    break;
}
```

task/jefe/src/dump.rs
```diff
@@ -323,24 +323,50 @@ pub fn dump_task_region(
...
+        // Otherwise, iterate over task regions.   We will start with `mem`
+        // representing the full memory range to be dumped, then adjust
+        // `mem.start` as we find overlaps within the task dump regions. If
+        // `mem` becomes empty, then we know that it is valid (because the
+        // entire `mem` region has overlapped with task dump regions).
+        //
+        // Note: we also implicitly trust that kipc gives us regions which are
+        // in sorted order by base address.
+        let mut mem = start..end;
+        for ndx in 1..=usize::MAX {
```
jamesmunns: Just checking, we DO allow dump requests that span multiple task ranges, but NOT one that spans the kernel and one or more task ranges? If this isn't for a specific reason: why do we need the first `if` block at all, and could this just be covered by `for ndx in 0..=usize::MAX`?

jamesmunns (approved):
Some notes, but I also admit to not fully understanding what makes this change somewhat spicy :)

task/jefe/src/dump.rs
jamesmunns: > it's got special behavior for the 0th region, which may not be sorted*

Oh, I missed that in the PR description, let me see again if I missed that in the docs, otherwise this might be worth putting there.

jamesmunns: Oh, my eyes glossed over the *two* other places you mentioned this. It might be worth mentioning this explicitly here again (for folks like myself that are oblivious outside local context, apparently), but you've definitely covered your bases. :D

mkeeter: I decided to make this more obvious by splitting the `userlib` functions into
- `fn get_task_desc_region(task_id: usize) -> TaskDumpRegion` (infallible, no index)
- `fn get_task_dump_region(task_id: usize, region_id: usize) -> Option<TaskDumpRegion>` (fallible, 0-indexed, sorted)

This means that users don't have to remember that index 0 is special at the raw KIPC layer; the raw KIPC wrapper is renamed to `get_task_dump_region_raw`.

task/jefe/src/dump.rs
mkeeter: Seems reasonable!  N is small here (which is why we don't mind quadratic behavior), but also has a context switch into the kernel, so I support cheap early-exit optimizations.

doc/kipc.adoc
evan-oxide, line 259: What is a "dump region" and how does it relate to an MPU region? How/where are a task's dump regions defined? What's a "dump area"?

doc/kipc.adoc
```diff
@@ -260,6 +260,13 @@ the dump region index.  If the dump region index is equal to or greater
 than the number of dump regions for the specified task, `None` will
 be returned.
 
+Passing a dump region of 0 returns the task's descriptor in kernel memory, and
```
evan-oxide, line 263: I'm confused. It sounds like this is a new feature, and now you need to pass different region indices to `GetTaskDumpRegionRequest` than before. But I don't see any change to the kernel's implementation of `get_task_dump_region`...

evan-oxide (commented):
Some questions before I keep reviewing

doc/kipc.adoc
mkeeter: "MPU region" is any memory region configured by the MPU for a task, which includes RAM, flash, peripherals, etc.

"Dump region" is vaguely defined as "a region that can / should be included in a RAM dump for that task".  In practice, this is (1) the task's descriptor in kernel RAM and (2) [writable, non-device](https://github.com/oxidecomputer/hubris/blob/ba49711e69234a3f3562da5e5c65bede2776fc3d/sys/kern/src/descs.rs#L127-L132) memory regions assigned to that task

doc/kipc.adoc
mkeeter: You're correct that nothing has changed; this is just documenting existing behavior to make it explicit!

task/jefe/src/dump.rs
evan-oxide: It feels confusing and error-prone to have the two `get_task_dump_region` and `get_task_dump_region_raw` functions that are both used by jefe but have `region` index args that are off by 1. Some alternate ideas:
- Have the only two userlib functions be `get_task_dump_region` (which does `GetTaskDumpRegionRequest(region+1)`) and `get_task_desc_region` (which does `GetTaskDumpRegionRequest(0)`). Then userspace has one consistent definition for the region index, though it's still off-by-one from the kernel's definition.
- Split them up into two syscalls: `GetTaskDumpRegionRequest` and `GetTaskDescriptorRegionRequest`. Then both userspace and the kernel would agree that dump region indices start at 0 and descriptor regions are something separate.

Sorry if we're doing the thing where a new person gets confused by code and requests silly refactors, even though it would be perfectly clear to someone who had worked here for more than a month.

mkeeter: That's reasonable – I had kept `get_task_dump_region_raw` public because I didn't want to duplicate code in `dump_task`, but have figured out a way to make that a closure instead (3855b408).  `userlib` now just has the two public functions `get_task_dump_region` and `get_task_desc_region`, with no overlapping functionality (and they both use a private `get_task_dump_region_inner` to actually invoke the syscall).

evan-oxide (approved):

task/jefe/src/dump.rs
```diff
@@ -323,24 +329,54 @@ pub fn dump_task_region(
...
-        };
-
-        if in_dump_area(region.base, region.size) {
-            continue;
-        }
-
-        ringbuf_entry!(Trace::DumpRegion(region));
+    // Get the task descriptor region (in kernel memory)
+    let desc = kipc::get_task_desc_region(task);
+    // Note: we implicitly trust kipc won't give us a region that wraps,
```
cbiffle: smol comment nit: the line of code after this comment includes an overflow check, so it doesn't actually trust that the region won't wrap -- it's checking and panicking.

cbiffle (approved):
This looks generally good, I liked James's suggestions but use your judgment.

task/jefe/src/dump.rs
mkeeter: This was a pre-existing comment, but I tweaked it slightly ("trust" here is used to justify doing an unchecked add).

PR #2488, opened 2026-04-23
Add sequencer timeout ereports

mkeeter:
This PR adds two ereport flavors for Cosmo sequencer issues:

- `hw.seq.regs` is a raw dump of the registers, along with a reason why it was produced (MAPO, timeout, or "suddenly we're not in A0")
- `hw.seq.timeout.*` is a family of ereports representing timeouts at each stage of sequencing, along with a diagnosis.  It includes an `Option<u64>` that refers to a `hw.seq.regs` ENA, so that we can also look at raw register values (if the diagnosis is not helpful).

For failure modes other than timeout, this PR doesn't send any additional ereports (just the `hw.seq.regs` value).

Consider this a starting point for brainstorming how we want to organize this data (e.g. one alternative is sending the high-level ereport first, then have the `hw.seq.regs` refer to _it_).

### DIFF
--- a/drv/cosmo-seq-server/src/diagnose.rs
+++ b/drv/cosmo-seq-server/src/diagnose.rs
@@ -102,7 +102,9 @@ pub(crate) struct RegisterDump {
     status: StatusView,
 }
 
+/// Raw registers to be sent as an ereport
 #[derive(Copy, Clone, PartialEq, microcbor::Encode)]
+#[ereport(class = "hw.seq.regs", version = 0)]
 pub(crate) struct RawRegisterDump {
     seq_api_status: u32,
     seq_raw_status: u32,
@@ -115,6 +117,8 @@ pub(crate) struct RawRegisterDump {
     rail_pgs_max_hold: u32,
     sp5_readbacks: u32,
     status: u32,
+
+    reason: DiagnoseReason,
 }
 
 #[derive(Copy, Clone, PartialEq, counters::Count)]
@@ -143,6 +147,7 @@ pub(crate) enum RailIssue {
 #[ereport(class = "hw.seq.timeout.group_a", version = 0)]
 pub(crate) struct GroupATimeoutEreport {
     err: WhyWaitingForGroupA,
+    regs_ena: Option<u64>,
 }
 
 #[derive(Copy, Clone, PartialEq, counters::Count, microcbor::Encode)]
@@ -155,6 +160,7 @@ pub(crate) enum WhyWaitingForGroupA {
 #[ereport(class = "hw.seq.timeout.slp_checkpoint", version = 0)]
 pub(crate) struct SlpCheckpointTimeoutEreport {
     err: WhyWaitingForSlpCheckpoint,
+    regs_ena: Option<u64>,
 }
 
 #[derive(Copy, Clone, PartialEq, counters::Count, microcbor::Encode)]
@@ -171,6 +177,7 @@ pub(crate) enum WhyWaitingForSlpCheckpoint {
 #[ereport(class = "hw.seq.timeout.group_b", version = 0)]
 pub(crate) struct GroupBTimeoutEreport {
     err: WhyWaitingForGroupB,
+    regs_ena: Option<u64>,
 }
 
 #[derive(Copy, Clone, PartialEq, counters::Count, microcbor::Encode)]
@@ -183,6 +190,7 @@ pub(crate) enum WhyWaitingForGroupB {
 #[ereport(class = "hw.seq.timeout.group_c", version = 0)]
 pub(crate) struct GroupCTimeoutEreport {
     err: WhyWaitingForGroupC,
+    regs_ena: Option<u64>,
 }
 
 #[derive(Copy, Clone, PartialEq, counters::Count, microcbor::Encode)]
@@ -196,6 +204,7 @@ pub(crate) enum WhyWaitingForGroupC {
 #[ereport(class = "hw.seq.timeout.power_ok", version = 0)]
 pub(crate) struct PowerOkTimeoutEreport {
     err: WhyWaitingForPowerOk,
+    regs_ena: Option<u64>,
 }
 
 #[derive(Copy, Clone, PartialEq, counters::Count, microcbor::Encode)]
@@ -210,6 +219,7 @@ pub(crate) enum WhyWaitingForPowerOk {
 #[ereport(class = "hw.seq.timeout.reset_l_release", version = 0)]
 pub(crate) struct ResetLReleaseTimeoutEreport {
     err: WhyWaitForResetLRelease,
+    regs_ena: Option<u64>,
 }
 
 #[derive(Copy, Clone, PartialEq, counters::Count, microcbor::Encode)]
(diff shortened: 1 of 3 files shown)


drv/cosmo-seq-server/src/diagnose.rs
```diff
@@ -102,6 +102,25 @@ pub(crate) struct RegisterDump {
     status: StatusView,
 }
 
+/// Raw registers to be sent as an ereport
+#[derive(Copy, Clone, PartialEq, microcbor::Encode)]
+#[ereport(class = "hw.seq.regs", version = 0)]
+pub(crate) struct RawRegisterDump {
+    seq_api_status: u32,
+    seq_raw_status: u32,
```
hawkw, line 110: take it or leave it: i think we could scrape off a couple bytes of ereport if this was:
```rust
seq_status: SeqStatusRegs

// ...

struct SeqStatusRegs {
   api: u32,
   raw: u32,
}
```
since it trades a couple bytes of nested K/V pairs for 11 bytes of repeating `seq_` and `_status` for both fields?

drv/cosmo-seq-server/src/diagnose.rs
```diff
@@ -102,6 +102,25 @@ pub(crate) struct RegisterDump {
...
+    early_power_rdbks: u32,
+    ifr: u32,
+    debug_enables: u32,
```
hawkw, line 113: nit, take it or leave it: could save 2 bytes if this was 
```suggestion
    dbg_enables: u32,
```

drv/cosmo-seq-server/src/diagnose.rs
```diff
@@ -102,6 +102,25 @@ pub(crate) struct RegisterDump {
...
+    power_ctrl: u32,
+    rail_enables: u32,
+    rail_pgs: u32,
+    rail_pgs_max_hold: u32,
```
hawkw, line 117: could save a few bytes if this was like
```rust
    rails: RailRegs,
    // ...
}
    
// ...

struct RailRegs {
    enables: u32,
    pgs: u32,
    pgs_max_hold: u32,
}
```

drv/cosmo-seq-server/src/diagnose.rs
```diff
@@ -102,6 +102,25 @@ pub(crate) struct RegisterDump {
...
+    sp5_readbacks: u32,
+    status: u32,
+
+    reason: DiagnoseReason,
```
hawkw, line 121: part of me wants to say that these should be part of the ereport's class, but...i get why they're not. hm. 🤔

mkeeter:
@hawkw After a little pondering, I'm leaning towards _not_ saving bytes in the `RawRegisterDump` ereport.  Right now, it's a flat array of values with names that match registers in the FPGA; adding fictional hierarchy or renaming registers would make it less obvious.

drv/cosmo-seq-server/src/diagnose.rs
```diff
@@ -102,6 +102,25 @@ pub(crate) struct RegisterDump {
     status: StatusView,
 }
 
+/// Raw registers to be sent as an ereport
+#[derive(Copy, Clone, PartialEq, microcbor::Encode)]
+#[ereport(class = "hw.seq.regs", version = 0)]
```
rmustacc, line 107: This doesn't inherently make sense to me as an ereport class. The class is supposed to be about an event that occurs. A snapshot of registers isn't an event that occurs. This might make sense as something to do, but it feels very much in a different spirit of the illumos FMA design.

drv/cosmo-seq-server/src/diagnose.rs
```diff
@@ -102,6 +102,25 @@ pub(crate) struct RegisterDump {
...
+pub(crate) struct RawRegisterDump {
```
rmustacc, line 108: What is the anticipated stability of this structure? The way that you're phrasing it it seems like this is going to basically be an uncommitted interface. How does the control plane know how to consume and interpret this data? How do we evolve it with changes in the sequence and Hubris code?

drv/cosmo-seq-server/src/diagnose.rs
rmustacc: The reason that this feels weird is because there really are two different things that would happen in FMA traditionally. We would issue an ereport that is about a specific timeout occurrence as one ereport. Then we would issue a **fault** class event that is the diagnosis. Right now hubris is combining the two into one class which is why this feels weird. I think.

Looking more at this, I think another thing that feels a bit off is that it doesn't tell us what the actual root cause or faulty component is, it tells us where we stopped and what the error is, but not necessarily what we should replace or consider the faulty component. Now because almost all of this has a FRU of the mainboad, maybe that makes it less meaningful.

drv/cosmo-seq-server/src/diagnose.rs
```diff
@@ -117,20 +136,34 @@ pub(crate) struct RailStatus {
...
+#[derive(Copy, Clone, PartialEq, counters::Count, microcbor::Encode)]
 pub(crate) enum RailIssue {
     NotEnabled,
     NotPowerGood,
     PowerGoodIntermittent,
 }
 
-#[derive(Copy, Clone, PartialEq, counters::Count)]
+#[derive(Copy, Clone, PartialEq, microcbor::Encode)]
+#[ereport(class = "hw.seq.timeout.group_a", version = 0)]
```
rmustacc, line 147: I feel like all of these variants really want to be a single ereport because that's the actual fact that occurred: we had a timeout. The `Why` part should be a committed enumeration that we are including in that payload.

Notably, the `hw.seq.timeout` as a broad class probably needs to be thought through a bit more. In particular, there are numerous things we sequence across different boards including the T6, Tofino 2, eventually the Versal, etc.

Given the way we've structured the existing CPU ereports, I don't know that a sequencer class makes sense, but we also aren't following the illumos FM structure where every part of the hierarchy adds more payload, so I'm not sure how to come up with better suggestions, but given that this is inherently specific to the SP5 sequencer, a couple bad options are `hw.seq.sp5.timeout`, `hw.cpu.sp5.timeout`, etc.

If we fail to sequence, are timeouts the only way that can happen? Are there other ways that it can fail? I guess put differently how much is the `timeout` an artifact of the design versus representative of a different failure? Are there any other ways it can fail while sequencing other than MAPO?

rmustacc (commented):
Thanks for starting this @mkeeter. I've left a few notes in here focused around the design of the ereports in question. To summarize a few of the notes and it may help to talk live:

* I don't think we should have a separate `hw.seq.regs`.
* I think we need to think a bit more about payload stability and how the control plane will use it.
* I think we want to consolidate the different existing ereports into one for all SP5 sequencing timeouts.
* I think we probably want to think a bit more about class naming and how this fits into the broader picture of sequencer failures.

drv/cosmo-seq-server/src/diagnose.rs
hawkw: I think putting this stuff under the `hw.cpu.sp5` hierarchy feels right to me. I do think we also want to distinguish between failures to sequence that occurs while the system is trying to come up, and some of the other faults we can detect that send a running system _back_ to A2...`timeout` does feel like a bit of an implementation detail, but I haven't thought of a better way to express "we were trying to sequence and couldn't get to A0"...

drv/cosmo-seq-server/src/diagnose.rs
hawkw: I agree that the register dump feels weird and should probably be a blackbox-type thing. We probably want to be saving the register snapshot when the failure occurs, in case the FPGA regs change between when we saw a sequencing failure and when the control plane actually came around to ask for them.

I'm also not sure how we're going to solve the control plane knowing how to decode those registers yet. Since they're coming from the FPGA, they can potentially change a _lot_ across versions...

PR #2490, opened 2026-04-27
Remove power sensor from BMR491

### LINKED ISSUE #1581: BMR491 reports NaN for power
While investigating https://github.com/oxidecomputer/hardware-gimlet/issues/1988 , we noticed that the BMR491 power sensor (`112  0x70 power          4  F   - 0x67 bmr491        V12_SYS_A2`) is reading `NaN`.  That seems weird!

mkeeter:
Fixes #1581, which is tagged with R20 for some reason 🤷🏻‍♂️ 

Two follow-up questions:

- Will removing this sensor make anything in the control plane mad?  cc @hawkw 
- We no longer have a power sensor ID to send in `host-sp-comms` inventory data, but we can't change the `InventoryData::Bmr491` shape for compatibility reasons.  I'm now sending `u32::MAX` as a "no such sensor" marker, which I _think_ should be fine (`bmr_pout` is never used in `stlouis`), but cc @citrus-it to double-check

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -1150,7 +1150,7 @@ device = "bmr491"
 name = "IBC"
 description = "Intermediate bus converter"
 power = { rails = [ "V12_SYS_A2" ] }
-sensors = { temperature = 1, power = 1, voltage = 1, current = 1 }
+sensors = { temperature = 1, voltage = 1, current = 1 }
 refdes = "U80"
 
 [[config.i2c.devices]]
--- a/app/gimlet/base.toml
+++ b/app/gimlet/base.toml
@@ -1069,7 +1069,7 @@ device = "bmr491"
 name = "IBC"
 description = "Intermediate bus converter"
 power = { rails = [ "V12_SYS_A2" ] }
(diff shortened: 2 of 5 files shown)


citrus-it:
> * We no longer have a power sensor ID to send in `host-sp-comms` inventory data, but we can't change the `InventoryData::Bmr491` shape for compatibility reasons.  I'm now sending `u32::MAX` as a "no such sensor" marker, which I _think_ should be fine (`bmr_pout` is never used in `stlouis`), but cc @citrus-it to double-check

That should be fine. The consumer of these will be sled-agent at some point but they aren't used today. Defining `u32::MAX` for this makes sense.

hawkw:
> Will removing this sensor make anything in the control plane mad? cc @hawkw

It shouldn't, we will just stop putting those metrics in Oximeter. MGS' metric collection code enumerates devices from the SP on every pass so if SPs stop reporting a measurement, nothing will break.

hawkw (approved):

hawkw (approved):
seems fine to me

Issue #2409, opened 2026-03-03
Basic Observer app.toml


rmustacc:
This covers creating a basic app.toml for the new observer PSC.

jamesmunns:
Please bring along @jamesmunns and/or @evan-oxide!

Aaron-Hartwig:
@sdonnan will be the EE at EMY for this testing

mkeeter:
This is happening in the [`observer-app` branch](https://github.com/oxidecomputer/hubris/tree/observer-app); Evan is going to take over from here.

evan-oxide:
https://github.com/oxidecomputer/hubris/pull/2493 is merged, with a bunch of todo comments. Features like pmbus will be implemented in later PRs.

PR #2493, opened 2026-04-30
Support for observer-a

evan-oxide:
This creates a draft `observer` app in preparation for bringup. Issue: #2409


app/observer/base.toml
```diff
@@ -0,0 +1,622 @@
...
+
+[[config.i2c.devices]]
+bus = "local"
+address = 0b1010_000
+device = "at24csw080"
+name = "local_vpd"
+description = "FRU ID EEPROM"
+refdes = "U27"
+
+# TODO is this the same part number as for the PSC?
```
mkeeter: After checking with Ian, it isn't the same part number – Observer is using MWOCP67-5500-B-RM.

We should delete all of the PSU stuff from the manifest for now, and leave a comment to the effect of
```toml
# TODO add rectifiers (once we have their SMBus spec)
```

task/power/src/bsp/observer_a.rs
```diff
@@ -0,0 +1,62 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::{
+    PowerControllerConfig, PowerState,
+    i2c_config::{self, sensors},
+};
+
+pub(crate) const CONTROLLER_CONFIG_LEN: usize = 12;
+pub(crate) static CONTROLLER_CONFIG: [PowerControllerConfig;
```
mkeeter, line 9: Similarly, let's just make this an empty array for now (if that works).

If that doesn't work, then we could remove the power task from the manifest until we get the new MWOCP67 implemented.

task/sensor-polling/src/main.rs
```diff
@@ -134,7 +134,11 @@ fn main() -> ! {
 
 use i2c_config::{devices, sensors};
 
-#[cfg(any(target_board = "psc-b", target_board = "psc-c"))]
+#[cfg(any(
+    target_board = "psc-b",
+    target_board = "psc-c",
+    target_board = "observer-a"
+))]
 static SENSORS: [TemperatureSensor; 6] = [
```
mkeeter, line 141: Let's make a new empty array for the `observer-a`, with a comment to fill it in once we know how to talk to the MWOCP67

task/net/src/bsp/observer_a.rs
```diff
@@ -0,0 +1,149 @@
...
+        let spi = bsp_support::claim_spi(sys);
+        let ksz8463_dev = spi.device(drv_spi_api::devices::KSZ8463);
+        let bsp = mgmt::Config {
+            // A2_EN is now controlled by Ignition
+            power_en: None,
+            slow_power_en: false,
+            power_good: &PG_PINS,
+
+            ksz8463: Ksz8463::new(ksz8463_dev),
+            ksz8463_nrst: Port::C.pin(2), // SP_TO_VSC8562_RESET_L_V3P3
```
mkeeter:
```suggestion
            ksz8463_nrst: Port::E.pin(1), // SP_TO_KSZ8463_RESET_L
```

isobering (approved):

app/observer/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+    p.GPIOG.moder.modify(|_, w| w
+        .moder0().input()
+        .moder1().input()
+        .moder2().input());
+    // Enable the pullups.
+    #[rustfmt::skip]
+    p.GPIOG.pupdr.modify(|_, w| w
+        .pupdr0().pull_up()
+        .pupdr1().pull_up()
+        .pupdr2().pull_up());
```
Aaron-Hartwig: @mkeeter do we still need to do this now that we have external pull ups/downs on the board as appropriate? For context Evan, I think this may be vestigial from early Gimlet revs where we had pull-downs in place, but no option for an external pull-up. So we would DNI the pull-down and allow the internal pull-ups to charge the pin.

Image

drv/psc-seq-server/src/main.rs
evan-oxide, line 285: These are wrong - are SP_TO_POWER_SHELF_PSUx_ON_L the right signals to use?

app/observer/base.toml
evan-oxide: That breaks `PSU_PMBUS_DEVS` in psc-seq-server/src/main.rs and I don't see an obvious clean way to fix it.

drv/psc-seq-server/src/main.rs
isobering: Yep, the enable signals are PSU[0:7]_ON_L:

image

mkeeter:
@evan-oxide can you also add this to the set of boards built in CI?

app/observer/src/main.rs
mkeeter: Yeah, I think we can skip this part of pin configuration.  The "how long to wait" is copy-pasted from board to board, so it could also be revisited (but I'm also fine leaving it be).

evan-oxide: Can we remove the wait entirely? It doesn't seem necessary if there are external pullups/downs. Maybe it could matter if the pin was previously configured as an output and driven in the opposite direction as the pullup...

evan-oxide: I removed the wait and the pullups.

app/observer/base.toml
evan-oxide: It's not exactly clean, but I ripped all the pmbus stuff out of the `psc-seq-server` so that it compiles without the rectifier config in base.toml. That commit breaks the original PSC and should obviously not be merged, but hopefully it will let us test other features of `psc-seq-server` without using pmbus.

labbott (commented):
If the EEs are satisfied with initial bring up I vote to merge so we aren't carrying a branch

app/observer/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+fn system_init() {
+    let cp = cortex_m::Peripherals::take().unwrap();
+    let p = device::Peripherals::take().unwrap();
+
+    // We want to measure PG0-2 to determine if we're running on the correct
+    // board.  On rev A, these pins are left floating; on later revisions, they
+    // are pulled either high or low.
+    //
+    // This code matches that in gimlet/src/main.rs; see detailed comments over
+    // there about how times were calculated.
```
mkeeter: Most of this comment is outdated now!

drv/psc-seq-server/src/main.rs
```diff
@@ -317,16 +320,27 @@ const PSU_PWR_OK_NOTIF: [u32; PSU_COUNT] = [
...
-/// In order to get the PMBus devices by PSU index, we need a little lookup table guy.
-const PSU_PMBUS_DEVS: [fn(TaskId) -> (I2cDevice, u8); PSU_COUNT] = [
-    i2c_config::pmbus::v54_psu0,
-    i2c_config::pmbus::v54_psu1,
-    i2c_config::pmbus::v54_psu2,
-    i2c_config::pmbus::v54_psu3,
-    i2c_config::pmbus::v54_psu4,
-    i2c_config::pmbus::v54_psu5,
+// The Observer numbers its notification bits starting at 0 instead of 1
+#[cfg(target_board = "observer-a")]
```
mkeeter: There are a bunch of changes in this file which would break the PSC, so we can't merge as-is.  Since we're using a new rectifier, it will probably be different enough to warrant a new server – want to make an `observer-seq-server` to float these changes while waiting for the SMBus specs?

app/observer/src/main.rs
evan-oxide: Removed.

drv/psc-seq-server/src/main.rs
evan-oxide: I created `observer-seq-server`. Let me know if there's anything obviously outdated in there to remove.

app/observer/dev.toml
```diff
@@ -0,0 +1,21 @@
+# Configuration fragment for -dev images
+
+[tasks.jefe.config.allowed-callers]
+request_reset = ["udprpc"]
+
```
mkeeter, line 5: Let's also add `net` support to the `hiffy` task in the `dev` image, e.g.

```toml
[tasks.hiffy]
features = ["net", "vlan"]
notifications = ["socket"]

[config.net.sockets.hiffy]
kind = "udp"
owner = {name = "hiffy", notification = "socket"}
port = 11115
tx = { packets = 3, bytes = 32 }
rx = { packets = 1, bytes = 4096 }
```

drv/observer-seq-server/src/main.rs
```diff
@@ -0,0 +1,1149 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
```
mkeeter, line 4: Add a big comment here, something like
```rust
// THIS CODE IS COPIED FROM PSC-SEQ_SERVER AND NOT YET FULLY IMPLEMENTED
```

drv/psc-seq-server/build.rs
```diff
@@ -4,6 +4,7 @@
 
 fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     build_util::build_notifications()?;
+    build_util::expose_target_board();
```
mkeeter: Undo this change, since we're no longer using `psc-seq-server` in the Observer app?

task/sensor-polling/src/main.rs
```diff
@@ -132,9 +132,10 @@ fn main() -> ! {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+#[cfg(any(target_board = "psc-b", target_board = "psc-c",))]
 use i2c_config::{devices, sensors};
 
-#[cfg(any(target_board = "psc-b", target_board = "psc-c"))]
+#[cfg(any(target_board = "psc-b", target_board = "psc-c",))]
```
mkeeter:
```suggestion
#[cfg(any(target_board = "psc-b", target_board = "psc-c"))]
```

task/sensor-polling/src/main.rs
```diff
@@ -132,9 +132,10 @@ fn main() -> ! {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+#[cfg(any(target_board = "psc-b", target_board = "psc-c",))]
```
mkeeter:
```suggestion
// The `observer` app uses `sensor-polling` but has no I2C devices yet,
// so this import is conditional to avoid warnings
#[cfg(any(target_board = "psc-b", target_board = "psc-c"))]
```

mkeeter (approved):

Issue #2498, opened 2026-05-05
Hubris' CI is failing because a node dependency has been deprecated


hawkw:
It looks like our CI build on `master` is failing because, apparently, the `swatinem/rust-cache@v2` GitHub action, which runs _after_ we have successfully built Hubris, depends on a Node.js dependency which has been deprecated, [as seen here](https://github.com/oxidecomputer/hubris/actions/runs/25405354453/job/74514872907):

```
Run Swatinem/rust-cache@v2
(node:7288) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

```

It looks like https://github.com/Swatinem/rust-cache/pull/294 has allegedly fixed this (but [perhaps not?](https://github.com/Swatinem/rust-cache/issues/331)), so hopefully updating our dependency from @v2 to @v2.9.1 _may_ fix it?

hawkw:
Welp. @mkeeter has informed me that this may be some kind of GitHub-related flake, and indeed, upon restarting the failed build, it has worked: https://github.com/oxidecomputer/hubris/actions/runs/25405354453

So that's cool. I guess I'll close this one out.

PR #2499, opened 2026-05-05
empty commit to make sure it's broken

hawkw:


hawkw:
This commit was intended as the first step in fixing what I thought was a sticky CI failure (#2498), which turned out to be a flake. It turned out that it was [not actually broken](https://github.com/oxidecomputer/hubris/actions/runs/25406345598/job/74518065525) and the empty commit to make sure it was broken made it through CI happily. So...we can close this.

PR #2501, opened 2026-05-07
Set VLAN FID in configuration

mkeeter:
The MAC tables in the VSC7448 contain a map from `(source mac, filtering id)` tuples to port.  The filtering id field ("FID") allows for VLAN-aware learning.

Unfortunately, you have to explicitly set a flag to copy the classified VID to the FID.

If you're in a situation where you see the same MAC address on multiple ports (which are on different VLANs), properly setting the FID is load-bearing; otherwise, that MAC address will randomly wander between those ports, and packets directed to that MAC address have unpredictable deliverability.

Fixes https://github.com/oxidecomputer/mfg/issues/224


labbott (approved):

mkeeter:
Tested on `london`, with no behavior change (which is expected).  This also fixes issues on `silver` (with Reversos populated), so let's bring it in.

PR #2502, opened 2026-05-07
move front-io related sequencing logic into the server

Aaron-Hartwig:
Act III in the effort to decouple front-io related logic from board specific sequencers involves moving all the actual sequencing logic (e.g. FPGA bitstream loading) into the front-io-server. This diff is very noisy as a result of that, but the core logic in play here is not touched. The last piece to do here will be move all the transceivers and LED stuff over, but that warrants its own commit I think. I am bringing Medusa along for the ride for now even though it is untested. After the decoupling is complete I will have a follow on PR which was used to actually bring up that board.


drv/front-io-server/build.rs
```diff
@@ -6,6 +6,12 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     build_util::expose_target_board();
     build_util::build_notifications()?;
 
+    let disposition = build_i2c::Disposition::Devices;
+    if let Err(e) = build_i2c::codegen(disposition) {
+        println!("code generation failed: {e}");
+        std::process::exit(1);
+    }
+
```
mkeeter:
```suggestion
    build_i2c::codegen(build_i2c::Disposition::Devices)?;
```

drv/sidecar-seq-server/src/main.rs
```diff
@@ -144,12 +116,10 @@ struct ServerImpl {
     mainboard_controller: MainboardController,
     clock_generator: ClockGenerator,
     tofino: Tofino,
-    front_io_hsc: HotSwapController,
```
mkeeter, line 147: You can now delete the `struct HotSwapController` entirely

drv/front-io-server/src/main.rs
```diff
@@ -81,20 +388,133 @@ impl NotificationHandler for ServerImpl {
...
 #[unsafe(export_name = "main")]
 fn main() -> ! {
-    let mut server = ServerImpl {
-        seq: Sequencer::from(SEQUENCER.get_task_id()),
-    };
+    let mut server = ServerImpl::default();
+
+    // TODO: there will be more timers when I2C gets moved into this server
+    #[derive(Copy, Clone, Enum)]
+    #[allow(clippy::upper_case_acronyms)]
```
mkeeter: Not needed?
```suggestion
```

drv/front-io-server/src/main.rs
```diff
@@ -7,54 +7,335 @@
...
+            PhyOscState::Bad => {
+                // The PHY was attempted to be initialized but its
+                // oscillator was deemed not functional. Unfortunately
+                // the only course of action is to power cycle the
+                // entire front IO board, so do so now.
+                ringbuf_entry!(Trace::PowerEnabled(false));
+                self.bsp.set_power_enable(false)?;
+                // After removing power to the board we must reset its
+                // server
+                self.do_server_reset();
```
mkeeter, line 285: Can you say more about why this is necessary?

drv/front-io-server/src/main.rs
```diff
@@ -63,10 +344,36 @@ impl idl::InOrderFrontIOImpl for ServerImpl {
...
+                ringbuf_entry!(if good {
+                    Trace::PhyOscGood
+                } else {
+                    Trace::PhyOscBad
+                });
+
+                self.phy_smi
+                    .set_osc_good(good)
+                    .map_err(FrontIOError::from)
+                    .map_err(RequestError::from)
```
mkeeter, line 370: We're basically using the FPGA as external memory that persists through task restarts?

(I know this hasn't changed, just clarifying my understanding)

drv/front-io-server/src/main.rs
```diff
@@ -81,20 +388,133 @@ impl NotificationHandler for ServerImpl {
...
+                            if server.is_board_present_and_powered() {
+                                server.board_status = FrontIOStatus::FpgaInit;
+                            } else {
+                                server.board_status = FrontIOStatus::NotPresent;
+                            }
+                        }
+
+                        // Once we know there is a board present, configure its
+                        // FPGAs and wait for its PHY oscillator to be functional.
+                        FrontIOStatus::FpgaInit => {
```
mkeeter: Should this be called `FrontIOStatus::ReadyForFpgaInit`?

drv/front-io-server/src/main.rs
```diff
@@ -81,20 +388,133 @@ impl NotificationHandler for ServerImpl {
...
+    let now = sys_get_timer().now;
+    multitimer.set_timer(
+        Timers::Seq,
+        now,
+        Some(Repeat::AfterDeadline(SEQ_INTERVAL)),
+    );
 
     let mut buffer = [0; idl::INCOMING_SIZE];
     loop {
+        multitimer.poll_now();
```
mkeeter, line 442: The multitimer notification bit should be folded into `NotificationHandler` / `handle_notification`; otherwise, we only respond to Idol messages at 100ms intervals.

drv/sidecar-seq-server/src/main.rs
Aaron-Hartwig: Thanks for catching this! I can get rid of `drv/sidecar-mainboard-controller/src/front_io.rs` entirely. I'd missed cleaning out that particular corner.

drv/front-io-server/src/main.rs
Aaron-Hartwig: You're right. I was referencing my original decoupling branch and then paring things back to fit the set of changes I need. In there, the `Timers` enum contains an `I2C` variant which lead to me adding that clippy allow. I'll remove it in the meantime.

drv/front-io-server/src/main.rs
Aaron-Hartwig: Since this task handles sequencing and maintains state related to components on the front-io board (FPGA configuration, LED driver state, etc), clearing any state and starting over is a clean way to start from scratch when it is powered back up. Do you take issue with that as a strategy, or just want more context in the comment above it?

drv/front-io-server/src/main.rs
Aaron-Hartwig: Correct! I'll adjust the comment at the top of this IPC to make that clearer.

drv/front-io-server/src/main.rs
Aaron-Hartwig: Fair. I also adjusted `FrontIOStatus::OscInit` to be `FrontIOStatus::WaitForOscGood` since that is actually what is happening. The oscillator's role in all this is still a bit strange, but I think we could come up with a more straightforward way of managing that in the future.

drv/front-io-server/src/main.rs
Aaron-Hartwig: Chatted briefly via DMs. Matt pointed out that without `notifications::TIMER_MASK` in my `NotificationHandler` that the timer would never be able to fire without there being idol activity to my server (which there is as everyone checks to wait and see if we've sequenced, effectively sequencing us).

drv/front-io-server/src/main.rs
mkeeter: Just wanted more context, since this was new in the PR (as far as I can tell by looking at where code moved from).  That seems reasonable, as long as the server's default state matches what we get when power-cycling...

drv/front-io-server/src/main.rs
```diff
@@ -7,94 +7,517 @@
...
+// notifications stubbed out to support multitimer firing
 impl NotificationHandler for ServerImpl {
     fn current_notification_mask(&self) -> u32 {
-        0
+        notifications::TIMER_MASK
     }
 
-    fn handle_notification(&mut self, _bits: NotificationBits) {
-        unreachable!()
+    fn handle_notification(&mut self, _bits: NotificationBits) {}
```
mkeeter, line 391:
```suggestion
    fn handle_notification(&mut self, _bits: NotificationBits) {
        // Do nothing here (timer notifications are handled in the main loop)
    }
```

drv/front-io-server/src/main.rs
```diff
@@ -7,94 +7,517 @@
...
+                        // make a judgement about the oscillator.
+                        FrontIOStatus::WaitForOscGood => {
+                            ringbuf_entry!(Trace::SeqStatus(
+                                server.board_status
+                            ));
+                            if server
+                                .phy_smi
+                                .osc_state()
+                                .unwrap_or(PhyOscState::Unknown)
+                                == PhyOscState::Good
```
mkeeter, line 496: Nit: you could also write this as
```suggestion
                                .osc_state()
                                == Ok(PhyOscState::Good)
```

mkeeter:
This all looks pretty good!  Have you tested on Sidecar hardware?  Dunno if you've thought about how to test the bad oscillator case – I'd like to confirm that the PR still handles it correctly, but it's hard to artificially trigger...

drv/front-io-server/src/main.rs
Aaron-Hartwig: It is https://github.com/oxidecomputer/hubris/pull/2502/changes#diff-277fe5cd83d114680b360ae4816921809829fab6f91fea1f9040aad2059f1c60R425

Aaron-Hartwig:
Yeah, this was tested on my Sidecar/Cosmo setup and things worked. My unit does not have a chronically bad oscillator, but `sidecar-b` on `niles` does and is able to correctly do `monorail`'s PHY reset dance (something I'd like to move into the front-io-server at some point...). There were some ringbuf improvement opportunities to give you a clear picture of what was happening, so I'm toying with those now.

Issue #2337, opened 2025-12-31
HostToSp::HostBootFailure needs to be accessible beyond humility


rmustacc:
There's a TODO in handling of various host failures that says to forward to MGS. Whether that, an ereport, or something else, we should make something discoverable as we cannot get access to this in an in-rack context and don't always have the luxury of hoping that we get another lap to confirm the information about what didn't happen.

This also applies to the panic and similar cases.

hawkw:
I opened #2140 for generating an ereport for host panics and boot failures. I think we will also need to add something in the `control-plane-agent` protocol for getting the host panic/boot failure data out though, as the buffers for those are too large to fit in an ereport. 

I think the approach here will have to be one where the ereport _informs_ upstack software about the boot failure/panic, but a separate API is used to actually read the data out. That will probably also make it easier to consume it via `faux-mgs` in bench and test use cases.

hawkw:
This is somewhere on my todo list (and would be happy to bump it up if it's become urgent). But, if anyone else gets around to it before me, please feel free to ping me if you have any questions about the ereport bit!

jamesmunns:
Picking this up, happy if anyone has any specific notes, I'll probably be asking more questions as I figure out the context here.

jamesmunns:
This landed as part of #2504

PR #2503, opened 2026-05-08
Add ereports for host panic/boot failure

### LINKED ISSUE #2140: ereport: host panicked
When the host OS panics, the `host-sp-comms` task should send an ereport to `packrat` about it.

jamesmunns:
This PR adds an `ereport` for a host panic, as well as boot failures. This PR does not currently implement the interface necessary to retrieve the panic message, which I will do in a follow-up commit/PR.

For @rmustacc, I've chosen the ereport class `hw.host.panic` and `hw.host.btfail`. Open to suggestions on this. Also worth noting, we will truncate any Host Panics or Boot Failures that are each larger than `MAX_HOST_FAIL_MESSAGE_LEN`, which is currently 4KiB. I will probably need to figure out how to paginate this for access in a follow up issue, but I wanted to make sure that 4KiB limit wasn't concerning to you (either not enough, or excessive).

For @hawkw, is there a guideline for whether `declare_ereporter` should be done in the relevant task, or in a central place like `lib/ereports`? The current `hw.cpu` ereports are there, but `drv/xxx-seq-server`s do it in their own crate (like this PR). I can move this to `lib/ereports` if that's the preferred approach.

I intend to include the `ttl_ct`/`panic_len` fields in the host comms API, either having the host include it in the "get" request, or just including it in the send, to ensure that we are actually retrieving the panic expected. We might also crosscheck the SP's boot count, to make sure we aren't getting any extra-stale requests. I'll note that in the follow-up PR.

Closes #2140.
Related to #2337.


hawkw:
@jamesmunns:

> For @rmustacc, I've chosen the ereport class `hw.host.panic` and `hw.host.btfail`.

The `hw.*` hierarchy is for ereports related to hardware components (`hw` stands for HardWare). In general we try to abide by the convention that all `hw.*` ereports have a refdes --- I think I mentioned this in the watercooler a bit ago? In this case, I don't think that the host boot fail and panic ereports should be under `hw.*` since they are not indicating an event related to a particular hardware component, but an event related to upstack software (the host OS). IMO, we should probably just define a new top-level `host.*` hierarchy for these ereports, but perhaps Robert will have an opinion.

> For @hawkw, is there a guideline for whether `declare_ereporter` should be done in the relevant task, or in a central place like `lib/ereports`? The current `hw.cpu` ereports are there, but `drv/xxx-seq-server`s do it in their own crate (like this PR). I can move this to `lib/ereports` if that's the preferred approach.

I would only put ereport type definitions in `lib/ereports` when the same ereport type is reported by multiple tasks. The `hw.cpu` ereports are defined in `lib/ereports` because both the Gimlet and Cosmo sequencer tasks will report some of the same ereport types (but with different type parameters in some cases, which you can see if you look at their `declare_ereporter!` invocations). As for the `declare_ereporter!` macro, it should _always_ be invoked in a task, and not in a library crate. That macro is responsible for creating some static allocations for the ereport buffer and ereport debugging counters, so if you invoke it in a lib crate, that will exist in *every* task that depends on that lib. And, it's supposed to be invoked with the complete list of ereport types that _that task_ will want to be able to report. For example, if you look at different tasks that report different sets of ereports, you can see that they invoke `declare_ereporter!` with a different list of ereport message types:

https://github.com/oxidecomputer/hubris/blob/33a6acba64fab05248dc6dbf6ba3df98f00e3d07/drv/gimlet-seq-server/src/main.rs#L1637-L1651

https://github.com/oxidecomputer/hubris/blob/33a6acba64fab05248dc6dbf6ba3df98f00e3d07/drv/psc-seq-server/src/main.rs#L1087-L1094

If the intended use of the macro wasn't clear in its docs, I should probably go improve that!

task/host-sp-comms/src/main.rs
```diff
@@ -2012,3 +2071,40 @@ mod idl {
 }
 
 include!(concat!(env!("OUT_DIR"), "/notifications.rs"));
+
+ereports::declare_ereporter! {
+    struct Ereporter<Ereport> {
+        HostPanic(HostPanic),
+        BootPanic(HostBootFail),
+    }
+}
```
hawkw, line 2096: This looks right!

task/host-sp-comms/src/main.rs
```diff
@@ -2012,3 +2071,40 @@ mod idl {
...
+
+/// An ereport represent a host reported panic
+#[derive(Encode)]
+#[ereport(class = "hw.host.panic", version = 0)]
```
hawkw: Should probably not be in the `hw` hierarchy, as it's a host OS event, not a hardware event:
```suggestion
#[ereport(class = "host.panic", version = 0)]
```

task/host-sp-comms/src/main.rs
```diff
@@ -2012,3 +2071,40 @@ mod idl {
...
+struct HostPanic {
+    /// The total number of panics observed this boot cycle.
+    ///
+    /// This count will wrap, but is guaranteed to never be zero.
+    ttl_ct: u32,
```
hawkw: Hm, I really don't love abbreviating "total" to "ttl" --- to me, `ttl` always means Time To Live, not "total". I might just say `cnt`, or even `n`?

task/host-sp-comms/src/main.rs
```diff
@@ -2012,3 +2071,40 @@ mod idl {
...
+    /// The length, in bytes, of the stored panic message.
+    ///
+    /// This quantity may be less than the amount received, as it is capped
+    /// by the available storage space allocated (`MAX_HOST_FAIL_MESSAGE_LEN`).
+    pan_len: u32,
```
hawkw: how about
```suggestion
    msglen: u32,
```
since we use it for both the panic and boot fail messages?

task/host-sp-comms/src/main.rs
```diff
@@ -2012,3 +2071,40 @@ mod idl {
...
+}
+
+/// An ereport represent a host reported boot failure
+#[derive(Encode)]
+#[ereport(class = "hw.host.btfail", version = 0)]
```
hawkw: similarly, this should be
```suggestion
#[ereport(class = "host.btfail", version = 0)]
```

task/host-sp-comms/src/main.rs
```diff
@@ -2012,3 +2071,40 @@ mod idl {
...
+        HostPanic(HostPanic),
+        BootPanic(HostBootFail),
+    }
+}
+
+/// An ereport represent a host reported panic
+#[derive(Encode)]
+#[ereport(class = "hw.host.panic", version = 0)]
+struct HostPanic {
+    /// The total number of panics observed this boot cycle.
```
hawkw: this comment seems potentially confusing --- when you say "boot cycle", are you referring to a boot cycle of the _SP_, or of the _host system_? we would not expect the host OS to panic multiple times in one boot cycle of the host CPU.

technically, it's neither, it's a count of host panics _since `host-sp-comms` has restarted_, which we would _hope_ it doesn't panic, but ...

task/host-sp-comms/src/main.rs
```diff
@@ -2012,3 +2071,40 @@ mod idl {
...
+    ///
+    /// This count will wrap, but is guaranteed to never be zero.
+    ttl_ct: u32,
+    /// The length, in bytes, of the stored panic message.
+    ///
+    /// This quantity may be less than the amount received, as it is capped
+    /// by the available storage space allocated (`MAX_HOST_FAIL_MESSAGE_LEN`).
+    pan_len: u32,
+}
```
hawkw, line 2136: The host boot fail message also includes a `reason` byte:
https://github.com/oxidecomputer/hubris/blob/f80c52d14a5922af661bdd7f46e32578d31184f1/task/host-sp-comms/src/main.rs#L997

I think we should include this in the ereport as well (and I would prefer it to be under the name `reason` for consistency rather than abbreviated).

hawkw (changes requested):
Some suggestions about the ereport messages and data contained therein. 

I also wonder somewhat if we might want to try to include which BSU (host boot slot) we were booting from in the boot fail (and also panic) messages. See https://github.com/oxidecomputer/hubris/blob/f80c52d14a5922af661bdd7f46e32578d31184f1/task/host-sp-comms/src/main.rs#L954-L966 for an example of getting it from the host-flash task. I'm not sure whether we expect the host to include this state in the panic messages someplace, but I don't think it will...

jamesmunns:
@hawkw should be ready for re-review now! I think I've addressed all comments.

jamesmunns:
@hawkw CI is passing now!

jamesmunns:
Closing this in favor of #2518 which contains these changes as well.

PR #2511, opened 2026-05-13
cosmo_seq: mask RAA229620A Iout overcurrent warnings

### LINKED ISSUE #2510: cosmo sequencer must learn to chill out about RAA229620A SVI3 fast sum overcurrent warnings
As described in https://github.com/oxidecomputer/customer-support/issues/1150, we have seen a whole bunch (i.e. in excess of 25k) of `hw.pwr.pmbus.alert` ereports for RAA229620A output overcurrent warnings emitted by Cosmos at a customer site. I initially was quite shocked and disturbed by these, since the `IOUT_OC_WARN_LIMIT` register, which is the PMBus standard register for configuring the overcurrent warning threshold, is set to a default value of 3000A on the RAA229620A; this is an absurdly high value, and higher than the _fault_ limit by default.

However, @ericaasen informs me that there is also an additional SVI3 `OCP_WARN_THRESH` value on these parts (or at least on their SVI3 rails?), which is set to a value of 297 amps. This is less than the EDC of AMD IRM Group G CPUs, which is 330A. So, these warnings are basically spurious and we shouldn't be constantly screaming about them, at least in systems with Group G parts. 

According to Eric in https://github.com/oxidecomputer/customer-support/issues/1150#issuecomment-4441444948, this threshold is "required[^1] to be set to 90% of the EDC", which I take to mean that we cannot change the threshold value. Instead, I think we should be disabling PMBus alerts for overcurrent warnings on the RAA229620A's SVI3 rails, which ought to suppress this provided that masking out the PMBus warning but also masks warnings from the SVI3 OC warning, which...well, I sure _hope_ it does. If that doesn't work, I guess we can always make the sequencer's PMBus warning interrupt handler check for overcurrent warnings and just decide to do nothing when it sees them, which should at least prevent the ereport spew. I'd really prefer to just tell the RAA229620A to stop asserting `SMBALERT_L` though, since that way we don't spend a bunch of time in the interrupt handler just to figure out that an alert is spurious...

[^1]: editor's note: required by whom?

hawkw:
See #2510 --- evidentially, AMD requires that the SVI3 fast sum overcurrent threshold for VRMs supplying SVI3 power rails to Turin CPUs be set at a threshold of 90% of 90% of the part's EDC amperage. This results in the VRM setting the overcurrent warning bit in `STATUS_IOUT` and tugging on its fault pin to request attention pretty incessantly when an IRM Group G CPU is under heavy (but importantly, safe!) load. AMD suggests that BMC's "mask or ignore" this warning, since the threshold has to be set to a value that results in it being asserted during normal operation due to reasons I am no doubt to provincial to understand. Therefore, this commit makes Hubris mask out the `STATUS_IOUT` bit for overcurrent warnings on the RAA229620A `VDDCR_CPU0` and `VDDCR_CPU1` SVI3 regulators. This felt like the best solution because it (a) shouldn't suppress any fault bits that actually *do* represent something bad, and (b) unlike ignoring this specific warning, doesn't result in Hubris taking a bunch of IRQs that it just ultimately decides to do nothing about.

With help from @ericaasen, we have reproduced the warnings by manually adjusting the threshold to a lower value, as the CPU load testing script I was running wasn't actually able to make a 9755 hit the 297 amp threshold. I've validated that masking the relevant bit in the `SMBALERT_MASK` for the `STATUS_IOUT` register makes the IRQing go away. So that's good.

I don't _love_ the function I wrote for masking this bit: it's specific for setting `SMBALERT_MASK` for `STATUS_IOUT` and not for any other register. I feel like it ought to be possible to write a single function for setting `SMBALERT_MASK` for *any* status register, but the abstractions provided by `pmbus` at time of writing made it difficult to do so generically --- or perhaps I'm holding it wrong! Anyway, I'd really like to get the immediate fix in first and then try to come up with a nicer abstraction for `SMBALERT_MASK` later, since I'd like to stop the ereport spew in R20 if possible.

In addition, I have made some changes to the `initialize_pmbus_alerts` (née `initialize_pmbus_warnings`) method in `Vcore`. Since that method now both sets the input undervoltage warning threshold _and_ masks the output overcurrent warning, I've changed it to not return a `Result` so that it cannot accidentally `?` out early without trying to set _all_ the PMBus configurations it tries to set. We probably want to attempt to set the `SMBALERT_MASK` even if we encountered an amount of I2C Weather that caused setting the UV threshold for one of the VRMs to fail. So now we put the error in the ringbuf and keep going there.

Fixes #2510.


drv/i2c-devices/src/raa229620a.rs
```diff
@@ -143,6 +143,32 @@ impl Raa229620A {
...
+            .write(&[
+                CommandCode::SMBALERT_MASK as u8,
+                CommandCode::STATUS_IOUT as u8,
+                mask.0,
+            ])
+            .map_err(|code| Error::BadWrite {
+                cmd: CommandCode::SMBALERT_MASK as u8,
+                code,
+            })
+    }
```
hawkw, line 174: I really do not particularly care for this method, since it _feels_ like it ought to be possible to come up with a function that takes _any status register_ as input and sets the mask for that status register, in a typed-y way. I think it's probably possible to come up with such a thing generically, but I wanted to at least get this to work first in the interest of expedience.

drv/i2c-devices/src/raa229620a.rs
```diff
@@ -143,6 +143,32 @@ impl Raa229620A {
...
+    pub fn set_status_iout_smbalert_mask(
+        &self,
+        mask: STATUS_IOUT::CommandData,
+    ) -> Result<(), Error> {
+        // Unfortunately, SMBALERT_MASK is kinda hard to use with the
+        // `pmbus_write!` macro family, due to not having its own `CommandData`
+        // type, and because it requires writing both the name of the status
+        // register being masked *and* the value of that register (as the mask).
+        // It's a bit odd. So we just do it "manually" --- luckily, this
+        // register is not paged, at least.
```
mkeeter: Are you sure about that?

Image

drv/i2c-devices/src/raa229620a.rs
mkeeter: I agree with both the sadness and the lack of a better way right now (I suspect a proper fix would require PMBus traits, which don't yet exist).

drv/i2c-devices/src/raa229620a.rs
hawkw: AGH

drv/i2c-devices/src/raa229620a.rs
hawkw: Yeah, I was imagining some kind of `StatusRegister` trait, but didn't want to spend time on it yet.

drv/i2c-devices/src/raa229620a.rs
hawkw: I guess this was just accidentally working 'cause we were already on the right page. I'll fix that.

hawkw: fixed it, my bad: c40c51959e772f0c3878bd97e34fd96fa15b7dc7

drv/cosmo-seq-server/src/main.rs
```diff
@@ -773,7 +773,7 @@ impl ServerImpl {
         // few times, so we should power through any transient I2C messiness,
         // and any I2C errors that occur get logged in the `vcore` module's
         // ringbuf.
-        let _ = self.vcore.initialize_uv_warning();
+        let _ = self.vcore.initialize_pmbus_warnings();
```
jamesmunns: Do we want to yell a bit if this fails? Currently if we early-return from setting the mask (I'm not sure how tenacious `retry_i2c_txn` is, seems like 3x attempts?), we'll still have the overcurrent warnings present, and someone would have to notice that there *isn't* a ringbuf entry for "set the mask".

jamesmunns (approved):
One very small "are you sure", as a wholly uninformed individual.

mkeeter (approved):

drv/cosmo-seq-server/src/main.rs
hawkw: So, as the comment above points out, we will record the I2C errors in the `vcore` ringbuf, and our only real options here are to either "ignore it" or "bail out in a way that stops the entire computer from turning on", which are both not great. However, I think you _are_ correct that the current code will bail out of we can't enable an undervolt warning and _not_ try to mask out the overcurrent warning, which seems a bit sad now that the function is doing multiple things --- we should probably always try to set all the desired bits even if a previous thing fails. I will change that

jamesmunns: Yeah, I was mostly suggesting that we might want to explicitly log if this happens, so someone can figure it out after-the-fact. The *absence* of the ringbuf event seems like something that could be looked for, and I'm definitely not familiar with what the "meta" is for including enough breadcrumbs for debugging stuff like this.

You likely have better opinions here than me! Appreciate you taking a second look.

hawkw: > Yeah, I was mostly suggesting that we might want to explicitly log if this happens, so someone can figure it out after-the-fact.

We already do explicitly log, it's just in the `vcore` module. That's what this comment was trying to explain: https://github.com/oxidecomputer/hubris/blob/1754c7412b252ca3bf984700e790c3e0a8b88d54/drv/cosmo-seq-server/src/main.rs#L768-L775

However, I've changed this to just not return an error at all, in commit https://github.com/oxidecomputer/hubris/pull/2511/changes/b69e211915ac36b576273ce0d3c1f2870c8f9c11. That also changes `initialize_pmbus_alerts` to try to set all the desired config bits even if one step fails, rather than bailing with `?` --- it seemed nicer to try to set every desired config even if we weren't able to do one of them for "i2c reasons". The errors are still always getting recorded in the `vcore` ringbuf.

drv/cosmo-seq-server/src/vcore.rs
```diff
@@ -198,17 +205,63 @@ impl VCore {
...
+            // the default value of the PMBus `IOUT_OC_WARN_LIMIT` register
+            // on the RAA229620A is 3000A, which is *well above* the overcurrent
+            // *fault* threshold.
+            mask.set_output_overcurrent_warning(
+                STATUS_IOUT::OutputOvercurrentWarning::Warning,
+            );
+            mask
+        };
+        let mut all_ok = true;
+        all_ok |=
```
jamesmunns: Should this be `&=`? Right now this will always be true.

drv/cosmo-seq-server/src/vcore.rs
```diff
@@ -198,17 +205,63 @@ impl VCore {
...
+        // Each step will retry internally a few times, so we should power
+        // through any transient I2C messiness, and any I2C errors that occur
+        // get logged in the ringbuf. We also do *not* bail out early if any
+        // other step fails, because we would still like to do every other thing
+        // if possible.
+
         // Set our warn limit
-        retry_i2c_txn(Rail::VddcrCpu0, PmbusCmd::LoadLimit, || {
+        let mut all_ok = true;
+        all_ok |= retry_i2c_txn(Rail::VddcrCpu0, PmbusCmd::LoadLimit, || {
```
jamesmunns: Same as below, &=

drv/cosmo-seq-server/src/vcore.rs
hawkw: WOW I'M STUPID

Issue #2505, opened 2026-05-08
Crash of Supervisor does not lead to system reboot


jamesmunns:
The [hubris reference](https://hubris.oxide.computer/reference/#_what_is_the_supervisor) states:

> If the supervisor task crashes, the system reboots.

The behavior is also referred to in code comments, for example in `sys/kern/src/kipc.rs`:

https://github.com/oxidecomputer/hubris/blob/33a6acba64fab05248dc6dbf6ba3df98f00e3d07/sys/kern/src/kipc.rs#L314-L316

However, there is currently no implementation in the kernel that makes good on this promise. If the supervisor ("jefe") were to panic, for example due to [bad indexing of a slice](https://github.com/oxidecomputer/hubris/pull/2469), the following approximate code path would be taken:

1. In `jefe`'s context:
    1. `userlib::panic_handler` is called, a panic message is prepared
    2. `userlib::sys_panic` is called
    3. `userlib::sys_panic_stub` is called
    4. a `Sysnum::Panic` syscall is made
2. In the `kernel`'s context:
    1. `syscall_entry` is entered
    2. `safe_syscall_entry` is called
    3. `explicit_panic` is called
    4. `force_fault` is called
      * The task (`index` == `0`) is updated as `TaskState::Faulted`
      * We attempt to `Task::post` to the supervisor (because a task faulted!)
      * `TaskState::can_accept_notification` is checked and decided as false, and not awoken
      * Since the supervisor was not awoken, `NextTask::Other` is returned
    5. Normal scheduling resumes

As the supervisor would now be in the "Faulted" state, there is no supervisor to restart itself, nor restart any subsequently failing tasks. This could lead to the system slowly degrading as tasks are halted and not restarted, unless some more fatal error occurs.

The fix for this is likely detecting somewhere in the kernel's context chain that the faulting task is `index` == `0`, and immediately triggering a restart of the system.

There is some potential risk to this approach, if supervisor death was more common than we thought, and restarting of SPs leads to louder failures than expected. However, this is still probably the right thing to do: we should just take care to test this change to shake out and resolve any knock-on impacts it has.

PR #2514, opened 2026-05-14
doc: clarify that the supervisor crashing doesn't reboot yet

cbiffle:
Thanks to @jamesmunns for the feedback on this text still being in the reference manual. There may be other places where we make this assertion that I didn't find.


jamesmunns:
CC https://github.com/oxidecomputer/hubris/issues/2505

jamesmunns:
@cbiffle there's at least one other mention I've found in the kipc source (linked in the issue), I'll let you decide if merging this change is enough to close that issue as well.

Issue #2335, opened 2025-12-30
Need better way to distinguish intended host state from current state


rmustacc:
I was debugging some cases earlier today where some sleds appear to have been rebooting. I appear to have caught a case where the sled was briefly in A2. I got this via `pilot sp status`, which is using faux-mgs to understand the current power state. However, the thing that I didn't know was that this was probably in the process of rebooting due to a host boot / panic loop. We need some way to better distinguish what is the current power state from what is happening or transient. In general when I see us in A2, I assume we have come to rest in A2 not that we're in the process of making a transition. Perhaps there's some additional set of APIs or way we can expose this. I'm not sure what makes the most sense or how we want to denote intent versus reality.

hawkw:
Last year, I merged #1953, which added a "reason" enum to the `Sequncer.set_power_state` IPCs that records why the power state changed. However, right now this only makes it into the sequencer's ringbuf and isn't exposed anywhere that's accessible without Humility. We could wire that value through into the MGS status messages, perhaps?

hawkw:
This would probably look like either the sequencer tasks hanging onto the last state change reason and exposing an IPC to read it that MGS would call, or eagerly posting it to MGS when the state changes. Without actually having thought too hard about it, the first seems nicer.

I'd want to put the state change reason in the same MGS message as the power state itself, rather than in a separate request to read the state change reason. Otherwise, there could be a race where MGS sees a state, and the power state changes again before MGS asks why we ended up in that state. At present, though, this would be a compute-sled-only field, so it wouldn't be included by sidecars/PSC. We might consider adding a similar thing to Sidecar's sequencer, but some of the state change reasons wouldn't apply there (there's no "host OS" to panic...).

rmustacc:
> We might consider adding a similar thing to Sidecar's sequencer, but some of the state change reasons wouldn't apply there (there's no "host OS" to panic...).

A big challenge I've had is where we apply things to only of many things that something applies to, e.g. just gimlet when also cosmo, compute sleds, when no it also applies to switches, etc.. I would really want to make sure that we do have this for all the sequencers and state. Even the PSC where there is no transition going on.

PR #2515, opened 2026-05-14
Report reason for power state change

evan-oxide:
This addresses https://github.com/oxidecomputer/hubris/issues/2335 and requires https://github.com/oxidecomputer/management-gateway-service/pull/492. Compute sleds were already tracking this reason for the last power state change but not exposing it to MGS. I added reason-tracking to sidecar as well, and now all boards using the control-plane-agent task can report a reason along with their current power state.

For the sake of backwards compatibility, I added new IPC calls and MGS messages instead of modifying the existing ones. It sounds like this is unfortunately the best approach for now.

**The sidecar-seq-server needs extra attention in the review**. I'm not confident that it's recording the correct reasons or making the right distinction between "policy changes" and "state changes". sidecar-seq-server tracks policy changes internally but then control-plane-agent renames them to state changes, to be consistent with the compute sleds.

Todo:

- [ ] Figure out the correct reason for some of the state/policy changes.
- [x] Merge https://github.com/oxidecomputer/management-gateway-service/pull/492 and then depend on it.

Testing:

I tested `GetPowerStateWithReason`, `GetPowerState`, and `SetPowerState` on grapefruit, cosmo, gimlet, psc, and sidecar. They are all able to report either `InitialPowerOn` or `ControlPlane` as the reason (except for the psc, which can't change state.) On sidecar, I also hardcoded an error into the thermal control loop and confirmed that it transitions to A2 with the reason `Overheat`.


drv/gimlet-seq-server/src/main.rs
```diff
@@ -622,7 +627,12 @@ impl<S: SpiServer> NotificationHandler for ServerImpl<S> {
                     self.seq
                         .clear_bytes(Addr::NIC_CTRL, &[cld_rst])
                         .unwrap_lite();
-                    self.update_state_internal(PowerState::A0PlusHP, now);
+                    // TODO what's the right reason?
+                    self.update_state_internal(
+                        PowerState::A0PlusHP,
+                        StateChangeReason::Unknown,
```
hawkw: I think this would be whatever the reason for the overall state transition we are currently undergoing. A0 to A0PlusHp means we have finished the ongoing transition to A0.
```suggestion
                    self.update_state_internal(
                        PowerState::A0PlusHP,
                        self.reason,
```

drv/gimlet-seq-server/src/main.rs
```diff
@@ -654,8 +664,10 @@ impl<S: SpiServer> NotificationHandler for ServerImpl<S> {
                     self.seq
                         .set_bytes(Addr::NIC_CTRL, &[cld_rst])
                         .unwrap_lite();
+                    // TODO what's the right reason?
                     self.update_state_internal(
                         PowerState::A0,
+                        StateChangeReason::Unknown,
```
hawkw, line 673: This one, on the other hand, should maybe be a new thing? It's saying that --- for whatever reason --- the NIC was powered on and is now being powered off. On Cosmo we have a "nic mapo" reason, but I'm not sure if this is necessarily a MAPO, especially since on Gimlet, the NIC is actually in the A0 MAPO domain due to "reasons". So I _think_ this means the host has decided to PERST the NIC? But don't take my word on it.

drv/sidecar-seq-server/src/main.rs
```diff
@@ -1127,7 +1155,11 @@ fn main() -> ! {
 
     // Power on, unless suppressed by the `stay-in-a2` feature
     if !cfg!(feature = "stay-in-a2") {
-        server.tofino.policy = TofinoSequencerPolicy::LatchOffOnFault;
+        // TODO what's the right reason?
+        server.tofino.set_policy(
+            TofinoSequencerPolicy::LatchOffOnFault,
+            PolicyChangeReason::InitialPowerOn,
+        );
```
hawkw: I think that `InitialPowerOn` feels correct to me here!

drv/cpu-seq-api/src/lib.rs
```diff
@@ -81,6 +86,24 @@ pub enum StateChangeReason {
...
+    Immutable,
+    KnownLayout,
+    Deserialize,
+    Serialize,
+    SerializedSize,
+)]
+pub struct PowerStateWithReason {
+    pub state: PowerState,
+    pub reason: StateChangeReason,
+}
```
hawkw, line 106: since we are adding a new message for this anyway, i wonder if it's worth also including the `since` timestamp (the hubris uptime, in milliseconds, at which we transitioned to this state), and maybe also the delta between that and the _current_ uptime? it's not strictly necessary just to solve the problem of "want to know why power state changed" but feels like a handy thing to throw in, and it's just 8 bytes in what's going to be a pretty tiny UDP datagram...

hawkw (commented):
Some notes, which I hope are helpful!

evan-oxide:
I removed `SpStateV4` as discussed offline, to avoid messing with an update-critical interface.

drv/cpu-seq-api/src/lib.rs
evan-oxide: Good idea, I added the timestamp. I assume it's easy enough for people to separately get the current timestamp and compare them.

drv/gimlet-seq-server/src/main.rs
evan-oxide: That makes sense, changed. Now the `InitialPowerOn` reason doesn't get overwritten with `Unknown` on startup.

drv/gimlet-seq-server/src/main.rs
evan-oxide: Any suggestions for who else we should consult?

Aaron-Hartwig: @evan-oxide I tagged in @nathanaelhuffman here since he has been heavily involved in the sled sequencing design on both Gimlet and Cosmo.

drv/sidecar-seq-api/src/lib.rs
```diff
@@ -68,6 +71,58 @@ pub enum TofinoSequencerPolicy {
...
+    Other = 1,
+    /// The system has just received power and has entered this state as part of
+    /// its normal startup process.
+    InitialPowerOn,
+    /// A power state change was requested by the control plane.
+    ControlPlane,
+    /// The system powered off because a component has overheated and/or the
+    /// thermal control loop has failed.
+    Overheat,
+}
```
Aaron-Hartwig, line 106: This is a good start, but we can actually get even more detailed scenarios from the FPGA should it have detected an error condition and powered down the Tofino. The FPGA has an [error register](https://github.com/oxidecomputer/quartz/blob/f5d4bc95a723450a79237d3d717a4531a1cc3486/hdl/projects/sidecar/mainboard/SidecarMainboardControllerSpiServer.bsv#L221) which can contain any of the [`Tofino2Sequencer::Error`](https://github.com/oxidecomputer/quartz/blob/f5d4bc95a723450a79237d3d717a4531a1cc3486/hdl/projects/sidecar/mainboard/Tofino2Sequencer.bsv#L97-L106) definitions. The way this is currently implemented in the FPGA is rather fragile; there is a type defined in the RTL, mapped into a fixed number of bits in the register, and then would have to be manually decoded properly in Hubris when read. We have patterns to handle this properly and I will get you an updated FPGA image + register definition to support this (tracked as https://github.com/oxidecomputer/quartz/issues/521).

Aaron-Hartwig (changes requested):
>The sidecar-seq-server needs extra attention in the review. I'm not confident that it's recording the correct reasons or making the right distinction between "policy changes" and "state changes". sidecar-seq-server tracks policy changes internally but then control-plane-agent renames them to state changes, to be consistent with the compute sleds.

Early on it was envisioned that we may take different dispositions towards faults encountered in A0 and may want the sequencer to automatically try to go from A2 -> A0 even when a fault had occurred. That is how we [ended up with](https://github.com/oxidecomputer/hubris/blob/d08aed03e0c8e8f497a2e6be14084255ac564fc6/drv/sidecar-seq-api/src/lib.rs#L65-L69) `Disabled` (A2), `LatchOffOnFault` (basically A0 and we go back to A2 on fault), and `RestartOnFault` (A0 but we try to bounce back after a fault). It turns out that [in practice](https://github.com/oxidecomputer/hubris/blob/d08aed03e0c8e8f497a2e6be14084255ac564fc6/drv/sidecar-seq-server/src/tofino.rs#L442-L444) **we never ended up using `RestartOnFault`**. With that background, I think there's a decent case for a broader changing the verbiage throughout the Sidecar sequencer to the idea of a `state` rather than a `policy`, but that can probably be a separate change after this.

I'm asking for a code change here because I will get you a Sidecar FPGA image which relies in RDL-encoded types, not RTL-encoded ones, which we can use to map the FPGA's error register to your `PolicyChangeReason` (or maybe `StateChangeReason someday 😉) type.

jamesmunns:
CC @evan-oxide, is it problematic that https://github.com/oxidecomputer/management-gateway-service/pull/492 landed if we might potentially need to make changes to the MGS API based on @Aaron-Hartwig's comments? I saw this come up in #2650, CC @hawkw for visibility. I want to avoid shipping a version of MGS with incompatible API types.

jgallagher:
> CC @evan-oxide, is it problematic that [oxidecomputer/management-gateway-service#492](https://github.com/oxidecomputer/management-gateway-service/pull/492) landed if we might potentially need to make changes to the MGS API based on @Aaron-Hartwig's comments? I saw this come up in #2650, CC @hawkw for visibility. I want to avoid shipping a version of MGS with incompatible API types.

Some historical context - we've treated MGS's API as "append only" to avoid having to worry much about backwards compatibility, and I think that's been largely successful, at the cost of having to deal with an "append only" API from a development perspective. That does usually mean "develop an MGS PR and hubris PR together, and wait to merge both of them until they're both ready to land", to avoid this kind of question.

That said, if we're very careful, we _can_ go back and change already-landed (or even already-shipped, with even more care) MGS APIs; we're just driving without seatbelts at that point and need to manually confirm it's okay (e.g., if no control plane consumer of the new API has been added yet, changing it is fine because there's no client with which we have to maintain compatibility).

evan-oxide:
Ok, in the future I'll make sure to merge them at the same time. In this case, neither hubris nor MGS is using the new API yet, so it should be safe to go back and add more variants to the `StateChangeReason` enum, right?

jgallagher:
> Ok, in the future I'll make sure to merge them at the same time. In this case, neither hubris nor MGS is using the new API yet, so it should be safe to go back and add more variants to the `StateChangeReason` enum, right?

Yep! If all we're doing is _adding_ variants, that could be done at any point in the future in a new MGS API version, too, since that doesn't change the serialization of existing variants.

PR #2518, opened 2026-05-18
Obtain Host Panic/Boot Failure Messages

### LINKED ISSUE #2504: control-plane-agent: Add interface(s) for retrieving Host Panic/Boot Failure messages
This is part of #2337, and a follow-up to #2140. We will need to update the protocol to handle retrieval of these messages, based on the information included in the `ereport`.

jamesmunns:
Not yet complete.

Closes #2504 
Currently based on #2503


task/packrat-api/src/lib.rs
```diff
@@ -72,6 +72,60 @@ pub enum EreportWriteError {
...
 }
 
+#[derive(
+    Copy, Clone, Debug, FromPrimitive, Eq, PartialEq, IdolError, counters::Count,
+)]
+pub enum HostInfoWriteError {
+    /// Storage of Host Info is not supported (this is a device not expected to
+    /// have a host)
+    Unsupported = 1,
+
```
hawkw: This should probably just be a `ReplyFault` that kills the caller --- how else are they going to deal with it? See for instance https://github.com/oxidecomputer/hubris/blob/3e2b6557e456b286d8f69de643d60bd36eb9806d/task/packrat/src/main.rs#L599-L612 where we return an `idol_runtime` error that turns into a reply fault.

task/packrat-api/src/lib.rs
```diff
@@ -72,6 +72,60 @@ pub enum EreportWriteError {
...
+    /// Failed to read from a provided lease, which probably means that the client
+    /// died before we serviced the request, so they'll probably never see this
+    LeaseLost,
```
hawkw: If the lease has gone away, this should probably just be `idol_runtime::ClientError::WentAway` rather than a client-visible error. See for example https://github.com/oxidecomputer/hubris/blob/1754c7412b252ca3bf984700e790c3e0a8b88d54/task/packrat/src/ereport.rs#L497-L498

task/packrat-api/src/lib.rs
```diff
@@ -72,6 +72,60 @@ pub enum EreportWriteError {
...
+
+    #[idol(server_death)]
+    ServerRestarted,
```
hawkw: ...and, if we eliminate the `LeaseLost` and `Unsupported` errors, which (as I mentioned above) the client has no reasonable response to and (in the `LeastLost` case) really should not ever be able to handle if it occurs, we don't really need a custom error type for this as the only remaining error is server death.

task/packrat-api/src/lib.rs
```diff
@@ -72,6 +72,60 @@ pub enum EreportWriteError {
...
+    /// The requested byte-offset is beyond the range of the currently stored
+    /// host info.
+    InvalidOffset,
+
+    /// Requested index does not match the currently stored host information
+    InvalidIndex,
+
+    /// Failed to write to a provided lease, which probably means that the client
+    /// died before we serviced the request, so they'll probably never see this
+    LeaseLost,
```
hawkw: Again, this is probably `ClientError::WentAway`?

task/host-sp-comms/src/main.rs
```diff
@@ -978,20 +970,32 @@ impl ServerImpl {
...
-                    *b = 0;
-                }
-                self.host_kv_storage.last_boot_fail_reason = reason;
+
+                // Store the bootfail message in packrat so it can be accessed by MGS in the future
+                // TODO: What to do if this call fails? Without it, we don't have a proper index, but
+                // this would only happen if packrat crashes. We could store some fake info here to
+                // continue preparing an ereport, but THAT is going to be a problem anyway because
+                // we send ereports to, you guessed it: packrat, which just crashed.
+                let response = self
```
jamesmunns, line 986: There's probably something better we can do here as well, the only failure modes for a `write_host_bootfail` (and host panic) call are:

1. That's illegal (not supported, OR client dies and revokes the lease)
2. Packrat died

Neither of these are *particulary* good things (the former doesn't even matter here), and there might be a better way to express the idolatry IPC API to make this more clear in the generated code.

jamesmunns:
MGS message changes are tracked in https://github.com/oxidecomputer/management-gateway-service/pull/495.

task/packrat-api/src/lib.rs
```diff
@@ -72,6 +72,59 @@ pub enum EreportWriteError {
...
+    pub offset: usize,
+    pub total_len: usize,
+    pub index: u32,
+}
+
+#[derive(Copy, Clone, Debug, FromBytes, IntoBytes, Immutable, KnownLayout)]
+#[repr(C)]
+pub struct HostInfoRequest {
+    pub offset: u32,
+    pub index: u32,
```
hawkw: sorry in advance if this feels bikesheddy, but i don't love calling these an `index` --- it's more of a sequence number or count of the number of times the host has panicked/bootfailed. `index` to me kind of implies we are storing several of them and you can read any of them, but you really can't.

task/control-plane-agent/src/mgs_compute_sled.rs
```diff
@@ -1259,6 +1260,101 @@ impl SpHandler for MgsHandler {
...
+            self.common.packrat().read_host_panic_fragment(
+                task_packrat_api::HostInfoRequest {
+                    offset: req.offset,
+                    index: req.index,
+                },
+                dest,
+            )
+        } else {
+            self.common.packrat().read_first_host_panic_fragment(dest)
+        };
```
hawkw: hm, it seems to me that this would be a bit simpler if we just made the packrat API take an option, too? do we really need two totally separate IPCs?

jamesmunns:
Should be able to move this out of WIP tomorrow, progress update here: https://github.com/oxidecomputer/hubris/issues/2504#issuecomment-4858712047. I'll address Eliza's comments before marking as ready for review.

task/packrat-api/src/lib.rs
jamesmunns: Changed `index` to `seqno`

task/control-plane-agent/src/mgs_compute_sled.rs
jamesmunns: You're correct, but the reason I avoided this was that unfortunately you can't use `Option` with `zerocopy`, and `packrat` wasn't already explicitly using hubpack/serde.

I have moved it over in https://github.com/oxidecomputer/hubris/pull/2518/commits/0af5cda4dd6799c3e207bf96050cb9d3bc456fb4, and I do agree this is nicer (despite adding another wire format).

jamesmunns:
This is IMO ready to review, there are still a couple of `TODO` comments, but they are items that I could probably stand to have feedback on!

task/host-sp-comms/src/main.rs
```diff
@@ -978,20 +970,32 @@ impl ServerImpl {
...
+                    .packrat
+                    .write_host_bootfail(reason, data)
+                    .unwrap_lite();
+
+                // TODO: Do we want to give this to packrat too?
+                // TODO: Update `humility host boot-fail` to use packrat API!
```
jamesmunns: This is still pending. I might need to open a `humility` PR that handles this elegantly.

task/packrat/src/main.rs
```diff
@@ -569,6 +637,298 @@ impl idl::InOrderPackratImpl for ServerImpl {
...
+        let new_seq = self
+            .host_info
+            .host_bootfail_state
+            .take()
+            .map(|s| s.sequence_number.wrapping_add(1))
+            .unwrap_or(0)
+            .max(1);
+        self.host_info.host_bootfail_state = Some(HostBootFailMetadata {
+            total_length: to_copy,
+            sequence_number: new_seq,
```
hawkw, line 655: take it or leave it: it occurs to me that this code and the similar code for writing a host panic is pretty duplicative, and the only real difference here is that `HostBootFailMetadata` also has a `reason`, making it a different type. I wonder a bit if we might be able to do something where most of the code can be shared. I was imagining something like this:

```rust
fn copy_host_info(
    data: Leased<idol_runtime::R, [u8]>,
    buf: &mut [u8; 4096],
    current_seq: Option<u32>,
) -> Result<
    HostInfoWriteOutput>,
    idol_runtime::RequestError<core::convert::Infallible>,
> {
   // ...
}
```
but then I realized that we really want to avoid blowing away the current state if the copy fails, including the `reason`, so there isn't really a good way to make it generic over the two metadata types without messing around with a trait or something, which is maybe not worth it.

maybe we could do

```rust
fn copy_host_info(
    data: Leased<idol_runtime::R, [u8]>,
    buf: &mut [u8; 4096],
) -> Result<
    usize,
    idol_runtime::RequestError<core::convert::Infallible>,
> {
   // ...
}
```

which _just_ does the copying part, but I dunno if that's enough code to be worth refactoring just to de-duplicate. so maybe none of this is a good idea, IDK. what do you think?

task/host-sp-comms/src/main.rs
```diff
@@ -2020,3 +2034,50 @@ mod idl {
...
+    /// This count will wrap, but is guaranteed to never be zero.
+    seqno: u32,
+    /// The length, in bytes, of the stored panic message.
+    ///
+    /// This quantity may be less than the amount received, as it is capped
+    /// by the available storage space allocated (`MAX_HOST_FAIL_MESSAGE_LEN`).
+    msglen: u32,
+    /// The flash boot index, directly correlated to which boot slot we are
+    /// operating from. Currently 0 (BSU: A), 1 (BSU: B), or 0xFF (unknown).
+    flashidx: u8,
```
hawkw: i think i would prefer to get a CBOR `null` instead of a sentinel value that's specific to this ereport, even if it makes it a byte bigger.

task/host-sp-comms/src/main.rs
```diff
@@ -2020,3 +2034,50 @@ mod idl {
...
+    seqno: u32,
+    /// The length, in bytes, of the stored panic message.
+    ///
+    /// This quantity may be less than the amount received, as it is capped
+    /// by the available storage space allocated (`MAX_HOST_FAIL_MESSAGE_LEN`).
+    msglen: u32,
+    /// The reported reason code for the host boot failure
+    reason: u8,
+    /// The flash boot index, directly correlated to which boot slot we are
+    /// operating from. Currently 0 (BSU: A), 1 (BSU: B), or 0xFF (unknown).
```
hawkw: similarly, i don't love `ff`

hawkw (commented):
i have more reviewing to do but here are a few things i left comments on

task/host-sp-comms/src/main.rs
```diff
@@ -1003,19 +1007,29 @@ impl ServerImpl {
...
-                }
+                // Store the panic message in packrat so it can be accessed by MGS in the future
+                // TODO: What to do if this call fails? Without it, we don't have a proper index, but
+                // this would only happen if packrat crashes. We could store some fake info here to
+                // continue preparing an ereport, but THAT is going to be a problem anyway because
+                // we send ereports to, you guessed it: packrat, which just crashed.
+                let response =
+                    self.packrat.write_host_panic(data).unwrap_lite();
+
+                // TODO: Do we want to give this to packrat too?
```
labbott: What would patrack do with this information? Or is the thought this would be useful for the human debugging?

task/host-sp-comms/src/main.rs
labbott: I think this is covered by humility#710 ?

idl/packrat.idol
```diff
@@ -144,5 +144,60 @@ Interface(
...
+            args: {
+                "reason": "u8",
+            },
+            leases: {
+                "data": (type: "[u8]", read: true),
+            },
+            reply: Result(
+                ok: "HostInfoWriteOutput",
+                err: ServerDeath,
+            ),
```
labbott, line 159: The actual functions are `Infallible`.  Other functions model this as a `Simple` reply type to avoid the need for the extra unwrap. otoh, these functions also copy from a lease so they do have a possibility of a `ServerDeath`.  I think you're right that we might want idolatry to handle this.

task/host-sp-comms/src/main.rs
jamesmunns: This is indeed covered! Removed.

task/host-sp-comms/src/main.rs
jamesmunns: This is now `Option<u16>`, where that was chosen to match with some other MGS "slot" data types.

task/host-sp-comms/src/main.rs
jamesmunns: good news! it's gone!

task/host-sp-comms/src/main.rs
jamesmunns: It would be stored to pass along to MGS, which we now do!

task/packrat/src/main.rs
jamesmunns: I think I will leave it as dumb and duplicative now, instead of trying to be clever. Let me know if you start feeling stronger about this.

jamesmunns:
Just confirming that this still works after testing: https://github.com/oxidecomputer/management-gateway-service/pull/495#issuecomment-4972072947

drv/stm32h7-rng/Cargo.toml
```diff
@@ -23,7 +23,7 @@ idol = { workspace = true }
 [features]
 h743 = ["stm32h7/stm32h743", "drv-stm32xx-sys-api/h743"]
 h753 = ["stm32h7/stm32h753", "drv-stm32xx-sys-api/h753"]
-ereport = ["task-packrat-api"]
+packrat = ["task-packrat-api"]
```
hawkw, line 26: This is a super unimportant turbo-nit, but I am not sure if I love calling this feature "packrat", since it enables a very specific _behavior_ which includes but is not only talking to Packrat. My previous name, "ereport", also didn't really explain what the actual thing it enables was, but at the time, it made sense since it was only required for ereports. I wonder if we want to call the feature "packrat-restart-id" or something that makes it specific what it controls?

idl/packrat.idol
```diff
@@ -144,5 +144,64 @@ Interface(
             ),
             idempotent: true,
         ),
+        "write_host_bootfail": (
+            doc: "Write a host's boot failure message and return the index of this failure",
```
hawkw: nitpicky nitpicky, but perhaps:
```suggestion
            doc: "Write the host's boot failure message and return the index of this failure",
```
or even

```suggestion
            doc: "Write a host boot failure message and return the index of this failure",
```
since "write a host's" kind of implies to me that we might receive such messages from multiple hosts?

idl/packrat.idol
```diff
@@ -144,5 +144,64 @@ Interface(
...
+            },
+            reply: Result(
+                ok: "HostBootfailReadOutput",
+                err: CLike("HostInfoReadError"),
+            ),
+            idempotent: true,
+            encoding: Hubpack,
+        ),
+        "write_host_panic": (
+            doc: "Write a host's panic message and return the index of this panic",
```
hawkw: similarly, I might say
```suggestion
            doc: "Write the host's panic message and return the index of this panic",
```

idl/packrat.idol
```diff
@@ -144,5 +144,64 @@ Interface(
...
+                "data": (type: "[u8]", read: true),
+            },
+            reply: Result(
+                ok: "HostInfoWriteOutput",
+                err: ServerDeath,
+            ),
+            encoding: Hubpack,
+        ),
+        "read_host_bootfail_fragment": (
+            doc: "Read a portion of the host's boot failure message",
```
hawkw, line 163: maybe this ought to explain how the request works? although I suppose the request type has RustDoc, it could be nice to say explicitly what `None` means...

task/host-sp-comms/src/main.rs
```diff
@@ -1003,19 +1006,25 @@ impl ServerImpl {
...
-                    self.host_kv_storage.last_panic.len(),
-                );
-                self.host_kv_storage.last_panic[..n]
-                    .copy_from_slice(&data[..n]);
-                for b in &mut self.host_kv_storage.last_panic[n..] {
-                    *b = 0;
-                }
+                // TODO: The flashidx *at panic time* may not be the *flashidx
+                // used when the panicking host booted*.
+                let flashidx = None;
```
hawkw, line 1013: figuring out a nicer solution for this probably deserves a ticket in addition to a TODO?

task/host-sp-comms/src/main.rs
```diff
@@ -978,20 +970,31 @@ impl ServerImpl {
...
-                    *b = 0;
-                }
-                self.host_kv_storage.last_boot_fail_reason = reason;
+
+                // Get the flash index used for the currently booting host
+                let flashidx = match self.hf.get_dev() {
+                    Ok(HfDevSelect::Flash0) => Some(0),
+                    Ok(HfDevSelect::Flash1) => Some(1),
+                    Err(_) => None,
+                };
```
hawkw, line 979: This kiiinda makes me want to change the ereport and MGS messages to represent this as an enum instead of an arbitrary u16, but...

task/host-sp-comms/src/main.rs
```diff
@@ -2020,3 +2029,52 @@ mod idl {
...
+    seqno: u32,
+    /// The length, in bytes, of the stored panic message.
+    ///
+    /// This quantity may be less than the amount received, as it is capped
+    /// by the available storage space allocated (`MAX_HOST_FAIL_MESSAGE_LEN`).
+    msglen: u32,
+    /// The flash boot index, directly correlated to which boot slot we are
+    /// operating from. Currently Some(0) (BSU: A), Some(1) (BSU: B), or None
+    /// (unknown).
+    flashidx: Option<u16>,
```
hawkw, line 2059: why does this get encoded as a 16-bit integer if it's only ever two values? if we're not going to represent it as an enum (which I'm fine with), it seems like it could be a `u8`?

task/host-sp-comms/src/main.rs
```diff
@@ -2020,3 +2029,52 @@ mod idl {
...
+}
+
+/// An ereport represent a host reported boot failure
+#[derive(Encode)]
+#[ereport(class = "host.btfail", version = 0)]
```
hawkw: take it or leave it, but i think expanding this to `bootfail` is maybe worth the two bytes? if we wanted that change to be encoded-length-neutral, we could steal them from `seqno` by changing it to just `seq` (which I personally think is just as clear, but YMMV...)

task/packrat-api/src/lib.rs
```diff
@@ -72,6 +75,63 @@ pub enum EreportWriteError {
...
+    Copy, Clone, Debug, FromPrimitive, Eq, PartialEq, IdolError, counters::Count,
+)]
+pub enum HostInfoReadError {
+    /// We have never received the requested Host Info, or this is a platform
+    /// that is not expected to have a host.
+    NoHostInfo = 1,
+
+    /// The requested byte-offset is beyond the range of the currently stored
+    /// host info.
+    InvalidOffset,
```
hawkw, line 122: nit, take it or leave it: if "invalid" specifically means "too long", I might say that explicitly?

task/packrat/src/main.rs
```diff
@@ -154,12 +161,62 @@ enum TraceSet<T> {
...
+impl HostInfo {
+    const fn new() -> Self {
+        Self {
+            host_panic_payload: [0u8; _],
+            host_bootfail_payload: [0u8; _],
+            host_panic_state: None,
+            host_bootfail_state: None,
+        }
+    }
+}
```
hawkw: take it or leave it: the amount of things that we are feature flagging here make me wonder if maybe we should try to stuff all this stuff inside of a feature flagged module? up to you.

task/packrat/src/main.rs
```diff
@@ -154,12 +161,62 @@ enum TraceSet<T> {
...
+    /// Boot slot
+    slot: Option<u16>,
+}
+
+#[cfg(any(feature = "gimlet", feature = "grapefruit", feature = "cosmo"))]
+pub struct HostInfo {
+    host_panic_payload: [u8; 4096],
+    host_bootfail_payload: [u8; 4096],
+    host_panic_state: Option<HostPanicMetadata>,
+    host_bootfail_state: Option<HostBootFailMetadata>,
```
hawkw: take it or leave it: imo, we can drop `host_` from all of these since it lives in a struct named "host info", but you do you

task/packrat/src/main.rs
```diff
@@ -154,12 +161,62 @@ enum TraceSet<T> {
...
+    /// (hopefully not) Rolling counter of panic messages observed this power cycle
+    sequence_number: u32,
+    /// Bootfail reason
+    reason: u8,
+    /// Boot slot
+    slot: Option<u16>,
+}
+
+#[cfg(any(feature = "gimlet", feature = "grapefruit", feature = "cosmo"))]
+pub struct HostInfo {
```
hawkw: i might consider renaming "host info" to something that makes it a little clearer that this is specifically host crash debugging buffers, "host info" could mean be a lot of things...

task/packrat-api/src/lib.rs
```diff
@@ -72,6 +75,63 @@ pub enum EreportWriteError {
...
+    pub offset: u32,
+    pub total_len: u32,
+    pub seqno: u32,
+    pub restart_id: RestartId,
+    pub slot: Option<u16>,
+}
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, SerializedSize)]
+#[repr(C)]
+pub struct HostInfoRequest {
```
hawkw, line 107: i might consider renaming "host info" to something that makes it a little clearer that this is specifically host crash debugging buffers, "host info" could mean be a lot of things...

task/packrat/src/main.rs
```diff
@@ -569,6 +638,318 @@ impl idl::InOrderPackratImpl for ServerImpl {
...
+    > {
+        // Do we have a restart ID?
+        let Some(restart_id) = self.restart_id else {
+            return Err(HostInfoReadError::MissingRestartId.into());
+        };
+
+        // Do we *have* a panic to report?
+        let Some(bfs) = self.host_info.host_panic_state.as_ref() else {
+            return Err(HostInfoReadError::NoHostInfo.into());
+        };
```
hawkw: there is a bit of a design question about which of these errors is more "relevant" to return first. IMO, if we haven't gotten the restart ID yet but we also know we don't actually need it to say "lol there's no panic", i might communicate the "no panic to read" bit instead of the fact that we are (hopefully transiently) without a restart ID...

hawkw (approved):
Overall, this looks great! I commented on a bunch of naming quibbles and other unimportant things, but the implementation looks solid. Do what you want with my suggestions. :)

drv/stm32h7-rng/Cargo.toml
jamesmunns: The reason I called it "packrat" is because this feature should be enabled whenever "packrat" is present. We could get more granular and split out packrat's capabilities, but that would also then require packrat to have a "restart-id" feature which can be selected, and would be required when "ereports" is enabled.

Maybe we punt on this until we actually need/want that granularity?

idl/packrat.idol
jamesmunns: No complaint, my intent was to signal "the boot failure message is from the host". I can remove it though, I use too many apostrophes anyway.

task/host-sp-comms/src/main.rs
jamesmunns: Oooh, yes, I meant to open a follow-on issue, will do.

task/host-sp-comms/src/main.rs
jamesmunns: Yeah, I even bumped it up from a `u8` to better match the rest of the (IIRC?) inventory APIs that express this as a `u16`.

task/host-sp-comms/src/main.rs
jamesmunns: Mentioned above, there are other places in gateway-messages that treat the "boot slot" as a u16, and I thought it was worth being consistent in the face of brevity.

task/host-sp-comms/src/main.rs
jamesmunns: Sure, happy to mess with any of the names. I think at some point you suggested I make things as short as possible to save space, but I definitely could use some Opinions On The Matter Of Taste for what is Short Enough and what is Too Short.

task/packrat-api/src/lib.rs
jamesmunns: I might keep this as-is as the MGS change has already landed and this keeps the names consistent.

task/packrat/src/main.rs
jamesmunns: I think that's reasonable, we do that for other stuff, couldn't hurt to split this out as well.

task/packrat/src/main.rs
jamesmunns: Yeah, I was looking for a name for "info we got from the host", but I'll see if I can find a better name.

task/packrat/src/main.rs
jamesmunns: True, that's reasonable, I was thinking in the order of "what is the cheapest to check", but honestly both are, and what you said is probably more reasonable to report in practice.

jamesmunns:
@hawkw no major arguments here, I'll make the minor cleanups tomorrow and hit merge unless you want me to wait until you can sign off again.

Thanks for the comments!

drv/stm32h7-rng/Cargo.toml
hawkw: Yeah, I think that's fair, I can accept it.

task/host-sp-comms/src/main.rs
hawkw: Ah, well, in that case I guess that makes sense. Hmm.

task/packrat-api/src/lib.rs
hawkw: ah, yup, nvm

task/host-sp-comms/src/main.rs
jamesmunns: Done

PR #2520, opened 2026-05-19
cosmo_seq: mask STATUS_CML bit 1 to not get annoying pmbus alerts about it

hawkw:
It turns out that PMBus devices like to set this bit bascially any time they saw any kind of I2C behavior that makes them feel uncomfortable. We probably don't want to produce an ereport every time an I2C Thing makes a VRM uncomfortable. So this commit masks out PMBus alerts for STATUS_CML's "other communication fault" bit, since it's annoying and not a big deal.

This branch does a bit of extra refactoring on top of the change from #2511. Since we are now setting even more `SMBALERT_MASK`s in `initialize_pmbus_alerts`, I thought it would be nice to have some shared utilities for trying to do this.

### DIFF
--- a/drv/cosmo-seq-server/src/vcore.rs
+++ b/drv/cosmo-seq-server/src/vcore.rs
@@ -18,6 +18,7 @@ use drv_i2c_api::ResponseCode;
 use drv_i2c_devices::raa229620a::{self, Raa229620A};
 use ereports::pwr::{PmbusAlert, PmbusStatus};
 use fixedstr::FixedStr;
+use pmbus::commands::raa229620a::STATUS_CML;
 use pmbus::commands::raa229620a::STATUS_IOUT;
 use pmbus::commands::raa229620a::STATUS_WORD;
 use ringbuf::*;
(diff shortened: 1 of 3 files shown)


hawkw:
hm [this build failure](https://github.com/oxidecomputer/hubris/actions/runs/26118734747/job/76815009169?pr=2520) is, i believe, due to GitHub Jank and is not my fault?

hawkw:
looks like [either the same CI weather has occurred again, or Windows CI is now busted in a sticky way that has nothing to do with this PR](https://github.com/oxidecomputer/hubris/actions/runs/26123004110/job/76829593512?pr=2520). imagine that i'm crying, screaming, and throwing up about this, because i certainly am emotionally, if not literally.

mkeeter (approved):

hawkw:
okay, so i have not actually been able to come up with a way to reproduce a condition that actually asserts this particular bit --- the best i could manage with `humility i2c`, (writing too many bytes to a register or scanning registers that don't exist) assert bits 5 (PEC failure) and 7 (invalid command) instead. i think asserting bit 1 requires glitching the bus in a way that Humility is too well-behaved to do, such as sending a STOP mid-transaction or just messing with SCL at unexpected moments. i am, however, confident enough that this probably does do the right thing, and _have_ validated that all the refactoring has, at least, not broken the previously working masking of the spurious overcurrent errors.  i am going to exercise my best judgement and declare this good enough.

PR #2521, opened 2026-05-19
Bump dependencies

mkeeter:
I ran into a similar issue as https://github.com/oxidecomputer/hubris/pull/2480, because`tlvc` and `tlvc-text` specified versions which are no longer at the tip of the repo.  After bumping them, a bunch of other stuff broke, so I bumped it as well!  

The only logic change is that we're now building the initial Hubris archive with `bzip2` compression (instead of `deflate`).  This doesn't change the end result, because `hubpack` was modifying our archives (e.g. adding a caboose) and _already_ used `bzip2` compression when writing them back.  It _does_ let us remove a few dependencies by removing the `deflate` feature from `zip`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,24 +14,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "adler2"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
-
-[[package]]
-name = "ahash"
-version = "0.8.12"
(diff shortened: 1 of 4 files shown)


hawkw:
> This doesn't change the end result, because `hubpack` was modifying our archives (e.g. adding a caboose) and _already_ used `bzip2` compression when writing them back

hub_pack_, or hub_tools_? I think you meant hubtools...

mkeeter:
I did mean `hubtools`!  `hubpack` is blameless (because it's distributed through crates.io)

hawkw (approved):

jamesmunns:
CC @labbott - this PR requires a RoT group signoff because it touches `/lib/dice`

labbott (approved):

PR #2524, opened 2026-05-20
Add `cargo xtask rust-analyzer`

mkeeter:
This PR adds `cargo xtask rust-analyzer`, which points `rust-analyzer` at a particular `(manifest, task)` tuple (set by a CLI argument).  It can be used as a drop-in replacement in your editor, to get LSP goodness within the Hubris build system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ hubris.testout.*
 !.vscode/tasks.json
 !.vscode/extensions.json
 .nvim.lua
+.zed/settings.json
+.dir-locals.el
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ hubris.testout.*
 !.vscode/launch.json
 !.vscode/tasks.json
 !.vscode/extensions.json
+.nvim.lua
--- a/build/xtask/src/rust_analyzer.rs
+++ b/build/xtask/src/rust_analyzer.rs
@@ -48,6 +48,12 @@ impl LspConfig {
                 .map(|(k, v)| (k.clone(), v.clone().into())),
         );
         cargo.insert("target".to_owned(), self.target.to_owned().into());
+
+        // Only check the package being edited, to avoid errors from all the
+        // other crates in the workspace (which may be incompatible with this
+        // specific task's build configuration)
+        let check = get_or_insert(options, "check");
+        check.insert("workspace".to_owned(), false.into());
     }
 }
 
--- a/build/xtask/src/rust_analyzer.rs
+++ b/build/xtask/src/rust_analyzer.rs
@@ -64,14 +64,18 @@ impl LspConfig {
         let check = get_or_insert(options, "check");
         check.insert("workspace".to_owned(), false.into());
 
-        // Okay, this is where it gets awkward.  By default, the flycheck
-        // program calls `cargo check` (or `cargo clippy`) on a single package,
-        // but uses the full set of features defined in
+        // Okay, this is where it gets awkward.  When invoked on a binary crate,
+        // rust-analyzer's flycheck pass calls `cargo check` (or `cargo clippy`)
+        // on that single package, but uses the full set of features defined in
         // `rust-analyzer.cargo.features`.  This fails if you're enabling
-        // features for multiple crates, which we want to do!  Instead, we'll
-        // replace the `check` command wholesale with one that does what we
-        // want.  This is less efficient, because it checks every crate in our
-        // tree, but has the advantage of working.
+        // features for multiple crates, which we want to do!
+        //
+        // Instead, we'll replace the `check` command wholesale with one that
+        // does what we want.  This is less efficient, because it checks every
+        // crate in our tree, but has the advantage of working.
+        //
+        // We have fixed this upstream (rust-lang/rust-analyzer#22432), so can
+        // remove this workaround once that fix ships.
         let check_cmd = check
             .get("command")
             .and_then(|c| c.as_str())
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -14,3 +14,4 @@ hubris.testout.*
 !.vscode/launch.json
 !.vscode/tasks.json
 !.vscode/extensions.json
+.nvim.lua
```
jamesmunns, line 17:
```suggestion
.nvim.lua
.zed/settings.json
```

jamesmunns:
This definitely works *better* for me in Zed, there are still some hiccups, I'll try to enumerate them a bit better once I've gotten a better chance to take this for a spin.

Looking at the logs, I see some weirdness: https://gist.github.com/jamesmunns/59c51489e6e62676bef118514a3fddbf

Some of this is attempting to build crates that aren't relevant for cosmo+control_plane_agent, e.g. task-attest.

There are some recurring errors that seem odd to me, like:

```
  Error: app.toml missing task config section [tasks.control_plane_agent.config]
```

And also:

```
Error: reading env var $HUBRIS_KCONFIG
```

That all being said, I *am* able to use rust-analyzer in the `control-plane-agent` task, as well as seemingly other cosmo-relevant tasks.

For context, this is my current `.zed/settings.json`:

```json5
// Folder-specific settings
//
// For a full list of overridable settings, and general information on folder-specific settings,
// see the documentation: https://zed.dev/docs/configuring-zed#settings-files
{
  "lsp": {
    // Specify the LSP name as a key here.
    "rust-analyzer": {
      // // A special flag for rust-analyzer integration, to use server-provided tasks
      // enable_lsp_tasks": true,
      "binary": {
        "path": "cargo",
        "arguments": [
          "run",
          "-pxtask",
          "--",
          "rust-analyzer",
          "--log",
          "./target/xtask-ra.log",
          "./app/cosmo/rev-b-dev.toml:control_plane_agent",
        ],
      },
      // These initialization options are merged into Zed's defaults
      "initialization_options": {
        "check": {
          "command": "clippy", // rust-analyzer.check.command (default: "check")
        },
        "cargo": {
          "targetDir": "target",
        },
      },
    },
  },
}
```

jamesmunns (approved):
Happy to do a follow-on PR to document setting up Zed as well.

mkeeter:
I pushed a fix to set `check { workspace: false }`, which should reduce some of the chattiness (because it will only run `cargo check` for the package being edited).

mkeeter:
> I pushed a fix to set `check { workspace: false }`, which should reduce some of the chattiness (because it will only run `cargo check` for the package being edited).

This was a bad idea due to https://github.com/rust-lang/rust-analyzer/pull/22432 ; the current implementation builds a full `overrideCommand` for flycheck.

build/xtask/src/rust_analyzer.rs
```diff
@@ -0,0 +1,505 @@
...
+                && let Ok(file) = path.canonicalize()
+            {
+                let is_prefix =
+                    cfg.packages.iter().any(|(root, _)| file.starts_with(root));
+                if !is_prefix {
+                    let id = format!("xtask-0/{}", self.msg_id);
+                    self.msg_id += 1;
+                    let msg = format!(
+                        "This file is not used by {}:{}; \
+                         LSP support will be degraded",
```
evan-oxide, line 434: 2 issues: 
- Having this interactive message pop up and demand a response when I open a file degrades the experience more than the thing it's telling me about. Consider using `window/showMessage` instead.
- It seems to trigger the first time I open *any* file. For example when I open `psc-seq-server/src/main.rs` it says "This file is not used by psc-c-dev:sequencer", which is false.

.gitignore
evan-oxide: Can we add `.dir-locals.el` too?

mkeeter: Done!

build/xtask/src/rust_analyzer.rs
mkeeter: - Switched to `showMessage` instead, thanks!  (The LSP docs are not the most discoverable)
- This was actually a bug triggered by `psc-seq-server` specifically because it enables no features, so the graph search didn't have a starting point!  It should be fixed now.

evan-oxide: `showMessage` isn't working for me, I think because of a distinction between requests and notifications. From the log of LSP communication, this message was shown correctly:
```
[Trace - 11:10:22 AM] Received notification 'window/showMessage'.
Params: {
  "message": "Failed to run build scripts of some packages.\n\nPlease refer to the logs for more details on the errors.",
  "type": 2
}
```
and this triggered "Warning (lsp-mode): Unknown request method: window/showMessage":
```
[Trace - 11:10:29 AM] Received request 'window/showMessage - (xtask-0/0).
Params: {
  "message": "This file is not used by observer-a:sequencer; LSP support will be degraded",
  "type": 2
}
[Trace - 11:10:29 AM] Sending response 'window/showMessage - (xtask-0/0)'. Processing request took 16ms
Params: {
  "jsonrpc": "2.0",
  "id": "xtask-0/0",
  "result": null
}
```

mkeeter: Ah, that's because I included an `id` field (which makes it a request rather than a notification).  I pushed a fix, let me know how it looks!

evan-oxide: Yup that fixed it, thanks!

Issue #2535, opened 2026-05-27
CI fails intermittently on Windows while restoring cache


jamesmunns:
We've noticed that *occasionally*, a windows build will fail while running the `build-boards.yml` `build` step.

The symptoms are that the cache restore step will fail silently, generally within <0.5s, during the "Cache build output" step, which uses `Swatinem/rust-cache@v2`. At the time of writing, all instances of this that I looked into (not exhaustively) were running the latest version, v2.9.1.

This looks like this in the succinct/GUI logs:

```
Run Swatinem/rust-cache@v2
(node:3960) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Cache Configuration

... Restoring cache ...
```

That step then ends in failure, and subsequent steps other than cleanup are skipped.

and in the raw logs:

```
2026-05-26T21:29:02.9618339Z ##[group]Cache Configuration
2026-05-26T21:29:02.9618712Z Cache Provider:
2026-05-26T21:29:02.9618947Z     github
2026-05-26T21:29:02.9619197Z Workspaces:
2026-05-26T21:29:02.9619488Z     D:\a\hubris\hubris
2026-05-26T21:29:02.9619782Z Cache Paths:
2026-05-26T21:29:02.9620117Z     C:\Users\runneradmin\.cargo\bin
2026-05-26T21:29:02.9620524Z     C:\Users\runneradmin\.cargo\.crates.toml
2026-05-26T21:29:02.9621084Z     C:\Users\runneradmin\.cargo\.crates2.json
2026-05-26T21:29:02.9621630Z     C:\Users\runneradmin\.cargo\registry
2026-05-26T21:29:02.9622098Z     C:\Users\runneradmin\.cargo\git
2026-05-26T21:29:02.9622517Z     D:\a\hubris\hubris\target
2026-05-26T21:29:02.9622882Z Restore Key:
2026-05-26T21:29:02.9623218Z     v0-rust-build-Windows_NT-x64-a9551aa0
2026-05-26T21:29:02.9623653Z Cache Key:
2026-05-26T21:29:02.9624024Z     v0-rust-build-Windows_NT-x64-a9551aa0-70a78afc
2026-05-26T21:29:02.9624557Z .. Prefix:
2026-05-26T21:29:02.9624923Z   - v0-rust-build-Windows_NT-x64
[...]

PR #2536, opened 2026-05-27
Allow windows to continue on cache restore error

jamesmunns:
Mitigation for https://github.com/oxidecomputer/hubris/issues/2535

### DIFF
--- a/.github/workflows/build-boards.yml
+++ b/.github/workflows/build-boards.yml
@@ -89,6 +89,10 @@ jobs:
 
       - name: Cache build output
         uses: Swatinem/rust-cache@v2
+        # See: https://github.com/oxidecomputer/hubris/issues/2535 - allow windows to still continue
+        # if cache fails. Make this conditional because we don't *think* we've observed this on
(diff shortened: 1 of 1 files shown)


jamesmunns:
One tricky thing about this approach is that I don't know how we'd be able to determine if this mitigation is REALLY working or not (as the issue is intermittent!), short of looking at "builds that took a bit longer" in the actions history.

We'll definitely know if it *doesn't* work, as we'll still see CI failures on Windows.

hawkw (approved):
Personally, I think this is better than not having it, even though I agree that it's gonna be kinda hard to tell if it's working..

emilyalbini:
Before merging this [I'd rather enable debug logging in all workflows](https://github.com/oxidecomputer/hubris/issues/2535#issuecomment-4562883737). If we merge this too it's going to be harder to notice when it fails (and when to grab the logs).

jamesmunns:
Converting to draft so we remember that we might not want to merge this until we've had time to debug this. See #2540 for context.

PR #2538, opened 2026-05-27
PMBus Status Registers via MGS

### LINKED ISSUE #2463: want to read PMBus status registers via MGS
This is motivated by two different needs:

1. we recently had a rectifier die on us at a customer site ([customer-support#898](https://github.com/oxidecomputer/customer-support/issues/898)), and we were not able to determine if the rectifier was so dead that we were unable to talk PMBus at it or not, because it was in a state where the PSC sequencer does not automatically attempt to read and record the PMBus status. It would have been nice to be able to use faux-mgs to talk to it while debugging.
2. for automated fault management in the control plane, reading the status registers of a PMBus device is an important [health endpoint (as described in RFD 589)](https://rfd.shared.oxide.computer/rfd/0589#_health_endpoints) for that device.

It would be nice if there was a way to read the contents of all of the standard PMBus status registers over the management network. Ideally, this would happen automagically via codegen for any device with a 
```toml
power = { /* ... */ pmbus = true }
```
in its declaration in the TOML file.

This could either be done using a new `ComponentDetails` entry for PMBus status, or a separate MGS API that's PMBus-specific. I'm a bit on the fence as to which is better --- personally, I think hanging it off of component details might be a bit of a shame as it means we would do a bunch more I2C traffic when trying to access _other_ component details from that device --- every PMbus thing already has sensor data in its `ComponentDetails` as well. But, it is kinda the standard interface for most of this sort of thing. 

Since status registers are paged in PMBus, we would want to ensure that we can make it clear to the caller which rail a status register value refers to. This could be done either by having the control plane _request_ a specific rail when reading, or by having the caller read them by device ID and reading every rail on the device, but including the rail names in the response. Doing it by rail kinda seems nicer, especially since ereports for power-related faults will include the rail name, but this would require `control-plane-agent` to maintain an additional index of devices by rail name in addition to devices by refdes, which probably eats up a bunch more flash/RAM, and means we can't just hang it off `ComponentDetails`.

Personally, I think it would also make sense to add a new [`DeviceCapabilities`](https://github.com/oxidecomputer/management-gateway-service/blob/177c9c719e12896c566a1b6b5416c9bc686531d3/gateway-messages/src/sp_to_mgs.rs#L903-L913) flag for PMBus things, to advertise that they have the standard set of PMBus status registers in addition to sensor values.

### LINKED ISSUE #2548: PMBus Status: Consider skipping known-absent status subfields
See https://github.com/oxidecomputer/hubris/pull/2538#discussion_r3319064838 for more context.

#2538 added the ability to poll PMBus `STATUS_*` registers on-demand. As of that PR, we always poll all status subfields for every PMBus device. Not all devices will support all status registers, for example, reads of `STATUS_FANS_3_4` will fail on a device that has no fans.

The cost of this is a little extra I2C traffic, both to set the PMBus paging, as well as read the absent status register.

It's feasible we could track "what status fields does this device support" in the `pmbus` crate, and present this information through code generation to the control-plane-agent code, immediately placing an `Err(Absent)` in the status report.

It is ALSO feasible we could decide to be *extra* clever and only query sub-fields if the top level `STATUS_WORD` bits are set, as if the top level bit is `0`, all bits of the sub-field should *also* be zero. In this case, we could lie, and say `Ok(0)` for the subfield, or report `Err(Skipped)` or something. At the time of working on 2538, it was decided it was worth the extra traffic to be extra sure we catch unexpected quirky-ness where "what the PMBus *should* do" and "what the PMBus *actually* does", but this suspicion is not founded on any past trauma.

For now, this is all a "nice to have", unless we find that the volume of I2C/PMBus traffic generated becomes problematic.

jamesmunns:
~~This is a (still fairly early) PR towards https://github.com/oxidecomputer/hubris/issues/2463. Opening as draft PR to solicit some feedback.~~

Closes #2463 
Closes #2548 

This PR implements an API to get PMBus Status Registers. This uses information from the `pmbus` crate to query the correct supported registers, and provides an API over MGS to be queried by the control plane.


drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +278,52 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+        Ok(PmbusStatus {
+            status_word: pmbus_rail_read!(dev, rail_idx, STATUS_WORD).map_err(drop)?.0,
+            status_vout: pmbus_rail_read!(dev, rail_idx, STATUS_VOUT).map_err(drop)?.0,
+            status_iout: pmbus_rail_read!(dev, rail_idx, STATUS_IOUT).map_err(drop)?.0,
+            status_temperature: pmbus_rail_read!(dev, rail_idx, STATUS_TEMPERATURE).map_err(drop)?.0,
+            status_cml: pmbus_rail_read!(dev, rail_idx, STATUS_CML).map_err(drop)?.0,
+            status_other: pmbus_rail_read!(dev, rail_idx, STATUS_OTHER).map_err(drop)?.0,
+            status_input: pmbus_rail_read!(dev, rail_idx, STATUS_INPUT).map_err(drop)?.0,
+            status_fans_1_2: pmbus_rail_read!(dev, rail_idx, STATUS_FANS_1_2).map_err(drop)?.0,
+            status_fans_3_4: pmbus_rail_read!(dev, rail_idx, STATUS_FANS_3_4).map_err(drop)?.0,
```
hawkw: Hm...I'd be interested to know what happens if we try to read either `STATUS_FANS` register from a device that doesn't have fans, or if we try to read `STATUS_FANS_3_4` from a device that only has 1 or 2 fans. Do we get back a read that's all zeroes, or does the device return a `NAK` or something?

§17.10 and §17.10 in the _PMBus Specification, Part II, Version 1.3.1_ doesn't explicitly say what a device without fans is supposed to do, so, if I had to guess, it's probably left up to the manufacturer...

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +278,52 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+
+            // Unfortunately, STATUS_MFR_SPECIFIC *is* defined in the pmbus crate, but doesn't have a
+            // "structured" representation, so instead use a raw representation. It *could* be argued
+            // that since we're not actually peeking at any of the introspection stuff, we could do
+            // the same for all the items above, and save ourselved a little indirection, but for now
+            // just hole-punch the minimum amount necessary
+            status_mfr_specific: pmbus_rail_read!(@raw => dev, rail_idx, CommandCode::STATUS_MFR_SPECIFIC as u8, 1).map_err(drop)?[0],
+        })
```
hawkw, line 404: I don't love that this will bail out if _any_ read returns an error. I think it would be much better to change the `PmbusStatus` type so that each register's field was a `Result<u8, i2c::ResponseCode>` or something. That way, we can always communicate whatever registers we were able to successfully read to upstack software even if we encounter some weird I2C bus weather or something. Also, if (as I mentioned in my other comment) a device without fans _does_ choose to NAK one of the `STATUS_FANS` registers, that doesn't result in us being permanently unable to read its status.

drv/i2c-devices/src/lib.rs
hawkw: Attempting to find out what happens when one does this experimentally, and I have discovered that, using the `humility pmbus` command to ask a RAA229620A for its `STATUS_FANS_1_2`, it just...prints nothing:

```console
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r VDDCR_CPU0_A0 -C STATUS_FANS_1_2
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
eliza@alfred ~ $
```

I'm not actually sure what "nothing" means here!

hawkw: The verbose mode also makes it say "nothing":

```console
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r VDDCR_CPU0_A0 -C STATUS_FANS_1_2 --verbose
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
eliza@alfred ~ $
```

hawkw: OH, you have to pass the `--errors` flag to `humility pmbus` to make it show errors. So the RAA229620A returns `Noregister` when you ask for its `STATUS_FANS_1_2` or `STATUS_FANS_3_4`:

```console
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r VDDCR_CPU0_A0 -C STATUS_FANS_1_2,STATUS_FANS_3_4 --verbose --errors
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
0x81 STATUS_FANS_1_2           Err(NoRegister)
0x82 STATUS_FANS_3_4           Err(NoRegister)
eliza@alfred ~ $
```

The ISL68224, ADM127X, TPS546B24A also do the same when asked for their `STATUS_FANS` registers, while the BMR491 appears to...still do something that results in Humility printing nothing?

```console
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r VDDCR_CPU0_A0 -C STATUS_FANS_1_2 --verbose --errors
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
0x81 STATUS_FANS_1_2           Err(NoRegister)
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r VDDCR_CPU0_A0 -C STATUS_FANS_1_2,STATUS_FANS_3_4 --verbose --errors
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
0x81 STATUS_FANS_1_2           Err(NoRegister)
0x82 STATUS_FANS_3_4           Err(NoRegister)
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r V1P8_SP5_A0 -C STATUS_FANS_1_2,STATUS_FANS_3_4 --verbose --errors
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
humility pmbus failed: rail V1P8_SP5_A0 not found
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r V1P8_SP5_A1 -C STATUS_FANS_1_2,STATUS_FANS_3_4 --verbose --errors
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
0x81 STATUS_FANS_1_2           Err(NoRegister)
0x82 STATUS_FANS_3_4           Err(NoRegister)
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r V54P5_IBC_A3 -C STATUS_FANS_1_2,STATUS_FANS_3_4 --verbose --errors
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
0x81 STATUS_FANS_1_2           Err(NoRegister)
0x82 STATUS_FANS_3_4           Err(NoRegister)
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r V3P3_SP_A2 -C STATUS_FANS_1_2,STATUS_FANS_3_4 --verbose --errors
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
0x81 STATUS_FANS_1_2           Err(NoRegister)
0x82 STATUS_FANS_3_4           Err(NoRegister)
eliza@alfred ~ $ pfexec /staff/eliza/humility -t cosmo-hubris-sp pmbus -r V12_SYS_A2 -C STATUS_FANS_1_2,STATUS_FANS_3_4 --verbose --errors
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP
eliza@alfred ~ $
```

This is sort of what I had figured, and suggests that we need to either have some additional configuration option that says whether or not to read the fans, and/or (and this would be my preference) change `PmbusStatus::read_from` to not bail out if any read returns an error, as I suggested in https://github.com/oxidecomputer/hubris/pull/2538#discussion_r3313107539.

task/validate-api/build.rs
```diff
@@ -5,6 +5,11 @@
 use std::io::Write;
 
 fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    if let Err(e) = build_i2c::codegen(build_i2c::Disposition::Devices) {
+        println!("cargo::error=failed to generate I2C devices: {e}");
+        std::process::exit(1);
+    }
```
hawkw: Soooo, I realize that in my VPD PR, I added all the code generation to `task-validate-api`, but I'm not sure if we need to do that here. In the VPD branch (which I need to resurrect), it was necessary to add the VPD-reading code to `task-validate-api`'s codegen, because `validate` generates the list of `DeviceDescriptions`, and I wanted to add device VPD to the `component_details` MGS API, which is driven by the list of `DeviceDescriptions` from `task-validate-api`.

Since we're looking up devices for reading PMBus status using a _new_ lookup table of rail names to devices, I think there isn't actually any reason that that _needs_ to go in `task-validate-api`. I think it might be nicer to just put that in `control-plane-agent`'s build script, or make it a `Disposition` for `build_i2c::codegen` or something. That way, the validate API crate need not actually inflict all the I2C codegen on _every_ task that depends on it. What do you think?

task/validate-api/build.rs
```diff
@@ -114,6 +129,39 @@ fn write_pub_device_descriptions() -> anyhow::Result<()> {
     }
     writeln!(file, "];")?;
 
+    let max_len = pmbus_rail_names.iter().map(String::len).max();
+    if let Some(_len) = max_len {
+
+        // TODO: do we need fixed-length bstrings? If we want to binary search, we probably want them to
+        // be truncated to some maximum length, so we either need to define the max length at the
+        // protocol level so mgs knows how long of a string to send us, or we could instead trim the
+        // trailing nulls and search by that instead.
```
hawkw: i suspect the lookup code is going to end up looking something like the code in `control-plane-agent` for looking up a component by ID: https://github.com/oxidecomputer/hubris/blob/96975a6b6314df88324acac6be76f688cdd8c703/task/control-plane-agent/src/inventory.rs#L177-L196

as for maximum length, i think that the wire message will certainly have to have one, a la `SpComponent` in `gateway-messages`: https://github.com/oxidecomputer/management-gateway-service/blob/6c0aca2545a73fd75536e149d29faa7108be5862/gateway-messages/src/lib.rs#L284-L294 

I really don't love the null-padding, but even if we were to adopt an encoding that let us not do that part, we would still need to have some max length _somewhere_ so we can limit the buffer size...

task/validate-api/build.rs
jamesmunns: I think that sounds reasonable! I put it in `task-validate-api` because I was copying your VPD PR, but I admit to wondering "why does it need to be here/like this?".

I'll move it over to control plane agent, or as a new codegen path.

task/validate-api/build.rs
jamesmunns: Yeah, I think the distinction that I wanted to make (or at least still need to shake out) is that just because it needs to be zero-padded *on the wire*, that doesn't necessarily mean that internally we ALSO need to zero-pad it.

When receiving the command over the wire, we could first trim any trailing nulls, and then use the slice locally instead. That being said, if I bump into anywhere else where we need fixed length arrays, I'll probably revert back to zero-padding in the codegen.

drv/i2c-devices/src/lib.rs
jamesmunns: Will do! I wondered whether to make these `Option`s or `Result`s, and I should definitely do one of those.

I wonder if we could capture "this device supports these statuses" in the `pmbus` crate and maybe codegen based on this so we don't spend time doing I2C we know is going to be NAK'd, but this could also be a good follow-up.

Since we're not using device-specific methods here, I would imagine we would want to pass in some kind of bit-flag of which ones to check or not, and just have a `None` or `Err(Unsupported)`.

hawkw: > I wonder if we could capture "this device supports these statuses" in the `pmbus` crate and maybe codegen based on this so we don't spend time doing I2C we know is going to be NAK'd, but this could also be a good follow-up.

Yeah, we could either do this in the `pmbus` crate or in the codegen itself. Since we have a string saying which device it is, the codegen could have its own understanding of which registers to skip for certain devices. It probably should fall back to querying all registers for device strings it doesn't recognize, maybe along with a build warning.

I do worry _slightly_ about a device firmware revision changing its behavior to something that doesn't match the codegen's understanding, but that feels pretty unlikely. For instance, an RAA229620A VRM is not going to suddenly grow fans if its firmware is updated!

---

Incidentally, there _is_ another option, which is the PMBus `QUERY` command (§11.13 _PMBus Specification, Part II, Version 1.3.1_). This allows you to _ask_ a device whether it supports a particular PMBus command, such as `STATUS_FANS_1_2` and get back a response that tells you whether the device supports it. I'm not sure how useful this is to us here, though --- if the goal is mostly to avoid I2C traffic for unsupported registers, using `QUERY` to ask "hey, do you have `STATUS_FANS_1_2`? okay, do you have `STATUS_FANS_3_4`?" results in just as much I2C traffic if the device doesn't have those registers, and *more* if it does. We could probably avoid that by `QUERY`ing once and caching the result somewhere, but the current code is not structured in a way that gives us a place to stash that, so it would be a bit more complex. 

At the end of the day, I just don't really think that's particularly worthwhile: since we know all the devices on the board at build time, we can just decide which registers to read when we do the code generation, and we don't have to dynamically detect capabilities. I think that `QUERY` is more intended for the use case of commodity motherboards which might end up talking to any arbitrary commodity PSUs, where the firmware doesn't know what PMBus commands are supported until the system boots up, and therefore must dynamically detect them. But, I felt like it was worth mentioning, if only to write down why we probably _don't_ need to use it.

task/validate-api/build.rs
hawkw: I think avoiding zero-padding here is probably nicer if you can manage it --- it should save us a bit of flash in the lookup table even if it doesn't save us the stack space when we're passing around the wire message.

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +278,69 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+    BadData { cmd: u8, },
+}
+
+impl PmbusStatus {
+    /// Attempt to read a [`PmbusStatus`] from the given device and rail.
+    ///
+    /// This function returns an error if the initial attempt to obtain `STATUS_WORD` from the
+    /// device fails, otherwise returnining successfully even if "leaf" status bytes were unable
+    /// to be read, either due to ephemeral hiccups, or that status byte being unsupported by
+    /// the device queried.
```
hawkw: I'm not *entirely* convinced this behavior for `STATUS_WORD` is correct. If you look at the bits in `STATUS_WORD`, you'll see that it's basically a summary of which other registers have bits set --- there's a bit in `STATUS_WORD` that is set if any fault bit in `STATUS_IOUT`, one for faults in `STATUS_VOUT`, and so on. So, we actually _can_ reasonably interpret all the other registers even if we're missing `STATUS_WORD`.

That said, I think it's also reasonable-ish to fail fast if we get an error that indicates there is no device to talk to at all; if we get a `NAK` that indicates its not present, we don't need to spend more time doing all the other I2C reads. I dunno.

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,64 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    writeln!(file, "pub const PMBUS_RAIL_TO_I2C_DEVICE_MAP: [(&[u8], fn(userlib::TaskId) -> (drv_i2c_api::I2cDevice, u8)); {}] = [", pmbus_rail_names.len())?;
+    for rail in pmbus_rail_names.iter() {
+        write!(file, "    (b\"{rail}\", ")?;
+        // build_i2c *also* only to-lowercases the rail names to make functions
+        write!(file, "crate::i2c_config::pmbus::{}", rail.to_lowercase())?;
+        writeln!(file, "),")?;
+    }
+    writeln!(file, "];")?;
+
+    assert_eq!(pmbus_rail_dupes, 0);
```
hawkw: nitpick: it would be nice if this had a panic message that said something to the effect of "your config is invalid"...also, it's maybe worth checking whether panicking or returning an `anyhow::Error` here results in a nicer UX for error reporting in build scripts?

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,64 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    let mut pmbus_rail_names = std::collections::BTreeSet::new();
+    let mut pmbus_rail_dupes = 0;
+    for dev in devices {
+        // We only need to map PMBus devices
+        let Some(ref pmbus) = dev.pmbus else {
+            continue;
+        };
+
+        // pmbus devices must have a refdes
+        assert!(dev.device_id.is_some());
```
hawkw: do we actually care about this here?

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,64 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+
+        // Aggregate a list of all PMBus-visible rails
+        for rail in pmbus.rails.iter() {
+            // `BTreeSet::insert` return value means "is unique", which is the inverse of
+            // `BTreeMap::insert().is_some()`!
+            if !pmbus_rail_names.insert(rail.name.clone()) {
+                pmbus_rail_dupes += 1;
+                println!("cargo::error=Duplicate Rail name: {:?}", rail.name);
```
hawkw: nit: this error message is what the programmer will see if they have made an error in the `app.toml`. i think we can probably improve the user experience here a bit, at least by making the following changes:

- downcasing the first letter to match other cargo error messages
- including the name of the PMBus device that has the duplicate rail.
- it would probably be nice if we could _also_ say the refdes of the other device that has that rail, so we could say "both U420 and U69 define a power rail called V69P7_TURBOENCABULATOR_A2, your config file is invalid", but that would require a bigger change to also track which device a rail name in the map came from...
 
```suggestion
                println!("cargo::error=PMBus device {} defines a power rail {:?} which already exists in the manifest", dev.device_id, rail.name);
```
or something like that?

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,64 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+            }
+        }
+    }
+
+    // Create a mapping between rail names and generated accessor functions for obtaining
+    // the device handle and rail index
+    writeln!(file)?;
+    writeln!(file, "pub const PMBUS_RAIL_TO_I2C_DEVICE_MAP: [(&[u8], fn(userlib::TaskId) -> (drv_i2c_api::I2cDevice, u8)); {}] = [", pmbus_rail_names.len())?;
```
hawkw: nit, take it or leave it: part of me wants to suggest that this be a struct rather than a tuple, in the hopes that the non-generated code consuming the LUT is a bit more readable. but i'm not sure if this is worth the ffort or not. up to you.

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,64 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    for rail in pmbus_rail_names.iter() {
+        write!(file, "    (b\"{rail}\", ")?;
+        // build_i2c *also* only to-lowercases the rail names to make functions
+        write!(file, "crate::i2c_config::pmbus::{}", rail.to_lowercase())?;
```
hawkw: eventually, if we wanted to be fancy, we could probably change up `build-i2c` so that it actually _provides_ the name of the `i2c_config::pmbus` function for looking up that device, instead of relying on that code not changing. but also, that code is not going to change without breaking a bunch of other stuff, so 🤷‍♀️ 

could be worth adding to the backlog of "non-important non-blockers"...

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,64 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    let dest_path =
+        std::path::Path::new(&out_dir).join("pmbus_mapping.rs");
+    let file = std::fs::File::create(&dest_path)?;
+    let mut file = std::io::BufWriter::new(file);
+
+    // Generate the necessary rail names
+    if let Err(e) = build_i2c::codegen(build_i2c::Disposition::Devices) {
+        println!("cargo::error=failed to generate I2C devices: {e}");
+        std::process::exit(1);
+    }
```
hawkw: nitpicky: i might put this part in `main` rather than in the pmbus-specific codegen? we will want it if other stuff also starts wanting i2c config in control-plane-agent...

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,64 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+
+    let devices = build_i2c::device_descriptions()
+        .collect::<Vec<_>>();
+
+    let mut pmbus_rail_names = std::collections::BTreeSet::new();
+    let mut pmbus_rail_dupes = 0;
+    for dev in devices {
```
hawkw: nit: since all we do with `devices` is iterate over it, and `build_i2c::device_descriptions()` returns an iterator, i think the `collect` is unnecessary:
```suggestion
    let devices = build_i2c::device_descriptions();

    let mut pmbus_rail_names = std::collections::BTreeSet::new();
    let mut pmbus_rail_dupes = 0;
    for dev in devices {
```

or even just:

```suggestion
    let mut pmbus_rail_names = std::collections::BTreeSet::new();
    let mut pmbus_rail_dupes = 0;
    for dev in build_i2c::device_descriptions() {
```

i doubt the build time hit for iterating twice and allocating here is all that noticeable, but...every nanosecond counts, i guess?

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,64 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+
+    do_pmbus()?;
+
+    Ok(())
+}
+
+fn do_pmbus() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let out_dir = std::env::var("OUT_DIR")?;
+    let dest_path =
+        std::path::Path::new(&out_dir).join("pmbus_mapping.rs");
```
hawkw: unimportant turbo-nit: `std::path::PathBuf` is imported at the top of the file, so perhaps we should just change that import to `use std::path::{PathBuf, Path}` and change this to 
```suggestion
        Path::new(&out_dir).join("pmbus_mapping.rs");
```

build/i2c/src/lib.rs
```diff
@@ -1381,6 +1381,8 @@ impl ConfigGenerator {
             all.sort();
 
             for (rail, (device, index)) in &all {
+                // if we update this code to be more clever than just to-lowercase'ing the rail names,
+                // you might need to go update the mapping in `validate-api`!
```
hawkw: nit: this is now in the `control-plane-agent` build script...

hawkw (commented):
This is looking good --- I'm looking forward to seeing the control-plane-agent code that actually consumes it. I had some smallish suggestions on the codegen.

drv/i2c-devices/src/lib.rs
jamesmunns: My assumption here is that IF a device supports getting PMBus status, then it MUST support querying `STATUS_WORD`. I need to actually read the spec to decide if this is true or not.

I think your point about "fail fast" is reasonable, but I was actually mostly hoping to just reduce the "state space" for MGS: rather than having to handle `Ok({ all fields err })`, AND `Err(failed)`, I just compress the former into the latter. I can make the first field failable, especially if you think there could be a case where STATUS_WORD would fail, but the "child" fields would still be present.

Even in that case though, I think UX wise, we should probably flatten `Ok({ all fields err })` to `Err(failed)`, just because it's a little silly in a "successfully failed" kind of way if we don't.

task/control-plane-agent/build.rs
jamesmunns: So, I'll change this to a `bail`, but this assert is actually (currently) impossible to hit, we guard against dupes in the i2c generation phase:

https://github.com/oxidecomputer/hubris/blob/8f5e0f7c2417aa70a341120cc02c4ceb581156af/build/i2c/src/lib.rs#L1358-L1360

Which means we'll fail earlier in this function on generation, rather than here.

task/control-plane-agent/build.rs
jamesmunns: Nope, this is leftover copy-paste. will remove.

task/control-plane-agent/build.rs
jamesmunns: Will fix

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +278,85 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+        rail_idx: u8,
+    ) -> Result<Self, PmbusStatusError> {
+        use PmbusStatusError as Error;
+        use pmbus::commands::*;
+
+        Ok(PmbusStatus {
+            // Status word *must* succeed, otherwise we don't have reasonable data to return.
+            // We may want to consider making some/all of these retryable, but for now you either
+            // get them or you don't.
+            status_word: pmbus_rail_read!(dev, rail_idx, STATUS_WORD)?.0,
```
jamesmunns: @hawkw do you have opinions about whether it's "worth it" to round trip through the actual structured pmbus machinery if I'm just going to immediately shuck the `u8` back out of it? I could replace all of these `pmbus_rail_read`s to use the `@raw =>` variant I added for `MFR_SPECIFIC` below, and probably spare a little code space and cpu time.

I originally used the existing method for consistency reasons, but it does seem a little silly to me when I step back.

drv/i2c-devices/src/lib.rs
hawkw: > My assumption here is that IF a device supports getting PMBus status, then it MUST support querying `STATUS_WORD`. I need to actually read the spec to decide if this is true or not.

I think this is a safe assumption. I was thinking more about transient I2C glitchiness here, rather than "device doesn't support STATUS_WORD".

drv/i2c-devices/src/lib.rs
hawkw: Hm, I think that it's probably better to avoid reading the structured registers and then ignore the result of doing any of that, personally. The only real benefit of not doing it that way is that we don't get validation of the register value, but...this one is just bitflags, and I'd honestly prefer to just report whatever bits the device said and let the upstack software sort it out.

build/i2c/src/lib.rs
```diff
@@ -1381,6 +1381,8 @@ impl ConfigGenerator {
             all.sort();
 
             for (rail, (device, index)) in &all {
+                // if we update this code to be more clever than just to-lowercase'ing the rail names,
+                // you might need to go update the mapping in `control-plane-agent`!
                 write!(
```
hawkw, line 1395: nit: mind wrapping at column 80?

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +278,85 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+    BadData { cmd: u8 },
+}
+
+impl PmbusStatus {
+    /// Attempt to read a [`PmbusStatus`] from the given device and rail.
+    ///
+    /// This function returns an error if the initial attempt to obtain `STATUS_WORD` from the
+    /// device fails, otherwise returnining successfully even if "leaf" status bytes were unable
+    /// to be read, either due to ephemeral hiccups, or that status byte being unsupported by
+    /// the device queried.
```
hawkw: nit: can we wrap this?

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +278,85 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+        dev: &I2cDevice,
+        rail_idx: u8,
+    ) -> Result<Self, PmbusStatusError> {
+        use PmbusStatusError as Error;
+        use pmbus::commands::*;
+
+        Ok(PmbusStatus {
+            // Status word *must* succeed, otherwise we don't have reasonable data to return.
+            // We may want to consider making some/all of these retryable, but for now you either
+            // get them or you don't.
```
hawkw: nit: can we wrap this?

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +278,85 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+                .map(|v| v.0),
+            status_fans_3_4: pmbus_rail_read!(dev, rail_idx, STATUS_FANS_3_4)
+                .map(|v| v.0),
+
+            // Unfortunately, STATUS_MFR_SPECIFIC *is* defined in the pmbus crate, but doesn't have a
+            // "structured" representation, so instead use a raw accessor. It *could* be argued
+            // that since we're not actually peeking at any of the introspection stuff, we could do
+            // the same for all the items above, and save ourselves a little indirection, but for now
+            // just hole-punch the minimum amount necessary. Indexing is fine here because we know
+            // (statically) the returned value is `[u8; 1]`.
```
hawkw: wrapmeplz sorry

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,85 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+        };
+
+        // Aggregate a list of all PMBus-visible rails
+        for rail in pmbus.rails.iter() {
+            // `BTreeSet::insert` return value means "is unique", which is the inverse of
+            // `BTreeMap::insert().is_some()`!
+            if !pmbus_rail_names.insert(rail.name.clone()) {
+                pmbus_rail_dupes += 1;
+                println!(
+                    "cargo::error=PMBus device {} defines a power rail {:?} which already exists in the manifest",
```
hawkw: wrapme plz

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,85 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+
+    do_pmbus()?;
+
+    Ok(())
+}
+
+fn do_pmbus() -> Result<(), anyhow::Error> {
+    let out_dir = std::env::var("OUT_DIR")?;
+    let dest_path = Path::new(&out_dir).join("pmbus_mapping.rs");
+    let file = std::fs::File::create(&dest_path)?;
```
hawkw: maybe worth including the path to the file we were trying to create here in an anyhow `context`, in case this doesn't work?

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,85 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+                    "cargo::error=PMBus device {} defines a power rail {:?} which already exists in the manifest",
+                    dev.device_id.as_deref().unwrap_or("(no ID)"),
+                    rail.name
+                );
+            }
+        }
+    }
+
+    // Create a mapping between rail names and generated accessor functions for obtaining
+    // the device handle and rail index
```
hawkw: wrap

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,85 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+            rail.to_lowercase()
+        )?;
+        writeln!(file, "}},")?;
+    }
+    writeln!(file, "];")?;
+
+    // This is supposed to be caught during I2C generation
+    if pmbus_rail_dupes != 0 {
+        bail!(
+            "duplicate PMBus rails: invalid application toml. This is a bug in build_i2c."
```
hawkw: wrap me too please

task/control-plane-agent/build.rs
```diff
@@ -35,6 +35,85 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    writeln!(file, "/// Type returned by generated pmbus rail functions")?;
+    writeln!(
+        file,
+        "pub type SummonFn = fn(userlib::TaskId) -> (drv_i2c_api::I2cDevice, u8);"
+    )?;
+    writeln!(file)?;
+    writeln!(file, "pub struct PmbusRailBinding {{")?;
+    writeln!(file, "    pub name: &'static str,")?;
+    writeln!(file, "    pub summon_fn: SummonFn,")?;
+    writeln!(file, "}}")?;
```
hawkw: FWIW, we don't _really_ have to generate the types in the codegen. Instead, we could just have the file where the codegen is imported (which currently looks like `control-plane-agent`'s `main.rs`) do this:

```rust
/// Type returned by generated pmbus rail functions
pub type SummonFn = fn(userlib::TaskId) -> (drv_i2c_api::I2cDevice, u8);

pub struct PmbusRailBinding {
    pub name: &'static str,
    pub summon_fn: SummonFn,
}

include!(concat!(env!("OUT_DIR"), "/pmbus_mapping.rs"));
```

Personally, I'd kinda prefer we do this, because I'd rather be able to see the type definition in the rust code, along with the comment that the generated code is just an array of these. That way it's easy to see what the fields on the type are when I read the code.

I might also stuff all of that stuff in a submodule but it's up to you.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -746,6 +747,64 @@ impl MgsCommon {
...
+                drv_i2c_devices::PmbusStatusError::BadRead { cmd: _, code } => {
+                    PmbusStatusReadError::DriverReadFailed {
+                        response_code: code as u8,
+                    }
+                }
+                drv_i2c_devices::PmbusStatusError::BadData { cmd: _ } => {
+                    PmbusStatusReadError::InvalidBusData
+                }
+            }
+        }
```
hawkw, line 786: I believe that if we just do raw reads for everything (as discussed in https://github.com/oxidecomputer/hubris/pull/2538/changes#r3350341678), this all goes away (since it's basically just necessary to handle `BadData`, which the raw read shouldn't return)?

app/minibar/rev-b.toml
```diff
@@ -105,4 +106,3 @@ owner = {name = "control_plane_agent", notification = "socket"}
 port = 11111
 tx = { packets = 3, bytes = 2048 }
 rx = { packets = 3, bytes = 2048 }
-
```
hawkw, line 108: spurious whitespace change?

hawkw (commented):
Sorry for leaving a bunch of obnoxious line wrapping nits, but beyond that, I think this is going in the right direction. Sadly, we try to wrap everything at the 80th column, but rustfmt cannot automatically do this for strings or comments unless you enable a bunch of nightly features. So wrapping comments and strings manually is just a habit one kinda gets used to.

I think the error munging gets a lot simpler if we change this to just do raw reads, as you discussed. Besides that, looking good!

app/minibar/rev-b.toml
jamesmunns: My editor automatically trims duplicate trailing newlines (and adds one if missing) on save, there were two here before. I can revert it if we liked the extra newlines.

drv/i2c-devices/src/lib.rs
jamesmunns: Leaving it as "todo", I'll open up a follow-up issue.

jamesmunns: Opened https://github.com/oxidecomputer/hubris/issues/2548 to track this.

task/control-plane-agent/build.rs
jamesmunns: Added a helper function for this.

jamesmunns:
This was discussed with @hawkw and @rmustacc today, it's looking like https://github.com/oxidecomputer/hubris/issues/2548 is moving towards "not optional", particularly because querying status fields that *don't* exist seems to be particularly problematic for some devices.

Some of them just NAK the register read (code 7):

```
jamesmunns@castle ~ $ ./faux-mgs-pmbus-please --interface berlin_sw0tp0 --discovery-addr [fe80::aa40:25ff:fe04:482]:11111 get-pmbus-status V3P3_SP_A2
Jun 04 17:56:19.712 INFO creating SP handle on interface berlin_sw0tp0, component: faux-mgs
Jun 04 17:56:19.714 INFO initial discovery complete, addr: [fe80::aa40:25ff:fe04:482%12]:11111, interface: berlin_sw0tp0, socket: control-plane-agent, component: faux-mgs
Rail 'V3P3_SP_A2':
WORD          0x0202
VOUT          0x00
IOUT          0x00
TEMPERATURE   0x00
CML           0x82
OTHER         0x01
INPUT         0x00
MFR_SPECIFIC  0x00
FANS_1_2      Err(DriverReadFailed { response_code: 7 })
FANS_3_4      Err(DriverReadFailed { response_code: 7 })
```

Some of them get *really* mad (code 18 => `BusLocked`, "I2C bus locked up and was reset"):

```
jamesmunns@castle ~ $ ./faux-mgs-pmbus-please --interface berlin_sw0tp0 --discovery-addr [fe80::aa40:25ff:fe04:482]:11111 get-pmbus-status V12_SYS_A2
Jun 04 18:09:56.805 INFO creating SP handle on interface berlin_sw0tp0, component: faux-mgs
Jun 04 18:09:56.807 INFO initial discovery complete, addr: [fe80::aa40:25ff:fe04:482%12]:11111, interface: berlin_sw0tp0, socket: control-plane-agent, component: faux-mgs
Rail 'V12_SYS_A2':
WORD          0x0002
VOUT          0x00
IOUT          0x00
TEMPERATURE   0x00
CML           0x82
OTHER         0x00
INPUT         0x00
MFR_SPECIFIC  0x00
FANS_1_2      Err(DriverReadFailed { response_code: 18 })
FANS_3_4      Err(DriverReadFailed { response_code: 18 })
```

For both of these cases, you can also note that the CML error has been set, which is regrettable.

I will be looking into whether I can use the `pmbus` crate (and its metadata) to tell us whether we should be asking a given device for a status subfield. I'm going to look into doing this tomorrow.

jamesmunns:
Also speaking with @hawkw, https://github.com/oxidecomputer/hubris/pull/2538/commits/582b456c545b60a8fab72301d294427a56b57837 (make two versions of the rail-getters) doesn't look like the preferred solution. She suggested either going back to "keep the 'single rail' info separate" like https://github.com/oxidecomputer/hubris/pull/2538/commits/1006fea27e93a2811dfcb92267d1671bc3f3c96c, or moving *everything* to the new getter that returns an `Option<u8>`.

This might be a moot point, if we can pull this "does support paging" info from the `pmbus` crate. Otherwise I might go back to looking at one of the other two solutions.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -936,3 +1000,35 @@ impl From<SpImageError> for MgsImageError {
...
+        ResponseCode::BusResetMux => true,
+        ResponseCode::BusLocked => true,
+        ResponseCode::BusLockedMux => true,
+        ResponseCode::ControllerBusy => true,
+        ResponseCode::BusError => false,
+        ResponseCode::BadDeviceState => false,
+        ResponseCode::OperationNotSupported => false,
+        ResponseCode::IllegalLeaseCount => false,
+        ResponseCode::TooMuchData => false,
+    }
```
hawkw: I feel like the match that maps each variant to a bool is a little bit hard to read, IMO, since the `true` and `false` branches are mixed together. Personally, I might prefer to express it more like this:


```suggestion
    match code {
        // Non-retryable errors
        ResponseCode::BadResponse
        | ResponseCode::BadArg
        | ResponseCode::NoDevice
        | ResponseCode::BadController
        | ResponseCode::ReservedAddress
        | ResponseCode::BadPort
        | ResponseCode::NoRegister
        | ResponseCode::BadMux
        | ResponseCode::BadSegment
        | ResponseCode::MuxNotFound
        | ResponseCode::SegmentNotFound
        | ResponseCode::SegmentDisconnected
        | ResponseCode::MuxDisconnected
        | ResponseCode::MuxMissing 
        | ResponseCode::BadMuxRegister
        | ResponseCode::BusError
        | ResponseCode::BadDeviceState
        | ResponseCode::OperationNotSupported
        | ResponseCode::IllegalLeaseCount
        | ResponseCode::TooMuchData => false,
        // Retryable errors
        ResponseCode::BusReset
        | ResponseCode::BusResetMux
        | ResponseCode::BusLocked
        | ResponseCode::BusLockedMux
        | ResponseCode::ControllerBusy => true,
    }
```

This way, there's two separate lists of retryable and non-retryable errors, which I think is easier to understand.

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -936,3 +1000,35 @@ impl From<SpImageError> for MgsImageError {
...
+        ResponseCode::SegmentDisconnected => false,
+        ResponseCode::MuxDisconnected => false,
+        ResponseCode::MuxMissing => false,
+        ResponseCode::BadMuxRegister => false,
+        ResponseCode::BusReset => true,
+        ResponseCode::BusResetMux => true,
+        ResponseCode::BusLocked => true,
+        ResponseCode::BusLockedMux => true,
+        ResponseCode::ControllerBusy => true,
+        ResponseCode::BusError => false,
```
hawkw: it's probably worth double-checking the semantics of `BusError` --- this _might_ belong on the retryable list?

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -936,3 +1000,35 @@ impl From<SpImageError> for MgsImageError {
         })
     }
 }
+
+fn retry_hint(code: drv_i2c_api::ResponseCode) -> bool {
+    use drv_i2c_api::ResponseCode;
```
hawkw: This is probably best left to a separate PR, but I wonder if it's worth factoring this out and putting it someplace where we can use it elsewhere --- perhaps making it an `is_retryable()` method on `ResponseCode`. There are a bunch of places in hubris where we unconditionally retry all I2C errors, including ones which should always be fatal such as as nonexistent bus/segment/controller...

task/control-plane-agent/src/mgs_common.rs
jamesmunns: fwiw, this is just "I expanded the match and still hadn't figured out what was good here", I will make sure to clean it up before it gets merged!

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +284,113 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+            ),
+            status_fans_3_4: get_byte(
+                Cc::STATUS_FANS_3_4,
+                Capabilities::STATUS_FANS_3_4,
+            ),
+            status_mfr_specific: get_byte(
+                Cc::STATUS_MFR_SPECIFIC,
+                Capabilities::STATUS_MFR_SPECIFIC,
+            ),
+        })
```
hawkw, line 404: So...well, I don't completely hate the bitmap thing, but I do wonder about an alternative approach. This approach generates a bitmask per device type at build time, but is ultimately deciding whether to read each register at runtime. I wonder a bit if it would be nicer to not do that. I know that back when we were planning to unconditionally read every register, I told you to just have one function for any `I2cDevice`, but now, I kind of wonder if it would be better to generate a function for each device that reads the status registers we believe that device supports. Then,`control-plane-agent` declare a type like

```rust
pub type SummonFn = fn(userlib::TaskId) -> (drv_i2c_api::I2cDevice, u8);
pub type ReadFn = fn(drv_i2c_api::I2cDevice, u8) -> Result<PmbusStatus, PmbusStatusReadResult>;

pub struct PmbusRailBinding {
    pub name: &'static str,
    pub summon_fn: SummonFn,
    pub read_fn: ReadFn,
}
```

and the codegen could use the device name in the TOML to generate bindings like, say, 

```rust
// ... 
PmbusRailBinding { 
    name: "VDDCR_CPU0_A0",
    summon_fn: crate::i2c_config::pmbus::vddcr_cpu0_a0,
    read_fn: drv_i2c_devices::raa229620a::read_pmbus_status,
},
// ...
```
(assuming that there's a device-specific `read_pmbus_status` function in the `raa229620a` module --- that would imply you hand-wrote it, but it also could be auto-generated in `drv_i2c_devices` probably?)

this also lets us get rid of the `Option<u8>` thing, since the `read_pmbus_status` function for each device could decide whether or not the page needs to be sent.

task/control-plane-agent/build.rs
```diff
@@ -27,29 +29,111 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+
+        // If it's a pmbus device, we need to get it's status capabilities
+        let Some(func) = pmbus_status_caps.get(dev.device.as_str()) else {
+            panic!("Unknown device: {}", dev.device);
+        };
+        let caps = (func)();
+
+        // Aggregate a list of all PMBus-visible rails
+        for rail in pmbus.rails.iter() {
+            // `BTreeSet::insert` return value means "is unique", which is the
```
jamesmunns: Stale comment

task/control-plane-agent/build.rs
```diff
@@ -27,29 +29,111 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+
+    let mut pmbus_rail_names = std::collections::BTreeMap::new();
+    let mut pmbus_rail_dupes = 0;
+    for dev in build_i2c::device_descriptions() {
+        // We only need to map PMBus devices
+        let Some(ref pmbus) = dev.pmbus else {
+            continue;
+        };
+
+        // If it's a pmbus device, we need to get it's status capabilities
```
jamesmunns:
```suggestion
        // If it is a pmbus device, we need to get its status capabilities
```

build/i2c/src/lib.rs
```diff
@@ -1381,6 +1384,14 @@ impl ConfigGenerator {
             all.sort();
 
             for (rail, (device, index)) in &all {
+                let raw_bank = index.unwrap_or(0);
+
+                // ----
+                //
```
jamesmunns:
```suggestion
                // First accessor, returns `(I2cDevice, u8)`
```

build/i2c/src/lib.rs
```diff
@@ -1390,12 +1401,33 @@ impl ConfigGenerator {
                 )?;
 
                 let out = self.generate_device(device, 16);
-                writeln!(&mut self.output, "({out}, {index})\n        }}")?;
+                writeln!(&mut self.output, "({out}, {raw_bank})\n        }}")?;
+
+                // ---
+                //
```
jamesmunns:
```suggestion
                // Second accessor, returns `(I2cDevice, Option<u8>)`
```

task/control-plane-agent/src/main.rs
```diff
@@ -627,3 +629,16 @@ mod idl {
 }
 
 include!(concat!(env!("OUT_DIR"), "/notifications.rs"));
+
+include!(concat!(env!("OUT_DIR"), "/i2c_config.rs"));
+
+/// Type returned by generated pmbus rail functions")?;
```
jamesmunns:
```suggestion
/// Type returned by generated pmbus rail functions
```

task/control-plane-agent/build.rs
```diff
@@ -27,29 +29,111 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
-fn write_keys(
-    cfg: Config,
-) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+fn do_pmbus() -> Result<()> {
+    let out_dir = std::env::var("OUT_DIR")?;
+    let dest_path = Path::new(&out_dir).join("pmbus_mapping.rs");
+    let out = context_create_file(&dest_path)?;
+    let mut file = std::io::BufWriter::new(out);
+
+    let pmbus_status_caps = {
```
jamesmunns, line 59:
```suggestion
    // Build a mapping from "pmbus device name" to "supported status regs"
    let pmbus_status_caps = {
```

task/control-plane-agent/build.rs
```diff
@@ -27,29 +29,111 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+        let mut map = HashMap::new();
+        for (name, func) in PMBUS_GENERATOR {
+            map.insert(*name, *func);
```
jamesmunns: todo: I could just call `(func)()` here and stick the `Capabilities` into the hashmap instead of calling the function for each device.

task/control-plane-agent/build.rs
```diff
@@ -90,3 +174,54 @@ fn write_keys(
...
     Ok(())
 }
+
+/// Create a file with anyhow context
+fn context_create_file(path: &Path) -> Result<File> {
+    File::create(path)
+        .with_context(|| format!("failed to create file '{}'", path.display()))
+}
+
+macro_rules! bitmaker {
```
jamesmunns:
```suggestion
/// Look at the `pmbus` crate metadata to see if a specific command is "Illegal"
/// and set the capability bit if not.
macro_rules! bitmaker {
```

task/control-plane-agent/build.rs
```diff
@@ -90,3 +174,54 @@ fn write_keys(
...
+        use pmbus::{Command, Operation};
+        if pmbus::commands::$module::CommandCode::$cmd.read_op()
+            != Operation::Illegal
+        {
+            $out |= Capabilities::$cmd.0;
+        }
+    }};
+}
+
+macro_rules! generator {
```
jamesmunns, line 203:
```suggestion
/// For a given device, calculate the `Capabilities` for each of the
/// status registers.
///
/// The pmbus functions are not const, so generate a closure instead.
macro_rules! generator {
```

task/control-plane-agent/build.rs
```diff
@@ -27,29 +29,111 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    let mut pmbus_rail_dupes = 0;
+    for dev in build_i2c::device_descriptions() {
+        // We only need to map PMBus devices
+        let Some(ref pmbus) = dev.pmbus else {
+            continue;
+        };
+
+        // If it's a pmbus device, we need to get it's status capabilities
+        let Some(func) = pmbus_status_caps.get(dev.device.as_str()) else {
+            panic!("Unknown device: {}", dev.device);
```
jamesmunns: TODO: Nicer error message that point to the fact that `PMBUS_GENERATOR` needs to be updated

task/control-plane-agent/build.rs
```diff
@@ -90,3 +174,54 @@ fn write_keys(
...
+            bitmaker!(out, $module, STATUS_FANS_1_2);
+            bitmaker!(out, $module, STATUS_FANS_3_4);
+            Capabilities(out)
+        })
+    };
+}
+
+use drv_i2c_types::pmbus_status::Capabilities;
+type StatusRow = (&'static str, fn() -> Capabilities);
+const PMBUS_GENERATOR: &[StatusRow] = &[
```
jamesmunns, line 271: TODO: Add a note here that before you add a pmbus device to this list, you should make sure that you have reviewed the `pmbus` crate to make sure that any unsupported status registers are marked as illegal, similar to https://github.com/oxidecomputer/pmbus/pull/35. Remind future impl'ers that failure to do so could cause CML or OTHER error bits to be set.

jamesmunns: TODO: I've only checked `gimlet` devices! That's:

* adm127x
* bmr491 (already had "Illegal" fields discovered)
* isl68224
* raa229618
* tps546b24a

The other devices on this list were required to make CI happy, but I haven't checked these at runtime! I should probably work with @evan-oxide or @hawkw to help find other pmbus-haver SPs (PSC, Observer, ?) to cover the remaining devices, and make a PR to the pmbus crate.

Cargo.toml
```diff
@@ -139,7 +139,7 @@ apob = { git = "https://github.com/oxidecomputer/apob", default-features = false
 # for the migration.
 attest-data = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.4.0", rev = "a0811d06c75c757a6e12c91ed6ea81fde137ba43" }
 dice-mfg-msgs = { git = "https://github.com/oxidecomputer/dice-util", default-features = false, version = "0.2.1", rev = "a0811d06c75c757a6e12c91ed6ea81fde137ba43" }
-gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", default-features = false, features = ["smoltcp"] }
+gateway-messages = { git = "https://github.com/oxidecomputer/management-gateway-service", branch = "james/pmbus-please", default-features = false, features = ["smoltcp"] }
```
jamesmunns: TODO: remember to set this back before merging

Cargo.toml
```diff
@@ -150,7 +150,7 @@ idol-runtime = { git = "https://github.com/oxidecomputer/idolatry.git", default-
 lpc55_sign = { git = "https://github.com/oxidecomputer/lpc55_support", default-features = false }
 measurement-token = { git = "https://github.com/oxidecomputer/lpc55_support", default-features = false }
 ordered-toml = { git = "https://github.com/oxidecomputer/ordered-toml", default-features = false }
-pmbus = { git = "https://github.com/oxidecomputer/pmbus", default-features = false }
+pmbus = { git = "https://github.com/oxidecomputer/pmbus", branch = "james/pmbus-please", default-features = false }
```
jamesmunns: TODO: remember to set this back before merging

drv/i2c-devices/src/lib.rs
jamesmunns: Replied in the Hubris chat, there was some discussion of this already. I'm on the fence between "well the current impl is already here", and "yeah, we're probably at the break even point between 'bespoke codegen per-func' and 'generate metadata, call one-size-fits-all function'". If you prefer the former, I can make the switch over.

I probably wouldn't even have separate summon_fns and read_fns at that point, just the name and read_fn, and codegen the device-and-rail-getting part too.

jamesmunns (commented):
Self-review comments for Monday

task/control-plane-agent/build.rs
hawkw: > I should probably work with @evan-oxide or @hawkw to help find other pmbus-haver SPs (PSC, Observer, ?)

You're probably going to want to test all the prod boards, there are PMBus components on Sidecar as well.

drv/i2c-devices/src/lib.rs
hawkw: Honestly, it's kind of up to you. I would _not_ recommend generating a separate function for each rail, because of code size, but I would consider generating a function for each component type that takes a rail. I'm also fine with the current approach, and it's probably more code size efficient. I'm going to leave this one up to you — whatever you think feels best is fine with me.

build/i2c/src/lib.rs
```diff
@@ -1390,12 +1401,33 @@ impl ConfigGenerator {
...
+                writeln!(&mut self.output, "({out}, {raw_bank})\n        }}")?;
+
+                // ---
+                // Second accessor, returns `(I2cDevice, Option<u8>)`
+
+                write!(
+                    &mut self.output,
+                    r##"
+        #[allow(dead_code)]
+        pub fn {}_banked(task: TaskId) -> (I2cDevice, Option<u8>) {{"##,
```
hawkw: i still really don't love this, and in particular, i wish it was called something else. I don't really understand what "_banked" means here.

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +293,118 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+            }
+            if let Some(rail_idx) = rail_idx {
+                pmbus_rail_read!(@raw => dev, rail_idx, cmd as u8, 2)
+            } else {
+                pmbus_read!(@raw => dev, cmd as u8, 2)
+            }
+            .map(u16::from_le_bytes)
+        };
+
+        let get_byte = |cmd, cap| {
```
hawkw: nit, unimportant: i might call this
```suggestion
        let read_u16 = |cmd, cap| {
            if !device_caps.supports(&cap) {
                return Err(PmbusStatusError::Unsupported);
            }
            if let Some(rail_idx) = rail_idx {
                pmbus_rail_read!(@raw => dev, rail_idx, cmd as u8, 2)
            } else {
                pmbus_read!(@raw => dev, cmd as u8, 2)
            }
            .map(u16::from_le_bytes)
        };

        let read_byte = |cmd, cap| {
```
since `get_` feels like an accessor rather than something that does I/O?

drv/i2c-devices/src/lib.rs
```diff
@@ -271,6 +293,118 @@ pub trait Validate<T: core::convert::Into<drv_i2c_api::ResponseCode>> {
...
+            if !device_caps.supports(&cap) {
+                return Err(PmbusStatusError::Unsupported);
+            }
+            if let Some(rail_idx) = rail_idx {
+                pmbus_rail_read!(@raw => dev, rail_idx, cmd as u8, 2)
+            } else {
+                pmbus_read!(@raw => dev, cmd as u8, 2)
+            }
+            .map(u16::from_le_bytes)
+        };
```
hawkw: i'm a little unconvinced that `get_u16` really needs to exist, since there's really only one 16-bit register, but... 🤷‍♀️

task/control-plane-agent/build.rs
jamesmunns: Tested prod boards!

drv/i2c-types/src/lib.rs
```diff
@@ -233,3 +275,35 @@ pub enum Segment {
...
+        pub const STATUS_FANS_3_4: Self = Self(1 << 9);
+
+        /// Does this capability support all capabilities of `other`?
+        ///
+        /// `self` may support *more* capabilities than `other`, but
+        /// not the other way around.
+        #[inline]
+        pub const fn supports(&self, other: &Self) -> bool {
+            (self.0 & other.0) == other.0
+        }
```
hawkw, line 307: take it or leave it: we use the `bitflags` crate pretty pervasively throughout Hubris; we could use it here if you want to. doesn't make a huge different in this case.

drv/i2c-devices/src/lib.rs
jamesmunns: Leaving as-is for now.

task/control-plane-agent/src/main.rs
```diff
@@ -627,3 +629,16 @@ mod idl {
...
+/// Type returned by generated pmbus rail functions
+pub type SummonFn = fn(userlib::TaskId) -> (drv_i2c_api::I2cDevice, Option<u8>);
+
+pub struct PmbusRailBinding {
+    pub name: &'static str,
+    pub summon_fn: SummonFn,
+    pub status_bits: Capabilities,
+}
+
+include!(concat!(env!("OUT_DIR"), "/pmbus_mapping.rs"));
```
hawkw: i might consider stuffing all of this in a module like:
```suggestion

pub(crate) mod pmbus {
    use super::*;

    /// Type returned by generated pmbus rail functions
    pub type SummonFn = fn(userlib::TaskId) -> (drv_i2c_api::I2cDevice, Option<u8>);

    pub struct PmbusRailBinding {
        pub name: &'static str,
        pub summon_fn: SummonFn,
        pub status_bits: Capabilities,
    }

    include!(concat!(env!("OUT_DIR"), "/pmbus_mapping.rs"));
}
```

in particular, `SummonFn` doesn't make a ton of sense without a `pmbus::` in front. but 🤷‍♀️

task/control-plane-agent/src/mgs_common.rs
```diff
@@ -746,6 +747,69 @@ impl MgsCommon {
     ) -> Result<Option<DumpSegment>, GwSpError> {
         self.dump_state.task_dump_read_continue(key, seq, buf)
     }
+
+    pub(crate) fn get_pmbus_status(
+        &mut self,
+        rail: &PowerRailName,
+    ) -> Result<PmbusStatus, GwSpError> {
+        // Get the name as a binary string, up to the first zero
```
hawkw: turbo nit: "the first zero" could mean `0` _or_ `\0`, maybe be more explicit here?
```suggestion
        // Get the name as a binary string, up to the first null byte
```

task/control-plane-agent/build.rs
```diff
@@ -27,29 +29,113 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    let mut pmbus_rail_dupes = 0;
+    for dev in build_i2c::device_descriptions() {
+        // We only need to map PMBus devices
+        let Some(ref pmbus) = dev.pmbus else {
+            continue;
+        };
+
+        // If it is a pmbus device, we need to get its status capabilities
+        let Some(caps) = pmbus_status_caps.get(dev.device.as_str()) else {
+            print!("cargo::error=unknown pmbus device: {}, ", dev.device);
```
hawkw: should we maybe also include the refdes and/or rail of the device or something here, so that you can actually track down which thing in the toml was the wrong one?

task/control-plane-agent/build.rs
```diff
@@ -27,29 +29,113 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+                pmbus_rail_dupes += 1;
+                print!("cargo::error=PMBus device ");
+                print!(
+                    "{} defines a power rail ",
+                    dev.device_id.as_deref().unwrap_or("(no ID)")
+                );
+                println!(
+                    "{:?} which already exists in the manifest",
+                    rail.name
+                );
```
hawkw: i don't really get why this is split across a whole bunch of `print!` invocations instead of being like
```suggestion
                println!(
                    "cargo::error=PMBus device {} defines a power rail {:?} \
                     which already exists in the manifest",
                    dev.device_id.as_deref().unwrap_or("(no ID)"),
                    rail.name
                );
```

task/control-plane-agent/build.rs
```diff
@@ -27,29 +29,113 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+        let Some(ref pmbus) = dev.pmbus else {
+            continue;
+        };
+
+        // If it is a pmbus device, we need to get its status capabilities
+        let Some(caps) = pmbus_status_caps.get(dev.device.as_str()) else {
+            print!("cargo::error=unknown pmbus device: {}, ", dev.device);
+            println!(
+                "add entry to PMBUS_GENERATOR for status register support."
+            );
```
hawkw: i don't really get why this is split across a whole bunch of `print!` invocations instead of being like
```suggestion
            println!(
                "cargo::error=unknown pmbus device: {}, add entry to \
                 PMBUS_GENERATOR for status register support.",
                dev.device
            );
```

hawkw: also, maybe we should say 
```suggestion
            println!(
                "cargo::error=unknown pmbus device: {}, add entry to \
                 PMBUS_GENERATOR in {} for status register support.",
                dev.device,
                file!(),
            );
```
so you know where to find the thing you're supposed to edit?

task/control-plane-agent/build.rs
```diff
@@ -90,3 +176,71 @@ fn write_keys(
...
+
+/// Create a file with anyhow context
+fn context_create_file(path: &Path) -> Result<File> {
+    File::create(path)
+        .with_context(|| format!("failed to create file '{}'", path.display()))
+}
+
+/// Look at the `pmbus` crate metadata to see if a specific command is "Illegal"
+/// and set the capability bit if not.
+macro_rules! bitmaker {
```
hawkw: can this be called something other than `bitmaker`? it's setting the bit if it sees that the register is not marked as `Illegal` by `pmbus`, maybe it could be like `check_if_legal_command` or some such, i dunno.

hawkw (commented):
some last minute suggestions (sorry)

build/i2c/src/lib.rs
jamesmunns: changed `_banked` to `_with_opt_page_idx`, I'll make a follow-up issue to unify this

drv/i2c-types/src/lib.rs
jamesmunns: I'm not going to change this today, let me know if you want me to open a tracking issue to fix it.

task/control-plane-agent/build.rs
jamesmunns: it is now `check_if_marked_as_illegal_by_pmbus_metadata`

task/control-plane-agent/build.rs
jamesmunns: The name of the device itself, e.g. "raa229620a", is the thing that would be missing here. It's not necessarily tied to a single refdes/rail, and is searchable in the toml.

build/i2c/src/lib.rs
```diff
@@ -1410,9 +1410,10 @@ impl ConfigGenerator {
                     &mut self.output,
                     r##"
         #[allow(dead_code)]
-        pub fn {}_banked(task: TaskId) -> (I2cDevice, Option<u8>) {{"##,
+        pub fn {}_with_opt_page_idx(task: TaskId)"##,
```
hawkw, line 1413: yeah i hate this one too, but...i hate it less! :)

task/control-plane-agent/build.rs
```diff
@@ -204,16 +204,54 @@ macro_rules! generator {
...
-            bitmaker!(out, $module, STATUS_VOUT);
-            bitmaker!(out, $module, STATUS_IOUT);
-            bitmaker!(out, $module, STATUS_TEMPERATURE);
-            bitmaker!(out, $module, STATUS_CML);
-            bitmaker!(out, $module, STATUS_OTHER);
-            bitmaker!(out, $module, STATUS_INPUT);
-            bitmaker!(out, $module, STATUS_MFR_SPECIFIC);
-            bitmaker!(out, $module, STATUS_FANS_1_2);
-            bitmaker!(out, $module, STATUS_FANS_3_4);
+            check_if_marked_as_illegal_by_pmbus_metadata!(
```
hawkw, line 207: okay on second thought i don't like how making it that wordy caused the whole thing to be reformatted from like 10 lines into a gazillion lines. i feel like there's gotta be a happy middle ground between `bitmaker!` and `check_if_marked_as_illegal_by_pmbus_metadata!`. but whatever.

hawkw (approved):
let's get this merged!

PR #2539, opened 2026-05-27
Remove `used_with_arg`; add Grapefruit test suite

mkeeter:
We would like to minimize our usage of nightly features.  The only place we use `#[used(linker)]` is in the test suite, and it's not necessary: we need both `#[unsafe(no_mangle)]` and `pub static mut`, but together they prevent `TESTS` from being compiled out of the image.  It's a bit awkward that we now need an `unsafe` to access tests, but that's the cost of `pub static mut`.

### DIFF
--- a/test/test-suite/src/main.rs
+++ b/test/test-suite/src/main.rs
@@ -18,7 +18,6 @@
 //!
 //! The Idol server, `test-idol-server`, tests Idol-mediated IPC.  It must be
 //! included in the image with the name `idol`, but its ID is immaterial.
-#![feature(used_with_arg)]
 #![no_std]
 #![no_main]
 #![forbid(clippy::wildcard_imports)]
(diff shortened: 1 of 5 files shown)


test/test-suite/src/main.rs
```diff
@@ -47,15 +46,14 @@ ringbuf!(Trace, 64, Trace::None);
 /// secure fault, and a different constant will be required.)
 const BAD_ADDRESS: u32 = 0x0;
 
-/// Helper macro for building a list of functions with their names.
-/// We use the humility debug processing to get the name of each
-/// test case and the total number of tests. The #[used(linker)]
-/// is to ensure that this actually gets emitted as a symbol.
+/// Helper macro for building a list of functions with their names. We use the
+/// humility debug processing to get the name of each test case and the total
+/// number of tests. The `unsafe(no_mangle)` and `pub static mut` are to ensure
```
hawkw, line 51: i suppose it can't just be `pub static`, and the `mut` is necessary because...if it's externally visible then you can't stop non-Rust code from `mut`ating it? annoying but reasonable!

hawkw (approved):
looks good to me!

test/test-suite/src/main.rs
mkeeter: Yeah, just `pub static` gets compiled out

labbott (approved):

PR #2545, opened 2026-06-01
Use new NullStr based SpComponent

jamesmunns:
Codegen impl changes necessary to support https://github.com/oxidecomputer/management-gateway-service/pull/496. ~~Should wait for that to merge first.~~ This has now merged. This PR bumps the git version used.

### DIFF
--- a/task/control-plane-agent/src/inventory.rs
+++ b/task/control-plane-agent/src/inventory.rs
@@ -129,8 +129,21 @@ impl Inventory {
         if !device.sensors.is_empty() {
             capabilities |= DeviceCapabilities::HAS_MEASUREMENT_CHANNELS;
         }
+
+        // NOTE: the `from_bstr_unchecked` method expects that:
+        //
+        // 1. The given bytes contain utf-8 data
+        // 2. The given slice is <= SpComponent::MAX_ID_LENGTH
+        //
+        // Since we pass the bytes of a `str` (always good utf-8!), and our
+        // `str`s are built (and length-checked) at compile time, use of this
+        // method is justified. You don't see an unsafe block here, because
+        // SpComponent can be received over the wire, so even if we violated
+        // the rules above, there would be no potential soundness concerns.
+        let component = SpComponent::from_bstr_unchecked(device.id.as_bytes());
+
         DeviceDescription {
-            component: SpComponent::from_bstr_unchecked(device.id.as_bytes()),
+            component,
             device: device.device,
             description: device.description,
             capabilities,
@@ -178,6 +191,9 @@ impl TryFrom<&'_ SpComponent> for Index {
     type Error = SpError;
 
     fn try_from(component: &'_ SpComponent) -> Result<Self, Self::Error> {
+        // TODO(AJM): implement PartialEq/PartialOrd for `SpComponent` et. al,
+        // then make this nicer. We'll want this for some follow-up PMBus
+        // changes as well.
         if let Ok(entry_idx) = task_validate_api::DEVICE_INDICES_BY_SORTED_ID
             .binary_search_by_key(&component.as_bstr(), |&(id, _)| {
                 id.as_bytes()
--- a/task/control-plane-agent/src/inventory.rs
+++ b/task/control-plane-agent/src/inventory.rs
@@ -130,7 +130,7 @@ impl Inventory {
             capabilities |= DeviceCapabilities::HAS_MEASUREMENT_CHANNELS;
         }
         DeviceDescription {
-            component: SpComponent { id: device.id },
+            component: SpComponent::from_bstr_unchecked(device.id.as_bytes()),
             device: device.device,
             description: device.description,
             capabilities,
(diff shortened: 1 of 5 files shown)


jamesmunns:
(no more force pushes, promise)

jamesmunns:
Note to self, to test this, I should probably:

* Make a `hubris` image of this
* Test it on the switch zone using both the pre- and post- https://github.com/oxidecomputer/management-gateway-service/pull/496 version of `faux-mgs`.

jamesmunns:
This has now been tested against a local `grapefruit` board.

* old mgs (`6c0aca2545a73fd75536e149d29faa7108be5862`), pre-https://github.com/oxidecomputer/management-gateway-service/pull/496
* new mgs (`9dd63c0d7533e333c2ebf522f1929443041fd726`), post-496

`read-component-caboose` was used to exercise mgs-to-sp, `inventory` was used to exercise `sp-to-mgs`, output of both tests was ~identical:

```
james@fool:~/oxide/management-gateway-service/faux-mgs git:(main) cargo run --release -- --interface en7 read-component-caboose --component sp --slot=active GITC
    Finished `release` profile [optimized] target(s) in 0.11s
     Running `/Users/james/oxide/management-gateway-service/target/release/faux-mgs --interface en7 read-component-caboose --component sp --slot=active GITC`
Jun 02 19:40:01.974 INFO creating SP handle on interface en7, component: faux-mgs
Jun 02 19:40:01.976 INFO initial discovery complete, addr: [fe80::c1d:98ff:fe5f:c881%22]:11111, interface: en7, socket: control-plane-agent, component: faux-mgs
4c7b83529b034b2e99e754945b5bc4f47e2b749d

james@fool:~/oxide/management-gateway-service/faux-mgs git:(main) cargo run --release -- --interface en7 inventory
    Finished `release` profile [optimized] target(s) in 0.32s
     Running `/Users/james/oxide/management-gateway-service/target/release/faux-mgs --interface en7 inventory`
Jun 02 19:39:56.193 INFO creating SP handle on interface en7, component: faux-mgs
Jun 02 19:39:56.196 INFO initial discovery complete, addr: [fe80::c1d:98ff:fe5f:c881%22]:11111, interface: en7, socket: control-plane-agent, component: faux-mgs
COMPONENT        STATUS       DEVICE           DESCRIPTION (CAPABILITIES)
sp               Present      sp               Service Processor (DeviceCapabilities(1))
sp-aux-flash     Present      sp-aux-flash     Service Processor auxiliary flash (DeviceCapabilities(0))
sp5-host-cpu     Present      sp5-host-cpu     Cosmo SP5 host cpu (DeviceCapabilities(4))
sp5-post-codes   Present      sp5-post-codes   Cosmo SP5 POST code buffer (DeviceCapabilities(0))
host-boot-flash  Present      host-boot-flash  Cosmo host boot flash (DeviceCapabilities(1))
host-boot-apob   Present      host-boot-apob   Cosmo host boot APOB region (DeviceCapabilities(0))
system-led       Present      system-led       System attention LED (DeviceCapabilities(8))
```

task/control-plane-agent/src/inventory.rs
```diff
@@ -179,7 +179,9 @@ impl TryFrom<&'_ SpComponent> for Index {
 
     fn try_from(component: &'_ SpComponent) -> Result<Self, Self::Error> {
         if let Ok(entry_idx) = task_validate_api::DEVICE_INDICES_BY_SORTED_ID
-            .binary_search_by_key(&component.id, |&(id, _)| id)
+            .binary_search_by_key(&component.as_bstr(), |&(id, _)| {
+                id.as_bytes()
+            })
```
hawkw, line 200: it occurs to me that maybe `SpComponent` and the underlying `NullStr` type could have `PartialEq`/`PartialOrd` implementations for `&str` and `&[u8]` so that we have to do a bit less of this...but, maybe not worth the effort to make _oen_ callsite a little nicer.

hawkw: I suppose we will end up using the impl for `NullStr` again later with the PMBus change, though.

task/control-plane-agent/src/inventory.rs
```diff
@@ -130,7 +130,7 @@ impl Inventory {
             capabilities |= DeviceCapabilities::HAS_MEASUREMENT_CHANNELS;
         }
         DeviceDescription {
-            component: SpComponent { id: device.id },
+            component: SpComponent::from_bstr_unchecked(device.id.as_bytes()),
```
hawkw: maybe worth a comment explaining why the use of `_unchecked` is okay? i realize it's not memory-unsafe, but seeing something named "_unchecked" feels like it deserves a "we don't have to check this because it's always codegenned to be correct" type thing.

hawkw (approved):
Well, this turns out to be a surprisingly small change, looks good to me!

As an aside, it would be interesting to see how much smaller this makes the `control-plane-agent` task, now that the strings in the generated `DeviceDescription` array are no longer being null-padded...

task/control-plane-agent/src/inventory.rs
jamesmunns: I'll add a comment!

jamesmunns:
> it would be interesting to see how much smaller this makes the control-plane-agent task, now that the strings in the generated DeviceDescription array are no longer being null-padded

I did realize that it's less of a *clear* win, because we're actually now building an array of slices instead of an array of arrays, which means the strings themselves are no longer padded, but we now have an extra array of 8 bytes per string, which means we win some when the components are < 8 bytes of text, and lose some when the components are > 8 bytes of text.

We'll spend a bit less time searching/comparing (we only compare used bytes, I would bet the search will also match lens first too), but this is probably in the category of "too small to be meaningful".

We could be more clever, but imo it's probably a tossup.

I do think it'll be a bigger win for pmbus rail names, since those are 32-byte max instead of 16-byte max.

task/control-plane-agent/src/inventory.rs
jamesmunns: I'll sneak this in a follow-on PR.

PR #2546, opened 2026-06-01
Add rust-analyzer component

jamesmunns:
Since we support `xtask rust-analyzer`, this means that we should have the rust-analyzer component installed.


mkeeter (approved):

mkeeter:
Seems reasonable; should we also be installing `clippy` since we have `cargo xtask clippy`?

mkeeter (approved):

Issue #2547, opened 2026-06-03
TMP117 on PSC is not a temperature sensor


hawkw:
```console
root@oxz_switch0:~# faux-mgs --interface psc0 inventory
Jun 03 17:23:17.951 INFO creating SP handle on interface psc0, component: faux-mgs
Jun 03 17:23:17.952 INFO initial discovery complete, addr: [fe80::aa40:25ff:fe06:107%76]:11111, interface: psc0, socket: control-plane-agent, component: faux-mgs
COMPONENT        STATUS       DEVICE           DESCRIPTION (CAPABILITIES)
sp               Present      sp               Service Processor (DeviceCapabilities(1))
system-led       Present      system-led       System attention LED (DeviceCapabilities(8))
U31              Present      tmp116           Temperature sensor (DeviceCapabilities(0))
U32              Present      at24csw080       FRU ID EEPROM (DeviceCapabilities(0))
PSU0/ID          Present      m24c02           PSU 0 EEPROM (DeviceCapabilities(0))
PSU0             Present      mwocp68          PSU 0 MCU (DeviceCapabilities(2))
PSU1/ID          Present      m24c02           PSU 1 EEPROM (DeviceCapabilities(0))
PSU1             Present      mwocp68          PSU 1 MCU (DeviceCapabilities(2))
PSU2/ID          Present      m24c02           PSU 2 EEPROM (DeviceCapabilities(0))
PSU2             Present      mwocp68          PSU 2 MCU (DeviceCapabilities(2))
PSU3/ID          NotPresent   m24c02           PSU 3 EEPROM (DeviceCapabilities(0))
PSU3             NotPresent   mwocp68          PSU 3 MCU (DeviceCapabilities(2))
PSU4/ID          NotPresent   m24c02           PSU 4 EEPROM (DeviceCapabilities(0))
PSU4             NotPresent   mwocp68          PSU 4 MCU (DeviceCapabilities(2))
PSU5/ID          NotPresent   m24c02           PSU 5 EEPROM (DeviceCapabilities(0))
PSU5             NotPresent   mwocp68          PSU 5 MCU (DeviceCapabilities(2))
root@oxz_switch0:~# faux-mgs --interface psc0 component-details U31
Jun 03 17:23:45.766 INFO creating SP handle on interface psc0, component: faux-mgs
[...]

Issue #2410, opened 2026-03-03
MWOCP67-5500 PMBus observability and inventory


rmustacc:
This covers making sure that we can get the full PMBus features updated to cover the MWOCP67-5500. Likely requires https://github.com/oxidecomputer/pmbus/issues/28. The end goal here is to ensure that we can successfully talk to all of the devices, have the expected stats that we need defined, and can support control plane inventory of the rectifiers.

hawkw:
@evan-oxide and I discussed this today, and here are some notes:
- With the exception of an unfortunate bug that Evan is writing up an issue for, we should already be close to feature-parity with PSC Classic already, however, we need to do some testing with a live Observer to make sure it actually works.
- Feature-parity with PSC Classic is pretty insufficient for fault management purposes, and there is going to be more work to improve PSC inventory coming soon (see #2212, #2216); this will apply to both PSC Classic and Observer. 
- We should test that the current code actually works.
  + This will require an Observer connected to a real MWOCP67 power shelf with at least some rectifiers present. Ideally, we would also remove/insert them while testing.
  + We will need management network access, either through a Sidecar or through a commodity Ethernet switch (if we have some kind of IX connector to RJ45 adapter?). Hopefully this can be done in @ericaasen's basement using his Sidecar?
 + In particular, I would at least test the following operations using `faux-mgs`:
   * `faux-mgs inventory` should show `Present` for PSUs that are present and `NotPresent` for PSUs that are ... not present. Inserting/removing a PSU should change that PSU's state if you run it again. The PSUs should report that they have the `HAS_MEASUREMENT_CHANNELS` and `IS_PMBUS` `DeviceCapabilities`.
   * `faux-mgs component-details <the component ID of a PSU>` should give you the PMBus sensor readings from that PSU. These are going to look pretty similar to the sensors on the MWOCP68 but may be slightly different since I think the new rectifiers changed some stuff. @evan-oxide will probably remember the differences here.

evan-oxide:
@hawkw Thanks for the testing recommendations! Both `faux-mgs inventory` and `faux-mgs component-details` work as expected, so I'll close this ticket.

PR #2549, opened 2026-06-04
observer: MWOCP67 PMBus Support

evan-oxide:
This PR implements https://github.com/oxidecomputer/pmbus/issues/28. It:
- Adds support for PMBus communication with the MWOCP67 PSUs.
- Combines the `psc-seq-server` and `observer-seq-server` back into one task, since the differences can be managed by a few `cfg`s. (I considered using an `Mwocp6X` trait instead to make it generic over PSU models, but that's hard because of the methods that return device-specific `CommandData` types).
- Polls the onboard tmp117 temperature sensor.
- Adds the `framulator` task to dev builds, to test that the FRAM works.

Known issues / remaining todos:
- It's still unclear how many power rails the MWOCP67 actually has, what their voltages are, and what pages are used to access them. It looks suspiciously like it's reporting stats for the _same_ 50V rail on both pages 0 and 1. There's also insufficient documentation about which other registers are paged.
- The MWOCP67 has 2 new temperature sensors on the positive and negative bus bar clips and they both report ~170°C. This is probably incorrect.
- I wasn't able to manually control the fan speed. I don't know if we care.
- I'm not sure what model of EEPROM is built into the PSUs. I'm assuming it's still the M24C02, like the MWOCP68 had, and I was able to read some bytes off it.
- I skipped implementing the blackbox feature because it looks complicated and we weren't using the (somewhat different) blackbox feature on the PSC.
- PSU firmware updates aren't supported yet.

Testing:
- [X] Basic PMBus reads:
  - [X] Can read model number, manufacturer ID, and serial number
  - [X] Can read temperatures 1, 2 and 3
  - [X] Can read both bus bar clip temperatures
  - [X] Can read fan speed
  - [X] Can read status_word
- [X] humility dashboard shows 5 temps and 1 fan speed per PSU, plus the onboard tmp117
- [X] humility power shows stats for power rails (though they might be wrong, see above)
- [X] When someone removes and reinserts a PSU, I see appropriate events in the ringbuf
- [X] Can read something from the at24csw080 VPD eeprom
- [X] Can read something from the PSUs' m24c02 eeproms
- [x] rail name looks right in ereports


app/observer/base.toml
```diff
@@ -408,19 +408,132 @@ af = 4
...
 refdes = "U29"
+name = "onboard"
+sensors = { temperature = 1 }
 
 [[config.i2c.devices]]
 bus = "local"
+# This one device actually appears at 8 different i2c addresses. 0b1010_000 -
+# 0b1010_011 let you access the main memory registers, where the 2 LSBs of the
+# device address are the 2 MSBs of the memory address. 0b1011_000 - 0b1011_011
+# let you access the security registers and write protection registers.
```
hawkw, line 422: FWIW --- and I am not saying that we should get rid of this comment, for the record --- this is an EEPROM we use kinda all over the place (there's one on pretty much every board), so perhaps we should document this in a more centralized location.

app/observer/dev.toml
```diff
@@ -31,3 +31,11 @@ owner = {name = "udprpc", notification = "socket"}
...
 tx = { packets = 3, bytes = 1024 }
 rx = { packets = 3, bytes = 1024 }
+
+[tasks.framulator]
+name = "task-framulator"
+priority = 5
+start = true
+task-slots = [{spi_driver = "spi2_driver"}]
+stacksize = 2048
+notifications = ["hey-you"]
```
hawkw: we probably don't want to include this in a release image, I think.

drv/i2c-devices/src/mwocp67.rs
```diff
@@ -0,0 +1,579 @@
...
+/// this is always a 17-byte ASCII string. It should be "MWOCP67-5500-B-RM".
+#[derive(Copy, Clone, PartialEq, Eq, Default)]
+pub struct ModelNumber(pub [u8; 17]);
+
+/// Manufacturer ID.
+///
+/// Per Murata Application Note ACAN-157 "PMBus Communication Protocol",
+/// this is always a 9-byte ASCII string. It should be "Murata-PS".
+#[derive(Copy, Clone, PartialEq, Eq, Default)]
+pub struct MfrId(pub [u8; 9]);
```
hawkw: I kind of wonder if we want to factor these out into a separate "Murata FRU ID" module or something, since the FRU ID strings from the MWOCP68 are also the same length. Not a big deal.

drv/i2c-devices/src/mwocp6x/mwocp67.rs
```diff
@@ -0,0 +1,579 @@
...
+            _ => {
+                return Err(Error::InvalidData {
+                    err: pmbus::Error::InvalidCode,
+                });
+            }
+        };
+        Ok(r)
+    }
+
+    pub fn read_speed(&self) -> Result<Rpm, Error> {
```
hawkw: Personally, I might have suggested this be `read_fan_speed`, but...I see that the method on the `mwocp68` driver is `read_speed`, so 🤷‍♀️. Probably worth consistency.

drv/i2c-devices/src/mwocp67.rs
```diff
@@ -0,0 +1,579 @@
...
+
+            if data[index] != expected[index] {
+                return Err(Error::BadFirmwareRev { index: index as u8 });
+            }
+        }
+
+        //
+        // Return the primary MCU version
+        //
+        Ok(FirmwareRev([data[0], data[1], data[2], data[3]]))
```
hawkw: This looks like the same format as the one used by the MWOCP68, right? That observation kinda makes me want to push for pulling out the parsing code into a common "Murata FRU ID" thingy. But, well...I don't think that actually gains us all that much, I just have some aesthetic desire to not duplicate things.

drv/i2c-devices/src/mwocp6x/mwocp67.rs
```diff
@@ -0,0 +1,579 @@
...
+        let _ = self
+            .device
+            .read_block(CommandCode::MFR_ID as u8, &mut id.0)
+            .map_err(|code| Error::BadMfrIdRead { code })?;
+        Ok(id)
+    }
+
+    pub fn status_word(&self) -> Result<STATUS_WORD::CommandData, Error> {
+        // ACAN-157 doesn't specify what page this is on.
+        // Assume it's on page 0, as it is on the better-documented mwocp68.
```
hawkw, line 386: Out of curiosity, have we tested what happens when we read this (and the other status registers) on page 1? Does it also work, or does the MCU get mad at us?

drv/psc-seq-server/src/main.rs
```diff
@@ -100,7 +100,12 @@
 #![no_main]
 
 use drv_i2c_api::I2cDevice;
-use drv_i2c_devices::mwocp68::{self, Mwocp68};
+
+#[cfg(target_board = "observer-a")]
+use drv_i2c_devices::mwocp67::{self as mwocp6x, Mwocp67 as Mwocp6x};
+#[cfg(any(target_board = "psc-b", target_board = "psc-c"))]
+use drv_i2c_devices::mwocp68::{self as mwocp6x, Mwocp68 as Mwocp6x};
```
hawkw: cute :)

drv/psc-seq-server/src/main.rs
```diff
@@ -999,18 +1030,15 @@ impl Psu {
...
+            let psu_label = match self.slot {
+                Slot::Psu0 => '0',
+                Slot::Psu1 => '1',
+                Slot::Psu2 => '2',
+                Slot::Psu3 => '3',
+                Slot::Psu4 => '4',
+                Slot::Psu5 => '5',
             };
-            FixedString::try_from_utf8(&v54_psu[..]).unwrap_lite()
+            Mwocp6x::rail_name(psu_label)
```
hawkw: I found the decision to move this into the drivers a bit weird at first once I saw it, but now that I see how it's being used, it's not unreasonable. However, I don't _love_ putting the rail name in the `drv-i2c-devices` crate, since we could _imagine_ a universe where we have a MWOCP6x PSU on a board where we use different rail names (although admittedly, that's unlikely). Personally, I might have considered just leaving all the code here, and having one of two different "base" strings (`"V54_PSUx"` or `"V50_PSUx"`) based on the board cfg. I'm not super bothered by the current approach, but I feel like it might be a bit simpler that way. Up to you!

task/sensor-polling/src/main.rs
```diff
@@ -96,6 +138,21 @@ impl TemperatureSensor {
...
+                        Err(e) => {
+                            let e = Error::Tmp117Error(e);
+                            ringbuf_entry!(Trace::TemperatureReadFailed(s, e));
+                            sensor_api.nodata_now(s, e.into())
+                        }
+                    }
+                }
+                // The tmp117 is just a temperature sensor, not a speed sensor.
+                assert_eq!(self.speed_sensors.len(), 0);
+            }
```
hawkw, line 146: Am I correct that, after this PR merges, if we were to add a temperature sensor to the TMP117 in the PSC config, we would also be able to fix #2547?

app/observer/dev.toml
evan-oxide: What does release image mean exactly? I thought it wasn't in the release image because this is the dev.toml, but I'm a bit confused about release vs dev vs lab vs those same words in other contexts. (Note that framulator is in the PSC's toml too, though maybe it shouldn't be: https://github.com/oxidecomputer/hubris/blob/master/app/psc/rev-c-dev.toml)

drv/i2c-devices/src/mwocp6x/mwocp67.rs
evan-oxide: I tried reading a bunch of these registers on both pages and always got the same value on both. I don't know if that's because the page is ignored or they happened to be the same on both pages.

task/sensor-polling/src/main.rs
evan-oxide: Yeah you'd need to update the PSC's base.toml, add an entry to the SENSORS array in sensor-polling, and test that the ComponentDetails part works (not just `humility sensors`).

drv/psc-seq-server/src/main.rs
evan-oxide: Yeah I don't love this approach either. On one hand it seems like the Mwocp67 class shouldn't need to know about rail naming conventions, and on the other hand scattering too many `cfg`s in the psc-seq-server could get confusing. So is this what you're suggesting?
```rust
let rail = {
    #[cfg(any(target_board = "psc-b", target_board = "psc-c"))]
    let mut rail_name = *b"V54_PSUx";
    #[cfg(target_board = "observer-a")]
    let mut rail_name = *b"V50_MAIN_PSUx";

    rail_name[rail_name.len() - 1] = match self.slot {
        Slot::Psu0 => b'0',
        ...
    };
    FixedString::try_from_utf8(&rail_name[..]).unwrap_lite()
};
```

evan-oxide: Actually it's a problem that the rail names are different lengths now but EreportFields is still using `FixedString<8>`.

app/observer/dev.toml
hawkw: oh, I didn't see that this was the dev config, never mind --- sorry, I guess I can't read. I think we should leave it in there, then! We'll want to delete it if we start actually using the FRAM, though, so we don't clobber anything we might have actually written to it on purpose.

drv/psc-seq-server/src/main.rs
hawkw: Yeah, we will have to either just make it longer and null pad it on the MWOCP68 version, or make the const size depend on which board we're building. But I do think having all of this right there in that file feels better to me, rather than hiding it in the driver.

mkeeter: FWIW I agree that the names should be here in `psc-seq-server`, rather than in the I2C driver (which should be rail-agnostic).

app/observer/dev.toml
evan-oxide: :+1: I added comments about that

drv/i2c-devices/src/mwocp67.rs
evan-oxide: Done

drv/i2c-devices/src/mwocp67.rs
evan-oxide: I pulled it out into a mwocp6x module along with the shared types. What do you think of the new module structure? It might need to change later if the firmware update process ends up being significantly different, but hopefully not.

evan-oxide: It's unfortunate that we can't share more of the implementations - they look similar but use different types from the pmbus crate.

drv/psc-seq-server/src/main.rs
evan-oxide: Done. I should probably test this... any suggestions for how to generate an ereport and see if the rail name looks good?

drv/i2c-devices/src/mwocp6x/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+
+/// Manufacturer ID.
+///
+/// Per Murata Application Note ACAN-114.A01.D03 "PMBus Communication Protocol",
+/// this is always a 9-byte ASCII string. It should be "Murata-PS".
+#[derive(Copy, Clone, PartialEq, Eq, Default)]
+pub struct MfrId(pub [u8; 9]);
+
+//
+// The boot loader command -- sent via BOOT_LOADER_CMD -- is unfortunately odd
```
mkeeter, line 42: Should this be in the `mwocp68.rs` file instead, or is it the same across models?

evan-oxide: Unclear if it's the same across models, we don't have documentation on it yet for the mwocp67. My thought was to stick it in here for now (since the now-shared Error type uses it) and reorganize if necessary once we implement firmware updates for the mwocp67.

drv/i2c-devices/src/mwocp6x/mwocp67.rs
evan-oxide: We can have both consistency and a good name! I renamed it for the mwocp68 too.

drv/psc-seq-server/src/main.rs
evan-oxide: Tested! I generated a spurious ereport and read it with humility. The rail name looks good.

app/observer/base.toml
evan-oxide: Any thoughts about where it should go? This info is in the datasheet, I just added a note here because I was confused about why I saw 8 devices on the bus.

task/control-plane-agent/build.rs
```diff
@@ -274,6 +274,7 @@ const PMBUS_GENERATOR: &[StatusRow] = &[
     generator!("isl68224", isl68224),
     generator!("lm5066", lm5066),
     generator!("lm5066i", lm5066i),
+    generator!("mwocp67", mwocp67),
```
evan-oxide, line 277: I added mwocp67 to the list so that observer compiles again, but note that `get_pmbus_status` will trigger a CML error until https://github.com/oxidecomputer/pmbus/pull/35 is merged.

evan-oxide: Done, the pmbus PR is merged and bumped here.

app/observer/base.toml
hawkw: Perhaps in a doc comment in `drv_i2c_devices::at24csw080` or something? We can add that in a separate PR.

app/observer/base.toml
```diff
@@ -409,19 +409,132 @@ af = 4
...
+refdes = ["PSU0", "ID"]
+
+[[config.i2c.devices]]
+bus = "backplane"
+name = "psu0mcu"
+address = 0b1011_000
+device = "mwocp67"
+description = "PSU 0 MCU"
+# TODO it's not clear what rails the mwocp67 has or how they are paged.
+# https://github.com/oxidecomputer/hubris/issues/2410
```
hawkw, line 444: nitpick, considered unimportant: might we want to repeat this todo comment on all the `power` stanzas for all the PSUs? just so that we know that if something is eventually done about this, it gets addressed everywhere?

app/psc/rev-c-dev.toml
```diff
@@ -1,6 +1,8 @@
 name = "psc-c-dev"
 inherit = ["rev-c.toml", "dev.toml"]
 
+# This writes test data to the FRAM, so if we start actually using the FRAM then
+# we should remove this task to avoid clobbering the real data.
```
hawkw, line 5: <3 for adding this

drv/i2c-devices/src/mwocp6x/mod.rs
```diff
@@ -0,0 +1,222 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Functionality shared by multiple models of Murata MWOCP6x power supplies
```
hawkw: nit, take it or leave it: perhaps worth explicitly mentioning the specific driver types?
```suggestion
//! Functionality shared by multiple models of Murata MWOCP6x power supplies.
//!
//! Currently, this module supports the [`Mwocp68`], used by the Rack Model 0
//! power shelf, and [`Mwocp6`], used by the Rack Model 1 power shelf.
```

drv/i2c-devices/src/mwocp6x/mod.rs
hawkw: can we add a comment saying that we don't know if this is identical across both devices, perhaps?

Personally, I'd prefer to not put it in the shared module unless we *know* it's the same for both devices, and keep it in `mwocp68.rs`. 

I realize that it's included in the error type, but I think we should probably look at refactoring the error type: I don't love having one big enum for every error any `Mwocpx` device can return from any operation. Currently, we are forcing callers of methods that are not trying to update the firmware to handle all the update-specific errors in *every* call to the driver, which feels messy to me. I think that we should refactor this in a follow-up change. In the meantime, I wonder if we should just rename the variant that can include `BootLoaderCommand` to `Error::BadMwocp68BootLoaderCommand`, and move the update stuff to the MWOCP68 module?

drv/i2c-devices/src/mwocp6x/mwocp67.rs
hawkw: Based on experience with other PMBus devices, I would _guess_ that it's just that the value was the same on both pages; I'd expect that it should be possible to see a fault on one rail and not the other. I think we could probably test this by setting a warning threshold on one page (say `VIN_WARN_LIMIT` or something) but not the other, and seeing if the warning bit gets set on one page and not the other. I could be wrong about this, but I think it's worth investigating, because the current behavior of both MWOCPx drivers always sending page 0 feels sketchy to me.

I'm fine with merging this as-is and investigating as a follow-up though.

drv/i2c-devices/src/lib.rs
```diff
@@ -23,6 +23,7 @@
 //! - [`max6634`]: MAX6634 temperature sensor
 //! - [`max31790`]: MAX31790 fan controller
 //! - [`mcp9808`]: MCP9808 temperature sensor
+//! - [`mwocp67`]: Murata power shelf
 //! - [`mwocp68`]: Murata power shelf
```
hawkw: I think these links are dead now that both `mwocp67` and `mwocp68` live in the `mwocp6x` module?

drv/i2c-devices/Cargo.toml
```diff
@@ -16,6 +16,7 @@ derive-idol-err = { path = "../../lib/derive-idol-err" }
 drv-i2c-api = { path = "../i2c-api" }
 drv-i2c-types = { path = "../i2c-types" }
 drv-onewire = { path = "../onewire" }
+fixedstr = { path = "../../lib/fixedstr", features = ["microcbor"] }
```
hawkw: now that the code for constructing rail names no longer lives in here, i think we no longer need this dependency
```suggestion
```

drv/psc-seq-server/src/main.rs
```diff
@@ -326,6 +350,15 @@ const PSU_PMBUS_DEVS: [fn(TaskId) -> (I2cDevice, u8); PSU_COUNT] = [
...
 ];
+#[cfg(target_board = "observer-a")]
+const PSU_PMBUS_DEVS: [fn(TaskId) -> (I2cDevice, u8); PSU_COUNT] = [
+    i2c_config::pmbus::v50_main_psu0,
+    i2c_config::pmbus::v50_main_psu1,
+    i2c_config::pmbus::v50_main_psu2,
+    i2c_config::pmbus::v50_main_psu3,
+    i2c_config::pmbus::v50_main_psu4,
+    i2c_config::pmbus::v50_main_psu5,
+];
```
hawkw: not a huge deal, but a fairly common practice in tasks that support multiple slightly different boards with mostly common code is to put all of the board specific stuff in a BSP (Board Support Package) submodule which is `#[cfg(...)]` flagged for the appropriate board, rather than having to put a bunch of `#[cfg(...)]` attributes on _every_ definition. so something like this:

```rust
#[cfg(target_board = "observer-a")]
mod bsp {
    pub(crate) const PSU_ENABLE_L_PORT = ...;
    pub(crate) const PSU_ENABLE_L_PINS = ...;
    pub(crate) const ALL_PSU_ENABLE_L_PINS = ...;
    pub(crate) const PSU_PWR_OK_NOTIF = ...;
    pub(crate) const PSU_PMBUS_DEVS = ...;
}
#[cfg(any(target_board = "psc-b", target_board = "psc-c"))]
mod bsp {
    pub(crate) const PSU_ENABLE_L_PORT = ...;
    pub(crate) const PSU_ENABLE_L_PINS = ...;
    pub(crate) const ALL_PSU_ENABLE_L_PINS = ...;
    pub(crate) const PSU_PWR_OK_NOTIF = ...;
    pub(crate) const PSU_PMBUS_DEVS = ...;
}
use self::bsp::*;
```

this ends up being a little neater most of the time, IMO.

drv/psc-seq-server/src/main.rs
```diff
@@ -1001,16 +1034,21 @@ impl Psu {
         let rail = {
             // This is a little silly, but it stops us from having to 6 separate
             // instances of the string "V54_PSU" in the binary...
-            let mut v54_psu = *b"V54_PSUx";
-            v54_psu[7] = match self.slot {
+            // If you add a new name, make sure it still fits in the FixedString.
```
hawkw: nitpicky: mayhaps this ought to have a note about where that size is defined

app/observer/base.toml
evan-oxide: Looks like that file already has doc comments that explain the address situation. So I'll just leave things as they are.

drv/i2c-devices/src/mwocp6x/mwocp67.rs
evan-oxide: I'll leave a comment on https://github.com/oxidecomputer/hubris/issues/2410 about investigating this more

task/sensor-polling/src/main.rs
```diff
@@ -16,28 +18,43 @@ task_slot!(I2C, i2c_driver);
...
-                | Mwocp68Error::BadWrite { code, .. }
-                | Mwocp68Error::BadValidation { code, .. } => code.into(),
-                Mwocp68Error::BadData { .. }
-                | Mwocp68Error::InvalidData { .. } => Self::DeviceError,
+                Mwocp6xError::BadRead { code, .. }
+                | Mwocp6xError::BadWrite { code, .. }
+                | Mwocp6xError::BadValidation { code, .. } => code.into(),
+                Mwocp6xError::BadData { .. }
+                | Mwocp6xError::InvalidData { .. } => Self::DeviceError,
                 _ => Self::DeviceError,
```
mkeeter: You can capture `e` once here to avoid duplication:
```suggestion
            Error::Mwocp67Error(e) | Error::Mwocp68Error(e) => match e {
                Mwocp6xError::BadRead { code, .. }
                | Mwocp6xError::BadWrite { code, .. }
                | Mwocp6xError::BadValidation { code, .. } => code.into(),
                Mwocp6xError::BadData { .. }
                | Mwocp6xError::InvalidData { .. } => Self::DeviceError,
                _ => Self::DeviceError,
            },
```

drv/psc-seq-server/src/main.rs
mkeeter: Another pattern is to have an explicit `bsp` module, then change its path depending on board, e.g.
```rust
#[cfg_attr(
    any(target_board = "minibar-a", target_board = "minibar-b"),
    path = "bsp/minibar.rs"
)]
#[cfg_attr(
    any(target_board = "cosmo-a", target_board = "cosmo-b"),
    path = "bsp/cosmo_ab.rs"
)]
mod bsp;
```

evan-oxide: I created a bsp module, but it got a bit messy and might need some tweaks.
1) I duplicated the PRESENT and OK pin arrays in both files even though they happen to be the same on both boards. I _think_ that's cleaner than interleaving shared and board-specific arrays, but maybe not.
2) The bsp files need to use the PSU_COUNT constant in their array definitions. I had them each define their own constant then I assert that's its equal to 6. I guess I could define a shared PSU_COUNT somewhere else instead.
3) The bsp files need to use the codegen'ed `notifications` and `i2c_config` modules, so I stuck `include!(concat!(env!("OUT_DIR"), "/notifications.rs"));` etc in both of them. Not sure how weird that is.

hawkw: > 1. I duplicated the PRESENT and OK pin arrays in both files even though they happen to be the same on both boards. I _think_ that's cleaner than interleaving shared and board-specific arrays, but maybe not.

I think this is fine; just because it _happens_ to be the same on both boards doesn't mean it _needs_ to be.


> 2. The bsp files need to use the PSU_COUNT constant in their array definitions. I had them each define their own constant then I assert that's its equal to 6. I guess I could define a shared PSU_COUNT somewhere else instead.

Hmm, I was _going_ to say that we should actually not have to do this and that all the code should basically just work with arrays of any length, but then I remembered about the `Slot` enum which has six slot numbers. That kinda makes me think that it would be nicer to just define `PSU_COUNT` in `main.rs` and have the BSPs use `super::PSU_COUNT`, rather than asserting that every BSP uses 6, which feels a bit convoltued.

> 3. The bsp files need to use the codegen'ed `notifications` and `i2c_config` modules, so I stuck `include!(concat!(env!("OUT_DIR"), "/notifications.rs"));` etc in both of them. Not sure how weird that is.

This seems fine to me. You could also leave them in `main.rs` and have the BSPs use `super::i2c_config` and `super::notifications`, but since (at a glance) they are now basically just referenced in the BSP modules, making them belong to the BSP is also totally fine.

drv/psc-seq-server/src/main.rs
```diff
@@ -272,62 +272,19 @@ enum Slot {
...
-const PSU_PMBUS_DEVS: [fn(TaskId) -> (I2cDevice, u8); PSU_COUNT] = [
-    i2c_config::pmbus::v54_psu0,
-    i2c_config::pmbus::v54_psu1,
-    i2c_config::pmbus::v54_psu2,
-    i2c_config::pmbus::v54_psu3,
-    i2c_config::pmbus::v54_psu4,
-    i2c_config::pmbus::v54_psu5,
-];
+// Currently the psc-seq-server only supports boards with 6 PSU slots.
+const _: () = assert!(bsp::PSU_COUNT == 6);
```
hawkw: As I mentioned, I think I'd really prefer we just define `PSU_COUNT` in `main` and make all the BSPs reference it as `super::PSU_COUNT`. Making the BSP define the constant and then asserting that it defined the right thing feels a bit goofy.

(also, the place where we initialize a `[Slot; bsp::PSU_COUNT]` to a 6-element array literal also serves as a slightly more implicit version of this assertion...but again, I think we should just not be asserting it)

hawkw (commented):
One last thing. After that, I think this will be all good!

drv/psc-seq-server/src/main.rs
evan-oxide: Ok I moved `PSU_COUNT` to main. I've tended to think of `use super::` as a code-smell but I don't think that's actually true, and it's definitely an improvement in this situation.

evan-oxide: I moved the codegen out too because main was using `bsp::notifications::TIMER_MASK`.

mkeeter (approved):

hawkw (approved):
looks good to me! thank you for all your hard work on this!

PR #2555, opened 2026-06-09
new FPGA bits to provide proper diagnosis when DDR bulk rails don't come up at the start.

nathanaelhuffman:
See https://github.com/oxidecomputer/quartz/issues/504 and https://github.com/oxidecomputer/quartz/pull/505 for details.

The FPGA was incorrectly enforcing DDR bulk rails in the MAPO logic before we were supposed to which led to incorrect diagnosis of failures to sequence. This fixes the enforcement and as tested hubris properly diagnoses bulk DDR faults now as expected.

Hubris logic was always correct here, so no sw changes were necessary.


hawkw (approved):
I'll take your word for it about the FPGA changes!

I presume we'll want to get a new SP release out with this so that mfg can pick it up?

nathanaelhuffman:
Yeah I think at the next reasonable moment. We don't sequence so this is mostly just annoying to troubleshoot on the line until this lands, so I think it's not super urgent but would be nice as soon as possible.

PR #2558, opened 2026-06-10
Generate a CoRIM

labbott:


build/xtask/src/dist.rs
```diff
@@ -695,6 +695,18 @@ pub fn package(
...
+    // Generate a CoRIM for our measurements for testing
+    let mut corim_builder = rats_corim::CorimBuilder::new();
+    corim_builder.vendor("Hubris Engineering Build".to_string());
+    corim_builder.tag_id(format!("ENGINEERING {}", cfg.toml.name));
+    corim_builder.id("ENGINEERING-BUILD-NOT-TRUSTED".to_string());
+    corim_builder.version(
+        caboose_args
+            .version_override
+            .clone()
+            .unwrap_or("0.0.0-testing".to_string()),
```
mkeeter, line 707: Is there any benefit to using the git revision here?

mkeeter:
From pedantically browsing the lockfile changes, there are a few upstream crates that we could update to minimize the number of different versions in the dependency tree:

- `attest-data` in `dice-util` is still on the older `sha3`
- `rats-corim` is on an old version of `strum`; we may also want to reorganize so that `clap` isn't in the library crate's dependency tree

(this shouldn't block the PR, obviously)

build/xtask/src/dist.rs
labbott: We could, this is also designed to be used as a placeholder so :shrug:

labbott:
>     * `attest-data` in `dice-util` is still on the older `sha3`
> 

I spotted that too and was going to take a pass at that after I got a few more things done

>     * `rats-corim` is on an old version of `strum`; we may also want to reorganize so that `clap` isn't in the library crate's dependency tree
> 

Ah good point. That repo needs some restructuring.

flihp (approved):

PR #2561, opened 2026-06-16
Store measurement result for debug

mkeeter:
When debugging https://github.com/oxidecomputer/customer-support/issues/1271, knowing whether the SP was measured on boot would bifurcate the problem space.  Unfortunately, the measurement token is in a section of RAM used by `net`, so it's been wiped with our stack fill.

```
$ humility -d rack-0801-cosmo-209MYPCP-hubris.dump.0 readmem 0x20000000 64
humility: attached to dump
             \/  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
0x20000000 | fe ca dd ba fe ca dd ba fe ca dd ba fe ca dd ba | ................
0x20000010 | fe ca dd ba fe ca dd ba fe ca dd ba fe ca dd ba | ................
0x20000020 | fe ca dd ba fe ca dd ba fe ca dd ba fe ca dd ba | ................
0x20000030 | fe ca dd ba fe ca dd ba fe ca dd ba fe ca dd ba | ................
```

This PR adds `HUBRIS_MEASUREMENT_RESULT`, which caches the result and can be read with a debugger (or from a full-system dump).

I tested all of the variants on my desktop:
```console
# reflash, which sends the SKIP token
$ humility flash -F
humility: verified auxflash in slot 0
humility: archive appears to be already flashed; forcing re-flash
humility: skipping measurement token handoff
humility: auxiliary flash data is already loaded in slot 0; skipping programming
humility: done with auxiliary flash
humility: flashing done
$ humility readvar HUBRIS_MEASUREMENT_RESULT
HUBRIS_MEASUREMENT_RESULT (0x24000404) = Some(Skipped)

# reset the SP; the RoT measures it immediately
$ humility readvar HUBRIS_MEASUREMENT_RESULT
HUBRIS_MEASUREMENT_RESULT (0x24000404) = Some(Valid {
        count: None
    })

# reset the SP and RoT; release the RoT slightly after SP so the SP resets a few times
$ humility readvar HUBRIS_MEASUREMENT_RESULT
HUBRIS_MEASUREMENT_RESULT (0x24000404) = Some(Valid {
        count: Some(0xe)
    })

# hold the RoT in reset while resetting SP, so the SP times out
$ humility readvar HUBRIS_MEASUREMENT_RESULT
HUBRIS_MEASUREMENT_RESULT (0x24000404) = Some(RetryCountExceeded)
```

### DIFF
--- a/drv/stm32h7-startup/src/lib.rs
+++ b/drv/stm32h7-startup/src/lib.rs
@@ -114,9 +114,17 @@ pub fn system_init_custom(
     // Before doing anything else, check for a measurement handoff token
     #[cfg(feature = "measurement-handoff")]
     unsafe {
+        // After each delay, we'll wait roughly 200 ms.  We double the naive
+        // cycle count because the STM32H7 may (under some circumstances)
+        // dual-issue instructions in the delay loop, which would make the loop
+        // run twice as fast as expected.  We'd rather the loop sometimes run
+        // twice as *slow*, because that just slows down SP boot in cases where
+        // the RoT is not present; if the loop is twice as fast, the SP can time
+        // out before RoT comes up at all, which is a much worse failure mode.
+        const DELAY_CYCLES: u32 = 12860000 * 2;
         const RETRY_COUNT: u32 = 20;
         measurement_handoff::check(RETRY_COUNT, || {
-            cortex_m::asm::delay(12860000); // about 200 ms
+            cortex_m::asm::delay(DELAY_CYCLES);
             cortex_m::peripheral::SCB::sys_reset()
         });
     }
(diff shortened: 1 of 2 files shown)


labbott (approved):

mkeeter:
@cbiffle discovered that the timing of the delay loop can be half as many cycles if correctly aligned (see https://github.com/rust-embedded/cortex-m/issues/430#issuecomment-4723016729), so this PR also doubles the cycle count to fix the actual customer-support issue.  Note that we're also behind on the `cortex-m` crate, and it has both alignment and divide-by-two changes in subsequent releases.

PR #2562, opened 2026-06-16
Update versioning scheme

emilyalbini:
This PR updates Brussels and the release workflow to support the proposed versioning scheme in [RFD 705](https://rfd.shared.oxide.computer/rfd/0705) (tl;dr moving the version number to the minor number rather than the patch, to support backports). The new "Custom version number" field in the release workflow will allow both to migrate to the new versioning scheme, and to issue patch releases (where we'll want to bump the patch number rather than let Brussels bump the minor number).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,9 @@ on:
       brussels-run-id:
         description: Download Brussels from a Run ID
         type: string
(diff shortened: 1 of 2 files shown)


hawkw (approved):
thank you for taking this on <3

PR #2570, opened 2026-06-19
lib/ringbuf: fix two macro syntax bugs in no_dedup and all-disabled arms

NeonPhantom123:
Two syntactically invalid macro arms in lib/ringbuf/src/lib.rs.

The first is in the `#[cfg(not(feature = "disabled"))]` `ringbuf!` macro,
`no_dedup` named arm (line 295). The type expression reads
`Ringbuf<$t, () $n>` — two adjacent generic arguments with no separator.
Rust requires a comma between them; the correct form is `Ringbuf<$t, (), $n>`.
The parallel default arm on line 283 already has this right (`u16, $n`).

The second is in the `#[cfg(all(counters, counters-disabled, disabled))]`
`counted_ringbuf!` macro, `no_dedup` named arm (line 434). The recursive
call uses `%name` instead of `$name` — `%` is not a Rust macro metavariable
sigil. Every other forwarding arm in the file uses `$name`; this one was
a typo.

Neither arm has an active caller in this repo under the currently enabled
feature combinations, so both are latent compile errors. Adds three
integration tests (one function per file, one static per test) covering
the named no_dedup arm, the unnamed delegation form, and the no-collapse
behaviour that distinguishes no_dedup from the default dedup mode.

### DIFF
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -292,7 +292,7 @@ macro_rules! ringbuf {
             });
     };
     ($name:ident, $t:ty, $n:expr, $init:expr, no_dedup $(,)?) => {
-        static $name: $crate::StaticCell<$crate::Ringbuf<$t, () $n>> =
+        static $name: $crate::StaticCell<$crate::Ringbuf<$t, (), $n>> =
             $crate::StaticCell::new($crate::Ringbuf {
                 last: None,
                 buffer: [$crate::RingbufEntry {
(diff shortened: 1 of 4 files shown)


Dev-X25874:
@hawkw — happy to take a look when you get a chance. Let me know if anything needs adjusting.

NeonPhantom123:
@hawkw — happy to take a look when you get a chance. Let me know if anything needs adjusting.

Issue #2506, opened 2026-05-08
Consider requiring a higher resolution platform-specific timer


jamesmunns:
## Background

Currently, the operating system's timers provide [one millisecond resolution](https://hubris.oxide.computer/reference/#_timestamp_format). For most functional uses, this is sufficient. On `cortex-m` based platforms, this timing is provided by recurring `SysTick` increments, and on `risc-v` this is provided using `mtime`.

However, there have been discussions of measuring information such as CPU utilization, or even task-specific counters. It is currently assumed that one-millisecond precision will not be generally suitable for accurate book-keeping, as individual tasks may execute for less than a millisecond (and therefore counted as zero), or in relatively low millisecond durations where the lack of precision would compound (either over or under-estimating the duration).

The SysTick timer is also not particularly well designed to permit sub-tick measurements, particularly as operations within the kernel (e.g. `SVCall`/`PendSV`) run at the same priority as the `SysTick` interrupt:

https://github.com/oxidecomputer/hubris/blob/33a6acba64fab05248dc6dbf6ba3df98f00e3d07/sys/kern/src/arch/arm_m.rs#L682-L704

This makes it difficult to tell whether there has *already* been a rollover.

## Potential Techniques

One potential source of inspiration we could use would be to mimic a technique used by `embassy-stm32`'s implementation of the [`embassy-time-driver`](https://docs.rs/embassy-time-driver/latest/embassy_time_driver/) interface (from [here](https://github.com/embassy-rs/embassy/blob/5dbc479fce23c046177e3c5c719be8b2d631de53/embassy-stm32/src/time_driver/gp16.rs#L66-L85)):

```rust
// Clock timekeeping works with something we call "periods", which are time intervals
// of 2^15 ticks. The Clock counter value is 16 bits, so one "overflow cycle" is 2 periods.
//
// A `period` count is maintained in parallel to the Timer hardware `counter`, like this:
// - `period` and `counter` start at 0
// - `period` is incremented on overflow (at counter value 0)
// - `period` is incremented "midway" between overflows (at counter value 0x8000)
//
// Therefore, when `period` is even, counter is in 0..0x7FFF. When odd, counter is in 0x8000..0xFFFF
// This allows for now() to return the correct value even if it races an overflow.
//
// To get `now()`, `period` is read first, then `counter` is read. If the counter value matches
// the expected range for the `period` parity, we're done. If it doesn't, this means that
[...]

PR #2571, opened 2026-06-19
Conscript TIM5 for early boot

jamesmunns:
This change implements a very simple hardware timer using the 32-bit TIM5.

I'm open to better ideas of how to test this, but using the `HUBRIS_MEASUREMENT_RESULT` as a signal:

```
# Current `master`, measured on a grapefruit
$ humility -d ./hubris.core.0 readvar HUBRIS_MEASUREMENT_RESULT
humility: attached to dump
HUBRIS_MEASUREMENT_RESULT (0x24000404) = Some(Valid {
        count: Some(0x7)
    })

# This change, as-is (20x200ms delays)
$ humility -d ./hubris.core.1 readvar HUBRIS_MEASUREMENT_RESULT
humility: attached to dump
HUBRIS_MEASUREMENT_RESULT (0x24000404) = Some(Valid {
        count: Some(0xd)
    })

# This change, but with `DELAY_MICROS` set to `1_000 * 1_000`, or 20x1s delays
$ humility -d ./hubris.core.2 readvar HUBRIS_MEASUREMENT_RESULT
humility: attached to dump
HUBRIS_MEASUREMENT_RESULT (0x24000404) = Some(Valid {
        count: Some(0x3)
    })
```

This PR is intended as a lay-up to https://github.com/oxidecomputer/hubris/issues/2506, though for that future use we would need to:

1. Re-configure and re-enable the timer for the increased APB1L frequency (100MHz, instead of 64MHz)
2. Implement a "get current rolling time" free function, provide that to the kernel
3. Find a way to tell the kernel the range and frequency of the rolling timer (e.g. 32-bit, 1MHz)
4. Implement the embassy-time "rolling timer extension" algorithm described in #2506
5. (optional) add a "get micros since boot -> u64" syscall for use in userspace


drv/stm32h7-startup/src/lib.rs
```diff
@@ -111,22 +124,49 @@ pub fn system_init_custom(
...
+        const DELAY_MICROS: u32 = 200 * 1_000;
         const RETRY_COUNT: u32 = 20;
-        measurement_handoff::check(RETRY_COUNT, || {
-            cortex_m::asm::delay(DELAY_CYCLES);
+        measurement_handoff::check(RETRY_COUNT, &p, |p| {
+            while p.TIM5.cnt.read().bits() < DELAY_MICROS {}
             cortex_m::peripheral::SCB::sys_reset()
         });
+
+        // Disable timer, we will need to reconfigure it after setting up PLL
```
hawkw: unimportant nitpick: this comment felt a *little* unclear to me, since it could be interpreted as saying that we *will* reconfigure it after PLL init or saying that we don't do that, which is currently the case. what do you think of maybe 
```suggestion
        // Disable timer. If we want to use it again, we will need to reconfigure it after setting up the PLL
```

or something?

drv/stm32h7-startup/src/lib.rs
```diff
@@ -111,22 +124,49 @@ pub fn system_init_custom(
...
+    // resets to 0 AFTER counting this number.
+    p.TIM5.psc.write(|w| w.psc().bits(63));
+    // Set auto-reload to u32::MAX
+    p.TIM5.arr.write(|w| w.arr().bits(u32::MAX));
+    // Generate update (latch the PSC and ARR values)
+    p.TIM5.egr.write(|w| w.ug().set_bit());
+
     // Before doing anything else, check for a measurement handoff token
     #[cfg(feature = "measurement-handoff")]
     unsafe {
```
hawkw, line 129: is this still unsafe? i had kind of assumed that the unsafe block was because of the `asm::delay`...but perhaps the SCB reset call is also unsafe?

drv/stm32h7-startup/src/lib.rs
```diff
@@ -111,22 +124,49 @@ pub fn system_init_custom(
...
+    p.TIM5.cr1.modify(|_r, w| w.cen().disabled());
+    // Set counter to zero
+    p.TIM5.cnt.modify(|_r, w| w.cnt().bits(0));
+    // Set prescaler to (FREQ / 1M) - 1, (64M / 1M) - 1 = 63, as the counter
+    // resets to 0 AFTER counting this number.
+    p.TIM5.psc.write(|w| w.psc().bits(63));
+    // Set auto-reload to u32::MAX
+    p.TIM5.arr.write(|w| w.arr().bits(u32::MAX));
+    // Generate update (latch the PSC and ARR values)
+    p.TIM5.egr.write(|w| w.ug().set_bit());
```
hawkw: take it or leave it: it seems like all of this could be behind the `#[cfg(feature = "measurement-handoff")]` since that's all we're using TIM5 for, and if we wanted to use it again in the future we would have to reconfigure it again after PLL initialization, right? so if we're not going to try and check for a RoT measurement handoff, we don't need to do any of this, AFAICT?

drv/stm32h7-startup/src/lib.rs
jamesmunns: `measurement_handoff::check` is the actual `unsafe` code in question here. asm::delay and SCB sys_reset are both safe.

drv/stm32h7-startup/src/lib.rs
jamesmunns: That's a reasonable question, I was planning to discuss this with folks to see what the feedback was, and whether to take this further. There are a few other places we use asm::delay, where we might want to knock out as well. These are in the platform-specific main apps `app/gimlet/src/main.rs`, `app/minibar/src/main.rs`, etc., so we would need to export EITHER:

* A "get raw microseconds -> u32" function
* A "delay microseconds(n: u32)" function

The former would ALSO be useful for later impl of #2506, BUT would mean that we need to potentially handle rolling math manually in each of these callsites. That being said: I don't expect boot to take the 71 minutes for a rollaround to occur, that feels a little unsatisfying though.

The latter would help here, but probably wouldn't be necessary for #2506. That being said, it's not *that* hard to do, and since we want it in early boot anyway, I've really just convinced myself that I should just impl both for now. So, early Monday, I'll update this PR to:

1. Actually re-configure the timer to have a PSC of 99 (100MHz / 1MHz - 1), *maybe* reset the `cnt` back to 0, trigger an `egr.ug` latch, then start the timer again after RCC setup is complete
2. Provide public "get raw micros wrapping" and "delay micros" functions in the `stm32h7_startup` crate, and use `delay_micros` in the `check` closure
3. Update the platform-specific mains and fully banish `cortex_m::asm::delay`

drv/stm32h7-startup/src/lib.rs
```diff
@@ -15,6 +15,19 @@ use stm32h7::stm32h753 as device;
 #[cfg(any(feature = "h743", feature = "h753"))]
 #[pre_init]
 unsafe fn system_pre_init() {
+    // /!\ EXTREME DANGER WARNING /!\
+    //
+    // We are running this function *before* the startup routine has completed,
+    // meaning that `static`s have NOT been initialized. This is extremely
+    // likely to be unsound in the general case, and should probably be
+    // rewritten in `global_asm!` some day, as the `pre_init` macro is now
+    // deprecated.
```
labbott, line 24: 😬  I think `pre_init` is deprecated deserves its own bug

labbott (commented):
I'll confess to initially being skeptical of using timers but this looks really nice. I'll take a closer look later this week.

drv/stm32h7-startup/src/lib.rs
jamesmunns: Done, as https://github.com/oxidecomputer/hubris/issues/2572.

drv/stm32h7-startup/src/lib.rs
```diff
@@ -347,8 +374,105 @@ pub fn system_init_custom(
...
+    /// SAFETY: Calling this function will invalidate any current timer counts
+    /// obtained through `get_rolling_micros_since_boot()`, as it resets our
+    /// "epoch". If TIM5 is re-used for another purpose, you MUST NOT call
+    /// `get_rolling_micros_since_boot()` or `blocking_delay_micros()` after
+    /// reconfiguring or disabling TIM5. If APB1 is reconfigured to another
+    /// frequency after calling `configure_tim5`, you must call it again with
+    /// the new frequency for timing measurements to be accurate.
+    pub unsafe fn configure_tim5(p: &device::Peripherals, apb1_mhz: u16) {
+        // Hand-build TIM5 as a 32-bit rolling timer at 1 MHz. Start by enabling
+        // TIM5 is on APB1, which at boot is undivided from the default 64MHz
```
jamesmunns: todo: update this comment, it is a little stale

drv/stm32h7-startup/src/lib.rs
```diff
@@ -347,8 +374,105 @@ pub fn system_init_custom(
...
+        // Set prescaler to (FREQ / 1M) - 1, as the counter resets to 0 AFTER
+        // counting this number.
+        p.TIM5.psc.write(|w| w.psc().bits(apb1_mhz - 1));
+        // Generate update (latch the PSC and ARR values)
+        p.TIM5.egr.write(|w| w.ug().set_bit());
+        // Start counting!
+        p.TIM5.cr1.modify(|_r, w| w.cen().enabled());
+    }
+
+    pub fn assert_tim5_running() {
```
jamesmunns: todo: add doc comment

drv/stm32h7-startup/src/lib.rs
```diff
@@ -114,19 +136,20 @@ pub fn system_init_custom(
...
+        const DELAY_MICROS: u32 = 200 * 1_000;
         const RETRY_COUNT: u32 = 20;
         measurement_handoff::check(RETRY_COUNT, || {
-            cortex_m::asm::delay(DELAY_CYCLES);
+            rolling_timer::blocking_delay_micros(DELAY_MICROS);
             cortex_m::peripheral::SCB::sys_reset()
         });
+
+        // Disable timer, we will need to reconfigure it after setting up PLL
+        p.TIM5.cr1.modify(|_r, w| w.cen().disabled());
```
labbott: Maybe pull this into the `rolling_timer` module? It feels a little off to be poking it directly here.

drv/stm32h7-startup/src/lib.rs
```diff
@@ -347,8 +374,116 @@ pub fn system_init_custom(
     p.RCC.d2ccip2r.modify(|_, w| w.rngsel().pll1_q());
 
     // Hello from target speed!
+    //
+    // Set up the clock for the new nominal APB1 frequency of 100MHz
+    unsafe {
+        rolling_timer::configure_tim5(&p, 100);
+    }
```
labbott: If we're not intending to use this past early startup should we just disable the timer now?

drv/stm32h7-startup/src/lib.rs
```diff
@@ -347,8 +374,116 @@ pub fn system_init_custom(
...
+    pub fn get_rolling_micros_since_boot() -> u32 {
+        let tim5 = unsafe { &*device::TIM5::ptr() };
+        tim5.cnt.read().bits()
+    }
+
+    /// Perform a blocking delay for the given number of microseconds.
+    ///
+    /// If TIM5 is not already enabled and configured, this function will panic.
+    #[inline]
+    pub fn blocking_delay_micros(micros: u32) {
```
labbott: gosh I wish that `core::time::Duration` was a thing because I feel like we want APIs for both microseconds and miliseconds

labbott (commented):
It would be good to do at least one boot/check on a gimlet for the revision strap pins since debugging that was awful.

drv/stm32h7-startup/src/lib.rs
```diff
@@ -387,8 +374,105 @@ pub fn system_init_custom(
...
+        // Set prescaler to (FREQ / 1M) - 1, as the counter resets to 0 AFTER
+        // counting this number.
+        p.TIM5.psc.write(|w| w.psc().bits(apb1_mhz - 1));
+        // Generate update (latch the PSC and ARR values)
+        p.TIM5.egr.write(|w| w.ug().set_bit());
+        // Start counting!
+        p.TIM5.cr1.modify(|_r, w| w.cen().enabled());
+    }
+
+    pub fn assert_tim5_running() {
```
hawkw: ...better go catch it!

drv/stm32h7-startup/src/lib.rs
hawkw: [`core::time::Duration` _is_ a thing](https://doc.rust-lang.org/stable/core/time/struct.Duration.html), and I feel like we should probably be using it here (and everywhere else!)

drv/stm32h7-startup/src/lib.rs
```diff
@@ -387,8 +374,105 @@ pub fn system_init_custom(
...
+    /// be reset to zero. This function may be called multiple times, modulo the
+    /// safety concerns listed below.
+    ///
+    /// SAFETY: Calling this function will invalidate any current timer counts
+    /// obtained through `get_rolling_micros_since_boot()`, as it resets our
+    /// "epoch". If TIM5 is re-used for another purpose, you MUST NOT call
+    /// `get_rolling_micros_since_boot()` or `blocking_delay_micros()` after
+    /// reconfiguring or disabling TIM5. If APB1 is reconfigured to another
+    /// frequency after calling `configure_tim5`, you must call it again with
+    /// the new frequency for timing measurements to be accurate.
```
hawkw: Does this result in a memory safety violation, or does it just make any previously held timestamps incorrect? If this function cannot violate Rust's memory safety invariants if used incorrectly, I feel (somewhat strongly) that it probably should not be `unsafe`.

drv/stm32h7-startup/src/lib.rs
hawkw: (the one reason we _might_ not want to do this is that `core::time::Duration` is 16 bytes on the stack instead of 4)

jamesmunns: It is a thing, but I (somewhat strongly) don't think this is the place for it. I think the later, higher level API exposed by the kernel could/should use it, but core::time::Duration [has both](https://doc.rust-lang.org/stable/src/core/time.rs.html#81-84):

* nanosecond level precision (we don't here)
* a u64 second range (we don't here)

`embassy-time` also doesn't tend to use `core::time::Duration` either, as it means doing more scaling math to convert to/from platform-chosen time base to base-ten seconds and nanos.

I don't think it *gets* us anything in practice, and it means being a lot more careful with enforcing invariants at runtime that are "naturally" held by exposing the unit as a natively rolling `u32`.

drv/stm32h7-startup/src/lib.rs
jamesmunns: I'm open to that. Honestly, `get_rolling_micros_since_boot` is more of a crime to memory safety than `configure_tim5` is.

There's probably a larger conversation to have about "what kinds of peripheral interactions should be marked unsafe", embassy has moved more to "that's a HAL level concern", whereas classic embedded rust has been more proactive at the PAC/register level.

"global timers" are one of those areas that end up having a lot of friction to create proper safety invariants around, and my opinion here is that TIM5 is an impl detail that hopefully will not be seen outside of startup routines, and it will then be the kernel's job to present a reasonable (and safe) user interface around.

drv/stm32h7-startup/src/lib.rs
jamesmunns: 👈👈 ayyyy

drv/stm32h7-startup/src/lib.rs
jamesmunns: We could disable it here. It's not particularly hurting anything, but we're also not *using* it for anything. Happy to be persuaded.

drv/stm32h7-startup/src/lib.rs
jamesmunns: Resolved, no longer requires manual poking.

drv/stm32h7-startup/src/lib.rs
jamesmunns: This is now disabled automatically.

drv/stm32h7-startup/src/lib.rs
jamesmunns: It is no longer unsafe to create the timer.

jamesmunns:
> It would be good to do at least one boot/check on a gimlet for the revision strap pins since debugging that was awful.

I've now tested this on:

* gimlet-d
* sidecar-b
* psc-c

Using both power-off resets (using `niles`' power control), and with soft resets (`humility reset`). I wasn't able to test a `minibar`, but it seems to use the same delay timing as `sidecar`.

Let me know if you'd like me to chase down a `minibar` as well, or if this is close enough for comfort. Either way, I might hack together a test on Monday to toggle a pin using the timer and watch it with a saleae, just to make sure that it really is ticking at 64MHz as I expect.

jamesmunns:
Hacked in some GPIO toggling:

```rust
// Before doing anything else, check for a measurement handoff token
#[cfg(feature = "measurement-handoff")]
unsafe {
    // After each delay, we'll wait roughly 200 ms.
    //
    // You might ask yourself, "how do we have a RETRY_COUNT if the closure
    // diverges"? Well! `measurement_handoff::check` stores the iteration
    // counter in a linker location that persists across soft-reboots.
    const DELAY_MICROS: u32 = 200 * 1_000;
    const RETRY_COUNT: u32 = 20;

    // APB1 is currently 64MHz. Create a rolling timer we can use for now.
    let timer = rolling_timer::RollingTimer::new_tim5(&p, 64);
    measurement_handoff::check(RETRY_COUNT, || {
        let start = timer.get_rolling_micros();
        let mut tickr = start;
        while timer.get_rolling_micros().wrapping_sub(start) <= DELAY_MICROS
        {
            // Low 750us
            p.GPIOK.odr.modify(|_r, w| w.odr1().clear_bit());
            while timer.get_rolling_micros().wrapping_sub(tickr) <= 750 {}
            tickr = tickr.wrapping_add(750);

            // High 250us
            p.GPIOK.odr.modify(|_r, w| w.odr1().set_bit());
            while timer.get_rolling_micros().wrapping_sub(tickr) <= 250 {}
            tickr = tickr.wrapping_add(250);
        }
        cortex_m::peripheral::SCB::sys_reset()
    });
}
```

Screenshot 2026-06-29 at 15 09 29
Screenshot 2026-06-29 at 15 10 18

I'm pretty happy with this accuracy, and is probably within our expected tolerance for HSI64M!

labbott (approved):

PR #2108, opened 2025-06-23
gimlet: add missing temperature sensors for the ISL68224 controller

Aaron-Hartwig:
CC #1573

edit(jamesm): this does not fully fix 1573, moving it to "CC".

PR #2578, opened 2026-06-24
Add missing cosmo temperature sensors

### LINKED ISSUE #2577: Cosmo is missing `isl68224` temperature
#2108 added the missing temperature sensors for gimlet, but cosmo is also missing them:

https://github.com/oxidecomputer/hubris/blob/b931e44d5e340a5dfda2c02153507db1bfa38ed1/app/cosmo/base.toml#L1075-L1083

I'll make a similar change for cosmo, and see if there are any other similar items.

It seems somewhat unfortunate that this needs to be done on a per-device basis, however I'm unsure of other feasible approaches we could take.

jamesmunns:
Closes #2577.

This is a similar change to #2108, but for Cosmo instead of gimlet. I took a look, and the only other apps with this ISL sensor are the gimlet-bu target (which has no sensors), and sidecar (which already has these temp sensors mapped).

### DIFF
--- a/task/power/src/bsp/cosmo_ab.rs
+++ b/task/power/src/bsp/cosmo_ab.rs
@@ -22,9 +22,9 @@ pub(crate) static CONTROLLER_CONFIG: [PowerControllerConfig;
     rail_controller!(Core, raa229620A, vddcr_soc_a0, A0),
     rail_controller!(Core, raa229620A, vddcr_cpu1_a0, A0),
     rail_controller!(Core, raa229620A, vddio_sp5_a0, A0),
-    rail_controller_notemp!(Core, isl68224, v1p1_sp5_a0, A0),
-    rail_controller_notemp!(Core, isl68224, v1p8_sp5_a1, A0), // XXX A0 or A2?
-    rail_controller_notemp!(Core, isl68224, v3p3_sp5_a1, A0), // XXX A0 or A2?
+    rail_controller!(Core, isl68224, v1p1_sp5_a0, A0),
+    rail_controller!(Core, isl68224, v1p8_sp5_a1, A0), // XXX A0 or A2?
+    rail_controller!(Core, isl68224, v3p3_sp5_a1, A0), // XXX A0 or A2?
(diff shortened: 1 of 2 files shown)


task/power/src/bsp/cosmo_ab.rs
```diff
@@ -22,9 +22,9 @@ pub(crate) static CONTROLLER_CONFIG: [PowerControllerConfig;
     rail_controller!(Core, raa229620A, vddcr_soc_a0, A0),
     rail_controller!(Core, raa229620A, vddcr_cpu1_a0, A0),
     rail_controller!(Core, raa229620A, vddio_sp5_a0, A0),
-    rail_controller_notemp!(Core, isl68224, v1p1_sp5_a0, A0),
-    rail_controller_notemp!(Core, isl68224, v1p8_sp5_a1, A0), // XXX A0 or A2?
-    rail_controller_notemp!(Core, isl68224, v3p3_sp5_a1, A0), // XXX A0 or A2?
+    rail_controller!(Core, isl68224, v1p1_sp5_a0, A0),
+    rail_controller!(Core, isl68224, v1p8_sp5_a1, A0), // XXX A0 or A2?
```
jamesmunns, line 26: @mkeeter do we know better now whether this should be A0 or A2?

mkeeter: I think it should be A2, because the chip generates the A0 rail (but is itself powered by A2 rails).  We should test it, though!

jamesmunns: Is there a specific way you'd suggest to test this, or just "make the change and see if a cosmo sled still boots happily about it"?

mkeeter: Basically!  Make the change above (with `A2` instead of `A0`), put a Cosmo sled into A2 (`humility hiffy -c Sequencer.set_state -a state=A2`), and confirm that it's still happily reading temperatures (with `humility sensors`)

labbott (approved):
change LGTM

jamesmunns:
I'll test the scenario Matt suggested today and get this merged.

PR #2582, opened 2026-06-29
Update to Brussels 0.6.0

emilyalbini:
This PR updates the release flow to pull in oxidecomputer/brussels#21. This should be the last release process change needed before we can integrate Brussels with the omicron releng tooling!

This also changes the configuration to match the current behavior of sprot-release: sidecar and observer images are not included in the CoRIM measurements anymore, and reverso and observer images will not be pulled into omicron.

### DIFF
--- a/brussels.toml
+++ b/brussels.toml
@@ -40,22 +40,22 @@ gimlet-e = {}
 gimlet-e-lab = {}
 gimlet-f = {}
 gimlet-f-lab = {}
-psc-b = {}
-psc-c = {}
-sidecar-b = {}
-sidecar-b-lab = {}
-sidecar-b-reverso = {}
-sidecar-c = {}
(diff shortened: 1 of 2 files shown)


brussels.toml
```diff
@@ -42,20 +42,20 @@ gimlet-f = {}
 gimlet-f-lab = {}
 psc-b = {}
 psc-c = {}
```
labbott: I think you want `corim = false` here too

labbott (commented):
Eventually we'll want to generate CoRIMs for sidecar and psc/observer but I think that will require some serious rework

brussels.toml
emilyalbini: Done!

labbott (approved):

PR #2584, opened 2026-07-01
Don't panic on bad nonce_size

evan-oxide:
Before, the SP could cause the RoT to panic by sending it an `attest` request with a `nonce_size` larger than the actual blob of nonce data. Now the RoT replies with a `BadMessageLength` error instead.

This is slightly different than the other uses of `BadMessageLength`, but I don't think it's worth making a new error variant just for this case and plumbing it through MGS. Tell me if you disagree.

Tested on a grapefruit.

### DIFF
--- a/drv/lpc55-sprot-server/src/handler.rs
+++ b/drv/lpc55-sprot-server/src/handler.rs
@@ -571,20 +571,17 @@ impl<'a> Handler {
                 nonce_size,
                 write_size,
             }) => {
-                // Ensure the requester didn't send a bad nonce_size that's
-                // larger than the actual blob.
-                let nonce_size = nonce_size as usize;
-                if nonce_size > req.blob.len() {
+                let Some(nonce) = req.blob.get(..nonce_size as usize) else {
+                    // The requester sent a bad nonce_size that's larger than
+                    // the actual blob.
                     return Err(SprotError::Protocol(
                         SprotProtocolError::BadMessageLength,
                     ));
-                }
+                };
+
                 Ok((
                     RspBody::Attest(Ok(AttestRsp::Attest)),
-                    Some(TrailingData::Attest {
-                        nonce: &req.blob[..nonce_size],
-                        write_size,
-                    }),
+                    Some(TrailingData::Attest { nonce, write_size }),
                 ))
             }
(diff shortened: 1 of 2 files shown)


drv/lpc55-sprot-server/src/handler.rs
```diff
@@ -570,13 +570,23 @@ impl<'a> Handler {
...
-                Some(TrailingData::Attest {
-                    nonce: &req.blob[..nonce_size as usize],
-                    write_size,
-                }),
-            )),
+            }) => {
+                // Ensure the requester didn't send a bad nonce_size that's
+                // larger than the actual blob.
+                let nonce_size = nonce_size as usize;
+                if nonce_size > req.blob.len() {
```
jamesmunns: Tiny nit, I usually prefer something like:

```rust
let Some(nonce) = req.blob.get(..nonce_size) else {
    return Err(SprotError::Protocol(
        SprotProtocolError::BadMessageLength,
    ));
};

Ok((
    RspBody::Attest(Ok(AttestRsp::Attest)),
    Some(TrailingData::Attest { nonce, write_size }),
))
```

which lets you do the "length check" and "subslice" in one step.

drv/sprot-api/src/error.rs
```diff
@@ -94,7 +94,8 @@ pub enum SprotProtocolError {
     UnsupportedProtocol,
     /// Unknown message
     BadMessageType,
-    /// Transfer size is outside of maximum and minimum lenghts for message type.
+    /// Transfer size is outside of maximum and minimum lengths for message
```
jamesmunns, line 97: I did go and check, we do check that the nonce is a valid size elsewhere (I think min and max are checked in different places?).

jamesmunns (approved):
This LGTM, modulo one tiny style nit and one "made me look".

drv/lpc55-sprot-server/src/handler.rs
labbott: I have also begun to find joy in `let / else`

labbott (approved):

drv/lpc55-sprot-server/src/handler.rs
evan-oxide: Yeah that's nicer!

PR #2585, opened 2026-07-02
thermal: surgically excise power-down timeout

hawkw:
Presently, the thermal loop will power down the system if any component remains above its "critical" threshold for 60 seconds. We believe this behavior to be excessively conservative, and unnecessary, as an uncontrollable thermal runaway event will eventually result in something tripping its "power-down" threshold and shutting the system down immediately.

This commit makes the smallest possible change to the thermal loop to remove the timeout. This allows us to operate at the edge of our thermal envelope for as long as is necessary, while still ensuring that the fans go full-throttle when we hit 90 Funny AMD Temperature Units, or when anything else hits its crit threshold. The rest of the control loop behavior should be identical.

### DIFF
--- a/task/thermal/src/control.rs
+++ b/task/thermal/src/control.rs
@@ -452,9 +452,6 @@ pub(crate) struct ThermalControl<'a> {
     /// Controller state
     state: ThermalControlState,
 
-    /// How long to wait in the `Overheated` state before powering down
-    overheat_timeout_ms: u64,
-
     /// Most recent power mode mask
     power_mode: PowerBitmask,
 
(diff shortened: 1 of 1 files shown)


task/thermal/src/control.rs
```diff
@@ -1437,10 +1415,6 @@ impl<'a> ThermalControl<'a> {
                     // nominal.
                     let values = *values;
                     self.transition_to_fan_party(now_ms, values)
-                } else if now_ms > *start_time + self.overheat_timeout_ms {
-                    // If blasting the fans hasn't cooled us down in this amount
-                    // of time, then something is terribly wrong - abort!
-                    self.transition_to_uncontrollable(now_ms)
```
hawkw, line 1443: this is the important part

labbott:
this tentatively looks okay to me but I don't have the full context so I'll let others review/comment

bcantrill (approved):
Looks great -- thanks for updating the surrounding comments as well.

PR #2587, opened 2026-07-04
Split out `elf` and `stack` components from `xtask`

jamesmunns:
I intend to do some rework of the max-stack analysis code, and would like to break out the max-stack analysis into its own crate to make testing this in isolation easier. As this code also depends on the `xtask::elf` module, I broke that out to avoid circular deps.

This is the minimum change necessary to split things out before any refactoring.


build/elf/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "elf"
```
hawkw: nit...most of the other stuff in `build/` is named `build-foo`...but most of the other stuff is also the `build-foo` for some task/lib/api called `foo`, and this is really part of the xtask infrastructure more than a build script. i'm on the fence as to whether it should even live in `build` at all, but i can't think of anywhere better...

build/stack/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "stack"
```
hawkw: similar nits about naming and also whether or not this is the right directory for this crate to live, but honestly, i don't actually care.

hawkw (approved):
sure! up to you whether you want to wait for others to bikeshed over the crate names and paths...

build/elf/Cargo.toml
jamesmunns: Renamed and left where it is for now

labbott (approved):
This looks much less intimidating to review, thanks for doing this

Issue #2588, opened 2026-07-06
Current max stack analysis code is (significantly) underestimating necessary stack space for tasks


jamesmunns:
TL;DR:

* Our current max-stack analysis is more limited than we thought
* This caused task crashes on the current `master` branch (but not in currently-released versions)
* A reasonable fix of the main new defect is known, though we might want a few more improvements while we're at it
* We should get this fixed ASAP and bump some max stack numbers across a few app manifests.

## Current Approach

The current [static analysis](https://github.com/oxidecomputer/hubris/blob/2f318dc6770e69ac8ccc442bc55b12693cf8f1c0/build/xtask/src/dist.rs#L1465-L1726) that we perform on all tasks works roughly as follows:

1. We configure LLVM to emit `.stack_sizes` debug information using a `-Z` flag, which means that the `elf` file compiled for each task contains a debuginfo section listing the size of stack frames on a per-function basis.
2. We use the `goblin` crate to parse the compiled `elf` files, and obtain:
    1. The per-function stack frame sizes
    2. The list of all function symbols contained in the elf file
    3. The `.text` section of the elf file, which contains executable instructions
    4. Information about where data is inlined within the `.text` section
3. For each function in the elf file, we:
    1. Use the [Capstone library](https://www.capstone-engine.org/) (by way of the [`capstone`](https://docs.rs/capstone/latest/capstone/) FFI crate) to disassemble the instruction code
    2. Iterate through each instruction in the function, and determine all "outgoing" branches which are calls to other functions
    3. Produce a report called `FunctionData` for that function which contains the name, local stack usage, and a list of all called functions by that function
4. We then start from the entrypoint of the task, `_start`, and recurse through each node, to calculate the deepest stack usage of any chain of function calls.
5. This "max stack" usage is compared against the `stacksize` for the task in the manifest, and compilation fails if the calculated max stack exceeds the written stacksize.

## Known Limitations

This approach already had a couple of known limitations, as they are typical for this approach of static analysis.

1. This approach does not handle recursion, as we have no way to annotate a potential upper bound of recursive iterations. Currently, the code silently skips considering any instances of recursion found.
2. This approach does not handle "indirect" branching, which looks something like `blx r5` in assembly, and is often (but not exclusively) generated when calling a function through a vtable method, like `dyn Format`. Currently, the code silently skips considering instances of indirect branching found.
[...]

PR #2589, opened 2026-07-06
Rework max-stack analysis crate, address some limitations

jamesmunns:
Note: this builds on top of #2587, which broke out the relevant functions into a standalone crate.

This PR consists of four main parts:

1. A direct fix for https://github.com/oxidecomputer/hubris/issues/2588's "Program Counter Relative Branching" defect. This is a ~1 line change when filtering the potential instructions for branching
2. An attempt to address https://github.com/oxidecomputer/hubris/issues/2588's "Missing sanity cross-checks", which looks for any functions *not* present in the call graph, and adds the largest call stack of those functions to the evaluated call graph, as a potential "fudge factor" for indirect branching/dyn dispatch we were unable to follow
3. Updating of manifests to increase stack sizes based on 1 + 2
4. A major refactor of the current max stack analysis code

The 4th part isn't strictly necessary, however while investigating and making myself familiar with how our current code works, it was useful to me to split things out so I could follow what was going on, and restructure the analysis to allow for some more interactive debugging/diffing. I *personally* think the code is a bit more accessible now, but that could be more personal preference than anything else.

The 1st and 3rd parts are achievable with a much smaller (a few lines) diff. If this PR is too difficult to review, happy to bail on this or at least defer it.

The 2nd part would be more difficult (but still possible) to achieve without the refactorings made in the 4th part, as the new call graph analysis lets us more easily keep track of which functions have been visiting when traversing the graph.

If we merge this PR, I'd consider the "urgent" part of #2588 resolved, though there are still some "known defects", and we might want to increase the baseline stack requirements to handle register stacking (this would bump most task stack sizes in most manifests by 104 bytes, and +36 for gimletlet).


jamesmunns:
Issues I've found so far:

## At least some code does meaningfully have recursion cycles

```
$ cargo xtask dist test/tests-stm32g0/app-g070.toml
thread 'main' (17436042) panicked at build/stack/src/lib.rs:285:27:
unable to resove 0x08004BD8: Refusing to handle recursion: [0x08004BD8, 0x08004BE8] -> 0x08004BD8
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
 08004bd8 <core::str::slice_error_fail>:
 8004bd8: b580          push    {r7, lr}
 8004bda: af00          add     r7, sp, #0x0
 8004bdc: b082          sub     sp, #0x8
 8004bde: 68bc          ldr     r4, [r7, #0x8]
 8004be0: 9400          str     r4, [sp]
 8004be2: f000 f801     bl      0x8004be8 <core::str::slice_error_fail_rt> @ imm = #0x2
 8004be6: d4d4          bmi     0x8004b92 <<u32 as core::fmt::Display>::fmt+0xaa> @ imm = #-0x58

08004be8 <core::str::slice_error_fail_rt>:
 8004be8: b580          push    {r7, lr}
 8004bea: af00          add     r7, sp, #0x0
...
 8004c92: 9200          str     r2, [sp]
 8004c94: 4622          mov     r2, r4
 8004c96: f7ff ff9f     bl      0x8004bd8 <core::str::slice_error_fail> @ imm = #-0xc2
```

## At least some code is missing proper symbol table metadata

```
objdump -xSC ../../target/oxide-rot-1-selfsigned/dist/attest.tmp | grep '161f4'
000161f4 l       .text	00000000 $t
000161f4 g     F .text	00000000 fe25519_add_asm
                        ^^^^^^^^ -> size = 0
```

the function gets filtered out because of this

```
thread 'main' (17388305) panicked at build/stack/src/main.rs:43:48:
called `Result::unwrap()` on an `Err` value: 000161f4: no function data
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

// The original `.a` doesn't have info for this object. Some functions do, this one doesn't.

```
$ objdump -xSC ~/Downloads/salty-asm.a
SYMBOL TABLE:
00000000 l    d  .text  00000000 .text
00000000 l    d  .data  00000000 .data
00000000 l    d  .bss   00000000 .bss
00000000 l       .text  00000000 $t
00000000 l    d  .ARM.attributes        00000000 .ARM.attributes
00000000 g     F .text  00000000 fe25519_add_asm

Disassembly of section .text:

00000000 <fe25519_add_asm>:
       0: b4f0          push    {r4, r5, r6, r7}
       2: f04f 0701     mov.w   r7, #0x1
```

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
-    task_name: &str,
-    verbose: bool,
-) -> Result<Vec<(u64, String)>> {
-    // Open the statically-linked ELF file
-    let data = std::fs::read(elf).context("could not open ELF file")?;
-    let elf = goblin::elf::Elf::parse(&data)?;
+/// Typically stored in an [`Rc`] as each node may appear multiple times across
+/// the call graph of a program.
+pub struct ResolvedNode {
+    pub addr: u32,
```
hawkw, line 136: Take it or leave it: it seems like these are (almost?) always being stored in `BTreeMap`s with the `addr` as the key. This seems like it might be a perfect candidate for using [`iddqd::IdOrdMap`](https://docs.rs/iddqd/latest/iddqd/struct.IdOrdMap.html) and implementing `IdOrdItem for Rc<ResolvedNode>`, rather than `std::collections::BTreeMap`, so that we can ensure that the value of the `addr` field stored by the `ResolvedNode` is _always_ what we use as the key into those maps.

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+    pub recursive_calls: usize,
+}
+
+/// The resulting report of all metadata obtained for functions after
+/// extracting from the elf, but before attempting to "resolve" or visit each
+/// function to determine max stack usage.
+pub struct FunctionReport {
+    pub function_items: BTreeMap<u32, FunctionData>,
+    pub addr_to_frame_size: BTreeMap<u32, u64>,
+    pub names: BTreeMap<u32, String>,
```
hawkw: do we expect that a given address will always have an entry in all these maps, or not? what is the difference between something having an entry in `addr_to_frame_size` or `names` from a `FunctionData`'s `name` or `frame_size` values? It would be nice if there was a comment explaining this.

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
-        addr_to_frame_size.insert(addr, size);
+/// A tool for turning a [`FunctionReport`] into a hydrated call graph.
+///
+/// This visits a call graph of [`FunctionData`] recursively, turning each node
+/// into a memoized `Rc<ResolvedNode>`, and keeps a mapping of all nodes by
+/// address.
+pub struct Resolver {
+    pub call_stack: Vec<u32>,
+    pub all_resolved: BTreeMap<u32, Rc<ResolvedNode>>,
+    pub fn_items: BTreeMap<u32, FunctionData>,
```
hawkw, line 151: similarly, I wonder if we might want `FunctionData` to have an `addr` field, ane make this an `IdOrdMap` as well.

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+            all_resolved: BTreeMap::new(),
+            fn_items,
+        }
+    }
+
+    /// Attempt to resolve a function by name into a [`ResolvedNode`].
+    pub fn resolve_by_name(&mut self, entry: &str) -> Result<Rc<ResolvedNode>> {
+        let Some(item) = self.fn_items.iter().find(|(_k, v)| &v.name == entry)
+        else {
+            bail!("Not found");
```
hawkw: nit: maybe
```suggestion
            bail!("function `{entry}` not found");
```

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+    /// Resolve a function by address into a [`ResolvedNode`].
+    pub fn resolve_addr(&mut self, addr: u32) -> Result<Rc<ResolvedNode>> {
+        // Have we already resolved this node?
+        if let Some(node) = self.all_resolved.get(&addr) {
+            return Ok(node.clone());
+        }
+
+        // no, we havent. Get the node info from the fn data
+        let Some(item) = self.fn_items.get(&addr) else {
+            bail!("no function data");
```
hawkw: nit: maybe
```suggestion
            bail!("no function data for {addr:#x}");
```

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+        for child in children {
+            if self.call_stack.contains(&child) {
+                bail!("Refusing to handle recursion");
+            }
+
+            // Resolve the child
+            let rchild = self.resolve_addr(child)?;
+
+            let ttl_child = rchild.max_stack();
+            max_children = max_children.max(ttl_child);
```
hawkw: nit/style: is `ttl_child` meant to be interpreted as `total_child`? in my brain, `ttl` means "Time To Live", which is nonsensical here. perhaps I ought to be able to interpret it from context cluse, but I'd kinda prefer we call this:
```suggestion
            let total_child = rchild.max_stack();
            max_children = max_children.max(total_child);
```

or, now that i'm suggesting changes, it occurs to me that it could also be:


```suggestion
            max_children = max_children.max(rchild.max_stack());
```

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+            if !self.all_resolved.contains_key(&addr) {
+                missing.push(addr);
+            }
+        }
+
+        let mut found = vec![];
+        for addr in missing {
+            let node = self.resolve_addr(addr).unwrap();
+            found.push(node);
+        }
```
hawkw: take it or leave it, but I think this could all be golfed down a bit using iterator stuff
```suggestion
        let mut found = all_addrs.iter()
           .filter(|addr| !self.all_resolved.contains_key(&addr))
           .collect::<Vec<_>>() // necessary to avoid double borrowing self
           .into_iter()
           .map(|addr| self.resolve_addr(addr).unwrap())
           .collect::<Vec<_>>();
```

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+                    to_keep.push(val);
+                }
+            }
+            found = to_keep;
+            if found.len() == last_len {
+                break;
+            } else {
+                last_len = found.len();
+            }
+        }
```
hawkw: could we maybe get a comment explaining what's going on here? i found it a bit hard to follow...

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+
+/// Load and parse the `elf` file at the given path, producing a report of
+/// metadata about all functions in the `elf` file.
+pub fn extract_function_items(
+    elf: &Path,
+    task_name: &str,
+    verbose: bool,
+) -> Result<FunctionReport> {
+    // Open the statically-linked ELF file
+    let data = std::fs::read(elf).context("could not open ELF file")?;
```
hawkw: nit: might be nice if this said
```suggestion
    let data = std::fs::read(elf).with_context(|| format!("could not open ELF file {}", path.display()))?;
```
so you could know which elf was the one it couldn't open?

build/stack/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use std::{path::Path, rc::Rc};
+
+use stack::{FunctionReport, ResolvedNode, Resolver};
+
+fn main() {
+    let file = "../../target/gimlet-c/dist/host_sp_comms.tmp";
+    let items =
+        stack::extract_function_items(Path::new(file), "host_sp_comms", false)
+            .unwrap();
```
hawkw: should this maybe be an xtask that can be called with any task's elf file? or, if the intent is just to show how the code works, maybe it ought to be an example?

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+    let elf = goblin::elf::Elf::parse(&data)?;
```
hawkw: similarly, maybe
```suggestion
    let elf = goblin::elf::Elf::parse(&data).with_context(|| format!("could not parse {}", path.display())?;
```

build/stack/src/lib.rs
```diff
@@ -92,178 +379,270 @@ pub fn get_max_stack(
...
-                        panic!("bad operand type: {op:?}");
-                    };
                     // We can't resolve indirect calls, alas
+                    //
+                    // TODO: We could consider keeping track of register ops
+                    // and potentially figure out the location of the register
+                    // based jump here
                     let arm::ArmOperandType::Imm(target) = op.op_type else {
+                        if verbose {
+                            println!(
```
hawkw: nit: should this perhaps be
```suggestion
                            eprintln!(
```

build/stack/src/lib.rs
```diff
@@ -92,178 +379,270 @@ pub fn get_max_stack(
...
-                    g == &InsnGroupId(InsnGroupType::CS_GRP_CALL as u8)
-                        || (g == &InsnGroupId(InsnGroupType::CS_GRP_JUMP as u8)
-                            && can_tail)
-                }) {
+                let is_tail_call = |g| is_grp_jump(g) && can_tail;
+                let is_branching_instr = detail.groups().iter().any(|g| {
+                    // Check if this is now not needed anymore
+                    if !is_grp_rel(g) && is_tail_call(g) {
+                        panic!("this check is NOT obsolete!");
+                    }
```
hawkw: can you explain this?

build/stack/src/lib.rs
```diff
@@ -92,178 +379,270 @@ pub fn get_max_stack(
...
+                    // which is where we are jumping to. Get the last operand so
+                    // we can determine if we can follow this.
                     let arch = detail.arch_detail();
-                    let ops = arch.operands();
-                    let op = ops.last().unwrap_or_else(|| {
+                    let ArchDetail::ArmDetail(details) = arch else {
+                        panic!("Unsupported arch");
+                    };
+                    let op = details.operands().last().unwrap_or_else(|| {
                         panic!("missing operand!");
```
hawkw: nit, take it or leave it: maybe the panic message could be where we say that ARM jumps always have an operand?

build/stack/src/lib.rs
```diff
@@ -92,178 +379,270 @@ pub fn get_max_stack(
...
                     let target = u32::try_from(target).unwrap();
 
-                    // Avoid recursive calls into the same function (or midway
-                    // into the function, which is a thing we've seen before!
-                    // it's weird!)
-                    if !(base_addr..base_addr + sym.st_size as u32)
-                        .contains(&target)
-                    {
+                    // Avoid recursive calls to the same function, and ignore
+                    // any control flow that directs back inside this function.
```
hawkw: nit: maybe this comment ought to be reworded a bit --- "avoid recursive calls" felt confusing to me especially since we are now _counting_ the recursive calls, which feels like a substantially different thing than "avoiding" them...

build/stack/src/lib.rs
```diff
@@ -92,178 +379,270 @@ pub fn get_max_stack(
...
+        })
+}
 
-    let mut out = vec![];
-    for m in max_stack {
-        let f = fns.get(&m).unwrap();
-        let name = &f.short_name;
-        out.push((f.frame_size.unwrap_or(0), name.clone()));
+fn name_shortener(name: &str) -> String {
+    // Strip the trailing hash from the name for ease of printing
```
hawkw: are there situations in which this might be collidey?

hawkw (commented):
this is mostly style nitpicking because i didn't really understand the old code and i am not sure if i fully understand the new code either.

build/stack/src/lib.rs
jamesmunns: This is true, I *used* to `bail!` here, until I realized that we couldn't actually do this in practice. Will update the comment.

build/stack/src/lib.rs
jamesmunns: I'll look into `iddqd`, or at least clarify why they are separate if I can't do that. The short answer is "these didn't used to be bundled up, it was more sequential, but now I put them all together so you could look at all the intermediary analysis stages", but that's more of an explanation rather than a defense of a good API.

build/stack/src/main.rs
jamesmunns: Yes, that would probably be smart. The *intent* was to be a demonstration driver here, and maybe make a standalone cli interface for using this (without xtask).

The actual thing that happened was "this is where james is manually hacking on analysis when something goes weird", and this probably should be removed or made into something nicer.

build/stack/src/lib.rs
jamesmunns: I *think* the old code was printing to stdout with `--verbose`, not stderr, but I will double check.

build/stack/src/lib.rs
jamesmunns: will do, I think the new change superceded the old "tail call" check, but wasn't very confident about this. This was a tombstone to tell me if that was in fact not true. I will add a comment or remove this.

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+//!
+//! ## Known Limitations
+//!
+//! This approach already had a couple of known limitations, as they are
+//! typical for this approach of static analysis.
+//!
+//! 1. This approach does not handle recursion, as we have no way to annotate a
+//!    potential upper bound of recursive iterations. Currently, the code
+//!    counts the number of direct recursion instances detected (e.g. self-calls
+//!    of a function), and refuses to resolve call stacks with cycles.
```
labbott: Would this also let us detect recursion? I almost want to say we should be disallowing recursion in hubris given the difficulties with stack sizing but I would settle for a warning.

build/stack/src/lib.rs
```diff
@@ -2,86 +2,373 @@
...
+//! 2. This approach does not handle "indirect" branching, which looks something
+//!    like `blx r5` in assembly, and is often (but not exclusively) generated
+//!    when calling a function through a vtable method, like `dyn Format`.
+//!    Currently, the code silently skips considering instances of indirect
+//!    branching found.
```
labbott, line 55: Insert rant about panic and formatting (I agree we can't fix that here)

labbott (commented):
I think this is a good approach, I'll take another pass later

build/stack/src/lib.rs
jamesmunns: Unfortunately, I tried to disallow this, but at least `tests-stm32g0` does meaningfully have recursion: https://github.com/oxidecomputer/hubris/pull/2589#issuecomment-4895751197

jamesmunns: Added an allow-list, currently limited to one `core` function and one `smoltcp` function.

jamesmunns:
@hawkw I will likely punt on switching over to `iqqdq`, I agree with you it's a good fit, but I probably need to stop working on this for a bit, and would like to get it landed.

build/stack/src/lib.rs
jamesmunns: Removed, we no longer need this.

jamesmunns:
For the sake of verification, I coerced a couple different generations of the stack analysis to output similar reports, and the easiest way to multi-file-diff (edit: and share it, locally I use [`bcompare`](https://www.scootersoftware.com/)) I could think of was making a git repo: https://github.com/jamesmunns/stack-compare/commits/main/

largely: they change in the way we expect:

- Commit 1/4, which is https://github.com/oxidecomputer/hubris/commit/2f318dc6770e69ac8ccc442bc55b12693cf8f1c0, just before the "fire drill"
- Commit 2/4, which is after https://github.com/oxidecomputer/hubris/pull/2590, adds a whole bunch of missing function calls
- Commit 3/4, which is after https://github.com/oxidecomputer/hubris/pull/2589, now includes functions that aren't "reachable" with recursive analysis
- Commit 4/4, which is after https://github.com/oxidecomputer/hubris/commit/6b4e417a81c968bcdfa69bbb861f61ee4ef54168, doesn't really change anything, but that's good, because the "what" is the same but the "how" is different (this isn't part of 2589 yet)

Commit 2->3 has some meaningful code changes (I tried to back these out, but it was annoying and I gave up), so there are some changed addresses, but 1->2 and 3->4 are relatively the same code, just with analysis changed.

build/stack/src/lib.rs
```diff
@@ -2,12 +2,219 @@
...
+//!
+//! ## Known Limitations
+//!
+//! This approach already had a couple of known limitations, as they are
+//! typical for this approach of static analysis.
+//!
+//! 1. This approach does not handle recursion, as we have no way to annotate a
+//!    potential upper bound of recursive iterations. Currently, the code
+//!    counts the number of direct recursion instances detected (e.g. self-calls
+//!    of a function), and refuses to resolve call stacks with cycles.
```
labbott: Does this comment need to be updated to match our current approach with "blessed recursion"?

labbott (approved):
I think this is a really nice improvement

build/stack/src/lib.rs
jamesmunns: Yes! Updated this comment.

PR #2590, opened 2026-07-06
Increase stack limits, check relative branches

jamesmunns:
This is a more targeted, partial fix for #2588, and contains the most immediate relief from the larger refactoring in #2589.

This catches the largest "novel" defect: ignoring of relative branches, and raises all the stack numbers to what the new analysis in #2589 shows.

This is intended as a stop-gap to prevent seeing stack overflows in test benches. Split into two commits, one with manifest updates, and one with the stack analysis change.


hawkw (approved):
looks good to me

PR #2591, opened 2026-07-07
Record image ID and name in BOOTED_IMAGE at startup

evan-oxide:
See the companion humility PR: https://github.com/oxidecomputer/humility/pull/712


evan-oxide:
Closed in favor of https://github.com/oxidecomputer/humility/pull/711

PR #2592, opened 2026-07-08
Kernel accounting of task CPU usage

jamesmunns:
This is currently "minimum viable experiment" state, it did work while testing, but I was distracted by discovering #2588 while implementing this.


jamesmunns:
Oh, and there are a bunch of semi-unrelated changes/refactoring to kernel unsafe code that I should probably factor out into a separate PR, or revert. I was mostly just putzing around while looking at things understanding how scheduling and startup worked in process.

@cbiffle you might be interested in these, outside of the context of time accounting.

jamesmunns:
Misc note, this PR (and #2571 before it) uses the hardware system timer at 1MHz resolution. However, there is practically little downside IMO for *this* use case to not use the maximum possible speed, 200MHz (the full speed of the APB1 timer clock) for ticking.

With a 32-bit timer, this would roll-over every 21.47 seconds, meaning that as long as we perform "timekeeping" every 10.73 seconds, this frequency is suitable. One potential risk is if a debugger causes the core to be halted longer than that, we may observe "weird" time (potentially breaking monotonic guarantees). At 1MHz, this is 200x longer, meaning our max acceptable interruption time is roughly 35.79 minutes.

With a 64-bit extended timing range at 200MHz, we would have a full range of 2922 years before encountering a roll-over. It seems acceptable to qualify this as "forever", as it is likely we will not experience that duration of uptime. If we are concerned about this, we could operationally require a reboot every two millennia or so for good measure.

This increased precision would allow for fewer rounding errors when calculating time spent in each task. At 1MHz, I'd say our practical resolution is compromised when there is less than 2us between scheduling operations, which is approximately 800 CPU cycles. It is probably not THAT often we are scheduling a task for less than that time, but it does seem at least potentially possible. There is probably some floor to this: whatever our total time spent entering and exiting a task, e.g. one that immediately yields or is interrupted, probably measured in the dozens-to-low-hundreds of cycles. For this, even bumping up to 10(s) of MHz is likely "good enough", even if we don't go all the way up to 200MHz.

One downside of not picking 1MHz as "the official timebase" is potential variance in precision between targets. The LPC55 likely can't run a timer all the way up to 200MHz. That being said, the current impl can handle this reasonably.

drv/stm32h7-startup/Cargo.toml
```diff
@@ -8,6 +8,7 @@ cortex-m = { workspace = true }
 cortex-m-rt = { workspace = true }
 stm32h7 = { workspace = true }
 measurement-handoff = { path = "../../lib/measurement-handoff", optional = true }
+kern = { path = "../../sys/kern" }
```
labbott: It's unfortunate to introduce a dep on `kern` for something in `drv` since we've been able to have that dep be limited to the core app so far.

sys/ptime/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+pub fn set_ptime_vtable(vtable: &'static PTimeVTable) {
+    let vtable: *const PTimeVTable = vtable;
+    let vtable: *mut PTimeVTable = vtable.cast_mut();
+    PTIME_VTABLE.store(vtable, Ordering::Release)
+}
+
+pub fn ptimer() -> Option<&'static PTimeVTable> {
+    let vtable: *mut PTimeVTable = PTIME_VTABLE.load(Ordering::Acquire);
+    let vtable: *const PTimeVTable = vtable.cast_const();
+    unsafe { vtable.as_ref() }
```
labbott, line 50: Had to go re-read the Rust doc to refresh myself on this but it looks great :)

Also could use  // SAFETY comment

sys/kern/src/startup.rs
```diff
@@ -64,21 +65,22 @@ pub unsafe fn start_kernel(tick_divisor: u32) -> ! {
...
+        task_table.as_mut();
+
+    task_table
+        .iter_mut()
+        .zip(task_descs.iter())
+        .for_each(|(t, d)| {
+            let task = t.write(Task::from_descriptor(d));
+
+            // With init done, set up initial register state etc.
+            crate::arch::reinitialize(task);
```
labbott, line 83: The was originally written to be very very explicit about when were were dealing with init vs uninit code. I think that deserves a bigger comment/explanation

sys/kern/src/task.rs
```diff
@@ -379,6 +385,48 @@ impl Task {
...
+    ///
+    /// 1. It will reconfigure the MPU according to instructions in `task`,
+    ///    which could break kernel invariants if done maliciously.
+    /// 2. It will leak a pointer to `task` into static space, where it will be
+    ///    used on next kernel entry.
+    ///
+    /// To avoid causing problems, ensure that `task` is a member of the task
+    /// table, with memory protection generated by the build system, and that
+    /// your access to `task` goes out of scope before next kernel entry.
+    pub(crate) unsafe fn switch_to(&self) {
```
labbott: semantically, this is the same as the old `switch_to` but do we need to be concerned about optimizations doing something unexpected here? I may be overly worried here.

labbott (commented):
Some of this is good cleanup which might be good to review/merge separately.

drv/stm32h7-startup/Cargo.toml
jamesmunns: I think I can fix this, but it will require splitting some things into a separate crate.

jamesmunns:
Rebased on top of #2600, and set that as the base merge target to make the diff more clear.

sys/kern/src/startup.rs
jamesmunns: Discussed in #2600

drv/stm32h7-startup/Cargo.toml
jamesmunns: Handled in https://github.com/oxidecomputer/hubris/pull/2592/commits/ab43f26ef83774fafd9b7bdeec1ab02524c9a64b

jamesmunns:
There are a couple of targets that are failing:

* some `sidecar` targets, which currently have a smaller `kernel` RAM/Flash allowance than other `cosmo`/`gimlet` boards. I'm guessing this isn't intentional, and we can probably bump these somewhat freely
* some `stm32g0` targets. These might be able to be bumped up to the necessary size, but also raises the question whether this should be a toggle-able feature, as it requires `TASKS * size_of(u64)` additional RAM, which isn't a *ton*, but everything is a lot when you don't have very much

I can think of two ways to gate this:

* use the existing kernel `"nano"` feature, disable the `ptime` stuff when it is active. This would gate things out for I think all of the `stm32g0` targets, which already use `"nano"`.
* Introduce a new `"ptime"` feature in the kernel. We can either make this an active-by-default feature (and opt-out in `stm32g0` apps), or NOT make it default, and make all apps that want it opt-in. Either of these options is going to touch a lot of manifests.

jamesmunns:
@labbott this is probably ready for a second look.

PR #2600, opened 2026-07-17
Kernel: Adjustments to `unsafe` code

jamesmunns:
The following is a number of tweaks to the kernel code that in my opinion, make some of the unsafe code a bit more direct to follow. In particular, these changes:

1. Consolidates `switch_to` as a single method on `Task`, rather than one version in `syscalls.rs`, and a handful of semantically-equal copy/pastes
2. Removes as many single/double blind `as` casts as is reasonable, e.g. `x as *const _ as *mut _`, replacing them with discrete steps that note the implicit casts (references to pointer), and explicit casts, using `ptr::cast()`
3. Reduces `unsafe` usage a bit in `startup` when initializing each task slot
4. Makes some methods take `&mut Task` instead of `&Task`, as these are used for setting `CURRENT_TASK_PTR`, which semantically allows us to write-through to that address (often in assembly) when returning to the kernel, and that feels a little sketch to me doing from a pointer with shared provenance
5. Adjusts the codegen for Task slots from `MaybeUninit<[Task; N]>` to `[MaybeUninit<Task>; N]`, as this makes startup a little less awkward

In my opinion, none of these are strictly necessary for soundness, so if we don't want to touch it, I could definitely understand that!

I think 1, 2, and 3 are pretty clear readability/clarity wins.

I think 4 is maybe a little nitpicky, and is a bit more semantically correct, but also unlikely to cause miscompilations in practice.

I think 5 is maybe a mixed bag (we can still make the changes for 3 without 5), and I need to check it isn't going to make `humility` upset as the types have changed (even though the before/after are guaranteed to have the same layout and semantics).

Happy to revert any chunks, or rework these into separate commits if that makes reviewing easier.

This is extracted from #2592, and most of these were noticed while I was hacking around on that.


jamesmunns:
Annoyingly, the RAM usage for g070 didn't increase because we are using more, it's just a quirk that the linker isn't packing statics as intelligently as it possibly could.

Screenshot 2026-07-17 at 15 14 40

sys/kern/build.rs
```diff
@@ -426,8 +426,8 @@ fn generate_statics(generated: &Generated) -> Result<()> {
             pub static HUBRIS_IMAGE_ID: u64 = #image_id;
 
             static mut HUBRIS_TASK_TABLE_SPACE:
-                core::mem::MaybeUninit<[crate::task::Task; HUBRIS_TASK_COUNT]> =
-                core::mem::MaybeUninit::uninit();
+                [core::mem::MaybeUninit<crate::task::Task>; HUBRIS_TASK_COUNT] =
+                [const { core::mem::MaybeUninit::uninit() }; HUBRIS_TASK_COUNT];
```
labbott, line 430: You certainly know more about this area than I do but is this a functional change? I don't think so but this is also accessed via humility so is there any chance this could be disruptive there?

app/donglet/app-g031-i2c.toml
```diff
@@ -78,7 +78,7 @@ address = 0x73
 [[config.i2c.devices]]
 controller = 1
 mux = 1
-segment = 1
```
aapoalas: **thought:** Meh, I wish there was toml formatting. (Mostly because my editor formats it automatically, and has caused annoyance in PRs :) )

sys/kern/src/startup.rs
```diff
@@ -63,22 +64,40 @@ pub unsafe fn start_kernel(tick_divisor: u32) -> ! {
...
+        // a compiler error. We are not observing the contents of any of the
+        // MaybeUninit members of the array, only the length of the array.
+        let htts_len = unsafe {
+            (*core::ptr::addr_of_mut!(HUBRIS_TASK_TABLE_SPACE)).len()
+        };
+
+        assert!(
+            htd_len == htts_len,
+            "Task Descriptor and Task Table Length Mismatch!"
+        );
```
labbott: I'm assuming that putting this in the `const` block forces this to be a compile time check and won't generate an extra `assert` at runtime?

sys/kern/src/task.rs
```diff
@@ -379,6 +379,33 @@ impl Task {
...
+    ///
+    /// 1. It will reconfigure the MPU according to instructions in `task`,
+    ///    which could break kernel invariants if done maliciously.
+    /// 2. It will leak a pointer to `task` into static space, where it will be
+    ///    used on next kernel entry.
+    ///
+    /// To avoid causing problems, ensure that `task` is a member of the task
+    /// table, with memory protection generated by the build system, and that
+    /// your access to `task` goes out of scope before next kernel entry.
+    pub(crate) unsafe fn switch_to(&mut self) {
```
labbott, line 413: I know I asked this question elsewhere (or thought I did?) but I seem to have lost track of it: is there any reason we could be surprised by the compiler optimizing something we didn't expect? I don't think there should be a reason but somehow having task switching code be a method makes me nervous.

app/donglet/app-g031.toml
```diff
@@ -6,7 +6,7 @@ board = "donglet-g031"
 
 [kernel]
 name = "app-donglet"
-requires = {flash = 19168, ram = 1824}
+requires = {flash = 19168, ram = 1832}
```
labbott, line 9: Do you know what caused the increase in RAM? I'm not necessarily against it but I'm very very curious!

labbott: oh whoops I saw the comment above never mind

sys/kern/build.rs
jamesmunns: To the best of my knowledge an "executable" perspective, this is not a functional change: `MaybeUninit` is [`repr(transparent)`](https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent), therefore:

* `MaybeUninit<T>` is the same size + align as `T`
* `MaybeUninit<[T; N]>` is the same size + align as `[T; N]`

If this *wasn't* the case, [this `AsRef` impl](https://doc.rust-lang.org/stable/core/mem/union.MaybeUninit.html#impl-AsRef%3C%5BMaybeUninit%3CT%3E;+N%5D%3E-for-MaybeUninit%3C%5BT;+N%5D%3E) which is stable and safe, wouldn't be sound.

It *might* cause problems for humility, depending on how we access this symbol. The name and size are the same, but it might be a noticeable type change. I'll test this before going to merge it.

sys/kern/src/startup.rs
jamesmunns: yep, exactly! Inside of a const block (and storing it into a const), means that we guarantee this evaluation is made at compile time, and not at runtime. Because we never use `_CHECK`, it should not exist at all at runtime.

sys/kern/src/task.rs
jamesmunns: Could you explain what you mean by this?

Generally, anywhere that the optimizer CAN break our unsafe code, means that either:

* There is a miscompilation/compiler bug (happens, but is very rare)
* We were doing undefined behavior, and it was our fault (much more likely)

In general, "we tested it and it worked before" is never a reasonable validation against doing undefined behavior, as undefined behavior:

* isn't *guaranteed* to cause problems, it can still work even if we're doing UB
* is allowed to only *sometimes* cause problems, like every third compilation, or only compilation on Tuesdays. In practice, it's usually for a much more benign reasons like "you exceeded some limit on a function's size, and now we split the optimization, and that allows bad behavior to appear", or "the path in a debug symbol size changed the total size and that made us iterate optimization paths in a different order".

Even in unsafe code, we need to follow ALL of Rust's safety rules, including:

* Modification of data should never be done through a shared `&` reference, unless it was specifically done through [`UnsafeCell::get()`](https://doc.rust-lang.org/stable/core/cell/struct.UnsafeCell.html#method.get)
* There must never be two live `&mut` to the same data. Practically here, that means that we should never be holding an `&mut Task`, created from the pointer stored in `CURRENT_TASK_PTR`, AND be looking at the `task_table` given to us by `with_task_table`. Since the current task is IN the task table, that would mean that we have two `&mut` covering the same space, which is UB.

It is worth noting that while we may never have two aliasing mutable references, *pointers* are allowed to alias "at rest", and the aliasing only comes into play when reading/writing through them (or creating references from them).

So, the way this usually plays out in practice (from my understanding!) is:

1. When we enter an interrupt, usually through an assembly based shim that stores off the context to the task "save" area. We then enter rust code, and do whatever we need to do with `CURRENT_TASK_PTR`, which we are allowed to mutate because we "gave away" the exclusive access we had to it last time we swapped away. From this point onwards, we should NOT dereference CURRENT_TASK_PTR.
2. We then usually use `with_task_table` to get access to the FULL table, do whatever we need, and then pick a "next" task, and store that in CURRENT_TASK_PTR.
3. We then return out of Rust code, back into the assembly shim, which then restores the state from the task "save" area, and resumes back into the task code

The `&mut` change is working to ensure that the [provenance](https://doc.rust-lang.org/std/ptr/index.html#provenance) of the access to the CURRENT_TASK_PTR's contents is reasonable. In step 1 when our asm (or our rust!) is writing to the task's save area, for example to store off registers while context switching, we are writing to it! Because our Task entities are not in unsafecells, it doesn't seem generally sound to me to be writing through to it, using a pointer we got through a shared reference. By using an &mut reference, we do have that!

I would be particularly astounded if the compiler could "see through" all of the interrupt-entry-point and asm we move through, which is why I mention that I think this change is more pedantic than some of the other ones. I do think it is generally more correct though.

labbott: thanks for spelling this out. I think any concerns I have are probably in the category of compiler bug territory.

sys/kern/build.rs
hawkw: I agree that this is probably not a functional change from Hubris' perspective and shouldn't actually impact how this is initialized. But, I think we should definitely check if it changes the DWARF in a way that makes Humility sad. If it does, it should be okay to change Humility to handle either type, just a bit annoying.

sys/kern/src/arch/arm_m.rs
```diff
@@ -371,7 +371,8 @@ pub fn reinitialize(task: &mut task::Task) {
 
     // Set the initial stack pointer, *not* to the stack top, but to the base of
     // this frame.
-    task.save_mut().psp = frame as *const _ as u32;
+    let frame: *const ExtendedExceptionFrame = frame;
+    task.save_mut().psp = frame as u32;
```
hawkw, line 375: i wonder if it would be more pedantically correct to use [`expose_provenance`](https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.expose_provenance) here? i do not think it actually matters, since from strict provenance's perspective, we are never actually turning the u32 back into a pointer in the kernel (while this does happen, it happens in the task when we return from the syscall and not in a way that's visible to Rust)... but it seems like this could technically be treated as a provenance exposing cast?

sys/kern/src/task.rs
aapoalas: From the sidelines but with a toe in some active opsem discussions: yeah, this sounds like the absolutely right choice.

Technically the `&mut Task` means that a sufficiently advanced compiler could f.ex. see that (hypothetically) `next_task()` will eventually perform a `task.stack_offset = 0` write before doing anything else with the Task, and could therefore move that into this task switching code to be more efficient - ie. `&mut Task` means that the compiler can technically write into the `Task` at any time while that `&mut` lives, but of course it must not cause behaviour to change.

Taking `&Task`, casting that into a pointer to store in `CURRENT_TASK_PTR`, and then casting that back into a `&mut Task` was definitely fully UB (albeit impossible for the compiler to take advantage of) before so I definitely think this is beyond nitpicking :)

sys/kern/src/arch/arm_m.rs
```diff
@@ -1206,15 +1214,17 @@ unsafe extern "C" fn pendsv_entry() {
 
     // Safety: we're dereferencing the current task pointer, which we're
     // trusting the rest of this module to maintain correctly.
-    let current = usize::from(unsafe { (*current).descriptor().index });
+    let current = unsafe {
+        let current = &*current;
+        usize::from(current.descriptor().index)
+    };
```
hawkw, line 1234: this is maybe a bit of a silly nitpick, but i actually found the previous code, where the `usize::from` was outside the unsafe block, slightly clearer because it was obvious what part of the operation was unsafe. though upon actually thinking about it, i realize that a `from` conversion can never be unsafe, and because it's `usize`, it couldn't be a function named `from` that we defined that happens to be unsafe (which we shouldn't have done anyway). so maybe i'm being ridiculous.

jamesmunns: You're not ridiculous, it's actually this shape because [in the other PR](https://github.com/oxidecomputer/hubris/pull/2592/changes#diff-071eb4f76fe8d7a0d61bcc0454c6a2d03527130e5227e17f58dd758419469fefR1213-R1217) I put some more stuff there. I can change it back, this wasn't really a meaningful change other than leaving it will make it easier to put things back when I rebase the other PR.

sys/kern/src/arch/arm_m.rs
jamesmunns: I'm somewhat of the opinion that "that's another program's data", and it isn't necessarily relevant to us, the kernel.

The hardware itself swaps from psp back to msp before re-entering the kernel in an interrupt, so I don't think it's necessary. That explanation is all a little handwavey though.

sys/kern/src/startup.rs
jamesmunns: I simplified this a bit, because I realized we already are implicitly asserting the length of the task table by some of the explicit typing left in previous variables.

sys/kern/build.rs
jamesmunns: Hilariously, I think we're good on this. Running `humility tasks` and `humility dump`, both over SWD and net-hiffy, works to totally fine, and I think it's because for legacy compat reasons, we return a base pointer and task count here:

https://github.com/oxidecomputer/humility/blob/68e7d145f633706fcf51a28071c637b48c70ca96/humility-core/src/hubris.rs#L1929-L1957

We then read back `Task` sized objects from the read blobs, which means it doesn't care about this shape change!

sys/kern/src/arch/arm_m.rs
hawkw: ah, i really don't know if this is actually important, fwiw!

sys/kern/src/arch/arm_m.rs
hawkw: for the record, i also don't think it's _necessary_, i was thinking more from a position of  "is this something we should generally be in the habit of doing?"

aapoalas: The `expose_provenance` docs specifically state: "This is equivalent to `self as usize` ... Furthermore, this (like the as cast) has the [...] effect of marking the provenance of 'exposed'" so there doesn't seem to be absolute any difference between these to.

What would be interesting if you could use [addr()](https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.addr): probably not if it's the hardware using this data.

aapoalas (commented):
Please excuse my unrequested comments; the topic called out to me.

sys/kern/src/task.rs
```diff
@@ -379,6 +380,44 @@ impl Task {
...
+
+    /// Performs the necessary pointer offset math to convert a pointer to a
+    /// task into a pointer to the arch-specific saved state. This can be used
+    /// to obtain the store/restore address *without* taking a reference to the
+    /// task, which could invalidate provenance.
+    ///
+    /// SAFETY: `this` must point to a valid instance of `Self`, e.g. not null
+    /// and where `this.save` does not wrap around.
+    pub unsafe fn save_ptr(this: *mut Self) -> *mut crate::arch::SavedState {
+        unsafe { addr_of_mut!((*this).save) }
```
hawkw: aren't we supposed to spell this as
```suggestion
        unsafe { &raw mut (*this).save }
```
or something now?

jamesmunns: Oh, probably, I still reach for addr of out of habit, but yeah &raw is probably the right move here, I'll adjust that. I just get (illogically?) nervous about accidentally dereferencing something, whereas with addr of I have the comfort of "inside the parenthesis it's fine".

jamesmunns: Updated all of our usages for consistency.

sys/kern/src/arch/arm_m.rs
jamesmunns: I'll leave it for now, to make my life easier when rebasing.

jamesmunns:
Hey @labbott or @hawkw, any chance I could get an approve on this? I'd like to get it merged early in this release cycle if possible.

labbott (approved):
thought I already approved it whoops

PR #2601, opened 2026-07-17
Make psc-seq-server work without the enable pins

### LINKED ISSUE #2564: Make psc-seq-server work without the enable pins
The Observer's `PSUx_ON_L` signals were supposed to enable/disable the MWOCP67 PSU output, but Murata has confirmed that they're not connected and we need to use pmbus commands instead (see https://github.com/oxidecomputer/hardware-observer/issues/89). We should update psc-seq-server to use pmbus commands to disable output and clear faults on the MWOCP67 PSUs, while still using the pins on the MWOCP68 PSUs.

evan-oxide:
The PSC sequencer enables and disables the old MWOCP68 PSUs by toggling GPIOs. Unfortunately that doesn't work on the Observer's new MWOCP67 PSUs, so we need to use pmbus commands instead. This PR moves the now hardware-specific enable/disable code into the BSP modules.

I found another difference during testing: the MWOCP67 takes much longer to assert it's `PWR_OK` signal after being enabled or inserted. It often takes 1-2s, whereas code comments say that the MWOCP68 only took 100ms. This meant that when the sequencer re-enabled the PSU after clearing a fault, `PWR_OK` would stay de-asserted for too long and the sequencer would think the fault condition was still present. I fixed this by increasing the length of the probation period at the end of fault recovery, and adding a similar probation period after insertion.

This closes https://github.com/oxidecomputer/hubris/issues/2564

## Testing
Tested with a fault-injection setup (ie. asking Eric to repeatedly dead-short the main rail).

### Observer + MWOCP67:

- [X] when briefly shorting the rail:
  - [X] a fault is detected
  - [X] the fault recovery process happens once
  - [X] the fault is cleared and the PSU is re-enabled
- [X] when shorting the rail for 20s:
  - [X] the recovery process fails multiple times
  - [X] when the short removed, next recovery process succeeds
- [X] when I manually disable it with the OPERATION command, it detects the "fault" and recovers
- [X] hot-plugging works
  - [X] when a PSU is removed, the removal is detected
  - [X] when a PSU is inserted, it powers on and is not disabled by spurious fault detection
- [X] if the SP restarts during a fault, it logs FoundAlreadyDisabled then recovers
- [X] pmbus retry mechanism works
  - (when `set_enable` is hardcoded to return an error on alternating calls, fault recovery works)
- [X] if `do_action` fails, it does eventually recover
  - (when the first few `set_enable` calls are hardcoded to return an error, fault recovery eventually works)

### PSC + MWOCP68:
- [X] when I manually disable it by toggling the enable pin, it detects the "fault" and recovers
- [X] hot-plugging works
  - [X] when a PSU is removed, the removal is detected
  - [X] when a PSU is inserted, it powers on and is not disabled by spurious fault detection


drv/i2c-devices/src/mwocp6x/mwocp67.rs
```diff
@@ -324,6 +329,42 @@ impl Mwocp67 {
...
+        Ok(state == OPERATION::OnOffState::On)
+    }
+
+    /// Clears faults and status registers, allowing the PSU to resume operation
+    /// if it was latched off due to a fault.
+    pub fn clear_faults_and_latch(&self) -> Result<(), Error> {
+        let mut data = pmbus_read!(self.device, MB_PSU_SETTING)?;
+        data.set_clear_faults(MB_PSU_SETTING::ClearFaults::Clear);
+        pmbus_write!(self.device, MB_PSU_SETTING, data)
+    }
```
hawkw, line 366: Just double-checking, we're sure this command isn't paged, right? The PMBus standard's `CLEAR_FAULTS` is paged, and...silently does Not What You Expect if you don't send a page code. I presume the Murata-specific one actually doesn't expect a page, but I wanted to double-check.

drv/psc-seq-server/src/bsp/observer_a.rs
```diff
@@ -59,3 +51,82 @@ pub const PSU_PMBUS_DEVS: [SummonFn; PSU_COUNT] = [
...
+
+/// Performs the enable/disable action that was requested by the state machine.
+///
+/// The unused arguments are needed by other models of PSU.
+pub fn do_action(
+    action: super::ActionRequired,
+    _sys: &sys_api::Sys,
+    psu_index: usize,
+    dev: &mut Mwocp67,
+    now: u64,
```
hawkw, line 108: it occurs to me that if we're irritated by having to pass different arguments depending on which BSP we're calling into, we _could_ let the BSP define its own type which has the same methods but consists of different values, or something. Not sure if it's worth torturing the code over though.

drv/psc-seq-server/src/main.rs
```diff
@@ -305,50 +347,55 @@ const STARTUP_SETTLE_MS: u64 = 500; // Current value is somewhat arbitrary.
 /// How long to leave a PSU off on fault before attempting to re-enable it.
 const FAULT_OFF_MS: u64 = 5_000; // Current value is somewhat arbitrary.
 
-/// How long to wait after a PSU is inserted, before we attempt to turn it on.
-/// This does double-duty in both debouncing the presence line, and ensuring
-/// that things are firmly mated before activating anything.
+/// How long to wait after a PSU is inserted, before we attempt to turn it on
+/// (MWOCP68 only). This does double-duty in both debouncing the presence line,
+/// and ensuring that things are firmly mated before activating anything.
 const INSERT_DEBOUNCE_MS: u64 = 1_000; // Current value is somewhat arbitrary.
```
hawkw, line 445: Nitpicky: perhaps we ought to rename this to `MWOCP68_INSERT_DEBOUNCE_MS` or something, if it's only used on the '68?

drv/psc-seq-server/src/main.rs
```diff
@@ -366,30 +413,42 @@ enum PresentState {
...
+    /// to assert OK after being enabled. The MWOCP67 is much slower to assert
+    /// OK than the MWOCP68 is.
     ///
     /// Once the deadline elapses, we'll transition to the `On` state and start
     /// requiring OK to be asserted.
-    OnProbation { deadline: u64 },
+    OnProbation {
+        deadline: u64,
+        reason: ProbationReason,
+    },
```
hawkw, line 533: Hm, I think I get why we're using the `OnProbation` state to represent the "newly inserted, waiting for PSU to assert PWR_OK of its own free will", rather than the `NewlyInserted` state --- we want behavior more like the "probation" behavior than the "wait for MWOCP68 pin debounce" behavior. But it _feels_ weird to not represent this using the `NewlyInserted` state. I'm not convinced this code isn't the right thing, for the record, it just...makes me feel weird about the semantic meaning of and relationship between these two states...

drv/psc-seq-server/src/main.rs
```diff
@@ -366,30 +413,42 @@ enum PresentState {
     },
 
     /// The PSU has just appeared and we're waiting a bit to confirm that it's
-    /// stable before turning it on. (Waiting in this state provides some
-    /// debouncing for contact scrape.)
+    /// stable before turning it on (MWOCP68 only). (Waiting in this state
+    /// provides some debouncing for contact scrape.)
```
hawkw: Can we maybe explain in this comment the reason this behavior is MWOCP68 only? I realize it's because the '67 tries to turn itself on immediately, but it might be nice if the doc said that a bit more explicitly.

hawkw (commented):
Still have stuff I want to go through, but figured I'd leave the comments I have thus far.

drv/i2c-devices/src/mwocp6x/mwocp67.rs
evan-oxide: Yup the datasheet says that `CLEAR_FAULTS` is paged but `MB_PSU_SETTING` is not, that's why I picked it. And it has been working without us setting a page, so I think we're good.

drv/psc-seq-server/src/main.rs
evan-oxide: The state machine does wait for the debounce time for both PSUs, it's just that the MWOCP67 is automatically enabled before the debounce period and the MWOCP68 is manually enabled after. I'll make the comment clearer.

drv/psc-seq-server/src/main.rs
evan-oxide: I was thinking about that too. We could extend the time spent in the `NewlyInserted` state (`INSERT_DEBOUNCE_MS = 4000`) then transition directly to `On`, instead of going `NewlyInserted` -> `OnProbation` -> `On`. It would work correctly for the mwocp67, but the downsides are:
- After you plug in an mwocp68, it would sit idle for 4s before turning on. That's not terrible, but might be inconvenient or confusing for an operator.
- That approach (and the old code) don't leave any grace period between when the mwocp68 is enabled on insertion and when it's `PWR_OK` is required to be asserted. Apparently that's been working fine - either `PWR_OK` is always asserted very quickly, or we just haven't noticed that sometimes a hot-inserted mwocp68 triggers a fault recovery loop and takes an extra 5s to turn on. But using `OnProbation` means that both PSU types get a grace period, which seems nice and robust.

In terms of the semantic meanings of the states, in an ideal world I think the meanings would be:
   -  `NewlyInserted`: debouncing the contacts, not yet enabled.
   - `OnProbation`: enabled, not yet monitoring `PWR_OK`.
   - `On`: enabled, requiring `PWR_OK` to be asserted.

Of course the mwocp67 is enabled during `NewlyInserted`, but I'm treating that as an unfortunate accident. I think those definitions are still better and more consistent than:
   -  `NewlyInserted`: 
      - mwocp68: debouncing the contacts, not yet enabled.
      - mwocp67: enabled, not yet monitoring `PWR_OK`.
   - `On`: enabled, requiring `PWR_OK` to be asserted.

drv/psc-seq-server/src/bsp/observer_a.rs
evan-oxide: I don't think that's worth it. It would move the irritation to wherever we construct the different structs for each board.

drv/psc-seq-server/src/main.rs
evan-oxide: I ended up making it _less_ specific, does that work?

drv/i2c-devices/src/mwocp6x/mwocp67.rs
hawkw: Great!

drv/psc-seq-server/src/main.rs
hawkw: Thanks, I like the new comment here!

drv/i2c-devices/src/mwocp6x/mwocp68.rs
```diff
@@ -315,6 +315,11 @@ impl Mwocp68 {
             Operation::MfrMaxTemp3 => PmbusValue::from(
                 pmbus_read!(self.device, MFR_MAX_TEMP_3)?.get()?,
             ),
+            Operation::Operation => {
```
hawkw, line 318: lmao

drv/psc-seq-server/src/main.rs
```diff
@@ -183,24 +202,34 @@ counted_ringbuf!(Event, 128, Event::None, no_dedup);
...
         #[count(children)]
         psu: Slot,
     },
+    /// The new state of the PWR_OK pin.
+    PowerGoodChanged {
+        now: u64,
+        status: Status,
+        #[count(children)]
+        psu: Slot,
+    },
```
hawkw: So a downside to doing it like this instead of the previous `PowerGoodAsserted`/`PowerGoodDeasserted` is that this way, we don't have separate counters for `PWR_OK` and `PWRR_`...not `OK`. We will just have a counter for `PowerGoodChanged`, with child counters for each `Slot`. I think I would prefer the old structure if that's achievable.

drv/psc-seq-server/src/main.rs
```diff
@@ -248,17 +277,30 @@ enum Trace {
...
+}
+
+// Since entries in this ringbuffer contain timestamps, they will never be
+// de-duplicated. Thus, disable it.
+counted_ringbuf!(__TRACE, Trace, 32, Trace::None, no_dedup);
+
+/// I2c error logging is even more verbose than the `Trace` ring buffer because
+/// there are so many failed and retried i2c reads when a PSU is hot-plugged. We
+/// put them in their own ring buffer so they don't drown out all the other
+/// `Trace` events.
```
hawkw, line 374: good call

drv/psc-seq-server/src/main.rs
```diff
@@ -153,20 +170,22 @@ enum Event {
...
+    /// Emitted after failing to do an enable/disable action that was requested
+    /// by the state machine.
+    ActionFailed {
+        action: ActionRequired,
         now: u64,
         #[count(children)]
         psu: Slot,
-        present: bool,
+        err: mwocp6x::Error,
     },
```
hawkw: I feel like this might be nicer if we had a separate ringbuf variant for each `Action` that can have `Succeeded` or `Failed`, even though it will make the code that actually records them much grosser. This is mainly because it would be nice to have a separate counter for each action we perform with child counters for each PSU. Structuring it the way we are currently, we have counters of `ActionSucceeded` and `ActionFailed` with child counters for which slot succeeded and failed, but no way to count the individual actions.

Normally I would also suggest that this might make the ringbuf use less RAM, since we are using one enum discriminant instead of two. I almost said that here. Bu, then I remembered that we have ringbuf entries that have 12-byte serial numbers in them in this ringbuf, and that's almost certainly the long pole that determines how big the ringbuf entries are, so meh whatever.

drv/psc-seq-server/src/main.rs
hawkw: Okay, after reading what you wrote here, I think I'm sold on the current approach. It does feel a bit odd that the `NewlyInserted` state sometimes means "on" and sometimes means "not on" depending on which PSC...I almost wonder if we should just not use `NewlyInserted` on MWOCP67, or make the timeout 0s. I think the goal of the debouncing is to ensure the thing is fully plugged in before we try turning it on, but we don't turn it on ourselves in the MWOCP67. Anyway, I'm fine with leaving this as-is, I'm just rotating it in my mind a bit more.

drv/psc-seq-server/src/main.rs
```diff
@@ -525,7 +547,17 @@ fn main() -> ! {
...
                     serial: fruid.serial
                 });
-                PresentState::On { was_faulted: false }
+                // If this was a cold boot, then _maybe_ the PSU hasn't had time
+                // to fully power on and assert PWR_OK yet. We wouldn't want to
+                // interpret that as a fault and disable the PSU, so we ignore
+                // its PWR_OK for a little while by putting it in the
+                // OnProbation state. (We haven't actually seen this happen, but
+                // it seems possible since the MWOCP67 takes so long to assert
+                // PWR_OK.)
```
hawkw, line 641: I had to think through this for a sec to make sure it wouldn't stop the MWOCP68 from enabling PSUs, but it seems right to me. Good call!

drv/psc-seq-server/src/main.rs
```diff
@@ -554,66 +586,66 @@ fn main() -> ! {
...
+                {
+                    // The state machine doesn't know that we failed to
+                    // enable/disable the PSU as it requested, so its state will
+                    // be out-of-sync with reality. But this should resolve
+                    // itself eventually. If we fail to disable the PSU after a
+                    // fault, then either the fault recovery sequence will work
+                    // despite that, or it won't and we'll start another fault
+                    // recovery sequence. If we fail to enable the PSU at the
+                    // end of fault recovery, the deasserted OK signal will
+                    // trigger another fault recovery sequence soon.
```
hawkw: Not updating the state machine with the outcome it requested does feel a bit unfortunate, but I agree with your reasoning that this ought to be fine, and it will just go and do the right thing on the next times step regardless.

hawkw (approved):
Okay, after reading through this again, I feel good about the overall structure of this change, and it sounds like it's been adequately tested on both the MWOCP68 and MWOCP67 hardware. I left some minor comments: mainly, I think the ringbuf things are worth addressing if possible. But, I'm good with moving forward with this overall. Nice work here, this code is pretty complex to navigate!

drv/psc-seq-server/src/main.rs
evan-oxide: Having the state machine transition logic vary by PSU model would introduce a new type of complexity, and I think that complexity is more trouble than its worth (though I could be convinced otherwise).

drv/psc-seq-server/src/main.rs
evan-oxide: Good point, done. The grossness can be contained in a constructor.

drv/psc-seq-server/src/main.rs
evan-oxide: Done

PR #2607, opened 2026-07-23
Add new tokens to support authoried_keys

inickles:


plotnick:
Related: oxidecomputer/support-piv-tokens#10, oxidecomputer/omicron#10909. I have validated that the public keys here match the ones there.

jamesmunns (approved):

PR #2608, opened 2026-07-23
Check for overflow in update size calculations

evan-oxide:
Before, MGS could make the control-plane-agent task panic by claiming that the total size of an update was greater than `u32::MAX`. Now the SP replies with an "update is too large" error instead. Tested on a grapefruit.

### DIFF
--- a/task/control-plane-agent/src/mgs_compute_sled.rs
+++ b/task/control-plane-agent/src/mgs_compute_sled.rs
@@ -565,7 +565,7 @@ impl SpHandler for MgsHandler {
         update: SpUpdatePrepare,
     ) -> Result<(), SpError> {
         ringbuf_entry_root!(Log::MgsMessage(MgsMessage::UpdatePrepare {
-            length: update.aux_flash_size + update.sp_image_size,
+            length: update.aux_flash_size.saturating_add(update.sp_image_size),
             component: SpComponent::SP_ITSELF,
             id: update.id,
             slot: 0,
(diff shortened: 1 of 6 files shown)


task/control-plane-agent/src/mgs_compute_sled.rs
```diff
@@ -565,7 +565,7 @@ impl SpHandler for MgsHandler {
         update: SpUpdatePrepare,
     ) -> Result<(), SpError> {
         ringbuf_entry_root!(Log::MgsMessage(MgsMessage::UpdatePrepare {
-            length: update.aux_flash_size + update.sp_image_size,
+            length: update.aux_flash_size.saturating_add(update.sp_image_size),
```
jamesmunns: Do we want this to be saturating add? I guess it's just for the log, but it seems odd to me that we do a saturating add here for the log and then do the actual check later.

Maybe change the `UpdatePrepare` ringbuf message to track each size individually, or note here that we'll do the actual overflow check later in `fn prepare`.

jamesmunns (approved):
Overall looks good to me, one note about saturating in one place for the logs and checking in another.

task/control-plane-agent/src/mgs_compute_sled.rs
evan-oxide: Logging the sizes individually does sidestep the question nicely. I had to split `UpdatePrepare` into two different variants, which is a good change anyway because it was being used for two different MGS messages.

Issue #2603, opened 2026-07-21
Generate ereport if fan tachometer indicates fan failure


jamesmunns:
Currently, we generate a ringbuf entry when the fan tachometer falls to 0 RPM, and is not showing the "power on reset" value:

https://github.com/oxidecomputer/hubris/blob/53a149bc5f3451b1e1ce030a7a0043012d0d57e2/drv/i2c-devices/src/max31790.rs#L324-L335

We likely want to escalate this to an ereport when this first occurs, as an edge-triggered report.

We may want to consider what we count as the lower threshold of "deviancy", according to a [recent FAQ friday](https://bsky.app/profile/oxide.computer/post/3mqtzifq7ao2h), our fans should be idling at around 2000 RPM. We should confirm this is the case, and pick a reasonable number below that (1000 RPM? 1500 RPM?) that is very unlikely to occur in a happily operating system. This may also require some debouncing or filtering to pick the right number.

We may *also* want to report if the fan RPM exceeds some reasonable upper limit as well.

jamesmunns:
One potential question is whether we should be "noticing" that a fan is deviant in the i2c drivers themselves (as linked above), or in the thermal task.

The latter is probably better empowered to retain state (and potentially not report when multiple fans are failing), but currently the `thermal` loop doesn't particularly *care* about the RPMs, it just reports it to the sensor_api.

hawkw:
I would probably have the thermal task be responsible for the ereport. This is mostly because it feels right at a "vibes" level: it feels like the right separation of concerns, the I2C drivers are intended to be "mechanism" rather than policy. There are also some practical reasons, mostly in the context of future work:

- eventually we may want to track fan state in thermal so we can implement changes in thermal behavior in the event of a fan failure (i.e. putting the other fans at max PWM to compensate).
- we would also like the thermal task to emit ereports for other thermal events, and putting it there probably results in less duplication of the ereport code across a whole bunch of tasks

jamesmunns:
Looking at the design of this, there are a couple of ereports I could see us emitting:

## Kinds of ereports

I think there's probably some value in separating the following potential cases to report about:

1. Fans added/removed (for devices with removable fans)
    * NOTE: non-removable fans still emit a ringbuf entry on first presence poll, do we want that?
2. Fans go from "nominal" to "deviant" or "deviant" to "nominal", including:
    * **Too Fast**: could be over the "total" max speed or over the "expected" current speed
[...]

PR #2609, opened 2026-07-23
Refactor of the `thermal` task

jamesmunns:
This is a WIP refactor in service of simplifying the current state tracking performed by the thermal task.

Some of this is still Not Great, particularly the places where I pass 2-4 closures into a single function.

However, by pushing most state tracking down into the BSP, I hope to make it easier to Trait-ify the BSP interface and make it easier to build a "host native" thermal loop for testing purposes.

(most of the builds are failing because I've only refactored `cosmo` and none of the other bsps yet)


task/thermal/src/bsp/common/emc2305.rs
```diff
@@ -0,0 +1,132 @@
...
+        })
+    }
+}
+
+/// Helper function to retry initialization several times, logging errors
+pub(crate) fn retry_init<F: FnMut() -> Result<(), ControllerInitError>>(
+    mut init: F,
+) {
+    // When we first start up, try to initialize the fan controller a few
+    // times, in case there's a transient I2C error.
```
labbott, line 106: Is this still the case? I see this was originally added for grapefruit as part of `725bdfdd8b01de8e095199ee3f0ee218ffe71414`

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,225 @@
...
+                        | ChannelType::RemovableAndErrorProne
+                );
+                let removed =
+                    e == SensorReadError::I2cError(ResponseCode::NoDevice);
+                let unexpected_failure = !(removable && removed);
+                if unexpected_failure {
+                    InputReadingOutcome::UnacceptableMissing { id, err: e }
+                } else {
+                    InputReadingOutcome::AcceptableMissing { id, err: e }
+                }
```
labbott: The comments are very good but I still find this tricky to follow. I don't have any great ideas here (might take a pass later)

labbott (commented):
I forgot to hit submit yesterday, I'll take another pass today

task/thermal/src/bsp/common/i2c_temp_input.rs
jamesmunns: I think this could be simplified IMO, this is a little awkward in service of preserving the old behavior.

jamesmunns:
@labbott if you're making another pass, https://github.com/oxidecomputer/hubris/pull/2609/changes/5c803a76e4db05edf8505ef1c5c64ce518b1751b added some trait docs as well as some overall concepts, which might be useful to read first!

task/thermal/src/bsp/common/emc2305.rs
jamesmunns: I'm not sure! I haven't actually tested this on hardware yet, and tried as best as possible to preserve the existing behavior, whatever it was.

jamesmunns:
Taking this out of draft/WIP state, and marking it as ready for review. Builds are now passing for all targets, and I think this change is Good Actually, even if a bit large.

It definitely need on-target testing still, but I'll need to coordinate that. It would be good to make sure there aren't any major review items to address first.

jamesmunns:
Smoke testing on Hubris seems to work, the system is unloaded so there's not a lot of heat to react to, but the current software handles the transitions from A2 -> A0 and A0 -> A2 as I expect, and `ControlPwm` is going up:

```
   TOTAL VARIANT
     632 ControlPwm
       4 AutoState(Boot)
       3 AutoState(Running)
       6 FanAdded
       3 PowerModeChanged
       1 Start
       1 ThermalMode(Auto)
       1 SensorReadFailed
       1 FanControllerInitialized
       1 SetFanWatchdogOk
   0  393        1        1 Start
   1   72        1        1 FanControllerInitialized
   2  198        1        1 ThermalMode(Auto)
   3  676        1        1 AutoState(Boot)
   4  692        1        1 SetFanWatchdogOk
   5  704        1        1 FanAdded(0x0)
   6  704        1        1 FanAdded(0x1)
   7  704        1        1 FanAdded(0x2)
   8  704        1        1 FanAdded(0x3)
   9  704        1        1 FanAdded(0x4)
  10  704        1        1 FanAdded(0x5)
  11  808        1        1 PowerModeChanged(PowerBitmask(0b1))
  12  676        1        1 AutoState(Boot)
  13 1000        1        1 AutoState(Running)
  14  967        1      322 ControlPwm(0x0)
  15  775        1        1 SensorReadFailed(SensorId(0x37), I2cError(NoDevice))
  16  808        1        1 PowerModeChanged(PowerBitmask(0b10))
  17  676        1        1 AutoState(Boot)
  18  967        1        1 ControlPwm(0x64)
  19 1000        1        1 AutoState(Running)
  20  967        1      303 ControlPwm(0x0)
  21  808        1        1 PowerModeChanged(PowerBitmask(0b1))
  22  676        1        1 AutoState(Boot)
  23 1000        1        1 AutoState(Running)
  24  967        1        6 ControlPwm(0x0)
```

drv/i2c-devices/src/emc2305.rs
```diff
@@ -148,6 +148,27 @@ pub const MAX_FANS: u8 = 5;
...
 
+impl Fan {
+    /// Create a new fan
+    ///
+    /// This panics if idx is out of range, and is intended to be used
+    /// at compile time for building constant values. Prefer `TryFrom`
+    /// at runtime when handling user provided data.
+    pub const fn new_const(idx: u8) -> Self {
+        if idx >= MAX_FANS {
+            panic!("Out of range!");
```
labbott, line 159: I don't love the extra potential runtime panic but it also didn't seem to increase the code size so 🤷

task/thermal/src/bsp/common/emc2305.rs
```diff
@@ -0,0 +1,151 @@
...
+
+use crate::{
+    __RINGBUF, Trace,
+    control::{Fan, FanStatus},
+};
+
+/// Tracks whether a Emc2305 fan controller has been initialized, and
+/// initializes it on demand when accessed, if necessary.
+///
+/// This is copy-pasted from [`Max31790`]
```
labbott: Still a useful comment?

task/thermal/src/bsp/common/max31790.rs
```diff
@@ -0,0 +1,158 @@
...
+//! Common types and helpers for Max31790 Fan Controller
+
+use drv_i2c_api::{I2cDevice, ResponseCode};
+use drv_i2c_devices::max31790::Max31790;
+use ringbuf::ringbuf_entry;
+use task_sensor_api::SensorId;
+use task_thermal_api::{SensorReadError, ThermalError};
+
+use crate::{
+    // control::{ControllerInitError, retry_init},
```
labbott: dead code

task/thermal/src/bsp/minibar.rs
```diff
@@ -43,71 +20,119 @@ bitflags::bitflags! {
...
+    const USE_CONTROLLER: bool = false;
+
+    const PID_CONFIG: PidConfig = PidConfig {
+        zero: 0.,
+        gain_p: 0.,
+        gain_i: 0.,
+        gain_d: 0.,
+        min_output: 0.,
+        max_output: 100.,
+    };
```
labbott, line 37: We lost the comment about having no fans which I think is important to explain these settings

task/thermal/src/control.rs
```diff
@@ -1758,20 +1217,146 @@ impl<'a> ThermalControl<'a> {
...
+///   these sensors are also reported to the Sensors API.
+/// * Misc Sensors: Temperature sensors that are actively polled by the thermal
+///   control loop and reported to the Sensors API, but are not used as inputs
+///   to the PID control loop. Misc sensors are active in all power states.
+/// * Dynamic Inputs: Temperature sensors that are NOT actively polled by the
+///   thermal control loop, from which readings are instead obtained by querying
+///   the sensor API. These readings are used as inputs to the PID control loop.
+///   By default, all Dynamic Inputs are not marked as present, and require an
+///   external command (via IPC) to provide the necessary thermal model, and
+///   inform the control loop that the sensors are active and should be queried.
```
labbott: I think both Misc Sensors and Dynamic inputs could use one concrete example

labbott (commented):
This is a big diff but I do think it's an improvement. I think keeping the existing behavior is the correct approach. I would also like to capture areas that are particularly weird and you think should be improved (maybe a single issue with some links code lines?)

drv/i2c-devices/src/emc2305.rs
jamesmunns: This is a const-fn, and is only currently used to generate constants, which means that any panics that occur would be compile-time errors.

You are right, if someone inadvertently calls it at runtime, it would potentially panic in that case.

task/thermal/src/bsp/common/emc2305.rs
jamesmunns: It was there before, but I can probably remove it now!

task/thermal/src/bsp/common/emc2305.rs
jamesmunns: Also worth noting, thermal is only active for grapefruit-ruby, and I'm not sure where to even obtain one of those for testing.

task/thermal/src/bsp/minibar.rs
jamesmunns: Let me look at this and see if it would be feasible to change this to `Option<PidConfig>` to merge `PID_CONFIG` and `USE_CONTROLLER` into a single entity. That might make this less awkward.

If not I'll find and restore the comment!

jamesmunns:
Capturing a list of "some things that are 'eh' about the current state, and would like to fix later, but didn't because I didn't want to churn the diff even more":

## All the `dead_code` annotations

Right now, we don't use proper features for a bunch of things, like fans, fan controller types, etc., and some things are only used in some bsps.

I think this would be fixed by either:

1. Splitting thermal-core out into a crate with pub items, making the dead-code no longer relevant
2. Adding proper feature flagging to thermal

## The `#[path]` module stuff for `bsps/common`

https://github.com/oxidecomputer/hubris/pull/2609/changes#diff-a9b9fa16760e6bde87da1614252f333f4eef08e39dba9c5e1697fc1fc3e65f05R20-R29

This is awkward. It should probably be feature-flagged or somewhere that multiple BSPs can use it. For now it's an awkward pseudo thing.

tbf the thermal task was already doing this for bsps. I think that need to be fixed too, same options as above

## Be okay with some changes to ringbufs

Some of the trait API surface was made pretty much exclusively to keep us able to do ringbuf logging in the same/place way as we currently do. A lot of this is "log an event when updating status/presence". We might decide it is okay to ringbuf this differently, or let some of that information live in the bsps (or the state types like `TemperatureInput`s/`Fan`s) themselves).

This also includes things [like this](https://github.com/oxidecomputer/hubris/pull/2609/changes#diff-790d4852cbe91dad37420685649135e900fc7fedc897c1bb368b21e3c6b05ad5R193-R238).

That being said, where we ringbuf today, we probably want to ereport [tomorrow](https://github.com/oxidecomputer/hubris/issues/2603), so that might inform what is/isn't reasonable.

## Revisit how we do state transitions

https://github.com/oxidecomputer/hubris/pull/2609/changes#diff-f4eab2a6583ffdb099d7d21aff862a465dc6db83665b99618329f357f8a1cad2R891-R893

This could be a bit simpler I think, where we check for the "deviant" states in descending importance, and go directly there vs matching on the current state.

This isn't that big of a deal though, might just dedupe some repeated code.

## Duplication between emc2305 and max31790

These are very similar, there might be a way to dedupe these better with a trait or something, and have the bsps use that. Idk, I ran out of time, but it's also not hugely problematic IMO.

## Duplication/verbosity across bsps

They are currently a little wordy, with a couple of impls being *almost* the same. Again, with some more time we might find a better to "compose" the common parts, but the total volume of bsps is pretty low, and it seemed like the current amount of duplication was within the threshold of acceptable.

## Things that still need to be done to make this portable

1. I need to do something with `SeqError` and `PowerBitmask`, these are bsp-specific types, and they probably need to be surfaced as associated types on the trait.
2. I probably need to make the sensor-api stuff and maybe ringbufs a separate trait so we can replace them in hosted configurations.
3. Actually pull out most of `control.rs` into a `thermal-core` crate, use that from the thermal task

Once we do all that, I *think* we will be at a place where the logic of `thermal` is largely testable!

jamesmunns:
> I would also like to capture areas that are particularly weird and you think should be improved (maybe a single issue with some links code lines?)

Wrote up https://github.com/oxidecomputer/hubris/pull/2609#issuecomment-5120333046 for now, I can turn that into an issue (or issues) with correct line links once this merges!

jamesmunns:
@labbott addressed most of the comments so far!

task/thermal/src/bsp/common/emc2305.rs
```diff
@@ -0,0 +1,149 @@
...
+//! Common types and helpers for Emc2305 Fan Controller
+
+use drv_i2c_api::{I2cDevice, ResponseCode};
+use drv_i2c_devices::emc2305::Emc2305;
+use ringbuf::ringbuf_entry;
+use task_sensor_api::SensorId;
+use task_thermal_api::{SensorReadError, ThermalError};
+
+use crate::{
+    __RINGBUF, Trace,
```
hawkw: FWIW, the `ringbuf` crate also has [a `ringbuf_entry_root!` macro](https://github.com/oxidecomputer/hubris/blob/a41448f5a2fd6e291aa8b3eb6bfbc6c0763ed992/lib/ringbuf/src/lib.rs#L473-L485) which is intended to allow submodules to add entries to ringbufs declared in the crate root without having to import `__RINGBUF`. Not a big deal, as what you're doing also works fine in this case (as there is no additional ringbuf declared in this module, which `ringbuf_entry_root!` helps with). I just figured I'd mention it.

I've also seen code `use ringbuf::ringbuf_entry_root as ringbuf_entry;` in cases where a module doesn't have its own ringbuf. I dunno whether or not I love that pattern, but it's an existing convention... 🤷‍♀️

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,242 @@
...
+#[allow(dead_code, clippy::upper_case_acronyms)]
+pub enum Device {
+    Tmp117,
+    Tmp451(drv_i2c_devices::tmp451::Target),
+    CPU,
+    Dimm,
+    U2,
+    M2,
+    LM75,
+}
```
hawkw, line 35: turbo nitpick: I think the two "upper case acronyms" clippy doesn't like here are `LM75` and `CPU`; what do you think about just doing this instead:
```suggestion
#[allow(dead_code)]
pub enum Device {
    Tmp117,
    Tmp451(drv_i2c_devices::tmp451::Target),
    Cpu,
    Dimm,
    U2,
    M2,
    Lm75,
}
```

I note that every other acronym in this enum (TMP117, TMP451, DIMM) are camel-cased...we could also fully embrace upper-case acronyms here, but if we're going to do that, I'd kinda like to see DIMM and CPU in the same case?)

hawkw: also, take it or leave it, but I kind of wonder if we ought to change this so that it's the name of the actual I2C device we use for each thing (i.e. `Tse2005Av` instead of `Dimm`, `Sbtsi` instead of `CPU`, and `NvmeBmc` instead of `U2` and `M2`)? i'm thinking that because this enum is really being used to determine which I2C device's driver to use, rather than which component's thermal model to use (AFAICT?), we might be better off making that a bit more explicit --- in this case, the `U2` and `M2` variants don't actually seem necessary, since they're both just NVME BMC devices, and I can kind of imagine some future product where DIMM and/or CPU temperatures are read through a different I2C device (though we're probably not going to release, say, an intel compute sled any time soon, so that's probably not worth optimizing for...)

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,242 @@
...
+
+pub(crate) struct InputChannelMetadata {
+    /// Temperature sensor
+    sensor: TemperatureSensor,
+
+    /// Thermal properties of the associated component
+    model: ThermalProperties,
+
+    /// Mask with bits set based on the Bsp's `power_mode` bits
```
hawkw: nitpick:
```suggestion
    /// Mask with bits set based on the BSP's `power_mode` bits
```

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// InputChannelMetadata is the constant description portion of an InputChannel.
+///
+/// We split it off because InputChannel is mutable to contain the last state,
+/// and if we left it inlined, then we would end up including all of this
+/// metadata in RAM, despite never changing it! So instead, we break it off and
+/// have InputChannel hold an `&'static` reference instead, so we only waste
+/// a wee little pointer (4 bytes) to flash space in each InputChannel entry,
+/// instead of (at the time of writing) 36 bytes, which for example in Cosmo
+/// that has 14 input channels, is over 500 bytes of RAM!
```
hawkw: style nit: this feels like it ought to be on the type declaration, rather than on the `impl` block?

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,242 @@
...
+impl InputChannelMetadata {
+    pub const fn new(
```
hawkw: style nit: the type is `pub(crate)` but the constructor is `pub` (i thought clippy didn't like this, but i guess it doesn't seem to care...)
```suggestion
    pub(crate) const fn new(
```

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,242 @@
...
+
+impl InputChannel {
+    pub const fn new(metadata: &'static InputChannelMetadata) -> Self {
+        Self {
+            metadata,
+            last_reading: TemperatureReading::NotYetQueried,
+        }
+    }
+
+    pub fn has_been_queried(&self) -> bool {
```
hawkw: style nit:
```suggestion
    pub(crate) const fn new(metadata: &'static InputChannelMetadata) -> Self {
        Self {
            metadata,
            last_reading: TemperatureReading::NotYetQueried,
        }
    }

    pub(crate) fn has_been_queried(&self) -> bool {
```

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,242 @@
...
+            TemperatureReading::Unpowered => true,
+            TemperatureReading::Disconnected => true,
+            TemperatureReading::ValidAtLeastOnce(..) => true,
+        }
+    }
+
+    /// Get current stored status.
+    ///
+    /// Returns None if we do not have a reading stored.
+    pub fn status(&self) -> Option<InputStatus<'_>> {
```
hawkw: nit:
```suggestion
    pub(crate) fn status(&self) -> Option<InputStatus<'_>> {
```

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,242 @@
...
+                self.last_reading = TemperatureReading::ValidAtLeastOnce(
+                    TimestampedTemperatureReading {
+                        time_ms: now,
+                        value,
+                    },
+                );
+                InputReadingOutcome::Success { id, now, value }
+            }
+            Err(e) => {
+                // This is mimicing the old state value logic for deciding if
```
hawkw: spelling nit:
```suggestion
                // This is mimicking the old state value logic for deciding if
```

task/thermal/src/bsp/common/i2c_temp_input.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    pub fn reset_value(&mut self, mode: PowerBitmask) {
+        if !mode.intersects(self.metadata.power_mode_mask) {
+            self.last_reading = TemperatureReading::Unpowered;
+        } else {
+            self.last_reading = TemperatureReading::NotYetQueried;
+        }
+    }
+
+    pub fn do_reading(
```
hawkw: nit: IMO, it would be nice if there was a doc comment on this explaining its relationship with `status()` --- it looks like `do_reading` actually asks the temperature sensor for what temperature it is, but returns a thing that represents the result of the I2C request, and `status()` returns a thing representing the current temperature or its non-existence, but I'd like to see that written down.

task/thermal/src/bsp/common/i2c_temp_input.rs
hawkw: also, take it or leave it, but I'm not sure if i _love_ that the way that we actually get the temperature of a thing is in a method called `status()`...i think it's technically _correct_ to call it that, because the temperature is part of the status of the device, but i find it a bit weird, part of me kind of expects "status" to be the thing that tells you that the device is disconnected/errored/ok, but not how you actually get the temperature from it. with that said, i think coupling these two things is good, so...i don't know. maybe ignore me.

task/thermal/src/control.rs
```diff
@@ -220,21 +110,30 @@ pub(crate) enum ChannelType {
...
+    /// and not there, or Removable and Error Prone and had any kind of error.
+    AcceptableMissing { id: SensorId, err: SensorReadError },
+    /// Any read error that didn't match the "Acceptable" cases listed above.
+    UnacceptableMissing { id: SensorId, err: SensorReadError },
+    /// We read the data! Hooray!
+    Success {
+        id: SensorId,
+        now: u64,
+        value: Celsius,
+    },
```
hawkw, line 320: this is the type of thing where normally I would say "huh, it seems kind of like it might be easier to use this if we pulled the `SensorId` out into a struct so that you didn't have to unpack it from each variant"...but after actually looking at the code, I see that it doesn't matter and it's probably simpler to just leave it this way. carry on.

task/thermal/src/control.rs
```diff
@@ -2,184 +2,74 @@
...
+        sensor_id: SensorId,
+    },
+    PresentError {
+        error: SensorReadError,
+        sensor_id: SensorId,
+    },
+    #[allow(dead_code)] // Some bsps don't have removable fans
+    NotPresent {
+        sensor_id: SensorId,
+    },
```
hawkw: very unimportant style nit: in `InputReadingOutcome` and `InputStatus`, the sensor ID field is called `id`; here, it's called `sensor_id`. would maybe be worth making it consistent everywhere

task/thermal/src/control.rs
```diff
@@ -250,8 +149,56 @@ impl InputChannel {
 /// many of them could be present, but their thermal properties could vary
 /// depending on what's plugged in.
 #[derive(Clone, Copy)]
+#[allow(dead_code)] // Not all bsps have dynamic inputs
 pub(crate) struct DynamicInputChannel {
-    model: ThermalProperties,
+    pub sensor_id: SensorId,
```
hawkw, line 351: similarly, `InputStatus` calls this `id`, while `DynamicInputChannel` calls it `id`...maybe worth using the same name?

task/thermal/src/control.rs
```diff
@@ -250,8 +149,56 @@ impl InputChannel {
...
+impl DynamicInputChannel {
+    pub(crate) const fn new(sensor_id: SensorId) -> Self {
+        Self {
+            sensor_id,
+            model: ThermalProperties {
+                target_temperature: Celsius(0.),
+                critical_temperature: Celsius(0.),
+                power_down_temperature: Celsius(0.),
+                temperature_slew_deg_per_sec: 0.,
+            },
```
hawkw: I get why we're representing the thermal model of a dynamic input which hasn't been added as a `ThermalProperties` where everything is 0, and it saves at least a byte of enum discriminant and probably some padding versus making this an `Option`. But, it really doesn't _feel_ good. We are initializing this to an invalid state: an input channel that is at its target temperature, its critical temperature, _and_ its power-down temperature, all at the same time. And we're just relying on other code respecting the `has_been_queried` flag to avoid making the system thermal shutdown immediately...which it _does_, but it feels sketchy to me.

If we're going to do this, I'd feel a bit less sketched out if we initialized these to NaN instead of 0. That way, at least the comparisons against the threshold values would always compare false while we're in the uninitialized state, which feels less scary than initializing all dynamic channels to always be over the power-down temp...

hawkw (commented):
I haven't fully read through this change, but I wanted to leave the notes I have so far. Overall, I like the refactor, but have some nitpicks.

task/thermal/src/bsp/common/i2c_temp_input.rs
hawkw: ah, i see that this was copied directly from the old code, it just lived in a different file. in that case, feel free to ignore my nits for now, though it might be worth cleaning up later...

task/thermal/src/control.rs
```diff
@@ -491,19 +289,28 @@ pub(crate) struct ThermalControl<'a> {
 /// Represents the state of a temperature sensor, which either has a valid
 /// reading or is marked as inactive (due to power state or being missing)
 #[derive(Copy, Clone, Debug)]
-enum TemperatureReading {
-    /// Normal reading, timestamped using monotonic system time
-    Valid(TimestampedTemperatureReading),
+#[allow(dead_code)] // Not all bsps have inputs!
```
hawkw, line 491: diversity is an Oxide value :)

task/thermal/src/bsp/common/i2c_temp_input.rs
jamesmunns: Will leave as-is for now, will consider poking at a later date.

task/thermal/src/control.rs
jamesmunns: Resolved

task/thermal/src/bsp/common/i2c_temp_input.rs
jamesmunns: Reworked in https://github.com/oxidecomputer/hubris/pull/2609/commits/e478bc31aa583602867114f2a301cb6fa9bbc2c4

task/thermal/src/bsp/common/i2c_temp_input.rs
jamesmunns: Reworked in https://github.com/oxidecomputer/hubris/pull/2609/commits/e478bc31aa583602867114f2a301cb6fa9bbc2c4, which hopefully switched to more consistent verbiage to make this easier to follow, also take a look at the trait definition, which does describe this IMO.

hawkw: Maybe that's a little bit on me for reading this in the order that GitHub showed me the files, instead of looking at the trait first... :)

labbott (approved):

task/thermal/src/control.rs
hawkw, line 1: @jamesmunns one note is that I feel like this module could really use a high-level theory of operation type comment explaining the relationship between all the different input types and input state types? I realize this was not there *before* you showed up, but part of the reason the previous code was hard to understand was that it never really had an overview of how all these pieces fit together. I feel like the refactor presents a nice opportunity to add some overview documentation while it's all fresh in your head. What do you think?

task/thermal/src/control.rs
```diff
@@ -250,8 +161,52 @@ impl InputChannel {
...
 #[derive(Clone, Copy)]
+#[allow(dead_code)] // Not all bsps have dynamic inputs
 pub(crate) struct DynamicInputChannel {
-    model: ThermalProperties,
+    pub sensor_id: SensorId,
+    pub state: DynamicTemperatureState,
+}
+
+/// Represents the state of a temperature sensor, which either has a valid
+/// reading or is marked as inactive (due to power state or being missing)
```
hawkw: this is specific to _dynamic_ temperature inputs, right?
```suggestion
/// Represents the state of a dynamic temperature sensor (which are added 
/// and removed at runtime by IPCs from outside the thermal loop). Such a 
/// sensor either has a valid reading or is marked as inactive (due to power 
/// state or not having been added to the thermal loop).
```

task/thermal/src/control.rs
```diff
@@ -1122,17 +708,19 @@ impl<'a> ThermalControl<'a> {
...
-                    {
-                        ringbuf_entry!(Trace::FanRemoved(fan));
+        match self.bsp.poll_fan_presence() {
+            Ok(piter) => {
+                for fan in piter {
+                    match fan {
+                        FanPresence::Present { fan_id, new } if new => {
+                            ringbuf_entry!(Trace::FanAdded(fan_id));
+                        }
+                        FanPresence::NotPresent { fan_id, new } if new => {
```
hawkw: perhaps nitpicky, but I'm not sure if I love naming this "new" when it means a fan was just *removed*...to me "new" implies that something was *added*, specifically. what do you think about calling it "changed" instead?

task/thermal/src/control.rs
```diff
@@ -1122,17 +708,19 @@ impl<'a> ThermalControl<'a> {
...
+                            ringbuf_entry!(Trace::FanRemoved(fan_id));
+                        }
```
hawkw, line 909: it is worth noting that this kind of implies that it is now the BSP, rather than the control loop, which is responsible for tracking fan presence states across polls, so that it can indicate whether or not the presence change is "new". I don't think that's terrible, but it _does_ mean that if we are testing the control loop code in simulation eventually, the logic for determining if presence has changed is not being tested, since it lives in the BSP, which is the part we're simulating.

task/thermal/src/control.rs
```diff
@@ -1190,38 +773,33 @@ impl<'a> ThermalControl<'a> {
...
+                    value,
+                } => {
+                    self.sensor_api.post(sensor_id, value.0, now);
+                }
+                InputPollingOutcome::AcceptableMissing { sensor_id, err } => {
+                    self.sensor_api.nodata_now(sensor_id, err.into());
+                }
+                InputPollingOutcome::UnacceptableMissing { sensor_id, err } => {
+                    // Record an error errors if the sensor is not removable
+                    // or we get a unexpected error from a removable sensor
```
hawkw: typo:
```suggestion
                    // Record an error if the sensor is not removable, or if
                    // we got an unexpected error from a removable sensor
```

task/thermal/src/control.rs
```diff
@@ -1235,135 +813,118 @@ impl<'a> ThermalControl<'a> {
...
-                    if let Ok(r) = self.sensor_api.get_reading(*sensor_id) {
-                        self.state.write_temperature(index, r);
-                    }
-                }
-                None => self.state.write_temperature_inactive(index),
+        // `register_dynamic_input` and `remove_dynamic_input`.
+        //
+        // TODO(AJM): Should we be doing this in `read_sensors` instead of
+        // `run_control`?
+        self.bsp.poll_dynamic_inputs(&self.sensor_api);
```
hawkw, line 1065: conceptually, that does sound like that's where this belongs --- I presume that we're leaving this here because it's what the old code did and we're uncomfortable about potentially changing behavior?

task/thermal/src/control.rs
```diff
@@ -1235,135 +813,118 @@ impl<'a> ThermalControl<'a> {
...
+
+        // Run a common analysis pass first, regardless of state. Don't take any
+        // side effectful actions yet though.
+        let mut all_nominal = true;
+        let mut any_power_down = None;
+        let mut any_critical = None;
+        let mut worst_margin = f32::MAX;
+        let all_inputs_queried = self.bsp.all_inputs_queried();
+
+        // TODO(AJM): Assert all queried if !Boot
```
hawkw: it looks like we are now doing that, so this todo can probably be removed?
```suggestion
```

task/thermal/src/control.rs
```diff
@@ -1235,135 +813,118 @@ impl<'a> ThermalControl<'a> {
...
+                // We allow boot to have not yet queried all items successfully
+            }
+            ThermalControlState::Running { .. }
+            | ThermalControlState::Critical { .. }
+            | ThermalControlState::FanParty => {
+                // TODO: Should this just move us back to the Boot state? This
+                // should not be possible by construction, as moving from
+                // Invalid to Valid reading is "latching" unless we reset the
+                // state.
+                assert!(all_inputs_queried);
```
hawkw: panicking here does *basically* seem to be "moving us back to the boot state but with more steps", so maybe we should be doing that? Thinking through it: 

Pros of panicking:
- The panic totally resets the thermal loop's state, rather than just changing this to "boot"
- The panic takes a dump so that we have a complete snapshot of what the error condition looked like, which could be helpful for debugging

Cons of panicking: 
- **If whatever thing causes us to not poll a sensor that we expected to have polled is 'sticky' upon restarting the task, we will panic again when we're restarted, and we make this assertion _before_ we have ever actually run the thermal loop.** IIUC, this basically means hat a programmer error that results in us failing to query any particular input can basically permanently kill the thermal loop. That seems...not great (although fortunately we have a fan controller watchdog we'll probably just fail to pet here --- the failure mode is that we're stuck in full fan mode forever).

I kinda think we should probably just be emitting a ringbuf entry (with a counter, so we can see it forever), and continuing, or *possibly* going back to the `Boot` state here.

task/thermal/src/control.rs
```diff
@@ -1235,135 +813,118 @@ impl<'a> ThermalControl<'a> {
...
+                let worst_case = reading.worst_case(now_ms, model);
+                let temperature = worst_case.worst_case_temp;
+                all_nominal &= model.is_nominal(temperature);
+                if model.should_power_down(temperature) {
+                    any_power_down = Some((sensor_id, worst_case));
+                }
+                if model.is_critical(temperature) {
+                    any_critical = Some((sensor_id, worst_case));
                 }
+                worst_margin = worst_margin.min(model.margin(temperature).0);
```
hawkw: super nitpicky, but I feel like personally I might have put the comment on the part where we actually compare the margins:
```suggestion
        self.bsp.all_active_inputs().for_each(
            |ActiveInputState {
                 sensor_id,
                 reading,
                 model,
             }| {
                let worst_case = reading.worst_case(now_ms, model);
                let temperature = worst_case.worst_case_temp;
                all_nominal &= model.is_nominal(temperature);
                if model.should_power_down(temperature) {
                    any_power_down = Some((sensor_id, worst_case));
                }
                if model.is_critical(temperature) {
                    any_critical = Some((sensor_id, worst_case));
                }

                // Remember, positive margin means that all parts are happily
                // below their max temperature; negative means someone is
                // overheating.  We want to pick the _smallest_ margin, since
                // that's the part which is most overheated.
                worst_margin = worst_margin.min(model.margin(temperature).0);
```

task/thermal/src/control.rs
```diff
@@ -1235,135 +813,118 @@ impl<'a> ThermalControl<'a> {
...
+            return Ok(self.transition_to_uncontrollable_due_to(due_to, now_ms));
+        }
 
-                if let Some(due_to) = any_power_down {
-                    self.transition_to_uncontrollable_due_to(due_to, now_ms)
-                } else if let Some(values) = values.as_temperature_array() {
-                    self.transition_to_running(worst_margin, now_ms, values)
+        let all_nominal = all_nominal;
+        let any_critical = any_critical;
+        let worst_margin = worst_margin;
```
hawkw: this shadowing is essentially just intended to declare "beyond here we will never mutate these again"?

task/thermal/src/control.rs
hawkw: also...I _love_ seeing that the common logic for scanning all the temperature readings is now no longer repeated almost-but-not-entirely-identically across every state. That's something I kept wanting to do while working on this code in past but never got around to. Thanks for fixing it!

task/thermal/src/control.rs
```diff
@@ -1235,135 +813,118 @@ impl<'a> ThermalControl<'a> {
...
+
+        // TODO(AJM): I think we could dedupe some of the code below, basically
+        // working backwards and checking if we "qualify" for each state, though
+        // that's a bit more invasive of a change
```
hawkw, line 1133: yeah, I think that would be interesting to do once we have some testing wih the current thing. maybe worth opening an issue describing this change?

task/thermal/src/control.rs
```diff
@@ -1381,94 +942,51 @@ impl<'a> ThermalControl<'a> {
...
+            ThermalControlState::Uncontrollable => ControlResult::PowerDown,
+        })
+    }
 
+    /// An extremely simple thermal control loop.
+    ///
+    /// Returns an error if the control loop failed to read critical sensors;
+    /// the caller should set us to some kind of fail-safe mode if this
+    /// occurs.
+    pub fn run_control(&mut self) -> Result<(), ThermalError> {
```
hawkw: this is a super unimportant nit, so feel free to ignore me, but personally, i like to sort of topologically sort methods in Rust so that the public "entry points" are *above* the internal functions they call into, so the reader can sort of read through the logic in top-to-bottom order...so what i'm really saying is that I might have put this _above_ `run_control_inner`. but, that would be a bigger diff, so, it's up to you. 🤷‍♀️

task/thermal/src/control.rs
```diff
@@ -1711,44 +1200,35 @@ impl<'a> ThermalControl<'a> {
...
-        if index >= bsp::NUM_DYNAMIC_TEMPERATURE_INPUTS {
-            return Err(ThermalError::InvalidIndex);
-        }
         // If we're adding a new dynamic input, then reset the state to `Boot`,
         // ensuring that we'll wait for that channel to provide us with at least
         // one valid reading before resuming the PID loop.
-        if self.dynamic_inputs[index].is_none() {
+        //
+        // TODO(AJM): We just ignore it if there was already a dynamic input
+        // there already?
```
hawkw: This is one of those cases where basically ignoring something that looks like an error does kind of seem like the most correct things to do, but I agree that it feels uncomfortable. If we think about it, though, there are basically two circumstances where a calling task might register the same dynamic input multiple times: 

- Programmer error in the calling task that just makes it do this multiple times due to a bug
- _Or_, because the calling task has just panicked and came back and realized they were in a state where that input is live, so they sent it to us...because they have no way of tracking whether or not they had ever previously sent it to us. This is probably the likelier case, and it is something we need to handle gracefully.

If we consider the other options beyond silently ignoring it:

- We could REPLY_FAULT the caller. That might be a reasonable thing to do if the *only* reason this might happen was programmer error, but in the case where the calling task has just restarted, this would actually be Super Duper Ultra Wrong, because now we have turned a one-time crash into a crash loop. I'm pretty sure the one task that actually adds dynamic inputs is the sidecar sequencer, so if we were to reply-fault it for adding an input we already knew about, we would basically break the entire switch. Let's not do that.
- *We* could panic ourselves, which could potentially turn another task panicking into the thermal task also panicking, when there's no real reason to do that since we can just ignore it. Also not awesome.
- We could reset the state to `Boot`...but again, there's no actual reason to do that, since we have already been reading from this input channel. In that case, we're just resetting the PID state without needing to.
- We could just ringbuf something and keep going. This is the only option besides ignoring it that is not strictly worse than just ignoring it, IMO. This way, if it _does_ happen due to programmer error, there's something in the ringbuf that would indicate that this is happening a whole bunch of times. I dunno if that's super necessary, since the IPC client counters should already be recording how many times this has gotten called, but it could be useful for debugging why it's getting spammed and which index keeps getting re-added.

In sum, just doing nothing here *is* almost certainly the right thing to do. With that said, we *could* do nothing loudly instead of doing nothing silently, but it's up to you.

task/thermal/src/control.rs
```diff
@@ -1758,20 +1238,149 @@ impl<'a> ThermalControl<'a> {
...
+    /// Run the PID loop on startup
+    const USE_CONTROLLER: bool;
+
+    /// Instruct the sequencer to power down the system
+    fn power_down(&self) -> Result<(), crate::SeqError>;
+
+    /// The current power mode reported by the sequencer
+    fn power_mode(&self) -> PowerBitmask;
+
+    /// An iterator representing the presence of each Fan of the system
```
hawkw: turbo nit: we don't capitalize this elsewhere
```suggestion
    /// An iterator representing the presence of each fan of the system
```

task/thermal/src/control.rs
```diff
@@ -1758,20 +1238,149 @@ impl<'a> ThermalControl<'a> {
...
+    ///
+    /// Typically reported by the sequencer, or always present in non-variable
+    /// configurations.
```
hawkw:
```suggestion
    /// Typically reported by the sequencer, or always present in non-variable
    /// configurations (or an empty iterator, if the board does not have fans).
```
:)

task/thermal/src/control.rs
```diff
@@ -1758,20 +1238,149 @@ impl<'a> ThermalControl<'a> {
...
+    /// query was successful. This updates the state of the sensor, which is
+    /// obtained by calling [`Self::all_active_inputs()`], which unlike
+    /// this API, retains the latest valid value, in case of transient read
+    /// failures.
+    fn poll_inputs(
+        &mut self,
+        mode: PowerBitmask,
+    ) -> impl Iterator<Item = InputPollingOutcome>;
+
+    /// Reads back all dynamic inputs that are currently marked as present. The
```
hawkw: "Reads back..." feels a bit unclear. To me, it kind of suggests that this *returns* inputs, but it really asks the `sensor` task for them and puts them in the BSP's internal state, so that calls to `all_active_inputs()` gives them back. Maybe we should say that here?

task/thermal/src/control.rs
```diff
@@ -1758,20 +1238,149 @@ impl<'a> ThermalControl<'a> {
...
+    fn poll_misc_sensors(
+        &self,
+    ) -> impl Iterator<Item = MiscSensorPollingOutcome>;
+
+    /// Return an iterator of the outcome of polling each input. The iterator
+    /// reports whether the input is powered, present, and whether the latest
+    /// query was successful. This updates the state of the sensor, which is
+    /// obtained by calling [`Self::all_active_inputs()`], which unlike
+    /// this API, retains the latest valid value, in case of transient read
+    /// failures.
```
hawkw: nit: this sentence structure feels kind of weird. maybe:
```suggestion
    /// query was successful. This updates the state of the sensor, which is
    /// obtained by calling [`Self::all_active_inputs()`]. Unlike this API, 
    /// that function retains the latest valid value in case of transient 
    /// read failures.
```

hawkw: it's maybe also worth having a note here explaining when one or the other are used by the control loop? up to you though.

task/thermal/src/control.rs
```diff
@@ -1758,20 +1238,149 @@ impl<'a> ThermalControl<'a> {
...
+        mode: PowerBitmask,
+    ) -> impl Iterator<Item = InputPollingOutcome>;
+
+    /// Reads back all dynamic inputs that are currently marked as present. The
+    /// information from this querying can be obtained by calling
+    /// [`Self::all_active_inputs()`].
+    fn poll_dynamic_inputs(&mut self, sensor_api: &task_sensor_api::Sensor);
+
+    /// Set the given dynamic input as present, and configured with the given
+    /// model.
```
hawkw: turbo nit: maybe
```suggestion
    /// Set the given dynamic input as present, and configures it with the given
    /// model.
```

task/thermal/src/control.rs
```diff
@@ -1758,20 +1238,149 @@ impl<'a> ThermalControl<'a> {
...
+    /// Does not indicate whether the input was previously present or not.
+    fn remove_dynamic_input(
+        &mut self,
+        index: usize,
+    ) -> Result<SensorId, ThermalError>;
+
+    /// Have all powered inputs (regular and dynamic) been queried?
+    ///
+    /// This is used to determine whether it is appropriate to leave the `Boot`
+    /// state. A sensor is considered to be queried if it is:
```
hawkw: turbo nit:
```suggestion
    /// state. A sensor is considered to have been queried if it is:
```

hawkw (approved):
Okay, having given this a more thorough review of the changes to the core thermal loop, I love the refactoring. Overall, this looks great! I commented on some minor but important issues, and a few docs nits. The main thing is that the assertion that `all_inputs_queried` is making me kiiiiinda nervous, but I agree that it's pretty unlikely to hit in practice. Once those notes are addressed, you're free to merge this whenever you're happy with it!

task/thermal/src/control.rs
jamesmunns: Yeah, that's probably worth calling out somewhere. I do think it's conceptually right, because the general direction this PR is taking is that entity state is a property of the BSP to avoid having awkward "how much state do I need" juggling between the control loop and the bsp. I'll see where I can add docs for this.

task/thermal/src/control.rs
jamesmunns: Yeah, the intent of this PR was mostly to keep the functional behavior the same, and address any fixups/changes *after* this PR has landed, so we can discuss those with way less diff noise.

One side effect of polling dynamic inputs in `read_sensors` I realized later is that it would be done when we are in "manual" mode (which runs read_sensors but not run_control), but this is probably not a significant change either.

task/thermal/src/control.rs
jamesmunns: Yeah, I think ringbuf+return to boot is a reasonable alternative, I'll do that.

task/thermal/src/control.rs
jamesmunns: Oh, this was something I did temporarily while refactoring for exactly that reason. I'll remove these lines, they are not really doing anything at this point.

task/thermal/src/control.rs
jamesmunns: Yep, I did it like that exactly to preserve the diff. Now that review is mostly done, I'll bump it back up,

task/thermal/src/control.rs
jamesmunns: I think I agree with your reasoning, though what I would have expected is "take the new model and reset to boot". It might even be worth checking if the model is the same, and still nop-ing in that case.

It was *surprising* to me, but I think reasoned out leaving it as-is is fair. I'll remove the TODO part of the comment here and leave it as documentation of what we do.

task/thermal/src/control.rs
jamesmunns: changed

jamesmunns:
I've now tested this on a gimlet, cosmo, and sidecar on london, and addressed all the comments from Laura and Eliza.

In general, all of the devices qualitatively generally act as expected, I'm going to go ahead and merge these changes, and see if we can cut a release of hubris soon and get this tested on a wider subset of devices.

PR #2614, opened 2026-07-27
ringbuf: don't allow length-0 ringbufs

hawkw:
It turns out that, if you try to make a ringbuf with length 0, surprisingly bad things will happen. This is because ringbuf entries are written to using `get_unchecked_mut()`, in order to suppress bounds checks. Suppressing bounds checks is okay to do, because we have already modulo'd the index to the ringbuf's length, so it is guaranteed to be in bounds...AS LONG AS THE LENGTH IS GREATER THAN ZERO. If it is zero, well...`0 % 0 == 0`, so we will do `get_unchecked_mut(0)`, and uncheckedly dereference the first index of the ringbuf array...which it doesn't have, so you end up writing out of bounds to whatever happens to be next to the ringbuf.

Luckily, no one has ever actually tried to make a ringbuf with zero entries, because...why would you want that? But, just in case, it's better if we stop you, so this branch adds a const assertion to the ringbuf macro that checks that the length is greater than zero.

Now, if you try to make one, it fails like this: 

```
  Compiling task-ereportulator v0.1.0 (/home/eliza/Code/oxide/hubris/task/ereportulator)
error[E0080]: evaluation panicked: ringbufs may not be zero-length
    --> /rustlib/panic.rs:62:9
     |
  51 | pub macro panic_2021 {
     | -------------------- in this expansion of `$crate::panic::panic_2021!` (#3)
...
  62 |         $crate::panicking::panic_fmt($crate::const_format_args!($($t)+));
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `_` failed here
     |
    ::: lib/ringbuf/src/lib.rs:282:1
     |
 282 | macro_rules! ringbuf {
     | -------------------- in this expansion of `ringbuf::ringbuf!` (#1)
...
 293 |             assert!($n > 0, "ringbufs may not be zero-length");
     |             --------------------------------------------------
     |             |
     |             in this macro invocation (#2)
     |             in this macro invocation (#3)
     |
    ::: task/ereportulator/src/main.rs:72:1
     |
  72 | ringbuf::ringbuf!(__LOL, Trace, 0, Trace::None);
     | ----------------------------------------------- in this macro invocation (#1)
     |
    ::: /rustlib/macros/mod.rs:1691:5
     |
1691 |     macro_rules! assert {
     |     ------------------- in this expansion of `assert!` (#2)

For more information about this error, try `rustc --explain E0080`.
error: could not compile `task-ereportulator` (bin "task-ereportulator") due to 1 previous error
```

### DIFF
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -655,19 +655,13 @@ impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
     // *also* an out of bounds read. Luckily, nobody has ever wanted to make a
     // zero-length ringbuf, so this hasn't been an issue in practice, but let's
     // stop you from doing it here just in case.
-    //
-    // Sadly, using a `const` block with an `assert!` in it somehow does not
-    // fail to build, but the "assertion"-using-a-bool-as-an-array-index trick
-    // does work. I don't know why.
-    const RINGBUF_LEN_MUST_BE_NONZERO: () = [()][(N > 0) as usize];
+    pub const RINGBUF_LEN_MUST_BE_NONZERO: bool = {
+        let is_nonzero = N > 0;
+        assert!(is_nonzero, "ringbuf length must be greater than 0");
+        is_nonzero
+    };
 
     fn do_record(&mut self, last: usize, line: u16, count: C, payload: &T) {
-        // Force the compile-time assertion that the length is nonzero to be
-        // evaluated. Clippy complains (not unreasonably), that this let-binding
-        // is not actually doing anything, which is true, but that's on purpose
-        // here.
-        #[allow(clippy::let_unit_value)]
-        let _ = Self::RINGBUF_LEN_MUST_BE_NONZERO;
         // Either we were unable to reuse the entry, or the last index was out
         // of range (perhaps because this is the first insertion). We're going
         // to advance last and wrap if required. This uses a wrapping_add
@@ -700,7 +694,9 @@ impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
             // in order to actually be safe, as a zero-length ringbuf is the one
             // exception to the guarantee that having modulo'd the index (or
             // having done our weird modulo-like thing, technically) ensures
-            // that it is in bounds.
+            // that it is in bounds. So, force that to actually be evaluated
+            // here, so it fails at the location of the unsafe block comment. :)
+            let _is_nonzero = Self::RINGBUF_LEN_MUST_BE_NONZERO;
             self.buffer.get_unchecked_mut(ndx)
         };
         *ent = RingbufEntry {
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -280,6 +280,17 @@ macro_rules! ringbuf {
 #[macro_export]
 macro_rules! ringbuf {
     ($name:ident, $t:ty, $n:expr, $init:expr $(,)?) => {
+        // `N` may not be 0, and constructing a zero-length ringbuf results in a
+        // compiler error. This is necessary because ringbuf entries are written
+        // to using unchecked indexing, which is always valid because the index
+        // has already been modulo'd to the ringbuf's length...unless the length
+        // is 0, in which case reading the 0th index is *also* an out of bounds
+        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
+        // so this hasn't been an issue in practice, but let's stop you from
+        // doing it here just in case.
+        const _: () = {
+            assert!($n > 0, "ringbufs may not be zero-length");
+        };
         static $name: $crate::StaticCell<$crate::Ringbuf<$t, u16, $n>> =
             $crate::StaticCell::new($crate::Ringbuf {
                 last: None,
@@ -292,6 +303,17 @@ macro_rules! ringbuf {
             });
     };
     ($name:ident, $t:ty, $n:expr, $init:expr, no_dedup $(,)?) => {
+        // `N` may not be 0, and constructing a zero-length ringbuf results in a
+        // compiler error. This is necessary because ringbuf entries are written
+        // to using unchecked indexing, which is always valid because the index
+        // has already been modulo'd to the ringbuf's length...unless the length
+        // is 0, in which case reading the 0th index is *also* an out of bounds
+        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
+        // so this hasn't been an issue in practice, but let's stop you from
+        // doing it here just in case.
+        const _: () = {
+            assert!($n > 0, "ringbufs may not be zero-length");
+        };
         static $name: $crate::StaticCell<$crate::Ringbuf<$t, () $n>> =
             $crate::StaticCell::new($crate::Ringbuf {
                 last: None,
@@ -336,6 +358,17 @@ macro_rules! ringbuf {
 #[macro_export]
 macro_rules! counted_ringbuf {
     ($name:ident, $t:ident, $n:expr, $init:expr $(,)?) => {
+        // `N` may not be 0, and constructing a zero-length ringbuf results in a
+        // compiler error. This is necessary because ringbuf entries are written
+        // to using unchecked indexing, which is always valid because the index
+        // has already been modulo'd to the ringbuf's length...unless the length
+        // is 0, in which case reading the 0th index is *also* an out of bounds
+        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
+        // so this hasn't been an issue in practice, but let's stop you from
+        // doing it here just in case.
+        const _: () = {
+            assert!($n > 0, "ringbufs may not be zero-length");
+        };
         static $name: $crate::CountedRingbuf<$t, u16, $n> =
             $crate::CountedRingbuf {
                 ringbuf: $crate::StaticCell::new($crate::Ringbuf {
@@ -351,6 +384,17 @@ macro_rules! counted_ringbuf {
             };
     };
     ($name:ident, $t:ident, $n:expr, $init:expr, no_dedup $(,)?) => {
+        // `N` may not be 0, and constructing a zero-length ringbuf results in a
+        // compiler error. This is necessary because ringbuf entries are written
+        // to using unchecked indexing, which is always valid because the index
+        // has already been modulo'd to the ringbuf's length...unless the length
+        // is 0, in which case reading the 0th index is *also* an out of bounds
+        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
+        // so this hasn't been an issue in practice, but let's stop you from
+        // doing it here just in case.
+        const _: () = {
+            assert!($n > 0, "ringbufs may not be zero-length");
+        };
         static $name: $crate::CountedRingbuf<$t, (), $n> =
             $crate::CountedRingbuf {
                 ringbuf: $crate::StaticCell::new($crate::Ringbuf {
@@ -675,6 +719,12 @@ impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
             // `self.buffer.get_mut(ndx)` and then silently nop'ing if it
             // returns `None`...but it seems nicer to also elide the bounds
             // check, given that we *just* did one of our own!
+            //
+            // Note that this requires the `const` assertion that `N > 0` in the
+            // ringbuf macro in order to actually be safe, as a zero-length
+            // ringbuf is the one exception to the guarantee that having
+            // modulo'd the index (or having done our weird modulo-like thing,
+            // technically) ensures that it is in bounds.
             self.buffer.get_unchecked_mut(ndx)
         };
         *ent = RingbufEntry {
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -280,17 +280,6 @@ macro_rules! ringbuf {
 #[macro_export]
 macro_rules! ringbuf {
     ($name:ident, $t:ty, $n:expr, $init:expr $(,)?) => {
-        // `N` may not be 0, and constructing a zero-length ringbuf results in a
-        // compiler error. This is necessary because ringbuf entries are written
-        // to using unchecked indexing, which is always valid because the index
-        // has already been modulo'd to the ringbuf's length...unless the length
-        // is 0, in which case reading the 0th index is *also* an out of bounds
-        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
-        // so this hasn't been an issue in practice, but let's stop you from
-        // doing it here just in case.
-        const _: () = {
-            assert!($n > 0, "ringbufs may not be zero-length");
-        };
         static $name: $crate::StaticCell<$crate::Ringbuf<$t, u16, $n>> =
             $crate::StaticCell::new($crate::Ringbuf {
                 last: None,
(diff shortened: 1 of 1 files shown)


lib/ringbuf/src/lib.rs
```diff
@@ -675,6 +719,12 @@ impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
             // `self.buffer.get_mut(ndx)` and then silently nop'ing if it
             // returns `None`...but it seems nicer to also elide the bounds
             // check, given that we *just* did one of our own!
+            //
+            // Note that this requires the `const` assertion that `N > 0` in the
+            // ringbuf macro in order to actually be safe, as a zero-length
+            // ringbuf is the one exception to the guarantee that having
+            // modulo'd the index (or having done our weird modulo-like thing,
+            // technically) ensures that it is in bounds.
```
jamesmunns: Could we just do the static assert here on the `N` parameter instead of all the other places?

hawkw: I tried that, but surprisingly, no: it can't go _inside_ the function because consts within a scope cannot access const-generic parameters from the enclosing scope. You can put a
```rust
const _DONT_DO_THAT: () = { assert!(N > 0); }
```
in the `impl` block[^1], but, as I was surprised to discover, this...doesn't actually do anything, and an image that declares a length-0 ringbuf compiles if it's asserted inside the `impl Ringbuf` block. 

In hindsight, it's possible it didn't hit the assertion because my test code was just _constructing_ a length-0 ringbuf, not trying to write to one? But, I would think the const would still be evaluated in any case?

[^1]: It has to be named, because you can't use `const _` as an associated const apparently.

hawkw: shockingly, even doing it like this (so that the const is forced to be evaluated) still compiles just fine:
```patch
diff --git a/lib/ringbuf/src/lib.rs b/lib/ringbuf/src/lib.rs
index 7d2bde7b4..f80e97814 100644
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -280,17 +280,6 @@ macro_rules! ringbuf {
 #[macro_export]
 macro_rules! ringbuf {
     ($name:ident, $t:ty, $n:expr, $init:expr $(,)?) => {
-        // `N` may not be 0, and constructing a zero-length ringbuf results in a
-        // compiler error. This is necessary because ringbuf entries are written
-        // to using unchecked indexing, which is always valid because the index
-        // has already been modulo'd to the ringbuf's length...unless the length
-        // is 0, in which case reading the 0th index is *also* an out of bounds
-        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
-        // so this hasn't been an issue in practice, but let's stop you from
-        // doing it here just in case.
-        const _: () = {
-            assert!($n > 0, "ringbufs may not be zero-length");
-        };
         static $name: $crate::StaticCell<$crate::Ringbuf<$t, u16, $n>> =
             $crate::StaticCell::new($crate::Ringbuf {
                 last: None,
@@ -303,17 +292,6 @@ macro_rules! ringbuf {
             });
     };
     ($name:ident, $t:ty, $n:expr, $init:expr, no_dedup $(,)?) => {
-        // `N` may not be 0, and constructing a zero-length ringbuf results in a
-        // compiler error. This is necessary because ringbuf entries are written
-        // to using unchecked indexing, which is always valid because the index
-        // has already been modulo'd to the ringbuf's length...unless the length
-        // is 0, in which case reading the 0th index is *also* an out of bounds
-        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
-        // so this hasn't been an issue in practice, but let's stop you from
-        // doing it here just in case.
-        const _: () = {
-            assert!($n > 0, "ringbufs may not be zero-length");
-        };
         static $name: $crate::StaticCell<$crate::Ringbuf<$t, () $n>> =
             $crate::StaticCell::new($crate::Ringbuf {
                 last: None,
@@ -358,17 +336,6 @@ macro_rules! ringbuf {
 #[macro_export]
 macro_rules! counted_ringbuf {
     ($name:ident, $t:ident, $n:expr, $init:expr $(,)?) => {
-        // `N` may not be 0, and constructing a zero-length ringbuf results in a
-        // compiler error. This is necessary because ringbuf entries are written
-        // to using unchecked indexing, which is always valid because the index
-        // has already been modulo'd to the ringbuf's length...unless the length
-        // is 0, in which case reading the 0th index is *also* an out of bounds
-        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
-        // so this hasn't been an issue in practice, but let's stop you from
-        // doing it here just in case.
-        const _: () = {
-            assert!($n > 0, "ringbufs may not be zero-length");
-        };
         static $name: $crate::CountedRingbuf<$t, u16, $n> =
             $crate::CountedRingbuf {
                 ringbuf: $crate::StaticCell::new($crate::Ringbuf {
@@ -384,17 +351,6 @@ macro_rules! counted_ringbuf {
             };
     };
     ($name:ident, $t:ident, $n:expr, $init:expr, no_dedup $(,)?) => {
-        // `N` may not be 0, and constructing a zero-length ringbuf results in a
-        // compiler error. This is necessary because ringbuf entries are written
-        // to using unchecked indexing, which is always valid because the index
-        // has already been modulo'd to the ringbuf's length...unless the length
-        // is 0, in which case reading the 0th index is *also* an out of bounds
-        // read. Luckily, nobody has ever wanted to make a zero-length ringbuf,
-        // so this hasn't been an issue in practice, but let's stop you from
-        // doing it here just in case.
-        const _: () = {
-            assert!($n > 0, "ringbufs may not be zero-length");
-        };
         static $name: $crate::CountedRingbuf<$t, (), $n> =
             $crate::CountedRingbuf {
                 ringbuf: $crate::StaticCell::new($crate::Ringbuf {
@@ -691,6 +647,10 @@ where
 }

 impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
+    const CHECKED_BUFFER_LEN: usize = {
+        assert!(N > 0, "ringbufs may not be zero-length");
+        N
+    };
     fn do_record(&mut self, last: usize, line: u16, count: C, payload: &T) {
         // Either we were unable to reuse the entry, or the last index was out
         // of range (perhaps because this is the first insertion). We're going
@@ -703,7 +663,7 @@ impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
             // This is because none of our target platforms currently have
             // hardware modulus, and many of them don't even have hardware
             // divide, making remainder quite expensive.
-            if last_plus_1 >= self.buffer.len() {
+            if last_plus_1 >= Self::CHECKED_BUFFER_LEN {
                 0
             } else {
                 last_plus_1
diff --git a/task/ereportulator/src/main.rs b/task/ereportulator/src/main.rs
index 60cddd685..4aa627d94 100644
--- a/task/ereportulator/src/main.rs
+++ b/task/ereportulator/src/main.rs
@@ -69,6 +69,8 @@ enum Trace {

+ringbuf::ringbuf!(__LOL, Trace, 0, Trace::None);
+
 #[unsafe(export_name = "main")]
 fn main() -> ! {
```

jamesmunns: Apparently it does work but only if you "use" the impl-scoped const?

```diff
diff --git a/lib/ringbuf/src/lib.rs b/lib/ringbuf/src/lib.rs
index cb9c3d2c7..c5bc6e2df 100644
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -647,7 +647,15 @@ where
 }
 
 impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
+    #[doc(hidden)]
+    pub const _NONZERO: bool = const {
+        let is_nonzero = N > 0;
+        assert!(is_nonzero);
+        is_nonzero
+    };
+
     fn do_record(&mut self, last: usize, line: u16, count: C, payload: T) {
+        let _nz = Self::_NONZERO;
         // Either we were unable to reuse the entry, or the last index was out
         // of range (perhaps because this is the first insertion). We're going
         // to advance last and wrap if required. This uses a wrapping_add
diff --git a/task/jefe/src/main.rs b/task/jefe/src/main.rs
index b5d109072..9f3b1213f 100644
--- a/task/jefe/src/main.rs
+++ b/task/jefe/src/main.rs
@@ -47,6 +47,8 @@ pub enum Disposition {
     Hold,
 }
 
+ringbuf::ringbuf!(u32, 0, 4);
+
 // We install a timeout to periodically check for an external direction
 // of our task disposition (e.g., via Humility).  This timeout should
 // generally be fast for a human but slow for a computer; we pick a
@@ -81,6 +83,7 @@ const MIN_RESTART_DELAY: u64 = 5;
 
 #[unsafe(export_name = "main")]
 fn main() -> ! {
+    ringbuf::ringbuf_entry!(15);
     let mut task_states = [TaskStatus::default(); hubris_num_tasks::NUM_TASKS];
     for held_task in generated::HELD_TASKS {
         task_states[held_task as usize].disposition = Disposition::Hold;
```

```
   Compiling task-jefe v0.1.0 (/Users/james/oxide/hubris/task/jefe)
error[E0080]: evaluation panicked: assertion failed: N > 0
    --> lib/ringbuf/src/lib.rs:652:9
     |
 652 |         assert!(N > 0);
     |         ^^^^^^^^^^^^^^
     |         |
     |         evaluation of `ringbuf::Ringbuf::<u32, u16, 0>::_NONZERO::{constant#0}` failed here
     |         in this macro invocation
     |
    ::: /rustlib/macros/mod.rs:1691:5
     |
1691 |     macro_rules! assert {
     |     ------------------- in this expansion of `assert!`

note: erroneous constant encountered
   --> lib/ringbuf/src/lib.rs:651:30
    |
651 |       pub const _NONZERO: () = const {
    |  ______________________________^
652 | |         assert!(N > 0);
653 | |     };
    | |_____^

note: erroneous constant encountered
   --> lib/ringbuf/src/lib.rs:656:19
    |
656 |         let _nz = Self::_NONZERO;
    |                   ^^^^^^^^^^^^^^

note: the above error was encountered while instantiating `fn Ringbuf::<u32, u16, 0>::do_record`
   --> lib/ringbuf/src/lib.rs:602:9
    |
602 |         ring.do_record(last, line, 1, payload);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0080`.
error: could not compile `task-jefe` (bin "task-jefe") due to 1 previous error
Error: failed to build jefe
```

Huh.

hawkw: oh, that's nicer than the solution i came up with (https://github.com/oxidecomputer/hubris/pull/2614/changes/7c5e1bc66e23bec630ef38335576c91cc47627ba)

hawkw: ...what the heck, how come your thing works but this doesn't?
```patch
eliza@hekate ~/Code/oxide/hubris $ git diff
diff --git a/lib/ringbuf/src/lib.rs b/lib/ringbuf/src/lib.rs
index 694902534..109ce2534 100644
--- a/lib/ringbuf/src/lib.rs
+++ b/lib/ringbuf/src/lib.rs
@@ -659,7 +659,10 @@ impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
     // Sadly, using a `const` block with an `assert!` in it somehow does not
     // fail to build, but the "assertion"-using-a-bool-as-an-array-index trick
     // does work. I don't know why.
-    const RINGBUF_LEN_MUST_BE_NONZERO: () = [()][(N > 0) as usize];
+    pub const RINGBUF_LEN_MUST_BE_NONZERO: usize = {
+        assert!(N > 0, "ringbuf length must be greater than 0");
+        N
+    };

     fn do_record(&mut self, last: usize, line: u16, count: C, payload: &T) {
         // Force the compile-time assertion that the length is nonzero to be
@@ -667,7 +670,7 @@ impl<T: Copy, C, const N: usize> Ringbuf<T, C, N> {
         // is not actually doing anything, which is true, but that's on purpose
         // here.
         #[allow(clippy::let_unit_value)]
-        let _ = Self::RINGBUF_LEN_MUST_BE_NONZERO;
+        let _n = Self::RINGBUF_LEN_MUST_BE_NONZERO;
         // Either we were unable to reuse the entry, or the last index was out
         // of range (perhaps because this is the first insertion). We're going
         // to advance last and wrap if required. This uses a wrapping_add
diff --git a/task/ereportulator/src/main.rs b/task/ereportulator/src/main.rs
index 60cddd685..4aa627d94 100644
--- a/task/ereportulator/src/main.rs
+++ b/task/ereportulator/src/main.rs
@@ -69,6 +69,8 @@ enum Trace {

+ringbuf::ringbuf!(__LOL, Trace, 0, Trace::None);
+
 #[unsafe(export_name = "main")]
 fn main() -> ! {
```

hawkw: Oh, I see, the problem is that evaluating it in the function only fails if the code actually calls `do_record` somewhere. I guess that's fine, since it still stops us from doing the OOB write, and just _constructing_ a zero-length ringbuf is _technically_ fine if you never actually use it. What do you think of https://github.com/oxidecomputer/hubris/pull/2614/commits/55ae04c659f9d9fa5f5be6821bdf811fbb49d099?

That produces an error message like this:
```
   Compiling task-ereportulator v0.1.0 (/home/eliza/Code/oxide/hubris/task/ereportulator)
error[E0080]: evaluation panicked: ringbuf length must be greater than 0
    --> /rustlib/panic.rs:62:9
     |
  51 | pub macro panic_2021 {
     | -------------------- in this expansion of `$crate::panic::panic_2021!` (#2)
...
  62 |         $crate::panicking::panic_fmt($crate::const_format_args!($($t)+));
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `ringbuf::Ringbuf::<Trace, u16, 0>::RINGBUF_LEN_MUST_BE_NONZERO` failed here
     |
    ::: /rustlib/macros/mod.rs:1691:5
     |
1691 |     macro_rules! assert {
     |     ------------------- in this expansion of `assert!` (#1)
     |
    ::: lib/ringbuf/src/lib.rs:660:9
     |
 660 |         assert!(is_nonzero, "ringbuf length must be greater than 0");
     |         ------------------------------------------------------------
     |         |
     |         in this macro invocation (#1)
     |         in this macro invocation (#2)

note: erroneous constant encountered
    --> /rustlib/panic.rs:62:9
     |
  51 | pub macro panic_2021 {
     | -------------------- in this expansion of `$crate::panic::panic_2021!` (#2)
...
  62 |         $crate::panicking::panic_fmt($crate::const_format_args!($($t)+));
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
    ::: lib/ringbuf/src/lib.rs:660:9
     |
 660 |         assert!(is_nonzero, "ringbuf length must be greater than 0");
     |         ------------------------------------------------------------
     |         |
     |         in this macro invocation (#1)
     |         in this macro invocation (#2)
     |
    ::: /rustlib/macros/mod.rs:1691:5
     |
1691 |     macro_rules! assert {
     |     ------------------- in this expansion of `assert!` (#1)

note: erroneous constant encountered
   --> lib/ringbuf/src/lib.rs:699:31
    |
699 |             let _is_nonzero = Self::RINGBUF_LEN_MUST_BE_NONZERO;
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

note: the above error was encountered while instantiating `fn Ringbuf::<Trace, u16, 0>::do_record`
   --> lib/ringbuf/src/lib.rs:602:9
    |
602 |         ring.do_record(last, line, 1, payload);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0080`.
error: could not compile `task-ereportulator` (bin "task-ereportulator") due to 1 previous error
```

...which, I hope, is sufficiently descriptive that anyone who accidentally makes a zero-entry ringbuf understands why they shouldn't have done that.

jamesmunns (approved):
I like it!

Issue #2216, opened 2025-09-05
expose PSC sequencer's understanding of rectifier states to the management network


hawkw:
#2214 adds code to the `psc_seq` task for generating ereports when a rectifier's presence or POWER_GOOD states change. Ereports represent edge-triggered notifications of an event. As discussed in RFD 589, we must also provide [health endpoints](https://rfd.shared.oxide.computer/rfd/0589#_health_endpoints) that expose the same component health information in a manner that the control plane can poll. This is necessary in the event that an ereport is lost, either due to a SP reset or a full ereport buffer in the SP. If either of those conditions are detected by the control plane, it will poll the health of the SP's components to determine whether there may be a fault (or state change, more broadly) for which an ereport may have been lost.

Currently, we do expose presence information for rectifiers over the `control-plane-agent` inventory interface. Presence is detected by the `validate` task by reading the rectifier's model number over I2C, and returning an error if that doesn't work:

https://github.com/oxidecomputer/hubris/blob/b50c3c74e55cf14bc22fb1ef1350844952d1bd9e/drv/i2c-devices/src/mwocp68.rs#L764-L770

This is different from the PSC sequencer's understanding of presence, which is detected by reading the rectifier presence GPIO pins. In most cases, though, this should be more or less in line.

What _isn't_ currently known to `control-plane-agent` is whether the rectifier is healthy or has faulted, and whether the `psc_seq` task has enabled or disabled the rectifier. The sequencer determines if a rectifier is faulted based on whether or not `POWER_GOOD` is asserted, and enables and disables rectifiers by toggling GPIOs. The `validate` task could potentially read PMBus status registers to determine if the rectifier is faulted (and we may want to do that anyway to expose PMBus status information to the control plane). However, this is a different mechanism from the `psc_seq` task's understanding of faults, and it's the sequencer that actually controls whether a rectifier is turned on or not. Therefore, I think it might be better if we treat the sequencer as an authoritative source for this and just ask it whether it believes a rectifier is okay or not. 

However, this might be a bit tricky to wire into the existing control-plane-agent inventory interface, which basically expects all devices are accessed through `validate`. We could probably special-case the PSUs by putting them in `control-plane-agent`'s "devices with static validation" list, but presently, that would mean _always_ treating them as present, which is actually _worse_. What we really need is a way to tell control plane agent that "this thing needs special validation behavior".

PR #2617, opened 2026-07-29
Remove silent fallback behavior from i2c codegen

### LINKED ISSUE #2616: Rectifiers may be included in inventory even when not valid
When the SP reports the inventory of devices to the control plane, it includes rectifiers that are "valid". The validation process was meant to be "check that we can read the expected model number over pmbus". However, because of a bug introduced in https://github.com/oxidecomputer/hubris/pull/2549, the actual validation process is now "check that an i2c device exists at this address and ACKs some arbitrary read". This doesn't cause a problem during normal operation, but it means that the validation is less strict than we would like. 

The problem is that when the [i2c codegen](https://github.com/oxidecomputer/hubris/blob/4fed8c5dbcc6fe9cc4bf2d6bc566708f12c4d5a3/build/i2c/src/lib.rs#L1242-L1306) needs a `validate()` function for a device named `foo`, it searches for an appropriate driver by looking for a file with the exact path `src/foo.rs` in the `drv-i2c-devices` crate. If it finds that file, it calls the driver's handwritten, device-specific `validate()` function. If it doesn't, it silently generates a fallback `validate()` function that just checks if a read is ACKed.

The problem with https://github.com/oxidecomputer/hubris/pull/2549 was that it moved `src/mwocp68.rs` to `src/mwocp6x/mwocp68.rs` so that it could share code with the Observer's new mwocp67 rectifier. This meant the codegen couldn't find the driver anymore and started using the fallback instead.

We should check if any devices are actually intentionally using the fallback. If not, we could just delete that code path and fail to build if the driver is not found. If we do need the fallback, it should become opt-in through the device config in the manifest.

Note that we are planning to make big changes to how rectifier inventory and validation work in https://github.com/oxidecomputer/hubris/issues/2216, but we should still fix this smaller bug first.

evan-oxide:
This closes https://github.com/oxidecomputer/hubris/issues/2616.

### DIFF
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -477,6 +477,7 @@ address = 0x70
 device = "oximux16"
 description = "Front FPGA virtual mux"
 refdes = ["U31"]
+validate-with-raw-read = true
 
 # Welcome to the sharkfin zone
 # SMBUS_FPGA2_TO_CEMA
@@ -902,6 +903,7 @@ address = 0x70
 device = "oximux16"
 description = "Main FPGA virtual mux"
 refdes = ["U27", "MUX"]
+validate-with-raw-read = true
 
 # SMBUS_FPGA1_TO_M2A
 [[config.i2c.devices]]
--- a/app/cosmo/base.toml
+++ b/app/cosmo/base.toml
@@ -477,6 +477,8 @@ address = 0x70
 device = "oximux16"
 description = "Front FPGA virtual mux"
 refdes = ["U31"]
+# TODO Audit i2c devices using validate-with-raw-read.
+# https://github.com/oxidecomputer/hubris/issues/2629
 validate-with-raw-read = true
 
 # Welcome to the sharkfin zone
@@ -903,6 +905,8 @@ address = 0x70
 device = "oximux16"
 description = "Main FPGA virtual mux"
 refdes = ["U27", "MUX"]
+# TODO Audit i2c devices using validate-with-raw-read.
+# https://github.com/oxidecomputer/hubris/issues/2629
 validate-with-raw-read = true
 
 # SMBUS_FPGA1_TO_M2A
--- a/build/i2c/src/lib.rs
+++ b/build/i2c/src/lib.rs
@@ -108,6 +108,15 @@ struct I2cDevice {
     /// device is removable
     #[serde(default)]
     removable: bool,
+
+    /// We typically expect that each device will have a driver in
+    /// `drv-i2c-devices` that implements the `Validate` trait, doing some
+    /// device-specific validation (like checking the model number via PMBus).
+    /// If there is no driver, then codegen will fail - _unless_ you set this to
+    /// true to opt in to a generic, fallback implementation of `Validate` that
+    /// just checks whether the device ACKs a PMBus read.
+    #[serde(default)]
+    validate_with_raw_read: bool,
 }
 
 impl I2cDevice {
@@ -1217,6 +1226,9 @@ impl ConfigGenerator {
         // breaking -- with apologies, dear reader, if that's what brings you
         // here!
         //
+        // Apology accepted. Perhaps this should be completely redesigned, but
+        // I've at least made it so the build will now fail with a helpful error
+        // message if the device driver isn't found.
         use cargo_metadata::MetadataCommand;
 
         let metadata = MetadataCommand::new()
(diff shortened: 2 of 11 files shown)


build/i2c/src/lib.rs
```diff
@@ -108,6 +108,15 @@ struct I2cDevice {
...
     removable: bool,
+
+    /// We typically expect that each device will have a driver in
+    /// `drv-i2c-devices` that implements the `Validate` trait, doing some
+    /// device-specific validation (like checking the model number via PMBus).
+    /// If there is no driver, then codegen will fail - _unless_ you set this to
+    /// true to opt in to a generic, fallback implementation of `Validate` that
+    /// just checks whether the device ACKs a PMBus read.
+    #[serde(default)]
+    validate_with_raw_read: bool,
```
jamesmunns, line 123: It might be worth documenting what "raw read" does. I *think* this does a zero-sized-read address check, but I'm not sure based on the docs added here.

app/cosmo/base.toml
```diff
@@ -477,6 +477,7 @@ address = 0x70
 device = "oximux16"
 description = "Front FPGA virtual mux"
 refdes = ["U31"]
+validate-with-raw-read = true
```
jamesmunns, line 482: It *might* be worth documenting for each of the `validate-with-raw-read = true` examples *why* this is acceptable, and/or opening a follow-on issue to review whether each of these are appropriate.

I'm guessing we won't prioritize that now-ish, so it would be good to leave breadcrumbs to know what the scope of the effort is to improve this for future us.

It might be worth cross-linking that issue to https://github.com/oxidecomputer/hubris/issues/2216 which might resolve some of this (though I'm not sure about that).

evan-oxide: Added breadcrumbs

jamesmunns: Perfect, thank you!

jamesmunns (approved):

build/i2c/src/lib.rs
evan-oxide: Added some more details to that comment

build/i2c/src/lib.rs
```diff
@@ -1217,6 +1230,9 @@ impl ConfigGenerator {
         // breaking -- with apologies, dear reader, if that's what brings you
         // here!
         //
+        // Apology accepted. Perhaps this should be completely redesigned, but
+        // I've at least made it so the build will now fail with a helpful error
+        // message if the device driver isn't found.
```
hawkw, line 1235: lol, humor truly is an oxide value.

build/i2c/src/lib.rs
```diff
@@ -108,6 +108,19 @@ struct I2cDevice {
...
+    /// `drv-i2c-devices` that implements the `Validate` trait, doing some
+    /// device-specific validation (like checking the model number via PMBus).
+    /// If there is no driver, then codegen will fail - _unless_ you set this to
+    /// true to opt in to a generic, fallback implementation of `Validate` that
+    /// just checks whether the device ACKs a single-byte i2c read (with no
+    /// write beforehand). This is meant to detect whether the device is
+    /// present. However, not all devices react to such a read in the same way,
+    /// so this fallback implementation may be incorrect or cause unwanted side
+    /// effects on some devices.
+    #[serde(default)]
```
hawkw, line 122: i briefly panicked about the `#[serde(default)]` enabling this behavior for everything but then i remembered that `bool::default()` is `false`, so this seems correct. disregard me, i am stupid.

build/i2c/src/lib.rs
```diff
@@ -1280,6 +1296,26 @@ impl ConfigGenerator {
         // here, it must be updated there as well.
         for (index, device) in self.devices.iter().enumerate() {
             if drivers.contains(&device.device) {
+                if device.validate_with_raw_read {
+                    bail!(
+                        "Device '{}{}{}' set `validate-with-raw-read = true`, \
+                        but that was probably a mistake because this device \
+                        already has a driver in `drv-i2c-devices` that should \
```
hawkw, line 1303: turbo nit: maybe we should print the name of the driver here?

hawkw (approved):
great, looks good to me!

build/i2c/src/lib.rs
evan-oxide: We don't know the real name of the driver at this point. We assume it's an upper camel case version of the device name, and rustc will yell at us later if that was wrong. So I don't want to risk having this message gaslight someone about a driver name that doesn't exist.

PR #2618, opened 2026-07-30
host-sp-comms: Check for bad cert length

evan-oxide:
When host-sp-comms handles the `GetTqCertificates` request, it doesn't explicitly check whether the certificate length reported by the RoT fits in its buffer. This was flagged because we thought it could cause a panic, and I added a check that would return a `CommsBufTooSmall` error instead. However, after I got the hardware set up for testing it, I found that the SP already reported an `AttestOutOfRange` error rather than panicking. 

So this PR is kinda pointless, but I did find a few other small things to fix along the way. Changes:
- `GetTqCertificates` can now report a `CommsBufTooSmall` error, but in practice it never will (unless some surrounding code changes someday.) I guess I'll leave the check in place. I tested this with a grapefruit and faux-ipcc.
- `TqSign` can now report a `CommsBufTooSmall` error. I couldn't figure out a good way to test this. 
- I fixed an off-by-one error in all the existing `CommsBufTooSmall` checks. I didn't test that either.

### DIFF
--- a/task/host-sp-comms/src/main.rs
+++ b/task/host-sp-comms/src/main.rs
@@ -1470,12 +1470,22 @@ impl ServerImpl {
                             .sprot
                             .tq_cert_len(index)
                             .map_err(AttestDataSprotError::from)?;
-                        for o in (0..cert_len).step_by(512) {
+                        for o in (0..cert_len).step_by(SPROT_TRANSFER_SIZE) {
                             let idx: usize = buf_idx + o as usize;
-                            let len = usize::min(512, (cert_len - o) as usize);
+                            let end = usize::min(
+                                SPROT_TRANSFER_SIZE,
+                                (cert_len - o) as usize,
+                            );
+
+                            if idx + end >= buf.len() {
+                                return Err(
+                                    AttestDataSprotError::CommsBufTooSmall
+                                        .into(),
+                                );
+                            }
 
                             self.sprot
-                                .tq_cert(index, o, &mut buf[idx..idx + len])
+                                .tq_cert(index, o, &mut buf[idx..idx + end])
                                 .map_err(AttestDataSprotError::from)?;
                         }
                         cnt += cert_len;
--- a/task/host-sp-comms/src/main.rs
+++ b/task/host-sp-comms/src/main.rs
@@ -1363,7 +1363,7 @@ impl ServerImpl {
                                 SPROT_TRANSFER_SIZE,
                                 (cert_len - o) as usize,
                             );
-                            if idx + end >= buf.len() {
+                            if idx + end > buf.len() {
                                 return Err(
                                     AttestDataSprotError::CommsBufTooSmall
                                         .into(),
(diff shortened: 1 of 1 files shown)


task/host-sp-comms/src/main.rs
```diff
@@ -1470,12 +1470,22 @@ impl ServerImpl {
                             .sprot
                             .tq_cert_len(index)
                             .map_err(AttestDataSprotError::from)?;
-                        for o in (0..cert_len).step_by(512) {
+                        for o in (0..cert_len).step_by(SPROT_TRANSFER_SIZE) {
```
jamesmunns, line 1473: Take it or leave it, I think it would be possible to do a bit less book-keeping using a slightly different technique:

```rust
let f = |buf: &mut [u8]| {
    // Make a sliding window of our buffer, and remember the
    // original length so we can figure out how much we've used
    let orig_len = buf.len();
    let mut window = buf;

    // This is the number of certificates to pull
    let cert_chain_len = self
        .sprot
        .tq_cert_chain_len()
        .map_err(AttestDataSprotError::from)?;

    // For each certificate index in the chain...
    for index in 0..cert_chain_len {
        // ...get the length of the certificate
        let cert_len = self
            .sprot
            .tq_cert_len(index)
            .map_err(AttestDataSprotError::from)?
            as usize;

        // If we don't have room for the full certificate, don't
        // even ask the ROT
        if cert_len > window.len() {
            return Err(
                AttestDataSprotError::CommsBufTooSmall.into()
            );
        }

        // Keep track of how much of the certificate we've
        // consumed. We've already checked we have room for the
        // whole certificate.
        let mut got = 0;
        while got < cert_len {
            // Get the max transfer size or the remaining amt,
            // whichever is less.
            let remain = cert_len - got;
            let to_take = remain.min(SPROT_TRANSFER_SIZE);

            // Break our remaining buffer off for the current
            // cert fragment, and request it.
            let (now, later) = window.split_at_mut(to_take);
            self.sprot
                .tq_cert(index, got as u32, now)
                .map_err(AttestDataSprotError::from)?;

            // Update tracking variables
            got += to_take;
            window = later;
        }
    }
    // The amount we used is the original length minus the
    // number of unused bytes.
    let used = orig_len - window.len();
    Ok(used)
};
```

jamesmunns (approved):
Looks good! I have one style suggestion that might make the buffer ops more clear IMO.

FWIW, the `HostToRotCommand::GetCertificates` handler has roughly the same shape, it might be worth sharing some code between them, but that might be Too Clever.

task/host-sp-comms/src/main.rs
evan-oxide: I'd prefer not to mess with it, but you're welcome to open a separate refactoring PR.

labbott:
> FWIW, the `HostToRotCommand::GetCertificates` handler has roughly the same shape, it might be worth sharing some code between them, but that might be Too Clever.

Yeah when I added the TQ code I think I concluded that there were just enough differences that sharing code was annoying and I'd make that a problem for future us.

PR #2622, opened 2026-08-03
tofino-seq-server: add ringbuf counters

hawkw:
As noted in oxidecomputer/customer-support#1485, the `sidecar-seq-server` task has one very big ringbuf which logs a bunch of stuff. This can make it difficult to see if a specific event has ever occurred since boot, since stuff often falls off the end of the ringbuf. We can solve this using counters, which allow us to determine if something was *ever* recorded even if enough other events have been recorded that the thing we're looking for is no longer present in the ringbuf.

This commit adds counters to the Sidecar sequencer server's ringbuf, and derives the `Count` trait for various types in some of its dependencies as needed.

In a few cases, I needed to choose which of a pair of enums in a particular event ought to be the one that got the `#[count(children)]` annotation. I hope that I picked the thing that makes the most sense to count in those cases. If it's necesary to count both of the enums in some of those events, we can restructure the rignbuf entries to nest the enums so that we can count each set of combinations, but that increases the size of the generated counters table.

Surprisingly, it seems like the current RAM and flash allocations for this task are big enough to contain the rather large counters table we are now generating, so it didn't have to be increased.


task/ereportulator/src/main.rs
```diff
@@ -69,6 +69,8 @@ enum Trace {
 
 counted_ringbuf!(Trace, 16, Trace::None);
 
+ringbuf::ringbuf!(__LOL, Trace, 0, Trace::None);
```
Aaron-Hartwig: What is this?

Aaron-Hartwig (commented):
Left one question,but otherwise this LGTM. Thank you!

task/ereportulator/src/main.rs
hawkw: an unstaged diff for testing an unrelated change (#2614) which i committed by mistake. whoops!

Aaron-Hartwig (approved):

Aaron-Hartwig (changes requested):
gotta remove this too

task/ereportulator/src/main.rs
```diff
@@ -79,6 +79,7 @@ fn main() -> ! {
 
     let mut buffer = [0; idl::INCOMING_SIZE];
 
+    ringbuf_entry!(__LOL, Trace::None);
```
Aaron-Hartwig: gotta remove this too

Aaron-Hartwig (changes requested):

task/ereportulator/src/main.rs
hawkw: yeah i also noticed that lol 
image

Aaron-Hartwig (approved):

jamesmunns:
> one very big ringbuf which logs a bunch of stuff

Is it maybe worth shattering this into a couple of ringbufs, or maybe shunting some of the noiser events out to a separate buffer than the higher signal events?

hawkw:
> > one very big ringbuf which logs a bunch of stuff
> 
> Is it maybe worth shattering this into a couple of ringbufs, or maybe shunting some of the noiser events out to a separate buffer than the higher signal events?

That could be worth doing eventually. I was considering that as a possible thing to do if it turns out that the counters table generated for the current ringbuf ended up being really huge. But, I don't actually know enough about how the sequencer works on sidecar to have the "taste" for which entries go where. And, correlating a timeline of events across multiple ringbufs can be difficult unless there's a structure that indicates "each time this entry in the "main ringbuf" occurs, it logs exactly these entries in the "verbose"ringbuf" or similar.

PR #2626, opened 2026-08-05
Update max-stack analysis to account for interrupt stacking requirements

### LINKED ISSUE #2595: Max Stack Analysis code does not account for interrupts
The current max-stack code does not account for the possibility for an interrupt to occur when estimating max stack usage. I believe that this is up to 104 bytes on Cortex-M7/M33, and up to 36 bytes on Cortex-M0(+), when considering the automatically stacked registers and potential alignment corrections.

Interrupts can happen at any time from a task's perspective, either when actively making a syscall, or when the task is pre-empted by SysTick or another hardware interrupt.

We should add this as a fixed number on top of the calculated call-graph stack usage. This likely should be a follow-on to #2589, and will likely cause a lot of manifests to need to be incremented.

CC #2588

jamesmunns:
Closes https://github.com/oxidecomputer/hubris/issues/2595

### DIFF
--- a/build/stack/src/lib.rs
+++ b/build/stack/src/lib.rs
@@ -232,6 +232,7 @@ struct FunctionCallCollector<'a> {
 /// `B` never calls `C` if called by `A`).
 pub fn get_max_stack(
     config: Config,
+    target: &str,
     elf: &Path,
     verbose: bool,
 ) -> Result<Vec<(u64, String)>> {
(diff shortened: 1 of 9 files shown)


build/stack/src/lib.rs
```diff
@@ -253,6 +254,49 @@ pub fn get_max_stack(
...
+                "{{core register interrupt stacking}}".to_string(),
+            ));
+        }
+        "thumbv7m-none-eabihf"
+        | "thumbv7em-none-eabihf"
+        | "thumbv8m.main-none-eabihf" => {
+            chain_compat.push((
+                CORE_REGISTER_SIZE + WORST_CASE_ALIGNMENT_SIZE,
+                "{{core register interrupt stacking}}".to_string(),
+            ));
```
hawkw, line 286: nit: it seems like we could maybe compute `CORE_REGISTER_SIZE + WORST_CASE_ALIGNMENT_SIZE` once, and make that part unconditional, so that we can't somehow forget it? not sure if this matters.

hawkw (approved):
looks good to me, good catch!

build/stack/src/lib.rs
jamesmunns: Open to it, I'll save it for future improvement.

PR #2630, opened 2026-08-10
thermal: enhance fan state tracking

jamesmunns:
This is a step towards #2603, and enhances how we track the state of fans. The main states we now track are:

1. Not Present (according to physical presence detection)
2. Present but unresponsive to I2C queries
3. Present but below a minimum speed threshold
4. Present but above a maximum speed threshold
5. Present and all good

This also introduces a slight "debounce" for reporting state, currently 3s, that prevents us from emitting ringbufs if fans are momentarily deviant/non-nominal, but quickly return to a nominal state. Although this PR doesn't add ereports yet, this is intended to avoid spamming back-to-back ereports for minor hiccups. Right now, the plan is to put ereports where we currently ringbuf. In the future, we might choose to ringbuf immediately, but only send ereports after debouncing.

Marked as draft for now, until I see whether CI passes, and after I make a final docs pass for these changes.

### DIFF
--- a/task/thermal-api/src/lib.rs
+++ b/task/thermal-api/src/lib.rs
@@ -83,6 +83,7 @@ pub enum ThermalAutoState {
     FanParty,
 }
 
+/// Properties for a particular fan in the system
 pub struct FanProperties {
     /// The speed which a fan should never exceed, even at maximum control
     /// input level. May be higher than the expected max RPM of a fan.
@@ -92,10 +93,11 @@ pub struct FanProperties {
     pub underspeed_rpm: Rpm,
 }
 
+/// Generic Sanyo Denki fan limits use in Oxide products
 pub const SANYO_DENKI_FAN_PROPERTIES: FanProperties = FanProperties {
-    // Nominal max speed: 12_000 RPM, +1_500 RPM (total guess)
+    // Nominal max speed: 12_000 RPM, +1_500 RPM variance allowed (total guess)
     overspeed_rpm: Rpm(13_500),
-    // Nominal min speed:  2_000 RPM, -1_500 RPM (total guess)
+    // Nominal min speed:  2_000 RPM, -1_500 RPM variance allowed (total guess)
     underspeed_rpm: Rpm(500),
 };
 
--- a/task/thermal-api/src/lib.rs
+++ b/task/thermal-api/src/lib.rs
@@ -84,6 +84,7 @@ pub enum ThermalAutoState {
 }
 
 /// Properties for a particular fan in the system
+#[derive(Copy, Clone, Debug)]
 pub struct FanProperties {
     /// The speed which a fan should never exceed, even at maximum control
     /// input level. May be higher than the expected max RPM of a fan.
--- a/task/thermal-api/src/lib.rs
+++ b/task/thermal-api/src/lib.rs
@@ -10,7 +10,10 @@ use derive_idol_err::IdolError;
 use drv_i2c_api::ResponseCode;
 use hubpack::SerializedSize;
 use serde::{Deserialize, Serialize};
-use userlib::{units::Celsius, *};
+use userlib::{
+    units::{Celsius, Rpm},
+    *,
+};
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
 #[derive(
@@ -79,6 +82,22 @@ pub enum ThermalAutoState {
     Uncontrollable,
     FanParty,
 }
+
+pub struct FanProperties {
+    /// The speed which a fan should never exceed, even at maximum control
+    /// input level. May be higher than the expected max RPM of a fan.
+    pub overspeed_rpm: Rpm,
+    /// The speed which a fan should never fall below, even at minimum control
+    /// input level. May be lower than the expected min RPM of a fan.
+    pub underspeed_rpm: Rpm,
+}
+
+pub const SANYO_DENKI_FAN_PROPERTIES: FanProperties = FanProperties {
+    // Nominal max speed: 12_000 RPM, +1_500 RPM (total guess)
+    overspeed_rpm: Rpm(13_500),
+    // Nominal min speed:  2_000 RPM, -1_500 RPM (total guess)
+    underspeed_rpm: Rpm(500),
+};
 
 /// Properties for a particular part in the system
 #[derive(Clone, Copy, IntoBytes, FromBytes, Immutable, KnownLayout, Debug)]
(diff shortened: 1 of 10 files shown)


task/thermal-api/src/lib.rs
```diff
@@ -10,7 +10,10 @@ use derive_idol_err::IdolError;
 use drv_i2c_api::ResponseCode;
 use hubpack::SerializedSize;
 use serde::{Deserialize, Serialize};
-use userlib::{units::Celsius, *};
+use userlib::{
+    units::{Celsius, Rpm},
+    *,
```
labbott: If you have the motivation removing the `*` include would be nifty

task/thermal-api/src/lib.rs
```diff
@@ -80,6 +83,25 @@ pub enum ThermalAutoState {
...
+    pub underspeed_rpm: Rpm,
+}
+
+/// Generic Sanyo Denki fan limits use in Oxide products
+pub const SANYO_DENKI_FAN_PROPERTIES: FanProperties = FanProperties {
+    // Nominal max speed: 12_000 RPM, +1_500 RPM variance allowed (total guess)
+    overspeed_rpm: Rpm(13_500),
+    // Nominal min speed:  2_000 RPM, -1_500 RPM variance allowed (total guess)
+    underspeed_rpm: Rpm(500),
+};
```
labbott, line 101: Any docs we can cite for these numbers?

task/thermal/src/bsp/cosmo_ab.rs
```diff
@@ -119,20 +122,18 @@ impl crate::control::BspInterface for Bsp {
...
-    }
+    fn poll_fan_rpms(&mut self) -> impl Iterator<Item = &'_ mut Fan> {
+        if let Ok(fctl) = self.fctrl.try_initialize() {
+            let now = sys_get_timer().now;
+            for fan in self.fans.iter_mut() {
+                let bsp_data = fan.bsp_data;
+                fan.poll_rpm_with(now, &SANYO_DENKI_FAN_PROPERTIES, || {
+                    fctl.fan_rpm(bsp_data).map_err(SensorReadError::I2cError)
+                });
+            }
```
labbott, line 132: Will we be sad if `try_initialize()` fails repeatedly and we have no other error besides an empty iterator?

task/thermal/src/control.rs
```diff
@@ -200,57 +204,199 @@ pub trait BspInterface {
...
+                | (Fps::TooFast(_), Fps::TooSlow(_))
+                | (Fps::TooSlow(_), Fps::Unresponsive(_))
+                | (Fps::TooSlow(_), Fps::TooFast(_))
+                | (Fps::Unresponsive(_), Fps::TooFast(_))
+                | (Fps::Unresponsive(_), Fps::TooSlow(_)) => {
+                    // TODO: Do we want ringbuf entries and/or ereports for
+                    // deviant -> deviant transitions? If so, we should reset
+                    // the since_ms and state_acked values here too, basically
+                    // collapsing this with the Nominal -> Deviant +
+                    // Deviant -> Nominal case above.
```
labbott: Would we risk flooding ereports if we have a fan that is constantly going from, say, unresponsive to to slow? Otherwise I do feel like we would want at least some information about this state

task/thermal/src/control.rs
```diff
@@ -1440,3 +1549,68 @@ impl<'a, B: BspInterface> ThermalControl<'a, B> {
...
+        if !fan.presence_acked {
+            ringbuf_entry!(Trace::FanRemoved(fan.rpm_sensor_id));
+            fan.presence_acked = true;
+        }
+        return;
+    };
+    if !fan.presence_acked {
+        ringbuf_entry!(Trace::FanAdded(fan.rpm_sensor_id));
+        fan.presence_acked = true;
+    }
```
labbott: the `let`/`else` briefly confused me as I tried to follow the logic of why it was `FanRemoved` vs `FanAdded`.  I have a slight preference for a `match` to make it clearer but that would also create some nesting so 🤷

labbott (commented):
Generally LGTM, I appreciate the cleanup and commenting!

task/thermal-api/src/lib.rs
jamesmunns: Removed!

task/thermal/src/control.rs
jamesmunns: Cleaned up, should hopefully be more clear!

task/thermal/src/bsp/cosmo_ab.rs
jamesmunns: So, we won't have an empty iterator, this function returns all fans, even if never initialized, and on non-removable devices, fans start in the `Present(Unresponsive)` state.

We'll definitely be sad because we've never been able to talk to fans, but we will send an ereport for each fan once the fans' state becomes "stable" (after 3s), and really there's not much else to do IMO if we can't talk to the fans other than hope that they've entered their watchdog and are spinning at 100%.

Right now, that doesn't change our control mode, but it could, for example knocking us into "boot" mode.

~THAT BEING SAID, I did just realize that all fans start with a `since_ms` of 0, so if the thermal task restarts, or doesn't get started at system boot time, all of those error timers will fire immediately. I'm going to address that.~ This is now fixed.

task/thermal/src/control.rs
jamesmunns: Partially, yes. I've sort of been trying to categorize things into:

* Not Present
* Present and Good
* Present and Bad

Because I want to send ereports when:

* Things go from NotPresent to Present
* Things go from Present to NotPresent
* Things go from Good to Bad
* Things go from Bad to Good

Since TooFast/TooSlow/Unresponsive are all "Bad", I wasn't sure if it was reasonable to send more ereports when something went from "one kind of bad" to "another kind of bad", especially since they could observe that by monitoring the sensor API data, which they would probably start doing once there's an open fault management case for it.

In my understanding, we want to send ereports when things "become bad" (so a case can be opened), and when things "become good" (so a case can be closed), but I might be reading too much into this.

task/thermal-api/src/lib.rs
jamesmunns: Asking around, the nominal speeds are documented in RFD276 (not public), seeing if I can find a fan part number we can site here with a public datasheet.

task/thermal/src/bsp/cosmo_ab.rs
labbott: I think I missed the extra `iter_mut` below, thanks for the explanation too

task/thermal/src/control.rs
labbott: I think you're right about when we definitely want ereports. I guess if the general disposition is to just replace any fan that's stuck in Bad more ereports wouldn't help. If we don't want an ereport right now I think it's worth keeping the comment and maybe having an explicit ringbuf.

task/thermal/src/control.rs
labbott: Looks better thanks!

labbott (approved):
Fans ENHANCE

Issue #2625, opened 2026-08-04
`monorail` should degrade gracefully when the tech ports are not available


Aaron-Hartwig:
As discovered on https://github.com/oxidecomputer/hubris/issues/1410, the oscillator for the tech port PHY on the front IO board does not always work. We worked around this by power cycling the front IO board when `monorail` detects a problem (see https://github.com/oxidecomputer/hubris/blob/master/task/monorail-server/src/bsp/sidecar_bcd.rs#L270-L291). Something we've seen in the field a small number of times is this loop running infinitely. While we of course need to get to the bottom of that, a step in the right direction in the near term would be to let the rest of the management network operate while noting the tech ports are down. Ideally we'd have an ereport or some structured way of discovering the switch was in this state as well.

hawkw:
If we give up after turning it off and on again several times, do we want to continue trying again when the `monorail-server` loop gets woken up, if it's still down? IIUC, resetting the oscillator requests a power cycle of the whole front IO board, which (I assume) would also take down the transceivers for the data network. Is there a way to *just* mess with the oscillator's enable pin without having to power cycle the entire board, or is that out of the question?

hawkw:
@Aaron-Hartwig are there any registers or other values you would want to see included in an ereport if we can't get the oscillator to oscillate?

jamesmunns:
Noting that we've potentially observed this in R21/hubris v1.76.0 on `london` SW0's sidecar, BRM44220004:

```
jamesmunns@castle ~ $ humility -a ./thermal/build-sidecar-b-image-1_76_0.zip -i fe80::aa40:25ff:fe05:200%3 ringbuf sequencer
humility: connecting to fe80::aa40:25ff:fe05:200%3
humility: ring buffer drv_i2c_devices::bmr491::__RINGBUF in sequencer:
 NDX LINE      GEN    COUNT PAYLOAD
   0  186        1        1 MitigationApplied(NewlyApplied)
humility: ring buffer drv_oxide_vpd::__RINGBUF in sequencer:
humility: ring buffer drv_packrat_vpd_loader::__RINGBUF in sequencer:
humility: ring buffer drv_sidecar_seq_server::__RINGBUF in sequencer:
 NDX LINE      GEN    COUNT PAYLOAD
   8  317        6        1 FrontIOBoardPhyPowerEnable(false)
   9  261        6        1 FrontIOBoardPowerGood
  10  342        6        1 FrontIOBoardPhyPowerEnable(true)
[...]

PR #2631, opened 2026-08-10
sidecar-seq: do the front IO FPGAs in the opposite order

hawkw:
As discussed in [this comment][1] on #2625, it *might* help with The Sidecar Front IO Board PHY Oscillator Sadness Situation if we were to initialize FPGA 1 *before* FPGA 0 instead of in the current order. This would decrease the amount of time between when the VDD rail is powered and when the oscillator's EN pin is asserted, because it is enabled by FPGA 1. This might help reduce the chance of sadness maybe?

[1]: https://github.com/oxidecomputer/hubris/issues/2625#issuecomment-5193231743

Co-authored-by: Aaron Hartwig <aaron@oxidecomputer.com>

### DIFF
--- a/drv/sidecar-seq-server/src/front_io.rs
+++ b/drv/sidecar-seq-server/src/front_io.rs
@@ -45,6 +45,13 @@ impl FrontIOBoard {
     pub fn init(&mut self) -> Result<bool, FpgaError> {
         let mut controllers_ready = true;
 
+        // We reverse the order of interaction here, interacting with
+        // controller 1 and then 0. This is because controller 1 is connected
+        // to the techport PHY and its oscillator. An errata with that
+        // oscillator states that the oscillator should be enabled as close to
+        // power being applied as possible. For this reason, we want to program
+        // controller 1 prior to 0, allowing that to happen a few seconds
+        // sooner than it would if we'd program controller 0 first.
         for (i, controller) in self.controllers.iter_mut().enumerate().rev() {
             let state = controller.await_fpga_ready(25)?;
             let mut ident;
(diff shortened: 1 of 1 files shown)


drv/sidecar-seq-server/src/front_io.rs
```diff
@@ -45,7 +45,7 @@ impl FrontIOBoard {
     pub fn init(&mut self) -> Result<bool, FpgaError> {
         let mut controllers_ready = true;
 
-        for (i, controller) in self.controllers.iter_mut().enumerate() {
+        for (i, controller) in self.controllers.iter_mut().enumerate().rev() {
```
Aaron-Hartwig, line 55: Given the load-bearing `rev` call here, I think that warrants a comment? I'd actually done this as well, but you beat me to opening a PR haha. Here's the comment I had on my branch:
```
        // We reverse the order of interaction here, interacting with
        // controller 1 and then 0. This is because controller 1 is connected
        // to the techport PHY and its oscillator. An errata with that
        // oscillator states that the oscillator should be enabled as close to
        // power being applied as possible. For this reason, we want to program
        // controller 1 prior to 0, allowing that to happen a few seconds
        // sooner than it would if we'd program controller 0 first.
```

Aaron-Hartwig (approved):

drv/sidecar-seq-server/src/front_io.rs
hawkw: yeah, I think you're right, I like the comment --- i'll add that, or we can just use your PR. mostly i thought the six-character change was kinda funny.

PR #2633, opened 2026-08-12
Make jefe's server-impl a static, to make it easier to nab in dumps

### LINKED ISSUE #1980: It is currently hard to get a reset reason out of a dump.
The SP code reads the hardware's reset reason, boils it down (lossily -- see #1978), and then puts it in a variable on the stack. This makes it astonishingly difficult to find in a dump!

This information should wind up in a _static_ instead so we can easily find and print it.

jamesmunns:
This PR moves all of jefe's `ServerImpl` to be a static.

I potentially could have just made the reset reason a static, but we already had some parts in statics, so it seemed more reasonable to just chuck the whole thing in, and inline the parts that were previously separately static'd.

const-ify'ing `task_states` isn't strictly necessary, and was something I did to attempt to fully statically initialize the `ServerImpl`, but unfortunately `dump_areas` is not reasonable to statically initialize, as I was able to find some humpty code paths that implicitly trust that value as a Good Pointer Address, and didn't feel comfortable just marking it as `0x0000_0000` to start with. Making it an `Option<u32>` or so would have been more work than I was interested in doing now.

Closes #1980

### DIFF
--- a/task/jefe/src/main.rs
+++ b/task/jefe/src/main.rs
@@ -108,7 +108,7 @@ fn main() -> ! {
 
     external::set_ready();
 
-    let server = SERVER_IMPL.claim();
+    let server = JEFE_SERVER_IMPL.claim();
     let server = server.write(ServerImpl {
         state: 0,
         deadline,
--- a/task/jefe/src/main.rs
+++ b/task/jefe/src/main.rs
@@ -81,7 +81,7 @@ const MIN_RESTART_DELAY: u64 = 5;
 
 /// Store the entire server impl so that it is retrievable in system dumps
 #[unsafe(no_mangle)]
-static SERVER_IMPL: ClaimOnceCell<MaybeUninit<ServerImpl>> =
+static JEFE_SERVER_IMPL: ClaimOnceCell<MaybeUninit<ServerImpl>> =
     ClaimOnceCell::new(MaybeUninit::uninit());
 
 /// Generate a list of task states, considering the compile-time specified
--- a/task/jefe/src/main.rs
+++ b/task/jefe/src/main.rs
@@ -32,19 +32,31 @@ mod dump;
 
 mod external;
 
-use core::convert::Infallible;
+use core::{convert::Infallible, mem::MaybeUninit};
 
 use hubris_num_tasks::NUM_TASKS;
 use humpty::DumpArea;
 use idol_runtime::RequestError;
+use static_cell::ClaimOnceCell;
 use task_jefe_api::{DumpAgentError, ResetReason};
 use userlib::{Generation, TaskId, kipc};
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Default)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum Disposition {
-    #[default]
     Restart,
     Hold,
+}
+
+impl Disposition {
+    const fn const_default() -> Self {
+        Self::Restart
+    }
+}
+
+impl Default for Disposition {
+    fn default() -> Self {
+        Self::const_default()
+    }
 }
 
 // We install a timeout to periodically check for an external direction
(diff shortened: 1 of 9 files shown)


jamesmunns:
This probably wants a humility PR to add a CLI option to grab this.

jamesmunns:
Poked into the CI failure, at least for the g070 target, is due to now having the ~224 bytes of the `ServerImpl` as a static instead of on the stack. The max-stack analysis dropped from 404 bytes to 228 bytes.

For other targets (like cosmo), this isn't an issue: for example, cosmo uses 200 bytes less flash, 840 bytes more static ram, and 848 bytes less stack after this change.

I can chase down all the ram usage manifest updates if folks are happy with the change otherwise.

edit: fixed some numbers, I was missing a digit.

task/jefe/src/main.rs
```diff
@@ -79,31 +91,37 @@ const MIN_RUN_TIME: u64 = 50;
...
 
-#[unsafe(export_name = "main")]
-fn main() -> ! {
-    let mut task_states = [TaskStatus::default(); hubris_num_tasks::NUM_TASKS];
-    for held_task in generated::HELD_TASKS {
-        task_states[held_task as usize].disposition = Disposition::Hold;
+/// Store the entire server impl so that it is retrievable in system dumps
+#[unsafe(no_mangle)]
+static SERVER_IMPL: ClaimOnceCell<MaybeUninit<ServerImpl>> =
+    ClaimOnceCell::new(MaybeUninit::uninit());
```
labbott, line 85: I feel like this is not the first time we've wanted server data from a dump. I don't think it's a good plan to swap every server over but what's a good way to point others to this pattern in the future? Pulling this to a library/idolatry? Update the docs somewhere?

labbott (commented):
Generally looks okay but i will also let @hawkw give feedback

labbott:
> Poked into the CI failure, at least for the g070 target, is due to now having the ~224 bytes of the `ServerImpl` as a static instead of on the stack. The max-stack analysis dropped from 404 bytes to 228 bytes.
> 

If the g0 gets to be a blocker we can consider if it's time to hit the fork armv6m button (I'd still prefer to not hit that button if we can avoid it)

jamesmunns:
> If the g0 gets to be a blocker we can consider if it's time to hit the fork armv6m button (I'd still prefer to not hit that button if we can avoid it)

as much as I *do* want to hit that button, I don't think it's necessary. The failures are just that the usage moved from "stack" to "ram", so the net usage isn't really different. We're also still pretty healthily within acceptable range even after bumping the usage up, at least on that one target.

hawkw:
> > If the g0 gets to be a blocker we can consider if it's time to hit the fork armv6m button (I'd still prefer to not hit that button if we can avoid it)
> 
> as much as I _do_ want to hit that button, I don't think it's necessary. The failures are just that the usage moved from "stack" to "ram", so the net usage isn't really different. We're also still pretty healthily within acceptable range even after bumping the usage up, at least on that one target.

Yeah, I think we should just adjust the stack size.

task/jefe/src/main.rs
hawkw: yeah, this is generally a good idea, we probably want to do this more or less everywhere if it makes sense to.

task/jefe/src/main.rs
```diff
@@ -32,21 +32,33 @@ mod dump;
...
-    #[default]
     Restart,
     Hold,
 }
 
+impl Disposition {
+    const fn const_default() -> Self {
+        Self::Restart
+    }
+}
```
hawkw: tbh, i might just change anywhere where we can't use the `Default` impl because we're in a const context to just explicitly write `Disposition::Restart`. I'm not sure if indirecting that by having a `const_default()` is actually worth it; the benefit of the `Default` feels like it's mostly just that derived `Default` impls will automagically fill it in...

task/jefe/src/main.rs
```diff
@@ -386,6 +404,15 @@ struct TaskStatus {
     state: TaskState,
 }
 
+impl TaskStatus {
+    const fn const_default() -> Self {
+        Self {
+            disposition: Disposition::const_default(),
+            state: TaskState::const_default(),
+        }
+    }
```
hawkw: again, this kind of feels like we should just explicitly use the defaults for `Disposition` and `TaskState`, and maybe get rid of the `Default` impls entirely --- I don't think they're being used anywhere else. if that's the case, I'd just make this be the `default()` and explicitly say
```suggestion
    const fn new() -> Self {
        Self {
            disposition: Disposition::Restart,
            state: TaskState::Running { started_at: 0 },
        }
    }
```
and then delete all the `const_default()`s and `impl Default`s.

hawkw (approved):
one nitpick about the defaulting, which seems kinda unnecessarily complicated to me. otherwise, looks good.

Issue #2405, opened 2026-02-28
thermal: inlet air over design temp should definitely be an ereport


hawkw:
The thermal loop should emit an ereport if the inlet air is over our stated design temperature range of
> between 35º F (2º C) and 95º F (35º C). 
> (per https://oxide.computer/product/specifications, accessed at time of writing)

Inlet air temp would be best approximated by the temperature measurement from the "South", "Southwest", and "Southeast" sensors.

This is _in addition_ to any ereports we would emit due to thermal shutdowns or overheating events. This way, we can better attempt to distinguish between "computer overheated because something is wrong with it" and "computer overheated because it is sucking up 115ºF air straight out of a _different_ and also very hot computer.".[^1]

We may also want to include inlet temperatures in the ringbuf entries for various thermal loop decisions so that we can make similar determinations when debugging.

[^1]: And also a secret third possibility, which is "computer 'overheated' because the thermal loop is wrong and it was actually fine".

rmustacc:
We're going to have to be very careful with this because the air in question is all pre-heated by drives and therefore any assumption that it is actual inlet temperature may not be correct.

Issue #2316, opened 2025-12-10
ereport: thermal loop overheats or powers off the system


mkeeter:
When the thermal loop enters the `Overheated` or `Uncontrollable` state, it should send an ereport with details on which component is overheated and by how much.

PR #2639, opened 2026-08-13
thermal: Add fan ereports

### LINKED ISSUE #2603: Generate ereport if fan tachometer indicates fan failure

jamesmunns:
Adds ereports for the thermal task, as a follow up to #2630 

closes #2603


task/thermal/src/control.rs
```diff
@@ -1576,13 +1602,101 @@ fn report_fan_state<D>(fan: &mut Fan<D>, sensor_api: &SensorApi, now_ms: u64) {
...
+        FanRemoved(FanRemoved),
+        FanInserted(FanInserted),
+        FanNominal(FanNominal),
+        FanOverspeed(FanOverspeed),
+        FanUnderspeed(FanUnderspeed),
+        FanRpmReadFailed(FanRpmReadFailed),
+    }
+}
+
+#[derive(microcbor::EncodeFields)]
```
jamesmunns, line 1650: @hawkw I'm particularly open to feedback on what to put in these ereports.

jamesmunns:
@hawkw another open question is "should we send ereports on power-on"? Right now on sidecar which has 8 fans, we'll send 8x "fan is/is not present" messages, and 8x "fan is/is not nominal" messages.

I can add some more logic that suppresses this if we want. It seems consistent to send them, but also I don't know if we've observed any "mad rush" of this kind of state transmission, and what we should do if the outgoing ereport queue fills up.

We *also* could pay attention to whether `deliver_ereport` succeeds or not, and try again every fan tick (1hz) until it succeeds, though other services all seem to ignore whether it succeeded or not.

task/thermal/src/control.rs
jamesmunns: Right now I'm sending `SensorId` as `id`, which on one hand is nice, because that's something that could directly be used to query the fan RPM. On the other hand, it's an unstable generated ID, and does nothing to explain which fan this is.

We could include a more string-y "East South East" name (for sidecar), or some kind of refdes, or something. This was the *easiest* to do, but I can plumb other info through if we need it.

hawkw: > Right now I'm sending `SensorId` as `id`, which on one hand is nice, because that's something that could directly be used to query the fan RPM. On the other hand, it's an unstable generated ID, and does nothing to explain which fan this is.
>
> We could include a more string-y "East South East" name (for sidecar), or some kind of refdes, or something. This was the _easiest_ to do, but I can plumb other info through if we need it.

IIUC there is no way currently for the control plane to turn the sensor ID into anything that's actually meaningful to it. While the ID can be used to request fan RPM *from Hubris*, it cannot be used to query fan speeds that the control plane has *previously* read from Hubris and stored in Clickhouse by Oximeter, since those timeseries data points are not labeled by the sensor ID (which is on purpose, because as you mentioned, it is not stable across Hubris builds). 

Therefore, I think we should probably be including the refdes, as that can be used both for querying historical measurements from Clickhouse *and* for reading the current value via `component-details`. Since querying by refdes will give us the metrics for *all* the sensors associated with that fan controller, we will also need to include the _name_ assigned to that sensor in the app.toml. For instance, on sidecar:
https://github.com/oxidecomputer/hubris/blob/c1328f747c812d6da504a0a67d38193e8f4c4db6/app/sidecar/base.toml#L512 

I think that the combination of the refdes and sensor name should be sufficient for the control plane to both retrieve historical measurements from Clickhouse *and* poll new ones from the SP if it chooses to. I'm on the fence as to whether we ought to also keep the sensor ID or not. It's more or less  useless to the control plane, but it's also only a couple bytes, so we may as well throw it in.

I think that it will require a bit more work to plumb through the refdes and sensor names in a way that they can be retrieved from a sensor ID; I started describing that in #2364.

hawkw: My abortive PR https://github.com/oxidecomputer/hubris/pull/2383 included a codegen change for looking up a refdes from a sensor ID. That won't give us the sensor _name_, as written, but it could be added there. I think the reason I didn't finish that at the time is mostly because I had been hoping to figure out a way to do that that didn't require generating a giant LUT of sensor IDs that would use a lot of flash.

One or both of us could blow some of the dust out of that PR, perhaps pulling out just the codegen-related stuff in there?

task/thermal/src/control.rs
```diff
@@ -1576,13 +1602,101 @@ fn report_fan_state<D>(fan: &mut Fan<D>, sensor_api: &SensorApi, now_ms: u64) {
...
+/// An ereport representing a fan becoming underspeed
+#[derive(Encode)]
+#[ereport(class = "hw.fan.rpm.lo", version = 0)]
+struct FanUnderspeed {
+    #[cbor(flatten)]
+    info: FanInfo,
+    rpm: u16,
+}
+
+/// An ereport representing a failure to remove a fan
```
hawkw, line 1697: don't think this comment is quite right

task/thermal/src/control.rs
```diff
@@ -1576,13 +1602,101 @@ fn report_fan_state<D>(fan: &mut Fan<D>, sensor_api: &SensorApi, now_ms: u64) {
...
+#[derive(Encode)]
+#[ereport(class = "hw.fan.rpm.err", version = 0)]
+struct FanRpmReadFailed {
+    id: u32,
```
hawkw, line 1701: it might be nice to include the `i2c::ResponseCode` here as well, so we can indicate what actually went wrong?

task/thermal/src/control.rs
```diff
@@ -1576,13 +1602,101 @@ fn report_fan_state<D>(fan: &mut Fan<D>, sensor_api: &SensorApi, now_ms: u64) {
...
+}
```
hawkw, line 1702: part of me thinks we might really want to have a more general ereport in the shared ereports crate for "trying to read a sensor over I2C failed"?

task/thermal/src/control.rs
jamesmunns: I need to shake off the conversation we had about this with the pmbus-status, I feel like sending `i2c::ResponseCode`, which isn't ABI stable, got debated there.

I can match whatever we did there.

hawkw:
> We _also_ could pay attention to whether `deliver_ereport` succeeds or not, and try again every fan tick (1hz) until it succeeds, though other services all seem to ignore whether it succeeded or not.

This is something I have thought about doing in several places. I think it's not a bad idea, although one deficiency in the way ereports are currently implemented is that the timestamp is always decided by `packrat` when it _receives_ the ereport, which means that if a task is hanging onto an ereport and retrying it, the timestamp associated with the ereport will be the time it was *received* by packrat, not the time at which the thing actually occurred. We might want to have a way for the task to explicitly override the timestamp, for cases like this.

For now, we haven't actually ever dropped ereports due to full buffers, so I haven't been worrying about retries too much _yet_.

jamesmunns:
> For now, we haven't actually ever dropped ereports due to full buffers

Do we have a way to reasonably detect when this happens? I can add a little stress and see if we hit it, and back off if not.

task/thermal/src/control.rs
jamesmunns: Ah, okay, I remember now. We included it with a significant red flag of "DON'T ACTUALLY USE THIS JUST FOR LOGGING AND DEBUGGING PLZ":

https://github.com/oxidecomputer/management-gateway-service/blob/a19831fa86daa4ec253d50ccb4394bd9abb693de/gateway-messages/src/sp_to_mgs.rs#L799-L811

hawkw:
> Do we have a way to reasonably detect when this happens?

there are ringbuf counters in each task that will tell you precisely how many times it tried and failed to submit an ereport. but, more importantly, [the ereport ring buffer code in packrat will create a "loss report"](https://github.com/oxidecomputer/hubris/blob/c5a836ef6a319dd10c0cd3ec5bd1b27701090310/lib/snitch-core/src/lib.rs#L337-L372) for upstack software that tells it "hey, i have dropped this many ereports, sorry about that". this is part of the same stream as the actual ereports. as far as i know, we have never actually seen such a loss report in a production system.

> I can add a little stress and see if we hit it, and back off if not.

for what it's worth, such a test is really only going to be interesting if there _is_ a control plane collecting ereports (and therefore draining the buffer); if you run it against a bench system, you *will* see a bunch of stuff get dropped, but that's kind of anticipated.

task/thermal/src/control.rs
hawkw: okay, i think https://github.com/oxidecomputer/hubris/pull/2645 should've added all the requisite codegen :)

task/thermal/src/control.rs
```diff
@@ -1576,13 +1606,101 @@ fn report_fan_state<D>(fan: &mut Fan<D>, sensor_api: &SensorApi, now_ms: u64) {
 
     // Step three: handle state reporting, if unreported
     if !fan.state_acked {
-        let trace = match pres {
-            Fps::Unresponsive(e) => Trace::FanReadFailed(id, e),
-            Fps::Nominal(_) => Trace::FanNominal(id),
-            Fps::TooFast(rpm) => Trace::FanOverspeed(id, rpm),
-            Fps::TooSlow(rpm) => Trace::FanUnderspeed(id, rpm),
+        let fan_info = || FanInfo {
+            name: fan.name,
```
jamesmunns: So @hawkw interesting point here:

If I do `id.name()` here, it ends up inflating the `.rodata`, likely with the table of all sensor names:

```
james@fool:~/oxide/hubris git:(james/ereport-thermals) ✗ arm-none-eabi-size -A target/cosmo-b-dev/dist/thermal.elf
target/cosmo-b-dev/dist/thermal.elf  :
section                size   addr
.text                 17632      0
.rodata                4640      0
.data                   580      0
.bss                   2452      0
```

But with the change to make `.name()` a `const fn`, and storing `.name` in the const-fn constructor of `Fan::new()`, then we don't pull in the whole array:

```
james@fool:~/oxide/hubris git:(james/ereport-thermals) ✗ arm-none-eabi-size -A target/cosmo-b-dev/dist/thermal.elf
target/cosmo-b-dev/dist/thermal.elf  :
section                size   addr
.text                 17544      0
.rodata                2552      0
.data                   628      0
.bss                   2452      0
.uninit                   0      0
```

I don't think this is a make or break, a ~2KiB flash difference (on cosmo) isn't really THAT big of a deal, but it does seem lightly footgun-y if you aren't Too Clever about it.

task/thermal/src/control.rs
jamesmunns: Added in df409fbc7

task/thermal/src/control.rs
jamesmunns: I would be open to discuss this! How do we still maintain the context: "this i2c read is specifically for a fan and we are sad we don't have its data"?

jamesmunns:
@hawkw this should be ready for a fresh look, IMO.

task/thermal/src/control.rs
```diff
@@ -1576,13 +1611,125 @@ fn report_fan_state<D>(fan: &mut Fan<D>, sensor_api: &SensorApi, now_ms: u64) {
...
+    info: FanInfo,
+    rpm: u16,
+}
+
+/// An ereport representing a failure to read from a fan
+#[derive(Encode)]
+#[ereport(class = "hw.fan.rpm.err", version = 0)]
+struct FanRpmReadFailed {
+    name: fixedstr::FixedStr<'static, MAX_SENSOR_NAME_LEN>,
+    component_id: fixedstr::FixedStr<'static, MAX_COMPONENT_ID_LEN>,
```
hawkw: Every other ereport calls these `refdes`, not `component_id`, so I think we should make these consistent:
```suggestion
struct FanInfo {
    name: fixedstr::FixedStr<'static, MAX_SENSOR_NAME_LEN>,
    refdes: fixedstr::FixedStr<'static, MAX_COMPONENT_ID_LEN>,
    lo_rpm_lim: u16,
    hi_rpm_lim: u16,
}

/// An ereport representing a fan being removed
#[derive(Encode)]
#[ereport(class = "hw.remove.fan", version = 0)]
struct FanRemoved {
    name: fixedstr::FixedStr<'static, MAX_SENSOR_NAME_LEN>,
    refdes: fixedstr::FixedStr<'static, MAX_COMPONENT_ID_LEN>,
}

/// An ereport representing a fan being inserted
#[derive(Encode)]
#[ereport(class = "hw.insert.fan", version = 0)]
struct FanInserted {
    name: fixedstr::FixedStr<'static, MAX_SENSOR_NAME_LEN>,
    refdes: fixedstr::FixedStr<'static, MAX_COMPONENT_ID_LEN>,
}

/// An ereport representing a fan entering a nominal state
#[derive(Encode)]
#[ereport(class = "hw.fan.ok", version = 0)]
struct FanNominal {
    #[cbor(flatten)]
    info: FanInfo,
}

/// An ereport representing a fan becoming overspeed
#[derive(Encode)]
#[ereport(class = "hw.fan.rpm.hi", version = 0)]
struct FanOverspeed {
    #[cbor(flatten)]
    info: FanInfo,
    rpm: u16,
}

/// An ereport representing a fan becoming underspeed
#[derive(Encode)]
#[ereport(class = "hw.fan.rpm.lo", version = 0)]
struct FanUnderspeed {
    #[cbor(flatten)]
    info: FanInfo,
    rpm: u16,
}

/// An ereport representing a failure to read from a fan
#[derive(Encode)]
#[ereport(class = "hw.fan.rpm.err", version = 0)]
struct FanRpmReadFailed {
    name: fixedstr::FixedStr<'static, MAX_SENSOR_NAME_LEN>,
    refdes: fixedstr::FixedStr<'static, MAX_COMPONENT_ID_LEN>,
```

hawkw: I'm also not sure if I like the fairly generic `name`, either. I might consider either `sensor_name` or just `sensor`, since it's specifically the name we assign to the sensor in metrics etc.

hawkw: I would also really like to have a "slot: u8" or something with the fan slot number as marked on the chassis, as in https://rfd.shared.oxide.computer/rfd/0200#_naming_conventions. I'm not sure if that is as necessary on compute sled fan trays since the whole fan tray is the FRU, but fundamentally we need to be able to tie this to a physical location in the hardware in a consistent way.

task/thermal/src/control.rs
hawkw: So, I was thinking more that we might want to have a conventional field name/payload structure for an I2C read error that would include the raw response code and either a refdes or a controller/bus/mux/addr (I think refdes is more concise) and include that structure in any I2C error ereport. Having that payload struct be included in the "fan sensor read error" ereport is how we would indicate it was for a fan.

I think it's fine to punt on this now and go and factor it out only once we are trying to make ereports like this in multiple places.

hawkw (commented):
I left some suggestions on the ereport structures. I would really like it if we could figure out a way to get the fan slot numbers as marked on the Sidecar chassis here, but I realize there may not be a great way to do that with the way the sensor config works currently...

jamesmunns:
@hawkw one thing I noticed is that there is a slight discrepancy in how we number fans at the chassis level and how the thermal task "thinks" about fans. For example on sidecar:

Screenshot 2026-09-03 at 13 37 50

The chassis thinks in terms of "fan assemblies", each of which are a pair of counter-rotating fans with individual control and monitoring. When reporting in the ereport, would you like the slot to reflect the **chassis** numbering (e.g. we'll have `[0, 0, 1, 1, 2, 2, 3, 3]`), or the **logical** ordering (e.g. we'll have `[0, 1, 2, 3, 4, 5, 6, 7]`), OR add a second field (e.g. we'll have `[(0, 0), (0, 1), ... (3, 0), (3, 1)]`, or get cute and pack it into one `u8` still (e.g. we'll have `[00, 01, ... 30, 31]`), or even go full enum on it (e.g. we'll have `[Fan0A, Fan0B, ... Fan3A, Fan3B]`)?

In 04bc61802 I added plumbing for the "slot" field to the ereport, but the numbering currently is almost certainly wrong (I need to make sure that the ordering defined in the BSP initializer matches the schematic/chassis at least for all PCBAs).

edit: The fans are also definitely not numbered in the order that I've placed them in currently:

Screenshot 2026-09-03 at 13 48 46

jamesmunns:
Also @hawkw should we rename these functions to `refdes` and `sensor` respectively, to avoid other folks making the same mistake I did wrt naming?

https://github.com/oxidecomputer/hubris/blob/1cbf835c1db3c8d9a7a53274e5af38e002c1c244/task/sensor-api/src/lib.rs#L85-L102

jamesmunns:
Chatted with @hawkw:

- [ ] We probably want to focus on the "field replaceable unit", e.g. the fan assembly unit, when we talk about "slots". I need to hunt down this info from mechanical details
- [ ] As a short term hack, I MAY just revert the `slot` field for now, and add it back later in a versioned rev.
- [ ] This is actually a bit more complicated field to deal with, because a "fan assembly" isn't the refdes+name that we're reporting from, because it's coming from the fan controller's perspective
- [ ] We probably need to discuss how to report this upstack, and take into account things like the VPD eeprom on fans, which are a different refdes and such.
- [ ] I will open a separate hubris issue to discuss "how do we think about this", it may be more of a host software concern, but it's at least useful to capture here while it's fresh, and will be relevant to how hubris should report this in ereports and such.

jamesmunns:
Booted this on a cosmo, I see the "fan added" ereports as expected on boot:

```
jamesmunns@alfred ~ $ pfexec ./humility0908 -t cosmo-hubris-sp -a ./cosmo-b-5f51ba353.zip ereport dump
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP V1
task:      packrat (6)
timestamp: 0
ena:       1
  lost: null

task:      thermal (8)
timestamp: 9498
ena:       2
  class: hw.fan.ok,
  version: 0,
  sensor: "V12_U2A_A0",
  refdes: "J200/U1",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9498
ena:       3
  class: hw.fan.ok,
  version: 0,
  sensor: "V12_U2A_A0",
  refdes: "J200/U1",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9498
ena:       4
  class: hw.fan.ok,
  version: 0,
  sensor: "V12_U2A_A0",
  refdes: "J200/U1",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9498
ena:       5
  class: hw.fan.ok,
  version: 0,
  sensor: "V12_U2A_A0",
  refdes: "J200/U1",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9498
ena:       6
  class: hw.fan.ok,
  version: 0,
  sensor: "V12_U2A_A0",
  refdes: "J200/U1",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9498
ena:       7
  class: hw.fan.ok,
  version: 0,
  sensor: "V12_U2A_A0",
  refdes: "J200/U1",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500
```

jamesmunns:
Actually, those are *not* the details I expect.

```
[[config.i2c.devices]]
bus = "front"
mux = 1
segment = 1
address = 0b0111_000
device = "max5970"
description = "U.2 Sharkfin A hot swap controller"
power = { rails = [ "V12_U2A_A0", "V3P3_U2A_A0" ], pmbus = false }
sensors = { voltage = 2, current = 2 }
name = "sharkfin_a_hsc"
refdes = ["J200", "U1"]
removable = true
```

vs

```
[[config.i2c.devices]]
bus = "rear"
address = 0x20
device = "max31790"
description = "Fan controller"
sensors = { speed = 6, names = [
    "Southeast", "Northeast", "South", "North", "Southwest", "Northwest"
] }
refdes = "U58"
```

I'm gunna look into this.

jamesmunns:
Okay that's better:

```
jamesmunns@alfred ~ $ pfexec ./humility0908 -t cosmo-hubris-sp -a ./cosmo-b-d3e14015a.zip ereport dump
humility: attached to 1fc9:0143:XJACFXSEKTQJS via CMSIS-DAP V1
task:      packrat (6)
timestamp: 0
ena:       1
  lost: null

task:      thermal (8)
timestamp: 9497
ena:       2
  class: hw.fan.ok,
  version: 0,
  sensor: "Southeast",
  refdes: "U58",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9497
ena:       3
  class: hw.fan.ok,
  version: 0,
  sensor: "Northeast",
  refdes: "U58",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9497
ena:       4
  class: hw.fan.ok,
  version: 0,
  sensor: "South",
  refdes: "U58",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9497
ena:       5
  class: hw.fan.ok,
  version: 0,
  sensor: "North",
  refdes: "U58",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9497
ena:       6
  class: hw.fan.ok,
  version: 0,
  sensor: "Southwest",
  refdes: "U58",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500

task:      thermal (8)
timestamp: 9497
ena:       7
  class: hw.fan.ok,
  version: 0,
  sensor: "Northwest",
  refdes: "U58",
  lo_rpm_lim: 500,
  hi_rpm_lim: 13500
```

task/thermal/src/control.rs
```diff
@@ -308,31 +324,27 @@ impl<D> Fan<D> {
...
-                // Any of the following:
-                //
-                // - Nominal -> Deviant
-                // - Deviant -> Nominal
-                // - Deviant -> Deviant
-                //
-                // Take:
-                //
-                // - New state
-                // - Status ack state
```
hawkw, line 327: is this not still the case?

task/thermal/src/control.rs
```diff
@@ -1576,13 +1621,124 @@ fn report_fan_state<D>(fan: &mut Fan<D>, sensor_api: &SensorApi, now_ms: u64) {
...
+    basic_info: BasicFanInfo,
+}
+
+/// An ereport representing a fan being inserted
+#[derive(Encode)]
+#[ereport(class = "hw.insert.fan", version = 0)]
+struct FanInserted {
+    #[cbor(flatten)]
+    basic_info: BasicFanInfo,
+}
```
hawkw, line 1704: I would really like it if the fan insert and remove events could also include the FRU ID (i.e. the barcode) of the fan tray, but I'm fine with punting on this for now and saying that upstack software has to go and figure that out with the `component_vpd` API we just added. This is part of the difficulty with these specific component IDs we're reporting not actually representing the FRU but a specific IC (as we discussed in #2673), since there isn't a super easy way to map "this fan was inserted" -> "here is what you're supposed to read to find its serial". As much as I really want this, I'm fine with having it land in fan ereports V1 instead of V0.

task/thermal/src/control.rs
```diff
@@ -1576,13 +1621,124 @@ fn report_fan_state<D>(fan: &mut Fan<D>, sensor_api: &SensorApi, now_ms: u64) {
...
+#[ereport(class = "hw.fan.rpm.err", version = 0)]
+struct FanRpmReadFailed {
+    #[cbor(flatten)]
+    basic_info: BasicFanInfo,
+    /// The raw I2C driver code reported when this query failed. This value
+    /// is not stable across versions of the SP firmware, and should only
+    /// be logged or used for interactive or post-mortem debugging.
+    /// Requires knowledge of the exact firmware revision to meaningfully
+    /// decode.
+    raw_response_code: u8,
```
hawkw, line 1743: turbo nitpick: i might consider shortening this to save a few bytes of CBOR:
```suggestion
    raw_rsp_code: u8,
```

task/thermal/src/bsp/sidecar_bcd.rs
```diff
@@ -521,40 +516,26 @@ const MISC_SENSORS: [TemperatureSensor; NUM_TEMPERATURE_SENSORS] = [
...
+        fans[idx] = Fan::new(
+            sensors::MAX31790_SPEED_SENSORS[idx],
+            SANYO_DENKI_FAN_PROPERTIES,
+            MaxFan::new_const(FAN_ORDER[idx]),
+        );
+        idx += 1;
+    }
+
+    fans
+}
```
hawkw, line 539: man, i both love and hate const fn! :D

hawkw (approved):
Okay, this looks good to me: I think there is a great deal more we can (and should, hopefully soon!) do here, but I think we're better off having the ereports even if we don't currently have all the potential data we would want to include in them (i.e., #2673).

One procedural note: I think when we merge this, we should change the commit message to `thermal: Add fan ereports` or something, because every time I see "thermal: add ereports", I immediately think of ereports for stuff going overtemp (see 
#2316, #2405) which this doesn't add (and which we should probably do in a follow-up!).

task/thermal/src/control.rs
jamesmunns: Added a comment block about this referring to that issue.

task/thermal/src/control.rs
jamesmunns: No, we decided at some point that we do want to re-report if a fan goes from "too slow" to missing, for example, so we reset the `state_acked` flag to false.

PR #2644, opened 2026-08-17
xtask: I2C codegen

jamesmunns:
This adds a new command, `xtask i2c-codegen`, which allows for printing what *would* be generated using `build-i2c` for a given manifest.

This is intended to be used as a way of regression/snapshot testing some potential future I2C codegen improvements.

For example, running:

```sh
$ cargo xtask i2c-codegen app/cosmo/rev-b-dev.toml -d sensors --output target/i2c-xtask-codegen.rs --fmt
```

You would obtain this file as output: https://gist.github.com/jamesmunns/3588214c2933f90615ff2be24f0c7ac3


build/xtask/src/main.rs
```diff
@@ -263,6 +264,23 @@ enum Xtask {
...
+    I2cCodegen {
+        /// Path to the image configuration file, in TOML.
+        cfg: PathBuf,
+
+        #[clap(short, long)]
+        disposition: String,
+
+        /// File to write to
+        #[clap(long)]
+        output: Option<PathBuf>,
```
hawkw: What does this do if you don't pass the argument? I assume it prints it? It would be nice if that was written down in the CLI help text.

build/xtask/src/main.rs
```diff
@@ -263,6 +264,23 @@ enum Xtask {
...
         attestation: Option<PathBuf>,
     },
+
+    /// Codegen I2C information
+    I2cCodegen {
+        /// Path to the image configuration file, in TOML.
+        cfg: PathBuf,
+
+        #[clap(short, long)]
+        disposition: String,
```
hawkw: it would be nice if this was a [`clap` `ValueEnum`](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#enumerated-values), so that the help text told you which possible values are accepted here. 

Perhaps the easiest way to do that could be to add a feature-flagged `clap` dependency to `build-i2c` and derive the trait for `Disposition` there, so that we don't have to duplicate the whole enum here. But, we could also have a separate version of it in `xtask` and convert between them, alhough that kinda just feels worse IMO.

build/xtask/src/main.rs
```diff
@@ -263,6 +264,23 @@ enum Xtask {
...
+
+        /// File to write to
+        #[clap(long)]
+        output: Option<PathBuf>,
+
+        /// Invoke rustfmt. Requires `--output`.
+        #[clap(long)]
```
hawkw: adding
```suggestion
        #[clap(long, requires("output")]
```
will make clap enforce this.

build/util/src/lib.rs
```diff
@@ -306,11 +306,15 @@ fn toml_from_env<T: DeserializeOwned>(var: &str) -> Result<Option<T>> {
...
     };
 
-    let rval =
-        toml::from_str(&config).context("deserializing configuration")?;
+    let rval = toml_from_str(&config)?;
     Ok(Some(rval))
 }
 
+/// Partse the contents of text as toml.
+pub fn toml_from_str<T: DeserializeOwned>(contents: &str) -> Result<T> {
```
hawkw, line 314: turbo nitpick but how about `from_toml_str` instead? "TOML from str" sounds to me like it's taking a str and turning it *into* TOML, but it's actually taking a string which contains toml and parsing it.

hawkw: upon further consideration i see that there's already a `toml_from_env` so disregard this.

build/xtask/src/i2c_codegen.rs
```diff
@@ -0,0 +1,69 @@
...
+    if let Some(p) = output {
+        let mut f = File::create(p)?;
+        f.write_all(res.as_bytes())?;
+        f.flush()?;
+        drop(f);
+        if fmt {
+            call_rustfmt::rustfmt(p)?;
+        }
+    } else {
+        assert!(!fmt, "--fmt only works with --output");
```
hawkw: if we follow my note from https://github.com/oxidecomputer/hubris/pull/2644#discussion_r3797175059, this can be removed.

build/xtask/src/i2c_codegen.rs
```diff
@@ -0,0 +1,69 @@
...
+    // we do in normal builds where `xtask dist` will prepare the app toml and
+    // shove it in an env var, and then i2c codegen will pull it from there.
+    //
+    // We convert to a string using *xtask*'s notion of manifest tomls...
+    let config = toml::to_string(&cfg.config)?;
+
+    // ...and now that it's a string, parse the contents back as *i2c*'s
+    // different notion of what a manifest toml looks like (mostly just the
+    // i2c config section).
+    let i2c_cfg: build_i2c::Config = build_util::toml_from_str(&config)?;
```
hawkw, line 30: This is all kind of wacky, but I think the comments make sense. :thumbsup:

build/xtask/src/i2c_codegen.rs
```diff
@@ -0,0 +1,69 @@
...
+        "sensors" => Disposition::Sensors,
+        // devices are used, but only as validation
+        "validation" => Disposition::Validation,
+        // ???
+        other => bail!("Unknown disposition: '{other}'"),
+    })
+}
+
+/// Do I2C code generation
+pub fn run(
```
hawkw, line 12: It occurs to me that, eventually, we may also want a thing that prints some of the internal data structures like `DeviceDescription`, which are inputs to *other* tasks codegen. Not saying we have to do that now, but it could be worth adding subsequently.

build/xtask/src/i2c_codegen.rs
```diff
@@ -0,0 +1,69 @@
...
+        // devices are used (i.e., controller is not used), but not as sensors
+        "devices" => Disposition::Devices,
+        // devices are used, with some used as sensors
+        "sensors" => Disposition::Sensors,
+        // devices are used, but only as validation
+        "validation" => Disposition::Validation,
+        // ???
+        other => bail!("Unknown disposition: '{other}'"),
+    })
+}
```
hawkw: if you take my suggestion from https://github.com/oxidecomputer/hubris/pull/2644#discussion_r3797169748, this can all go away.

build/xtask/src/i2c_codegen.rs
jamesmunns: I 100% agree :)

hawkw (approved):
great, thank you for putting this together!

labbott (approved):

Issue #2624, opened 2026-08-04
Want to read or identify sensors by name over the management network


adamlouis:
Would like to be able to identify & read sensors by name over the management network.

We can read sensors over the management network with `faux-mgs read-sensor-value <ID>`. e.g.:

```
$ faux-mgs read-sensor-value 25
...
value:     52.0625
timestamp: 5122574
```

Alternatively, we can read the temperature if we know the parent component in `inventory` and then look at children in `component-details` - e.g.:

 ```
$ faux-mgs component-details U64
...
Measurement(Measurement { name: "tf2", kind: Temperature, value: Ok(51.875) })
```

However (as I understand!), to get the `<ID>` for `read-sensor-value`, we first need `humility manifest`, which unfortunately requires knowing the live hubris version & having the archive. e.g.:

```
$ humility --archive=/data/local/images/sidecar/d/sp/build-sidecar-d-image-default-v1.76.0.zip manifest | grep tf2
humility: attached to archive
                 25 tf2                     i2c id=11   temp
```

Any solution that removes the `humility` or `inventory` + `component-details` steps & allows identifying & reading sensors would help, for example, (1) listing sensors by name & ID or (2) reading sensors by name.

```
$ faux-mgs list-sensors
...
25 tf2
...
```

or

```
$ faux-mgs read-sensor-value tf2
...
value:     52.0625
timestamp: 5122574
```

For context, we are expanding sensor coverage in manufacturing - ideally in Rust via both MGS & humility.

adamlouis:
cc @jamesmunns - this may fit in with your recent sensors & thermals work!

jamesmunns:
I'll take a peek at what it would take to enable this, I agree it would be good to have. For the thermal stuff I was doing, I manually pulled the list of sensors out with `humility manifest` like you mentioned above.

Issue #501, opened 2022-04-14
I2C codegen could use some love


mkeeter:
The I2C codegen in `build-i2c` dates back to early Hubris development, and has languished as other systems have been improved.

For example, we do a [manual search](https://github.com/oxidecomputer/hubris/blob/master/drv/stm32h7-i2c-server/src/main.rs#L48-L73) through the I2C mux table, when this could be known at compile time.

Making this system more robust / ergonomic / obvious would have downstream benefits for future development.

jamesmunns:
Noting this is still relevant today. Marking as triaged.

I think at least it might be worth reworking `build/i2c/src/main.rs` some time soon-ish to:

1. Separate the "IR stage" of doing analysis on the input from the "codegen" stage
2. (maybe) Add some way of dumping that IR for debugging "why is/isn't this rail/device generating this code"
3. (maybe) replace the `write!` oriented codegen with `quote` or similar

The current shape of things of multiple pieces follow the pattern of:

1. read in the app tomls
2. do some sort of meta analysis (gather all pmbus device, gather all temperature sensors, etc.)
3. generate code based on that, using writeln!, with manual indenting in an attempt to pretty-format the output

Right now steps 2 + 3 are largely are done inline (and are repeated/re-implemented), "separate IR from codegen" means breaking these steps into discrete pieces.

hawkw:
> 1. Separate the "IR stage" of doing analysis on the input from the "codegen" stage

I'm strongly in favor of this. I have wanted for a longish time to separate the step of building the various indices of devices by name/kind/refdes from the step of actually generating the code. Currently, the `I2cDeviceDescriptions` stuff is *kind* of like an "IR" that's a validated and indexed representation of the parsed TOML contents, but it's only used for exposing it in the API to other tasks' build scripts, and the codegen functions build their _own_ indices which are in some cases duplicative of this and in other cases not duplicative. Which...isn't great.

[...]

PR #2645, opened 2026-08-17
sensors-api: add lookup tables from SensorID to component ID/name

### LINKED ISSUE #2364: unfortunately, a way to turn a SensorId into a refdes is probably required

hawkw:
As I described in #2364, there are several places (mostly in the `thermal` task for now, but this may change) where it is necessary to be able to go from a numeric, build-specific `SensorId` to a sensor _name_ (stable, defined in the app.toml), or a `ComponentId` (the same, but also used for SP/MGS communication). This branch implements such a mechanism by enhancing the `task-sensor-api` build script with the (optional) ability to generate lookup tables mapping a sensor ID index to the corresponding component ID and/or sensor name strings. This is feature flagged so that only tasks that depend on the ability to look up these values can generate the LUTs.

While working on this, I also did a bit of refactoring to `build-i2c`; in particular, I changed it up so that we are now storing `DeviceSensor` in an `Arc` and using `iddqd::IdOrdMap` to store them, so that the ID field defined in the `DeviceSensor` struct need not be duplicated to serve as the map key. I think there's room to continue along these lines by adding newtypes that wrap an `Arc<DeviceSensor>` and implement `IdOrdItem` with all of the different key types by which we might look up a sensor, and replace some of the other maps with a similar pattern. I didn't do that here to keep the change fairly minimal, but I think there are a lot of places where we might benefit from using Rain's work on `iddqd` in the I2C codegen datastructures.

Also, it was necessary to add refdes strings to a few sensors in `app.toml`s that didn't previously have one. So I did that too.

Fixes #2364


hawkw:
This is currently a draft because I need to figure out what to put in the (now mandatory) `refdes` field for Sidecar's QSFP temperature sensors.

hawkw:
using @jamesmunns's new xtask from #2644, we can see that there's no actual change to the `build-i2c` codegen. of course, the `task-sensors-api` will now include the new LUTs, if they're enabled:

```console
eliza@hekate ~/Code/oxide/hubris $ cargo xtask i2c-codegen -d sensors --fmt --output cosmo-i2c-before.rs app/cosmo/rev-b.toml
     Running `target/debug/xtask i2c-codegen -d sensors --fmt --output cosmo-i2c-before.rs app/cosmo/rev-b.toml`
will invoke: /home/eliza/.rustup/toolchains/1.95.0-x86_64-unknown-linux-gnu/bin/rustfmt

eliza@hekate ~/Code/oxide/hubris $ git co eliza/turn-an-id-into-a-refdes
Switched to branch 'eliza/turn-an-id-into-a-refdes'
Your branch is ahead of 'origin/eliza/turn-an-id-into-a-refdes' by 4 commits.
  (use "git push" to publish your local commits)

eliza@hekate ~/Code/oxide/hubris $ cargo xtask i2c-codegen -d sensors --fmt --output cosmo-i2c-after.rs app/cosmo/rev-b.toml
   Compiling build-i2c v0.1.0 (/home/eliza/Code/oxide/hubris/build/i2c)
   Compiling xtask v1.0.0 (/home/eliza/Code/oxide/hubris/build/xtask)
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.90s
     Running `target/debug/xtask i2c-codegen -d sensors --fmt --output cosmo-i2c-after.rs app/cosmo/rev-b.toml`
will invoke: /home/eliza/.rustup/toolchains/1.95.0-x86_64-unknown-linux-gnu/bin/rustfmt

eliza@hekate ~/Code/oxide/hubris $ diff cosmo-i2c-before.rs cosmo-i2c-after.rs

eliza@hekate ~/Code/oxide/hubris $ 
```

build/i2c/src/lib.rs
```diff
@@ -1829,7 +1865,7 @@ impl ConfigGenerator {
 
         self.generate_footer()?;
 
-        Ok(self.output)
+        Ok((self.output, outputs))
```
jamesmunns: Would it maybe make more sense to make `CodegenOutputs` contain both the `output` string as well as the `sensors` contents?

That way we could return a single struct instead of a tuple.

I guess then `codegen_to_string` doesn't make as much sense, but I'm open to other names.

task/sensor-api/build.rs
```diff
@@ -27,17 +28,29 @@ struct Sensor {
...
 }
 
 fn main() -> Result<()> {
     idol::client::build_client_stub("../../idl/sensor.idol", "client_stub.rs")
         .map_err(|e| anyhow!("idol error: {e}"))?;
 
-    build_i2c::codegen(build_i2c::Disposition::Sensors)?;
+    let i2c_outputs = build_i2c::codegen(build_i2c::Disposition::Sensors)?;
+
+    let i2c_sensors = i2c_outputs.sensors.expect(
```
jamesmunns, line 43: I don't love that we have an `Option` field vs a different codegen method, but I also don't know if it's worth a more invasive change (yet).

It might be worth documenting that `CodegenOutputs.sensors` is only populated in `Disposition::Sensors`.

I do wonder if at some point we want to get rid of the disposition enum and have one codegen method per disposition instead.

Mostly thinking out loud, feel free to ignore for this PR if it's not worth it.

task/sensor-api/build.rs
```diff
@@ -52,32 +65,31 @@ fn main() -> Result<()> {
...
         let mut sensors_text = String::new();
-        let mut sensor_id = 0;
         for d in &config_sensor.devices {
             for (sensor_type, &sensor_count) in d.sensors.iter() {
                 let sensor = format!(
-                    "{}_{}_{}",
+                    "{}_{}_{sensor_type}",
                     d.device.to_ascii_uppercase(),
                     d.name.to_ascii_uppercase(),
-                    sensor_type.to_ascii_uppercase()
```
jamesmunns, line 62: Does it matter we lost the `to_ascii_uppercase()`?

task/sensor-api/src/lib.rs
```diff
@@ -80,6 +81,25 @@ impl SensorId {
...
     }
+
+    /// Returns the component ID (refdes) corresponding to this sensor.
+    ///
+    /// Note that multiple sensor IDs may have the same component ID, when a
+    /// single device exposes multiple measurement channels.
+    #[cfg(feature = "component-id-lookup")]
+    pub fn component_id(
+        &self,
+    ) -> fixedstr::FixedStr<'static, { config::MAX_COMPONENT_ID_LEN }> {
```
jamesmunns, line 92: Could we get away with returning `&'static fixedstr::FixedStr<'static, { config::MAX_COMPONENT_ID_LEN }` here to avoid making a stack copy, if we're likely going to just copy this into some output structure anyway?

jamesmunns (commented):
Overall looks reasonable to me, added some suggestions.

At some point, I wonder if we should move the `sensor-api` and `validate` codegen into `build-i2c`, if for no other reason than to benefit from any refactoring and the ability to snapshot the output. I don't think that needs to be done in this PR, but might be something to keep in mind for #501.

jamesmunns:
re-kicked CI, it looks like it timed out, likely due to github jank during the outage yesterday.

jamesmunns:
https://github.com/oxidecomputer/hubris/issues/2624 seems to be looking for the *inverse* of this, I wonder if thats worth doing as a follow-up, or as part of this.

build/i2c/src/lib.rs
hawkw: yeah, i thought about this, was trying to do the smallest possible diff in the rebase. happy to move that!

I think maybe the move then would be to make `codegen_to_string` just be `codegen` and rename the current `codegen` to `codegen_to_file` or something?

task/sensor-api/build.rs
hawkw: The change here was due to my switching the parser from parsing the sensor type as a `String` to using the `Sensor` enum in `build_i2c`. This gets us validation that it's one of the known sensor types at parse time (the previous code wouldn't fail until compilation), but more importantly, it lets us use it to construct a `DeviceSensor` to put in the map of IDs to `DeviceSensor`s that `build_i2c` spits out.

Losing the `to_ascii_uppercase()` is not a problem here because the `Display` impl for `build_i2c::Sensor` already prints the uppercase string:
https://github.com/oxidecomputer/hubris/blob/4efa6f8d64ab76e3ac5de0eeb54743d07e84bda1/build/i2c/src/lib.rs#L458-L490

task/sensor-api/src/lib.rs
hawkw: `FixedStr` is a `&'static str` already: https://github.com/oxidecomputer/hubris/blob/4efa6f8d64ab76e3ac5de0eeb54743d07e84bda1/lib/fixedstr/src/lib.rs#L321-L324

It's `FixedString` that's the owned array. I tried to preserve the `String`/`Str` convention with these types' names.

task/sensor-api/build.rs
hawkw: I also didn't love the `Option` here. Another option (heh) could be a `DispositionOutputs` or something, but the consumer would still have to match that. I think the better change would be to move from the `Disposition` enum to separate codegen methods with shared code they all call into, as you suggested...and I think I'd rather do that refactoring in a separate PR, so I'm somewhat inclined to leave the option rather introducing more ceremony just to remove it later. What do you think?

jamesmunns: Very reasonable and sounds good to me! That would make the pieces easier to snapshot test as well, as then they become per-method instead of per-disposition.

hawkw:
I started an additional change to add an option to print the `I2cSensorsDescription` in `xtask i2c-codegen`, but it turns out the `fmt::Debug` output is like 500 lines long. I'll probably make it print as a table instead or something, but I think that belongs in a follow-up PR

build/i2c/src/lib.rs
hawkw: Done in https://github.com/oxidecomputer/hubris/pull/2645/commits/9c0d54198c204d2060cd119f48fad0335d1a6129

jamesmunns (approved):

evan-oxide (approved):

PR #2648, opened 2026-08-18
build-i2c: Add snapshot tests for i2c codegen

jamesmunns:
This adds `insta`-based snapshot tests for I2C codegen, as a follow-up to #2644.

I'm not entirely convinced this is a good idea yet, nor if we want to do it this way. The snapshot files are:

* generally huge (up to 1.6k-6.3k currently)
* Per target + per disposition (right now only Sensors and the current hardware revs)
* The test is in `xtask` for pragmatic reasons

I think in the future we are likely to have some kind of "IR" for codegen, that will be easier to snapshot + dump. However, until then, snapshotting the actual code output could still be useful, as this new testing will be done by CI.


evan-oxide:
You could consider constructing toml files that are just used for testing, so that the test will only break if codegen changes, not if someone intentionally adds a sensor to a product.

jamesmunns:
> You could consider constructing toml files that are just used for testing, so that the test will only break if codegen changes, not if someone intentionally adds a sensor to a product.

I think that's a reasonable idea! That being said, right now I think the production tomls are the best input corpus that we have, and coming up with inputs that would appropriately exercise the different codegen features/options might take more time than we want in the short term.

I do hope that *after* we've gotten things refactored a bit, we might consider phasing out the snapshot tests in favor of more directed Manifest -> IR, IR -> Codegen tests. Once we can write more targeted tests, it's probably a lot less value for using snapshot tests as a shotgun approach to regression checks.

Maybe it makes sense to have some kind of planned expiration date to evict the snapshot tests? I think they have more value For Now than In The Long Term, if that makes sense. If that's a reason to keep them out of the repo, I could also understand that, and we could just ask people to run it ad-hoc when actually making changes to codegen.

jamesmunns:
I think for this we might want to snapshot the `generate` functions instead of whole output or by disposition.

evan-oxide:
You could make copies of the production tomls that will live in the test directory. But it could get a bit confusing to have those copies floating around, and the production tomls probably won't change often anyway :shrug:.

I agree that we should think of these tests as a temporary tool for refactoring. It would be good to have a comment explaining that so that if we forget to remove it, somebody who breaks the test in a few years knows that it's ok to rip it out. And a benefit of committing them rather than running them ad hoc is that reviewers can see how a PR has changed the output.

> I think for this we might want to snapshot the generate functions instead of whole output or by disposition.

Why? So that the test failure message is more specific about where the change was?

jamesmunns:
> Why? So that the test failure message is more specific about where the change was?

There's a lot of repetition in the generation, if we wanted to cover more dispositions:

```rust
match self.settings.disposition {
    Disposition::Target => {
        let n = self.ncontrollers();

        if n != 1 {
            //
            // If we have the disposition of a target, we expect exactly
            // one controller to be configured as a target; if none have
            // been specified, the task should be deconfigured.
            //
            anyhow::bail!(
                "found {n} I2C controller(s); expected exactly one"
            );
        }

        self.generate_controllers()?;
        self.generate_pins()?;
        self.generate_ports()?;
    }

    Disposition::Initiator => {
        self.generate_controllers()?;
        self.generate_pins()?;
        self.generate_ports()?;
        self.generate_muxes()?;
    }

    Disposition::Devices => {
        self.generate_devices()?;
        self.generate_ports()?;
    }

    Disposition::Sensors => {
        self.generate_devices()?;
        let desc = self.generate_sensors()?;
        output_sensors = Some(desc);
    }

    Disposition::Validation => {
        self.generate_devices()?;
        self.generate_validation()?;
    }
}
```

Doing snapshots per-generate function would reduce the repetition (and size) of the snapshots.

build/xtask/tests/i2c-codegen.rs
```diff
@@ -0,0 +1,110 @@
...
+
+//! Okay, so this is actually testing `build/i2c`, and the test for this should
+//! probably live there, but right now the easiest way to invoke code generation
+//! is by using the xtask implementation. We can't depend on xtask from
+//! build-i2c because that would lead to circular dependencies.
+//!
+//! So, pragmatically, xtask is going to host the snapshot testing party for
+//! build-i2c to avoid churning a lot of other things.
+//!
+//! ALSO, note that the existence of this type is intended to be a temporary
```
evan-oxide: Did you mean "this test"?

jamesmunns: Thanks, fixed!

evan-oxide (approved):

hawkw:
> > You could consider constructing toml files that are just used for testing, so that the test will only break if codegen changes, not if someone intentionally adds a sensor to a product.
> 
> I think that's a reasonable idea! That being said, right now I think the production tomls are the best input corpus that we have, and coming up with inputs that would appropriately exercise the different codegen features/options might take more time than we want in the short term.

It occurs to me that another option is to snapshot _the production tomls as they currently exist_ and use them as the inputs for the snapshot tests, rather than the state of the production tomls at the time the tests are run. This way, changes to a board such as adding a sensor won't break the snapshot tests. On the other hand, in that universe, changes to the structure we require in the TOMLs (i.e. making a field mandatory for all i2c devices, renaming things, etc) will require updating both the real board TOMLs *and* the ones that are used only for snapshot testing, which is also kinda unfortunate. I dunno.

jamesmunns:
> It occurs to me that another option is to snapshot the production tomls as they currently exist and use them as the inputs for the snapshot tests, rather than the state of the production tomls at the time the tests are run. This way, changes to a board such as adding a sensor won't break the snapshot tests. On the other hand, in that universe, changes to the structure we require in the TOMLs (i.e. making a field mandatory for all i2c devices, renaming things, etc) will require updating both the real board TOMLs and the ones that are used only for snapshot testing, which is also kinda unfortunate. I dunno.

This would also require us to either pre-process the app toml to have all the includes, *and* potentially do... something for the validate logic, since the code we generate is dependent on that.

I'll keep the merge going, but if this gets annoying to maintain we can come up with a reasonable plan B.

build/xtask/tests/snapshots/i2c_codegen__app_cosmo_rev-b-dev.toml.controllers-Initiator.snap
hawkw, line 1: So, I don't know enough about `insta` to know if this is possible, but it is a bit unfortunate IMO that these have the filename extension `.snap`, rather than `.rs`. It would be nice if github rendered these with Rust syntax highlighting. With `expectorate`, the golden file's extension can be whatever you want it to be, and it doesn't require any metadata in the file itself, but it looks like `insta` adds some YAML frontmatter which will probably break the Rust syntax highligher even if the extension was correct? I wonder if it would let us put that inside a Rust comment or something...

build/xtask/tests/i2c-codegen.rs
```diff
@@ -0,0 +1,110 @@
...
+//! ALSO, note that the existence of this test is intended to be a temporary
+//! band-aid for a lack of any testing for `build-i2c`. The hope is to someday
+//! test this more directly using unit tests of analysis and much more targeted
+//! snapshot testing of generation behavior, which will allow for much smaller
+//! fragments and less breakage on intentional changes.
+//!
+//! Apologies to those that need to update these snapshots until that day comes.
+//! You will need to install `cargo-insta`, e.g. `cargo install cargo-insta`,
+//! run the tests with `cargo insta test -p xtask`, and then review+bless any
+//! new changes with `cargo insta review`.
```
hawkw, line 22: This also feels like a bit of a tradeoff between `insta` and `expectorate`, IMO: `expectorate` is *just* a library, so the workflow for updating the golden/snapshot files is just to re-run the tests with `EXPECTORATE=overwrite` set, and the user need not install any additional tools. On the other hand, it seems cool that `insta` has a review tool (which I presume shows you a diff between the new and old state of the golden file?), and that's part of the workflow for blessing new changes...

PR #2649, opened 2026-08-18
WIP: Slight i2c codegen retouch

jamesmunns:
This is currently staged as a follow-on to #2648, and does two main things:

* Splits the `output` out of `ConfigGenerator`, to make it easier to see methods that mutate that from ones that don't (currently none, analysis is done locally in the codegen functions themselves)
* Moves some of the inline "check the environment variables/project structure" stuff into `CodegenSettings` so that you could invoke codegen directly without actually being in the given project structure

This is likely to have merge conflicts, but I'm gunna work on other stuff for a while, so I wanted to push this up just to have it.

### DIFF
--- a/build/i2c/src/lib.rs
+++ b/build/i2c/src/lib.rs
@@ -502,9 +502,6 @@ pub struct ConfigGenerator {
     /// Settings
     settings: CodegenSettings,
 
-    /// output that we're building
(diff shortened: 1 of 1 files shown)


jamesmunns:
merged into #2648

PR #2652, opened 2026-08-19
sidecar: use RDL-generated types for sequencer state, step, and error

Aaron-Hartwig:
We have a few sequencing types which were shared across the FPGA and the SP which needed to be manually kept in sync. This change adjusts the SP-side of things to rely on the RDL-defined types (depends on https://github.com/oxidecomputer/quartz/pull/522 landing first) so if the FPGA-side definition changes, the SP will pick that up automatically. This is intended to support #2515.


drv/sidecar-mainboard-controller/src/tofino2.rs
```diff
@@ -10,86 +10,12 @@ use drv_fpga_user_api::power_rail::*;
...
-    PowerVrHot = 3,
-    PowerAbort = 4,
-    SoftwareAbort = 5,
-    VidAckTimeout = 6,
-    ThermalAlert = 7,
-}
+// Types to provide more ergonomic access to RDL defined types
+pub type TofinoSeqState = Reg::TOFINO_SEQ_STATE::ValueEncoded;
+pub type TofinoSeqStep = Reg::TOFINO_SEQ_STEP::ValueEncoded;
+pub type TofinoSeqError = Reg::TOFINO_SEQ_ERROR::ValueEncoded;
```
Aaron-Hartwig, line 16: @hawkw I removed the counter derive here because we're now relying on an the FPGA codegen. I could add that counter derive there too, but then this change really starts to reach out into other drivers.

drv/sidecar-mainboard-controller/sidecar_mainboard_controller.adoc
```diff
@@ -1,537 +0,0 @@
-
```
Aaron-Hartwig, line 1: I've stopped checking in the `.adoc` files we generate as I've not heard of them really being used, so they're just noise. I know Matt just relies on the JSON (also used for codegen), but I find more human centered formats (like HTML or adoc) to be useful. I'm leaving the HTML here for now.

drv/sidecar-seq-server/src/main.rs
```diff
@@ -83,7 +83,6 @@ enum Trace {
     TofinoSequencerAbort {
         state: TofinoSeqState,
         step: TofinoSeqStep,
-        #[count(children)]
```
Aaron-Hartwig, line 86: As noted above, I removed the count derivation on `TofinoSeqError` (affects L109 below too).

Aaron-Hartwig (commented):
@hawkw I botched the FPGA register map type plumbing to hiffy such that it can't find the types I specify in the register map:
```
$ humility hiffy -c Sequencer.tofino_seq_state                           humility: attached via ST-Link V3
humility hiffy failed: couldn't find type drv_sidecar_seq_api::reg_map::Reg::TOFINO_SEQ_STATE::ValueEncoded for Sequencer.tofino_seq_state
```

I'll keep poking at this, but just an FYI.

Aaron-Hartwig:
Okay I fixed the type location bug and things are happy now.
```
$ humility hiffy -c Sequencer.tofino_seq_state
Sequencer.tofino_seq_state() => A0

$ humility hiffy -c Sequencer.tofino_seq_error
Sequencer.tofino_seq_error() => None

$ humility hiffy -c Sequencer.tofino_seq_error_step
Sequencer.tofino_seq_error_step() => Init
```

drv/sidecar-mainboard-controller/sidecar_mainboard_controller.json
evan-oxide, line 1: Does this mean https://github.com/oxidecomputer/hubris/blob/master/drv/sidecar-mainboard-controller/README.md should be updated?

And out of curiosity, are you hand-editing the json or is it generated from something in quartz?

drv/sidecar-mainboard-controller/sidecar_mainboard_controller.adoc
hawkw: personally, if i had known the ASCIIDoc existed, i would have really wanted to use it --- i've often wished we had that for the Cosmo sequencer RDL types...

drv/sidecar-mainboard-controller/sidecar_mainboard_controller.html
```diff
@@ -1367,11 +1390,21 @@ <h1>sidecar_mainboard_controller</h1>
...
+        <br>
+        <br>
+            None = 0x0 <br>
+            PowerGoodTimeout = 0x1 <br>
+            PowerFault = 0x2 <br>
+            PowerVrHot = 0x3 <br>
+            PowerAbort = 0x4 <br>
+            SoftwareAbort = 0x5 <br>
+            VidAckTimeout = 0x6 <br>
+            ThermalAlert = 0x7
```
hawkw, line 1407: it would be neat (and don't worry about it in _this_ PR) if we could rig up a way to get the codegen to emit an `Option<{enum type for the variants starting at 0x1}>` here. Rust will still represent that in a single byte thanks to [niche optimization](https://niche.rs/about), and it makes the semantic difference between "None" and the actual errors a bit clearer.

don't worry about figuring that out in this branch though.

drv/sidecar-mainboard-controller/src/tofino2.rs
hawkw: It's kind of a shame that we'll no longer be able to generate counters for these. I see why you didn't just add it to the list of derives in the codegen crate, because we probably don't want to generate a `Count` implementation for _every_ RDL type (and indeed, some of them may not be able to derive it), but losing the more detailed error counters makes me sad.

I wonder if what we really ought to do is rig up a way to configure the codegen with a list of "please add these additional derives _on these specific type names_" someplace. That way, we could explicitly request that there be counters for the `TOFINO_SEQ_ERROR` enum, and not have to derive the trait for _every_ enum.

If you don't want to spend the time on the codegen change in this branch, I won't hold it against you, but i do think the counter is likely to be useful for debugging.

drv/sidecar-mainboard-controller/sidecar_mainboard_controller.adoc
Aaron-Hartwig: I'm happy to leave it here! And we can get that added for the Cosmo sequencer too.

drv/sidecar-mainboard-controller/sidecar_mainboard_controller.json
Aaron-Hartwig: It should be updated! I will do that once the PR with these changes (https://github.com/oxidecomputer/quartz/pull/522) gets merged.

No hand editing! quartz generates the VHDL, BSV, JSON, and adoc/HTML based off a BUCK build rule (see https://github.com/oxidecomputer/quartz/blob/f5d4bc95a723450a79237d3d717a4531a1cc3486/hdl/projects/sidecar/mainboard/BUCK#L4-L17 for an example). We manually copy them over to hubris as needed. Obviously some room for improvement on the hubris/quartz integration there!

drv/sidecar-mainboard-controller/src/tofino2.rs
Aaron-Hartwig: I think those particular errors will be less interesting as we get the #2515 merged and perhaps other subsequent ereport generation stuff here, but I agree removing them is not something we should take lightly -- which why I highlighted having done so!. I'll grab some time today to discuss this with you (@evan-oxide will throw you on the invite too if you can make it).

hawkw: We agreed that losing the counters for the error enum is not something that's worth blocking this PR over, but that it might be nice to do some work around configuring the codegen to request additional derives, separately from this PR. I wrote up https://github.com/oxidecomputer/hubris/issues/2656 to describe that.

Aaron-Hartwig:
@hawkw and @evan-oxide this is ready for another look

evan-oxide (approved):

Aaron-Hartwig:
Thank you for the approval here @evan-oxide! @hawkw do you care to give a +1 here at all?

PR #2653, opened 2026-08-19
observer: remove PSU's auxiliary rail

evan-oxide:
It turns out that the MWOCP67 PSU's auxiliary rail is not accessible over PMBus, so this PR removes it from the manifest. I had to update the i2c codegen snapshot tests, but it was actually quite nice to see what had changed.

 Tested by running `humility power` and confirming that now only the main rail of each PSU is shown.

### DIFF
--- a/app/observer/base.toml
+++ b/app/observer/base.toml
@@ -568,6 +568,9 @@ mux = "port_b"
 cs = [{port = "B", pin = 12}]
 
 # TODO SPI3 is the ignition flash reprogramming port
+# We validated that the hardware works but didn't add proper Observer support to
+# the ignition-flash task. For details and example code, see
+# https://github.com/oxidecomputer/hardware-observer/issues/90
 
 [config.spi.spi4]
 controller = 4
--- a/app/observer/base.toml
+++ b/app/observer/base.toml
@@ -440,10 +440,14 @@ name = "psu0mcu"
 address = 0b1011_000
 device = "mwocp67"
 description = "PSU 0 MCU"
-# TODO it's not clear what rails the mwocp67 has or how they are paged.
-# https://github.com/oxidecomputer/hubris/issues/2410
-power = { rails = [ "V50_MAIN_PSU0", "V50_AUX_PSU0" ], sensors = ["voltage", "current", "input-voltage", "input-current"] }
-sensors = { input-voltage = 2, input-current = 2, voltage = 2, current = 2, temperature = 5, speed = 1 }
+# The MWOCP67 PSU has two rails: a 50V main rail and a 12V auxiliary rail that
+# powers the Observer PSC even when the main rail is off. However, only the main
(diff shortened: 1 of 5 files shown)


app/observer/base.toml
```diff
@@ -440,10 +440,14 @@ name = "psu0mcu"
...
-# TODO it's not clear what rails the mwocp67 has or how they are paged.
-# https://github.com/oxidecomputer/hubris/issues/2410
-power = { rails = [ "V50_MAIN_PSU0", "V50_AUX_PSU0" ], sensors = ["voltage", "current", "input-voltage", "input-current"] }
-sensors = { input-voltage = 2, input-current = 2, voltage = 2, current = 2, temperature = 5, speed = 1 }
+# The MWOCP67 PSU has two rails: a 50V main rail and a 12V auxiliary rail that
+# powers the Observer PSC even when the main rail is off. However, only the main
+# rail is accessible over PMBus. The PSU does allow you to set the page to 0 or
+# 1, but this has no effect; in either case, commands like READ_VOUT just report
+# readings for the main rail, and Murata has confirmed that this is the intended
+# behavior. Therefore, we only define the main rail here.
```
hawkw, line 448: man, that's _so cool_. why would they design it like that???

drv/psc-seq-server/src/main.rs
```diff
@@ -1235,7 +1235,7 @@ struct PowerUngoodEreport {
 #[derive(microcbor::EncodeFields)]
 struct EreportFields {
     refdes: FixedStr<'static, 20>, // Component ID max length
-    rail: FixedString<13>,         // Example: "V54_PSU0"
+    rail: FixedString<8>,          // Example: "V54_PSU0"
```
hawkw, line 1238: well, that's nice, i guess, at least.

app/observer/base.toml
```diff
@@ -440,10 +440,14 @@ name = "psu0mcu"
 address = 0b1011_000
 device = "mwocp67"
 description = "PSU 0 MCU"
-# TODO it's not clear what rails the mwocp67 has or how they are paged.
-# https://github.com/oxidecomputer/hubris/issues/2410
-power = { rails = [ "V50_MAIN_PSU0", "V50_AUX_PSU0" ], sensors = ["voltage", "current", "input-voltage", "input-current"] }
```
hawkw, line 445: it has occurred to me that even if the aux rail _did_ work, we would have needed to change its name here, since it's not 50V.

hawkw (approved):
well, that's unfortunate (but in no way your fault). good job murata, very nice.

jamesmunns (approved):
Glad to hear the snapshot updating wasn't painful!

This only affects the mwocp67 and not the mwocp68, right?

evan-oxide:
> This only affects the mwocp67 and not the mwocp68, right?

Correct, we can still read the mwocp68's aux rail.

PR #2665, opened 2026-09-01
onewire: Remove onewire support

jamesmunns:
I noticed that `thermal` had a dependency on the `onewire` crate, which it wasn't using.

Looking back at the history, [RFD114](https://rfd.shared.oxide.computer/rfd/0114), an early treatise on thermal management, discussed evaluating the use of this sensor for the sharkfins, but we ultimately went a different direction.

For the sake of Less Code Is More Better, this PR removes the driver and associated pieces from the repo.


hawkw (approved):
i think this is pretty reasonable, personally; we can always get it back from the Git history if we need 1wire later...

Issue #2664, opened 2026-09-01
thermal/transceivers: QSFP temperature may not be reliable


jamesmunns:
We have observed QSFP transceivers reporting temperatures that have been hot enough to cause an immediate power-down of the sidecar as part of the thermal monitoring task.

However, we have also now seen cases where the reported temperatures are suspicious, for example:

```
  23 1470        7     4917 ControlPwm(0x0)
  24 1582        7        1 PowerDownDueTo { sensor_id: SensorId(0x4a), worst_case_temp: Celsius(123.08184) }
  25 1586        7        1 LastRealTemperature { sensor_id: SensorId(0x4a), temperature: Celsius(122.83984), age_s: 0.484 }
  26 1603        7        1 AutoState(Uncontrollable)
  27 1474        7        1 PowerDownAt(0x3aaa0f10)
  28 1247        7        1 PowerModeChanged(PowerBitmask(0b1))
  29 1102        7        1 AutoState(Boot)
  30 1506        7        1 AutoState(Running)
  31 1470        7    65535 ControlPwm(0x0)
```

In this case, the system was at 0% fans one second, meaning that the transceivers are all under their nominal 65C limit. The next second, we power down due to a transceiver reporting 122.8C. The next second, we are back in Running mode with the fans commanded at 0%, which means the transceiver was again under 65C.

It seems extremely unlikely that the QSFP would experience a 60C excursion over a few seconds.

We have also seen some reports of transceivers reporting *negative* temperatures, which are again extremely unlikely in a climate controlled datacenter.

For these reasons, we may want to modify `transceivers-server` to report or reject potential "glitches". Mechanisms proposed so far include:

1. ALWAYS poll each temperature sensor twice in a row, if the readings are outside some reasonable span (5C or so), reject both samples and try again on the next tick.
2. If a temperature sensor reports a suspicious value, THEN repoll. This could be:
    1. If one transceiver is significantly different than other ones (if present)
    2. If the temperature jumps more than some amount between 1hz samples (some kind of moving average?)
    3. Some more intelligent filter, like using a Kalman filter and monitoring the sample variance
3. Poll temperature sensors more often (currently 1Hz), but potentially report less often, discarding any suspicious samples

We may want to avoid pre-filtering/averaging potentially bad data: a very large bad-sample could distort the average, and will always introduce additional lag on top of the filtering delay in the thermal PID loop.

Issue #2656, opened 2026-08-20
want a way to add attribute to types generated by `fpga-regmap`


hawkw:
This originally came up in https://github.com/oxidecomputer/hubris/pull/2652#discussion_r3817895716.

It would be nice if there was a way for users of the `fpga-regmap` codegen to request that additional Rust attributes be added to generated types. This is mostly so that we can add derive attributes for additional traits we'd like the generated types to implement (such as `counters::Count`, `microcbor::Encode` for ereports,  serde traits, etc). 

I think it's probably better to be able to pass in an arbitrary attribute rather than "derive these additional traits", because that way, we can also request an attribute like `#[cfg(feature = "counters", derive(counters::Count)]`. This way, if the regmap codegen is being included in a library crate rather than a driver, the crate can feature flag things like counters or serialization. This is mostly important in the Tofino sequencer regmap currently, since types it generates are used by a couple different tasks.

I imagine the interface to this would be something like a `CodegenSettings` type in `fpga-regmap` that provides a builder interface to add additional attributes. Then, the [`drv-sidecar-mainboard-controller` build script](https://github.com/oxidecomputer/hubris/blob/4c371f01d1d52235ae5dfcb0ca15adf8e4d6e674/drv/sidecar-mainboard-controller/build.rs) could write something like this:

```rust
    let out_dir = build_util::out_dir();
    let out_file = out_dir.join("sidecar_mainboard_controller.rs");
    let mut file = fs::File::create(out_file)?;
    
    let fpga_codegen = fpga_regmap::CodegenSettings::default()
        // Naming subject to change...
        .additional_attribute(
           "Reg::TOFINO_SEQ_ERROR::ValueEncoded", // the name of the generated type
           "#[cfg_attr(feature = \"counters\", derive(counters::Count)]", // attribute
        )
        // example of multiple types
        .additional_attribute(
           "Reg::TOFINO_SEQ_STEP::ValueEncoded",
           "#[derive(microcbor::Encode)]",
        );
    write!(
        &mut file,
        "{}",
        fpga_codegen.fpga_regs(include_str!("sidecar_mainboard_controller.json"))?,
    )?;
```

Then, when generating the register map, the codegen would check if a generated type name is in the map of types that want additional attributes, and add those attributes to the generated code. 

[...]

PR #2668, opened 2026-09-02
transceivers: Implement double-polling for transceiver temperatures

jamesmunns:
CC https://github.com/oxidecomputer/hubris/issues/2664 (though probably not a permanent solution).

This PR:

1. Makes the transceiver server a static, so we can pull it from dumps
2. Now polls each transceiver twice, discarding samples that are >= 5.0C apart
3. Keeps stats (in the server) about each enabled port, including max sample variance and number of discarded samples

The hope for this PR is to get more trackable information and maybe catch a misbehaving I2C poll in action.

We may want to also add an ereport for this, I'd like to leave that for a follow-on PR though.

This is also currently based on #2667, we should merge that first.

This did require some refactoring of the transceiver server code, I tried to keep the diffs pretty stable and the commits pretty useful so far, if that makes things easier to review. I'll flag a couple of "wuts" I saw along the way.

### DIFF
--- a/drv/front-io-api/src/transceivers.rs
+++ b/drv/front-io-api/src/transceivers.rs
@@ -2,9 +2,12 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use core::sync::atomic::{AtomicU32, Ordering};
+
 use crate::{Addr, FrontIOError, Reg};
 use drv_fpga_api::{FpgaError, FpgaUserDesign, ReadOp, WriteOp};
 use drv_transceivers_api::{ModuleStatus, NUM_PORTS};
+use ringbuf::Count;
 use transceiver_messages::ModuleId;
 use userlib::UnwrapLite;
 use zerocopy::{
@@ -134,6 +137,24 @@ impl LogicalPort {
         PortLocation::from(*self)
     }
 }
+
+/// Implement the ringbuf trait on LogicalPort to allow for per-port metrics
+impl Count for LogicalPort {
+    type Counters = [AtomicU32; NUM_PORTS as usize];
+
+    #[allow(clippy::declare_interior_mutable_const)]
+    const NEW_COUNTERS: Self::Counters =
+        [const { AtomicU32::new(0) }; NUM_PORTS as usize];
+
+    fn count(&self, counters: &Self::Counters) {
+        // This should never happen, but just in case.
+        let Some(ctr) = counters.get(self.0 as usize) else {
+            return;
+        };
+        ctr.fetch_add(1, Ordering::Relaxed);
+    }
+}
+
 /// Represents a set of selected logical ports, i.e. a 32-bit bitmask
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 pub struct LogicalPortMask(pub u32);
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -138,7 +138,7 @@ counted_ringbuf!(Trace, 16, Trace::None);
 
 #[derive(PartialEq, Debug, Clone, Copy)]
 struct TempGlitch {
-    index: LogicalPort,
+    port: LogicalPort,
     first: Celsius,
     second: Celsius,
 }
@@ -152,7 +152,7 @@ impl TempGlitch {
     /// didn't feel like making this an Option, or using u8::MAX as the index,
     /// or even adding 1 to the port and then using NonZero or something.
     const RINGBUF_INIT: Self = Self {
-        index: LogicalPort(0),
+        port: LogicalPort(0),
         first: Celsius(0.0),
         second: Celsius(0.0),
     };
@@ -170,7 +170,7 @@ impl Count for TempGlitch {
 
     #[inline]
     fn count(&self, counters: &Self::Counters) {
-        self.index.count(counters);
+        self.port.count(counters);
     }
 }
 
@@ -240,9 +240,10 @@ struct PortData {
     /// Number of times the temperature has been samples. This counts resample
     /// times, not individul i2c queries
     total_temp_samples: u32,
-    /// Thermal models are populated by the host
-    // TODO(AJM): Is the above comment true? We actually fill this in with a
-    // basic model for each, and I don't *think* there's a UDP api to set this?
+    /// Thermal models for the given transceiver. Currently this is fixed to
+    /// a single basic model.
+    ///
+    /// See https://github.com/oxidecomputer/hubris/issues/2670.
     model: Option<ThermalModel>,
 }
 
@@ -496,6 +497,8 @@ impl XcvrApi {
             ManagementInterface::Sff8636 | ManagementInterface::Cmis => {
                 ringbuf_entry!(Trace::GotInterface(p.0, interface));
                 // TODO: this is made up
+                //
+                // See: https://github.com/oxidecomputer/hubris/issues/2670
                 Some(ThermalModel {
                     interface,
                     model: ThermalProperties {
@@ -533,7 +536,7 @@ impl XcvrApi {
             ringbuf_entry!(
                 TEMP_GLITCH_RINGBUF,
                 TempGlitch {
-                    index: port,
+                    port: port,
                     first: a,
                     second: b,
                 }
@@ -687,24 +690,17 @@ impl ServerImpl {
                     // Log error to ringbuf
                     e.ringbuf(port);
 
-                    // TODO(AJM): Old behavior here is a bit funky, we should
-                    // review this.
                     match e {
-                        // Previously, this did a "continue", and didn't affect
-                        // consecutive errors at all. This should never happen
-                        // because we only add models to known interface types
-                        TempReadError::UnknownInterface => {}
+                        // Failure to read neither increments nor resets the
+                        // counter of errors
+                        TempReadError::UnknownInterface
+                        | TempReadError::BadTempRead(_)
+                        | TempReadError::BadPortStatus(_)
+                        | TempReadError::UnexpectedFpgaErr(_) => {}
                         // This would *increment* consecutive errors
                         TempReadError::PotentialRemoval(_) => {
                             meta.consecutive_errors =
                                 meta.consecutive_errors.saturating_add(1);
-                        }
-                        // All of these *actually reset* the error count. Do
-                        // we want this?
-                        TempReadError::BadTempRead(_)
-                        | TempReadError::BadPortStatus(_)
-                        | TempReadError::UnexpectedFpgaErr(_) => {
-                            meta.consecutive_errors = 0;
                         }
                         // We probably don't want to count this against the
                         // device, since it could have been a momentary I2C
@@ -847,8 +843,9 @@ impl From<FpgaError> for TempReadError {
             ErrorEncoded::I2CAddressNack => true,
             ErrorEncoded::I2CSclStretchTimeout => true,
 
-            // TODO(AJM): why *don't* we consider these errors as worth
-            // potentially invalidating the QSFP?
+            // We expect these are transient errors, and will likely be resolved
+            // in short order, either by waiting to detect the QSFP device has
+            // been removed, or obtaining some other kind of error.
             ErrorEncoded::NoError => false,
             ErrorEncoded::NoModule => false,
             ErrorEncoded::NoPower => false,
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -5,7 +5,10 @@
 #![no_std]
 #![no_main]
 
-use core::mem::MaybeUninit;
+use core::{
+    mem::MaybeUninit,
+    sync::atomic::{AtomicU32, Ordering},
+};
 
 use counters::Count;
 use idol_runtime::{NotificationHandler, RequestError};
@@ -125,6 +128,65 @@ enum Trace {
 }
 
 counted_ringbuf!(Trace, 16, Trace::None);
+
+#[derive(PartialEq, Debug, Clone, Copy)]
+struct TempGlitch {
+    index: u8,
+    first: Celsius,
+    second: Celsius,
+}
+
+impl TempGlitch {
+    /// Okay so ringbuf requires an initializer, and this value is actually
+    /// somewhat obnoxious because it *is* a plausible value to observe.
+    /// However, "all zeroes" lets us put the initializer in .bss and not .data,
+    /// and since this is just for debugging, we hope that a reasonable person
+    /// would realize that the delta between the two entries is zero. I really
+    /// didn't feel like making this an Option, or using u8::MAX as the index,
+    /// or even adding 1 to the port and then using NonZero or something.
+    const RINGBUF_INIT: Self = Self {
+        index: 0,
+        first: Celsius(0.0),
+        second: Celsius(0.0),
+    };
+}
+
+/// Keep counters for how often each port has experienced temperature glitches.
+///
+/// This *does not* reset when ports are disabled or qsfp xcvrs are removed or
+/// re-added.
+impl Count for TempGlitch {
+    type Counters = [AtomicU32; NUM_PORTS as usize];
+
+    #[allow(clippy::declare_interior_mutable_const)]
+    const NEW_COUNTERS: Self::Counters =
+        [const { AtomicU32::new(0) }; NUM_PORTS as usize];
+
+    fn count(&self, counters: &Self::Counters) {
+        // This should never happen, but just in case.
+        let Some(ctr) = counters.get(self.index as usize) else {
+            return;
+        };
+        ctr.fetch_add(1, Ordering::Relaxed);
+    }
+}
+
+// Keep counters each time the delta between two samples exceeds the value
+// of `MAX_RESAMPLE_VARIANCE_CELSIUS`. This also counts how often it happens
+// on a per-port basis, as well as the last 32 instances it has been observed.
+//
+// This augments the per-port info kept in [`PortData`].
+//
+// We hope to observe the reported values differing in a way that can be
+// explained by data bits being flipped or shifted, which would be consistent
+// with "I2C weather" as opposed to a complete hiccup of the transceiver
+// itself.
+counted_ringbuf!(
+    TEMP_GLITCH_RINGBUF,
+    TempGlitch,
+    32,
+    TempGlitch::RINGBUF_INIT
+);
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -460,6 +522,15 @@ impl XcvrApi {
         // It has probably been milliseconds, we don't expect the temperature
         // to change significantly in this time.
         if diff >= MAX_RESAMPLE_VARIANCE_CELSIUS {
+            // Record glitches specifically in the ringbuf
+            ringbuf_entry!(
+                TEMP_GLITCH_RINGBUF,
+                TempGlitch {
+                    index: port.0,
+                    first: a,
+                    second: b,
+                }
+            );
             Err(TempReadError::UnexpectedVariance(diff))
         } else {
             Ok((a, diff))
--- a/drv/transceivers-server/src/main.rs
+++ b/drv/transceivers-server/src/main.rs
@@ -59,7 +59,7 @@ static XCVR_SERVER_IMPL: ClaimOnceCell<MaybeUninit<ServerImpl>> =
 include!(concat!(env!("OUT_DIR"), "/i2c_config.rs"));
 
 #[allow(dead_code)]
-#[derive(Copy, Clone, PartialEq, Eq, Count)]
+#[derive(Copy, Clone, PartialEq, Count)]
 enum Trace {
     #[count(skip)]
     None,
@@ -77,31 +77,47 @@ enum Trace {
     TransceiversError(#[count(children)] TransceiversError),
     GotInterface(u8, ManagementInterface),
     UnknownInterface(u8, ManagementInterface),
-    UnpluggedModule(usize),
-    RemovedDisabledModuleThermalModel(usize),
-    TemperatureReadError(usize, Reg::QSFP::PORT0_STATUS::ErrorEncoded),
-    TemperatureReadUnexpectedError(usize, FpgaError),
-    ThermalError(usize, ThermalError),
-    GetInterfaceError(usize, Reg::QSFP::PORT0_STATUS::ErrorEncoded),
-    GetInterfaceUnexpectedError(usize, FpgaError),
-    InvalidPortStatusError(usize, u8),
+    UnpluggedModule {
+        port: LogicalPort,
+    },
+    RemovedDisabledModuleThermalModel {
+        port: LogicalPort,
+    },
+    TemperatureReadError {
+        port: LogicalPort,
+        err: Reg::QSFP::PORT0_STATUS::ErrorEncoded,
+    },
+    TemperatureReadUnexpectedError {
+        port: LogicalPort,
+        err: FpgaError,
+    },
+    ThermalError {
+        port: LogicalPort,
+        err: ThermalError,
+    },
+    GetInterfaceError {
+        port: LogicalPort,
+        err: Reg::QSFP::PORT0_STATUS::ErrorEncoded,
+    },
+    GetInterfaceUnexpectedError {
+        port: LogicalPort,
+        err: FpgaError,
+    },
+    InvalidPortStatusError {
+        port: LogicalPort,
+        raw_err: u8,
+    },
     DisablingPorts(LogicalPortMask),
-    DisableFailed(usize, LogicalPortMask),
+    DisableFailed {
+        step: usize,
+        mask: LogicalPortMask,
+    },
     ClearDisabledPorts(LogicalPortMask),
     SeqError(SeqError),
-    TemperatureGlitch(usize, MilliCelsiusFixed),
-}
-
-/// Trace requires that types are Eq, and floats are not Eq. Make a little
-/// helper type that stores milli-celcius as a fixed point number
-#[derive(Copy, Clone, PartialEq, Eq)]
-struct MilliCelsiusFixed(u32);
-
-impl From<f32> for MilliCelsiusFixed {
-    fn from(value: f32) -> Self {
-        let value = value * 1000.0;
-        Self(value as u32)
-    }
+    TemperatureGlitch {
+        port: LogicalPort,
+        variance: Celsius,
+    },
 }
 
 counted_ringbuf!(Trace, 16, Trace::None);
@@ -481,7 +497,8 @@ impl ServerImpl {
         } = self;
 
         for (i, meta) in ports.iter_mut().enumerate() {
-            let port = LogicalPort(i as u8);
+            let port_idx = i as u8;
+            let port = LogicalPort(port_idx);
             let mask = 1 << i;
             let operational = (!status.modprsl & mask) != 0
                 && (status.power_good & mask) != 0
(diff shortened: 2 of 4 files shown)


drv/transceivers-server/src/main.rs
```diff
@@ -99,6 +120,19 @@ counted_ringbuf!(Trace, 16, Trace::None);
...
+/// that causes the sidecar's thermal protection loop to shut down immediately.
+///
+/// For this reason, we double-sample each transceiver, and if the two readings
+/// taken in short succession vary by more than this amount, we discard the
+/// reading entirely, trying again on the next tick.
+///
+/// This number is a "wild guess", but across two readings a few milliseconds
+/// apart, we expect very little difference (probably <1.0C), even factoring in
+/// potential sample noise and precision limitations.
+const MAX_RESAMPLE_VARIANCE_CELSIUS: f32 = 5.0f32;
```
jamesmunns, line 219: I have no idea if this is a good number. That being said, +/- 5C is probably not going to throw us into powerdown.

drv/transceivers-server/src/main.rs
```diff
@@ -111,11 +145,41 @@ enum FrontIOStatus {
...
 
+struct PortMetadata {
+    /// Number of consecutive NACKS seen on a given port
+    consecutive_errors: u8,
+    /// Largest observed temperature drift between two samples
+    peak_diff: f32,
+    /// Number of times the temperature has been discarded, saturates
+    discarded_temps: u32,
+    /// Thermal models are populated by the host
+    // TODO(AJM): Is the above comment true? We actually fill this in with a
```
jamesmunns: This was an old comment, and I think was more aspirational than actual. I probably should just rewrite it. Not sure if we have a tracking issue for "the host can set thermal models".

drv/transceivers-server/src/main.rs
```diff
@@ -431,66 +556,60 @@ impl ServerImpl {
...
+                        // Previously, this did a "continue", and didn't affect
+                        // consecutive errors at all. This should never happen
+                        // because we only add models to known interface types
+                        TempReadError::UnknownInterface => {}
+                        // This would *increment* consecutive errors
+                        TempReadError::ReallyBadTempRead(_) => {
+                            meta.consecutive_errors =
+                                meta.consecutive_errors.saturating_add(1);
+                        }
+                        // All of these *actually reset* the error count. Do
```
jamesmunns: CC @Aaron-Hartwig - this faithfully reproduces the old behavior, but I'm not sure if that's the behavior we actually want. There's one other place to look below to understand this as well.

drv/transceivers-server/src/main.rs
```diff
@@ -563,6 +682,99 @@ impl ServerImpl {
...
+
+/// Error while reading temperature from QSFP transceiver
+enum TempReadError {
+    /// Tried to read a transceiver that had no assigned thermal model
+    /// This is a programming error.
+    UnknownInterface,
+    /// We failed to read, and the FPGA reported an error code that makes us
+    /// suspicious that the xcvr is not present or operating correctly in a
+    /// way that might lead us to disable it.
+    ReallyBadTempRead(Reg::QSFP::PORT0_STATUS::ErrorEncoded),
```
jamesmunns: This needs a better name.

drv/transceivers-server/src/main.rs
```diff
@@ -563,6 +682,99 @@ impl ServerImpl {
...
+            return Self::BadPortStatus(code);
+        };
+
+        let is_read_err = match decoded {
+            // We consider these errors as potentially worth invalidating the
+            // QSFP over.
+            ErrorEncoded::I2CAddressNack => true,
+            ErrorEncoded::I2CSclStretchTimeout => true,
+
+            // TODO(AJM): why *don't* we consider these errors as worth
```
jamesmunns: CC @Aaron-Hartwig - this also needs a look at. This reproduces the old behavior, I'm not sure if that's what we want. If it is, could you help explain some context to improve the comments here?

lib/static-cell/src/lib.rs
```diff
@@ -99,7 +99,7 @@ pub struct ClaimOnceCell<T> {
 // Safety: because a `ClaimOnceCell` may only create a single mutable reference to
 // the inner value a single time, it can implement `Sync` freely, as the inner
 // `UnsafeCell`'s value cannot be mutably aliased.
-unsafe impl<T> Sync for ClaimOnceCell<T> where for<'a> &'a T: Send {}
+unsafe impl<T> Sync for ClaimOnceCell<T> where T: Send {}
```
jamesmunns, line 102: This is from #2667, should get rebased away after merging.

jamesmunns (commented):
Added some commentary.

drv/transceivers-server/src/main.rs
labbott: `UnrecoverableTempRead` ? (Or is unrecoverable already overloaded for thermal)

drv/transceivers-server/src/main.rs
labbott: I agree +/- 5C is at least a good first attempt and probably safe. I'm interested to see the data from letting this sit on racklette under "normal" conditions.

drv/transceivers-server/src/main.rs
```diff
@@ -81,6 +89,19 @@ enum Trace {
     DisableFailed(usize, LogicalPortMask),
     ClearDisabledPorts(LogicalPortMask),
     SeqError(SeqError),
+    TemperatureGlitch(usize, MilliCelsiusFixed),
+}
+
+/// Trace requires that types are Eq, and floats are not Eq. Make a little
+/// helper type that stores milli-celcius as a fixed point number
```
labbott: We have https://github.com/oxidecomputer/hubris/blob/master/sys/userlib-units/src/lib.rs which has `Celsius` which seems to be used already in `ringbuf` (see https://github.com/oxidecomputer/hubris/blob/4113fbc208a0e80c205c81d5f5bbe13a70ab214d/task/thermal/src/main.rs#L74-L88) Is there a reason that works for `Celcius` but not here?

jamesmunns: Oh, I just scrolled back in my terminal, and I misread the error:

```
   Compiling drv-transceivers-server v0.1.0 (/Users/james/oxide/hubris/drv/transceivers-server)
error[E0277]: the trait bound `Celsius: Eq` is not satisfied
   --> drv/transceivers-server/src/main.rs:85:30
    |
 55 | #[derive(Copy, Clone, PartialEq, Eq, Count)]
    |                                  -- in this derive macro expansion
...
 85 |     TemperatureGlitch(usize, Celsius, Celsius),
    |                              ^^^^^^^ the trait `Eq` is not implemented for `Celsius`
    |
   ::: /rustlib/cmp.rs:364:1
    |
364 | pub macro Eq($item:item) {
    | ------------ in this expansion of `#[derive(Eq)]`
    |
note: required by a bound in `core::cmp::AssertParamIsEq`
   --> /rustlib/cmp.rs:379:31
    |
379 | pub struct AssertParamIsEq<T: Eq + PointeeSized> {
    |                               ^^ required by this bound in `AssertParamIsEq`
```

We have `#[derive(Eq)]` on this `Trace`, which we probably don't need. I assumed it was something to do with Ringbuf/Counters. Good catch!

drv/transceivers-server/src/main.rs
```diff
@@ -81,6 +89,19 @@ enum Trace {
     DisableFailed(usize, LogicalPortMask),
     ClearDisabledPorts(LogicalPortMask),
     SeqError(SeqError),
+    TemperatureGlitch(usize, MilliCelsiusFixed),
```
hawkw: I was going to ask what the `usize` means here (I am guessing it's the port and not a timestamp?) and propose we made it a named field so its meaning would be obvious when looking at the ringbuf. But, I see that that there are a bunch of similarly shaped ringbuf entries here already. I'd kinda like to see them all become named fields, but 🤷‍♀️

drv/transceivers-server/src/main.rs
hawkw: > We have #[derive(Eq)] on this Trace, which we probably don't need. I assumed it was something to do with Ringbuf/Counters. Good catch!

I believe it's just `PartialEq` that ringbuf needs. I think we can get rid of the `Eq'.

drv/transceivers-server/src/main.rs
hawkw: 5C seems pretty reasonable to me, too. N.B. that in the thermal model for the transceivers, we are saying that we believe the "worst case" slew for them to be 0.5C/s: https://github.com/oxidecomputer/hubris/blob/755934d7d1ecf2b18440293710953a95f5e3565c/drv/transceivers-server/src/main.rs#L416

(I think that estimate is pretty conservative, but 🤷‍♀️)

drv/transceivers-server/src/main.rs
hawkw: Incidentally, it occurs to me --- and this is probably best saved for another PR --- it would be nice if we used an enum of port numbers rather than `usize` here, so that the ringbuf events could add `#[count(children)]` for the port. The `counters` crate doesn't support `#[count(children)]` on integer fields since well, we would have to generate a table of `usize::MAX` counters (or dynamically allocate them which kind of defeats the purpose!)

drv/transceivers-server/src/main.rs
```diff
@@ -111,11 +145,41 @@ enum FrontIOStatus {
     Ready,
 }
 
+struct PortMetadata {
```
hawkw: obnoxious and nitpicky: is this really _metadata_? it seems kinda like it's actually just _data_ about that port...

drv/transceivers-server/src/main.rs
hawkw: I, unfortunately, kind of _like_ `BadTempRead` and `ReallyBadTempRead`, but I agree that the important thing to communicate here is whether we distrust the transceiver as a result of that error or not. I think the use of `Unrecoverable` wouldn't be _too_ conflicty with the thermal loop, but, hmm. What about `TransientTempReadError`/`FatalTempReadError`?

hawkw: It looks like `ReallyBad` basically means an address NACK or a clock-stretching timeout; I guess the idea here is that those would conclusively indicate that the module is either not physically present or is totally incapable of talking to us over I2C ever?

drv/transceivers-server/src/main.rs
```diff
@@ -563,6 +682,99 @@ impl ServerImpl {
...
+    /// This is a programming error.
+    UnknownInterface,
+    /// We failed to read, and the FPGA reported an error code that makes us
+    /// suspicious that the xcvr is not present or operating correctly in a
+    /// way that might lead us to disable it.
+    ReallyBadTempRead(Reg::QSFP::PORT0_STATUS::ErrorEncoded),
+    /// We failed to read, but in some kind of forgivable way that *doesn't*
+    /// make us suspect that the xcvr should be disabled.
+    BadTempRead(Reg::QSFP::PORT0_STATUS::ErrorEncoded),
+    /// The FPGA gave us back a status codethat we were unable to decode. This
```
hawkw: typo:
```suggestion
    /// The FPGA gave us back a status code that we were unable to decode. This
```

drv/transceivers-server/src/main.rs
```diff
@@ -563,6 +682,99 @@ impl ServerImpl {
...
+    /// is probably a pretty serious mismatch in FPGA and SP versioning.
+    BadPortStatus(u8),
+    /// The FPGA gave us back an FpgaError that we didn't expect at all when
+    /// reading I2C.
+    UnexpectedFpgaErr(FpgaError),
+    /// We read the sensor multiple times, and the
```
hawkw: looks like this sentence wasn't finished?

drv/transceivers-server/src/main.rs
```diff
@@ -563,6 +682,99 @@ impl ServerImpl {
...
+
+impl TempReadError {
+    fn ringbuf(&self, idx: usize) {
+        let trace = match self {
+            // We don't ringbuf for this, should never happen
+            Self::UnknownInterface => return,
+
+            Self::ReallyBadTempRead(e) | Self::BadTempRead(e) => {
+                Trace::TemperatureReadError(idx, *e)
+            }
```
hawkw: Part of me wonders if these ought to be separate ringbuf entries, so that we can record separate counters of the two error classifications? Note that we are not able to record a counter for each possible value of `e` because we don't currently have a way to derive counters for stuff generated by FPGA codegen (see #2656). So, I think it _might_ be worthwhile for the counters to be able to distinguish between the number of bad and really bad errors.

drv/transceivers-server/src/main.rs
jamesmunns: I went with "PotenialRemoval", since this code path is counted when we think the qsfp was potentially removed (or totally inoperative)

drv/transceivers-server/src/main.rs
```diff
@@ -111,11 +145,41 @@ enum FrontIOStatus {
     Ready,
 }
 
+struct PortMetadata {
+    /// Number of consecutive NACKS seen on a given port
+    consecutive_errors: u8,
+    /// Largest observed temperature drift between two samples
+    peak_diff: f32,
+    /// Number of times the temperature has been discarded, saturates
+    discarded_temps: u32,
```
bnaecker, line 239: I might recommend recording either all attempted reads, or the number of successful reads. We can get a sense of the error rate using that.

drv/transceivers-server/src/main.rs
```diff
@@ -111,11 +145,41 @@ enum FrontIOStatus {
     Ready,
 }
 
+struct PortMetadata {
+    /// Number of consecutive NACKS seen on a given port
+    consecutive_errors: u8,
+    /// Largest observed temperature drift between two samples
+    peak_diff: f32,
```
bnaecker, line 237: Are any other statistical measures useful? Median or other percentiles would be nice but they're tricky to compute on streaming data. Possibly the mean and variance? I have no sense of the resources we have available to us on The Little CPU :)

drv/transceivers-server/src/main.rs
Aaron-Hartwig: I'm not aware of an issue for that. In general, we try to have a lot of the MPN-specific state and decision making in the host rather than the SP. The SP/FPGAs are primarily just conduits for host control (aside from thermal and power monitoring). I could imagine the comment author was thinking that the host would configure the model for a given MPN, similar to what we do with TX equalization settings on the Tofino.

drv/transceivers-server/src/main.rs
jamesmunns: We can probably keep a couple u32/f32s per transceiver. Buckets for percentiles would probably be harder. Mean would be challenging if we only look a few hours/days later.

I thought about keeping an exponential moving average and pitching samples that fall outside of that by too far (instead of double sampling).


I'm sort of guessing we'll end up seeing transient discarded samples that are power-of-two deviations if we are seeing i2c weather, rather than just wildly swinging values.

we could keep a separate ringbuf of just the disagreeing samples to see what the A/B difference is.

drv/transceivers-server/src/main.rs
Aaron-Hartwig: My comment here applies for your note higher up to. I think the errors which make us question "was the transceiver removed?" are things that we basically don't expect to ever see during "normal operation", but one could also envision cases where an `I2CByteNack` or an `I2CTransactionTimeout` could also be due to device removal. Rather than try to infer this from the I2C error register, could we just verify it by querying for port presence again on the port in question? Right before we update the thermal loop, we update device presence, so the window for a device to disappear _after_ that and _before_ we query temperature is extremely small but it of course could happen.

drv/transceivers-server/src/main.rs
bnaecker: Thanks @jamesmunns, I agree that most of that sounds more difficult than we need right now, given that we're mostly interested in speculative data collection.

> we could keep a separate ringbuf of just the disagreeing samples to see what the A/B difference is.

I like this one. If we have a small ringbuf of the "bad" samples, we could at least have some idea of its distribution. We already know it's a pretty unusual distribution, based on the wild swings (60C -> 120C -> 60C) and hugely negative values (-40C). So having a few raw data points is probably a good starting point.

jamesmunns: Added in d1e08661ff2aeb9f0c779275de195d72e02092e2

drv/transceivers-server/src/main.rs
jamesmunns: @hawkw I think I did this in 0c080db0f, since I wanted 90% of this for Ben's request as well.

drv/transceivers-server/src/main.rs
jamesmunns: Opened #2670

drv/transceivers-server/src/main.rs
jamesmunns: Got it, I'll leave it mostly as-is, but not decrease the error count if we see one of these. Like you said, if the device was really removed, we'll notice it on the next tick.

drv/front-io-api/src/transceivers.rs
```diff
@@ -134,6 +137,24 @@ impl LogicalPort {
         PortLocation::from(*self)
     }
 }
+
+/// Implement the ringbuf trait on LogicalPort to allow for per-port metrics
```
hawkw, line 141: counters trait?

drv/front-io-api/src/transceivers.rs
```diff
@@ -134,6 +137,24 @@ impl LogicalPort {
...
+        [const { AtomicU32::new(0) }; NUM_PORTS as usize];
+
+    fn count(&self, counters: &Self::Counters) {
+        // This should never happen, but just in case.
+        let Some(ctr) = counters.get(self.0 as usize) else {
+            return;
+        };
+        ctr.fetch_add(1, Ordering::Relaxed);
+    }
+}
```
hawkw, line 156: this is cute; I wonder if we might want the `counters` crate to have a macro or something for saying "yes, this looks like a `u32`, but it will always be in some range, so you can derive counters for it without having to worry"; elsewhere, I've tended to just use `repr(N)` enums for this sort of thing, like this thing:
https://github.com/oxidecomputer/hubris/blob/0d1ba0453a5d80470f07ea9949eb82bc3c291e01/drv/psc-seq-server/src/main.rs#L390-L403
but that then requires some weirdish boilerplate for converting between the enum and integers if you also want to index arrays or whatever.

very much not a blocker for this PR, but I wonder if we might throw together a little newtype-integer-counter derive or something that works like this.

drv/transceivers-server/src/main.rs
```diff
@@ -69,22 +77,120 @@ enum Trace {
...
+    type Counters = <LogicalPort as Count>::Counters;
+
+    #[allow(clippy::declare_interior_mutable_const)]
+    const NEW_COUNTERS: Self::Counters = <LogicalPort as Count>::NEW_COUNTERS;
+
+    #[inline]
+    fn count(&self, counters: &Self::Counters) {
+        self.port.count(counters);
+    }
+}
```
hawkw, line 175: Oh, huh, I think this is necessary because the `counters` crate doesn't support deriving `Count` for a struct, even if it has `count(Children)` on a field that implements `Count`? It could be worth adding that to better support cases like this (not in this PR, I'll go make a ticket).

drv/transceivers-server/src/main.rs
```diff
@@ -69,22 +77,120 @@ enum Trace {
...
+
+impl TempGlitch {
+    /// Okay so ringbuf requires an initializer, and this value is actually
+    /// somewhat obnoxious because it *is* a plausible value to observe.
+    /// However, "all zeroes" lets us put the initializer in .bss and not .data,
+    /// and since this is just for debugging, we hope that a reasonable person
+    /// would realize that the delta between the two entries is zero. I really
+    /// didn't feel like making this an Option, or using u8::MAX as the index,
+    /// or even adding 1 to the port and then using NonZero or something.
```
hawkw, line 153: Yeah, I agree that this is pretty reasonable; the offset one where we use `NonZero`, in particular, actually sounds like kind of a bad idea, because the primary consumer of ringbufs, humility, does not have any idea that we have done such an offsetting and will end up just lying to the reader about which port it is, unless they went and read the code to learn about this.

With that said, I'm not actually sure if making it an `Option` and initializing the ringbuf to `None` is actually all that bad, since we don't actually really work with the entries all that meaningfully? It might make things a bit less confusing for a reader...

drv/transceivers-server/src/main.rs
```diff
@@ -69,22 +77,120 @@ enum Trace {
...
 }
 
 counted_ringbuf!(Trace, 16, Trace::None);
 
+#[derive(PartialEq, Debug, Clone, Copy)]
+struct TempGlitch {
+    port: LogicalPort,
+    first: Celsius,
+    second: Celsius,
+}
```
hawkw, line 144: i wonder if we might also want to include a timestamp in these, so that a reader can get some sense of how frequently they occur/whether they are clustered temporally or not?

drv/transceivers-server/src/main.rs
```diff
@@ -351,77 +518,144 @@ impl ServerImpl {
...
+            ManagementInterface::Unknown(..) => {
+                // We should never get here, because we only assign
+                // `self.thermal_models[i]` if the management interface is
+                // known.
+                return Err(TempReadError::UnknownInterface);
+            }
+        }?;
+
+        Ok(res)
+    }
```
hawkw, line 567: i think the layering of `get_temperature_resample` -> `get_temperature_once` is conceptually reasonable, but it is a bit silly that we end up matching on `m.interface` twice every time we read temps from a thing. the performance hit from that is probably not actually important but it feels silly. probably not worth messing with.

drv/transceivers-server/src/main.rs
```diff
@@ -351,77 +518,144 @@ impl ServerImpl {
...
-                self.thermal_models[i] = None;
+                meta.model = None;
             }
         }
 
         // Accumulate ports to disable (but don't disable them in the loop), to
         // avoid issues with the borrow checker.
         let mut to_disable = LogicalPortMask(0);
-        for (i, m) in self.thermal_models.iter().enumerate() {
+        for (i, meta) in ports.iter_mut().enumerate() {
```
hawkw, line 655: bit odd that we renamed this to `PortData` but the variable is still `meta`...

jamesmunns:
Noting that this is still probably a thing we want to land, it's just lower prio than #2678 right now.

drv/transceivers-server/src/main.rs
jamesmunns: I think I'm going to punt on this, because `humility ringbuf` is already smart enough to not show counter entries that haven't actually been logged, so this "valid but not actually" state never gets shown to the user. e.g. when the ringbuf is empty, the user sees:

```
humility: ring buffer drv_transceivers_server::TEMP_GLITCH_RINGBUF in transceivers:
humility: ring buffer drv_transceivers_server::__RINGBUF in transceivers:
```

drv/transceivers-server/src/main.rs
jamesmunns: Refactored.

drv/transceivers-server/src/main.rs
jamesmunns: de-meta'd.

drv/transceivers-server/src/main.rs
jamesmunns: Yeah, I considered this, but originally avoided it because it doubles the size of the ringbuf entries (12B -> 24B, because `u64` has `align(8)`), but the data is valuable and worth the extra 384B of RAM usage.

Added.

jamesmunns: Technically we could probably slap `#[repr(packed(4))]` onto the struct, we never reference the timestamp field (only store), though I'm not sure this would play nice with `humility`, and it's probably not worth it to golf out 192B of static usage. I do wonder how much of this there is across our entire codebase.

Sadly, this is a requirement of the [AAPCS32 ABI](https://github.com/ARM-software/abi-aa/blob/main/aapcs32/aapcs32.rst), as `u64` is "naturally aligned" to its size, which is 8 bytes.

jamesmunns:
CC @hawkw @labbott, I have addressed all open comments, and this has run successfully overnight on `berlin`, I'd like to get this merged so it can be included in the next release. If I could get a review, I'd appreciate it!

drv/transceivers-server/src/main.rs
hawkw: Yeah, it still _feels_ a little bad, but I can live with it!

drv/transceivers-server/src/main.rs
```diff
@@ -69,22 +77,129 @@ enum Trace {
...
+
+    #[inline]
+    fn count(&self, counters: &Self::Counters) {
+        self.port.count(counters);
+    }
+}
+
+// Keep counters each time the delta between two samples exceeds the value
+// of `MAX_RESAMPLE_VARIANCE_CELSIUS`. This also counts how often it happens
+// on a per-port basis, as well as the last 32 instances it has been observed.
```
hawkw: Docs nit: this wording is a bit weird: "Keep counters ... This also counts ... as well as the last 32 instances". It seems like we are referring to both the ringbuf entries and the generated counters as "counters" by mistake, which seems a bit confusing. How about something like:
```suggestion
// Keep a record of the last 32 times the delta between two samples exceeded
// the value of `MAX_RESAMPLE_VARIANCE_CELSIUS`. This also counts the total
// number of times it has happened on a per-port basis.
```

drv/transceivers-server/src/main.rs
```diff
@@ -111,11 +243,46 @@ enum FrontIOStatus {
     Ready,
 }
 
+struct PortData {
+    /// Number of consecutive NACKS seen on a given port
+    consecutive_errors: u8,
+    /// Largest observed temperature drift between two samples
+    peak_diff: f32,
```
hawkw: turbo nit: since this is being read by Humility, where the reader won't see the comment, perhaps this ought to be:
```suggestion
    peak_temp_diff: f32,
```

drv/transceivers-server/src/main.rs
```diff
@@ -111,11 +243,46 @@ enum FrontIOStatus {
...
+            consecutive_errors: 0,
+            peak_diff: 0.0,
+            discarded_temps: 0,
+            total_temp_samples: 0,
+            model: None,
+        }
+    }
+
+    /// Like [`Self::new()`] but with the given model.
+    fn init(&mut self, model: Option<ThermalModel>) {
```
hawkw, line 276: unimportant nit, feel free to ignore: hm, this isn't _really_ "like `Self::new()`", because it initializes `self` in place...

drv/transceivers-server/src/main.rs
```diff
@@ -351,77 +531,139 @@ impl ServerImpl {
...
+                // `self.thermal_models[i]` if the management interface is
+                // known.
+                return Err(TempReadError::UnknownInterface);
+            }
+        };
+
+        // Attempt to get the temperature twice to determine if we get a stable
+        // temperature. If either attempt fails, just return the error.
+        let a = func(self, port)?;
+        let b = func(self, port)?;
```
hawkw: nit, feel free to ignore, but i might call this something like:
```suggestion
        let read_temp = match m.interface {
            ManagementInterface::Cmis => Self::read_cmis_temperature,
            ManagementInterface::Sff8636 => Self::read_sff8636_temperature,
            ManagementInterface::Unknown(..) => {
                // We should never get here, because we only assign
                // `self.thermal_models[i]` if the management interface is
                // known.
                return Err(TempReadError::UnknownInterface);
            }
        };

        // Attempt to get the temperature twice to determine if we get a stable
        // temperature. If either attempt fails, just return the error.
        let a = read_temp(self, port)?;
        let b = read_temp(self, port)?;
```

drv/transceivers-server/src/main.rs
```diff
@@ -563,6 +795,106 @@ impl ServerImpl {
...
         self.update_thermal_loop(status);
     }
+
+    pub(crate) fn transceivers(&mut self) -> &mut Transceivers {
+        &mut self.xcvr_api.transceivers
+    }
+}
+
+/// Error while reading temperature from QSFP transceiver
+enum TempReadError {
```
hawkw, line 805: i feel like some of these, in the fullness of time, should probably become ereports. not necessarily in _this_ PR, though.

hawkw (approved):
This looks good to me overall! I had some smallish docs and naming nits.

PR #2682, opened 2026-09-11
Switch to software hashing for async host flash hashing (and some cleanup)

labbott:


drv/gimlet-hf-server/src/main.rs
```diff
@@ -485,13 +492,14 @@ impl ServerImpl {
 
                 if addr + step_size >= end {
                     self.hash.state = HashState::Done;
-                    match self.hash.task.finalize_sha256() {
-                        Ok(v) => match dev {
+                    let mut out: [u8; 32] = [0; 32];
```
jamesmunns, line 495: Is this `drv_hash_api::SHA256_SZ`?

drv/gimlet-hf-server/src/main.rs
```diff
@@ -771,11 +779,14 @@ impl idl::InOrderHostFlashImpl for ServerImpl {
...
         }
+
+        let mut out: [u8; SHA256_SZ] = [0; SHA256_SZ];
+
         self.hash_range_update(begin, end)?;
-        self.hash
-            .task
-            .finalize_sha256()
-            .map_err(|_| HfError::HashError.into())
+        if self.hash.task.finalize_sha256(&mut out).is_err() {
```
jamesmunns, line 786: Do we want to log this error somehow? Right now we just discard it.

drv/stm32h7-hash/src/lib.rs
```diff
@@ -186,6 +187,47 @@ impl Hash {
         self.count += valid_bytes;
     }
 
+    /// Update hash with a full block of data. Requring exactly one block avoids
```
jamesmunns:
```suggestion
    /// Update hash with a full block of data. Requiring exactly one block avoids
```

drv/stm32h7-hash/src/lib.rs
```diff
@@ -186,6 +187,47 @@ impl Hash {
...
+            State::Initialized => {
+                self.state = State::Processing;
+            }
+            State::Processing => {}
+            _ => {
+                return Err(HashError::InvalidState);
+            }
+        };
+
+        if self.nvalid > 0 {
```
jamesmunns, line 206: Any worries that we could end up changing the state, then these checks cause us to remain in that state even though our current state or the input was invalid?

drv/stm32h7-hash/src/lib.rs
```diff
@@ -186,6 +187,47 @@ impl Hash {
...
+        }
+
+        for block in data.chunks(BLOCK_LEN_BYTES) {
+            while self.is_busy() {
+                // If the block is busy a write to `DATAIN` will stall the
+                // AHB bus. Past experience has shown that context switching
+                // away is too costly so just busy wait.
+            }
+            for w in block.chunks(SIZEOF_U32) {
+                let data = u32::from_le_bytes(w.try_into().unwrap());
```
jamesmunns, line 221: Would it be worth forcing the alignment of this buffer so we don't have to stack-copy through? Not sure if it matters much if we end up spinning on `self.is_busy()` anyway.

drv/gimlet-hf-server/src/main.rs
labbott: This error is not actually likely to be useful in practice. The only way we would get an error is if we forgot to call `init` or called finalize twice in a row.

drv/stm32h7-hash/src/lib.rs
labbott: I'm not following this comment. Can you clarify? You might be asking about interactions between `update` and `update_exact` and how they might interact?

drv/stm32h7-hash/src/lib.rs
labbott: It could be nicer! I've been debating the value of accepting arbitrary `[u8]` vs requiring `[u32]` everywhere in this driver. We would be limiting what we could accept and wouldn't work on some NIST test vectors. But the entire point of this PR is that it turns out we don't need a generic hashing mechanism in hubris: software hashing is Good Enough(tm) for most cases except for the 32 MB of QSPI.

drv/stm32h7-hash/src/lib.rs
```diff
@@ -243,15 +285,6 @@ impl Hash {
...
         }
-        while offset + SIZEOF_U32 <= data.len() {
-            self.write_word(
-                u32::from_le_bytes(
-                    (&data[offset..offset + SIZEOF_U32]).try_into().unwrap(),
-                ),
-                SIZEOF_U32,
-            );
-            offset += SIZEOF_U32;
-        }
```
labbott, line 254: For the record this was some dead code I saw floating around (increase the context to see this can never be called)

drv/gimlet-hf-server/src/main.rs
labbott: Yes. I was initially trying to see if I could move away from that since we don't have the separate task anymore but the constant is still handy. I'll fix it up.

drv/stm32h7-hash/src/lib.rs
```diff
@@ -186,6 +187,47 @@ impl Hash {
...
+            }
+        };
+
+        if self.nvalid > 0 {
+            return Err(HashError::InvalidState);
+        }
+
+        if !data.len().is_multiple_of(BLOCK_LEN_BYTES) {
+            return Err(HashError::InvalidState);
+        }
```
hawkw, line 212: this feels more like an invalid _input_ than an invalid _state_; i wonder if a different error would be more appropriate here?

drv/stm32h7-hash/src/lib.rs
```diff
@@ -186,6 +187,47 @@ impl Hash {
...
+
+        for block in data.chunks(BLOCK_LEN_BYTES) {
+            while self.is_busy() {
+                // If the block is busy a write to `DATAIN` will stall the
+                // AHB bus. Past experience has shown that context switching
+                // away is too costly so just busy wait.
```
hawkw, line 218: nit, sorry --- just passing through with the [comma shaker](https://www.newyorker.com/culture/culture-desk/in-defense-of-nutty-commas): 
```suggestion
                // If the block is busy, a write to `DATAIN` will stall the
                // AHB bus. Past experience has shown that context switching
                // away is too costly, so just busy wait.
```

drv/stm32h7-hash/src/lib.rs
labbott: I was very lazy about adding errors and you are correct

app/gimlet/base.toml
```diff
@@ -214,10 +202,10 @@ priority = 3
...
 stacksize = 3000
 start = true
-uses = ["quadspi"]
-interrupts = {"quadspi.irq" = "qspi-irq"}
-task-slots = ["sys", "hash_driver"]
-notifications = ["qspi-irq", "timer"]
+uses = ["quadspi", "hash"]
+interrupts = {"quadspi.irq" = "qspi-irq", "hash.irq" = "hash-irq"}
+task-slots = ["sys"]
+notifications = ["qspi-irq", "timer", "hash-irq"]
```
labbott, line 208: I guess I could move some of this out of here but it gets used in the follow on work

drv/gimlet-hf-server/src/main.rs
```diff
@@ -41,14 +41,24 @@ use stm32h7::stm32h743 as device;
...
-task_slot!(HASH, hash_driver);
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum Trace {
+    None,
+
+    AsyncTime(u64),
+}
+
+ringbuf::ringbuf!(Trace, 8, Trace::None);
```
labbott, line 61: I threw this in because it was nice for profiling/quick checks. I can also remove it (but might be nice to keep in when measuring)

hawkw: fine with me if you think this will be useful in the future! i suppose we could probably scrape off a few bytes if we got rid of the enum here and zero-initialized it, like:
```suggestion
ringbuf::ringbuf!(u64, 8, 0);
```

but that might be more confusing to a future reader (who would just see all zeros when we haven't logged stuff), and perhaps we will want to record things other than timings eventually? so maybe leave it as is...

PR #2683, opened 2026-09-11
user_leds: rework

jamesmunns:
~~I mostly want to see if this passes CI.~~

I think this is a good change, actually!


drv/user-leds/src/bsp/stm32xx_sys.rs
```diff
@@ -0,0 +1,207 @@
...
+            info: pinset,
+            active_low: true,
+        }
+    }
+
+    #[allow(dead_code)]
+    const fn act_hi(pinset: PinSet) -> PinInfo<PinSet> {
+        PinInfo {
+            info: pinset,
+            active_low: true,
```
jamesmunns: This is wrong

drv/user-leds/src/bsp/stm32xx_sys.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let PinInfo {
+            info,
+            active_low: _,
+        } = board::led_info(led);
+        sys.gpio_toggle(info.port, info.pin_mask).unwrap_lite();
+    }
+}
+
+pub mod board {
```
jamesmunns: I should double check all the mappings are the same before/after

drv/user-leds/src/bsp/stm32xx_sys.rs
```diff
@@ -0,0 +1,211 @@
...
+            sys.gpio_configure_output(
+                *info,
+                OutputType::PushPull,
+                Speed::High,
+                Pull::None,
+            );
+        }
+    }
+
+    fn led_on(led: Self::Led) {
```
jamesmunns: TODO: I can probably change the BSP to take `&mut`, and store the task ID.

This saves us a syscall on *every* GPIO op, though potentially if `sys` restarts it means that we will too. This is probably fine?

drv/user-leds/src/bsp/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+
+// The following are common LED types that are used for various boards.
+//
+// The level of variance here is somewhat unfortunate, but seems to
+// work in practice.
+
+#[derive(enum_map::Enum, Copy, Clone, FromPrimitive)]
+#[allow(clippy::enum_variant_names)]
+pub enum Led4Color {
+    // chassis LED is controlled by cosmo-seq
```
labbott: This comment feels out of place

drv/user-leds/src/bsp/stm32fx.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        let (code, _) = userlib::sys_send(
+            rcc_driver,
+            ENABLE_CLOCK,
+            gpio_pnum.as_bytes(),
+            &mut [],
+            &[],
+        );
+        assert_eq!(code, 0);
+
```
labbott, line 55: can we do a quick spot check that this (and other `sys_send`) is still the correct calling convention? I think it should be but it's possible this has bit rotted since its not using idol and if we're going to have this around it would be nice if it could work.

labbott (commented):
Going to take another pass later but I think this is a big improvement. Thanks for putting the time in here.

drv/user-leds/src/bsp/mod.rs
jamesmunns: This does get removed in #2684, fwiw, if we merge that first this will go away.

drv/user-leds/src/bsp/stm32fx.rs
jamesmunns: Spot: checked!

labbott (approved):
Did another pass that the pins all match. 

I _think_ we now get an explicit compile error if we fail to select an stm32h7 board (`cosmo-a`, `gimlet-f` etc) instead of needing to call `compile_error` which is also nicer

PR #2684, opened 2026-09-14
user-leds: Fix cosmo's LED mapping

jamesmunns:
The original tapeout rev1 of cosmo contained an active-low RGB LED. However, in all subsequent revisions, it contained a single mosfet-switched, active-high, green-yellow LED.

This change removes the "ghost" LED that we've basically never used on cosmo.

This is currently based on #2683.


labbott (approved):
👻

PR #2686, opened 2026-09-14
Ndh/fmc vibe check

nathanaelhuffman:


app/cosmo/src/main.rs
```diff
@@ -398,7 +398,7 @@ fn system_init() {
     //
     // Note from the clock config earlier in this function that AHB3 is running
     // at 200 MHz.
-    const CLKDIV: u8 = 3; // /4, for 50 MHz -- field is divisor minus 1
+    const CLKDIV: u8 = 1; // /2, for 100 MHz -- field is divisor minus 1
```
jamesmunns, line 401: Can we revert this to land the change sooner?

app/grapefruit/src/main.rs
```diff
@@ -412,7 +412,7 @@ fn system_init() {
     //
     // Note from the clock config earlier in this function that AHB3 is running
     // at 200 MHz.
-    const CLKDIV: u8 = 3; // /4, for 50 MHz -- field is divisor minus 1
+    const CLKDIV: u8 = 2; // /3, for 66.67 MHz -- field is divisor minus 1
```
jamesmunns, line 415: Same here?

drv/stm32h7-fmc-demo-server/src/main.rs
```diff
@@ -404,6 +434,48 @@ fn process_network_packet(
...
+                // PeekBlockChecksum / PeekBlockChecksumFixed: read `count`
+                // 32-bit words and reply with only a wrapping-sum checksum.
+                // The constant-size reply means the request's duration is
+                // dominated by the FMC accesses, which makes bus line rate
+                // measurable from the host; per-word peek ops can't do that
+                // because their network cost swamps the bus time. Op 17
+                // advances through memory, op 18 re-reads one address.
+                let count = u16::from_le_bytes(read_chunk(&mut packet)?);
+                let mut sum: u32 = 0;
+                if byte == 17 {
```
jamesmunns, line 447: If you wanted to get clever to dedupe here a bit:

```rust
let offset = if byte == 17 { 4 } else { 0 };
let count = u16::from_le_bytes(read_chunk(&mut packet)?);
let mut sum: u32 = 0;
for _ in 0..count {
    let b = unsafe {
        core::ptr::read_volatile(address as *const u32)
    };
    sum = sum.wrapping_add(b);
    address += offset;
}
write_chunk(sum.to_le_bytes(), &mut response)?;
```

jamesmunns (changes requested):
Added some notes to hopefully get this merged before https://github.com/oxidecomputer/quartz/pull/527 lands.

PR #2687, opened 2026-09-14
metro: partially restore hiffy, update-server, and dump-agent

jamesmunns:
Partially re-enable these important tasks, with some functionality still notched out.


app/metro/base.toml
```diff
@@ -7,7 +7,7 @@ stacksize = 896
 name = "metro"
 requires = {flash = 32768, ram = 8192}
 features = [
-    # "dump",
+    "dump",
     # "measurement-handoff",
```
labbott, line 11: I think we'll probably want to keep `measurement-handoff` off for a while just to improve dev performance but turn it on after bring up is done. What's the best way to track that? Edit the checklist?

Another todo: double check that `humility flash` works as expected without `measurement-handoff` (it should because I can still flash e.g. gimletlet and the RoT but let's confirm the SP too)

app/metro/base.toml
```diff
@@ -193,15 +193,36 @@ notifications = ["exti-wildcard-irq"]
...
+features = [
+    "h753",
+    "stm32h7",
+    "gpio",
+    "spi",
+    "turbo",
+
+    # TODO: Re-enable these once the tasks are available!
+    # "i2c",
+    # "qspi",
```
labbott, line 207: We're definitely going to want to get QSPI added quickly because that's needed for actually writing the host flash image.

labbott (approved):

app/metro/base.toml
jamesmunns: This was dependent on the `hf` task, which I didn't bring in immediately, but I could do so in the next wave!

app/metro/base.toml
jamesmunns: @labbott yes, please feel free to edit the checklist in the tracking issue!

PR #2691, opened 2026-09-15
metro: Re-enable `net` and basic `net` accessories

jamesmunns:
Re-enable the `net` task and things that depend on it.

One note is the fundamentally this won't *actually* work yet: the `net` BSP waits for jefe to tell us our power state, which is provided by the `cpu_seq` task (e.g. `metro-seq-server`), which doesn't exist yet, and may not be enabled until later in the process.

I chose to share the BSP between the cosmo + metro boards, open to thoughts on that or whether we'd prefer a copy-paste approach.


task/net/src/main.rs
```diff
@@ -55,8 +55,12 @@ mod server;
...
 )]
 #[cfg_attr(
-    any(target_board = "cosmo-a", target_board = "cosmo-b",),
-    path = "bsp/cosmo_ab.rs"
+    any(
+        target_board = "cosmo-a",
+        target_board = "cosmo-b",
+        target_board = "metro-a",
+    ),
+    path = "bsp/cosmo_ab_metro_a.rs"
```
labbott, line 63: I both appreciate the very clear name and think it's going to get unweildy

labbott (approved):

task/net/src/main.rs
jamesmunns: I am open to bikeshed paint swatches, if you have any :D

PR #2694, opened 2026-09-17
control-plane-agent: Reevaluate this, since it's now the future

hawkw:
In the past, we wrote down that we may want to reevaluate this. Now, it is the future, and we wanted to reevaluate it. So I have reevaluated it, and my evaluation was: we probably shouldn't do that.


task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -665,19 +666,7 @@ impl SpHandler for MgsHandler {
                                 LifecycleState::Development
                                 | LifecycleState::Unprogrammed
                                 | LifecycleState::EndOfLife,
-                            )
-                            | Err(_) => {
-                                // Right now, we fail open if we can't talk to
```
jamesmunns, line 670: We might want to retain some comment why Development/Unprogrammed/EndOfLife RoT states justify offering a `Trivial` `UnlockChallenge`.

task/control-plane-agent/src/mgs_sidecar.rs
```diff
@@ -687,6 +676,33 @@ impl SpHandler for MgsHandler {
...
+                            // Performing the Release ECDSA challenge does not
+                            // actually require talking to the RoT beyond this
+                            // code for determining the lifecycle state, so the
+                            // Release challenge we perform in this case will
+                            // still allow the techport to be unlocked. This
+                            // means that if the RoT failure is persistent, the
+                            // system can still be debugged, if a valid
+                            // credential is provided.
+                            Err(error) => {
+                                ringbuf_entry!(Trace::RotLifecycleReadFailed(
```
jamesmunns, line 720: Do we want to add an ereport for this now (probably once per boot-cycle, or at least edge-triggered)? It might be good to surface this upstack for future alerting.

hawkw: I was thinking about doing that in a follow-up, but I'm not sure if we really need to or not. The likeliest cause is still that the RoT is just transiently in reset because we are updating its firmware or something...but perhaps it's worth saying something about.

task/control-plane-agent/src/mgs_sidecar.rs
hawkw: we didn't really have that previously; i feel like the comment that used to be here was really just discussing the error match. but, I think you're right that we _should_ have a comment about that, so I'll add something.

task/control-plane-agent/src/mgs_sidecar.rs
hawkw: added a TODO in https://github.com/oxidecomputer/hubris/pull/2694/commits/7d6d0dee6f5f5c66964597d88bbe7384323897ab

task/control-plane-agent/src/mgs_sidecar.rs
hawkw: https://github.com/oxidecomputer/hubris/pull/2694/commits/7d6d0dee6f5f5c66964597d88bbe7384323897ab adds a bunch of comments; possibly a bit excessively long-winded. WDYT?

labbott (approved):
LGTM thank you for following up here

task/control-plane-agent/src/mgs_sidecar.rs
hawkw: https://github.com/oxidecomputer/hubris/issues/2698

jamesmunns (approved):

Issue #2689, opened 2026-09-15
packrat: Consider reworking "cfg sprawl"


jamesmunns:
The `packrat` task has a lot of `#[cfg]`s (64 instances of `#[cfg` at the time of this writing). I'm about to add a bunch more for `metro` support.

It might be worth giving it a treatment similar to #2683 to make the similarities and differences between boards more clear.

Issue #2678, opened 2026-09-10
metro: Bringup tracking issue


jamesmunns:
In #2676, we added/are adding a totally-commented-out definition for metro, based on cosmo.

However, we need everything to *actually* work! Here's a tracking list of things that still need to be reviewed and enabled. We can make PRs for chunks of these features at a time. This is currently structured in roughly the shape of the toml definitions.

I've marked a couple "high priority" tasks as 🚨. A couple of "will definitely be different" items are marked as ⚠️.

## `base.toml`

- [ ] `[mmio]` ⚠️ - FPGA map may differ
- [ ] `[caboose]`
- [ ] `[tasks]`
    - [x] `[tasks.jefe]`
        - [ ] `[tasks.jefe.config.on-state-change]`
        - [ ] `[tasks.jefe.config.on-task-fault]`
        - [ ] `[tasks.jefe.config.allowed-callers]`
    - [ ] `[tasks.net]`
    - [ ] `[tasks.sys]` 🚨 - "I'd vote to review/enable this first since this should mostly be STM32H7 specific things"
        - [ ] `[tasks.sys.interrupts]`
        - [ ] `[tasks.sys.config.gpio-irqs.rot_irq]`
        - [ ] `[tasks.sys.config.gpio-irqs.seq_irq]`
    - [ ] `[tasks.spi2_driver]`
    - [ ] `[tasks.spi3_driver]`
    - [ ] `[tasks.i2c_driver]`
        - [ ] `[tasks.i2c_driver.interrupts]`
    - [ ] `[tasks.packrat]`
    - [ ] `[tasks.rng_driver]`
    - [ ] `[tasks.thermal]`
    - [ ] `[tasks.power]`
    - [ ] `[tasks.hiffy]` 🚨 - we want debugging ASAP
    - [ ] `[tasks.metro_seq]` ⚠️ - A general comment is that metro will have a new sequencer FPGA image that is not common with cosmos, though for most purposes they'll look the same barring some new features around programming the Versal etc.
    - [ ] `[tasks.ignition_flash]`
    - [ ] `[tasks.spartan7_loader]` ⚠️ - A general comment is that metro will have a new sequencer FPGA image that is not common with cosmos, though for most purposes they'll look the same barring some new features around programming the Versal etc.
        - [ ] `[tasks.spartan7_loader.config]`
    - [ ] `[tasks.hash_driver]`
    - [ ] `[tasks.hf]`
[...]

PR #2695, opened 2026-09-17
Re-enable SPI, QSPI, Auxflash, RoT, VPD, and Sensor tasks

jamesmunns:
This *mostly* unblocks adding a `metro_seq` task, though `spartan7_loader`, and having a `metro-seq/*.json` and real auxflash blobs are now probably the main blockers for that.


app/cosmo/base.toml
```diff
@@ -174,7 +174,7 @@ priority = 7
 max-sizes = {flash = 32768, ram = 65536 }
 stacksize = 1328
 start = true
-task-slots = ["sys", "hf", "i2c_driver", "hash_driver", "update_server", "sprot", "net"]
```
jamesmunns, line 177: This is just something #2688 missed (also removed from metro below)

app/metro/base.toml
```diff
@@ -1589,18 +1587,22 @@ tx = { packets = 3, bytes = 1024 }
...
-# [[auxflash.blobs]]
-# file = "drv/metro-seq-server/metro-hp/metro_hp.bz2"
-# unzip = "bz2"
-# compress = true
-# tag = "ICE4"
+[config.auxflash]
+memory-size = 33_554_432 # 256 Mib / 32 MiB
+slot-count = 16 # 2 MiB slots
+
+# TODO! These are cosmo blobs for now, as we can't uncomment the auxflash-server
```
jamesmunns, line 1594: This is somewhat footgun-shaped, but I'll add a note about it to #2678 so we don't forget.

labbott: I was going to copy in placeholders with a different name in another PR but I think this is fine for now

drv/stm32h7-sprot-server/src/main.rs
```diff
@@ -115,25 +115,26 @@ use gpio_irq_pins::ROT_IRQ;
...
+        target_board = "observer-a",
+        target_board = "gemini-bu-1",
+        target_board = "grapefruit-a",
+        target_board = "grapefruit-b",
+        target_board = "minibar-a",
+        target_board = "minibar-b",
+        target_board = "cosmo-a",
+        target_board = "cosmo-b",
+        target_board = "metro-a",
+    ))] {
```
labbott, line 137: This is getting a bit unwieldy and we should probably think of something more clever later

labbott (approved):

drv/stm32h7-sprot-server/src/main.rs
jamesmunns: I'll tag in https://github.com/oxidecomputer/hubris/issues/2689, maybe we make that a general "cfg sprawl" issue.

I wonder if we add a "standard_(sp)rot" key to the cfg that says "yes we used the same pinout again" that exports something like `sprot_config = "standard"`.

PR #2696, opened 2026-09-17
Bring in Metro Sequencer and Host Flash

labbott:


app/cosmo/base.toml
```diff
@@ -178,7 +178,7 @@ task-slots = ["sys", "hf", "i2c_driver", "hash_driver", "update_server", "net"]
 notifications = ["timer"]
 
 [tasks.cosmo_seq]
-name = "drv-cosmo-seq-server"
```
labbott, line 181: Experimenting here with the name change. Opinions welcome!

app/metro/base.toml
```diff
@@ -252,43 +249,43 @@ notifications = ["timer"]
...
+max-sizes = {flash = 16384, ram=4096 }
+stacksize = 2048
+start = true
+uses = ["hash"]
+interrupts = {"hash.irq" = "hash-irq"}
+task-slots = ["sys"]
+notifications = ["hash-irq"]
+
+[tasks.hf]
+name = "drv-cosmo-hf"
```
labbott: This did _not_ get a name change because honestly I was feeling a bit meh. But it's also useful for review against the sequencer so we can decide which approach si better

drv/cosmo-metro-seq-server/metro-hp/metro_hp.bz2
nathanaelhuffman, line 1: If it's easier, I don't currently see a reason why the cosmo and metro front FPGAs would need to diverge so will likely remain common.  We could rename them to cosmo-metro-hp or something if helpful. We can also support a new name probably but for the foreseeable future these will be the same.

nathanaelhuffman (approved):
Only real comment is I'm not sure we have reason to diverge the front fpgas on these two boards, so doing this means we'll have 2 copies of the same thing in the repo which we'll have to remember, but isn't a show stopper either way.  I'm happy to rename my front fpga to cosmo-metro-hp if that is helpful.

drv/cosmo-metro-seq-server/metro-hp/metro_hp.bz2
labbott: If you think they are likely to be the same I'll keep it in common for now and if it diverges we can fix it (it was a useful exercise to see that it's pretty easy to make that happen)

app/metro/base.toml
labbott: I decided to do a rename

drv/cosmo-metro-seq-server/build.rs
```diff
@@ -10,12 +10,12 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
 
     let out_dir = build_util::out_dir();
-    let out_file = out_dir.join("cosmo_fpga.rs");
+    let out_file = out_dir.join("cosmo_metro_fpga.rs");
     let mut file = fs::File::create(out_file)?;
 
     // Check that a valid bitstream is available for this board.
     let board = build_util::env_var("HUBRIS_BOARD")?;
-    if board != "cosmo-a" && board != "cosmo-b" {
+    if board != "cosmo-a" && board != "cosmo-b" && board != "metro-a" {
```
jamesmunns: Could do something like this, in case we end up having more revs of either board in the future:

```suggestion
    let supported_boards = &[
        "cosmo-a",
        "cosmo-b",
        "metro-a",
    ];
    if !supported_boards.contains(&board) {
```

drv/spartan7-loader/build.rs
```diff
@@ -14,7 +14,8 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     // Check that a valid bitstream is available for this board.
     let board = build_util::target_board().expect("could not get target board");
     match board.as_str() {
-        "grapefruit-a" | "grapefruit-b" | "cosmo-a" | "cosmo-b" => (),
+        "grapefruit-a" | "grapefruit-b" | "cosmo-a" | "cosmo-b" | "metro-a" => {
```
jamesmunns: Same note as https://github.com/oxidecomputer/hubris/pull/2696/changes#r4046175729 about using `[&str]::contains()` to make this a little easier to read/maintain.

drv/spartan7-loader/metro-seq/README.md
```diff
@@ -0,0 +1 @@
+THIS IS NOT A REAL METRO IMAGE JUST A COPY OF COSMO AS A PLACEHOLDER
```
jamesmunns, line 1: 🤣

app/metro/base.toml
```diff
@@ -233,16 +230,16 @@ task-slots = [
...
-# stacksize = 3200
-# start = true
-# task-slots = ["sys", "i2c_driver", {spi_front = "spi3_driver"}, "jefe", "packrat", "auxflash", "spartan7_loader", "hf"]
-# uses = ["mmio_sequencer", "mmio_info", "mmio_espi", "mmio_debug_ctrl"]
-# notifications = ["timer", "vcore", "seq-irq"]
+[tasks.metro_seq]
+name = "drv-cosmo-metro-seq-server"
+features = ["h753"]
+priority = 7
+max-sizes = {flash = 131072, ram = 32768 }
```
jamesmunns, line 237: Just noting that these sizes are a little overlarge (probably the same for cosmo), but it's not a big deal.

```
metro_seq        axi_sram   17880  32768  32768
                 flash      43652  49152  131072

metro_seq (stack): 2428 bytes (limit is 3200)
```

app/metro/base.toml
```diff
@@ -252,43 +249,43 @@ notifications = ["timer"]
...
-# start = true
-# uses = ["mmio_spi_nor"]
-# task-slots = ["hash_driver", "spartan7_loader"]
-# stacksize = 4000
-# notifications = ["timer"]
+[tasks.spartan7_loader]
+name = "drv-spartan7-loader"
+features = ["h753"]
+priority = 4
+max-sizes = {flash = 131072, ram = 16384 }
```
jamesmunns, line 256: Same note about sizes, fine to ignore

```
spartan7_loader  axi_sram   7072   7168   16384
                 flash      11248  11264  131072

spartan7_loader (stack): 2412 bytes (limit is 2600)
```

drv/cosmo-metro-seq-server/src/main.rs
jamesmunns, line 1: I guess one question for @nathanaelhuffman:

Are the `NicSm` and related NIC status parts (`nic_raw_status::HwSm`) of the FMC interface going to be the same/similar here?

I guess our management of the NIC sequencing is pretty high level (and the FPGA takes care of it?), and I'm not sure what we're going to need to do to sequence (or monitor) the Versal.

We can probably cfg-gate that, but at least for bringup if we need to stub anything out or perform any basic states, we might want to get a head start on that today.

jamesmunns (approved):
Overall LGTM, a couple of nits, and made a comment for @nathanaelhuffman about my largest anxiety (NIC sequencing).

OK to merge from me, we can iterate on this in follow-up PRs.

PR #2697, opened 2026-09-17
thermal: Retain shutdown logs better

### LINKED ISSUE #2662: thermal: Better logging in "Uncontrollable" condition
Right now when we enter the `Uncontrollable` state in thermal, or `run_control_inner` resolves to `ControlResult::PowerDown`, we don't (semi-)permanently capture the reason for this. It is captured in a `ringbuf`, but in at least one set of customer logs, we have observed this being overwritten by repeated calls to:

```rust
ringbuf_entry!(Trace::PowerDownAt(sys_get_timer().now));
```

Which just fills up the log.

We probably should one or both of:

1. Store the most recent reason for entering Uncontrollable in somewhere that can be observed in a dump (or otherwise obtained)
2. Make the `Trace::PowerDownAt` log a one-shot

jamesmunns:
Moves `PowerDownAt` log from `run_control` to `transition_to_uncontrollable`.

This makes the log a one-shot, stores the shutdown reason into the control structure (to make it durable), and makes the server impl a static, so we can pull the entire structure if we want.

Closes #2662.


jamesmunns:
Tested this with a `sidecar` on Berlin, I was able to retrieve the server impl:

```
jamesmunns@castle ~ $ humility -d ./hubris.core.10 readvar THERMAL_SERVER_IMPL
humility: attached to dump
THERMAL_SERVER_IMPL (0x24021f18) = ClaimOnceCell<core::mem::maybe_uninit::MaybeUninit<task_thermal::ServerImpl<task_thermal::bsp::Bsp>>> {
    taken: AtomicBool {
        v: UnsafeCell<u8> {
            value: 0x1
        }
    },
    cell: UnsafeCell<core::mem::maybe_uninit::MaybeUninit<task_thermal::ServerImpl<task_thermal::bsp::Bsp>>> {
        value: MaybeUninit<task_thermal::ServerImpl<task_thermal::bsp::Bsp>> {
            value: MaybeDangling<task_thermal::ServerImpl<task_thermal::bsp::Bsp>>(ServerImpl<task_thermal::bsp::Bsp> {
                    mode: Auto,
                    control: ThermalControl<task_thermal::bsp::Bsp> {
                        bsp: Bsp {
                            inputs: 0x240217a0 (&mut [task_thermal::bsp::i2c_temp_input::InputChannel; 2]),
                            dynamic_inputs: 0x240217e8 (&mut [task_thermal::control::DynamicInputChannel; 32]),
                            misc_sensors: 0x8097264 (&[task_thermal::bsp::i2c_temp_input::TemperatureSensor; 7]),
                            fctrl_east: Max31790State {
                                max31790: Max31790 {
                                    device: I2cDevice {
                                        task: TaskId(0xb),
                                        controller: I2C1,
                                        port: PortIndex(0x0),
                                        segment: None,
                                        address: 0x23,
                                        component_id: &str {
                                            data_ptr: 0x80972c6 (<UNNAMED>),
                                            length: 0x3
                                        }
                                    }
                                },
                                initialized: true
                            },
                            fctrl_west: Max31790State {
                                max31790: Max31790 {
                                    device: I2cDevice {
                                        task: TaskId(0xb),
                                        controller: I2C3,
                                        port: PortIndex(0x1),
                                        segment: None,
                                        address: 0x20,
                                        component_id: &str {
                                            data_ptr: 0x80972c9 (<UNNAMED>),
                                            length: 0x3
                                        }
                                    }
                                },
                                initialized: true
                            },
                            fans: 0x24021df0 (&mut [task_thermal::control::Fan<drv_i2c_devices::max31790::Fan>; 8]),
                            seq: Sequencer {
                                current_id: Cell<abi::TaskId> {
                                    value: UnsafeCell<abi::TaskId> {
                                        value: TaskId(0x13)
                                    }
                                }
                            },
                            i2c_task: TaskId(0xb),
                            pid_config: PidConfig {
                                zero: 35,
                                gain_p: 1.75,
                                gain_i: 0.0135,
                                gain_d: 0.4,
                                min_output: 0,
                                max_output: 100
                            }
                        },
                        sensor_api: Sensor {
                            current_id: Cell<abi::TaskId> {
                                value: UnsafeCell<abi::TaskId> {
                                    value: TaskId(0xd)
                                }
                            }
                        },
                        ereporter: Ereporter {
                            packrat: Packrat {
                                current_id: Cell<abi::TaskId> {
                                    value: UnsafeCell<abi::TaskId> {
                                        value: TaskId(0x12)
                                    }
                                }
                            },
                            buf: 0x240226f5 (&mut [u8; 105])
                        },
                        target_margin: Celsius(0),
                        state: Running {
                            pid: OneSidedPidState {
                                prev_error: Some(-2.1245003),
                                integral: -0.9320132
                            }
                        },
                        power_mode: PowerBitmask(0b10),
                        pid_config: PidConfig {
                            zero: 35,
                            gain_p: 1.75,
                            gain_i: 0.0135,
                            gain_d: 0.4,
                            min_output: 0,
                            max_output: 100
                        },
                        err_blackbox: ThermalSensorErrors {
                            values: [
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None
                            ],
                            next: 0x0
                        },
                        prev_err_blackbox: ThermalSensorErrors {
                            values: [
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None,
                                None
                            ],
                            next: 0x0
                        },
                        last_pwm: PWMDuty(0x1e),
                        fan_watchdog_configured: true,
                        overheat_timer: None,
                        last_shutdown_reason: None
                    },
                    deadline: 0xecf0,
                    runtime: 0xd
                })
        }
    }
}
```

I'll fix the slightly-larger stack sizes now.

labbott (approved):
LGTM

hawkw:
> Tested this with a `sidecar` on Berlin, I was able to retrieve the server impl:

oh _hell yeah_, you have no idea how happy that makes me.

app/grapefruit/rev-a-ruby.toml
```diff
@@ -34,7 +34,7 @@ name = "task-thermal"
 features = ["grapefruit"]
 priority = 5
 max-sizes = {flash = 32768, ram = 5120 }
-stacksize = 1536
+stacksize = 2000
```
hawkw, line 37: huh, it's a bit strange that this made the stack bigger when we're moving the whole thing into a static? or is it just that CI doesn't build this grapefruit config and it had lagged behind?

hawkw: oh, i see, it's because we're initializing the static by putting the whole server on the stack. gotcha.

task/thermal/src/main.rs
```diff
@@ -168,16 +157,38 @@ counted_ringbuf!(Trace, 32, Trace::None);
...
+    /// sensor.
+    ///
+    /// Note that this may not be an *actual temperature measurement*
+    /// from this sensor. Instead, it is projected from the last successful
+    /// temperature reading, the lag since that measurement was received,
+    /// and the thermal model's slew rate for the component.
+    ///
+    /// This ringbuf entry is always followed by a [`LastActualTemperature`]
+    /// entry, which records the last actual temperature measurement
+    /// reported by the sensor.
```
hawkw: I wonder if this comment ought to move back to the "PowerDownDueTo" ringbuf entry, since it says "this ringbuf entry is ... "

task/thermal/src/main.rs
hawkw: note that this has identical contents to the new `ShutdownReason` struct, I wonder if this ought to also use that? We might need to rename it if it's also used in the `CriticalDueTo` ringbuf, though, since it's not representing a shutdown.

task/thermal/src/main.rs
```diff
@@ -395,27 +404,40 @@ impl<'a, B: control::BspInterface> NotificationHandler for ServerImpl<'a, B> {
...
 
-#[unsafe(export_name = "main")]
-fn main() -> ! {
+/// Initializing the server takes a lot of stack, it's big. Doing this in a
+/// non-inlined function allows the stack analyzer to see that main's stack
+/// frame doesn't retain the large space taken to initialize these structs,
+/// shrinking the total analyzed usage.
+///
+/// Rust loves to inline functions that are private and only ever called once,
+/// so we force inline never.
```
hawkw, line 401: hm, would this be any better if we used an `&'static mut Bsp` and stuck the `Bsp` in its own *separate* static? that way, we could do this without having to have a stack frame big enough to contain both `ThermalControl` *and* the `Bsp`'s fields at the same time, and instead we would only need a stack big enough for whichever one is bigger?

Not sure if this is really worth golfing, just a thought. Since this is explicitly outlined, it probably doesn't impact our max stack depth except, apparently, on grapefruit, so I don't actually know if I care.

hawkw: (another option would be to change this to do more initialization of `ThermalControl` in place, rather than in a `ThermalControl::new`, and just mutate the claimed `&'static mut`...but that's also probably a more invasive change and again, i guess it's only grapefruit-ruby where this bumps the _max_ stack depth, so who cares!)

hawkw (approved):
looks good to me, with a few notes

PR #2700, opened 2026-09-18
gimlet/cosmo-seq: shorten NMI pulse duration to avoid bursts

luqmana:
Both Cosmo and Gimlet use a 25ms delay between toggling the NMI pin. If the host's NMI handler manages to EOI in that 25ms window we'll immediately get another NMI and so on until the pin is released:

Confirm starting NMI count and set `nmi_action` to `NMI_ACTION_IGNORE` which stops us from dropping into the debugger and thus drastically shortens the time before we EOI:
```
BRM42220071 # mdb -ke 'apic_num_nmis/D'
apic_num_nmis:
apic_num_nmis:  0
BRM42220071 # mdb -kwe 'nmi_action/W 2'
nmi_action:     0x3             =       0x2
```

Trigger NMI:
```
$ pfexec humility -t c71 hiffy -c Sequencer.send_hardware_nmi
Sequencer.send_hardware_nmi() => ()
```
We managed to hit the NMI handler 376 times before it stopped. (On a Cosmo I saw ~400).
```
BRM42220071 # mdb -ke 'apic_num_nmis/D'
apic_num_nmis:
apic_num_nmis:  376
```

Removing the delay altogether still triggered the NMI without the subsequent storm and without losing any:
```
$ repeat 1000 pfexec humility -t cosmo-os hiffy -c Sequencer.send_hardware_nmi >/dev/null 2>/dev/null
```
```
271F4FKC # mdb -ke 'apic_num_nmis/D'
apic_num_nmis:
apic_num_nmis:  1000
```


drv/cosmo-seq-server/src/main.rs
```diff
@@ -1054,11 +1054,16 @@ impl idl::InOrderSequencerImpl for ServerImpl {
         &mut self,
         _: &RecvMessage,
     ) -> Result<(), RequestError<core::convert::Infallible>> {
-        // The required length for an NMI pulse is apparently not documented.
+        // The AMD PPRs and SP5 motherboard design guide do not specify a
+        // minimum duration for an NMI pulse. On the CPU side, we know the pin
+        // enters the FCH GPIO block and from there feeds the IOHC which turns
+        // it into an NMI (or SYNCFLOOD depending on configuration).
         //
-        // Let's try 25 ms!
```
hawkw, line 1059: Let's try a lot less than that!

hawkw (approved):
Seems fine to me!

Sorry you *just* missed the SP 1.80 release, but if we think this should get in for R23, we can cut another?

jamesmunns:
Hey @luqmana, is there a spec on what the *minimum* required pulse width is to trigger an NMI? 25ms definitely seems like way too much, but it would be good to make sure we're not sometimes working really fast (or in the future if we reduce syscall costs) and emit a pulse that's not long enough to trigger an NMI.

jamesmunns:
> is there a spec on what the _minimum_ required pulse width is to trigger an NMI? 25ms definitely seems like way too much

Oh, you answered this, sorry:

```rust
// The AMD PPRs and SP5 motherboard design guide do not specify a
// minimum duration for an NMI pulse. On the CPU side, we know the pin
```

PR #2703, opened 2026-09-20
abi/kern/userlib: Use `usize`-like data for addr+len

jamesmunns:
This is the first PR in a series to be able to potentially run the kernel and tasks on a 64-bit host for testing purposes.

This PR switches some things in abi/kern/userlib to use `usize` instead of `u32`. On 32-bit targets, this is not a functional change, and in the first commit (before changing the types), I added some static asserts, which continue passing after the first change.


jamesmunns:

sys/abi/src/lib.rs
```diff
@@ -181,6 +181,160 @@ pub struct Interrupt {
...
+
+impl serde::Serialize for Addr {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        // This is icky! However, kipc uses `ssmarshal` for serialization, which
+        // makes the serialization of addresses a kipc binary concern, and we
+        // don't want to use `usize` as our wire format as serde treats that
+        // as a u64 in all cases, and we've always serialized addresses as u32s.
```
hawkw, line 220: SIGH

sys/kern/src/arch/arm_m.rs
```diff
@@ -393,10 +393,12 @@ pub struct RegionDescExt {
 
 #[cfg(any(armv6m, armv7m))]
 pub const fn compute_region_extension_data(
-    base: u32,
-    size: u32,
+    base: usize,
```
hawkw, line 396: should base be an addr?

sys/kern/src/arch/arm_m.rs
```diff
@@ -543,10 +545,12 @@ pub struct RegionDescExt {
 
 #[cfg(armv8m)]
 pub const fn compute_region_extension_data(
-    base: u32,
-    size: u32,
+    base: usize,
```
hawkw, line 548: should base be an addr?

sys/kern/src/descs.rs
```diff
@@ -94,31 +94,61 @@ pub struct RegionDesc {
     /// Address of start of region. The platform likely has alignment
     /// requirements for this; it must meet them. (For example, on ARMv7-M, it
     /// must be naturally aligned for the size.)
-    pub base: u32,
+    pub base: usize,
```
hawkw, line 97: should base be an addr?

sys/kern/src/descs.rs
```diff
@@ -94,31 +94,61 @@ pub struct RegionDesc {
...
     /// Flags describing what can be done with this region.
     pub attributes: RegionAttributes,
 }
 
+// The `RegionDesc` type is fundamental to the ABI! We very much want to ensure
+// that it never breaks. This const assert checks that.
+//
+// It's possible in the future we want `RegionDesc` to be arch-specific, but for
+// now ensure that all 32-bit targets share the same ABI qualities.
+#[cfg(target_pointer_width = "32")]
```
hawkw, line 111: i wonder if perhaps all of this ought to just be in the arch module?

sys/abi/src/lib.rs
mkeeter: We could always switch to `hubpack`!  Last time I tried, there was a noticeable size regression, but that was literally years ago.

jamesmunns: I think if I made a jump, I'd aim for `postcard` or [`nprpc`](https://docs.rs/nprpc), but I would probably want to do that *after* we have something like #2706 to be able to smoke test it!

PR #2704, opened 2026-09-20
kernel: Introduce `Arch` trait

jamesmunns:
This is the second PR in a series to be able to potentially run the kernel and tasks on a 64-bit host for testing purposes. This is currently based on #2703 

This PR introduces an `Arch` trait, which captures the architecture-specific methods required for a new architecture impl. This encapsulates the behaviors that the rest of the kernel uses for scheduling and other behavior.

This follows the pattern we've started using for BSP impls lately: We define a trait, each BSP (or Arch in this case) implements it, and we expose the One Selected Impl using `cfg`s as a commonly named type (here: `Current`) which the rest of the crate uses.

This one might be worth review commit-by-commit, as the total diff is a bit large, but the individual commits are each pretty straightforward.


labbott:
More thoughts later but I'd like us to have an eye towards something that could let us have RISC-V support too (I think this should be sufficient but always watch out for foot guns or something)

jamesmunns:

PR #2705, opened 2026-09-21
userlib: Introduce `Arch` trait

jamesmunns:
This is the third PR in a series to be able to potentially run the kernal and tasks on a 64-bit host for testing purposes. This is currently based on #2704 (and #2703).

This PR introduces an `Arch` trait for `userlib`, similar to the `Arch` trait introduced for the kernel in #2704.

This follows the pattern we've started using for BSP impls lately: We define a trait, each BSP (or Arch in this case) implements it, and we expose the One Selected Impl using cfgs as a commonly named type (here: Current) which the rest of the crate uses.

Like #2703, this one might be worth review commit-by-commit, as the total diff is a bit large, but the individual commits are each pretty straightforward.


jamesmunns:

